PR #3801, opened 2019-07-12
GEODE-6897: async operation framework for ClusterManagementService

onichols-pivotal:
As a prerequisite to adding long-running operations like `rebalance` to `ClusterManagementService`, a new framework is needed.

This PR introduces a new `startOperation` API in `ClusterManagementService` and a registry of supported `ClusterManagementOperation`s and performers in `OperationManager`.  The Java API returns a `CompletableFuture` for callers to append next stages after their long-running operations complete, while the REST interface returns a status URI for further monitoring.


jinmeiliao:
Actually only `startOperation` is added to the API, the `checkStatus` is only available on the locator side and is not a public api.

geode-management/src/main/java/org/apache/geode/management/api/ClientAsyncOperationResult.java
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  @Override
+  public boolean isCancelled() {
+    return false;
+  }
+
+  @Override
+  public boolean isDone() {
+    ClusterManagementOperationResult<V> result = requestStatus();
```
jinmeiliao: I was thinking about a way to avoid making the extra trip to the server to get the completed result. We could save the result here as a member variable to get it after isDone returns true.

onichols-pivotal: updated to cache the first "done" result as suggested

onichols-pivotal:
@jinmei thanks, I updated the PR summary to remove mention of `checkStatus` since it is used only internally by the REST controller, not part of the public Java API.

geode-management/src/main/java/org/apache/geode/management/runtime/RebalanceInfo.java
```diff
@@ -0,0 +1,36 @@
...
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package org.apache.geode.management.runtime;
+
+import org.apache.geode.management.api.JsonSerializable;
+
+public class RebalanceInfo implements JsonSerializable {
```
jinmeiliao: this class could be renamed to be more general, like StringInfo

geode-management/src/main/java/org/apache/geode/management/operation/RebalanceOperation.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+
+import org.apache.geode.management.api.AsyncOperation;
+import org.apache.geode.management.runtime.RebalanceInfo;
+
+@JsonIgnoreProperties(value = {"uri"}, allowGetters = true)
```
jinmeiliao: found out another way to achieve this without having to annotate each and every implementation: On the `RestfulEndpoint.getUri` method, add an annotation: `@JsonProperty(access = JsonProperty.Access.READ_ONLY)`

dschneider-pivotal:
@kirklund, could you add more information about using the CompletableFuture class vs. the Future interface?
We discussed using the Future interface and were concerned about its "cancel" method. The concern was that the geode core may not have support for cancelling the long running operation. So the locator could stop waiting for the op to finish but the op itself would still be running. Would you recommend that in that case we implement cancel to throw an unsupported op exception?

geode-management/src/main/java/org/apache/geode/management/runtime/RebalanceInfo.java
onichols-pivotal: I suspect `rebalance` will return a more complex result than a simple String, but it was stubbed out this way since this PR is focused on proving out the framework, not implementing the actual rebalance command yet.

kirklund:
@dschneider-pivotal Dale encouraged me to use CompletableFuture in a similar requirement for server startup notification which involves multiple async operations. It's still very much a work-in-progress, but you can check out an early draft of our code (with bugs) in #3799 (you can view it there even if we close the draft PR). We found a [great YouTube video by Angelika Langer that includes some great coverage on CompletableFuture ](https://www.youtube.com/watch?v=Q_0_1mKTlnY) that I recommend checking out. 

I don't know enough about cancel or why you need to prevent it. I guess if you evaluate CompletableFuture and determine it won't work because of how or if it supports cancellation, then I'll ultimately approve not using it. I would think that Cache close or JVM shutdown would require cancellation support. Or you could maybe subclass CompletableFuture to override the cancel methods.

demery-pivotal:
> We discussed using the Future interface…
Do you mean that you discussed using `CompletableFuture`?

> … and were concerned about its "cancel" method. The concern was that the geode core may not have support for cancelling the long running operation. So the locator could stop waiting for the op to finish but the op itself would still be running. Would you recommend that in that case we implement cancel to throw an unsupported op exception?

The `cancel()` method is a bit of a wart in `CompletableFuture`, which has the method only because it inherits it from `Future`. As the `CompletableFuture` javadoc says: "Method `cancel` has the same effect as `completeExceptionally(new CancellationException())`." That is, it marks the `CompletableFuture` as completed with a cancellation exception, and does nothing else. In particular: It does not actually cancel the task itself. Indeed, it has absolutely no effect on the task.

This behavior seems likely to surprise callers. So if we really don't want to cancel the task, then override `cancel()` to throw unsupported op.

geode-management/src/main/java/org/apache/geode/management/operation/RebalanceOperation.java
onichols-pivotal: good find, I have made the change

geode-management/src/main/java/org/apache/geode/management/api/ClusterManagementOperationStatusResult.java
```diff
@@ -0,0 +1,43 @@
...
+ * the License.
+ */
+package org.apache.geode.management.api;
+
+
+
+import org.apache.geode.annotations.Experimental;
+
+@Experimental
+public class ClusterManagementOperationStatusResult<V extends JsonSerializable>
```
jinmeiliao: Now that you've used this to convey the completed result, I think we can combine the ClusterManagementOperationResult with the ClusterManagementOperationStatusResult.

onichols-pivotal: it's not obvious to me how you would combine them.  one is Future<V> and one is V.  We had them combined before and it necessitated the extra CompletedAsyncOperationResult with the hack to set it to null in one case.  We still have the same number of classes now, but less indirection and the null hack is gone.

onichols-pivotal:
reworked to return CompletableFuture

geode-core/src/main/java/org/apache/geode/management/internal/api/LocatorClusterManagementService.java
```diff
@@ -65,19 +71,21 @@
...
   private Map<Class, ConfigurationManager> managers;
   private Map<Class, ConfigurationValidator> validators;
+  private AsyncExecutorManager executorManager;
   private MemberValidator memberValidator;
   private CacheElementValidator commonValidator;
 
   public LocatorClusterManagementService(InternalCache cache,
       ConfigurationPersistenceService persistenceService) {
-    this(persistenceService, new HashMap<>(), new HashMap<>(), null, null);
+    this(persistenceService, new HashMap<>(), new HashMap<>(), null, null, null);
```
dschneider-pivotal: After this constructor completes are these HashMaps ever modified? If not it seems like this code should be refactored to fill the maps before passing them in. The other constructor that takes the Map params could then wrap then with Collections.unmodifiableMap.
If they can be modified later then we need to address thread-safety on them.

geode-management/src/main/java/org/apache/geode/cache/configuration/GatewayReceiverConfig.java
```diff
@@ -61,7 +60,6 @@
  */
 @XmlAccessorType(XmlAccessType.FIELD)
 @XmlType(name = "", propOrder = {"gatewayTransportFilters"})
-@JsonIgnoreProperties(value = {"uri"}, allowGetters = true)
```
dschneider-pivotal: Were these changes related to async operation framework or a different bug you found?

geode-management/src/main/java/org/apache/geode/management/api/ClusterManagementOperationStatusResult.java
dschneider-pivotal: I only see this class used by internal classes. So should this class be in an internal package?

geode-management/src/main/java/org/apache/geode/management/api/ClusterManagementOperationResult.java
```diff
@@ -0,0 +1,47 @@
...
+package org.apache.geode.management.api;
+
+import java.util.concurrent.CompletableFuture;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+
+import org.apache.geode.annotations.Experimental;
+
+@Experimental
+public class ClusterManagementOperationResult<V extends JsonSerializable>
```
dschneider-pivotal: This looks like a class that should instead be an interface. I don't think we want users of the java api to ever create instances of it. The problem is ClusterManagementResult is a class that should also be an interface.

geode-management/src/main/java/org/apache/geode/management/api/ClusterManagementOperationStatusResult.java
dschneider-pivotal: And this class can be internal but ClusterManagementOperationResult needs to be in the external api

geode-management/src/main/java/org/apache/geode/management/api/ClusterManagementResult.java
```diff
@@ -112,7 +116,8 @@ public void setUri(String uri) {
 
   @JsonIgnore
   public boolean isSuccessful() {
-    return statusCode == StatusCode.OK;
+    return statusCode == StatusCode.OK || statusCode == StatusCode.ACCEPTED
```
dschneider-pivotal: Why are ACCEPTED and IN_PROGRESS considered successful? Does "isSuccessful" just mean it has not failed yet?

geode-management/src/main/java/org/apache/geode/management/internal/FutureProxy.java
```diff
@@ -0,0 +1,132 @@
...
+
+import org.springframework.web.client.RestTemplate;
+
+import org.apache.geode.annotations.Experimental;
+import org.apache.geode.management.api.ClusterManagementOperationStatusResult;
+import org.apache.geode.management.api.ClusterManagementResult;
+import org.apache.geode.management.api.JsonSerializable;
+
+@Experimental
+public class FutureProxy<V extends JsonSerializable> implements Future<V> {
```
jinmeiliao: Maybe have this directly extends CompletableFuture<V>

geode-management/src/main/java/org/apache/geode/management/internal/FutureProxy.java
```diff
@@ -0,0 +1,132 @@
...
+   */
+  @SuppressWarnings("unchecked")
+  private ClusterManagementOperationStatusResult<V> requestStatus() {
+    return restTemplate
+        .getForEntity(uri, ClusterManagementOperationStatusResult.class)
+        .getBody();
+  }
+
+  @Override
+  public boolean cancel(boolean mayInterruptIfRunning) {
```
jinmeiliao: Maybe not worrying about `cancel` for now.

geode-management/src/main/java/org/apache/geode/management/api/ClusterManagementResult.java
```diff
@@ -47,6 +47,10 @@
     // configuration.
     // the operation is still deemed unsuccessful.
     FAIL_TO_PERSIST,
+    // async operation launched successfully
+    ACCEPTED,
```
dschneider-pivotal: Why do we have both ACCEPTED and IN_PROGRESS? I see ACCEPTED set once we have submitted the async op.
But IN_PROGRESS is only set if we call checkStatus. Would it be better to get rid of ACCEPTED and set IN_PROGRESS only when ACCEPTED is currently set?

geode-management/src/main/java/org/apache/geode/management/internal/FutureProxy.java
onichols-pivotal: with CompletableFuture's default implementation, if cancel is called we would continue polling the REST status endpoint even after the CompletableFuture returned exceptional state due to cancellation.  I don't like the idea of keeping that polling loop running if user has explicitly asked to stop waiting...

geode-management/src/main/java/org/apache/geode/management/internal/FutureProxy.java
```diff
@@ -0,0 +1,132 @@
...
+
+import org.apache.geode.annotations.Experimental;
+import org.apache.geode.management.api.ClusterManagementOperationStatusResult;
+import org.apache.geode.management.api.ClusterManagementResult;
+import org.apache.geode.management.api.JsonSerializable;
+
+@Experimental
+public class FutureProxy<V extends JsonSerializable> implements Future<V> {
+  static final int POLL_INTERVAL = 100; // millis between http status checks
+  static final ExecutorService pool = Executors.newFixedThreadPool(10);
```
dschneider-pivotal: Can we make this pool non-static?

dschneider-pivotal: Should LoggingExecutors be used here?

geode-management/src/main/java/org/apache/geode/management/runtime/RebalanceInfo.java
dschneider-pivotal: Would RebalanceResult be a better name?

geode-management/src/main/java/org/apache/geode/management/internal/FutureProxy.java
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  @Override
+  public boolean isCancelled() {
+    return cancelled;
+  }
+
+  @Override
+  public boolean isDone() {
+    if (completedResult != null) {
```
dschneider-pivotal: I think completedResult should be an AtomicReference so that this isDone check can be made thread safe with a concurrent cancel call.

geode-management/src/main/java/org/apache/geode/management/internal/FutureProxy.java
onichols-pivotal: done

geode-management/src/main/java/org/apache/geode/management/internal/FutureProxy.java
onichols-pivotal: this part of the code has been reworked, but definitely still need to consider concurrency...

geode-management/src/main/java/org/apache/geode/management/api/ClusterManagementOperationStatusResult.java
onichols-pivotal: explored some options but decided there is no viable way to combine them without CompletableFuture internals getting inadvertently serialized as well

geode-management/src/main/java/org/apache/geode/management/api/ClusterManagementOperationResult.java
```diff
@@ -0,0 +1,50 @@
...
+    super(statusCode, message);
+  }
+
+  @JsonIgnore
+  private CompletableFuture<V> operationResult;
+
+  @JsonIgnore
+  public CompletableFuture<V> getFutureResult() {
+    if (operationResult instanceof CompletableFutureProxy)
+      ((CompletableFutureProxy) operationResult).startPolling();
```
dschneider-pivotal: Could the call of startPolling be moved to a place that knows it is a CompletableFutureProxy?

onichols-pivotal: getFutureResult on the locator returns a CompletableFuture, while getFutureResult on the client returns the CompletableFutureProxy.  The only way we know if the user cares about the result is if they call getFutureResult().  The only other option to is to unconditionally start polling...

geode-core/src/main/java/org/apache/geode/management/internal/operation/OperationPerformer.java
```diff
@@ -0,0 +1,24 @@
...
+ * the License.
+ */
+package org.apache.geode.management.internal.operation;
+
+import org.apache.geode.annotations.Experimental;
+import org.apache.geode.management.api.AsyncOperation;
+import org.apache.geode.management.api.JsonSerializable;
+
+@Experimental
+public interface OperationPerformer<A extends AsyncOperation<V>, V extends JsonSerializable> {
```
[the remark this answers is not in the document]
onichols-pivotal: For naming consistency, that would seem reasonable.  However there is nothing async about OperationPerformers, they implement a strictly synchronous (blocking) `perform` method.  The async-ness is added in by the surrounding framework...

geode-core/src/main/java/org/apache/geode/management/internal/operation/AsyncExecutorManager.java
```diff
@@ -0,0 +1,82 @@
...
+import org.apache.logging.log4j.util.Strings;
+
+import org.apache.geode.annotations.Experimental;
+import org.apache.geode.management.api.AsyncOperation;
+import org.apache.geode.management.api.JsonSerializable;
+import org.apache.geode.management.operation.RebalanceOperation;
+
+@Experimental
+public class AsyncExecutorManager {
+  private Map<Class<? extends AsyncOperation>, OperationPerformer> performers;
```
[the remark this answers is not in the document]
onichols-pivotal: there is no perfect solution in Java 8 for heterogenous collections.  I tried a variety of permutations and all but this one caused problems.  I'm certainly open to a better suggestion if one exists.

geode-core/src/main/java/org/apache/geode/management/internal/operation/AsyncExecutorManager.java
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  public <A extends AsyncOperation<V>, V extends JsonSerializable> CompletableFuture<V> submit(
+      A op) {
+    if (!Strings.isBlank(op.getId()))
+      throw new IllegalArgumentException(
+          String.format("Operation type %s should not supply its own id",
+              op.getClass().getSimpleName()));
+    op.setId(UUID.randomUUID().toString());
+    OperationPerformer<A, V> performer = getPerformer(op);
```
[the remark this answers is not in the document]
onichols-pivotal: I find clear, up-front error checking easier to follow.  What is the benefit of obfuscating the error-checking?

geode-management/src/main/java/org/apache/geode/management/internal/CompletableFutureProxy.java
```diff
@@ -0,0 +1,71 @@
...
+  private ScheduledFuture<?> scheduledFuture;
+
+  public CompletableFutureProxy(RestTemplate restTemplate, String uri) {
+    this.restTemplate = restTemplate;
+    this.uri = uri;
+  }
+
+  public void startPolling() {
+    scheduledFuture = pool.scheduleWithFixedDelay(() -> {
+      if (isDone()) {
```
dschneider-pivotal: why do we call isDone here and then also have a done check a few lines later based on status?

geode-management/src/main/java/org/apache/geode/management/internal/CompletableFutureProxy.java
```diff
@@ -0,0 +1,71 @@
...
+      }
+    }, 0, POLL_INTERVAL, TimeUnit.MILLISECONDS);
+  }
+
+  /**
+   * this should be the only method to make the request to the locator to check the status of the
+   * operation
+   */
+  @SuppressWarnings("unchecked")
+  private ClusterManagementOperationStatusResult<V> requestStatus() {
```
dschneider-pivotal: How does this method behave if the locator goes down? Does it throw an exception?

geode-management/src/main/java/org/apache/geode/management/internal/CompletableFutureProxy.java
```diff
@@ -0,0 +1,71 @@
...
+import org.apache.geode.annotations.Experimental;
+import org.apache.geode.management.api.ClusterManagementOperationStatusResult;
+import org.apache.geode.management.api.ClusterManagementResult;
+import org.apache.geode.management.api.JsonSerializable;
+
+@Experimental
+public class CompletableFutureProxy<V extends JsonSerializable> extends CompletableFuture<V> {
+  static final int POLL_INTERVAL = 100; // millis between http status checks
+  static final ScheduledExecutorService pool = Executors.newScheduledThreadPool(1);
+  private RestTemplate restTemplate;
```
dschneider-pivotal: restTemplate and uri should be final

geode-management/src/main/java/org/apache/geode/management/internal/CompletableFutureProxy.java
```diff
@@ -0,0 +1,71 @@
...
+
+import org.apache.geode.annotations.Experimental;
+import org.apache.geode.management.api.ClusterManagementOperationStatusResult;
+import org.apache.geode.management.api.ClusterManagementResult;
+import org.apache.geode.management.api.JsonSerializable;
+
+@Experimental
+public class CompletableFutureProxy<V extends JsonSerializable> extends CompletableFuture<V> {
+  static final int POLL_INTERVAL = 100; // millis between http status checks
+  static final ScheduledExecutorService pool = Executors.newScheduledThreadPool(1);
```
dschneider-pivotal: can we have this pool live somewhere with an instance reference so that we will not add a static pool?

geode-management/src/main/java/org/apache/geode/management/internal/CompletableFutureProxy.java
onichols-pivotal: two different checks.  `isDone` may be caused externally by user cancellation, completing, or exceptional completion.  `done` means we are ready to complete it.

geode-management/src/main/java/org/apache/geode/management/api/ClusterManagementOperationResult.java
dschneider-pivotal: Do you see a problem with unconditionally calling startPolling? Do you see a use case in which you would start the async op but never want to check if it completed.

geode-core/src/main/java/org/apache/geode/management/internal/operation/AsyncExecutorManager.java
```diff
@@ -0,0 +1,82 @@
...
+import org.apache.commons.collections.map.LRUMap;
+import org.apache.logging.log4j.util.Strings;
+
+import org.apache.geode.annotations.Experimental;
+import org.apache.geode.management.api.AsyncOperation;
+import org.apache.geode.management.api.JsonSerializable;
+import org.apache.geode.management.operation.RebalanceOperation;
+
+@Experimental
+public class AsyncExecutorManager {
```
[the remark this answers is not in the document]
onichols-pivotal: renamed to OperationManager

geode-management/src/main/java/org/apache/geode/management/api/ClusterManagementResult.java
onichols-pivotal: isSuccessful just means not-an-error

geode-management/src/main/java/org/apache/geode/management/api/ClusterManagementResult.java
onichols-pivotal: they need to map to different HTTP STATUSes

geode-management/src/main/java/org/apache/geode/management/internal/FutureProxy.java
onichols-pivotal: good catch, moved to LocatorCMS scope

onichols-pivotal: not here, because this is in management module which doesn't have access to geode-core.  But on the locator side, YES, I have made that change.

geode-management/src/main/java/org/apache/geode/management/runtime/RebalanceInfo.java
onichols-pivotal: yes, changed

geode-management/src/main/java/org/apache/geode/management/api/ClusterManagementOperationResult.java
onichols-pivotal: it's an ugly optimization but....

geode-management/src/main/java/org/apache/geode/management/internal/CompletableFutureProxy.java
onichols-pivotal: good catch!  added logic to tolerate intermittent network errors on check calls, but return failure if >3 consecutive errors (suggesting the locator is actually down or major network outage has happened)

geode-management/src/main/java/org/apache/geode/management/internal/CompletableFutureProxy.java
onichols-pivotal: done!

geode-management/src/main/java/org/apache/geode/cache/configuration/GatewayReceiverConfig.java
onichols-pivotal: kindof-unrelated cleanup, we found a way to annotate this in one place instead of having to add it to every new class, including some new classes in this PR.  went ahead and cleaned up the existing classes too.

geode-core/src/main/java/org/apache/geode/management/internal/operation/AsyncExecutorManager.java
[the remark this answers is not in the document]
onichols-pivotal: done

geode-core/src/main/java/org/apache/geode/management/internal/operation/AsyncExecutorManager.java
```diff
@@ -0,0 +1,82 @@
...
+          String.format("Operation type %s should not supply its own id",
+              op.getClass().getSimpleName()));
+    op.setId(UUID.randomUUID().toString());
+    OperationPerformer<A, V> performer = getPerformer(op);
+    if (performer == null) {
+      throw new IllegalArgumentException(String.format("Operation type %s is not supported",
+          op.getClass().getSimpleName()));
+    }
+
+    CompletableFuture<V> future = CompletableFuture.supplyAsync(() -> performer.perform(op));
```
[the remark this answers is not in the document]
onichols-pivotal: done, but note that LoggingExecutors.newThreadOnEachExecute is an `Executor` not an `ExecutorService` so we lose the ability to shut down cleanly if operations are in progress...

geode-core/src/main/java/org/apache/geode/management/internal/operation/OperationPerformer.java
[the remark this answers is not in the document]
onichols-pivotal: The performer is just the unit of work, it has no concept that it is being managed.  Not sure that `Operation` where getId currently resides is the right place either tho...

geode-core/src/main/java/org/apache/geode/management/internal/operation/RebalanceOperationPerformer.java
```diff
@@ -0,0 +1,32 @@
...
+ * the License.
+ */
+package org.apache.geode.management.internal.operation;
+
+import org.apache.geode.annotations.Experimental;
+import org.apache.geode.management.operation.RebalanceOperation;
+import org.apache.geode.management.runtime.RebalanceInfo;
+
+@Experimental
+public class RebalanceOperationPerformer
```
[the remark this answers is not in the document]
onichols-pivotal: Should be more logical with some renames now, `Operation` is the base class of `RebalanceOperation` (i.e. the public struct that specifies "what" to do plus any parameters) and `OperationPerformer` is the interface for the internal code that actually performs the work.

geode-core/src/main/java/org/apache/geode/management/internal/api/LocatorClusterManagementService.java
```diff
@@ -65,19 +71,21 @@
...
 import org.apache.geode.management.internal.exceptions.EntityNotFoundException;
+import org.apache.geode.management.internal.operation.AsyncExecutorManager;
 import org.apache.geode.management.runtime.RuntimeInfo;
 
 public class LocatorClusterManagementService implements ClusterManagementService {
   private static final Logger logger = LogService.getLogger();
   private ConfigurationPersistenceService persistenceService;
   private Map<Class, ConfigurationManager> managers;
   private Map<Class, ConfigurationValidator> validators;
+  private AsyncExecutorManager executorManager;
```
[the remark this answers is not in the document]
onichols-pivotal: done

geode-core/src/main/java/org/apache/geode/management/internal/api/LocatorClusterManagementService.java
onichols-pivotal: the known ones are hardcoded for now, but extensions could register more so we don't want to make them immutable.  down the road when Geode is more modular, each module would register its own configurables and operations.

geode-management/src/main/java/org/apache/geode/management/api/ClusterManagementOperationStatusResult.java
onichols-pivotal: moved, good catch

geode-management/src/main/java/org/apache/geode/management/api/ClusterManagementOperationResult.java
onichols-pivotal: There's not much benefit of making an interface here, it's all internal and Jackson needs some concrete class to deserialize to.

geode-management/src/main/java/org/apache/geode/management/internal/CompletableFutureProxy.java
onichols-pivotal: done

geode-core/src/main/java/org/apache/geode/management/internal/operation/AsyncExecutorManager.java
```diff
@@ -0,0 +1,82 @@
...
+
+  public AsyncExecutorManager() {
+    this(10);
+  }
+
+  public AsyncExecutorManager(int historySize) {
+    history = new LRUMap(historySize);
+
+    // initialize the list of operation performers
+    performers = new HashMap<>();
```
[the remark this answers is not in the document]
onichols-pivotal: changed to ConcurrentHashmap

geode-core/src/main/java/org/apache/geode/management/internal/operation/AsyncExecutorManager.java
```diff
@@ -0,0 +1,82 @@
...
+public class AsyncExecutorManager {
+  private Map<Class<? extends AsyncOperation>, OperationPerformer> performers;
+  private Map<String, CompletableFuture> history;
+
+  public AsyncExecutorManager() {
+    this(10);
+  }
+
+  public AsyncExecutorManager(int historySize) {
+    history = new LRUMap(historySize);
```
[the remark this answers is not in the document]
onichols-pivotal: wrapped in Collections.synchronizedMap

geode-management/src/main/java/org/apache/geode/management/api/ClusterManagementOperationResult.java
onichols-pivotal: Let's explore this idea in a separate PR.  Interfaces are a good practice for "future-proofing" and might be worth considering across the Cluster Management API

geode-management/src/main/java/org/apache/geode/management/internal/FutureProxy.java
onichols-pivotal: added a synchronized where it was missing on `startPolling`

geode-core/src/main/java/org/apache/geode/management/internal/operation/RebalanceOperationPerformer.java
```diff
@@ -0,0 +1,32 @@
...
+    implements OperationPerformer<RebalanceOperation, RebalanceResult> {
+  @Override
+  public RebalanceResult perform(RebalanceOperation operation) {
+    try {
+      Thread.sleep(100);
+    } catch (InterruptedException ignore) {
+    }
+    return new RebalanceResult("fake success");
```
kirklund: Is this product code or test code in src/main?

geode-core/src/test/java/org/apache/geode/management/internal/operation/OperationHistoryManagerTest.java
```diff
@@ -0,0 +1,111 @@
...
+
+import java.util.concurrent.CompletableFuture;
+
+import org.junit.Before;
+import org.junit.Test;
+
+import org.apache.geode.management.api.ClusterManagementOperation;
+
+public class OperationHistoryManagerTest {
+  OperationHistoryManager history;
```
kirklund: I think we should narrow all fields and methods as private as possible, even in tests.

geode-core/src/main/java/org/apache/geode/management/internal/operation/RebalanceOperationPerformer.java
dschneider-pivotal: In a future PR we plan on the first async operation we add being rebalance. The Rebalance classes you see are just place holders.
@onichols-pivotal maybe it would be better to get rid of the fake Rebalance classes in the product and instead prove in a test that the framework allows the addition of an operation and performer. I think we saw that at least one of the managers was missing public methods (i.e. it had to add the rebalance performer in the manager's constructor). Having a public method on th manager that lets you register the op/performer seems better (for future extensions) and you could move these rebalance classes in the pr to a test.

kirklund: Remember to reset the interrupt bit on the Thread if InterruptedException was caught (assuming it matters for this Thread).

onichols-pivotal: removed from this PR :)

geode-core/src/test/java/org/apache/geode/management/internal/operation/OperationHistoryManagerTest.java
onichols-pivotal: done

geode-core/src/main/java/org/apache/geode/management/internal/api/LocatorClusterManagementService.java
```diff
@@ -358,11 +371,51 @@ public LocatorClusterManagementService(ConfigurationPersistenceService persisten
...
+    ClusterManagementOperationStatusResult<V> result =
+        new ClusterManagementOperationStatusResult<>();
+    if (!status.isDone()) {
+      result.setStatus(ClusterManagementResult.StatusCode.IN_PROGRESS, "in progress");
+    } else {
+      try {
+        result.setResult(status.get());
+        result.setStatus(ClusterManagementResult.StatusCode.OK, "finished successfully");
+      } catch (InterruptedException | ExecutionException e) {
+        throw new RuntimeException(e);
```
kirklund: Just a quick comment here... you should probably reset the interrupt bit iif InterruptedException, before rethrowing it.

geode-management/src/main/java/org/apache/geode/management/api/ClusterManagementOperation.java
```diff
@@ -0,0 +1,36 @@
...
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.apache.geode.management.api;
+
+import org.apache.geode.lang.Identifiable;
+
+/**
+ *
+ * @param <V> the result type of the operation
```
kirklund: I think we should fill in the Javadoc for any User API classes. Second preference would be to delete any incomplete Javadocs if we think it's unnecessary to write Javadocs on a class.

geode-core/src/main/java/org/apache/geode/management/internal/api/LocatorClusterManagementService.java
onichols-pivotal: InterruptedException can't actually be thrown by this `get()` since we only call it when `isDone()` is already true.  But, I will update the catch block anyway.

geode-core/src/test/java/org/apache/geode/management/internal/operation/OperationManagerTest.java
```diff
@@ -0,0 +1,122 @@
...
+import java.util.concurrent.CountDownLatch;
+
+import org.junit.Before;
+import org.junit.Test;
+
+import org.apache.geode.management.api.ClusterManagementOperation;
+import org.apache.geode.management.api.JsonSerializable;
+
+public class OperationManagerTest {
+  OperationManager executorManager;
```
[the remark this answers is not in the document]
onichols-pivotal: Will fix.  Is there an IntelliJ tool to find all variables that can be made private?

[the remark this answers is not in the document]
kirklund: Yep, there's an inspection you can turn on.

geode-management/src/main/java/org/apache/geode/management/api/ClusterManagementOperation.java
onichols-pivotal: added javadocs, thanks for the reminder

PR #3805, opened 2019-07-16
GEODE-6947: separate Pdx configuration from xml domain object

jinmeiliao:

### For all changes:

### Note:


jinmeiliao:
@jmelchio

geode-core/src/main/java/org/apache/geode/management/internal/configuration/mutators/PdxManager.java
```diff
@@ -19,31 +19,34 @@
 import java.util.List;
 
 import org.apache.geode.cache.configuration.CacheConfig;
-import org.apache.geode.cache.configuration.PdxType;
+import org.apache.geode.management.configuration.Pdx;
+import org.apache.geode.management.internal.configuration.converters.PdxConverter;
+
+public class PdxManager implements ConfigurationManager<Pdx> {
+  private PdxConverter pdxConverter = new PdxConverter();
```
dschneider-pivotal: make this field final

geode-management/src/main/java/org/apache/geode/management/configuration/Pdx.java
```diff
@@ -0,0 +1,97 @@
...
+import org.apache.geode.cache.configuration.CacheElement;
+import org.apache.geode.management.api.CorrespondWith;
+import org.apache.geode.management.api.RestfulEndpoint;
+import org.apache.geode.management.runtime.RuntimeInfo;
+
+public class Pdx extends CacheElement implements RestfulEndpoint, CorrespondWith<RuntimeInfo> {
+  public static final String PDX_ID = "PDX";
+  public static final String PDX_ENDPOINT = "/configurations/pdx";
+
+  protected Boolean readSerialized;
```
dschneider-pivotal: What is the reason for the Object "Boolean" here instead of the primitive "boolean"? This trickles down to the gettors and settors.

geode-management/src/main/java/org/apache/geode/management/configuration/Pdx.java
```diff
@@ -0,0 +1,97 @@
...
+
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+
+import org.apache.geode.cache.configuration.CacheElement;
+import org.apache.geode.management.api.CorrespondWith;
+import org.apache.geode.management.api.RestfulEndpoint;
+import org.apache.geode.management.runtime.RuntimeInfo;
+
+public class Pdx extends CacheElement implements RestfulEndpoint, CorrespondWith<RuntimeInfo> {
```
dschneider-pivotal: At some point we should add javadoc comments for all these external apis.

geode-management/src/main/java/org/apache/geode/management/configuration/Pdx.java
dschneider-pivotal: It seems like "private" would be better than "protected" on these fields. Any subclass can use the gettors and settors. Did you have any reason for using "protected" instead of "private"?

geode-management/src/main/java/org/apache/geode/management/configuration/ClassName.java
```diff
@@ -13,7 +13,7 @@
  * the License.
  */
 
-package org.apache.geode.management.internal.cli.domain;
+package org.apache.geode.management.configuration;
```
dschneider-pivotal: It looks like you are making this previously internal class external. It looks like it would need to have some additional cleanup. It has the generic parameter <T> that is never used.  It is just Serializable. Its javadocs talk about gfsh.
It is different from the other CacheElement subclasses. For example it does not have settor methods but instead has multiple constructors. It also makes the "jsonInitProperties" an external api but it is not clear what those are.
You could just make the cleanup of this class another story.

geode-core/src/main/java/org/apache/geode/management/internal/configuration/converters/ClassNameConverter.java
```diff
@@ -0,0 +1,48 @@
...
+  @Override
+  protected ClassName fromNonNullXmlObject(DeclarableType xmlObject) {
+    Properties properties = new Properties();
+    List<ParameterType> parameters = xmlObject.getParameters();
+    for (ParameterType parameter : parameters) {
+      // ClassName only supports string type values
+      if (parameter.getString() != null) {
+        properties.setProperty(parameter.getName(), parameter.getString());
+      }
+    }
```
jmelchio: Lines 33-39 could be written in functional style (preferred):
```java
xmlObject.getParameters().stream()
        .filter(pt -> pt.getString() != null)
        .forEach(p -> properties.setProperty(p.getName(), p.getString()));
```

geode-management/src/main/java/org/apache/geode/management/configuration/Pdx.java
jinmeiliao: the generated code is "Boolean", and there is no default defined in cache.xsd. Do you know what are the defaults of these booleans if they are not specified in the cache.xml?

jinmeiliao: It's just a personal habit that I make my fields protected so that all subclass don't have to reference them using getter/setters. Seems cleaner to me this way, but I can change it.

dschneider-pivotal: You can see the defaults for these in: org.apache.geode.internal.cache.CacheConfig
I think they just all use the standard java default of "false".
The xml parsing code has some logic that can ask if an attribute was specified in the xsd so it might use a null Boolean to determine if it was unspecified.

Since these are in our external java api I think we would want to go with the primitive "boolean". That way when a user asks what the value is they would always get "true" or "false". With Boolean they could also get "null" which I don't think we want to allow in this context.

dschneider-pivotal: I tend to favor "private" and prefer even subclasses to use the gettors and settors. This helps if the gettor or settor has additional behaviour and lets you reason about the code when looking at the class (you know that the fields are only accessible by the current class).

But in this case this class will be in the external api and protected stuff is part of the external api.
Another thing to consider with having classes in the external api; if the class is not final then we are allowing users to create subclasses of our external api classes. Do we want to support that or should we make the classes final or move the class to internal and use an interface.

geode-core/src/main/java/org/apache/geode/management/internal/configuration/converters/ConfigurationConverter.java
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package org.apache.geode.management.internal.configuration.converters;
+
+public abstract class ConfigurationConverter<T, V> {
```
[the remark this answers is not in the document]
jinmeiliao: there is no way to mark the implementation as final if make it an interface. I make this an abstract class so that I don't need to do null check for every converter implementation.

geode-core/src/main/java/org/apache/geode/management/internal/configuration/converters/ClassNameConverter.java
jinmeiliao: nice trick

PR #3806, opened 2019-07-16
GEODE-6975: Reply with exception if fail to persit sender's new id.

pivotal-eshu:
* Reply with ReplyException when processing PrepareNewPersistentMemberMessage
   failed.
 * Retry again if remote member failed to persist the new persistent member id.
 * Make sure atomicCreation flag is reset, so that redundancy could be satisfied
   during bucket creation retry if previous attempt failed.


geode-core/src/main/java/org/apache/geode/internal/cache/BucketPersistenceAdvisor.java
```diff
@@ -461,7 +461,18 @@ public void endBucketCreation(PersistentMemberID newId) {
       }
       atomicCreation = false;
     }
-    super.setOnline(false, true, newId);
+    try {
+      superSetOnline(false, true, newId);
+    } catch (ReplyException replyException) {
+      synchronized (lock) {
```
kirklund: It might be best to transform `atomicCreation` into an `AtomicBoolean` to avoid repeated uses of `synchronized`. The inside the previous `synchronized` block only seems to check or set the flag as well as some logging.

pivotal-eshu:
> I think it might be better to simply not rely in the case of these cancel exceptions - or maybe wait to process the member departure on the sending side of this PrepareNewPersistent...Message?
> 
> You added some retries in the case of this exception, but that looks like it's only for when we are creating the bucket for the very first time? What about when we are recovering the bucket from disk? We don't want that operation to fail if one of our peers is shutting down at the time we are recovering - and I don't think that has a retry.

You are right, retry for recovering case only occurs in getFromOne during GII. We do need to handle the case when recovering without GII.

geode-core/src/main/java/org/apache/geode/internal/cache/BucketPersistenceAdvisor.java
pivotal-eshu: There is other places in the code path that doing operations like setOnline when holding the lock. Whether we change atomicCreation to AtomicBoolean or not, we still need to hold the lock before modifying the value to avoid multiple threads creating same bucket concurrently.

PR #3813, opened 2019-07-18
GEODE-6875: Remove unused & deprecated API usage

jujoramos:
The pull request contains 3 different commits, one for each of the tasks requested through the GEODE ticket.

### For all changes:

### Note:


jujoramos:
Regarding the usage of Internal API, below is the list of classes for which the task couldn't be achieved and the actual reasons:
- `AbstractCache`: no replacements for used API in public classes.
- `AbstractSessionAttributes`: belongs to an internal package, it can use internal classes.
- `Banner`: no replacement for `GemFireVersion` in public packages.
- `CreateRegionFunction`: internal function that creates an internal region dealing with internal mechanisms, no replacements in public packages.
- `DeltaEvent`: belongs to an internal package, it can use internal classes.
- `DeltaSession`: no replacements for used API in public classes.
- `DeltaSessionManager`: no replacements for used API in public classes.
- `DeltaSessionStatistics`: belongs to an internal package, it can use internal classes.
- `GatewayDeltaCreateEvent`: no replacements for used API in public classes.
- `GatewayDeltaEventApplicationCacheListener`: no replacements for used API in public classes.
- `GatewayDeltaForwarderCacheListener`: no replacements for used API in public classes.
- `GemfireSessionManager`: belongs to an internal package, it can use internal classes.
- `Installer`: no replacements for used API in public classes.
- `RegionHelper`: no replacements for used API in public classes.
- `BootstrappingFunctionTest`: no replacements for used API in public classes.
- `ClientServerSessionCacheDUnitTest`: no replacements for used API in public classes.

extensions/geode-modules-session-internal/src/main/java/org/apache/geode/modules/session/internal/common/AbstractSessionCache.java
```diff
@@ -31,19 +29,19 @@
...
    * The operatingRegion is the <code>Region</code> used to do HTTP operations. if local cache is
    * enabled, then this will be the local <code>Region</code>; otherwise, it will be the session
    * <code>Region</code>.
    */
-  protected Region<String, HttpSession> operatingRegion;
+  Region<String, HttpSession> operatingRegion;
 
-  protected Map<CacheProperty, Object> properties =
-      new TypeAwareMap<CacheProperty, Object>(CacheProperty.class);
+  protected TypeAwareMap<CacheProperty, Object> properties =
```
mcmellawatt: For this one, I think that Map is actually a better choice (though it doesn't have any significant impact here).  Usually you just want to reference the interface wherever possible, that way we aren't coupled with a specific Map implementation (TypeAwareMap here).  Maybe I'm overlooking some other part of the code where this change was required so we could call methods on TypeAwareMap that aren't present on Map though, which would be a legitimate reason to change it.

extensions/geode-modules-session-internal/src/main/java/org/apache/geode/modules/session/internal/jmx/SessionStatistics.java
```diff
@@ -44,14 +44,6 @@ public long getRegionUpdates() {
     return regionUpdates.get();
   }
 
-  public void setActiveSessions(int sessions) {
```
mcmellawatt: You mentioned earlier that if a user were very sly that they could write their own JMX component to read our stats.  Would this change be of concern in that case?  I'm not sure we should even really code around that theoretical corner case anyways.

extensions/geode-modules-test/src/main/java/org/apache/geode/modules/session/EmbeddedTomcat.java
```diff
@@ -39,55 +36,34 @@
...
+    container.setCatalinaHome(catalinaHome);
     container.setRealm(new MemoryRealm());
 
     // create webapp loader
     WebappLoader loader = new WebappLoader(this.getClass().getClassLoader());
-    if (classesDir != null) {
-      loader.addRepository(new File(classesDir).toURI().toURL().toString());
-    }
+    // The classes directory for the web application being run.
+    String classesDir = "target/classes";
```
mcmellawatt: Can't easily tell from this diff, but maybe you can just pass the string literal to the method call if it is the only use.

extensions/geode-modules-tomcat8/src/distributedTest/java/org/apache/geode/modules/session/EmbeddedTomcat8.java
```diff
@@ -86,19 +58,18 @@ public EmbeddedTomcat8(String contextPath, int port, String jvmRoute)
...
     // create engine
-    engine = container.getEngine();
+    Engine engine = container.getEngine();
     engine.setName("localEngine");
     engine.setJvmRoute(jvmRoute);
 
     // create http connector
     container.setPort(port);
-    Connector httpConnector = container.getConnector();// ((InetAddress) null, port, false);
+    container.getConnector();// ((InetAddress) null, port, false);
```
mcmellawatt: Does this getter still have side effects if we don't assign it to anything?  If not, can we just delete this line?

extensions/geode-modules-tomcat8/src/distributedTest/java/org/apache/geode/modules/session/Tomcat8SessionsClientServerDUnitTest.java
```diff
@@ -21,41 +21,29 @@
...
 public class Tomcat8SessionsClientServerDUnitTest extends TestSessionsTomcat8Base {
+  private ClientCache clientCache;
 
-  // Set up the session manager we need
-  @Override
-  public void postSetUp() throws Exception {
-    setupServer();
-  }
-
-  @Override
```
mcmellawatt: I think generally speaking we try to put the setup and teardown methods next to eachother.  I've noticed that seems to be the organizational convention to keep these together in test files.

extensions/geode-modules-tomcat8/src/main/java/org/apache/geode/modules/session/catalina/DeltaSession8.java
```diff
@@ -23,6 +23,7 @@
    * Construct a new <code>Session</code> associated with no <code>Manager</code>. The
    * <code>Manager</code> will be assigned later using {@link #setOwner(Object)}.
    */
+  @SuppressWarnings("unused")
```
mcmellawatt: Thanks for adding these - I think they will help if we ever need to do another round of cleanup.

extensions/geode-modules/src/integrationTest/java/org/apache/geode/modules/session/Tomcat6SessionsJUnitTest.java
```diff
@@ -20,10 +20,10 @@
 import org.apache.geode.modules.session.catalina.Tomcat6DeltaSessionManager;
 import org.apache.geode.test.junit.categories.SessionTest;
 
-@Category({SessionTest.class})
+@Category(SessionTest.class)
+// TODO: Remove the entire class once Tomcat6DeltaSessionManager is removed from the product.
```
mcmellawatt: Generally I am anti-TODO because they tend to just languish in the code, but in this case it is a nice comment because it indicates that this code should really be deprecated.

extensions/geode-modules/src/main/java/org/apache/geode/modules/session/bootstrap/AbstractCache.java
```diff
@@ -46,26 +46,31 @@
...
-  protected static final float DEFAULT_EVICTION_HEAP_PERCENTAGE =
+  private static final float DEFAULT_EVICTION_HEAP_PERCENTAGE =
       LocalRegion.DEFAULT_HEAPLRU_EVICTION_HEAP_PERCENTAGE;
 
-  protected static final float DEFAULT_CRITICAL_HEAP_PERCENTAGE =
+  private static final float DEFAULT_CRITICAL_HEAP_PERCENTAGE =
       ResourceManager.DEFAULT_CRITICAL_PERCENTAGE;
 
-  protected static final Logger LOGGER = LoggerFactory.getLogger(AbstractCache.class);
+  // TODO: DistributionConfig is internal, change these fields when the attributes are public.
```
mcmellawatt: Not sure if this TODO was for you or the next person

extensions/geode-modules/src/main/java/org/apache/geode/modules/util/CreateRegionFunction.java
```diff
@@ -179,17 +176,14 @@ public boolean hasResult() {
...
-
   private RegionStatus createRegion(RegionConfiguration configuration) {
     // Get a distributed lock
     DistributedMemberLock dml = getDistributedLock();
     if (this.cache.getLogger().fineEnabled()) {
       this.cache.getLogger().fine(this + ": Attempting to lock " + dml);
     }
-    long start = 0, end = 0;
-    RegionStatus status = null;
+    long start = 0, end;
```
mcmellawatt: won't start also default to 0?

extensions/geode-modules/src/main/java/org/apache/geode/modules/util/RegionConfigurationCacheListener.java
```diff
@@ -99,14 +97,12 @@ public void afterRegionCreate(RegionEvent<String, RegionConfiguration> event) {
...
       this.cache.getLogger().fine(builder1.toString());
-      if (builder2 != null) {
+      if (builder2.length() != 0) {
         this.cache.getLogger().fine(builder2.toString());
       }
     }
   }
-
-  @Override
-  public void init(Properties p) {}
```
mcmellawatt: I assume we can get rid of this because the super also does nothing?

extensions/geode-modules-session-internal/src/main/java/org/apache/geode/modules/session/internal/common/AbstractSessionCache.java
jujoramos: Good catch. I didn't have any "valid" reason to change it, other than getting rid of the `unchecked assignment` warning.

extensions/geode-modules-session-internal/src/main/java/org/apache/geode/modules/session/internal/jmx/SessionStatistics.java
jujoramos: I don't think we should concern about this one: the actual interface `SessionStatisticsMXBean` (what a user should actually use) never had the method in the first place and, moreover, it does't make much sense to provide public JMX methods to set the actual number of total and active sessions anyway.

extensions/geode-modules-test/src/main/java/org/apache/geode/modules/session/EmbeddedTomcat.java
jujoramos: Good catch, thanks.

extensions/geode-modules-tomcat8/src/distributedTest/java/org/apache/geode/modules/session/EmbeddedTomcat8.java
jujoramos: :+1:

extensions/geode-modules-tomcat8/src/distributedTest/java/org/apache/geode/modules/session/Tomcat8SessionsClientServerDUnitTest.java
jujoramos: Yeah, I re-arranged the methods as part of the second commit (9136cdf1bbc9d2043e08860a7b42d5debbacdf4f).

extensions/geode-modules/src/main/java/org/apache/geode/modules/session/bootstrap/AbstractCache.java
jujoramos: Forgot to remove the comment, thanks.

extensions/geode-modules/src/main/java/org/apache/geode/modules/util/CreateRegionFunction.java
jujoramos: Not initialising the `start` variable to `0` makes the compilation fail with `CreateRegionFunction.java:196: error: variable start might not have been initialized`.

extensions/geode-modules/src/main/java/org/apache/geode/modules/util/RegionConfigurationCacheListener.java
jujoramos: Indeed, the `Declarable` interface has the method declared as `default void init(Properties props) {};`.

jujoramos:
Hey @mcmellawatt ,

Thanks for reviewing this, guess it wasn't an easy task due to the amount of files changed... :sweat_smile:. Regarding the Tomcat tests in the `org.apache.geode.session.tests` package, they should be included as part of CI execution under the `:geode-assembly:distributedTest` task:

```
./gradlewStrict -PcompileJVM=/usr/lib/jvm/java-8-openjdk-amd64 -PcompileJVMVer=8 -PtestJVM=/usr/lib/jvm/java-11-openjdk-amd64 -PtestJVMVer=11 -PparallelDunit -PdunitDockerUser=geode -PdunitDockerImage=$(docker images --format '{{.Repository}}:{{.Tag}}') -PdunitParallelForks=24 --parallel --console=plain --no-daemon -x javadoc -x spotlessCheck -x rat distributedTest
```

We should be fine to merge this PR once the last CI finishes (assuming everything's green):
Cheers.

extensions/geode-modules-session-internal/src/main/java/org/apache/geode/modules/session/internal/common/AbstractSessionCache.java
```diff
@@ -31,19 +31,20 @@
...
    * enabled, then this will be the local <code>Region</code>; otherwise, it will be the session
    * <code>Region</code>.
    */
-  protected Region<String, HttpSession> operatingRegion;
+  Region<String, HttpSession> operatingRegion;
 
-  protected Map<CacheProperty, Object> properties =
-      new TypeAwareMap<CacheProperty, Object>(CacheProperty.class);
+  @SuppressWarnings("unchecked")
+  protected Map<CacheProperty, Object> properties = new TypeAwareMap<>(CacheProperty.class);
```
mcmellawatt: I think the `unchecked` warning is because the underlying implementation of TypeAwareMap extends from a non-generic HashMap.  Not sure if that is something we can fix here or not.

jujoramos: Yeah, thanks, I don't think we can do anything about this one here.

PR #3816, opened 2019-07-19
[Do not review] Refactoring of code to use long stats.

mhansonp:


geode-core/src/integrationTest/java/org/apache/geode/internal/statistics/StatSamplerIntegrationTest.java
```diff
@@ -260,10 +260,10 @@ public RuntimeException generateCancelledException(Throwable e) {
...
     incDouble(st1_1, "double_gauge_3", 1);
     incDouble(st1_1, "double_gauge_4", 1);
-    incInt(st1_1, "int_counter_5", 1);
-    incInt(st1_1, "int_counter_6", 1);
-    incInt(st1_1, "int_gauge_7", 1);
-    incInt(st1_1, "int_gauge_8", 1);
+    incLong(st1_1, "int_counter_5", 1);
+    incLong(st1_1, "int_counter_6", 1);
+    incLong(st1_1, "int_gauge_7", 1);
+    incLong(st1_1, "int_gauge_8", 1);
```
pivotal-jbarrett: Either leave the tea of int counters intact to test that the deprecated methods still work or delete them. Changing them to long does t improve test coverage.

demery-pivotal: > Either leave the test of int counters intact to test that the deprecated methods still work or delete them. Changing them to long does t improve test coverage.

If `StatisticsImpl` had tests to verify that each `Int` method had exactly the same effect as the corresponding `Long` method, would you be okay with other classes testing only through the `Long` methods?

PR #3820, opened 2019-07-19
GEODE-6897: implement CMS rebalance operation

onichols-pivotal:
- Adds `operationStart` and `operationEnded` timestamps to CMS async operation framework
- Adds passing of `cache` to performer functions
- Registers `RebalanceOperation` and a performer function (with implementation based on calling the existing gfsh `RebalanceCommand`)
- maps the RebalanceCommand result model output into more JSON-friendly `RebalanceResult`
- breaks `ClusterManagementResult` into more subtypes


onichols-pivotal:
@jmelchio @jackw26

geode-core/src/main/java/org/apache/geode/management/internal/operation/OperationHistoryManager.java
```diff
@@ -96,5 +127,17 @@ public String getId() {
...
+
+    public Date getOperationStart() {
+      return operationStart;
+    }
+
+    public CompletableFuture<Date> getOperationEnded() {
+      return operationEnded;
+    }
+
+    public void setOperationEnded(Date operationEnded) {
```
dschneider-pivotal: I find the overloading of "operationEnded" here confusing. Maybe "this.operationEnded" should actually be named "operationEndedFuture" to make clear that it is a container of operationEnded.

geode-core/src/main/java/org/apache/geode/management/internal/operation/OperationManager.java
```diff
@@ -14,75 +14,95 @@
...
 
+  /**
+   * looks up the start time of an operation by id, or null if not found
+   */
+  public Date getOperationStart(String opId) {
+    return historyManager.getOperationStart(opId);
+  }
+
+  /**
+   * looks up the end time of an operation by id, or null if not found or in progress
```
dschneider-pivotal: I think this comment is wrong about null being returned if "in progress". I think it will return the future in that case.
Also, do the callers of this method want to handle "null" if opId is not found or should it instead throw an exception (same question for getOperationStart())?

geode-core/src/test/java/org/apache/geode/management/internal/operation/OperationHistoryManagerTest.java
```diff
@@ -103,9 +104,33 @@ public void getSomeCompleted() {
     assertThat(history.getStatus("3")).isNotNull();
   }
 
+  @Test
+  public void timestampsAreCorrect() throws Exception {
```
dschneider-pivotal: split this up into three methods. Instead of using "now = System.currentTImeMillis()" use the "new Date()" the test passed in when op() was called. Maybe you should just add a Date parameter to op(String, CompletableFuture)

geode-management/src/main/java/org/apache/geode/management/operation/RebalanceOperation.java
```diff
@@ -0,0 +1,71 @@
...
+import org.apache.geode.annotations.Experimental;
+import org.apache.geode.management.api.ClusterManagementOperation;
+import org.apache.geode.management.runtime.RebalanceResult;
+
+@Experimental
+public class RebalanceOperation implements ClusterManagementOperation<RebalanceResult> {
+  public static final String REBALANCE_ENDPOINT = "/operations/rebalance";
+  private List<String> includeRegions = new ArrayList<>();
+  private List<String> excludeRegions = new ArrayList<>();
+  private boolean simulate;
```
dschneider-pivotal: Are we supporting "simulate"? I thought I saw you remove some deadcode earlier in this review because "simulate" was always false?

geode-management/src/main/java/org/apache/geode/management/operation/RebalanceOperation.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.List;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+
+import org.apache.geode.annotations.Experimental;
+import org.apache.geode.management.api.ClusterManagementOperation;
+import org.apache.geode.management.runtime.RebalanceResult;
+
+@Experimental
+public class RebalanceOperation implements ClusterManagementOperation<RebalanceResult> {
```
dschneider-pivotal: This class is part of the external API so we need javadocs on its publics (class, methods, fields).

geode-management/src/main/java/org/apache/geode/management/operation/RebalanceOperation.java
```diff
@@ -0,0 +1,71 @@
...
+  public static final String REBALANCE_ENDPOINT = "/operations/rebalance";
```
dschneider-pivotal: This does not seem like a fully qualified endpoint. How would users know how to qualify it?

geode-management/src/main/java/org/apache/geode/management/operation/RebalanceOperation.java
dschneider-pivotal: Currently the code allows uses to pass "null" in the constructor and the list settors. Is that a legal value for these lists?

geode-management/src/main/java/org/apache/geode/management/operation/RebalanceOperation.java
```diff
@@ -0,0 +1,71 @@
...
+  private List<String> includeRegions = new ArrayList<>();
+  private List<String> excludeRegions = new ArrayList<>();
+  private boolean simulate;
+
+  public RebalanceOperation() {}
+
+  public RebalanceOperation(List<String> includeRegions,
```
dschneider-pivotal: It is not clear to me why we have these two constructors. If you had just the default one then settors for all the attributes makes sense. But if you had just the non-default one then it seems like the fields could all be final and we could drop the settors.

geode-management/src/main/java/org/apache/geode/management/runtime/RebalanceResult.java
```diff
@@ -0,0 +1,44 @@
...
+import org.apache.geode.annotations.Experimental;
+import org.apache.geode.management.api.JsonSerializable;
+import org.apache.geode.management.operation.RebalanceOperation;
+
+/**
+ * holds the final results of a {@link RebalanceOperation}
+ */
+@Experimental
+public class RebalanceResult implements JsonSerializable {
+  private Map<String, Map<String, Long>> rebalanceSummary = new LinkedHashMap<>();
```
dschneider-pivotal: could we make this field final and have the map passed in to the constructor and get rid of the settor? This is external API and users will never need to call setRebalanceSummary. 
Even better would be for this to be an interface and have the class part of the internal API

geode-core/src/main/java/org/apache/geode/management/internal/operation/OperationHistoryManager.java
onichols-pivotal: preferred naming convention seems for Future getXxx seems to be `getFutureXxx`.  I've updated all such CMS methods accordingly.

geode-core/src/main/java/org/apache/geode/management/internal/operation/OperationManager.java
onichols-pivotal: refactored to return all the instance information in one call so the caller can decide what to do in one place if the operation id is not found

geode-core/src/test/java/org/apache/geode/management/internal/operation/OperationHistoryManagerTest.java
onichols-pivotal: split into multiple tests and removed use of `now`, instead passing start date into constructor

geode-management/src/main/java/org/apache/geode/management/operation/RebalanceOperation.java
onichols-pivotal: yes, `simulate` is supported.  just removed duplicate code.  updated setters to accept null (as if empty list was passed).  removed constructor.

geode-management/src/main/java/org/apache/geode/management/operation/RebalanceOperation.java
onichols-pivotal: done, good catch

geode-management/src/main/java/org/apache/geode/management/operation/RebalanceOperation.java
onichols-pivotal: added javadocs.  This isn't really intended for the user, but we haven't figured out any better place to put it yet.

geode-management/src/main/java/org/apache/geode/management/operation/RebalanceOperation.java
onichols-pivotal: empty constructor is needed for JSON deserialization.  removed the other constructor so there's one standard way to build a rebalance request.

geode-management/src/main/java/org/apache/geode/management/runtime/RebalanceResult.java
onichols-pivotal: some things are necessary for JSON deserialization, but I've made an interface to hide as much as possible from the public API

geode-core/src/main/java/org/apache/geode/management/internal/operation/RebalanceOperationPerformer.java
```diff
@@ -0,0 +1,109 @@
...
+import org.apache.geode.management.runtime.RebalanceResult;
+
+@Experimental
+public class RebalanceOperationPerformer {
+  public static RebalanceResult perform(Cache cache, RebalanceOperation parameters) {
+    try {
+      RebalanceCommand rebalanceCommand = new RebalanceCommand();
+      rebalanceCommand.setCache(cache);
+
+      ResultModel result = rebalanceCommand
```
jinmeiliao: I would like to see us not dependent on the command implementation at all. I believe the command also starts up threads to do the real work, and this performer is already in a thread. Please try to see how much work it is to do the real work in the performer instead of delegating to the command. If it's too much, you can do it in a separate PR.

geode-core/src/main/java/org/apache/geode/management/internal/operation/OperationManager.java
```diff
@@ -14,75 +14,95 @@
...
 
-  public OperationManager(OperationHistoryManager historyManager) {
+  public OperationManager(InternalCache cache, OperationHistoryManager historyManager) {
+    this.cache = cache;
     this.historyManager = historyManager;
     this.executor = LoggingExecutors.newThreadOnEachExecute("CMSOpPerformer");
 
     // initialize the list of operation performers
     performers = new ConcurrentHashMap<>();
+    registerOperation(RebalanceOperation.class, RebalanceOperationPerformer::perform);
```
jinmeiliao: if we eventually replace this with service loader, (you can see how commands are loaded by each extension to load all the custom commands), what needs to be put in the map is an instance of Performer, so I do believe we need to register a stateless performer instance in the map. It probably would help if we have a  dedicated interface for all the performers.

geode-core/src/main/java/org/apache/geode/management/internal/operation/RebalanceOperationPerformer.java
onichols-pivotal: Do you want to use RebalanceCommand.ExecuteRebalanceWithTimeout directly?  Or go even deeper and invoke `executeFunction` directly?

geode-core/src/main/java/org/apache/geode/management/internal/operation/OperationHistoryManager.java
```diff
@@ -27,10 +28,16 @@
...
+ * Retains references to all running and some recently-completed operations.
+ *
+ * The policy for expiring completed operations is subject to change, but may be based on age,
+ * count, when it was last accessed, or some combination thereof.
+ */
 @Experimental
 public class OperationHistoryManager {
-  private final ConcurrentMap<String, CompletableFuture> inProgressHistory;
-  private final Map<String, CompletableFuture> completedHistory;
+  private final ConcurrentMap<String, OperationInstance> inProgressHistory;
```
jinmeiliao: what's the reason that we have to use two maps to hold the history?

geode-core/src/main/java/org/apache/geode/management/internal/operation/RebalanceOperationPerformer.java
jinmeiliao: That ExecuteRebalanceWithTimeout.executeRebalanceWithTimeout looks like it's doing the real thing. There is even comments there saying we need to pull it out to a separate class. We should pull that out and have both the command and performer call that.

geode-core/src/main/java/org/apache/geode/management/internal/operation/OperationHistoryManager.java
dschneider-pivotal: One keeps track of any commands still in progress (inProgessHistory). The other is limited to a certain number of completed commands (completedHIstory) and is implemented with an LRUMap that only remembers a certain number.

jinmeiliao: So it is to guarantee that an "in progress" operation would not get cleaned out if it gets stale? what if an operation "never" finishes somehow?

dschneider-pivotal: If it is still running in the locator then it will still be in the inProgress map. Can you define what you mean by "stale"? If it is still running then we already have a bunch of other resources allocated for it (like a jvm thread) so I don't think have any entry in a map is a problem. Once it completes in the locator it will move to the completed map which has a limit on how many things can be in it. It does not need a client to be checking the status for this move to happen.

geode-core/src/main/java/org/apache/geode/management/internal/operation/RebalanceOperationPerformer.java
```diff
@@ -0,0 +1,109 @@
...
+import org.apache.geode.management.internal.cli.commands.RebalanceCommand;
+import org.apache.geode.management.internal.cli.result.model.InfoResultModel;
+import org.apache.geode.management.internal.cli.result.model.ResultModel;
+import org.apache.geode.management.internal.cli.result.model.TabularResultModel;
+import org.apache.geode.management.operation.RebalanceOperation;
+import org.apache.geode.management.runtime.RebalanceResult;
+
+@Experimental
+public class RebalanceOperationPerformer {
+  public static RebalanceResult perform(Cache cache, RebalanceOperation parameters) {
```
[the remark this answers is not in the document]
onichols-pivotal: `perform` has no dependencies on anything but it's parameters, so it should be an appropriate use of the `static` keyword to declare this.

It's also slightly more succinct that it can be registered as `registerOperation(RebalanceOperation.class, RebalanceOperationPerformer::perform);`

If it were not static, we could still register, but would require a few more characters registerOperation(RebalanceOperation.class, new RebalanceOperationPerformer()::perform);`

geode-core/src/main/java/org/apache/geode/management/internal/operation/OperationHistoryManager.java
jinmeiliao: is it ever possible that an operation instance would stay in this "in progress" map for ever? like that thread just dies or that operation never completes...

geode-core/src/main/java/org/apache/geode/management/internal/operation/RebalanceOperationPerformer.java
dschneider-pivotal: @jinmeiliao was it just the extra thread you wanted to avoid on the RebalanceCommand or did you want avoid other things in the command also? For example is it good for us to get a ResultModel back and translate it into the json response? It seems like a good idea to reuse as much of this code as possible since it is already tested. Is this true?

geode-core/src/main/java/org/apache/geode/management/internal/operation/RebalanceOperationPerformer.java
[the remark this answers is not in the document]
dschneider-pivotal: Because we don't actually need an instance of RebalanceOperationPerformer. Since it has no state you can just reference the method like so: RebalanceOperationPerformer::perform

geode-core/src/main/java/org/apache/geode/management/internal/operation/OperationManager.java
dschneider-pivotal: The current: RebalanceOperationPerformer::perform is a stateless performer. I think you might be right that it could help if we had our own interface that defines this contract more explicitly. But this BiFunction one might work. Would it be okay to delay adding support for service loader until someone works on that story and then they could add the performer interface if needed?

geode-core/src/main/java/org/apache/geode/management/internal/operation/RebalanceOperationPerformer.java
jinmeiliao: ResultModel and Command are "gfsh" concept, I would love to see CMS not using any Gfsh construct at all, since eventually we want to get rid of gfsh v1. In my mind, I would love to see a RebalanceResult that would have those attributes like "regionRebalanced", "TotalBucketCreateBytes",  "TotalBucketCreatesCompleted" etc as fields in this RebalanceResult, and both gfsh command would call that to construct its own ResultModel. But looks like its going to too much work.

geode-core/src/main/java/org/apache/geode/management/internal/operation/OperationHistoryManager.java
onichols-pivotal: The only way I know for a thread to "die" is by throwing some `Throwable`.  We execute using `CompletableFuture.supplyAsync` which completes exceptionally if it catches Throwable.  On the other hand, if you're talking about an operation that hangs, it seems like it-progress forever is an accurate depiction of that state.  Are you requesting that we put in a global or per-operation-type timeout value beyond which any operations still in progress will be forcibly killed?

geode-core/src/main/java/org/apache/geode/management/internal/operation/RebalanceOperationPerformer.java
onichols-pivotal: Right now a lot of the implementation of the inner class `RebalanceCommand.ExecuteRebalanceWithTimeout` still depends on helper methods in the parent class `GfshCommand`.  For today I will refactor just enough to get the callable without incurring the overhead of an extra thread, then tomorrow we can look into teasing out ExecuteRebalanceWithTimeout into a standalone class.

geode-core/src/main/java/org/apache/geode/management/internal/operation/OperationManager.java
jinmeiliao: yeah, this can be delayed.

geode-core/src/main/java/org/apache/geode/management/internal/operation/RebalanceOperationPerformer.java
jinmeiliao: actually I don't mind you put all this work in another PR, as long as we have this work to be done captured in another ticket.

geode-core/src/main/java/org/apache/geode/management/internal/operation/OperationHistoryManager.java
jinmeiliao: No, I am just wondering why we need to use two maps for these operationInstances, can we use just one LRU map to keep all of them? So if one operation is "in-progress" forever, it can still be wiped out eventually.

geode-core/src/main/java/org/apache/geode/management/internal/operation/RebalanceOperationPerformer.java
onichols-pivotal: It makes a difference in the output structure (the rebalance result), so releasing rebalance with one format then totally changing it may be less than desirable.  If it's not too much work to leverage the existing org.apache.geode.cache.control.RebalanceResults interface (would require moving it from geode-core to geode-management) that might be desirable.

geode-core/src/main/java/org/apache/geode/management/internal/operation/OperationHistoryManager.java
onichols-pivotal: Because we would like to guarantee that status will not disappear before the operation has a chance to complete.  If you use a single LRU cache with size=100, then start 101 async operations, you are guaranteeing that you'll never be able to retrieve status for one of them.  Although we may still have then problem once they complete, you would still be guaranteed to lose one result.  Maybe this should be reworked to a single map with a time-based expiration policy e.g. all results are kept until 1 hour after operation completes?

onichols-pivotal: changed to 2-hour retention policy of completed operations and a single map rather than two separate ones

geode-core/src/main/java/org/apache/geode/management/internal/cli/commands/RebalanceCommand.java
```diff
@@ -105,6 +105,12 @@ public ResultModel rebalance(
     return result;
   }
 
+  public Callable<ResultModel> rebalanceCallable(String[] includeRegions, String[] excludeRegions,
+      boolean simulate) {
+    return new ExecuteRebalanceWithTimeout(includeRegions, excludeRegions, simulate,
```
dschneider-pivotal: It seems like rebalanceCallable or ExecuteRebalanceWithTimeout should make sure to check the ResultModel it is returning and if it has a single error (see line 102) to set the status to ERROR (line 103).

geode-core/src/main/java/org/apache/geode/management/internal/operation/RebalanceOperationPerformer.java
```diff
@@ -35,11 +35,11 @@ public static RebalanceResult perform(Cache cache, RebalanceOperation parameters
...
       rebalanceCommand.setCache(cache);
 
-      ResultModel result = rebalanceCommand
-          .rebalance(arr(parameters.getIncludeRegions()), arr(parameters.getExcludeRegions()), -1,
-              parameters.isSimulate());
+      ResultModel result = rebalanceCommand.rebalanceCallable(arr(parameters.getIncludeRegions()),
+          arr(parameters.getExcludeRegions()), parameters.isSimulate()).call();
 
-      if (result.getStatus().equals(Result.Status.ERROR)) {
+      if (result.getStatus().equals(Result.Status.ERROR)
```
dschneider-pivotal: If you make the previous change I requested then this code does not need to change.

geode-core/src/main/java/org/apache/geode/management/internal/operation/OperationHistoryManager.java
```diff
@@ -36,30 +36,55 @@
...
+        .filter(e -> e.getValue().getFutureOperationEnded().isDone()
+            && getOperationEnded(e.getValue()).getTime() < purgeThreshold)
+        .map(e -> e.getKey()).collect(Collectors.toSet());
+    expiredKeys.forEach(history::remove);
+  }
+
+  /**
+   * @return null if unavailable
+   */
+  private static Date getOperationEnded(OperationInstance<?, ?> operationInstance) {
```
dschneider-pivotal: how about making this "boolean hasExpired(operationInstance, purgeThreshold)"
Then you can do the test against purgeThreshold here and if you are not able to get a Date
because of the exceptions return either true or false instead of null.
Returning null will cause the caller of getOperationEnded to throw NPE since it always calls getTime() on the result.
It seems like it would be okay to return true and remove items whose get() method is throwing.

geode-core/src/test/java/org/apache/geode/management/internal/operation/OperationHistoryManagerTest.java
```diff
@@ -83,9 +90,13 @@ public void getLotsOfCompleted() {
     future2.complete(null);
     CompletableFuture<JsonSerializable> future3 = new CompletableFuture<>();
     history.save(op("3", future3));
```
dschneider-pivotal: It seems like having multiple items in this unit test is now just noise since we don't have an LRU limit. I think you want a unit test that shows that both save and getOperationInstance will expire things. I think it would be good to add another method, just for the unit test, to OperationHistoryManager that lets you get an operation instance without calling expire. That allows the unit test to look without causing a change

geode-core/src/main/java/org/apache/geode/management/internal/operation/RebalanceOperationPerformer.java
onichols-pivotal: Tried today to remove dependency on gfsh / ResultModel, but it's much more entangled than I realized.  That will have to be a project for another PR.

geode-core/src/main/java/org/apache/geode/management/internal/cli/commands/RebalanceCommand.java
onichols-pivotal: I pulled this common code inside of the callable.

geode-core/src/main/java/org/apache/geode/management/internal/operation/OperationHistoryManager.java
onichols-pivotal: In the usage it never would actually returns null, because isDone is checked first.  However, a `hasExpired` test is a little cleaner, so went ahead and made that change

geode-core/src/test/java/org/apache/geode/management/internal/operation/OperationHistoryManagerTest.java
onichols-pivotal: Instead I moved the call to System.currentTimeMillis into a `now()` method that can be mocked to give tests "virtual time" ability.

geode-core/src/main/java/org/apache/geode/management/internal/operation/RebalanceOperationPerformer.java
```diff
@@ -0,0 +1,152 @@
...
+import org.apache.geode.management.runtime.RebalanceResult;
+import org.apache.geode.util.internal.GeodeJsonMapper;
+
+@Experimental
+public class RebalanceOperationPerformer {
+  /**
+   * the {@link ResultModel} contains human-readable strings, however we want property names that
+   * more closely resemble {@link PartitionRebalanceInfo} (but with time units)
+   */
+  private static final Map<String, String> translations =
```
dschneider-pivotal: I think you need to add the Immutable annotation to this. That is what is breaking the builds.

geode-core/src/main/java/org/apache/geode/management/internal/operation/RebalanceResultImpl.java
```diff
@@ -21,16 +21,110 @@
...
+  public Map<String, PerRegionStats> getRebalanceStats() {
     return rebalanceSummary;
   }
 
-  void setRebalanceSummary(Map<String, Map<String, Long>> rebalanceSummary) {
+  void setRebalanceSummary(Map<String, PerRegionStats> rebalanceSummary) {
     this.rebalanceSummary = rebalanceSummary;
   }
+
+  public static class PerRegionStatsImpl implements PerRegionStats {
```
dschneider-pivotal: should we override toString and this class so java api users can easily print it out?
also on the owner class RebalanceResultImpl

onichols-pivotal: added toString

geode-management/src/main/java/org/apache/geode/management/runtime/RebalanceResult.java
```diff
@@ -0,0 +1,53 @@
...
+ * the final results of a {@link RebalanceOperation}
+ */
+@Experimental
+public interface RebalanceResult extends JsonSerializable {
+  /**
+   * @return for each region, the statistics reported by rebalance
+   */
+  Map<String, PerRegionStats> getRebalanceStats();
+
+  interface PerRegionStats extends JsonSerializable {
```
dschneider-pivotal: I think you should change the name "PerRegionStats". How about "RebalanceRegionResult"?
This would change the method "getRebalanceStats" to "getRebalanceRegionResults".
Stats is an abbreviation. It is also used in geode for live data.
I also don't think we need to narrow this interface to only having stats in it. We may want to include other info in the future.

geode-management/src/main/java/org/apache/geode/management/runtime/RebalanceResult.java
```diff
@@ -0,0 +1,53 @@
...
+    long getBucketCreateBytes();
+
+    long getBucketCreateTimeInMilliseconds();
```
dschneider-pivotal: This interface and its methods need javadocs since they are part of the external java api. You could cut and paste them from the interface the geode core has.

geode-management/src/main/java/org/apache/geode/management/runtime/RebalanceResult.java
onichols-pivotal: done

geode-management/src/main/java/org/apache/geode/management/runtime/RebalanceResult.java
onichols-pivotal: done

geode-management/src/main/java/org/apache/geode/management/runtime/RebalanceRegionResult.java
```diff
@@ -0,0 +1,67 @@
...
+   */
+  long getPrimaryTransferTimeInMilliseconds();
+
+  /**
+   * Returns the number of primaries that were transferred for this region.
+   */
+  int getPrimaryTransfersCompleted();
+
+  /**
+   * @return the time, in milliseconds, that the rebalance operation took for this region.
```
dschneider-pivotal: make this one a proper sentence (like the others) instead of @return

geode-management/src/main/java/org/apache/geode/management/runtime/RebalanceRegionResult.java
```diff
@@ -0,0 +1,67 @@
...
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.apache.geode.management.runtime;
+
+import org.apache.geode.annotations.Experimental;
+import org.apache.geode.management.api.JsonSerializable;
+
+@Experimental
+public interface RebalanceRegionResult extends JsonSerializable {
```
dschneider-pivotal: add a javadoc comment for the interface

geode-management/src/main/java/org/apache/geode/management/runtime/RebalanceRegionResult.java
onichols-pivotal: done

geode-management/src/main/java/org/apache/geode/management/runtime/RebalanceRegionResult.java
onichols-pivotal: done

geode-management/src/main/java/org/apache/geode/management/api/ClusterManagementListOperationsResult.java
```diff
@@ -0,0 +1,55 @@
...
+import org.apache.geode.annotations.Experimental;
+import org.apache.geode.management.runtime.OperationResult;
+
+/**
+ * returned from {@link ClusterManagementService#list(ClusterManagementOperation)}
+ *
+ * @param <V> the result type associated with the operation type being listed
+ */
+@Experimental
+public class ClusterManagementListOperationsResult<V extends OperationResult>
```
dschneider-pivotal: Could this class be made an interface and have an internal class that implements the interface? It would be nice for "setResult" and the constructors to not be an external api.

Note that some of the methods on the class are missing javadocs which is okay if it was internal but every method in external should hav e javadocs

geode-management/src/main/java/org/apache/geode/management/api/ClusterManagementOperation.java
```diff
@@ -14,12 +14,13 @@
...
-
 import com.fasterxml.jackson.annotation.JsonIgnore;
 
+import org.apache.geode.management.runtime.OperationResult;
+
 /**
  * Geode modules, extensions, or user code can define cluster management operations to use with
- * {@link ClusterManagementService#startOperation(ClusterManagementOperation)} by subclassing this
+ * {@link ClusterManagementService#start(ClusterManagementOperation)} by subclassing this
  * abstract base class and adding any necessary parameters. For example, a CompactDiskStore
```
dschneider-pivotal: update the comment to not talk about subclassing abstract base class

geode-management/src/main/java/org/apache/geode/management/api/ClusterManagementService.java
```diff
@@ -59,24 +60,52 @@
    */
   <T extends CacheElement> ClusterManagementResult update(T config);
 
+  /**
+   * This method will list instances of the element type in the cluster configuration, along with
+   * additional runtime information from cluster members
+   *
+   * @param config this may be used to filter the search results by id or group (check documentation
+   *        for individual element types to see if they support filtering by addition attributes)
```
dschneider-pivotal: add @return

geode-management/src/main/java/org/apache/geode/management/api/ClusterManagementService.java
```diff
@@ -59,24 +60,52 @@
...
+   */
   <T extends CacheElement & CorrespondWith<R>, R extends RuntimeInfo> ClusterManagementListResult<T, R> list(
       T config);
 
+  /**
+   * This method will list a single instance of the element type in the cluster configuration, along
+   * with additional runtime information from cluster members
+   *
+   * @param config this must be used to filter the search results to a single result, generally by
+   *        supplying the name (id) of the desired config element
```
dschneider-pivotal: add @return

geode-management/src/main/java/org/apache/geode/management/api/ClusterManagementListOperationsResult.java
onichols-pivotal: made it immutable, that's about as good as we can get for now and still be JSON-deserializable (without adding a `class` attribute exposing the implementation class in the JSON)

onichols-pivotal: added missing javadocs

geode-management/src/main/java/org/apache/geode/management/api/ClusterManagementOperation.java
onichols-pivotal: done

geode-management/src/main/java/org/apache/geode/management/api/ClusterManagementService.java
onichols-pivotal: done

geode-management/src/main/java/org/apache/geode/management/api/ClusterManagementService.java
onichols-pivotal: done

PR #3655, opened 2019-05-31
GEODE-3718: InternalResourceManager thread pool shutdown improvement

mivanac:
Thank you for submitting a contribution to Apache Geode.

In order to streamline the review of the contribution we ask you
to ensure the following steps have been taken:

### For all changes:
- [*] Is there a JIRA ticket associated with this PR? Is it referenced in the commit message?

- [*] Has your PR been rebased against the latest commit within the target branch (typically `develop`)?

- [*] Is your initial contribution a single, squashed commit?

- [*] Does `gradlew build` run cleanly?

- [x] Have you written or updated unit tests to verify your changes?

- [ ] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)?

### Note:
Please ensure that once the PR is submitted, check Concourse for build issues and
submit an update to your PR as soon as possible. If you need help, please send an
email to dev@geode.apache.org.

PR #3821, opened 2019-07-19
GEODE-3718: shutdownNow should also interrupt non executing runnables

jhuynh1:
This is a follow up commit for GEODE-3718.  shutdownNow was causing some threads waiting for future.get() to get hung.


jhuynh1:
@kirklund @pivotal-jbarrett @boglesby @dschneider-pivotal
Would you be able to review or determine if this change is ok?  It's either this change (or something similar) or reverting the previous change https://github.com/apache/geode/pull/3655

jhuynh1:
I'm going to merge this and if needed, we can revert.  Thanks @kirklund @onichols-pivotal @jujoramos for the reviews!

PR #3827, opened 2019-07-22
 GEODE-7002: use enable-management-rest-service property to enable/disable CMS

jinmeiliao:

### For all changes:

### Note:


jinmeiliao:
@jmelchio

geode-core/src/main/java/org/apache/geode/distributed/internal/InternalLocator.java
```diff
@@ -756,9 +757,9 @@ private void startClusterManagementService() throws IOException {
         new ImmutablePair<>(HttpService.CLUSTER_MANAGEMENT_SERVICE_CONTEXT_PARAM,
             clusterManagementService);
 
-    if (Boolean.getBoolean(ClusterManagementService.FEATURE_FLAG)) {
-      logger.info("System Property {}=true Geode Management API is enabled.",
-          ClusterManagementService.FEATURE_FLAG);
+    if (distributionConfig.getEnableManagementRestService()) {
+      logger.info("Geode Property {}=true Geode Management Rest Service is enabled.",
+          ConfigurationProperties.ENABLE_MANAGEMENT_REST_SERVICE);
       internalCache.getHttpService().ifPresent(x -> {
```
dschneider-pivotal: Should we log a warning here if the http service is not present or if cluster management service is disabled?

jinmeiliao: if the http service is disabled, there is a message saying "http service is not enabled with port = 0", and none of these messages will show up.

geode-core/src/main/java/org/apache/geode/distributed/ConfigurationProperties.java
```diff
@@ -461,17 +461,18 @@
...
-   * <U>Description</U>: "true" causes cluster management rest service to be enabled. this service
-   * requires cluster configuration service to be present. If "enable-cluster-configuration" is
-   * false and this is true, the management rest service would do nothing but report error if you
-   * try to use it. Also, if http-service-port is set to 0, this would not be enabled regardless of
-   * what value property.
+   * <U>Description</U>: "true" causes the cluster management rest service to be enabled. This
+   * service
+   * requires the cluster configuration service to be enabled. If "enable-cluster-configuration" is
+   * false and this is true, the management rest service would do nothing but report an error if you
+   * try to use it. Also even this property is set to "true", The management rest service will only
```
dschneider-pivotal: Add "if" after "even". "Also even if this".

geode-core/src/main/java/org/apache/geode/internal/cache/HttpService.java
```diff
@@ -64,7 +64,7 @@
 
   public HttpService(String bindAddress, int port, SSLConfig sslConfig) {
     if (port == 0) {
-      return;
+      throw new IllegalArgumentException("HttpService not started with port = 0.");
```
dschneider-pivotal: I don't like this change. Our docs say that setting http-service-port to 0 disables the HTTP service.
But throwing this exception will cause code in GemFireCacheImpl to log a warning. Setting the property to 0 should not cause a warning.
I suggest that instead of throwing an exception here you log an info message about the service being disabled and just return as the code did before.
The other option would be to change this code in GemFireCacheImpl to not even call new HttpService if systemConfig.getHttpServicePort() is zero. Maybe that is better because it seems odd for GemFireCacheImpl to have an instance of HttpService if it has been disabled.

dschneider-pivotal: Here is the code in GemFireCacheImpl:        try {
          httpService = Optional.of(new HttpService(systemConfig.getHttpServiceBindAddress(),
              systemConfig.getHttpServicePort(), SSLConfigurationFactory
                  .getSSLConfigForComponent(systemConfig, SecurableCommunicationChannel.WEB)));
        } catch (Throwable ex) {
          logger.warn("Could not enable HttpService: {}", ex.getMessage());
        }

jinmeiliao: Noted. I just want to httpService to be null if port=0.

PR #3828, opened 2019-07-22
GEODE-6883 Move the membership code into a separate gradle sub-project

bschuchardt:
This PR narrows the effort of GEODE-6883 to focus only on the "gms"
package and its subpackages.  Other membership classes in
org.apache.geode.distributed.internal.membership are Geode interfaces,
adapters and factories.  Classes in "gms" are membership
classes/interfaces that should not refer to anything outside of "gms"
(though they still do).

This PR is otherwise focused on removing references to
InternalDistributedMember and DistributionMessage from "gms".

    GMS classes only refer to GMSMember
    GMS classes use GMSMembershipView.  NetView is now an interface
    GMS classes do not refer to DistributionMessage.  JGroupsMessenger
      expects GMSMessage instances.  Geode messages to be sent over UDP
      are wrapped in a GMSMessageAdapter.
    Message classes in the "gms" packages extend AbstractGMSMessage which implements GMSMessage
    GMSMembershipManager has an inner class that implements the GMS Manager
      interface and is now in the "adapter" package
    GMSMembershipManager translates GMSMembershipView into a MembershipView
      for the rest of Geode to use (this is the old NetView class)
    GMS instantiation allows us to inject the Manager into the new Services
      instance.
    Other adapter classes have been added to translate between Geode
      and GMS.

    GMSUtil has new methods for marshalling/unmarshalling
    InternalDistributedMember instances for backward-compatibility.
    GMSMember now has the same on-wire form as
    InternalDistributedMember.  This allows the GMS classes to
    deserialize a message from a pre-1.10 member whose code writes
    InternalDistributedMembers when serializing something like a
    JoinRequest.

### For all changes:

@Bill 

### Note:

### DIFF
--- a/geode-core/src/integrationTest/java/org/apache/geode/distributed/internal/membership/MembershipJUnitTest.java
+++ b/geode-core/src/integrationTest/java/org/apache/geode/distributed/internal/membership/MembershipJUnitTest.java
@@ -31,6 +31,7 @@ import static org.mockito.Mockito.when;
 
 import java.io.File;
 import java.net.InetAddress;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Properties;
 
@@ -50,21 +51,14 @@ import org.apache.geode.distributed.internal.DistributionConfigImpl;
 import org.apache.geode.distributed.internal.InternalDistributedSystem;
 import org.apache.geode.distributed.internal.InternalLocator;
 import org.apache.geode.distributed.internal.SerialAckedMessage;
+import org.apache.geode.distributed.internal.membership.adapter.GMSMembershipManager;
 import org.apache.geode.distributed.internal.membership.adapter.auth.GMSAuthenticator;
+import org.apache.geode.distributed.internal.membership.gms.GMSMember;
+import org.apache.geode.distributed.internal.membership.gms.GMSMembershipView;
 import org.apache.geode.distributed.internal.membership.gms.ServiceConfig;
 import org.apache.geode.distributed.internal.membership.gms.Services;
 import org.apache.geode.distributed.internal.membership.gms.interfaces.JoinLeave;
 import org.apache.geode.distributed.internal.membership.gms.membership.GMSJoinLeave;
-import org.apache.geode.distributed.internal.membership.gms.messages.HeartbeatMessage;
-import org.apache.geode.distributed.internal.membership.gms.messages.HeartbeatRequestMessage;
-import org.apache.geode.distributed.internal.membership.gms.messages.InstallViewMessage;
-import org.apache.geode.distributed.internal.membership.gms.messages.JoinRequestMessage;
-import org.apache.geode.distributed.internal.membership.gms.messages.JoinResponseMessage;
-import org.apache.geode.distributed.internal.membership.gms.messages.LeaveRequestMessage;
-import org.apache.geode.distributed.internal.membership.gms.messages.RemoveMemberMessage;
-import org.apache.geode.distributed.internal.membership.gms.messages.SuspectMembersMessage;
-import org.apache.geode.distributed.internal.membership.gms.messages.ViewAckMessage;
-import org.apache.geode.distributed.internal.membership.gms.mgr.GMSMembershipManager;
 import org.apache.geode.internal.AvailablePortHelper;
 import org.apache.geode.internal.admin.remote.RemoteTransportConfig;
 import org.apache.geode.internal.net.SocketCreator;
@@ -184,14 +178,14 @@ public class MembershipJUnitTest {
         }
       }
 
-      NetView view = jl1.getView();
-      InternalDistributedMember notCreator;
+      GMSMembershipView view = jl1.getView();
+      GMSMember notCreator;
       if (view.getCreator().equals(jl1.getMemberID())) {
         notCreator = view.getMembers().get(1);
       } else {
         notCreator = view.getMembers().get(0);
       }
-      List<String> result = notCreator.getGroups();
+      List<String> result = Arrays.asList(notCreator.getGroups());
 
       System.out.println("sending SerialAckedMessage from m1 to m2");
       SerialAckedMessage msg = new SerialAckedMessage();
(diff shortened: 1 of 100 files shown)


geode-core/src/integrationTest/java/org/apache/geode/distributed/internal/membership/MembershipJUnitTest.java
```diff
@@ -452,65 +446,4 @@ public void testMulticastDiscoveryNotAllowed() {
       // expected
     }
   }
-
```
echobravopapa: was this replaced?

geode-core/src/integrationTest/java/org/apache/geode/distributed/internal/membership/adapter/GMSMembershipManagerJUnitTest.java
```diff
@@ -157,51 +163,48 @@ public void initMocks() throws Exception {
...
   public void tearDown() throws Exception {
     if (manager != null) {
-      manager.stop();
-      manager.stopped();
+      manager.getGMSManager().stop();
+      manager.getGMSManager().stopped();
     }
   }
 
-  @Test
```
echobravopapa: is this test case still valid?

geode-core/src/integrationTest/java/org/apache/geode/distributed/internal/membership/gms/locator/GMSLocatorIntegrationTest.java
```diff
@@ -53,14 +53,14 @@ public void viewFileIsNullByDefault() {
 
   @Test
   public void initDefinesViewFileInSpecifiedDirectory() {
-    gmsLocator.init(tcpServer);
+    gmsLocator.init("" + tcpServer.getPort());
```
echobravopapa: I've got feedback from the Community that `String.valueOf(int)` is preferred over empty quotes concatenation...

geode-core/src/integrationTest/java/org/apache/geode/distributed/internal/membership/MembershipJUnitTest.java
bschuchardt: These classes no longer subclass DistributionMessage, so their "process" messages were removed and that made these tests unnecessary.

geode-core/src/integrationTest/java/org/apache/geode/distributed/internal/membership/adapter/GMSMembershipManagerJUnitTest.java
bschuchardt: yes, that test is still valid.  I don't recall why I removed it but I'll put it back.

```
@Test
  public void testSendMessage() throws Exception {
    HighPriorityAckedMessage m = new HighPriorityAckedMessage();
    m.setRecipient(mockMembers[0]);
    manager.getGMSManager().start();
    manager.getGMSManager().started();
    GMSMember myGMSMemberId = ((GMSMemberAdapter)myMemberId.getNetMember()).getGmsMember();
    List<GMSMember> gmsMembers = members.stream().map(x -> ((GMSMemberAdapter)x.getNetMember()).getGmsMember()).collect(
        Collectors.toList());
    manager.getGMSManager().installView(new GMSMembershipView(myGMSMemberId, 1, gmsMembers));
    Set<InternalDistributedMember> failures = manager.send(m.getRecipients(), m, this.services.getStatistics());
    verify(messenger).send(isA(GMSMessageAdapter.class));
    if (failures != null) {
      assertEquals(0, failures.size());
    }
  }
```

bschuchardt:
Note: ClusterCommunicationsDUnitTest > performARollingUpgrade will fail in Stress Tests because v1.9 locators attempt to read a membership view persisted to disk using v1.10.  This is due to a flaw in v1.9 that doesn't prevent the locator from reading this information if the version stamp on that data is greater than Version.CURRENT.

geode-core/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSUtil.java
```diff
@@ -51,6 +64,35 @@
...
   }
 
+  public static GMSMember readMemberID(DataInput in) throws IOException, ClassNotFoundException {
+    Object id = DataSerializer.readObject(in);
+    if (id == null || id instanceof GMSMember) {
+      return (GMSMember) id;
+    }
+    // return ((GMSMemberAdapter)((InternalDistributedMember)id).getNetMember()).getGmsMember();
+    try {
+      Method getNetMember = id.getClass().getMethod("getNetMember");
```
upthewaterspout: Why is this method using reflection to invoke methods?

geode-core/src/main/java/org/apache/geode/management/internal/configuration/messages/ConfigurationResponse.java
```diff
@@ -144,6 +144,6 @@ public void setMember(DistributedMember member) {
 
   @Override
   public Version[] getSerializationVersions() {
-    return new Version[] {Version.CURRENT};
+    return null;
```
upthewaterspout: How does this change relate to the rest of this PR?

geode-core/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSUtil.java
bschuchardt: It's using reflection because gms classes can't reference InternalDistributedMember or its getNetMember method.  This is backward-compatibility code for rolling-upgrade support.

geode-core/src/main/java/org/apache/geode/management/internal/configuration/messages/ConfigurationResponse.java
bschuchardt: I ran into a failure caused by this code in testing rolling-upgrade code where I purposefully introduced an older version into a cluster already in the process of upgrading.

geode-core/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSUtil.java
bschuchardt: wrt GMSMembershipManager being an adapter, it doesn't actually make any GMS decisions - doesn't elect a coordinator doesn't decide whether a member is there or not, etc.  It mostly handles "surprise members" and "shunned members" and the TCPConduit communications service.  It needs to know about HighPriorityDistributionMessage, SerialDistributionMessages and the like.
I put a lot of effort into trying to make it not be an adapter and not know about InternalDistributedMember before deciding it wasn't an integral part of the gms package and that the Manager interface in gms is a contract between gms and the user of gms (Geode's DistributionManager).
This is completely in line with the old GemFire source base that used JGroups for gms.  In that implementation JGroupMembershipManager was the glue between GemFire and JGroups and translated between JGroups views/identifiers/messages and the versions used by GemFire.

geode-core/src/main/java/org/apache/geode/distributed/internal/membership/adapter/GMSMembershipManager.java
```diff
@@ -840,11 +782,14 @@ private void removeWithViewLock(InternalDistributedMember dm, boolean crashed, S
    * @param member the member
    */
   protected void handleOrDeferSurpriseConnect(InternalDistributedMember member) {
-    synchronized (startupLock) {
+    startupLock.readLock().lock();
```
upthewaterspout: It looks like this PR is also changing the way this class does synchronization. I'm not sure we want to combine these synchronization related changes with refactoring the dependencies of the membership code.

bschuchardt: I've removed that change.  It needs to be done sometime though - synchronizing on this object for every message dispatched has got to be hurting performance.

geode-core/src/main/java/org/apache/geode/management/internal/configuration/messages/ConfigurationResponse.java
bschuchardt: This is GEODE-6956, so I'll link that ticket to this one.

geode-core/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSUtil.java
bschuchardt: We're going to look into, in a separate effort, whether GMSMembershipManager can be moved into gms so that MembershipManager is the contract between Geode and membership.

pivotal-jbarrett: Be careful here. I know you say this is for backwards compatibility but invoking `getMethod` involves some allocations. If this or similar access is in a hot path consider caching the Method `object`.

bschuchardt: Good thing to consider Jake.   This method is only used for membership messages during rolling upgrade.  It isn't used for messages defined outside of the gms packages, like Geode cache operation messages.  After rolling upgrade it won't take the Method lookup path because gms messages use GMSMember identifiers instead of InternalDistributedMember identifiers.

bschuchardt:
I have run full performance tests on this branch wrt develop and it passes with minor +/- variations that seem to be in line with what I see in the CI runs.

bschuchardt:
I'm going to wait until we cut a 1.10 branch to merge this into develop.

PR #3829, opened 2019-07-22
GEODE-7001: Add region entry count gauge

demery-pivotal:
Add 'member.region.entries' gauge in RegionPerfStats
- Add region.name and data.policy tags
- Add a `LongAdder` to track the entry count
- Configure the member.region.entries gauge to fetch from the `LongAdder`
- Configure the 'entryCount' stat to be supplied by the `LongAdder`
    
Also refactored the following:

- Reorganize CachePerfStats/RegionPerfStats constructors
- Remove CachePerfStats.getEntries()
- Remove use of CachePerfStats.getEntries() from PartitionedRegionStatus.
- Add @Override to ValidatingStatisticsType methods in StatisticsTypeImpl
- Extracted invokeSuppliers() to new SuppliableStatistics interface and rename as updateSuppliedValues()
- Move responsibility to prepend 'RegionStats-' onto region statistics textID into the classes that create RegionPerfStats and CachePerfStats.

    
### For all changes:

@aaronlindsey please review

### DIFF
--- a/geode-assembly/src/acceptanceTest/java/org/apache/geode/metrics/RegionEntriesGaugeTest.java
+++ b/geode-assembly/src/acceptanceTest/java/org/apache/geode/metrics/RegionEntriesGaugeTest.java
@@ -1,0 +1,336 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package org.apache.geode.metrics;
+
+import static java.util.Arrays.asList;
+import static java.util.stream.Collectors.toList;
+import static org.apache.geode.cache.RegionShortcut.PARTITION;
+import static org.apache.geode.cache.RegionShortcut.PARTITION_REDUNDANT;
+import static org.apache.geode.cache.RegionShortcut.REPLICATE;
+import static org.apache.geode.cache.client.ClientRegionShortcut.PROXY;
+import static org.apache.geode.test.awaitility.GeodeAwaitility.await;
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.IOException;
+import java.io.StringReader;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Stream;
+
+import io.micrometer.core.instrument.Gauge;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.rules.TemporaryFolder;
+
+import org.apache.geode.cache.Region;
+import org.apache.geode.cache.client.ClientCache;
+import org.apache.geode.cache.client.ClientCacheFactory;
+import org.apache.geode.cache.execute.Function;
+import org.apache.geode.cache.execute.FunctionContext;
+import org.apache.geode.internal.AvailablePortHelper;
+import org.apache.geode.metrics.rules.MetricsPublishingServiceJarRule;
+import org.apache.geode.metrics.rules.SingleFunctionJarRule;
+import org.apache.geode.test.junit.categories.MetricsTest;
+import org.apache.geode.test.junit.rules.gfsh.GfshExecution;
+import org.apache.geode.test.junit.rules.gfsh.GfshRule;
+
+@Category(MetricsTest.class)
+public class RegionEntriesGaugeTest {
+
+  @Rule
+  public GfshRule gfshRule = new GfshRule();
+
+  @Rule
+  public TemporaryFolder temporaryFolder = new TemporaryFolder();
+
+  // TODO: refactor to JarRule
+  @Rule
+  public MetricsPublishingServiceJarRule metricsPublishingServiceJarRule =
+      new MetricsPublishingServiceJarRule("metrics-publishing-service.jar",
+          SimpleMetricsPublishingService.class);
+
+  @Rule
+  public SingleFunctionJarRule functionJarRule =
+      new SingleFunctionJarRule("function.jar", GetMemberRegionEntriesGaugeFunction.class);
+
+  private static final String SPACE = " ";
+
+  private final List<File> serverFolders = new ArrayList<>();
+  private File folderForLocator;
+  private ClientCache clientCache;
+  private String connectToLocatorCommand;
+  private String locatorString;
+
+  @Before
+  public void startMembers() throws Exception {
+    int[] availablePorts = AvailablePortHelper.getRandomAvailableTCPPorts(5);
+    int locatorPort = availablePorts[0];
+    int locatorHttpPort = availablePorts[1];
+    int locatorJmxPort = availablePorts[2];
+    int serverPort1 = availablePorts[3];
+    int serverPort2 = availablePorts[4];
+
+    locatorString = "localhost[" + locatorPort + "]";
+
+    folderForLocator = temporaryFolder.newFolder("locator");
+    File folderForServer1 = temporaryFolder.newFolder("server1");
+    File folderForServer2 = temporaryFolder.newFolder("server2");
+    serverFolders.add(folderForServer1);
+    serverFolders.add(folderForServer2);
+
+    String startLocatorCommand = String.join(SPACE,
+        "start locator",
+        "--name=" + "locator",
+        "--dir=" + folderForLocator.getAbsolutePath(),
+        "--port=" + locatorPort,
+        "--J=-Dgemfire.jmx-manager-start=true",
+        "--J=-Dgemfire.jmx-manager-http-port=" + locatorHttpPort,
+        "--J=-Dgemfire.jmx-manager-port=" + locatorJmxPort);
+
+    String startServer1Command = startServerCommand("server1", serverPort1, folderForServer1);
+    String startServer2Command = startServerCommand("server2", serverPort2, folderForServer2);
+
+    gfshRule.execute(startLocatorCommand, startServer1Command, startServer2Command);
+
+    connectToLocatorCommand = "connect --locator=" + locatorString;
+
+    String deployCommand = functionJarRule.deployCommand();
+    String listFunctionsCommand = "list functions";
+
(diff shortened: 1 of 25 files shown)


geode-core/src/main/java/org/apache/geode/internal/cache/RegionPerfStats.java
```diff
@@ -463,7 +491,7 @@ protected void incEventThreads(int items) {
 
   @Override
   public void incEntryCount(int delta) {
-    stats.incLong(entryCountId, delta);
+    entryCount.addAndGet(delta);
```
pivotal-jbarrett: Why are we removing the region perf stat for entry count?

geode-core/src/main/java/org/apache/geode/internal/cache/RegionPerfStats.java
```diff
@@ -14,27 +14,55 @@
...
     this.cachePerfStats = cachePerfStats;
+    this.meterRegistry = meterRegistry;
+    entryCount = new AtomicLong();
+    entriesGauge = Gauge.builder("member.region.entries", entryCount::get)
+        .description("Current number of entries in the region.")
+        .tag("region.name", regionName)
+        .tag("data.policy", dataPolicy.toString())
+        .baseUnit("entries")
+        .register(meterRegistry);
+    stats.setLongSupplier(entryCountId, entryCount::get);
```
pivotal-jbarrett: Why is this being re-implemented as a supplier?

geode-core/src/main/java/org/apache/geode/internal/cache/RegionPerfStats.java
```diff
@@ -14,27 +14,55 @@
...
 import org.apache.geode.annotations.VisibleForTesting;
+import org.apache.geode.cache.DataPolicy;
 import org.apache.geode.internal.NanoTimer;
 
 class RegionPerfStats extends CachePerfStats {
 
   private final CachePerfStats cachePerfStats;
+  private final MeterRegistry meterRegistry;
+  private final Gauge entriesGauge;
+  private final AtomicLong entryCount;
```
pivotal-jbarrett: Converting from a `LongAdder` to `AtomicLong` will result in poor entry update performance. Unless you need to set specific values you should be using `LongAdder` for multithreaded increments and decrements.

demery-pivotal: Done

demery-pivotal:
> Can you explain why the RegionPerfStats is now keeping a local copy of the entry count? Unless there is a non-stats use for this counter then it should just be kept by stats/meters and not yet another thing.

Here's our thinking:
- Micrometer gauges do not accumulate values. They only fetch values from some source.
- We didn't want to invent a new meter type for this use case… yet. We may revisit this decision in the future as we get more experience linking Geode gauges to Micrometer gauges. (If we did create a new "cumulative gauge" meter type, that would also be "yet another thing.")
- Currently Geode accumulates these increments and decrements only in stats. There is no other available source for this accumulated value.
- We didn't want to configure the meter to fetch from the stat, because that would increase our dependence on stats. Given that we want to deprecate/remove stats eventually, we would eventually have to create some non-stat place to accumulate the increments and decrements. We chose to do that work now instead of deferring it until later.
- So we added a new field to accumulate the increments and decrements, and configured the meter and the stat to get their values from that.

geode-core/src/main/java/org/apache/geode/internal/cache/RegionPerfStats.java
demery-pivotal: To avoid updating two adders on each call, and to localize (to the constructor) the eventual work of removing stats.

geode-core/src/main/java/org/apache/geode/internal/cache/RegionPerfStats.java
demery-pivotal: So that the stat and the meter tracks the same value, without requiring us to update two things (the adder and the stat) on each increment/decrement. This does introduce a small lag into the stat.

geode-core/src/main/java/org/apache/geode/internal/cache/RegionPerfStats.java
pivotal-jbarrett: We had this conversation on the mailing list though and thought the consensus was to leave stats in place just wrap with meters. This approach of moving them out of stats into something else and having meters and stats poll it was clearly not accepted in the discussions.

pivotal-jbarrett:
> * Micrometer gauges do not accumulate values. They only fetch values from some source.
So is the point of this change the this "region perf stats" is the source of authority for the count of items in the region? This doesn't feel like the correct place for this. Shouldn't the region know its current size? Why is it being tracked externally?

> * We didn't want to invent a new meter type for this use case… yet. We may revisit this decision in the future as we get more experience linking Geode gauges to Micrometer gauges. (If we did create a new "cumulative gauge" meter type, that would also be "yet another thing.")
I see, its a mismatch between our stats definition of gauge and Micrometer's definition. 

> * We didn't want to configure the meter to fetch from the stat, because that would increase our dependence on stats. Given that we want to deprecate/remove stats eventually, we would eventually have to create some non-stat place to accumulate the increments and decrements. We chose to do that work now instead of deferring it until later.
> * So we added a new field to accumulate the increments and decrements, and configured the meter and the stat to get their values from that.

Understandable in this gauge case, though I think we need to consider locality and ownership of these gauges. Where you are putting this feels like just another form of stats.

Regions are just `Map` implementations right? `Map` has a `size` method. Shouldn't the gauges just be polling that method?

geode-core/src/main/java/org/apache/geode/internal/cache/RegionPerfStats.java
pivotal-jbarrett: See concerns in whole PR comment.

geode-core/src/main/java/org/apache/geode/internal/cache/RegionPerfStats.java
pivotal-jbarrett: See concerns in whole PR comment.

kirklund:
> Regions are just Map implementations right? Map has a size method. Shouldn't the gauges just be polling that method?

@pivotal-jbarrett PartitionedRegion.size() sends out messages to all the servers hosting buckets for that Region: `LocalRegion.size() -> LocalRegion.entryCount() -> LocalRegionDataView.entryCount(LocalRegion) -> PartitionedRegion.getRegionSize -> PartitionedRegion.entryCount(null) -> PartitionedRegion.entryCount(null, false) -> PartitionedRegion.getSizeRemotely(...) -> SizeMessage`

We didn't think it was a good idea to point the gauge at something that involves messaging.

kirklund:
> > We didn't want to configure the meter to fetch from the stat, because that would increase our dependence on stats. Given that we want to deprecate/remove stats eventually, we would eventually have to create some non-stat place to accumulate the increments and decrements. We chose to do that work now instead of deferring it until later.

> 
> Understandable in this gauge case, though I think we need to consider locality and ownership of these gauges. Where you are putting this feels like just another form of stats.

@pivotal-jbarrett We already tried to add meters outside the stats classes, but there were more people against that than one person (you) against us adding the meters in the stats classes.

demery-pivotal:
> > * We didn't want to configure the meter to fetch from the stat, because that would increase our dependence on stats. Given that we want to deprecate/remove stats eventually, we would eventually have to create some non-stat place to accumulate the increments and decrements. We chose to do that work now instead of deferring it until later.
> 
> Understandable in this gauge case, though I think we need to consider locality and ownership of these gauges. Where you are putting this feels like just another form of stats.

It is. In general, I would prefer to connect each gauge to an existing source of reliable information, rather than relying on an instrumentation class to store the information. In this case, I would love to get the information from the region. But currently we don't know of an inexpensive, reliable source other than the `RegionPerfStats` class.

Do you know of a reliable source of entry counts that we can connect the meter and stat to? Barring that, can you think of a better place than `RegionPerfStats` to accumulate the increments and decrements?

demery-pivotal:
> So is the point of this change the this "region perf stats" is the source of authority for the count of items in the region?

The source of authority was `Statistics`. We have moved the source of authority from there to the instrumentation class.

> This doesn't feel like the correct place for this.

Indeed. Is there an existing place that feels correct? If not, does the instrumentation class seem like a better place or worse place than the `Statistics`?

pivotal-jbarrett:
> Do you know of a reliable source of entry counts that we can connect the meter and stat to? Barring that, can you think of a better place than `RegionPerfStats` to accumulate the increments and decrements?

That's exactly why I am asking. We have either created another ephemeral accounting class or a redundant external state class and both smell bad to me. 

My question about `Map.size` in this really, why isn't that method returning the exact same results of this `LongAdder`? If it does then why not embed this `LongAdder` in the various region classes and update `Map.size` to return the contents of the `LongAdder`. If this new value is meant to reflect the current state of the region but it is being updated outside the region how are we keeping this state in sync, and if the state is in sync why do we need two things holding this state?
If the value is ephemeral, like stats themselves, and there really is no better place to put this sort of thing, then cool.

geode-assembly/src/acceptanceTest/java/org/apache/geode/metrics/RegionEntriesGaugeTest.java
```diff
@@ -0,0 +1,336 @@
...
+
+    commands.add("stop locator --dir=" + folderForLocator.getAbsolutePath());
+
+    gfshRule.execute(commands.toArray(new String[0]));
+  }
+
+  private static String stopServerCommand(File folder) {
+    return "stop server --dir=" + folder.getAbsolutePath();
+  }
+
```
aaronlindsey: Did you consider doing an acceptance test for a LOCAL region?

demery-pivotal:
> My question about `Map.size` in this really, why isn't that method returning the exact same results of this `LongAdder`? If it does then why not embed this `LongAdder` in the various region classes and update `Map.size` to return the contents of the `LongAdder`. If this new value is meant to reflect the current state of the region but it is being updated outside the region how are we keeping this state in sync, and if the state is in sync why do we need two things holding this state?
> If the value is ephemeral, like stats themselves, and there really is no better place to put this sort of thing, then cool.

I haven't found a better place yet.

CachePerfStats.incEntryCount(delta) is called from four places:

- `AbstractRegionMap.incEntryCount(delta)` increments if the map has an "owner" local region.
- `LocalRegion.incTombstoneCount(delta)` decrements the entry count by delta.
- `AbstractRegionEntry.makeTombstone(…)` increments if the entry is new (to offset the decrement that happened when the entry was marked as a tombstone).
- `RegionMapPut.updateSize()` increments if the entry is not a tombstone and not an update.

I can't say what percentage of calls come through each of these methods. But three of the methods are not in any subclass of `Region`, so if we put the adder in the `Region` classes, we would have to add a method specifically to update this count, which is currently used only to report metrics.

After a bit of research (not by any means exhaustive) I can't find a `size()` method that gives a simple answer.

LocalRegion overrides Map.size() to call getDateView().entryCount(this). Each data view class computes the answer differently. And for any given data view class, the entryCount() method gives different answers depending on various conditions.

I have explored only a few of the implementations so far, but I have not yet found a case where the answer is is the same as the unadjusted result of calling Map.size().

demery-pivotal:
We are investigating sources of "entry count" that are more direct and more reliable.

pivotal-jbarrett:
> We are investigating sources of "entry count" that are more direct and more reliable.

I am going to remove my block on this PR. I hope y'all can find a more direct method for getting the current size but if not then this approach is good too.

demery-pivotal:
Revised implementation will be submitted as a separate PR.

PR #3832, opened 2019-07-23
GEODE-7005: Fix Race Condition in Unit Test

jujoramos:
- Added a warn logging message so users know that the distributor
  thread has been interrupted.
- Use awaitility instead of trying to find the daemon thread right
  away, it might have not finished and, as such, the assertions
  would fail.

### For all changes:

### Note:

### DIFF
--- a/geode-wan/src/main/java/org/apache/geode/cache/client/internal/locator/wan/LocatorMembershipListenerImpl.java
+++ b/geode-wan/src/main/java/org/apache/geode/cache/client/internal/locator/wan/LocatorMembershipListenerImpl.java
@@ -307,6 +307,8 @@ public class LocatorMembershipListenerImpl implements LocatorMembershipListener 
                   Thread.sleep(memberTimeout);
                 } catch (InterruptedException e) {
                   Thread.currentThread().interrupt();
+                  logger.warn(
+                      "Locator Membership listener permanently failed to exchange locator information due to interruption.");
                 }
               }
             }
--- a/geode-wan/src/main/java/org/apache/geode/cache/client/internal/locator/wan/LocatorMembershipListenerImpl.java
+++ b/geode-wan/src/main/java/org/apache/geode/cache/client/internal/locator/wan/LocatorMembershipListenerImpl.java
@@ -42,7 +42,7 @@ import org.apache.geode.internal.logging.LoggingThreadFactory;
 public class LocatorMembershipListenerImpl implements LocatorMembershipListener {
   private static final Logger logger = LogService.getLogger();
   static final int LOCATOR_DISTRIBUTION_RETRY_ATTEMPTS = 3;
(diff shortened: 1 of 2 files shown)


geode-wan/src/test/java/org/apache/geode/cache/client/internal/locator/wan/LocatorMembershipListenerTest.java
```diff
@@ -92,9 +94,19 @@ private void verifyMessagesSentBothWays(DistributionLocatorId sourceLocator,
 
   private void joinLocatorsDistributorThread() throws InterruptedException {
     Set<Thread> threads = Thread.getAllStackTraces().keySet();
+
+    // Wait for the daemon thread to be alive.
+    try {
+      await().untilAsserted(() -> assertThat(
```
agingade: In try block, the "threads" is not getting re-set; it will be looking at the same set of entries. You may want to move getting threads in the wait loop.

jujoramos: Great catch @agingade, thanks!.

geode-wan/src/main/java/org/apache/geode/cache/client/internal/locator/wan/LocatorMembershipListenerImpl.java
```diff
@@ -74,6 +74,10 @@ public void setConfig(DistributionConfig config) {
     this.config = config;
   }
 
+  void launchLocatorsDistributorThread(Thread thread) {
```
agingade: It seems its overdone; just to get the start in a new method for testing.
How about moving entire thread launch logic over here:
launchLocatorsDistributorThread(Thread thread) {
// Launch Locators Distributor thread.
    Runnable distributeLocatorsRunnable = new DistributeLocatorsRunnable(config.getMemberTimeout(),
        tcpClient, localLocatorId, localCopy, locator, distributedSystemId);
    ThreadFactory threadFactory =
        new LoggingThreadFactory(LOCATORS_DISTRIBUTOR_THREAD_NAME, true);
    Thread distributeLocatorsThread = threadFactory.newThread(distributeLocatorsRunnable);
    distributeLocatorsThread.start();
}

Since the test is already spying on it; you could verify if this method is called once.

jujoramos: I agree in that a method to just start a thread is too much... but just verifying that the method was called isn't enough, by the time the assertions are executed the actual thread might not have been executed, the test needs an actual reference to the thread to make sure it's `TERMINATED` before executing the asserts.
I'll replace the method with another one, and use `ResultCaptor` instead of `ArgumentCaptor`, that might be better.

PR #3834, opened 2019-07-23
GEODE-6998: Fix for NPE at index update due to GII

mivanac:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/cache/query/internal/index/CompactRangeIndex.java
+++ b/geode-core/src/main/java/org/apache/geode/cache/query/internal/index/CompactRangeIndex.java
@@ -165,7 +165,7 @@ public class CompactRangeIndex extends AbstractIndex {
       // we know in this specific case, that a before op was called and stored oldKey/value
       // we also know that a regular remove won't work due to the entry no longer being present
       // We know the old key so let's just remove mapping from the old key
-      if (oldKeyValue != null) {
+      if (oldKeyValue != null && oldKeyValue.get() != null) {
         indexStore.removeMapping(oldKeyValue.get().getOldKey(), entry);
       } else {
         // rely on reverse map in the index store to figure out the real key
--- a/geode-core/src/main/java/org/apache/geode/cache/query/internal/index/CompactRangeIndex.java
+++ b/geode-core/src/main/java/org/apache/geode/cache/query/internal/index/CompactRangeIndex.java
@@ -165,7 +165,10 @@ public class CompactRangeIndex extends AbstractIndex {
       // we know in this specific case, that a before op was called and stored oldKey/value
       // we also know that a regular remove won't work due to the entry no longer being present
       // We know the old key so let's just remove mapping from the old key
-      if (oldKeyValue != null && oldKeyValue.get() != null) {
+      if (oldKeyValue != null) {
+        if (oldKeyValue.get() == null) {
+          return;
+        }
         indexStore.removeMapping(oldKeyValue.get().getOldKey(), entry);
       } else {
         // rely on reverse map in the index store to figure out the real key
--- a/geode-core/src/test/java/org/apache/geode/cache/query/internal/index/CompactRangeIndexTest.java
+++ b/geode-core/src/test/java/org/apache/geode/cache/query/internal/index/CompactRangeIndexTest.java
@@ -1,0 +1,107 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.apache.geode.cache.query.internal.index;
+
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+import org.junit.Before;
+import org.junit.Test;
+
+import org.apache.geode.Statistics;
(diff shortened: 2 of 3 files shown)


geode-core/src/main/java/org/apache/geode/cache/query/internal/index/CompactRangeIndex.java
```diff
@@ -165,7 +165,7 @@ void removeMapping(RegionEntry entry, int opCode) throws IMQException {
       // we know in this specific case, that a before op was called and stored oldKey/value
       // we also know that a regular remove won't work due to the entry no longer being present
       // We know the old key so let's just remove mapping from the old key
-      if (oldKeyValue != null) {
+      if (oldKeyValue != null && oldKeyValue.get() != null) {
```
jhuynh1: Is there a test case that isn't mocked that shows this problem?  It might be nice to know how the value in oldKeyValue is null.  The concern is that if we hit the other branch of code, the index updates will have a very noticeable decline in performance.  We would be crawling the entire index for every time the oldKeyValue.get() is returning null.

PartitionedRegionCompactRangeIndexDUnitTest has a few tests that were added when this line of code was added.  If we had something similar (even just for testing), we might be able to find a different fix and avoid the crawling of the entire index.

mivanac: Problem can be reproduced, as described in issue:
On REPLICATE_PERSISTENT region with defined indexes, in configuration with 2 servers, continues traffic (put, delete) is run. After one server is restarted, NullPointerException occurres. With UT is simulated scenario observed from logs.

jhuynh1:
I modified PartitionedRegionCompactRangeIndexDUnitTest locally to see if I could repro the issue.  It looks like this occurs if the current value in the cache is already a tombstone and the new value also happens to be a tombstone.  I think this change should be fine

geode-core/src/test/java/org/apache/geode/cache/query/internal/index/CompactRangeIndexTest.java
```diff
@@ -0,0 +1,107 @@
...
+import org.apache.geode.distributed.internal.InternalDistributedSystem;
+import org.apache.geode.internal.cache.CachePerfStats;
+import org.apache.geode.internal.cache.GemFireCacheImpl;
+import org.apache.geode.internal.cache.LocalRegion;
+import org.apache.geode.internal.cache.RegionEntry;
+
+public class CompactRangeIndexTest {
+
+  private CompactRangeIndex index;
+  LocalRegion region = mock(LocalRegion.class);
```
kirklund: Please change all fields to `private`. The access modifier should always be the most restrictive possible, and we should follow the same guidelines in tests as we do in product code.

kirklund: I also recommend you consider using the MockitoRule with STRICT_STUBS:
```
  @Rule
  public MockitoRule mockitoRule = MockitoJUnit.rule().strictness(STRICT_STUBS);
```
If you grep through Geode test code for this you'll find quite a few tests. The advantage of using this rule with STRICT_STUBS is that it will do some additional validating for you that really helps while creating or modifying the test. It will let you know if you have any mocks or when-clauses that are not necessary so you can delete them and streamline your test.

geode-core/src/test/java/org/apache/geode/cache/query/internal/index/CompactRangeIndexTest.java
```diff
@@ -0,0 +1,107 @@
...
+import org.apache.geode.cache.query.internal.DefaultQueryService;
+import org.apache.geode.cache.query.internal.index.AbstractIndex.InternalIndexStatistics;
+import org.apache.geode.cache.query.types.ObjectType;
+import org.apache.geode.distributed.internal.InternalDistributedSystem;
+import org.apache.geode.internal.cache.CachePerfStats;
+import org.apache.geode.internal.cache.GemFireCacheImpl;
+import org.apache.geode.internal.cache.LocalRegion;
+import org.apache.geode.internal.cache.RegionEntry;
+
+public class CompactRangeIndexTest {
```
kirklund: [Just a comment] You might want to consider using the MockitoRule with STRICT_STUBS:
```
  @Rule
  public MockitoRule mockitoRule = MockitoJUnit.rule().strictness(STRICT_STUBS);
```
You can grep through Geode test code for examples. The advantage of using this rule with STRICT_STUBS is that it will do some additional validating for you that really helps while creating or modifying the test. It will let you know if you have any mocks or when-clauses that are not necessary so you can delete them and streamline your test. Sometimes you'll find a mock or when-clause that should be moved to a single test.

jhuynh1:
@agingade What do you think about this change?  In particular, how do you feel about no-op'ing instead of crawling the index.

agingade:
@jhuynh1 @mivanac  
Jason, you are right no-oping is the right way to go here. We are getting into this situation, due to single GII processing thread updating index for batch of entries. And processing the entry that is removed. Since we don't store any removed-entries (TOMBSTONEs) in index, we could just return if oldKeyValue returns null.
 // We know the old key so let's just remove mapping from the old key
      if (oldKeyValue != null) {
          if (oldKeyValue.get() == null) {
            return;
          }
        indexStore.removeMapping(oldKeyValue.get().getOldKey(), entry);
      }

We should add few DUnit test, reproducing the issue.

geode-core/src/test/java/org/apache/geode/cache/query/internal/index/CompactRangeIndexTest.java
mivanac: Thanks for comments.

mivanac:
Thanks for comments. Code is updated. Do you have any proposals for DUnit test.

mivanac:
DUnit test added.

jhuynh1:
I'm currently working a dunit test but it might take another day to iron out.  If you want, I can create a new PR when it is done and reference this ticket number.

geode-core/src/test/java/org/apache/geode/cache/query/internal/index/CompactRangeIndexTest.java
mivanac: Updated

mivanac:
Thanks for help with DUnit test.

mivanac:
Could we merge this PR to develop now, and additionally add test, so this PR is included in 1.10 release.

geode-core/src/test/java/org/apache/geode/cache/query/internal/index/CompactRangeIndexTest.java
mivanac: Updated

PR #3840, opened 2019-07-24
GEODE-7006: Fixes function execution by id with transactions.

pivotal-jbarrett:
* Fixes test to force and wait for PR metadata to update.
* Run each test via function execution by id and object.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/cache/client/internal/ServerRegionProxy.java
+++ b/geode-core/src/main/java/org/apache/geode/cache/client/internal/ServerRegionProxy.java
@@ -756,13 +756,14 @@ public class ServerRegionProxy extends ServerProxy implements ServerRegionDataAc
         serverRegionExecutor, resultCollector, hasResult, isHA, optimizeForWrite);
 
     int retryAttempts = pool.getRetryAttempts();
+    boolean inTransaction = TXManagerImpl.getCurrentTXState() != null;
 
     final Supplier<AbstractOp> executeRegionFunctionOpSupplier =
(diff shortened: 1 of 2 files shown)


geode-core/src/main/java/org/apache/geode/cache/client/internal/ServerRegionProxy.java
```diff
@@ -756,13 +756,14 @@ public void executeFunction(String functionId,
...
     int retryAttempts = pool.getRetryAttempts();
+    boolean inTransaction = TXManagerImpl.getCurrentTXState() != null;
 
     final Supplier<AbstractOp> executeRegionFunctionOpSupplier =
         () -> new ExecuteRegionFunctionOp.ExecuteRegionFunctionOpImpl(region.getFullPath(),
             functionId, serverRegionExecutor, resultCollector, hasResult, isHA,
             optimizeForWrite, true, timeoutMs);
 
-    if (pool.getPRSingleHopEnabled()) {
+    if (pool.getPRSingleHopEnabled() && !inTransaction) {
```
agingade: This basically avoids using singleHop for transactional operation; the reason for that is, the single hop ops are performed in executor thread, losing the transaction context. Options could be to see, if tx context can be passed to the singlehop executors or to move the logic from SingleHopOperationCallable to separate method and call it in the same execution thread. This could be done as part of separate ticket/issue.

pivotal-jbarrett: Yeah, changing the logic should be a different ticket. This ticket is concerned with making the two code paths consistent. The logic came from https://issues.apache.org/jira/browse/GEODE-6016.

PR #3843, opened 2019-07-24
DO NOT REVIEW: GEODE-7010: Replace static globals in CachePerfStats with StatisticsClock

kirklund:
GEODE-7010: Replace static globals in CachePerfStats with StatisticsClock

* Create StatisticsClock during Cache creation
* Configure isEnabled() using "enable-time-statistics" config property
* Inject StatisticsClock dependency via constructor in classes that use it

I plan to make a second pass to hopefully inject StatisticsClock dependency
into more classes and remove getStatisticsClock methods from Cache and
Region.


kirklund:
Apparently LocatorMembershipListenerTest is a flaky test. I ran some tests and I can get two tests in the class to fail occasionally.

jujoramos:
Hey @kirklund,
The flakiness of `LocatorMembershipListenerTest` should be fixed now, I've merged the changes into `develop` earlier today (~3hs ago). Cheers.

kirklund:
@jujoramos good to know, I'll rebase soon. Thanks!

PR #3846, opened 2019-07-24
[DO NOT REVIEW] GEODE-7001: region size experiment

demery-pivotal:
Region entry gauge stat and meter get their values from a new `getLocalSize()` method added to `InternalRegion`.

We will want to run some performance tests on this before committing. The `getLocalSize()` method on `LocalRegion` iterates over the segments in the abstract region map. The `getLocalSize()` method in `PartitionedRegion` iterates over the local buckets, each of which iterates over its abstract region map. This computation may or may not be expensive. We want to find out.

### For all changes:


pivotal-jbarrett:
@dhemery I think the place to focus is `CustomEntryConcurrentHashMap::size`. It will attempt to get the count without locking but under update contention will resort to locking all the segments. I have a branch with some JMH benchmarks for this class already. If you want I can share and you can add one for calling size on an interval and see how that impacts puts.

pivotal-jbarrett:
@dhemery I made a small modification to the JMH I already had for comparing performance of `CustomEntryConcurrentHashMap`. I created a background thread that called `size` at 0ms (disabled, 1ms and 1s intervals. The benchmark used 100 threads to put or get concurrently into the maps. The `size` call had no measurable effect on the the put or get performance of `CustomEntryConcurrentHashMap`, which tops out on my machine 30M puts/s and 30M gets/s. It also shows no measurable performance impact on `ConcurrentHashMap`. So baring any serious locking going on at the `RegionBucket` down to the `CustomEntryConcurrentHashMap` you all should be good to go.

geode-core/src/main/java/org/apache/geode/internal/cache/LocalRegion.java
```diff
@@ -8257,6 +8258,11 @@ public int size() {
     }
   }
 
+  @Override
+  public int getLocalSize() {
+    return getRegionMap().size() - tombstoneCount.get();
```
metatype: Nice!

demery-pivotal:
Re-submitted as #3850

PR #3849, opened 2019-07-25
GEODE-7018: Enable hostname validation for SSL over NIO

sboorlagadda:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/integrationTest/java/org/apache/geode/internal/net/SSLSocketHostNameVerificationIntegrationTest.java
+++ b/geode-core/src/integrationTest/java/org/apache/geode/internal/net/SSLSocketHostNameVerificationIntegrationTest.java
@@ -196,7 +196,6 @@ public class SSLSocketHostNameVerificationIntegrationTest {
                 ByteBuffer.allocate(500),
                 new BufferPool(mock(DMStats.class)));
       } catch (Throwable throwable) {
-        throwable.printStackTrace(System.out);
         serverException = throwable;
       } finally {
         if (engine != null && socket != null) {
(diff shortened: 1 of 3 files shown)


geode-core/src/integrationTest/java/org/apache/geode/internal/net/SSLSocketHostNameVerificationIntegrationTest.java
```diff
@@ -0,0 +1,228 @@
...
+        SocketCreator sc = SocketCreatorFactory.getSocketCreatorForComponent(CLUSTER);
+        engine =
+            sc.handshakeSSLSocketChannel(socket.getChannel(),
+                sc.createSSLEngine(this.localHost.getHostName(), 1234),
+                timeoutMillis,
+                false,
+                ByteBuffer.allocate(500),
+                new BufferPool(mock(DMStats.class)));
+      } catch (Throwable throwable) {
+        throwable.printStackTrace(System.out);
```
bschuchardt: No need to print this stack trace

geode-core/src/main/java/org/apache/geode/internal/net/SocketCreator.java
```diff
@@ -933,6 +933,24 @@ public NioSslEngine handshakeSSLSocketChannel(SocketChannel socketChannel, SSLEn
       BufferPool bufferPool)
       throws IOException {
     engine.setUseClientMode(clientSocket);
+    if (!clientSocket) {
+      engine.setNeedClientAuth(sslConfig.isRequireAuth());
+    }
+
+    if (clientSocket) {
+      if (sslConfig.doEndpointIdentification()) {
```
dschneider-pivotal: this whole "if/else" statement is also in this class in configureClientSSLSocket. Should you extract it into a private method so both places use the same code?

sboorlagadda: Thanks @dschneider-pivotal I have refactored it to a method. Both `SSLSocket` and `SSLEngine` returns a copy of SSLParameters to I could not completely do it inside the method. This new method returns params which are re-set into socket or engine.

geode-core/src/integrationTest/java/org/apache/geode/internal/net/SSLSocketHostNameVerificationIntegrationTest.java
sboorlagadda: Thanks @bschuchardt. Removed the unwanted output

PR #3850, opened 2019-07-26
GEODE-7001: Add region entry count gauge

demery-pivotal:
Add 'member.region.entries' gauge in `RegionPerfStats`:
- Add region.name and data.policy tags
- Configure the member.region.entries gauge to fetch from new `getLocalSize()` method (see below).
- Configure the 'entryCount' stat to be supplied by new `getLocalSize()` method.

Add `getLocalSize()` method to `InternalRegion`, and implement it in `LocalRegion` and `PartitionedRegion`
- In `LocalRegion`, the method returns the region map size minus the number of tombstones.
- In `PartitionedRegion`, the method sums the local sizes of all local bucket regions.

Also refactored the following:
- Reorganize `CachePerfStats`/`RegionPerfStats` constructors
- Remove `CachePerfStats.getEntries()`
- Remove use of `CachePerfStats.getEntries()` from `PartitionedRegionStatus`.
- Add `@Override` to `ValidatingStatisticsType` methods in `StatisticsTypeImpl`
- Extracted `invokeSuppliers()` to new `SuppliableStatistics` interface and rename as `updateSuppliedValues()`
- Move responsibility to prepend 'RegionStats-' onto region statistics textID into the classes that create `RegionPerfStats` and `CachePerfStats`.

Co-authored-by: Michael Oleske <moleske@pivotal.io>
Co-Authored-By: Mark Hanson <mhanson@pivotal.io>
Co-Authored-By: Aaron Lindsey <alindsey@pivotal.io>

### For all changes:

### DIFF
--- a/geode-assembly/src/acceptanceTest/java/org/apache/geode/metrics/RegionEntriesGaugeTest.java
+++ b/geode-assembly/src/acceptanceTest/java/org/apache/geode/metrics/RegionEntriesGaugeTest.java
@@ -261,7 +261,6 @@ public class RegionEntriesGaugeTest {
     int expectedNumberOfEntries = numberOfEntries * totalNumberOfCopies;
 
     await()
-        .atMost(10, TimeUnit.SECONDS)
         .untilAsserted(() -> {
           GfshExecution execution =
               gfshRule.execute(connectToLocatorCommand, getGaugeValueCommand);
--- a/geode-assembly/src/acceptanceTest/java/org/apache/geode/metrics/RegionEntriesGaugeTest.java
+++ b/geode-assembly/src/acceptanceTest/java/org/apache/geode/metrics/RegionEntriesGaugeTest.java
@@ -1,0 +1,380 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package org.apache.geode.metrics;
+
+import static java.lang.Integer.parseInt;
+import static java.util.Arrays.asList;
+import static java.util.stream.Collectors.toList;
+import static org.apache.geode.cache.RegionShortcut.LOCAL;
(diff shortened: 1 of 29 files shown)


geode-core/src/main/java/org/apache/geode/internal/cache/CachePerfStats.java
```diff
@@ -1310,10 +1323,6 @@ public void incEntryCount(int delta) {
     stats.incLong(entryCountId, delta);
   }
 
-  public long getEntries() {
-    return stats.getLong(entryCountId);
-  }
-
```
aaronlindsey: This seems fine, but I'm just curious what the reasoning was for removing this. Was there a problem with the way it was retrieving the entry count from the stats?

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
```diff
@@ -4838,7 +4838,7 @@ public void fetchAllRemoteEntries(
   /**
    * Test Method: Get the number of entries in the local data store.
    */
-  public long getLocalSize() {
+  public long getLocalSizeForTest() {
```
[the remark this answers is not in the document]
demery-pivotal: We need to rename the method so that we could use its original name for our new functionality. We saw that numerous methods in Geode ended with 'ForTest()' and went with that. Then we figured that the annotation would be superfluous.

geode-core/src/main/java/org/apache/geode/internal/cache/CachePerfStats.java
demery-pivotal: In general, we're hoping to discourage using stats for anything other than publishing metrics. Removing the getter helps with that. In this case, we had an additional (minor) concern that someone might call this method expecting it to have an up-to-date value. If the `CachePerfStats` is actually a `RegionPerfStats`, this stat's value will lag, because the value is refreshed only periodically during sampling.

geode-assembly/src/acceptanceTest/java/org/apache/geode/metrics/RegionEntriesGaugeTest.java
```diff
@@ -0,0 +1,380 @@
...
+    region.destroy(keys.get(0));
+
+    String getGaugeValueCommand = memberRegionEntryGaugeValueCommand(regionName);
+
+    int numberOfEntries = keys.size() - 1;
+    int totalNumberOfCopies = numberOfRedundantCopies + 1;
+    int expectedNumberOfEntries = numberOfEntries * totalNumberOfCopies;
+
+    await()
+        .atMost(10, TimeUnit.SECONDS)
```
moleske: I think we've been trying to keep the default timeout, but this seems alright since the `gfsh` command should be quick

kirklund: We probably missed removing this 10 SECOND timeout. We should use the default.

moleske: just edited the file directly in github since it was such a small change

PR #3855, opened 2019-07-29
GEODE-6799: better gfsh error message when missing required parameters

onichols-pivotal:
This improves gfsh error in cases where required options are not given.

Before:
```describe offline-disk-store --name=foo```
```You should specify option (--disk-dirs, --pdx, --region) for this command```

After: 
```describe offline-disk-store --name=foo```
```You should specify option (--disk-dirs, --pdx, --region) for this command```
```  --disk-dirs=  is required```
```Use "help describe offline-disk-store" (without the quotes) to display detailed usage information.```

Note that the unhelpful first part that lists all possible options, including non-required, comes from a private method in spring shell, so there is no way to suppress it.  But hopefully the added helpful information will be enough to make the user less confused.


onichols-pivotal:
@jmelchio

PR #3857, opened 2019-07-29
GEODE-7022: Add SSL Alias Support for JMX Clients

jujoramos:
- Added tests.
- Fixed minor warnings.
- Use custom RMIClientSocketFactory that supports Geode SSL settings.

### For all changes:

### Note:


geode-core/src/main/java/org/apache/geode/management/internal/ContextAwareSSLRMIClientSocketFactory.java
```diff
@@ -44,18 +50,26 @@
...
       socketCreator =
           SocketCreatorFactory.getSocketCreatorForComponent(SecurableCommunicationChannel.JMX);
+      return socketCreator.connectForClient(host, port, 0);
     } catch (Exception exception) {
-      /*
-       * In the context of a gfsh VM the javax.net.ssl properties are used to configure SSL
-       * appropriately - see the constructor for JMXOperationInvoker.
-       */
-      return defaultFactory.createSocket(host, port);
+      try {
```
kirklund: I recommend extracting the nested try-catch to a new method. This example is small enough that it might not matter much, but usually that will be easier to read/maintain.

geode-core/src/main/java/org/apache/geode/management/internal/cli/shell/JmxOperationInvoker.java
```diff
@@ -407,7 +374,7 @@ public static String checkAndConvertToCompatibleIPv6Syntax(String hostAddress) {
 
   @Override
   public void handleNotification(Notification notification, Object handback) {
-    if (JMXConnectionNotification.class.isInstance(notification)) {
+    if (notification instanceof JMXConnectionNotification) {
```
kirklund: Just curious why you prefer `notification instanceof JMXConnectionNotification` over the newer syntax of `JMXConnectionNotification.class.isInstance(notification)`.

geode-web/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/ConnectCommandWithSSLMultiKeyTest.java
```diff
@@ -0,0 +1,204 @@
...
+public class ConnectCommandWithSSLMultiKeyTest {
+
+  @ClassRule
+  public static CleanupDUnitVMsRule cleanupDUnitVMsRule = new CleanupDUnitVMsRule();
+
+  private static final String multiKeyTrustStore;
+  private static final String multiKeyClientKeyStore;
+  private static final String multiKeyServerKeyStore;
+
+  static {
```
kirklund: Can you just move all of this static initializer type stuff to `@BeforeClass` and `@AfterClass`? That would be much more standard for a JUnit test like this. `createTempFileFromResource` should still behave fine as well.

geode-core/src/main/java/org/apache/geode/management/internal/cli/shell/JmxOperationInvoker.java
jujoramos: No particular reason at all... I guess old habits die hard, and the fact that `IntelliJ` flags `JMXConnectionNotification.class.isInstance` as a warning made me finally decide to use the old way.

geode-core/src/main/java/org/apache/geode/management/internal/ContextAwareSSLRMIClientSocketFactory.java
jujoramos: Agreed. In this particular case (which is short and readable) though, I'd prefer to leave the implementation as it is right now.

echobravopapa: Like the straight forward approach, is there a specific Exception type we could catch here for this case?

echobravopapa:
@jujoramos after your resolve @kirklund items looks good at the core

geode-core/src/main/java/org/apache/geode/management/internal/ContextAwareSSLRMIClientSocketFactory.java
jujoramos: Hey @echobravopapa,
We could catch `GemFireConfigException` (it extends `RuntimeException`), but I wanted to  prevent accidentally breaking backward compatibility... in previous versions we always ended up using the default `SslRMIClientSocketFactory` if any `Exception` was thrown when using our own socket factory, so I just wanted to respect that in my changes. Let me know if you think otherwise.

echobravopapa: that's a good point; can we verify if this would be an issue?

geode-web/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/ConnectCommandWithSSLMultiKeyTest.java
```diff
@@ -0,0 +1,199 @@
...
+    // keytool -certreq -alias server -keystore server.jks -file server.csr
+    // # Create a self-signed cert for the CA
+    // openssl req -new -x509 -key ca.key -out ca.crt
+    // # Sign them
+    // openssl x509 -req -in client.csr -CA ca.crt -CAkey ca.key -out client.crt -CAcreateserial
+    // openssl x509 -req -in server.csr -CA ca.crt -CAkey ca.key -out server.crt -CAcreateserial
+    // # Import both of these leaf certs into a trust store but don't import CA
+    // keytool -importcert -keystore trusted.jks -file client.crt -alias client
+    // keytool -importcert -keystore trusted.jks -file server.crt -alias server
+    // # Import the CA cert and the signed csrs back into the server and client keystores
```
sboorlagadda: Why was the CA imported into the key store? I guess to make it a multi key store. Just curious if there were other reasons.

geode-web/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/ConnectCommandWithSSLMultiKeyTest.java
```diff
@@ -0,0 +1,199 @@
...
+    // keytool -importcert -file ca.crt -keystore server.jks -storepass passw0rd -alias ca
+    // keytool -importcert -file server.crt -keystore server.jks -storepass passw0rd -alias server
+    // keytool -importcert -file ca.crt -keystore client.jks -storepass passw0rd -alias ca
+    // keytool -importcert -file client.crt -keystore client.jks -storepass passw0rd -alias client
+    String multiKeyTrustStore = createTempFileFromResource(ConnectCommandWithSSLMultiKeyTest.class,
```
sboorlagadda: This is more of an FYI. We have a way to programmatically create keystore and truststore, but this utility is not capable of signing through CA. You can look for usages of `geode-junit/src/main/java/org/apache/geode/cache/ssl/CertStores.java`.

geode-web/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/ConnectCommandWithSSLMultiKeyTest.java
```diff
@@ -0,0 +1,199 @@
...
+  public GfshCommandRule gfsh = new GfshCommandRule();
+
+  @Rule
+  public TemporaryFolder temporaryFolder = new TemporaryFolder();
+
+  @ClassRule
+  public static ClusterStartupRule clusterRule = new ClusterStartupRule();
+
+  @BeforeClass
+  public static void beforeClass() {
```
sboorlagadda: What if user provides multi-key keystore and fails to set alias?

geode-web/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/ConnectCommandWithSSLMultiKeyTest.java
jdeppe-pivotal: The CA needs to be available when importing a signed certificate. It could be deleted afterwards.

geode-web/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/ConnectCommandWithSSLMultiKeyTest.java
```diff
@@ -0,0 +1,199 @@
...
+    // keytool -certreq -alias client -keystore client.jks -file client.csr
+    // keytool -certreq -alias server -keystore server.jks -file server.csr
+    // # Create a self-signed cert for the CA
+    // openssl req -new -x509 -key ca.key -out ca.crt
+    // # Sign them
+    // openssl x509 -req -in client.csr -CA ca.crt -CAkey ca.key -out client.crt -CAcreateserial
+    // openssl x509 -req -in server.csr -CA ca.crt -CAkey ca.key -out server.crt -CAcreateserial
+    // # Import both of these leaf certs into a trust store but don't import CA
+    // keytool -importcert -keystore trusted.jks -file client.crt -alias client
+    // keytool -importcert -keystore trusted.jks -file server.crt -alias server
```
jdeppe-pivotal: I think this is confusing and not how certificates are typically handled. You should just need to import the `ca.crt` into the trust store and use that.

geode-web/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/ConnectCommandWithSSLMultiKeyTest.java
jujoramos: Thanks for the tip @sboorlagadda, I'll have a look and see if I can use it to simplify the tests and reduce the amount of files checked in.

geode-web/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/ConnectCommandWithSSLMultiKeyTest.java
jujoramos: Then the connection will fail, we don't need to handle that scenario.

geode-web/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/ConnectCommandWithSSLMultiKeyTest.java
jujoramos: Thanks @jdeppe-pivotal, 
I'll try to simplify the tests before merging the change.
Cheers.

jujoramos:
Thanks @jdeppe-pivotal,
I've changed the test to be more simple in my last commit, instead of multiple `key stores` and `trust stores` with/without the `CA` certificate there's now one single certificate in the `trust store` and multiple keys with aliases in the `key store`, which allows us to specifically test the correct usage of the the configured `alias` when obtaining the key. I'll merge the changes later today.

PR #3859, opened 2019-07-29
GEODE-7010: Replace static globals in CachePerfStats with StatisticsClock

kirklund:
GEODE-7010: Replace static globals in CachePerfStats with StatisticsClock

Create StatisticsClock during Cache creation with isEnabled configured by
the "enable-time-statistics" config property.

Inject StatisticsClock dependency via constructor in classes that use it.

This replaces the mutable global static in CachePerfStats.

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/AbstractRegion.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/AbstractRegion.java
@@ -15,6 +15,7 @@
 package org.apache.geode.internal.cache;
 
 import static org.apache.geode.internal.cache.LocalRegion.InitializationLevel.ANY_INIT;
+import static org.apache.geode.internal.statistics.StatisticsClockFactory.disabledClock;
 
 import java.io.DataInput;
 import java.io.DataOutput;
@@ -94,6 +95,7 @@ import org.apache.geode.internal.cache.extension.ExtensionPoint;
 import org.apache.geode.internal.cache.extension.SimpleExtensionPoint;
 import org.apache.geode.internal.cache.snapshot.RegionSnapshotServiceImpl;
 import org.apache.geode.internal.logging.LogService;
+import org.apache.geode.internal.statistics.StatisticsClock;
 import org.apache.geode.internal.util.ArrayUtils;
 import org.apache.geode.pdx.internal.PeerTypeRegistration;
 
@@ -274,10 +276,14 @@ public abstract class AbstractRegion implements InternalRegion, AttributesMutato
 
   private final PoolFinder poolFinder;
 
+  private final StatisticsClock statisticsClock;
+
   /** Creates a new instance of AbstractRegion */
   protected AbstractRegion(InternalCache cache, RegionAttributes<?, ?> attrs, String regionName,
-      InternalRegionArguments internalRegionArgs, PoolFinder poolFinder) {
+      InternalRegionArguments internalRegionArgs, PoolFinder poolFinder,
+      StatisticsClock statisticsClock) {
     this.poolFinder = poolFinder;
+    this.statisticsClock = statisticsClock;
 
     this.cache = cache;
     serialNumber = DistributionAdvisor.createSerialNumber();
@@ -387,6 +393,7 @@ public abstract class AbstractRegion implements InternalRegion, AttributesMutato
 
   @VisibleForTesting
   AbstractRegion() {
+    statisticsClock = disabledClock();
     cache = null;
     serialNumber = 0;
     isPdxTypesRegion = false;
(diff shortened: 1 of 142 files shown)


kirklund:
Previous draft PR passed GREEN: https://github.com/apache/geode/pull/3843

geode-core/src/main/java/org/apache/geode/internal/cache/AbstractRegion.java
```diff
@@ -1867,6 +1874,17 @@ public void incRecentlyUsed() {
     // nothing
   }
 
+  /**
+   * Only subclasses of {@code AbstractRegion} should use this supplier to acquire the
+   * {@code StatisticsClock}.
+   *
+   * <p>
+   * Please do not use this accessor from any class other than a Region.
+   */
```
demery-pivotal: What's the danger? (I don't necessarily need an explanation in the comment, but I am puzzled about this admonition.)

geode-core/src/main/java/org/apache/geode/internal/cache/AbstractUpdateOperation.java
```diff
@@ -169,7 +169,7 @@ public static boolean doPutOrCreate(LocalRegion rgn, EntryEventImpl ev, long las
       // from this message.
       if (doUpdate) {
         if (!ev.isLocalInvalid()) {
-          final long startPut = CachePerfStats.getStatTime();
+          final long startPut = rgn.getCachePerfStats().getTime();
```
demery-pivotal: What makes this preferable to `rgn.getStatisticsClock().getTime()`? (I realize that would violate your admonition in the `AbstractRegion.getStatisticsClock()` javadoc, but I don't yet understand the reason for that admonition.)

geode-core/src/main/java/org/apache/geode/internal/cache/AbstractRegion.java
kirklund: Not really any danger. I just really wanted every class to have its own `statisticsClock` field. So this is the field for `AbstractRegion` and I wanted to use an accessor from the subclasses instead of directly accessing the field. But I didn't want people to see this and think `oh, I'll just call region.getStatisticsClock() instead of injecting it into the constructor of this new/old class`. Maybe I should instead introduce a new annotation `@VisibleForSubclasses` or something like that?

geode-core/src/main/java/org/apache/geode/internal/cache/AbstractUpdateOperation.java
kirklund: There are two reasons. First, I wanted to limit `AbstractRegion.getStatisticsClock()` to being invoked only by subclasses of `AbstractRegion` but unfortunately `protected` won't protect it since everything under the sun lives in the package `cache`.

Also, many classes already need to invoke `region.getCachePerfStats()` so it's a well known accessor unlike the new `getStatisticsClock()`. The following pattern is really common throughout our code (especially `Operation` classes):
```
long start = rgn.getCachePerfStats().getTime();
try {
  // perform operation
} finally {
  rgn.getCachePerfStats().endOperation(start);
}
```
Some of the operations also follow the following variation, which isn't as clear but if you know of the above then it becomes natural to use `rgn.getCachePerfStats().getTime()`:
```
long start = rgn.getCachePerfStats().getTime();
performOperation(start);
```
I realize it would be cleaner to inject `CachePerfStats` into each of these classes that need to start/end stats that are defined in `CachePerfStats and that's exactly what I'd like to see happen (just not for this PR which is already quite long).

kirklund: Note: I also kept the list of calls to `cache.getStatisticsClock()` very slim on purpose. I tried to invoke that just when constructing and passing in `StatisticsClock` to various factory type classes, entities that the cache owns, and entities that the cache is essentially a factory for (such as regions).

kirklund:
I'm going to hold off resolving conflicts until code reviews are completed. I intend to resolve with a rebase and then do a force push. Then I'll wait for one more green precheckin before hopefully merging it.

geode-core/src/main/java/org/apache/geode/internal/cache/AbstractRegion.java
kirklund: Or maybe `@ProtectedForSubclass`

geode-core/src/main/java/org/apache/geode/internal/cache/AbstractUpdateOperation.java
demery-pivotal: Got it. Thanks for the explanations.

PR #3860, opened 2019-07-30
GEODE-6804: add tab completion for deploy --jar= and --dir=

onichols-pivotal:
Before, no tab completion was available, in contrast to other commands like `start server --properties-file=`

After, filesystem-based completion is available, with the following rules:
- `--jars` supports a comma-separated list of .jar files, although you must type the comma to start the next completion cycle
- `--dir` will avoid suggesting directories that don't contain any .jar files

During tab completion of either jars or dirs, it will look ahead one directory level to see if it can filter out directories that contain neither jars nor further subdirectories.

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/management/internal/cli/converters/JarDirPathConverter.java
+++ b/geode-core/src/main/java/org/apache/geode/management/internal/cli/converters/JarDirPathConverter.java
@@ -1,0 +1,89 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.apache.geode.management.internal.cli.converters;
+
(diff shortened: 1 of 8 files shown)


onichols-pivotal:
@jmelchio

geode-core/src/main/java/org/apache/geode/management/internal/cli/converters/JarDirPathConverter.java
```diff
@@ -0,0 +1,89 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+
+import org.springframework.shell.core.Completion;
+import org.springframework.shell.core.Converter;
+import org.springframework.shell.core.MethodTarget;
+
+import org.apache.geode.management.cli.ConverterHint;
+
+public class JarDirPathConverter implements Converter<String> {
```
jmelchio: Unit test missing for Class.

geode-core/src/main/java/org/apache/geode/management/internal/cli/converters/JarFilesPathConverter.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+
+import org.springframework.shell.core.Completion;
+import org.springframework.shell.core.Converter;
+import org.springframework.shell.core.MethodTarget;
+
+import org.apache.geode.management.cli.ConverterHint;
+
+public class JarFilesPathConverter implements Converter<String[]> {
```
jmelchio: Unit test missing for Class.

geode-core/src/main/java/org/apache/geode/management/internal/cli/converters/JarDirPathConverter.java
dschneider-pivotal: unit test added

geode-core/src/main/java/org/apache/geode/management/internal/cli/converters/JarFilesPathConverter.java
dschneider-pivotal: unit test added

PR #3861, opened 2019-07-30
GEODE-7024: raise a Java exception on non-success response from REST API

onichols-pivotal:
Before:
the server-side and client-side implementations of `ClusterManagementService` reported errors differently (by throwing various exceptions on the server side, but returning a status code on the client side).

After:
The goal is to be able to code to the `ClusterManagementService` interface without regard to server vs client environment. To make that possible, `ClientClusterManagementService` now raises `ClusterManagemementException` for any non-successful status, and `LocatorClusterManagementService` now wraps throw exceptions in `ClusterManagemementException` where necessary to be consistent with the client side.  REST clients should continue to receive the same status codes as they always did.

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/management/internal/api/LocatorClusterManagementService.java
+++ b/geode-core/src/main/java/org/apache/geode/management/internal/api/LocatorClusterManagementService.java
@@ -17,6 +17,10 @@
 
 package org.apache.geode.management.internal.api;
 
+import static org.apache.geode.management.api.ClusterManagementResult.StatusCode.ENTITY_NOT_FOUND;
+import static org.apache.geode.management.api.ClusterManagementResult.StatusCode.ERROR;
+import static org.apache.geode.management.api.ClusterManagementResult.StatusCode.ILLEGAL_ARGUMENT;
+
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -46,10 +50,12 @@ import org.apache.geode.distributed.DistributedMember;
 import org.apache.geode.internal.cache.InternalCache;
 import org.apache.geode.internal.cache.execute.AbstractExecution;
 import org.apache.geode.internal.logging.LogService;
+import org.apache.geode.management.api.ClusterManagementException;
 import org.apache.geode.management.api.ClusterManagementListOperationsResult;
 import org.apache.geode.management.api.ClusterManagementListResult;
 import org.apache.geode.management.api.ClusterManagementOperation;
 import org.apache.geode.management.api.ClusterManagementOperationResult;
+import org.apache.geode.management.api.ClusterManagementRealizationException;
 import org.apache.geode.management.api.ClusterManagementRealizationResult;
 import org.apache.geode.management.api.ClusterManagementResult;
 import org.apache.geode.management.api.ClusterManagementResult.StatusCode;
@@ -72,7 +78,6 @@ import org.apache.geode.management.internal.configuration.validators.Configurati
 import org.apache.geode.management.internal.configuration.validators.GatewayReceiverConfigValidator;
 import org.apache.geode.management.internal.configuration.validators.MemberValidator;
 import org.apache.geode.management.internal.configuration.validators.RegionConfigValidator;
-import org.apache.geode.management.internal.exceptions.EntityNotFoundException;
 import org.apache.geode.management.internal.operation.OperationHistoryManager;
 import org.apache.geode.management.internal.operation.OperationHistoryManager.OperationInstance;
 import org.apache.geode.management.internal.operation.OperationManager;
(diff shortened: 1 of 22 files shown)


onichols-pivotal:
Possibly we should go the other direction instead and change `LocatorClusterManagementService` to catch all exception and return a status object?

geode-management/src/main/java/org/apache/geode/management/internal/ClientClusterManagementService.java
```diff
@@ -194,9 +197,32 @@ private void checkIsRestful(CacheElement config) {
...
+      return false;
+    }
+  }
+
+  /**
+   * not public because user code should *not* try to catch this exception by name, since other
+   * implementations of {@link ClusterManagementService} will throw different subtypes of
+   * RuntimeException
+   */
+  private static class RemoteClusterManagementException extends RuntimeException {
```
onichols-pivotal: ClusterManagementService interface should declare that all methods throw RuntimeException.  And LocatorClusterManagementService should also have checks to ensure no unsuccessful result is ever returned.

geode-management/src/main/java/org/apache/geode/management/internal/ClientClusterManagementService.java
```diff
@@ -194,9 +197,32 @@ private void checkIsRestful(CacheElement config) {
...
+    public RemoteClusterManagementException(ClusterManagementResult result) {
+      super(result.getStatusCode() + (StringUtils.isEmpty(result.getStatusMessage()) ? ""
```
onichols-pivotal: Perhaps instead of extending RuntimeException, we should create a checked exception type ClusterManagementException with an additional method to get the full ClusterManagementResult.  It is possible for some operations, particularly realizations across multiple members, to be "partially successful" before failing, and the caller might need access to that rich failure information to take the appropriate cleanup steps.

geode-management/src/main/java/org/apache/geode/management/internal/ClientClusterManagementService.java
onichols-pivotal: fixed

geode-core/src/main/java/org/apache/geode/management/internal/api/LocatorClusterManagementService.java
```diff
@@ -468,11 +467,30 @@ public LocatorClusterManagementService(ConfigurationPersistenceService persisten
     return result;
   }
 
-  @Override
+  private <T extends ClusterManagementResult> T assertSuccessful(T result) {
```
onichols-pivotal: In the case where we know we are being called from the REST API, we need to return the unsuccessful cluster management result rather than throw exception (to avoid losing rich failure info).

geode-management/src/main/java/org/apache/geode/management/api/ClusterManagementService.java
```diff
@@ -35,6 +35,7 @@
    * @param config this holds the configuration attributes of the element to be created on the
    *        cluster, as well as the group this config belongs to
    * @return a {@link ClusterManagementRealizationResult} indicating the success of the creation
+   * @throws RuntimeException if unsuccessful
```
dschneider-pivotal: It does not seem right to throw the very general "RuntimeException". Instead a more specific subclass of RuntimeException should be part of our external API and that is what we would say we throw.
Note it does not need to be a checked exception. If it is a checked exception then all users of the api have to handle that exception. On some of the methods on this interface we throw other runtime exceptions (i.e. IllegalArgumentException, NoMemberException, EntityExistsException). So if we have some other reason for failing would should probably just have another, subclass of either RuntimeException or GemFireException. We should also include in the javadocs an explanation for why the exception will be thrown

geode-management/src/main/java/org/apache/geode/management/api/ClusterManagementService.java
```diff
@@ -47,6 +48,7 @@
    *        cluster
    * @return a {@link ClusterManagementRealizationResult} indicating the success of the deletion
    * @throws IllegalArgumentException, NoMemberException, EntityExistsException
```
dschneider-pivotal: Having the NoMemberException and EntityExistsException in the external api is a problem because they are internal. So users could not get code to compile that wants to handle these and the javadocs will have broken links. We could just have one external exception class that these apis throw when they fail. Internally we could catch the other exceptions and throw the external one.

geode-management/src/main/java/org/apache/geode/management/api/ClusterManagementService.java
onichols-pivotal: done

geode-management/src/main/java/org/apache/geode/management/api/ClusterManagementService.java
onichols-pivotal: replaced with our own exception

geode-core/src/main/java/org/apache/geode/management/internal/api/LocatorClusterManagementService.java
onichols-pivotal: fixed by adding controller advice to catch our own exception and unwrap it

onichols-pivotal:
@jmelchio

geode-management/src/main/java/org/apache/geode/management/api/ClusterManagementException.java
```diff
@@ -0,0 +1,41 @@
...
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.apache.geode.management.api;
+
+/**
+ * Base class of all exceptions thrown by {@link ClusterManagementService} implementations.
+ */
+public class ClusterManagementException extends RuntimeException {
+  protected final ClusterManagementResult result;
```
dschneider-pivotal: make this private. subclasses can use "getResult()"

geode-management/src/main/java/org/apache/geode/management/api/ClusterManagementException.java
```diff
@@ -0,0 +1,41 @@
...
+
+  public ClusterManagementException(ClusterManagementResult result) {
```
dschneider-pivotal: add javadocs on public methods since this is external api

geode-management/src/main/java/org/apache/geode/management/api/ClusterManagementException.java
```diff
@@ -0,0 +1,41 @@
...
+        + (org.springframework.util.StringUtils.isEmpty(result.getStatusMessage()) ? ""
+            : (": " + result.getStatusMessage())));
+    this.result = result;
+  }
+
+  public ClusterManagementResult getResult() {
+    return result;
+  }
+
+  public ClusterManagementResult.StatusCode getStatusCode() {
```
dschneider-pivotal: consider removing getStatusCode() and getStatusMessage() since they can be done using getResult().getStatusCode() ...

onichols-pivotal: I'd prefer the other direction, since getResult will have a lot of fields that are not applicable.  I would remove `getResult()` entirely but the controller advice needs it.  Marked as internal use only.

geode-management/src/main/java/org/apache/geode/management/api/ClusterManagementException.java
onichols-pivotal: done

geode-management/src/main/java/org/apache/geode/management/api/ClusterManagementException.java
onichols-pivotal: made private

PR #3870, opened 2019-07-31
GEODE-6804: Tab completion does not work when specifying path for --jar argument in gfsh

dschneider-pivotal:
Previous fix had issues on Windows with tests using '/'.

@jmelchio 

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/integrationTest/java/org/apache/geode/management/internal/cli/GfshParserConverterTest.java
+++ b/geode-core/src/integrationTest/java/org/apache/geode/management/internal/cli/GfshParserConverterTest.java
@@ -206,4 +206,28 @@ public class GfshParserConverterTest {
     result = parser.parse(command + "invalid_action");
     assertThat(result).isNull();
   }
+
+  @Test
+  public void testJarFilesPathConverter() {
+    String command = "deploy --jar=";
+    commandCandidate = parser.complete(command);
+    assertThat(commandCandidate.size()).isGreaterThan(0);
+    assertThat(commandCandidate.getCandidate(0)).isEqualTo(command + "/");
+  }
+
+  @Test
+  public void testJarFilesPathConverterWithMultiplePaths() {
+    String command = "deploy --jar=foo.jar,";
+    commandCandidate = parser.complete(command);
+    assertThat(commandCandidate.size()).isGreaterThan(0);
+    assertThat(commandCandidate.getCandidate(0)).isEqualTo(command + "/");
+  }
+
+  @Test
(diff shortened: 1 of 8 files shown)


geode-core/src/integrationTest/java/org/apache/geode/management/internal/cli/GfshParserConverterTest.java
```diff
@@ -206,4 +206,28 @@ public void testExpirationAction() {
...
+    assertThat(commandCandidate.size()).isGreaterThan(0);
+    assertThat(commandCandidate.getCandidate(0)).isEqualTo(command + "/");
+  }
+
+  @Test
+  public void testJarDirPathConverter() {
+    String command = "deploy --dir=";
+    commandCandidate = parser.complete(command);
+    assertThat(commandCandidate.size()).isGreaterThan(0);
+    assertThat(commandCandidate.getCandidate(0)).isEqualTo(command + "/");
```
onichols-pivotal: these still look windows-specific.  maybe the simplest thing here is just `    Assume.assumeFalse(System.getProperty("os.name").toLowerCase().contains("win"));`

dschneider-pivotal: We could change "/" to File.seperator. My only concern is we didn't see this test fail on Windows.

onichols-pivotal: The test that would be running it takes 4 hours, so may not have been completed yet, or may have died due to the pipeline instability today.

Changing this from / to File.separator also won't work.  / is the common prefix of all roots on a unix filesystem, but on Windows it would either be C:\ or maybe "" if there are other drive letters mapped...

dschneider-pivotal: It looks like FilePathStringConverter will call File.getRoots and give us a completion for each root. So if we had multiple roots on Windows the first completion would end with the first root. So I think the test can call File.getRoots() and expect the first completion to end with the first root.

onichols-pivotal: that sounds reasonable

dschneider-pivotal: test now uses getRoots

PR #3872, opened 2019-07-31
GEODE-130: Cleanup internal API warnings.

pivotal-jbarrett:
Isolate warnings to unsafe module.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/concurrent/CompactConcurrentHashSet2.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/concurrent/CompactConcurrentHashSet2.java
@@ -34,7 +34,6 @@ package org.apache.geode.internal.concurrent;
 
 import java.io.ObjectStreamField;
 import java.io.Serializable;
-import java.lang.reflect.Field;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.util.AbstractSet;
@@ -49,6 +48,7 @@ import java.util.concurrent.locks.ReentrantLock;
 
 import org.apache.geode.annotations.Immutable;
 import org.apache.geode.annotations.internal.MakeNotStatic;
+import org.apache.geode.unsafe.internal.sun.misc.Unsafe;
 
 /**
  * <p>
@@ -1961,13 +1961,10 @@ public class CompactConcurrentHashSet2<V> extends AbstractSet<V> implements Set<
     }
 
     @Immutable
-    private static final sun.misc.Unsafe U;
+    // private static final Unsafe U = new Unsafe();
     private static final long LOCKSTATE;
     static {
       try {
-        Field f = sun.misc.Unsafe.class.getDeclaredField("theUnsafe");
-        f.setAccessible(true);
-        U = (sun.misc.Unsafe) f.get(null);
         Class<?> k = TreeBin.class;
         LOCKSTATE = U.objectFieldOffset(k.getDeclaredField("lockState"));
       } catch (Exception e) {
@@ -2287,7 +2284,7 @@ public class CompactConcurrentHashSet2<V> extends AbstractSet<V> implements Set<
 
   // Unsafe mechanics
   @Immutable
-  private static final sun.misc.Unsafe U;
+  private static final Unsafe U = new Unsafe();
   private static final long SIZECTL;
   private static final long TRANSFERINDEX;
   private static final long BASECOUNT;
@@ -2298,9 +2295,6 @@ public class CompactConcurrentHashSet2<V> extends AbstractSet<V> implements Set<
 
   static {
     try {
-      Field f = sun.misc.Unsafe.class.getDeclaredField("theUnsafe");
-      f.setAccessible(true);
-      U = (sun.misc.Unsafe) f.get(null);
       Class<?> k = CompactConcurrentHashSet2.class;
       SIZECTL = U.objectFieldOffset(k.getDeclaredField("sizeCtl"));
       TRANSFERINDEX = U.objectFieldOffset(k.getDeclaredField("transferIndex"));
--- a/geode-core/src/main/java/org/apache/geode/internal/concurrent/CompactConcurrentHashSet2.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/concurrent/CompactConcurrentHashSet2.java
@@ -1961,7 +1961,6 @@ public class CompactConcurrentHashSet2<V> extends AbstractSet<V> implements Set<
     }
 
     @Immutable
-    // private static final Unsafe U = new Unsafe();
     private static final long LOCKSTATE;
     static {
       try {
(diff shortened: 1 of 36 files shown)


geode-unsafe/build.gradle
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+apply from: "${rootDir}/${scriptDir}/standard-subproject-configuration.gradle"
+apply from: "${project.projectDir}/../gradle/publish.gradle"
+
+dependencies {
+  compile(platform(project(':boms:geode-all-bom')))
+}
+
+ext.findJavac = { ->
```
upthewaterspout: What is this findJavac function for? Isn't gradle already doing this?

geode-core/src/main/java/org/apache/geode/internal/concurrent/CompactConcurrentHashSet2.java
```diff
@@ -1961,13 +1961,10 @@ else if ((xp = x.parent) == null) {
     }
 
     @Immutable
-    private static final sun.misc.Unsafe U;
+    // private static final Unsafe U = new Unsafe();
```
upthewaterspout: Remove commented out code

geode-junit/build.gradle
```diff
@@ -24,6 +24,7 @@ dependencies {
   compile(platform(project(':boms:geode-all-bom')))
 
   compileOnly(project(':geode-core'))
+//  compile(':geode-unsafe')
```
upthewaterspout: Remove commented out code

geode-core/src/main/java/org/apache/geode/management/internal/cli/shell/unsafe/GfshSignalHandler.java
```diff
@@ -80,9 +81,11 @@ protected void handleDefault(final sun.misc.Signal sig, final ConsoleReader cons
     }
   }
 
-  protected SignalHandler getOriginalSignalHandler(final Signal signal) {
+  private SignalHandler getOriginalSignalHandler(final Signal signal) {
     final SignalHandler handler = originalSignalHandlers.get(signal);
-    return (handler == SignalHandler.SIG_DFL || handler == SignalHandler.SIG_IGN ? null : handler);
+    // return (handler == SignalHandler.SIG_DFL || handler == SignalHandler.SIG_IGN ? null :
```
upthewaterspout: Commented out code. And this changed the behavior, is the new behavior correct? I'm not sure what this was doing.

geode-unsafe/src/main/java/org/apache/geode/unsafe/internal/sun/misc/Signal.java
```diff
@@ -0,0 +1,113 @@
...
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package org.apache.geode.unsafe.internal.sun.misc;
+
+import java.util.IdentityHashMap;
+
+public class Signal {
```
upthewaterspout: This looks like a brand new class and functionality. Probably needs a unit test? Where is this used?

geode-unsafe/build.gradle
pivotal-jbarrett: Gradle doesn't know where the Java command line compiler is. It uses the Java compiler APIs to invoke the compiler. You you want use the command line compiler then you have to provide the path to the executable. In our CI this executable in not in in the path because we have multiple versions of the JDK installed. We need to use forked compile here in order to send the flags to ignore internal APIs. You can't pass those flags to the compiler via the compiler API.

geode-core/src/main/java/org/apache/geode/internal/concurrent/CompactConcurrentHashSet2.java
pivotal-jbarrett: Good catch, thanks!

geode-unsafe/src/main/java/org/apache/geode/unsafe/internal/sun/misc/Signal.java
pivotal-jbarrett: I went back and forth on that. It's really just a 1:1 wrapper around the internal API Signal classes to avoid the compiler warnings on our other modules. I suppose I could whip one up anyway. Thanks!

geode-core/src/main/java/org/apache/geode/management/internal/cli/shell/unsafe/GfshSignalHandler.java
pivotal-jbarrett: The original code relied on the fact that the internal API handle method would return the default handlers for some events if they weren't ever initialized. Setting then back to `null` to the default has the same effect in the internal API, so this logic was really unnecessary. With the new wrapper API the original default is preserved and reset rather than nulled. Still producing the same effect.

pivotal-jbarrett: Scratch that... It looks like we were trying to chain singles here but not for these two defaults. You're right this will result in a change in behavior, for which we have no test for. Let me work out an integration test if possible for this. Not sure we can given the nature of the signals.

pivotal-jbarrett: And only if it is the default signal handler on these two signals.

geode-unsafe/build.gradle
rhoughton-pivot: Setting the executable is not standard. Shouldn't this use https://docs.gradle.org/current/userguide/building_java_projects.html#example_configure_java_7_build instead?

pivotal-jbarrett: Seems to effectively be doing the same thing. The link expects to use a different JDK than the one running Gradle and have it provided in a property. My approach here uses the Gradle JDK location to find the java executable. The link sets the executable to one of the java executables depending on task type.

pivotal-jbarrett: Ok, I see some magic in there. Let me see if I can do some of the same.

geode-unsafe/build.gradle
```diff
@@ -32,30 +32,12 @@ dependencies {
...
 
-ext.findJavac = { ->
-  def javac = new File(System.getProperty('java.home') + '/bin/javac')
-  if (javac.isFile() && javac.canExecute()) {
-    return javac.toString()
+tasks.withType(AbstractCompile) {
+  options.with {
+    fork = true
+    forkOptions.javaHome = file(System.getProperty('java.home'))
+    compilerArgs << '-XDignore.symbol.file'
```
rhoughton-pivot: Ah, I love how much more simple this is!

PR #3878, opened 2019-08-01
GEODE-7042: Wait until all startup tasks complete to update server status as "online"

aaronlindsey:
Currently, the server status is set to "online" before all of the startup tasks have completed. This tells users incorrect facts about the system and its availability.

This PR changes the behavior of `ServerLauncher.start()` so that it waits for all of the asynchronous startup tasks to complete before returning the "online" status. The GFSH "start server" command relies on this status, so this PR also has the effect of making the command wait for all startup tasks to complete before returning.

Please review @kirklund @moleske @demery-pivotal @mhansonp


metatype:
If the server being started has to recover really laaaarge disk stores and create indexes how will that affect the gfsh shell?  Will the user input be blocked until timeout or the disk stores are recovered?  Just curious what asynchronous startup tasks are affected by this change.

aaronlindsey:
> If the server being started has to recover really laaaarge disk stores and create indexes how will that affect the gfsh shell? Will the user input be blocked until timeout or the disk stores are recovered? Just curious what asynchronous startup tasks are affected by this change.

Specifically, the two asynchronous startup tasks it's waiting on are redundancy recovery and disk store recovery. At the moment it will wait indefinitely for these two startup tasks.

mstolz1:
It shouldn't hang waiting, but rather it should indicate what is happening.



--
Mike Stolz
Principal Engineer, Pivotal Cloud Cache
Mobile: +1-631-835-4771

kirklund:
@mstolz1 @metatype Note, `start server` already hangs indefinitely waiting for the process to startup unless the process goes away. We just added two more tasks to the list of things that it waits on: redundancy recover and recovery of persistent values (keys are already synchronous).

Changing the apparent interactiveness of GFSH during `start server` would require some changes to GFSH infrastructure which might be something this team shouldn't tackle (we could be looking at weeks of work for this). We should discuss in person.

mstolz1:
Thanks for the clarification.
Now I understand better what this is all about.

--
Mike Stolz
Principal Engineer, Pivotal Cloud Cache
Mobile: +1-631-835-4771

aaronlindsey:
@kirklund and I ran into issues merging this PR through GitHub, so we squashed and merged this manually as ab5c2df147e18a3f9750f8be9d40c368b5e9d920.

PR #3879, opened 2019-08-02
GEODE-6982: fix two error messages missed in first pass

onichols-pivotal:
- capitalize URI in sentence usage
- auto-capitalize messages that originate outside of CMS
- return success-with-informational-message as HTTP 200 OK rather than 500 Internal Error:

before: HTTP 500
```
{
  "statusCode": "ERROR",
  "statusMessage": "rebalance returned info: Distributed system has no regions that can be rebalanced",
}
```

after: HTTP 200
```
{
  "statusCode": "OK",
  "statusMessage": "Operation finished successfully.",
  "operationStart": "2019-08-02T01:08:22.019Z",
  "operationEnded": "2019-08-02T01:08:22.061Z",
  "result": {
    "statusMessage": "Distributed system has no regions that can be rebalanced."
  }
}
```


geode-management/src/main/java/org/apache/geode/management/api/ClusterManagementResult.java
```diff
@@ -90,9 +89,35 @@ public ClusterManagementResult(ClusterManagementResult copyFrom) {
...
+    this.statusCode = statusCode;
+    this.statusMessage = proper(message);
+  }
+
+  /**
+   * Capitalized the first letter and adds a period, if needed.
+   * This helps the CMS API give a consistent format to error messages, even when they sometimes
+   * come from parts of the system beyond our control.
+   */
+  private static String proper(String message) {
```
jmelchio: Can we give this a more descriptive name? Something like `formatErrorMessage`?

onichols-pivotal: done

PR #3882, opened 2019-08-02
GEODE-7036: Fix for NPE caused by ex-coordinator joining quorum

mivanac:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/distributedTest/java/org/apache/geode/distributed/internal/RestartOfMemberDistributedTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/distributed/internal/RestartOfMemberDistributedTest.java
@@ -58,7 +58,7 @@ public class RestartOfMemberDistributedTest {
   }
 
   @Test
-  public void restartofmembers() {
+  public void exCoordinatorJoiningQuorumDoesNotThrowNullPointerException() {
     IgnoredException exp1 =
         IgnoredException.addIgnoredException(ForcedDisconnectException.class.getName());
     IgnoredException exp2 =
--- a/geode-core/src/distributedTest/java/org/apache/geode/distributed/internal/RestartOfMemberDistributedTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/distributed/internal/RestartOfMemberDistributedTest.java
@@ -1,0 +1,83 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.apache.geode.distributed.internal;
+
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Properties;
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+
+import org.apache.geode.ForcedDisconnectException;
+import org.apache.geode.test.dunit.IgnoredException;
+import org.apache.geode.test.dunit.rules.ClusterStartupRule;
+import org.apache.geode.test.dunit.rules.MemberVM;
+
+
+public class RestartOfMemberDistributedTest {
+  public List<MemberVM> locators = new ArrayList<>();
+  public List<MemberVM> servers = new ArrayList<>();
+
+
+
+  @Rule
+  public ClusterStartupRule clusterStartupRule = new ClusterStartupRule();
+
+  @Before
+  public void before() {
+    Properties properties = new Properties();
+
+    locators.add(clusterStartupRule.startLocatorVM(0, properties));
+    servers.add(clusterStartupRule.startServerVM(1, properties, locators.get(0).getPort()));
+    locators.add(clusterStartupRule.startLocatorVM(2, properties, locators.get(0).getPort()));
+    servers.add(clusterStartupRule.startServerVM(3, properties, locators.get(0).getPort()));
+  }
+
+  @After
+  public void after() {
+    servers.clear();
+    locators.clear();
+  }
+
+  @Test
+  public void restartofmembers() {
+    IgnoredException exp1 =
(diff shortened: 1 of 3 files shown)


geode-core/src/main/java/org/apache/geode/distributed/internal/InternalLocator.java
```diff
@@ -1108,6 +1108,7 @@ private boolean attemptReconnect() throws InterruptedException, IOException {
       InternalDistributedSystem newSystem =
           (InternalDistributedSystem) system.getReconnectedSystem();
       if (newSystem != null) {
+        setLocator(this);
```
bschuchardt: If locator restart fails this will leave the static _locator_ variable set.  Previously this wasn't the case.

geode-core/src/distributedTest/java/org/apache/geode/distributed/internal/RestartOfMemberDistributedTest.java
```diff
@@ -0,0 +1,83 @@
...
+    try {
+      int locator2port = locators.get(1).getPort();
+      Properties properties = new Properties();
+      int locator0port = locators.get(0).getPort();
+      clusterStartupRule.crashVM(1);
+      clusterStartupRule.crashVM(0);
+      clusterStartupRule.startLocatorVM(0, locator0port, properties, locator2port);
+      clusterStartupRule.startServerVM(1, properties, locator2port);
+      locators.get(1).waitTilFullyReconnected();
+
```
aaronlindsey: I reverted the change to `InternalLocator` (to reproduce the error) and this test seemed to hang indefinitely. (I waited 15 min). Is there any way to make this test fail so that I can reproduce the issue?

mivanac: This is the error. In existing locator, NPE is reported, so it hangs, and restarted locator is waiting on existing locator. This is why everything is hanging.

geode-core/src/main/java/org/apache/geode/distributed/internal/InternalLocator.java
mivanac: This was only case, if it was not set, in code above, when checking quorum. I will correct this.

geode-core/src/distributedTest/java/org/apache/geode/distributed/internal/RestartOfMemberDistributedTest.java
bschuchardt: @aaronlindsey where did the test hang?  The waitTilFullyConnected methods should time out after 5 minutes.  If you can identify the place in the test that's hanging maybe an await() could be added to keep the test from hanging without the fix.

geode-core/src/main/java/org/apache/geode/distributed/internal/InternalLocator.java
```diff
@@ -1122,10 +1127,12 @@ private boolean attemptReconnect() throws InterruptedException, IOException {
           restartWithSystem(newSystem, GemFireCacheImpl.getInstance());
         } catch (CancelException e) {
           stoppedForReconnect = true;
+          if (noprevlocator) {
+            removeLocator(this);
```
bschuchardt: looks good

geode-core/src/distributedTest/java/org/apache/geode/distributed/internal/RestartOfMemberDistributedTest.java
aaronlindsey: Sorry for the lack of details here. I debugged the test and saw that it hangs on `clusterStartupRule.startLocatorVM(0, locator0port, properties, locator2port);`. I suggest wrapping this with something like
```
import static org.apache.geode.test.awaitility.GeodeAwaitility.await;

await().until(() -> {
  clusterStartupRule.startLocatorVM(0, locator0port, properties, locator2port);
  return true;
});
```

mivanac: Thanks for advice.

geode-core/src/distributedTest/java/org/apache/geode/distributed/internal/RestartOfMemberDistributedTest.java
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  @After
+  public void after() {
+    servers.clear();
+    locators.clear();
+  }
+
+  @Test
+  public void restartofmembers() {
```
aaronlindsey: Feel free to ignore this, but I think a more descriptive name that is properly capitalized would be helpful here. Perhaps something like `exCoordinatorJoiningQuorumDoesNotThrowNullPointerException`.

geode-core/src/distributedTest/java/org/apache/geode/distributed/internal/RestartOfMemberDistributedTest.java
aaronlindsey: I made a mistake in my last comment when I suggested wrapping that startLocatorVM call in an await(). The problem is that the startLocatorVM call is blocking and therefore the await() has no effect. I tried wrapping the entire thing inside an asynchronous invocation using ExecutorServiceRule and waiting on the result, but I was unable to fix the hanging problem. Maybe someone with more expertise in this area could suggest a fix.

geode-core/src/distributedTest/java/org/apache/geode/distributed/internal/RestartOfMemberDistributedTest.java
aaronlindsey: Also, see my [other comment about this test](https://github.com/apache/geode/pull/3882#discussion_r312280102)

geode-core/src/distributedTest/java/org/apache/geode/distributed/internal/RestartOfMemberDistributedTest.java
mivanac: When I test fault reproduction, it does not hang forever, TC fails after 6 and half minutes.

aaronlindsey: On my system, it sometimes fails after 6 and a half minutes with error "Suspicious strings were written to the log during this run", and sometimes hangs for 15+ minutes. It seems to alternate between these two outcomes each time I run it. My ideal would be that the test fails reliably when I remove the fix from the code. However, I don't see this as a gating issue necessarily because (1) this may only be reproducible in my environment, (2) the code as-is does not hang – it only hangs when I remove the fix, and (3) I don't know how much effort would be required to ensure that this test fails reliably without the fix.

geode-core/src/distributedTest/java/org/apache/geode/distributed/internal/RestartOfMemberDistributedTest.java
mivanac: Could we merge this PR, or do we need some additional changes in TC?

bschuchardt:
I did the merge just now

PR #3530, opened 2019-05-01
DISCUSSION: Add TimeUnit to timeout properties and parameters on public API.

pivotal-jbarrett:
I am using this PR as an example of a proposal to deprecate current timeout APIs with new APIs that take `TimeUnit`.

Current APIs leave an ambiguity when coding as to what the unit of time is. You must consult the documentation to determine the time unit. Many of the APIs like this documentation. Making things worse there isn't consistency throughout the API, some are in seconds, others milliseconds. Adding `TimeUnit` to these APIs removes the ambiguity since the caller dictates the time unit.

It also produces more readable code. Consider `setSocketConnectTimeout(86400000)` vs. `setSocketConnectTimeout(1, DAYS)`.

The goal would be to slowly add `TimeUnit` to all public APIs, and deprecate the original, as we are working in an area of code.

A similar task has already been completed in [geode-native](http://github.com/apache/geode-native).

PR #3888, opened 2019-08-06
GEODE-7049: Add timeout to Java client Execute::execute() methods

albertogpz:

### For all changes:

### Note:


geode-core/src/main/java/org/apache/geode/cache/execute/Execution.java
```diff
@@ -106,6 +107,22 @@
...
+   *
+   * @param functionId id of the function to execute
+   * @param timeoutMs milliseconds to wait for the operation to finish before timing out
+   * @return ResultCollector to retrieve the results received. This is different object than the
+   *         ResultCollector provided in {@link Execution#withCollector(ResultCollector)}. User has
+   *         to use this reference to retrieve results.
+   *
+   * @since GemFire 6.0
+   */
+  ResultCollector<OUT, AGG> execute(String functionId, int timeoutMs) throws FunctionException;
```
pivotal-jbarrett: There was a discussion months ago about migrating timeout params to `long` and `TimeUnit` pairs on all public APIs. Please use signature:
`ResultCollector<OUT, AGG> execute(String functionId, long timeout, TimeUnit unit) throws FunctionException;`

pivotal-jbarrett: Here is the discussion https://github.com/apache/geode/pull/3530

albertogpz: I will. Thanks.
By the way, the timeout passed to the execute and executeFunction methods of DistributedRegionFunctionExecutor, PartitionedRegionFunctionExecutor, MemberFunctionExecutor and MultiRegionFunctionExecutor will have no effect.
Is that ok? If so, should I highlight it in the code somehow?

geode-core/src/main/java/org/apache/geode/internal/cache/execute/AbstractExecution.java
```diff
@@ -435,7 +447,7 @@ public boolean isIgnoreDepartedMembers() {
     return ignoreDepartedMembers;
   }
 
-  protected abstract ResultCollector executeFunction(Function fn);
+  protected abstract ResultCollector executeFunction(Function fn, int timeoutMs);
```
pivotal-jbarrett: Changing this method to take the timeout and unit would remove the duplicate conversions from time an unit to milliseconds. Also in that conversion you need to account for values less than 1ms, like 1ns, as well as values greater than `Integer.MAX_VALUE`ms, like `Long.MAX_VALUE`ms. I would suggest a little timeout helper class to do this since we will need it later.

albertogpz: I added the TimeUnit parameter to the executeFunction method but still there are several places where the conversion must be done.
I have also added the helper class suggested.
Thanks!

upthewaterspout:
The execute method returns a `ResultCollector`. `ResultCollector` does  have a timeout - there is get(long, TimeUnit) method. So naively, these timeout parameters seem redundant.

The problem is that the execute method doesn't behave consistently when invoked on a peer vs. on a client. execute invoked on a peer doesn't block. It returns immediately, and the ResultCollector is how you wait for results. I think that's what you are seeing when you said your new parameters have no effect on DistributedRegionFunctionExecutor, etc.

I have seen in the past that invoking execute on a client *does* block, so I can see why you want to add these parameters to that method. Of course that means that the parameters to `ResultCollector.get` make no sense on the client (with or without these changes).

Maybe it would make more sense to make the execute on the client behave like it does when invoked on a peer - don't block, but return a result collector that can be used to wait for the results?

albertogpz:
@upthewaterspout It makes a lot of sense to make the execute method non blocking and leave the blocking to the result collector together with the timeout. As a matter of fact, I found a ticket written by you a couple of years ago in which you already talked about this although it seems nobody has worked on it (https://issues.apache.org/jira/browse/GEODE-3817).
Nevertheless, I currently see some problems with that approach:
- The C++ and C# clients already provide a timeout parameter in the execute methods and they are also blocking so in order to be consistent we would also have to change these two clients.
- In the Java client, as of today, when an execute call times out (because the execution takes longer than what is configured in the gemfire.CLIENT_FUNCTION_TIMEOUT property) after the configured retries have been reached, the connection is closed. We would lose this feature if we change the execute methods to be non-blocking.
- The current DefaultResultCollector ignores the timeout passed in the get method. It would be needed to implement the timeout handling in this method as well as in any custom collector. This is actually not a problem, just a finding.
As I am quite new in the community, I do not know how to progress with this PR. Should we go on with the changes proposed in this PR and at the same time start a discussion in the list about changing the execute methods to be non-blocking?

pivotal-jbarrett:
>   As I am quite new in the community, I do not know how to progress with this PR. Should we go on with the changes proposed in this PR and at the same time start a discussion in the list about changing the execute methods to be non-blocking?

I would lean towards resolving the the direction we want go with the API before merging this PR. It would unfortunate to introduce new API that introduces something we decide we don't want is immediately deprecated. Perhaps now would be a good time to close this PR, open a discussion on the dev@geode list with your ideas for normalizing the two clients, and then re-open this PR if it still makes sense after that discussion.

albertogpz:
Ok. I'll do it like that.
Thanks

albertogpz:
Ticket reopened in order to implement the changes as described in the
"Improvements on client Function execution API" Proposal in the https://cwiki.apache.org/confluence/display/GEODE/

geode-core/src/main/java/org/apache/geode/cache/execute/Execution.java
```diff
@@ -106,8 +110,29 @@
...
+   * {@link FunctionService#getFunction(String)} on the executing member.
+   *
+   * @param functionId id of the function to execute
+   * @param timeout time to wait for the operation to finish before timing out
+   * @param unit timeout unit
+   * @return ResultCollector to retrieve the results received. This is different object than the
+   *         ResultCollector provided in {@link Execution#withCollector(ResultCollector)}. User has
+   *         to use this reference to retrieve results.
+   *
+   * @since GemFire 6.0
```
pivotal-jbarrett: This should say Geode 1.11.

geode-core/src/main/java/org/apache/geode/cache/execute/Execution.java
```diff
@@ -122,4 +147,26 @@
...
+   * @param function instance to execute
+   * @param timeout time to wait for the operation to finish before timing out
+   * @param unit timeout unit
+   * @throws LowMemoryException if the {@link Function#optimizeForWrite()} returns true and there is
+   *         a low memory condition
+   * @return ResultCollector to retrieve the results received. This is different object than the
+   *         ResultCollector provided in {@link Execution#withCollector(ResultCollector)}. User has
+   *         to use this reference to retrieve results.
+   *
+   * @since GemFire 6.0
```
pivotal-jbarrett: Again, this is Geode 1.11.

upthewaterspout:
Looks like there are two +1 reviews, so I'll go ahead and merge this to develop.

PR #3889, opened 2019-08-06
GEODE-7015: Fixing redundant copies in test

mcmellawatt:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/control/RebalanceOperationDistributedTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/control/RebalanceOperationDistributedTest.java
@@ -2238,7 +2238,7 @@ public class RebalanceOperationDistributedTest extends CacheTestCase {
       RebalanceResults results = doRebalance(false, manager);
 
       // The rebalance should have done some work since the buckets were imbalanced
-      assertThat(results.getTotalPrimaryTransfersCompleted() > 0).isTrue();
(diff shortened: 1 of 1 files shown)


geode-core/src/distributedTest/java/org/apache/geode/internal/cache/control/RebalanceOperationDistributedTest.java
```diff
@@ -2314,7 +2314,7 @@ public void testBucketImageProviderBouncesDuringBucketMove()
       RebalanceResults results = doRebalance(false, getCache().getInternalResourceManager());
 
       // The rebalance should have done some work since the buckets were imbalanced
-      assertThat(results.getTotalPrimaryTransfersCompleted() > 0).isTrue();
+      assertThat(results.getTotalBucketTransfersCompleted() > 0).isTrue();
```
jhuynh1: We are ok with secondaries being transferred and not just primaries?

nabarunnag: I think Jason the test had redundantCopies set to 0 but in the createPersistentPartitionRegion function call, variable redundantCopies was completely ignored and was set to hard coded 1 .
Hence I think the test was originally designed to have no secondaries.

mcmellawatt: Naba is correct.  The concept of primaries and secondaries shouldn't even apply here since we don't have redundancy (redundant copies is 0).

PR #3892, opened 2019-08-06
GEODE-7050: Use Log4jAgent only if Log4j is using Log4jProvider

kirklund:
This change prevents Geode from using Log4jAgent if Log4j Core is
present but not using Log4jProvider.

For example, Log4j uses SLF4JProvider instead of Log4jProvider when 
log4j-to-slf4j is in the classpath.

By disabling Log4jAgent when other Log4j Providers are in use, this
prevents problems such as ClassCastExceptions when attempting to cast
loggers from org.apache.logging.slf4j.SLF4JLogger to
org.apache.logging.log4j.core.Logger to get the LoggerConfig or
LoggerContext.

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/logging/DefaultProviderChecker.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/logging/DefaultProviderChecker.java
@@ -1,0 +1,82 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.apache.geode.internal.logging;
+
+import java.util.function.Function;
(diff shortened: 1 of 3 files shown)


kirklund:
UnitTestOpenJDK11 hit the timeout exceeded problem that is under discussion on dev-list. Note that UnitTestOpenJDK8 passed.

geode-core/src/main/java/org/apache/geode/internal/logging/DefaultProviderChecker.java
```diff
@@ -0,0 +1,82 @@
...
+  private final Function<String, Boolean> isClassLoadableFunction;
+  private final Logger logger;
+
+  DefaultProviderChecker() {
+    this(() -> LogManager.getFactory().getClass(), DefaultProviderChecker::isClassLoadable,
+        StatusLogger.getLogger());
+  }
+
+  @VisibleForTesting
+  DefaultProviderChecker(Supplier<Class> contextFactoryClassSupplier,
```
demery-pivotal: Would it be useful to restrict the kind of class the supplier supplies? Something like `Supplier<Class<? extends LoggerContextFactory>>`. (I don't know if that's the right kind of class to require.)

geode-core/src/test/java/org/apache/geode/internal/logging/DefaultProviderCheckerTest.java
```diff
@@ -0,0 +1,152 @@
...
+
+    boolean value = checker.isAvailable();
+
+    assertThat(value).isFalse();
+  }
+
+  @Test
+  public void isAvailableReturnsFalseIfNotUsingLog4jProvider() {
+    DefaultProviderChecker checker = new DefaultProviderChecker(
+        () -> mock(LoggerContextFactory.class).getClass(), (a) -> false, mock(Logger.class));
```
aaronlindsey: As it is, this test will pass regardless of whether it's using the Log4j provider.
```suggestion
        () -> mock(LoggerContextFactory.class).getClass(), (a) -> true, mock(Logger.class));
```

PR #3894, opened 2019-08-07
GEODE-7055: Don't send failure replies from a P2P reader thread

upthewaterspout:
We were hitting a deadlock during startup if a P2P reader thread tried
to send a failure reply - it would block waiting for startup to finish,
but startup would not finish until the P2P reader thread could read a
startup response.

Send the failure reply in a separate thread, to make sure we always
unblock the P2P reader thread to read new messages.

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/tcp/Connection.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/tcp/Connection.java
@@ -1947,14 +1947,14 @@ public class Connection implements Runnable {
   }
 
   private void sendFailureReply(int rpId, String exMsg, Throwable ex, boolean directAck) {
-    ReplySender dm = null;
+    ReplyException exception = new ReplyException(exMsg, ex);
     if (directAck) {
-      dm = new DirectReplySender(this);
+      ReplySender dm = new DirectReplySender(this);
+      ReplyMessage.send(getRemoteAddress(), rpId, exception, dm);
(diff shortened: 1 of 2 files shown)


geode-core/src/main/java/org/apache/geode/internal/tcp/Connection.java
```diff
@@ -1947,14 +1947,14 @@ void releaseMsgDestreamer(short msgId, MsgDestreamer md) {
...
     if (directAck) {
-      dm = new DirectReplySender(this);
+      ReplySender dm = new DirectReplySender(this);
+      ReplyMessage.send(getRemoteAddress(), rpId, exception, dm);
     } else if (rpId != 0) {
-      dm = this.owner.getDM();
-    }
-    if (dm != null) {
-      ReplyMessage.send(getRemoteAddress(), rpId, new ReplyException(exMsg, ex), dm);
+      DistributionManager dm = this.owner.getDM();
```
bschuchardt: Functionally this will work but have you considered using `owner.getMembershipManager().send()` to send the reply message?  That will bypass the lock that you're currently hitting with dm.putOutgoing().  You'd need a new method in ReplyMessage to create one w/o sending it, I suppose.
I think that would be more in-line with what this code is doing with DirectReplySenders.  In that case we're not blocking on the waitUntilReadyToSendMessages lock but are immediately sending a reply.

PR #3900, opened 2019-08-08
GEODE-7064: Increase timeout to 30m for UnitTest Execute Tests task

mcmellawatt:
Bumps the timeout to 30 minutes but also changes the CALL_STACK_TIMEOUT to be 1200 seconds (20 minutes).  The latter configuration parameter controls when we declare the task "hung" and dump stacks.  We were not dumping stacks at all before these changes because the value was 1800 seconds which is well above the timeout of 10 minutes.  With my proposed change we will dump stacks before the task times out so we will have something to look at in the test artifacts to identify the hang.

### For all changes:

### Note:


onichols-pivotal:
Most of the other jobs allow 10 or 15 minutes between dumping stacks and the hard timeout.  There may be other housekeeping that eats into that buffer, I don't really know for sure, but to be on the safe side I would suggest dump stacks at 20m and timeout at 30m

mcmellawatt:
I'd personally be fine with a 30 minute timeout and 20 minute dump stacks...if nobody opposes I'll go ahead and do that.

PR #3902, opened 2019-08-09
GEODE-6946: separate GatewayReceiver configuration object from xml do…

jinmeiliao:
…main object

* add the converter
* the configuration object should not contain specific bind address and hostname for clients.

### For all changes:

### Note:


jinmeiliao:
@jmelchio @jackw26

geode-management/src/main/java/org/apache/geode/management/configuration/GatewayReceiver.java
```diff
@@ -0,0 +1,104 @@
...
+
+import org.apache.geode.cache.configuration.CacheElement;
+import org.apache.geode.cache.configuration.DeclarableType;
+import org.apache.geode.management.api.CorrespondWith;
+import org.apache.geode.management.api.RestfulEndpoint;
+import org.apache.geode.management.runtime.GatewayReceiverInfo;
+
+
+/**
+ * As a configuration object, user should not be able to set bind address and
```
dschneider-pivotal: This class is external. So it needs a good class comment that describes to users how to use it. Those javadocs should point to the apis on ClusterManagementService that you would pass an instance of this class to.
It should also describe, in general, what the purpose of this class is.
I think this sentence it currently has should be removed. If you think it is needed in this context, explain at the end of the comment why setting the bind address and hostname on a gateway is not currently supported by this api.

geode-management/src/main/java/org/apache/geode/management/configuration/GatewayReceiver.java
```diff
@@ -0,0 +1,104 @@
...
+
+public class GatewayReceiver extends CacheElement implements RestfulEndpoint,
+    CorrespondWith<GatewayReceiverInfo> {
+  @Override
+  @JsonIgnore
+  public String getId() {
+    return getConfigGroup();
+  }
+
+  public static final String GATEWAY_RECEIVERS_ENDPOINTS = "/gateways/receivers";
```
dschneider-pivotal: All the public methods on this class need javadocs.
A good thing to do is a command line build and then look at the generated html file for this class and see if it looks good.

geode-management/src/main/java/org/apache/geode/management/configuration/GatewayReceiver.java
```diff
@@ -0,0 +1,104 @@
...
+
+  @Override
+  public String getEndpoint() {
+    return GATEWAY_RECEIVERS_ENDPOINTS;
+  }
+
+  protected List<DeclarableType> gatewayTransportFilters = new ArrayList<>();
```
dschneider-pivotal: make all these protected fields private

geode-management/src/main/java/org/apache/geode/management/configuration/GatewayReceiver.java
```diff
@@ -0,0 +1,104 @@
...
+  protected String endPort;
+  protected String maximumTimeBetweenPings;
+  protected String socketBufferSize;
+  protected boolean manualStart = false;
+
+  public List<DeclarableType> getGatewayTransportFilters() {
+    return gatewayTransportFilters;
+  }
+
+  public void addGatewayTransportFilter(DeclarableType gatewayTransportFilter) {
```
dschneider-pivotal: Is this the standard pattern we will have on v2 config attributes that are collections? By "this" I mean having an "add" method instead of "set".
If so then gatewayTransportFilters should be declared "final" and should try to stick with this pattern on the other config classes.

geode-management/src/main/java/org/apache/geode/management/configuration/GatewayReceiver.java
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  protected List<DeclarableType> gatewayTransportFilters = new ArrayList<>();
+  protected String startPort;
+  protected String endPort;
+  protected String maximumTimeBetweenPings;
+  protected String socketBufferSize;
+  protected boolean manualStart = false;
+
+  public List<DeclarableType> getGatewayTransportFilters() {
```
dschneider-pivotal: Add a note to the javadoc on this method, that if you modify the returned list, you will be modifying the list owned by this config object.

geode-management/src/main/java/org/apache/geode/management/configuration/GatewayReceiver.java
```diff
@@ -0,0 +1,104 @@
...
+
+  public static final String GATEWAY_RECEIVERS_ENDPOINTS = "/gateways/receivers";
+
+  @Override
+  public String getEndpoint() {
+    return GATEWAY_RECEIVERS_ENDPOINTS;
+  }
+
+  protected List<DeclarableType> gatewayTransportFilters = new ArrayList<>();
+  protected String startPort;
```
dschneider-pivotal: Why are all these attributes that are numeric a String? Shouldn't our api use "int" ("long" if the value could be larger than 2G)?

geode-management/src/main/java/org/apache/geode/management/configuration/GatewayReceiver.java
```diff
@@ -0,0 +1,104 @@
...
+import org.apache.geode.management.api.RestfulEndpoint;
+import org.apache.geode.management.runtime.GatewayReceiverInfo;
+
+
+/**
+ * As a configuration object, user should not be able to set bind address and
+ * hostname for clients on it.
+ */
+
+public class GatewayReceiver extends CacheElement implements RestfulEndpoint,
```
dschneider-pivotal: Add a unit test for this new class

geode-core/src/main/java/org/apache/geode/management/internal/configuration/mutators/GatewayReceiverConfigManager.java
```diff
@@ -20,44 +20,41 @@
...
 public class GatewayReceiverConfigManager
-    implements ConfigurationManager<GatewayReceiverConfig> {
-  private final InternalCache cache;
-
-  public GatewayReceiverConfigManager(InternalCache cache) {
-    this.cache = cache;
-  }
+    implements ConfigurationManager<GatewayReceiver> {
 
+  private GatewayReceiverConverter converter = new GatewayReceiverConverter();
```
dschneider-pivotal: make this final

geode-management/src/main/java/org/apache/geode/management/configuration/GatewayReceiver.java
jinmeiliao: this class is derived from the xsd generated class, since they all use "String" instead of "int" or "long", it just followed that pattern. We should have a discussion about which is better.

geode-management/src/main/java/org/apache/geode/management/configuration/GatewayReceiver.java
jinmeiliao: done

geode-management/src/main/java/org/apache/geode/management/configuration/GatewayReceiver.java
jinmeiliao: done

geode-management/src/test/java/org/apache/geode/management/configuration/GatewayReceiverTest.java
```diff
@@ -23,24 +23,24 @@
 import org.apache.geode.management.api.RestfulEndpoint;
 
 
-public class GatewayReceiverConfigTest {
-  private GatewayReceiverConfig receiver;
+public class GatewayReceiverTest {
```
dschneider-pivotal: I think this unit test should have some coverage of each of the public methods on the class it is testing. It could be as simple as a call of the settor and then validating that the gettor returns what is expected. It currently looks like it is only testing getId and getUri

jinmeiliao: I thought we agreed that we don't test common getter/setters that contains no logic. I have tests for id and uri to make sure group names are used as ids.

geode-management/src/main/java/org/apache/geode/management/configuration/GatewayReceiver.java
dschneider-pivotal: this will be addressed in another ticket

geode-management/src/test/java/org/apache/geode/management/configuration/GatewayReceiverTest.java
dschneider-pivotal: I think it is worth testing the logic we do have, which is a little bit. For example if you just had one test method that verified that if you call a settor then the corresponding gettor will return what was set then at least we make sure we didn't copy & paste the wrong field names in the methods.

geode-management/src/main/java/org/apache/geode/management/configuration/GatewayReceiver.java
dschneider-pivotal: we decided to have a settor. If someone wants to add they can call the gettor and modify the returned collection in place

geode-management/src/main/java/org/apache/geode/management/configuration/GatewayReceiver.java
```diff
@@ -65,8 +65,9 @@ public String getEndpoint() {
     return gatewayTransportFilters;
   }
 
-  public void addGatewayTransportFilter(DeclarableType gatewayTransportFilter) {
-    this.gatewayTransportFilters.add(gatewayTransportFilter);
+  public void setGatewayTransportFilters(
```
dschneider-pivotal: We still need javadocs on all the public methods on this external class

geode-management/src/test/java/org/apache/geode/management/configuration/GatewayReceiverTest.java
jinmeiliao: these are IDE generated code, it would be a bit mundane to test them.

jinmeiliao: https://stackoverflow.com/questions/6197370/should-unit-tests-be-written-for-getter-and-setters 
looks like the industry is also having debate on this. I would go with the majority :-).

PR #3906, opened 2019-08-10
GEODE-6964: Move geode log4j-core classes to geode-log4j

kirklund:
Make geode-log4j an implementation dependency for geode-dunit.


pivotal-jbarrett:
@kirklund It looks like this new module will export two root packages, o.a.g.alerting and o.a.g.logging. As we move towards Java 9 modules we need to make sure that jars have unique package exports. Can you verify that geode-core will no longer exports these two packages? Would it make sense to move o.a.g.alerting to o.a.g.loggging.alerting so that this module only exports a single root package? See `geode-cq`gradle for adding module information to the jar. 

Also if the use of log4j is an implementation detail, and the root package is o.a.g.logging, then perhaps this module should be 'geode-logging` and not `geode-log4j`?

kirklund:
@pivotal-jbarrett The name `geode-log4j` is meant to describe this as the `log4j-core` implementation of Geode Logging as opposed to the yet-to-be-created `geode-logback` which would be the logback implementation of Geode Logging. The idea is that if you want to use `log4j-core` with Geode then drop this jar in your classpath. The pom dependency is implementation because these classes don't provide user APIs that return `log4j-core` classes, they simply install an SPI provider that enables Geode log-level, log-file, etc to interact with `log4j-core`. Without these provider classes, the logging properties and logging GFSH commands do nothing.

This module will export `org.apache.geode.logging.log4j`. `geode-core` will export `org.apache.geode.logging` and `org.apache.geode.logging.spi`. (I have a few changes I still need to make).

I can move the Alerting classes from `alerting` packages into `logging` packages. I was originally intending to keep them separate even though both are implemented with `log4j-core` -- and could thus be implemented in `geode-logback` with `logback` instead.

pivotal-jbarrett:
@kirklund this sounds very cool!

kirklund:
Closing this PR in favor of https://github.com/apache/geode/pull/3914

PR #3911, opened 2019-08-12
GEODE-7079: Prevent NPE During Queue Conflation

jujoramos:
- Added tests.
- Fixed minor warnings.
- Use the cached region name when doing conflation instead of the actual region so the processor doesn't need to wait for the actual region to be fully initialized.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySenderEventProcessor.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySenderEventProcessor.java
@@ -783,7 +783,7 @@ public abstract class AbstractGatewaySenderEventProcessor extends LoggingThread
         if (gsEvent.shouldBeConflated()) {
           // The event should be conflated. Create the conflation key
           // (comprised of the event's region, key and the operation).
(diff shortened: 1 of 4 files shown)


jujoramos:
@BenjaminPerryRoss

geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySenderEventProcessor.java
```diff
@@ -783,7 +783,7 @@ public List conflate(List<GatewaySenderEventImpl> events) {
         if (gsEvent.shouldBeConflated()) {
           // The event should be conflated. Create the conflation key
           // (comprised of the event's region, key and the operation).
-          ConflationKey key = new ConflationKey(gsEvent.getRegion().getFullPath(),
+          ConflationKey key = new ConflationKey(gsEvent.getRegionPath(),
```
agingade: How about a unit test; with mocked gsEvent.getRegion() returning null.

BenjaminPerryRoss: Seems like a good idea, will add.

PR #3912, opened 2019-08-12
GEODE-7080: Do not add removed entry into offline export file.

pivotal-eshu:
Co-authored-by: Donal Evans doevans@pivotal.io

### DIFF
--- a/geode-core/src/integrationTest/java/org/apache/geode/internal/cache/OfflineSnapshotJUnitTest.java
+++ b/geode-core/src/integrationTest/java/org/apache/geode/internal/cache/OfflineSnapshotJUnitTest.java
@@ -51,8 +51,8 @@ public class OfflineSnapshotJUnitTest {
         Region<Integer, MyObject> region =
             rgen.createRegion(cache, ds.getName(), rt, "test" + rcount++);
         final Map<Integer, MyObject> expected = createExpected(st, 1000);
-
         region.putAll(expected);
+        removeAndInvalidateKeys(region, expected);
         cache.close();
 
         DiskStoreImpl.exportOfflineSnapshot(ds.getName(), new File[] {new File(".")},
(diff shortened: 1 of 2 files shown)


geode-core/src/integrationTest/java/org/apache/geode/internal/cache/OfflineSnapshotJUnitTest.java
```diff
@@ -51,8 +51,8 @@ public void testExport() throws Exception {
         Region<Integer, MyObject> region =
             rgen.createRegion(cache, ds.getName(), rt, "test" + rcount++);
         final Map<Integer, MyObject> expected = createExpected(st, 1000);
-
         region.putAll(expected);
+        removeAndInvalidateKeys(region, expected);
```
agingade: It will be nice to create a separate test for testing tombstone (destroy). But I am fine with this too.

geode-core/src/integrationTest/java/org/apache/geode/internal/cache/OfflineSnapshotJUnitTest.java
```diff
@@ -65,6 +65,16 @@ public void testExport() throws Exception {
...
   }
 
+  private void removeAndInvalidateKeys(Region<Integer, MyObject> region,
+      Map<Integer, MyObject> expected) {
+    int removed = 5;
+    int invalidated = 7;
+    region.destroy(removed);
+    region.invalidate(invalidated);
+    expected.remove(removed);
+    expected.put(7, null);
```
metatype: Is `null` different than `Token.isInvalid`?  I'm wondering if we import a null value that will have the same behavior as if the Token is used.

metatype: Based on SnapshotPacket `convertToBytes` I think null is correct.

pivotal-eshu: During snapshot import, null value is converted to Token.INVALID to use the putAll operation.

PR #3915, opened 2019-08-13
GEODE-7082: geode-core:distributedTest timeout exceeded.

mhansonp:
The timeout for geode-core:distributedTest is not sufficiently long.

Lengthened the timeout as it appears the tests were successful at 2hrs and 20min. Adding a little padding just in case to 2hrs and 30min.


mhansonp:
Timeout is not the problem in this case.

PR #3921, opened 2019-08-13
 GEODE-7083: Adds statistics to PeerTypeRegistration

pivotal-jbarrett:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/pdx/internal/PeerTypeRegistration.java
+++ b/geode-core/src/main/java/org/apache/geode/pdx/internal/PeerTypeRegistration.java
@@ -12,12 +12,12 @@
  * or implied. See the License for the specific language governing permissions and limitations under
  * the License.
  */
+
 package org.apache.geode.pdx.internal;
 
 import java.io.IOException;
 import java.util.Collections;
 import java.util.HashMap;
-import java.util.Iterator;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
@@ -26,6 +26,7 @@ import org.apache.logging.log4j.Logger;
 
 import org.apache.geode.InternalGemFireError;
 import org.apache.geode.InternalGemFireException;
+import org.apache.geode.annotations.VisibleForTesting;
 import org.apache.geode.cache.AttributesFactory;
 import org.apache.geode.cache.CacheWriterException;
 import org.apache.geode.cache.DataPolicy;
@@ -68,10 +69,12 @@ public class PeerTypeRegistration implements TypeRegistration {
 
   public static final String REGION_NAME = "PdxTypes";
   public static final String REGION_FULL_PATH = "/" + REGION_NAME;
-  public static final int PLACE_HOLDER_FOR_TYPE_ID = 0xFFFFFF;
-  public static final int PLACE_HOLDER_FOR_DS_ID = 0xFF000000;
 
-  private int dsId;
+  @VisibleForTesting
+  public static final int PLACE_HOLDER_FOR_TYPE_ID = 0xFFFFFF;
+  private static final int PLACE_HOLDER_FOR_DS_ID = 0xFF000000;
+
+  private int distributedSystemId;
   private final int maxTypeId;
   private volatile DistributedLockService dls;
   private final Object dlsLock = new Object();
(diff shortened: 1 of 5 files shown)


pivotal-jbarrett:
I defaulted to the interface and class name but don’t have a strong preference either way. PDX is likely more recognizable by customers.

demery-pivotal:
Do we want meters in addition to (or instead of) statistics?

geode-core/src/main/java/org/apache/geode/pdx/internal/TypeRegistrationStatistics.java
```diff
@@ -0,0 +1,56 @@
...
+ * the License.
+ */
+
+package org.apache.geode.pdx.internal;
+
+import org.apache.geode.StatisticDescriptor;
+import org.apache.geode.Statistics;
+import org.apache.geode.StatisticsFactory;
+import org.apache.geode.StatisticsType;
+
```
demery-pivotal: With a new class, we have an opportunity to name the instrumentation methods to focus on reporting stuff, rather than on manipulating statistics. For example: Something like `pdxTypeDefined()` and `pdxTypeCreated()`.

Similar for the class name… Is there a good name that focuses on reporting occurrences of interesting events, rather than on statistics? I don't have a great suggestion for what to name the class. My not-necessarily-great attempt is `TypeRegistrationReporter`.

pivotal-jbarrett: I kind of like this idea but what other sort of stuff would we put in these methods? It wouldn't make sense for them to write to the log. So outside of stats, what are you thinking?

If we limit it to stats I still think changing the method names is a reasonable ask. It certainly makes it more readable.

pivotal-jbarrett:
Only CI failure is known issue [GEODE-6499](https://issues.apache.org/jira/browse/GEODE-6499).

geode-core/src/main/java/org/apache/geode/pdx/internal/TypeRegistrationStatistics.java
pivotal-jbarrett: I left the class name the same but did change the methods to be more readable. Thoughts?

pivotal-jbarrett:
@upthewaterspout changed the stats name and added enum. Also added a new distributed test to verify the behavior of the define/create/size across servers.

geode-core/src/main/java/org/apache/geode/pdx/internal/TypeRegistrationStatistics.java
demery-pivotal: My hope for the future is this: We can hide a certain amount of logging and measuring in the same instrumentation class. In my future fantasy world, the code reports an event to the instrumentation class, and the instrumentation class logs, measures, or both.

A suffix more general than 'stats' might invite us to put some logging in there. The best I can think of is 'Instrumentation' (as in some of the examples from [Pete Hodgson's instrumentation article](https://martinfowler.com/articles/domain-oriented-observability.html)).

I'd like to move in that direction at some point. I recognize that shifting the style will put an extra burden on whoever writes the first few new instrumentation classes.

So given all of that (non-blocking) input, do what you think best.

pivotal-jbarrett: I think out of scope for this because we really need a proposal and examples to drive this. I don't want to trailblaze this.

geode-core/src/main/java/org/apache/geode/pdx/internal/PeerTypeRegistration.java
```diff
@@ -107,18 +110,21 @@ public PeerTypeRegistration(InternalCache cache) {
     if (distributedSystemId == -1) {
       distributedSystemId = 0;
     }
-    this.dsId = distributedSystemId << 24;
-    this.maxTypeId = 0xFFFFFF;
+    this.distributedSystemId = distributedSystemId << 24;
+    maxTypeId = 0xFFFFFF;
```
gesterzhou: it's better to be this.maxTypeId = 0xFFFFFF;

pivotal-jbarrett: `this` is redundant in that context.

pivotal-jbarrett: Did some readability cleanup on the class, so the members being set from parameters are more clear.

PR #3922, opened 2019-08-14
 GEODE-7085: Ensuring the bitset stays within BIT_SET_WIDTH

upthewaterspout:
Ensuring that when we call recordVersion on a RegionVersionHolder,
we appropriately move the bitSet to match the new version we are
recording, rather than trying to expand it. In particular, if new
version is greater than Integer.MAX_VALUE, we can't record than in out
integer indexed bit set.

This change rewrites addBitSetExceptions. The logic is now broken into a
BitSetExceptionIterator, which converts some or all of the bit set into
RVVException objects, and the logic to slide the bit set forward to a
new bitSetVersion.

Adding unit tests that show that large versions cause an
IndexOutOfBounds exception from recordGCVersion. Adding more unit tests
for the internal state of the bitset.

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/versions/RegionVersionHolder.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/versions/RegionVersionHolder.java
@@ -162,7 +162,10 @@ public class RegionVersionHolder<T> implements Cloneable, DataSerializable {
     return getExceptions().toString();
   }
 
-  public void setVersion(long ver) {
+  /**
+   * Should only be called as part of cloning a RegionVersionHolder
+   */
+  void setVersion(long ver) {
     this.version = ver;
   }
 
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/versions/RegionVersionHolder.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/versions/RegionVersionHolder.java
@@ -250,8 +250,7 @@ public class RegionVersionHolder<T> implements Cloneable, DataSerializable {
       return; // it fits in this bitset
     }
 
-    int length = BIT_SET_WIDTH;
-    int bitCountToFlush = length * 3 / 4;
+    int bitCountToFlush = BIT_SET_WIDTH * 3 / 4;
 
     // We can only flush up to the last set bit because
     // the exceptions list includes a "next version" that indicates a received version.
@@ -263,7 +262,7 @@ public class RegionVersionHolder<T> implements Cloneable, DataSerializable {
     }
     // see if we can shift part of the bits so that exceptions in the recent bits can
     // be kept in the bitset and later filled without having to create real exception objects
-    if (version >= this.bitSetVersion + length + bitCountToFlush) {
+    if (bitCountToFlush == -1 || version >= this.bitSetVersion + BIT_SET_WIDTH + bitCountToFlush) {
       // nope - flush the whole bitset
       addBitSetExceptions(version);
     } else {
@@ -299,7 +298,6 @@ public class RegionVersionHolder<T> implements Cloneable, DataSerializable {
    *
    */
   private void addBitSetExceptions(long newVersion) {
-
     if (newVersion <= bitSetVersion) {
       return;
     }
@@ -374,7 +372,7 @@ public class RegionVersionHolder<T> implements Cloneable, DataSerializable {
     if (bitToSet > BIT_SET_WIDTH) {
       Assert.fail("Trying to set a bit larger than the size of the bitset " + bitToSet);
     }
-    this.bitSet.set((int) bitToSet);
+    this.bitSet.set(Math.toIntExact(bitToSet));
   }
 
   private void logRecordVersion(long version) {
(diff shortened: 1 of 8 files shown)


geode-core/src/main/java/org/apache/geode/internal/cache/versions/RegionVersionHolder.java
```diff
@@ -384,7 +370,11 @@ private void recordVersionWithBitSet(long version) {
   }
 
   private void setVersionInBitSet(long version) {
-    this.bitSet.set((int) (version - this.bitSetVersion));
+    long bitToSet = version - this.bitSetVersion;
+    if (bitToSet > BIT_SET_WIDTH) {
+      Assert.fail("Trying to set a bit larger than the size of the bitset " + bitToSet);
```
metatype: Instead of Assert should we use `Math.toIntExact()`?

geode-core/src/main/java/org/apache/geode/internal/cache/versions/RegionVersionHolder.java
```diff
@@ -280,63 +292,37 @@ private synchronized void mergeBitSet() {
...
    */
-  private void addBitSetExceptions(int numBits, long newVersion) {
-    final boolean isDebugEnabled_RVV = logger.isTraceEnabled(LogMarker.RVV_VERBOSE);
-    int lastSetIndex = -1;
+  private void addBitSetExceptions(long newVersion) {
 
-    if (isDebugEnabled_RVV) {
-      logger.trace(LogMarker.RVV_VERBOSE, "addBitSetExceptions({},{})", numBits, newVersion);
+    if (newVersion <= bitSetVersion) {
+      return;
```
bschuchardt: shouldn't we throw an exception instead of ignoring the version?

geode-core/src/main/java/org/apache/geode/internal/cache/versions/RegionVersionHolder.java
upthewaterspout: This assertion is a little stronger than toIntExact. We are asserting that the bitToSet is less than or equal to BIT_SET_WIDTH (around 1K).

geode-core/src/main/java/org/apache/geode/internal/cache/versions/RegionVersionHolder.java
```diff
@@ -245,6 +253,10 @@ void flushBitSetDuringRecording(long version) {
     int length = BIT_SET_WIDTH;
     int bitCountToFlush = length * 3 / 4;
 
+    // We can only flush up to the last set bit because
+    // the exceptions list includes a "next version" that indicates a received version.
+    bitCountToFlush = bitSet.previousSetBit(bitCountToFlush);
```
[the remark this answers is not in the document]
upthewaterspout: Good point. I will add a check.

geode-core/src/main/java/org/apache/geode/internal/cache/versions/RegionVersionHolder.java
upthewaterspout: I think the contract for this method is that it adds exceptions out of the bit set up to newVersion. So maybe it's reasonable that it just does nothing if newVersion < bitSetVersion? The old code also just did nothing in that case.

geode-core/src/main/java/org/apache/geode/internal/cache/versions/RegionVersionHolder.java
metatype: Good point.  I'm ambivalent about using `Assert` instead of throwing an exception always.

geode-core/src/test/java/org/apache/geode/internal/cache/versions/RegionVersionHolderBitSetJUnitTest.java
```diff
@@ -0,0 +1,165 @@
...
+    long version = ((long) Integer.MAX_VALUE) + 10L;
+    h.recordVersion(version);
+    assertEquals(version, h.getBitSetVersionForTesting());
+    assertEquals(bitSet(0), h.getBitSetForTesting());
+    assertContains(h, version);
+    assertHasExceptions(h, RVVException.createException(0, version));
+  }
+
+  @Test
+  public void recordVersionLessThanBitSetWidthShouldNotMoveBItSet() {
```
metatype: Decapitalize `BIt`...?

geode-core/src/main/java/org/apache/geode/internal/cache/versions/RegionVersionHolder.java
bschuchardt: It looks like the old method didn't even use the newVersion parameter outside of the trace-level log statement.  It only paid attention to the numBits parameter.

geode-core/src/main/java/org/apache/geode/internal/cache/versions/RegionVersionHolder.java
```diff
@@ -280,63 +291,36 @@ private synchronized void mergeBitSet() {
...
-    if (lastSetIndex > 0) {
-      this.bitSetVersion = this.bitSetVersion + (long) lastSetIndex;
+
+    // Move the data in the bitset forward to reflect the new version
+    if (newVersion > bitSetVersion + bitSet.size()) {
+      // Optimization - if the new version is past the end of the bitset, just clear the bitset
+      bitSet.clear();
+    } else {
+      // Otherwise slide the bitset over to the new offset
+      int offsetIncrease = (int) (newVersion - bitSetVersion);
```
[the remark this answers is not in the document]
upthewaterspout: Well, this is in the else clause - we we know that newVersion <= bitSetVersion+bitSet.size(). But maybe it would have been better to use Math.toIntExact here to validate that.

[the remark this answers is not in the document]
mcmellawatt: That makes sense, I was doing simple math wrong in my head ;P

PR #3926, opened 2019-08-14
GEODE-7087: Reset flag after unlock bucket primary lock.

pivotal-eshu:


geode-core/src/main/java/org/apache/geode/internal/cache/TXState.java
```diff
@@ -932,13 +932,19 @@ void doCleanup() {
       synchronized (this.completionGuard) {
         this.completionGuard.notifyAll();
       }
-
+      if (isBucketLocks()) {
```
agingade: not knowing complete logic; just wondering, if this check should come before notifying the completionGuard (above check).

geode-core/src/main/java/org/apache/geode/internal/cache/TXState.java
```diff
@@ -898,47 +898,65 @@ void doCleanup() {
...
+      synchronized (this.completionGuard) {
+        this.completionGuard.notifyAll();
+      }
+      if (exception != null && !this.proxy.getCache().isClosed()) {
+        throw exception;
+      }
+    }
+  }
+
+  private void cleanupTXRegionState() {
```
agingade: Now the logic is separated, can we add unit tests for cleanupTxRegionState() and unlockPrimaryMoveReadLock() functions.

PR #3927, opened 2019-08-14
GEODE-6805: Allow shorthand ~ to represent user home in jar deployment.

alex-grisham:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/management/internal/cli/GfshParser.java
+++ b/geode-core/src/main/java/org/apache/geode/management/internal/cli/GfshParser.java
@@ -203,6 +203,8 @@ public class GfshParser extends SimpleParser {
 
     // this tells the simpleParser not to interpret backslash as escaping character
     rawInput = rawInput.replace("\\", "\\\\");
+    // this allows the usage of home shorthand across platforms
+    rawInput = rawInput.replace("~", System.getProperty("user.home"));
     // User SimpleParser to parse the input
     ParseResult result = super.parse(rawInput);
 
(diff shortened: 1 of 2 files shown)


geode-core/src/main/java/org/apache/geode/management/internal/cli/GfshParser.java
```diff
@@ -203,6 +203,8 @@ public GfshParseResult parse(String userInput) {
 
     // this tells the simpleParser not to interpret backslash as escaping character
     rawInput = rawInput.replace("\\", "\\\\");
+    // this allows the usage of home shorthand across platforms
+    rawInput = rawInput.replace("~", System.getProperty("user.home"));
```
pivotal-jbarrett: I don't think you intend for `/~/`, `f~oo`, or other combinations of '~' to be replaced. The '~' should only be replaced if it anchors the beginning of the path and is immediately followed by a path separator. Something like this regex `^~(?:\/)` (untested) should do the trick.

geode-core/src/test/java/org/apache/geode/management/internal/cli/commands/DeployCommandTest.java
```diff
@@ -57,4 +57,10 @@ public void bothDirAndJar() {
   public void missingDirOrJar() {
     gfsh.executeAndAssertThat(command, "deploy").statusIsError().containsOutput("is required");
   }
+
+  @Test
+  public void convertShorthandToPathJar() {
+    gfsh.executeAndAssertThat(command, "deploy --jar=~/abc.jar").statusIsError()
```
pivotal-jbarrett: Add a test that `--jar=~foo/bar.jar` and `--jar=foo~/bar.jar` are not modified.

jdeppe-pivotal: Technically (at least considering shell semantics) `~foo` is legitimate (it expands to the home directory of the `foo` user if that user exists) but I don't think we should attempt to expand that.

PR #3929, opened 2019-08-14
GEODE-7089: Each client registration thread uses its own queue

mcmellawatt:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientRegistrationEventQueueManager.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientRegistrationEventQueueManager.java
@@ -15,21 +15,20 @@
 
 package org.apache.geode.internal.cache.tier.sockets;
 
-import java.util.Map;
 import java.util.Queue;
 import java.util.Set;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.locks.ReadWriteLock;
-import java.util.concurrent.locks.ReentrantLock;
 
 import org.apache.logging.log4j.Logger;
 
+import org.apache.geode.annotations.VisibleForTesting;
 import org.apache.geode.internal.cache.Conflatable;
 import org.apache.geode.internal.cache.EntryEventImpl;
 import org.apache.geode.internal.cache.FilterProfile;
 import org.apache.geode.internal.cache.FilterRoutingInfo;
 import org.apache.geode.internal.cache.InternalCacheEvent;
 import org.apache.geode.internal.cache.LocalRegion;
+import org.apache.geode.internal.concurrent.ConcurrentHashSet;
 import org.apache.geode.internal.logging.LogService;
 
 /**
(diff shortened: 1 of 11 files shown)


geode-core/src/test/java/org/apache/geode/internal/cache/tier/sockets/CacheClientNotifierTest.java
```diff
@@ -140,6 +142,35 @@ public void eventsInClientRegistrationQueueAreSentToClientAfterRegistrationIsCom
     verify(cacheClientProxy, times(1)).deliverMessage(isA(HAEventWrapper.class));
   }
 
+  @Test
+  public void clientRegistrationFailsQueueStillDrained()
```
nabarunnag: Hi Ryan, just wanted to know that the test mentions that clientRegistrationFails**QueueStillDrains**. Is there an assert needed to make sure that the queue actually drained? the last assert is that an IOException is thrown

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientRegistrationEventQueueManager.java
```diff
@@ -99,6 +97,132 @@ void add(final InternalCacheEvent event,
...
+
+      if (logger.isDebugEnabled()) {
+        logger.debug("Draining events from registration queue for client proxy "
+            + clientProxyMembershipID
+            + " without synchronization");
+      }
+
+      CacheClientProxy cacheClientProxy = cacheClientNotifier
+          .getClientProxy(clientProxyMembershipID);
+
```
nabarunnag: ryan, do you feel that a null check here will avoid couple of calls and locks ? I mean if cache client proxy is null we just decrement the put in progress counter and return from here.

geode-core/src/test/java/org/apache/geode/internal/cache/tier/sockets/CacheClientNotifierTest.java
mcmellawatt: You are right that this test is not really complete.  I think initially I was punting on that validation because the registration queue manager is not publicly accessible.  I am going to make a commit so we can inject and verify these calls on the queue manager.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientRegistrationEventQueueManager.java
mcmellawatt: So we still need to do the locking so that no new events are added while we are processing and removing the queue.  We need to decrement the put in progress counter on every event in the queue, and also need to ensure no new events are being added while we do that (otherwise we would remove the queue from the manager while it may still have events in it, thereby resulting in events leaking without the counter being decremented).  If you look at the drain logic, the only thing it does is decrement the putinprogress counter for each event if the proxy is null, so we aren't really doing any unnecessary work.  But we do need to do that for every event in the queue under synchronization.

mcmellawatt: I think maybe the confusion here is that the put in progress counter is incremented _per event received_ during registration, not just once for the whole registration.

mcmellawatt: I added a new test to ensure the put in progress counter is properly incremented and decremented for each event.  This conversation made me realize that we need a test for that case.

PR #3936, opened 2019-08-15
GEODE-6945: expand the type to related xml attributes

jinmeiliao:

### For all changes:

### Note:


jinmeiliao:
@jmelchio

jinmeiliao:
@dschneider-pivotal please re-review. I changed the approach: instead of not expanding the attributes in `CreateRegionCommand`, I chose to expand it everywhere, including the `RegionConverter`. This is because I found the subsequent code kind of relying on the DataPolicy being set all the time in xml configuration objects. If we don't do this, we get NPE everywhere, so it's easier just to expand it.

geode-management/src/main/java/org/apache/geode/cache/configuration/RegionType.java
```diff
@@ -47,5 +47,25 @@
 
   // this is used to represent regions not supported by the management V2 API. For example Gfsh can
   // create regions with "LOCAL*" types
-  UNSUPPORTED
+  UNSUPPORTED;
+
+  public boolean isProxy() {
```
dschneider-pivotal: I like these new "is" methods but add javadocs to make clear that it returns true if the type is any of the types that are proxies/partitioned/replicate.
In the past on the core apis we had some confusion on methods like these since a method like isReplicate could mean "it is the REPLICATE type" instead of "it is one of the replication types".

geode-core/src/test/java/org/apache/geode/management/internal/configuration/converters/RegionConverterTest.java
```diff
@@ -91,8 +100,27 @@ public void fromConfig() throws Exception {
...
     RegionAttributesType regionAttributes = config.getRegionAttributes();
+    assertThat(regionAttributes.getDataPolicy()).isEqualTo(RegionAttributesDataPolicy.REPLICATE);
     assertThat(regionAttributes.getKeyConstraint()).isEqualTo("bar");
     assertThat(regionAttributes.getValueConstraint()).isEqualTo("foo");
     assertThat(regionAttributes.getDiskStoreName()).isEqualTo("diskstore");
   }
+
+  @Test
+  public void checkDefaultRegionAttributesForShortcuts() throws Exception {
+    URL xmlResource = RegionConverterTest.class.getResource("RegionConverterTest.xml");
```
dschneider-pivotal: If is not clear to me how valuable this test is. It seems like we just duplicate everything that is in our management code that defines a type into the RegionConverterTest.xml. The xml sets both the refid and the nested region attributes.
Maybe this is worth testing.
But what seems more valuable, and what is not tested, is that the attributes returned from "createRegionAttributesByType" and the same as what the core defines for each RegionShortcut.

jinmeiliao: this xml was originally created by using the method in cache and saved as a master xml.

geode-management/src/main/java/org/apache/geode/cache/configuration/RegionType.java
jinmeiliao: remove these methods for now, we don't need them. And we might change these types anyway.

PR #3944, opened 2019-08-17
GEODE-7088: Using ConcurrentSets for interested clients

mcmellawatt:
This fix is to handle a very specific race condition scenario as described here:

Subscription HA is enabled and a server is providing a client queue image to a peer (serializing the queue). Meanwhile, a client is also just finished client subscription registration with that same server and is recalculating its filter info to determine if the client needs the event. Recalculating the filter info results in the client message in the HAContainer to be mutated, which causes a ConcurrentModificationException to occur in the GII provider thread.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/CacheClientNotifier.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/CacheClientNotifier.java
@@ -103,6 +103,7 @@ import org.apache.geode.internal.cache.tier.CommunicationMode;
 import org.apache.geode.internal.cache.tier.MessageType;
 import org.apache.geode.internal.cache.tier.OverflowAttributes;
 import org.apache.geode.internal.cache.versions.VersionTag;
+import org.apache.geode.internal.concurrent.ConcurrentHashSet;
 import org.apache.geode.internal.logging.InternalLogWriter;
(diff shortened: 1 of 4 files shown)


geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/CacheClientNotifier.java
```diff
@@ -907,7 +911,7 @@ private void singletonRouteClientMessage(Conflatable conflatable,
    * collection of non-durable identifiers of clients connected to this VM
    */
   Set<ClientProxyMembershipID> getProxyIDs(Set mixedDurableAndNonDurableIDs) {
-    Set<ClientProxyMembershipID> result = new HashSet<>();
+    Set<ClientProxyMembershipID> result = new ConcurrentHashSet<>();
```
agingade: Not seeing any requirement of making this ConcurrentHashSet; unless the returned set is modified, accessed in other parts at the same time.

mcmellawatt: It's hard to see in the diff, but we assign the result to _clientInterestList and _clientInterestListInv which are accessed concurrently by multiple client registration threads/GII threads.

PR #3957, opened 2019-08-21
GEODE-6944: separate regionType and region shortcut

jinmeiliao:

* we can create a region config object using a shortcut, but can not set it after it's created
* regionType is a more distilled enum that more corresponds to data policy now
* the converter will recognize more region types created by gfsh or cache.xml

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/management/internal/configuration/converters/RegionConverter.java
+++ b/geode-core/src/main/java/org/apache/geode/management/internal/configuration/converters/RegionConverter.java
@@ -28,25 +28,18 @@ public class RegionConverter extends ConfigurationConverter<Region, RegionConfig
   protected Region fromNonNullXmlObject(RegionConfig xmlObject) {
     Region region = new Region();
     region.setName(xmlObject.getName());
-    if (xmlObject.getType() == null) {
-      // older gfsh would generate the region xml without the refid/type. we will not
-      // support showing these regions in management rest api for now.
-      region.setType(RegionType.UNSUPPORTED);
-    } else {
-      try {
-        region.setType(RegionType.valueOf(xmlObject.getType()));
-      } catch (IllegalArgumentException e) {
-        // Management rest api will not support showing regions with "LOCAL*" types or user defined
-        // refids
-        region.setType(RegionType.UNSUPPORTED);
-      }
-    }
     RegionAttributesType regionAttributes = xmlObject.getRegionAttributes();
+    region.setType(getRegionType(regionAttributes));
 
     if (regionAttributes != null) {
       region.setDiskStoreName(regionAttributes.getDiskStoreName());
       region.setKeyConstraint(regionAttributes.getKeyConstraint());
       region.setValueConstraint(regionAttributes.getValueConstraint());
+      RegionAttributesType.PartitionAttributes partitionAttributes =
+          regionAttributes.getPartitionAttributes();
+      if (partitionAttributes != null && partitionAttributes.getRedundantCopies() != null) {
+        region.setRedundantCopies(Integer.parseInt(partitionAttributes.getRedundantCopies()));
+      }
     }
     return region;
   }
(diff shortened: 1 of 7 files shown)


geode-core/src/main/java/org/apache/geode/management/internal/configuration/converters/RegionConverter.java
```diff
@@ -68,27 +61,90 @@ protected RegionConfig fromNonNullConfigObject(Region configObject) {
     return region;
   }
 
+  /**
+   * Data policy to regionType is almost a 1-to-1 mapping, except in
+   * the case of DataPolicy.PARTITION, we will need to see the local max memory
+   * to determine if it's a PARTITION type or a PARTITION_PROXY type.
+   *
+   * we do our best to infer the type from the existing xml attributes. For data
+   * policies not supported by management rest api (for example, NORMAL and RELOADED)
```
dschneider-pivotal: typo: "RELOADED" -> "PRELOADED"

geode-core/src/main/java/org/apache/geode/management/internal/configuration/converters/RegionConverter.java
```diff
@@ -68,27 +61,90 @@ protected RegionConfig fromNonNullConfigObject(Region configObject) {
...
+   * it will show as UNSUPPORTED
+   */
+  public RegionType getRegionType(RegionAttributesType regionAttributes) {
+    if (regionAttributes == null) {
+      return RegionType.UNSUPPORTED;
+    }
+    RegionAttributesDataPolicy dataPolicy = regionAttributes.getDataPolicy();
+
+    if (regionAttributes.getDataPolicy() == null) {
```
dschneider-pivotal: use the "dataPolicy" local var here instead of fetching it again

geode-core/src/main/java/org/apache/geode/management/internal/configuration/converters/RegionConverter.java
```diff
@@ -68,27 +61,90 @@ protected RegionConfig fromNonNullConfigObject(Region configObject) {
...
+      return RegionType.UNSUPPORTED;
+    }
+
+    switch (dataPolicy.name()) {
```
dschneider-pivotal: don't use .name() here. Instead switch on the actual enum and each case can be an actual enum instead of a string

geode-core/src/main/java/org/apache/geode/management/internal/configuration/converters/RegionConverter.java
```diff
@@ -68,27 +61,90 @@ protected RegionConfig fromNonNullConfigObject(Region configObject) {
...
+      case "PARTITION": {
+        RegionAttributesType.PartitionAttributes partitionAttributes =
+            regionAttributes.getPartitionAttributes();
+        if (partitionAttributes != null && partitionAttributes.getLocalMaxMemory().equals("0")) {
```
dschneider-pivotal: Should you also handle a null and empty string returned by getLocalMaxMemory() here?
I think both of these would indicate a default value and the default for this one should be zero.

geode-management/src/main/java/org/apache/geode/cache/configuration/RegionType.java
```diff
@@ -18,34 +18,32 @@
...
-  REPLICATE_PERSISTENT_OVERFLOW,
-  REPLICATE_HEAP_LRU,
-
   REPLICATE_PROXY,
 
   // this is used to represent regions not supported by the management V2 API. For example Gfsh can
   // create regions with "LOCAL*" types
   UNSUPPORTED;
+
+  public boolean isProxy() {
```
dschneider-pivotal: add javadocs for these "is*" methods

geode-management/src/main/java/org/apache/geode/management/configuration/Region.java
```diff
@@ -89,6 +119,14 @@ public void setType(RegionType type) {
     this.type = type;
   }
 
+  public int getRedundantCopies() {
```
dschneider-pivotal: You have javadocs on the private field but then no javadocs on the external public methods

geode-management/src/main/java/org/apache/geode/management/configuration/RegionShortcut.java
```diff
@@ -0,0 +1,101 @@
...
+  REPLICATE_PERSISTENT_OVERFLOW,
+  REPLICATE_HEAP_LRU,
+
+  REPLICATE_PROXY;
+
+  /**
+   * @return the corresponding region type of each of these region shortcut
+   */
+  public RegionType getRegionType() {
+    if (withPartition()) {
```
dschneider-pivotal: Instead of implementing this with a bunch of logic, just add a "private final RegionType regionType" field to this enum class. You can initialize each corresponding regionType when you declare the enums and then this method can just return this.regionType.

jinmeiliao: I tried that, but this looks cleaner. The other way makes the shortcut enum class cluttered, but it's just my opinion.

PR #3958, opened 2019-08-21
GEODE-7106: Prevent NFE During PRid Generation

jujoramos:
- Added tests.
- Fixed minor warnings.
- While releasing the distributed lock acquired to generate a new
  global partitioned region ID, don't parse the String
  MAX_PARTITIONED_REGION_ID as an Integer.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/integrationTest/java/org/apache/geode/internal/cache/PartitionedRegionIntegrationTest.java
+++ b/geode-core/src/integrationTest/java/org/apache/geode/internal/cache/PartitionedRegionIntegrationTest.java
@@ -25,11 +25,13 @@ import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.spy;
 
-import java.io.File;
-import java.io.IOException;
-import java.nio.file.Files;
+import java.util.List;
 import java.util.concurrent.ScheduledExecutorService;
 
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.core.LogEvent;
+import org.apache.logging.log4j.core.Logger;
+import org.apache.logging.log4j.test.appender.ListAppender;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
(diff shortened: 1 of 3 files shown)


geode-core/src/integrationTest/java/org/apache/geode/internal/cache/PartitionedRegionIntegrationTest.java
```diff
@@ -49,10 +70,37 @@ public void bucketSorterShutdownAfterRegionDestroy() {
...
+    server.startServer();
     PartitionedRegion region =
         (PartitionedRegion) server.createRegion(RegionShortcut.PARTITION, "PR1",
             rf -> rf.setOffHeap(false));
     ScheduledExecutorService bucketSorter = region.getBucketSorter();
     assertThat(bucketSorter).isNull();
   }
+
+  @Test
+  // See GEODE-7106
```
agingade: I don't think we follow adding ticket number in source/test code.

geode-core/src/integrationTest/java/org/apache/geode/internal/cache/PartitionedRegionIntegrationTest.java
```diff
@@ -49,10 +70,37 @@ public void bucketSorterShutdownAfterRegionDestroy() {
...
+
+    server.withProperty("log-level", "FINE")
+        .withProperty("log-file", customLogFile.getAbsolutePath()).startServer();
+    PartitionedRegion region =
+        spy((PartitionedRegion) server.createRegion(RegionShortcut.PARTITION, methodName));
+    doReturn(mockLockService).when(region).getPartitionedRegionLockService();
+
+    assertThatCode(() -> region.generatePRId(server.getCache().getInternalDistributedSystem()))
+        .doesNotThrowAnyException();
+    assertThat(Files.lines(customLogFile.toPath())
```
agingade: Can this check be done by looking at logger.warn invocation? If there was any exception the logger.warn would not have successfully called. In that case we could remove checking for the exception string in the log file.

geode-core/src/integrationTest/java/org/apache/geode/internal/cache/PartitionedRegionIntegrationTest.java
jujoramos: Hey @agingade,
Thanks for reviewing this!.
Even though is not a "documented process", I saw a lot of places in the code and tests where we actually reference GEODE ticket numbers. I can remove it if you want, but I believe is a good practice to keep the context and history of changes.

geode-core/src/integrationTest/java/org/apache/geode/internal/cache/PartitionedRegionIntegrationTest.java
jujoramos: Done!

PR #3959, opened 2019-08-21
GEODE-7108: fixing a memory leak due to GCMetricsBinders not being closed

mhansonp:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/metrics/GeodeCompositeMeterRegistry.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/metrics/GeodeCompositeMeterRegistry.java
@@ -1,0 +1,36 @@
+package org.apache.geode.internal.metrics;
+
+import io.micrometer.core.instrument.binder.jvm.JvmGcMetrics;
+import io.micrometer.core.instrument.binder.jvm.JvmMemoryMetrics;
+import io.micrometer.core.instrument.binder.jvm.JvmThreadMetrics;
+import io.micrometer.core.instrument.binder.system.FileDescriptorMetrics;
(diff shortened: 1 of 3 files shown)


geode-core/src/main/java/org/apache/geode/internal/metrics/GeodeCompositeMeterRegistry.java
```diff
@@ -0,0 +1,36 @@
...
+
+    final private JvmMemoryMetrics jvmMemoryMetrics =  new JvmMemoryMetrics();
+    final private JvmThreadMetrics jvmThreadMetrics =  new JvmThreadMetrics();
+    final private JvmGcMetrics jvmGcMetrics =  new JvmGcMetrics();
+    final private ProcessorMetrics processorMetrics =  new ProcessorMetrics();
+    final private UptimeMetrics uptimeMetrics =  new UptimeMetrics();
+    final private FileDescriptorMetrics fileDescriptorMetrics =  new FileDescriptorMetrics();
+
+
+    void registerBinders() {
```
demery-pivotal: - Move this code and the rest of `CacheMeterRegistryFactory` into a constructor in `GeodeCompositeMeterRegistry`.
- Use `GeodeCompositeMeterRegistry::new` as the `CompositeMeterRegistryFactory`.
- Delete `CacheMeterRegistryFactory`.

Consider passing the set of binders into the constructor (from another constructor), so that we can pass mock binders from a unit tests, to ensure `bindTo(this)` is called on each during the constructor, and `close()` is called on each during `close()`.

mhansonp: Great suggestions! I don't know what I was thinking. Of course this should have a unit test.

PR #3960, opened 2019-08-21
GEODE-7108: fixing a memory leak

mhansonp:
- GCMetricsBinders not being closed was leaking memory

Co-Authored-By: Aaron Lindsey <alindsey@pivotal.io>

Closed original pull request. Moved to a draft pull request. Need to see how the windows testing goes and get some feedback.


aaronlindsey:
FYI I think we should probably merge #3952 before this one because it adds all of the tests for the added binders.

geode-core/src/main/java/org/apache/geode/internal/metrics/GeodeCompositeMeterRegistry.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.Set;
+
+import io.micrometer.core.instrument.binder.MeterBinder;
+import io.micrometer.core.instrument.composite.CompositeMeterRegistry;
+
+public class GeodeCompositeMeterRegistry extends CompositeMeterRegistry {
+
+  private final Set<MeterBinder> binders;
+
+  public GeodeCompositeMeterRegistry(Set<MeterBinder> binders) {
```
aaronlindsey: I think it would be more standard to make `binders` and `this.binders` a `List<MeterBinder>` since we're only ever iterating over them and not doing add/remove/contains operations.

geode-core/src/main/java/org/apache/geode/internal/metrics/GeodeCompositeMeterRegistry.java
```diff
@@ -0,0 +1,52 @@
...
+ * the License.
+ */
+package org.apache.geode.internal.metrics;
+
+import java.util.Set;
+
+import io.micrometer.core.instrument.binder.MeterBinder;
+import io.micrometer.core.instrument.composite.CompositeMeterRegistry;
+
+public class GeodeCompositeMeterRegistry extends CompositeMeterRegistry {
```
aaronlindsey: Should this class extend `AutoCloseable` since we're making it contain resources that need to be cleaned up?

geode-core/src/main/java/org/apache/geode/internal/metrics/GeodeCompositeMeterRegistry.java
```diff
@@ -0,0 +1,52 @@
...
+
+  @Override
+  public void close() {
+    for (MeterBinder binder : binders) {
+      if (binder instanceof AutoCloseable) {
+        AutoCloseable autoCloseable = (AutoCloseable) binder;
+        try {
+          autoCloseable.close();
+        } catch (Exception ignored) {
+          // Shutting down anyway.
```
pivotal-jbarrett: Probably worth logging.

geode-core/src/main/java/org/apache/geode/internal/cache/GemFireCacheImpl.java
```diff
@@ -2125,6 +2125,10 @@ public void close(String reason, Throwable systemFailureCause, boolean keepAlive
 
         // do this before closing regions
         resourceManager.close();
+        meterRegistry.close();
+
+        for (MeterRegistry tmpMeterRegistry : meterSubregistries)
+          tmpMeterRegistry.close();
```
pivotal-jbarrett: I would hesitate to close registries Geode didn't install. Consider a client application providing its own registry that records meters for itself and wants Geode recorded to the same registry. Now as the cache is closed so will the user's registry and their application meters will not record anymore.

geode-core/src/main/java/org/apache/geode/internal/metrics/GeodeCompositeMeterRegistry.java
aaronlindsey: Or perhaps `Iterable<MeterBinder>`

geode-core/src/main/java/org/apache/geode/internal/metrics/GeodeCompositeMeterRegistry.java
mhansonp: Will do.

geode-core/src/main/java/org/apache/geode/internal/cache/GemFireCacheImpl.java
mhansonp: Perhaps I am not understanding... If the cache is closed, isn't the result same? This just cleans up everything when the cache goes away. If we don't do that, I think they might leak memory like we were doing before.

geode-core/src/main/java/org/apache/geode/internal/metrics/GeodeCompositeMeterRegistry.java
mhansonp: ```AutoCloseable``` only closes in a try-with-resources block which this is not. At least, that is my understanding from the javadoc.

geode-core/src/main/java/org/apache/geode/internal/metrics/GeodeCompositeMeterRegistry.java
mhansonp: I thought about using a List. Conferring with Dale he suggested a Set. I think both are sufficient.  Set provides protection against duplicates which concerned Dale, but not me in the context(rightly or wrongly so). Hence, why I went with the Set.

geode-core/src/main/java/org/apache/geode/internal/metrics/GeodeCompositeMeterRegistry.java
aaronlindsey: In my mind, the primary advantage of extending the `AutoCloseable` interface in this case is that it tells the user of this class that the class encapsulates resources that need to be closed. A secondary benefit is that it provides a convenient way to allow this class to be used within a try-with-resources block (although it is not currently used like that, as you pointed out).

geode-core/src/main/java/org/apache/geode/internal/metrics/GeodeCompositeMeterRegistry.java
```diff
@@ -0,0 +1,52 @@
...
+public class GeodeCompositeMeterRegistry extends CompositeMeterRegistry {
+
+  private final Set<MeterBinder> binders;
+
+  public GeodeCompositeMeterRegistry(Set<MeterBinder> binders) {
+    super();
+    this.binders = binders;
+  }
+
+  void registerBinders() {
```
demery-pivotal: I'm not thrilled that the configuration is split between this method (to register the binders) and the factory (to configure everything else). Splitting the configuration seems troublesome. I have an idea that I'm also not sure I like: Let the factory do _all_ of the configuration (including binding the binders) and pass `Collection<Closeable>`  to this class's constructor. Then let this class simply close the `Closeable`s.

geode-core/src/main/java/org/apache/geode/internal/metrics/GeodeCompositeMeterRegistry.java
demery-pivotal: > Or perhaps `Iterable<MeterBinder>`

`Iterable` makes it hard to create a stream to process the elements.

geode-core/src/main/java/org/apache/geode/internal/cache/GemFireCacheImpl.java
mhansonp: So are you proposing to pull the closure of submeterregistries? If so, I have done that it appears fine. Partly because there appears to be no submeterregistries.

PR #3961, opened 2019-08-21
GEODE-7112: add wait until user transaction is timed out.

pivotal-eshu:


geode-core/src/distributedTest/java/org/apache/geode/internal/jta/dunit/TxnTimeOutDUnitTest.java
```diff
@@ -205,8 +207,11 @@ public static void runTest1() throws Exception {
       Context ctx = cache.getJNDIContext();
       UserTransaction utx = (UserTransaction) ctx.lookup("java:/UserTransaction");
       utx.begin();
+      assertThat(utx.getStatus() == Status.STATUS_ACTIVE);
       utx.setTransactionTimeout(2);
-      Thread.sleep(6000);
+      long start = System.currentTimeMillis();
+      Thread.sleep(2000);
+      waitUntilTransactionTimeout(utx, start);
```
agingade: Can it be replaced using GeodeAwaitility and remove all the sleeps...
GeodeAwaitility.await().until(() -> utx.getStatus() != Status.STATUS_NO_TRANSACTION);

pivotal-eshu: No, it checks the thread that begins the transaction. GeodeAwaitility is another thread and it won't work.

agingade: PollInSameThread allows to wait in same thread.

GeodeAwaitility.await().pollInSameThread().until(() -> {
        return utx.getStatus() == Status.STATUS_NO_TRANSACTION;
      });

PR #3963, opened 2019-08-22
GEODE-7107: Introduce HttpService interface

jdeppe-pivotal:
- This allows for Jetty to be optional on the classpath.
- Marked as Experimental as there will be sebsequent changes to
  formalize this interface.

Authored-by: Jens Deppe <jdeppe@pivotal.io>

### For all changes:

### Note:


jdeppe-pivotal:
@demery-pivotal Please be aware that this PR is specifically to address [1] and isn't supposed to relate to your RFC [2]. I would imagine that ultimately you would end up reworking and/or moving this interface which is why I've marked it as `@Experimental`.

[1] https://issues.apache.org/jira/browse/GEODE-7107 
[2] https://cwiki.apache.org/confluence/display/GEODE/%5BDiscussion%5D+Public+API+to+Add+Endpoints+to+Geode+HTTP+Server

geode-core/src/main/java/org/apache/geode/cache/HttpService.java
```diff
@@ -0,0 +1,46 @@
...
+  /**
+   * Add a new web application in the form of a war file. This method is also implicitly
+   * responsible for starting the container if necessary.
+   *
+   * @param webAppContext the context path to be exposed for the web application
+   * @param warFilePath the absolute path to the war file
+   * @param attributeNameValuePairs attributes to be set on the servlet context
+   */
+  void addWebApplication(String webAppContext, String warFilePath,
+      Pair<String, Object>... attributeNameValuePairs) throws Exception;
```
pivotal-jbarrett: Feels a little heavy handed to make a third party library a required part of the API interface. What's wrong with using a `Map<String, Object>`?

pivotal-jbarrett: What about using `Path warFilePath` to have a strong type?

pivotal-jbarrett: Should this method return some sort of context object that can be used to control the newly created context? 
```
interface WebAppContext {
  void stop();
  void start();
  void destroy();
  Path getWarFilePath();
  String getWebContextPath()
}

File: geode-core/src/main/java/org/apache/geode/cache/HttpService.java [hunk 2]
```diff
@@ -0,0 +1,46 @@
...
+
+import org.apache.commons.lang3.tuple.Pair;
+
+import org.apache.geode.annotations.Experimental;
+import org.apache.geode.internal.cache.InternalCache;
+
+/**
+ * This interface provides access to the simple http service exposed on the Geode
+ * {@link InternalCache}. An instance can be retrieved by calling
+ * {@code InternalCache.getHttpService()}.
```
pivotal-jbarrett: Why is this "public" API only accessible by using an internal class?

geode-core/src/main/java/org/apache/geode/cache/HttpService.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+@Experimental
+public interface HttpService {
+
+  /**
+   * Stop the service. This should also stop any transitively managed components. Starting the
+   * service is implicitly handled by {@link #addWebApplication(String, String, Pair[])}
+   */
+  void stop();
```
pivotal-jbarrett: Seems dangerous to have a global stop method on this API. This should only be executed by the internal lifecycle when shutting down. See below for comment on per-context lifecycle.

geode-core/src/main/java/org/apache/geode/cache/HttpService.java
pivotal-jbarrett: Can we have more specific `throws` and declare in the API docs what is thrown and why?.

geode-core/src/main/java/org/apache/geode/cache/HttpService.java
```diff
@@ -0,0 +1,46 @@
...
+
+  /**
+   * Add a new web application in the form of a war file. This method is also implicitly
+   * responsible for starting the container if necessary.
+   *
+   * @param webAppContext the context path to be exposed for the web application
+   * @param warFilePath the absolute path to the war file
+   * @param attributeNameValuePairs attributes to be set on the servlet context
+   */
+  void addWebApplication(String webAppContext, String warFilePath,
```
demery-pivotal: This signature is troublesome. If the caller doesn't pass any pairs, Java warns of "Unchecked generics array creation for varargs parameter."

Consider changing this varargs parameter to take a collection/set/list of pairs, or perhaps a map of attribute names and values.

pivotal-jbarrett: The `webAppContext` is is ambiguous to me if it is the name or path of the web context. Perhaps `contextPath` or something. Must it be prefixed with `/` or will that be assumed? Maybe some more contract around the content.

jdeppe-pivotal:
@pivotal-jbarrett This PR is really just a temporary fix to address GEODE-7107. As mentioned, @demery-pivotal 's RFC should formalize this interface. To address some of your concerns I've pulled `HttpService` back into an `internal` package namespace.

geode-core/src/main/java/org/apache/geode/cache/HttpService.java
jdeppe-pivotal: Fixed

PR #3996, opened 2019-08-30
GEODE-7146: Improve Aggregate Function Docs

jujoramos:
- Corrected the return type description for SUM, COUNT and AVG.
- Re-introduced the standalone page entirely dedicated to aggregate
  functions, along with the relevant menu item.
- Documented current limitations of aggregate functions in relation
  to overflow.

### For all changes:

### Note:


davebarnes97:
@jujoramos Just curious - what prompted you to restore the separate aggregate page? I have no strong preference, just curious about the motivation.

jujoramos:
Hey @davebarnes97,
I reverted to a separate page just because it seems easier to find and read from a user's perspective (👍). The older approach had everything related to the `SELECT` statement in one single page, the new one contains a separate section with sub-sections per aggregate function, making easier to read the usage and limitations of each one.

davebarnes97:
Thanks for the clarification. With regard to ease of use, a follow-up: How did you arrive at the order in which the aggregate functions are presented? They could have been alphabetical (with AVG first). Is it based on frequency of use?

jujoramos:
Hey @davebarnes97: I guess that's just a personal preference, but I tried to present them in order of "complexity" from _easy_ to _hard_.
The first functions (`MIN` and `MAX`) are the easiest to understand, that's why they're at the top. After that `COUNT` and `SUM`,  not much to say about them. `AVG` goes at the very end basically because it's a "complex" computation that "uses" other two previously seen concepts (to calculate the `average` we need to `sum` all matching records and divide the result by the amount - `count` - of matching results).
Makes sense?.

davebarnes97:
Yes, that makes sense. However, the order in which the topics are presented does not match the order in which they appear in your introductory sentence. I would like to suggest that you take advantage of Bookbinder's auto-generated page-local TofC capability. Simply remove the  class="no-quick-link" tags from the <a> tags, and you'll have a list of live links to follow at the top of the page. Just a suggestion. I will approve the PR without requiring this extra step.

jujoramos:
Thanks @davebarnes97, will remove the `class="no-quick-link"` and try it out locally before pushing the changes again!.

geode-docs/developing/query_select/aggregates.html.md.erb
```diff
@@ -0,0 +1,340 @@
...
+SELECT pf.status, MIN(pf.ID), MAX(pf.ID), COUNT(pf.ID), AVG(pf.ID), SUM(pf.ID)
+FROM /portfolio pf
+GROUP BY pf.status
+```
+
+## <a id="min"></a>MIN
+
+The `MIN` keyword returns the minimum or smallest value from the selected expression. The expression itself must always evaluate to `java.lang.Comparable`. The `MIN` statement returns the actual type of the selected element as its result. 
+
+The following are example `MIN` queries that return region entries (the entries implement the `java.lang.Comparable` interface).
```
jhuynh1: Personally, I think this example should go last of the MIN examples.  I think the one with min(pf.ID) is easiest to understand then then after seeing that, this one makes more sense?.  Just personal preference I think

geode-docs/developing/query_select/aggregates.html.md.erb
```diff
@@ -0,0 +1,340 @@
...
+SELECT pf.status, MAX(pf.ID)
+FROM /exampleRegion pf
+WHERE pf.ID > 0
+GROUP BY pf.status
+```
+
+## <a id="count"></a>COUNT
+
+The `COUNT` keyword returns the number of results that match the query selection conditions specified in the `WHERE` clause. Using `COUNT` allows you to determine the size of a results set.
+The `COUNT` statement always returns a `java.lang.Integer` or `java.lang.Long` as the result (depending on how big the value is); you should take this into consideration when executing the query: if an overflow occurs while computing the `COUNT` function because the value is higher than `Long.MAX_VALUE` (2<sup>63</sup> - 1), the result will be incorrect.
```
jhuynh1: Is this statement about returning integer true still?

geode-docs/developing/query_select/aggregates.html.md.erb
jujoramos: Thanks for reviewing @jhuynh1.
I can change this if you have strong opinions about it, but all the aggregate functions start with examples that show how to return the region entry directly (it's not uncommon to try to find the minimum/maximum entry within a region). In order to keep the documentation about aggregate function consistent with the rest, I included these as the first examples for `MIN` and `MAX`.
Anyway, let me know if you want it changed.

geode-docs/developing/query_select/aggregates.html.md.erb
jujoramos: Yes, it' still true. We internally downcast the value before returning it to the user, meaning that they can get a ` java.lang.Integer` or a `java.lang.Long` as the result, depending on "how big" the actual value is.

geode-docs/developing/query_select/aggregates.html.md.erb
jhuynh1: I'm ok, I just thought finding a min entry is interesting in that, how is in determined but I guess it's the same for an ID

jujoramos: Sure, thanks for the feedback.
Long story short: it doesn't matter what we use as the argument for `MIN` / `MAX` **as long as it implements the `java.lang.Comparable` interface**, that's what we use to calculate the final result:

```

  @Override
  public void accumulate(Object value) {
    if (value == null || value == QueryService.UNDEFINED) {
      return;
    }

    Comparable comparable = (Comparable) value;

    if (currentOptima == null) {
      currentOptima = comparable;
    } else {
      @SuppressWarnings("unchecked")
      int compare = currentOptima.compareTo(comparable);

      if (findMax) {
        currentOptima = compare < 0 ? comparable : currentOptima;
      } else {
        currentOptima = compare > 0 ? comparable : currentOptima;
      }
    }
  }

```

That requirement is already part of the improved docs so I'll merge this PR :-).
If you think there's anything missing, please let me know, I'll be happy to open a new ticket and improve the docs further!.
Cheers.

PR #4002, opened 2019-08-31
GEODE-7134: Replace bucket Set with bucket int[]

mivanac:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/BucketSetHelper.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/BucketSetHelper.java
@@ -1,0 +1,64 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.apache.geode.internal.cache;
+
+
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+
+import org.apache.commons.lang3.ArrayUtils;
+
+public class BucketSetHelper {
+  public static int get(int[] bucketSet, int index) {
+    return bucketSet[index + 1];
+  }
+
+  public static int length(int[] bucketSet) {
+    return null == bucketSet || bucketSet.length < 2 ? 0 : bucketSet[0];
+  }
+
+  public static void add(int[] bucketSet, int value) {
+    int index = bucketSet[0] + 1;
+    bucketSet[index] = value;
+    bucketSet[0] = index;
+  }
+
+  public static Set<Integer> toSet(int[] bucketArray) {
+    Set<Integer> bucketSet;
+    if (BucketSetHelper.length(bucketArray) > 0) {
+      bucketSet =
+          new HashSet(
+              Arrays.asList(
+                  ArrayUtils.toObject(Arrays.copyOfRange(bucketArray, 1, bucketArray[0] + 1))));
+    } else {
+      bucketSet = new HashSet();
+    }
+    return bucketSet;
+  }
+
+  public static int[] fromSet(Set<Integer> bucketSet) {
+    int[] bucketArray = new int[bucketSet.size() + 1];
+    bucketArray[0] = bucketSet.size();
+
+    if (bucketSet.size() > 0) {
+      System.arraycopy(ArrayUtils.toPrimitive(bucketSet.toArray(new Integer[bucketSet.size()])), 0,
+          bucketArray,
+          1, bucketSet.size());
+    }
+    return bucketArray;
+  }
+
+}
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/BucketSetHelper.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/BucketSetHelper.java
@@ -15,11 +15,9 @@
 package org.apache.geode.internal.cache;
 
 
-import java.util.Arrays;
 import java.util.HashSet;
 import java.util.Set;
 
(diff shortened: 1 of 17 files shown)


pivotal-jbarrett:
Please update your PR commits with properly formatted and descriptive commit messages. See https://cwiki.apache.org/confluence/display/GEODE/Commit+Message+Format

mivanac:
Bucket Set is replaced with bucket int[], where index 0 contains length of array.

pivotal-jbarrett:
@mivanac I know this is draft but I think some input could guide. Perhaps so we can distinguish any arbitrary int[] from a bucket set maybe BucketSet should encapsulate this int[]? Is the rational for just int[] to save the extra allocation of the BucketSet object? If so maybe a set of static methods on a BucketSet helper class to encapsulate some of the logic, like length, iteration, set operations, such that no operations are done directly on the int[] but only through the helper. 
```
class BucketSet {
static int get(int[] bucketSet, int index) {
   return bucketSet[I+1];
}

static int length(int[] bucketSet) {
  return null == bucketSet || bucketSet.length < 2 ? 0 : bucketSet[0];
}

}

**Comment by mivanac** [comment]
Thanks for advice.

File: geode-core/src/main/java/org/apache/geode/internal/cache/BucketSetHelper.java [hunk 1]
```diff
@@ -0,0 +1,64 @@
...
+public class BucketSetHelper {
+  public static int get(int[] bucketSet, int index) {
+    return bucketSet[index + 1];
+  }
+
+  public static int length(int[] bucketSet) {
+    return null == bucketSet || bucketSet.length < 2 ? 0 : bucketSet[0];
+  }
+
+  public static void add(int[] bucketSet, int value) {
```
pivotal-jbarrett: What if the bucketSet doesn't have capacity? Should it resize and return a new array?

geode-core/src/main/java/org/apache/geode/internal/cache/BucketSetHelper.java
```diff
@@ -0,0 +1,64 @@
...
+    int index = bucketSet[0] + 1;
+    bucketSet[index] = value;
+    bucketSet[0] = index;
+  }
+
+  public static Set<Integer> toSet(int[] bucketArray) {
+    Set<Integer> bucketSet;
+    if (BucketSetHelper.length(bucketArray) > 0) {
+      bucketSet =
```
pivotal-jbarrett: How frequently is this being invoked? This is really expensive in time and allocations. I'd implement a more direct conversion, `new HashSet(length(bucketSet))`, then iterator over the `bucketSet` array and add each element to the `HashSet`.

What about just wrapping the current array in a Set implementation?

geode-core/src/main/java/org/apache/geode/internal/cache/BucketSetHelper.java
```diff
@@ -0,0 +1,64 @@
...
+          new HashSet(
+              Arrays.asList(
+                  ArrayUtils.toObject(Arrays.copyOfRange(bucketArray, 1, bucketArray[0] + 1))));
+    } else {
+      bucketSet = new HashSet();
```
pivotal-jbarrett: `Collections.emptySet()`?

geode-core/src/main/java/org/apache/geode/internal/cache/BucketSetHelper.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+    return bucketSet;
+  }
+
+  public static int[] fromSet(Set<Integer> bucketSet) {
+    int[] bucketArray = new int[bucketSet.size() + 1];
+    bucketArray[0] = bucketSet.size();
+
+    if (bucketSet.size() > 0) {
+      System.arraycopy(ArrayUtils.toPrimitive(bucketSet.toArray(new Integer[bucketSet.size()])), 0,
```
pivotal-jbarrett: Like above, this has extra allocations we don't need. I would iterate directly.

geode-core/src/main/java/org/apache/geode/internal/cache/BucketSetHelper.java
mivanac: Before add method is used, array is allocated, with enough capacity, so it should never happen that we don`t have capacity. If it happens, it is fault.

geode-core/src/main/java/org/apache/geode/internal/cache/BucketSetHelper.java
mivanac: Thanks for comments. Solution is updated.

pivotal-jbarrett:
@mivanac It doesn't look like these test failures are your fault. Try merging the latest develop into this branch and pushing again.

pivotal-jbarrett:
@mivanac It is your PR after all. Your local build should be failing too. The `Version` class and some of the internal serialization stuff have moved.

geode-core/src/main/java/org/apache/geode/internal/cache/BucketSetHelper.java
```diff
@@ -0,0 +1,67 @@
...
+  public static Set<Integer> toSet(int[] bucketSet) {
+    Set<Integer> resultSet;
+    int arrayLength = length(bucketSet);
+    if (arrayLength > 0) {
+      resultSet = new HashSet(arrayLength);
+      for (int i = 1; i <= arrayLength; i++) {
+        resultSet.add(bucketSet[i]);
+      }
+    } else {
+      resultSet = Collections.emptySet();
```
pivotal-jbarrett: It looks like my recommendation to use `emptySet()` is causing some test failures. Looks like the consumer of this method expects to modify the set returned.
```
...
     final int arrayLength = length(bucketSet);
     final Set<Integer> resultSet = new HashSet(arrayLength);
     for (int i = 1; i <= arrayLength; i++) {
       resultSet.add(bucketSet[i]);
     }
     return resultSet;
...
```

geode-core/src/main/java/org/apache/geode/internal/cache/BucketSetHelper.java
```diff
@@ -0,0 +1,66 @@
...
+ * the License.
+ */
+package org.apache.geode.internal.cache;
+
+
+import java.util.HashSet;
+import java.util.Set;
+
+
+public class BucketSetHelper {
```
dschneider-pivotal: I think you should add a unit test for this new class. I know it is just a bunch of static methods, but you can have a unit test for them.

pivotal-jbarrett:
@mivanac I don't think those test failures are yours but you should run them locally and if they pass write up some flaky test JIRAs for them if they don't already have them.

mivanac:
Thanks for all the comments.

mivanac:
Could you merge this PR, if there is nothing else to add.

PR #4007, opened 2019-09-04
GEODE-7160: Remove AttributesFactory in geode-wan

jujoramos:
- Removed references to the deprecated AttributesFactory from the
  geode-wan module.
- Reduced default configured off-heap size in WanTestBase to 128m.

### For all changes:

### Note:

### DIFF
--- a/geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/CacheClientNotifierDUnitTest.java
+++ b/geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/CacheClientNotifierDUnitTest.java
@@ -38,6 +38,7 @@ import org.apache.geode.cache.Region;
 import org.apache.geode.cache.RegionAttributes;
 import org.apache.geode.cache.RegionFactory;
 import org.apache.geode.cache.RegionShortcut;
+import org.apache.geode.cache.Scope;
 import org.apache.geode.cache.client.Pool;
 import org.apache.geode.cache.client.PoolManager;
 import org.apache.geode.cache.server.CacheServer;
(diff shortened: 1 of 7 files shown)


jujoramos:
The `OutOfMemoryError` doesn't have anything to do with the changes introduced through this `PR`, it's caused by the fact that all classes inheriting from `WANTestBase` have `off-heap` [enabled by default](https://github.com/apache/geode/blob/develop/geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/WANTestBase.java#L3851-L3861):

```
  @Override
  public Properties getDistributedSystemProperties() {
    // For now all WANTestBase tests allocate off-heap memory even though
    // many of them never use it.
    // The problem is that WANTestBase has static methods that create instances
    // of WANTestBase (instead of instances of the subclass). So we can't override
    // this method so that only the off-heap subclasses allocate off heap memory.
    Properties props = new Properties();
    props.setProperty(OFF_HEAP_MEMORY_SIZE, "300m");
    return props;
  }
```

Since this `PR` modifies some tests in `geode-wan`, the `StressNewTest` job is kicking in and executing all those tests multiple times, thus causing the OOME.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/CacheClientNotifierDUnitTest.java
```diff
@@ -265,11 +265,9 @@ public static void createClientWithLocator(int port0, String host, String region
       CacheServerTestUtil.enableShufflingOfEndpoints();
     }
 
-    AttributesFactory factory = new AttributesFactory();
+    RegionFactory factory = cache.createRegionFactory(RegionShortcut.LOCAL);
```
boglesby: I'm not sure if this matters, but the regions created before and after your changes here are not the same.

The new code creates this region:
```
[vm4] RMI TCP Connection(1)-127.0.0.1: CacheClientNotifierDUnitTest.createClientWithLocator region=org.apache.geode.internal.cache.LocalRegion[path='/testNormalClient2MultipleCacheServer_PR';scope=LOCAL';dataPolicy=NORMAL; concurrencyChecksEnabled]
```
The old code creates this region:
```
[vm3] RMI TCP Connection(1)-127.0.0.1: CacheClientNotifierDUnitTest.createClientWithLocator region=org.apache.geode.internal.cache.DistributedRegion[path='/testNormalClient2MultipleCacheServer_PR';scope=DISTRIBUTED_NO_ACK';dataPolicy=NORMAL; concurrencyChecksEnabled]
```

jujoramos: Good catch @boglesby!. I don't think it matters for the test but the goal of this PR is to get rid of `AttributesFactory`, not to modify the actual tests, so I've changed the region creation to reflect what it was doing before. 
I'll merge the changes once CI is green, thanks again for reviewing!.

PR #4011, opened 2019-09-04
GEODE-7075: 'class' field should not be part of json results

onichols-pivotal:
ClientClusterManagementService now adds a header to request the class attributes, otherwise if the header is not present they are stripped out.


geode-web-management/src/main/java/org/apache/geode/management/internal/rest/controllers/ManagementControllerAdvice.java
```diff
@@ -33,9 +45,47 @@
...
+          "application/json; charset=" + Charset.defaultCharset().toString());
+      response.getBody().write(json.getBytes());
+      response.close();
+      return null;
+    } catch (Exception e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  private static String removeClassFromJsonText(String json) {
```
dschneider-pivotal: Add some unit tests for removeClassFromJsonText to make sure it does what is expected.

geode-web-management/src/main/java/org/apache/geode/management/internal/rest/controllers/ManagementControllerAdvice.java
```diff
@@ -33,9 +45,47 @@
...
-public class ManagementControllerAdvice {
+public class ManagementControllerAdvice implements ResponseBodyAdvice<Object> {
   private static final Logger logger = LogService.getLogger();
 
+  @Autowired
+  private Jackson2ObjectMapperFactoryBean objectMapperFactory;
+
+  @Override
+  public boolean supports(MethodParameter returnType, Class converterType) {
+    return AbstractJackson2HttpMessageConverter.class.isAssignableFrom(converterType);
```
dschneider-pivotal: I don't understand why we only return true from supports if the converterType could be assigned to AbstractJackson2HttpMessageConverter. Is it worth a comment?

geode-management/src/main/java/org/apache/geode/management/internal/ClientClusterManagementService.java
```diff
@@ -74,10 +76,17 @@
...
     return assertSuccessful(restTemplate
-        .postForEntity(endPoint, config, ClusterManagementRealizationResult.class)
+        .exchange(endPoint, HttpMethod.POST, makeEntity(config),
+            ClusterManagementRealizationResult.class)
         .getBody());
   }
 
+  static <T> HttpEntity<T> makeEntity(T config) {
+    HttpHeaders headers = new HttpHeaders();
+    headers.add("X-Include-Class", "true");
```
dschneider-pivotal: add a symbolic constant for "X-Include-Class"

geode-web-management/src/main/java/org/apache/geode/management/internal/rest/controllers/ManagementControllerAdvice.java
jinmeiliao: I believe this is because the class information is only added by the "AbstractJackson2HttpMessageConverter".

onichols-pivotal: added a comment

geode-web-management/src/main/java/org/apache/geode/management/internal/rest/controllers/ManagementControllerAdvice.java
onichols-pivotal: added

geode-management/src/main/java/org/apache/geode/management/internal/ClientClusterManagementService.java
onichols-pivotal: added

PR #4012, opened 2019-09-04
GEODE-7164: Fix issue with IntelliJ Gradle import

aaronlindsey:
When delegating build/run actions to IntelliJ instead of Gradle, IntelliJ 2019 fails to build geode with an error like "output path not specified for modules...". I was able to reproduce this issue on IntelliJ IDEA CE versions 2019.1.4 and 2019.2.1 (see JIRA ticket for details). The issue seems to occur when IntelliJ imports Gradle sub-projects as modules that have "Use module compile output path" checked but don't have an Output path specified.

The best solution I've found involves 2 parts. I tested this on both versions of IntelliJ 2019.
1. Set the Project compiler output as shown in this SO answer: <https://stackoverflow.com/a/40675201>.
2. Add `inheritOutputDirs = true` to the idea plugin configuration in `ide.gradle`. This causes IntelliJ to import gradle sub-projects with the "Inherit project compile output path" option checked by default. You may need to reimport gradle projects after making this change if you do not have auto-import enabled.

This PR addresses part 2 above. Part 1 has to be addressed manually as far as I'm aware. 

---

### For all changes:

### Note:

### DIFF
--- a/gradle/ide.gradle
+++ b/gradle/ide.gradle
@@ -70,6 +70,7 @@ tasks.eclipse.dependsOn(cleanEclipse)
 
 idea {
   module {
+    inheritOutputDirs = true
     downloadSources = true
   }
 }


jdeppe-pivotal:
This is awesome!!

pivotal-jbarrett:
I am surprised this actually does anything. Are you switching between the Gradle builder and the IntelliJ builder? These parameters only effect the IntelliJ builder, which is all sorts of broken. Though to be fair, so is the Gradle builder but less so.

aaronlindsey:
@pivotal-jbarrett yes, this only effects the IntelliJ builder. I'm wanting to use IntelliJ's builder because it's much faster than the Gradle builder when making small changes, e.g. for TDD. So far, this is the only issue I've seen that prevents me from using the IntelliJ builder. This works because the Gradle IDEA plugin affects how the IntelliJ gradle import happens — from the IDEA plugin site:
> If you simply want to load a Gradle project into IntelliJ IDEA, then use the IDE’s import facility. You do not need to apply this plugin to import your project into IDEA, although if you do, the import will take account of any extra IDEA configuration you have that doesn’t directly modify the generated files — see the Configuration section for more details.

pivotal-jbarrett:
By default IJ invokes `cleanTest` when running a test. I don't understand why but after it creates te run configuration you can edit it to remove the clean targets and things go much faster.

aaronlindsey:
> By default IJ invokes `cleanTest` when running a test. I don't understand why but after it creates te run configuration you can edit it to remove the clean targets and things go much faster.

I usually end up with many run configurations because I may run tests one-at-a-time or a whole class at a time. It seems cumbersome to have to edit the run configuration each time. Is there a way to change the default Gradle test run configuration?

pivotal-jbarrett:
> I usually end up with many run configurations because I may run tests one-at-a-time or a whole class at a time. It seems cumbersome to have to edit the run configuration each time. Is there a way to change the default Gradle test run configuration?

I haven't found a way to set the default. :(

PR #4014, opened 2019-09-05
GEODE-6973: Use cachelistener to synchronize typeToId with IdToType

gesterzhou:
Co-authored-by: Xiaojian Zhou <gzhou@pivotal.io>

### For all changes:

### Note:


geode-core/src/main/java/org/apache/geode/pdx/internal/PeerTypeRegistration.java
```diff
@@ -365,20 +379,32 @@ public int defineType(PdxType newType) {
...
       newType.setTypeId(id);
 
       updateIdToTypeRegion(newType);
-
-      typeToId.put(newType, id);
-
       return newType.getTypeId();
     } finally {
+      // flush the tmpTypeToId map for who introduced this new pdxType
+      if (!tmpTypeToId.isEmpty()) {
```
jhuynh1: Why is this finally needed?  How would the tmpTypeToId not already be empty here?

gesterzhou: We have discussed that, it's safe to do it when the creator holds the d-lock. 
The listener will only add into tmpTypeToId map now. We use it to make sure that the distribution should all finished before flushing anything into typeToId map (they stayed in tmpTypeToId map now)

jhuynh1: But wouldn't the line above (line 386) when we actually use the map, have cleared it?  Why do we need to reclear in the finally?  Why is the duplicated code needed in the finally block?  Same goes for the other method...

gesterzhou:
I changed the map's name based on bruce's suggestion.

gesterzhou:
To bruce: the whole idea is to use a listener to maintain the synchronization between local TypeToId map and the IdToType region, then we don't have to search the region any more to speed up creating a new pdxType. 

All the code changes are focused on this and based on current code structure. We do notice that some codes can be enhanced. And a refactoring of TypeRegistry can even improve the performance. 

However, I would rather to log another geode ticket for refactoring work. 

We showed the pull request to the whole commons team in a presentation, new developers all understood.

upthewaterspout:
@bschuchardt - To make your suggestion more concrete - would it help to move pendingTypeToId and typeToId to a separate data structure that encapsulates that functionality?

IMO this whole pendingTypeToId business is essentially working around a limitation in our transactions. I would expect that if a CacheListener fires in Member A as a result of a transaction, then a `get` in Member B *after* that CacheListener event should see the change. That is unfortunately not the case.

bschuchardt:
@upthewaterspout yes, that would help.  It's like we're writing to a structure but don't want to commit the changes in the current thread.  The changes need to be committed in another thread.  Encapsulating the collections in another object that requires that the correct locks are obtained before transferring from pending to committed would go a long way toward making these changes more understandable without hearing Gester's presentation on the solution.

DonalEvans:
@upthewaterspout 

I removed those tests. You're right, they were potentially flaky and not testing any behaviour that wasn't accounted for elsewhere. As for where the sample JSON came from, it was just an edited version of an existing JSON resource used in other tests.

gesterzhou:
The testJSON.txt is from ./geode-core/src/distributedTest/resources/org/apache/geode/pdx/jsonStrings/json4.txt. 

We can change it to dummy url and email.

gesterzhou:
I will create an inner class to wrap the typeToId, pendingTypeToId, pendingEnumToId.

gesterzhou:
I created an inner class to wrap all these local maps. Pls review again.

PR #4019, opened 2019-09-06
Feature/geode 7168

bschuchardt:
This modifies geode-old-versions to use the dot-notation product version when building the classpath and install property files.  VersionManager.java and tests are altered to use dot-notation, too.  This lets TomcatInstall forego looking up Version instances in order to get the dot-notation product version string when it builds a classpath to launch a server.

### For all changes:

### Note:

### DIFF
--- a/geode-junit/src/main/java/org/apache/geode/test/version/VersionManager.java
+++ b/geode-junit/src/main/java/org/apache/geode/test/version/VersionManager.java
@@ -129,9 +129,16 @@ public class VersionManager {
    */
   public String versionWithNoDots(String s) {
     StringBuilder b = new StringBuilder(10);
-    for (int i = 0; i < s.length(); i++) {
-      if (s.charAt(i) != '.')
+    int length = s.length();
+    for (int i = 0; i < length; i++) {
+      char ch = s.charAt(i);
+      if (ch != '.') {
+        // leave off any trailing stuff like "-incubating"
+        if (!Character.isDigit(ch)) {
+          break;
+        }
         b.append(s.charAt(i));
+      }
     }
     return b.toString();
   }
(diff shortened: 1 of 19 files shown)


bschuchardt:
Stress testing jobs keep dieing with the Gradle daemon disappearing.  I'm running the stress tests manually on a Linux machine to see what's up with that.  The test configuration is logged in the failed StressNewTestOpenJDK11 jobs.

bschuchardt:
Some of the tests are using hard-coded port numbers and aren't going to pass stress-testing without being reworked.  Given how stress-testing runs everything in parallel I can use its output to figure out what tests are having this problem.  My manually run tests hung after reporting bind exceptions and I killed them after 3 hours.

geode-junit/src/main/java/org/apache/geode/test/version/VersionManager.java
```diff
@@ -180,12 +195,15 @@ private void findInstalls(String fileName) {
...
     String parsedVersion = null;
-    if (version.startsWith("test") && version.length() >= "test".length()) {
-      if (version.equals("test")) {
-        parsedVersion = CURRENT_VERSION;
-      } else {
-        parsedVersion = version.substring("test".length());
-      }
+    if (version.length() > 0 && Character.isDigit(version.charAt(0))
+        && version.length() >= "1.2.3".length()) {
+      // for (int i = 1; i < version.length(); i++) {
```
mcmellawatt: Some new commented out code here we probably want to remove before merging (assuming its not needed anymore)

geode-junit/src/main/java/org/apache/geode/test/version/VersionManager.java
```diff
@@ -41,11 +39,9 @@
...
  */
 public class VersionManager {
-  public static final String CURRENT_VERSION = "000";
-  public static final String GEODE_110 = "110";
-  public static final String GEODE_120 = "120";
-  public static final String GEODE_130 = "130";
-  public static final String GEODE_140 = "140";
+  public static final String CURRENT_VERSION = "0.0.0";
+  public static final String GEODE_130 = "1.3.0";
+  public static final String GEODE_140 = "1.4.0";
```
mcmellawatt: I feel like I asked this before so I apologize if its a repeat...why could we get rid of versions 1.1 and 1.2 here?

geode-junit/src/main/java/org/apache/geode/test/version/VersionManager.java
```diff
@@ -126,6 +122,25 @@ public String getInstall(String version) {
...
+    StringBuilder b = new StringBuilder(10);
+    int length = s.length();
+    for (int i = 0; i < length; i++) {
+      char ch = s.charAt(i);
+      if (ch != '.') {
+        // leave off any trailing stuff like "-incubating"
+        if (!Character.isDigit(ch)) {
+          break;
+        }
+        b.append(s.charAt(i));
```
mcmellawatt: Can you just use the `ch` variable here instead of calling `s.charAt(i)` again?

geode-junit/src/main/java/org/apache/geode/test/version/VersionManager.java
bschuchardt: there were no references to them.  We should probably get rid of GEODE_130 and GEODE_140 as well since the common practice seems to be to hard-code version strings in our tests.

geode-junit/src/main/java/org/apache/geode/test/version/VersionManager.java
bschuchardt: okay - I'll change that

geode-junit/src/main/java/org/apache/geode/test/version/VersionManager.java
bschuchardt: okay

PR #4020, opened 2019-09-06
GEODE-7170: Move InetAddressUtils to internal.net package

kirklund:
* Rename class from InetAddressUtil to InetAddressUtils
* Move InetAddressUtils from admin.internal to internal.net


geode-core/src/main/java/org/apache/geode/admin/internal/InetAddressUtil.java
```diff
@@ -1,200 +0,0 @@
...
-   * Logs the stack trace for the given Throwable if logger is initialized else prints the stack
-   * trace using System.out. If logged the logs are logged at WARNING level.
-   *
-   * @param throwable Throwable to log stack trace for
-   */
-  private static void logStackTrace(Throwable throwable) {
-    AdminDistributedSystemImpl adminDS = AdminDistributedSystemImpl.getConnectedInstance();
-
-    logger.warn(throwable.getMessage(), throwable);
-  }
```
aaronlindsey: It looks like the old version of this class used to log warnings under certain error conditions but the new version of the class does not. Are we sure nobody is relying on these warnings being logged?

kirklund: No one is using Admin API. It was deprecated before Apache Geode 1.0.0. I removed the logging of stack traces mainly because it's bad form for simple low-level utils like this to perform logging. The main reason to move this class out of admin.internal is because we're going to delete everything under admin.* but we have lots of code outside of admin.* using this class. I could add the logging back in if you feel strongly about it, but I don't think it's valuable.

If we change it further, then my preference would be to propagate the caught exception instead of logging it:
```
} catch (UnknownHostException e) {
  throw new AssertionError("Failed to get InetAddress: " + host**, e**);
}
```

aaronlindsey: My concern was that we might be removing logging that users rely on. If we're sure that this logging isn't valuable/useful then I'm fine with removing it.

kirklund: I added InetAddressUtilsWithLogging which wraps two of the utils with versions that log call stack. I also updated all uses of both InetAddressUtils classes to use `import static`

PR #4024, opened 2019-09-09
GEODE-5940: Disable server port in many ServerLauncher tests

kirklund:
GEODE-5940: Disable server port in many ServerLauncher tests
    
Disable default server in all launcher tests that do not require it.
    
This will decrease the number of failures to launch the server caused
by an underlying BindException because a port is already in use.

GEODE-5940: Use destroyForcibly in ServerLauncherRemoteIntegrationTest tearDown


kirklund:
The IntegrationTestOpenJDK11 failure is JGroupsMessengerJUnitTest which is failing identically against develop in CI.

kirklund:
GREEN precheckin results

PR #4047, opened 2019-09-11
GEODE-7194: simplify CMS 'get' return type

onichols-pivotal:
before, you had to use ClusterManagementService "get" method like this:
`Index index = cms.get(index).getConfigResult().get(0);`

Now, it's slightly simpler:
`Index index = cms.get(index).getConfigResult();`

Note that this is a "breaking" API change for both REST and Java.  Since everything is still marked @experimental, breaking is ok, right?


onichols-pivotal:
Internally we still have "get" methods on some controllers (get members, get indexes) that are implemented using `list`.  Maybe these should also be rewritten to use `get` -- the only downside would be a more generic error message when not found, especially in the case of index (the current custom error message includes the region name in addition to the index name)

PR #4058, opened 2019-09-13
GEODE-7208: FlatFormatSerializer should index on inherited fields

gesterzhou:

### For all changes:

### Note:

### DIFF
--- a/geode-lucene/src/integrationTest/java/org/apache/geode/cache/lucene/NestedObjectSeralizerIntegrationTest.java
+++ b/geode-lucene/src/integrationTest/java/org/apache/geode/cache/lucene/NestedObjectSeralizerIntegrationTest.java
@@ -20,6 +20,7 @@ import static org.junit.Assert.assertEquals;
 
 import java.io.Serializable;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.concurrent.TimeUnit;
@@ -44,12 +45,22 @@ import org.apache.geode.test.junit.categories.LuceneTest;
 @Category({LuceneTest.class})
 public class NestedObjectSeralizerIntegrationTest extends LuceneIntegrationTest {
 
-  private static int WAIT_FOR_FLUSH_TIME = 10000;
-  private static final Logger logger = LogService.getLogger();
+  protected static int WAIT_FOR_FLUSH_TIME = 10000;
+  protected static final Logger logger = LogService.getLogger();
   LuceneQuery<Integer, Object> query;
   PageableLuceneQueryResults<Integer, Object> results;
 
-  private Region createRegionAndIndex() {
+  protected CustomerFactory getCustomerFactory() {
+    return Customer::new;
+  }
+
+  protected static interface CustomerFactory {
+    public Customer create(String name, Collection<String> phoneNumbers,
+        Collection<Person> contacts,
+        Page[] myHomePages);
+  }
+
+  protected Region createRegionAndIndex() {
     luceneService.createIndexFactory().setLuceneSerializer(new FlatFormatSerializer())
         .addField("name").addField("phoneNumbers").addField("myHomePages.content")
         .addField("contacts.name").addField("contacts.email", new KeywordAnalyzer())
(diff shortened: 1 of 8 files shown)


geode-lucene/src/main/java/org/apache/geode/cache/lucene/FlatFormatSerializer.java
```diff
@@ -140,13 +140,17 @@ private Object getFieldValue(Object value, String fieldName) {
...
-        return null;
-      }
+
+      do {
+        try {
+          Field field = clazz.getDeclaredField(fieldName);
+          field.setAccessible(true);
+          return field.get(value);
+        } catch (Exception e) {
+          clazz = clazz.getSuperclass();
```
upthewaterspout: Seems like it would be better to specifically check to see if the field is missing, rather than catch a blanket "Exception" here.

geode-lucene/src/integrationTest/java/org/apache/geode/cache/lucene/ToplevelObjectSerializerOnGrandSubclassIntegrationTest.java
```diff
@@ -0,0 +1,67 @@
...
+  public void queryOnTopLevelObjectCollectionField_AND_BothExist() {}
+
+  @Test
+  public void queryOnContactNameWithExactMath() {}
+
+  @Test
+  public void queryOnTopLevelObjectCollectionField_OR_OneExist() {}
+
+  @Test
+  public void queryOnTopLevelStringCollectionField_AND_BothExist() {}
```
upthewaterspout: The fact that is test is overriding so many methods with empty implementations looks like it means maybe this shoudl not extend NestedObjectSerializerOnGrandSubclassIntegrationTest. Maybe you need a different parent class for both tests?

geode-lucene/src/integrationTest/java/org/apache/geode/cache/lucene/NestedObjectSeralizerIntegrationTest.java
```diff
@@ -70,7 +77,16 @@ private Region createRegionAndIndexOnInvalidFields() {
     return region;
   }
 
-  private void feedSomeNestedObjects(Region region) throws InterruptedException {
+  protected void feedSomeNestedObjects(Region region) throws InterruptedException {
+    Class clazz = useWhichClass();
+    Constructor<Customer> constructor = null;
+    try {
+      constructor = clazz.getConstructor(String.class, Collection.class, Collection.class,
```
upthewaterspout: Rather than using a Class and finding the constructor through reflection, why not just use a Supplier which returns an instance of the class?

geode-lucene/geode-lucene-test/src/main/java/org/apache/geode/cache/lucene/test/SpecialCustomer.java
```diff
@@ -0,0 +1,26 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.apache.geode.cache.lucene.test;
+
+import java.util.Collection;
+
+public class SpecialCustomer extends LocalCustomer {
```
upthewaterspout: I find the names "LocalCustomer" and "SpecialCustomer" to not be very descriptive. Maybe InheritedCustomer or ChildCustomer or something like that which describes what you are trying to do with your test?

gesterzhou: I was using ChildCustomer before, then GrandChildCustomer. But I think LocalCustomer and SpecialCustomer makes more sense in naming. How do you think?

geode-lucene/src/main/java/org/apache/geode/cache/lucene/FlatFormatSerializer.java
gesterzhou: it's old behavior. I kept it.

PR #4061, opened 2019-09-16
Geode 6874: add unit tests to tomcat modules

BenjaminPerryRoss:

### For all changes:

### Note:


extensions/geode-modules-test/src/main/java/org/apache/geode/modules/session/catalina/DeltaSessionManagerJUnitTest.java
```diff
@@ -0,0 +1,471 @@
...
+import org.apache.geode.cache.Region;
+import org.apache.geode.cache.query.FunctionDomainException;
+import org.apache.geode.cache.query.NameResolutionException;
+import org.apache.geode.cache.query.Query;
+import org.apache.geode.cache.query.QueryInvocationTargetException;
+import org.apache.geode.cache.query.SelectResults;
+import org.apache.geode.cache.query.TypeMismatchException;
+import org.apache.geode.cache.query.internal.InternalQueryService;
+import org.apache.geode.cache.query.internal.LinkedResultSet;
+import org.apache.geode.internal.cache.GemFireCacheImpl;
```
jujoramos: Can we remove reference to internal classes and use the public interfaces (`QueryService`, `Cache` and `SelectResults`) instead?

extensions/geode-modules-test/src/main/java/org/apache/geode/modules/session/catalina/PeerToPeerSessionCacheJUnitTest.java
```diff
@@ -0,0 +1,205 @@
...
+import org.apache.geode.modules.util.TouchReplicatedRegionEntriesFunction;
+
+public class PeerToPeerSessionCacheJUnitTest extends AbstractSessionCacheJUnitTest {
+
+  private String localRegionName = sessionRegionName + "_local";
+  @SuppressWarnings("unchecked")
+  private RegionFactory<String, HttpSession> regionFactory = mock(RegionFactory.class);
+  @SuppressWarnings("unchecked")
+  private Region<String, HttpSession> localRegion = mock(Region.class);
+  private Cache cache = mock(GemFireCacheImpl.class);
```
jujoramos: Same as before, we should mock the `Cache` interface instead of the internal `GemFireCacheImpl` class.

extensions/geode-modules-test/src/main/java/org/apache/geode/modules/session/catalina/internal/DeltaSessionAttributeEventBatchJUnitTest.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+import org.junit.Test;
+
+import org.apache.geode.LogWriter;
+import org.apache.geode.cache.Cache;
+import org.apache.geode.cache.Region;
+import org.apache.geode.internal.cache.GemFireCacheImpl;
+import org.apache.geode.internal.logging.ManagerLogWriter;
```
jujoramos: Can we change this to `Cache` and `LogWriter` instead?.

extensions/geode-modules-test/src/main/java/org/apache/geode/modules/util/TouchPartitionedRegionEntriesFunctionJUnitTest.java
```diff
@@ -0,0 +1,84 @@
...
+import org.junit.Before;
+import org.junit.Test;
+
+import org.apache.geode.cache.Cache;
+import org.apache.geode.cache.Region;
+import org.apache.geode.cache.execute.FunctionContext;
+import org.apache.geode.cache.execute.RegionFunctionContext;
+import org.apache.geode.cache.execute.ResultSender;
+import org.apache.geode.internal.cache.GemFireCacheImpl;
+import org.apache.geode.internal.logging.LocalLogWriter;
```
jujoramos: Same as before, can we mock the public interfaces instead of the internal implementations?.

extensions/geode-modules-test/src/main/java/org/apache/geode/modules/util/TouchReplicatedRegionEntriesFunctionJUnitTest.java
```diff
@@ -0,0 +1,77 @@
...
+import org.junit.Before;
+import org.junit.Test;
+
+import org.apache.geode.cache.Cache;
+import org.apache.geode.cache.Region;
+import org.apache.geode.cache.execute.FunctionContext;
+import org.apache.geode.cache.execute.RegionFunctionContext;
+import org.apache.geode.cache.execute.ResultSender;
+import org.apache.geode.internal.cache.GemFireCacheImpl;
+import org.apache.geode.internal.logging.LocalLogWriter;
```
jujoramos: Can we mock the public interfaces instead of the internal implementations?.

extensions/geode-modules/src/main/java/org/apache/geode/modules/session/catalina/DeltaSessionManager.java
```diff
@@ -932,16 +945,57 @@ private void doLoad() throws ClassNotFoundException, IOException {
...
+
+  void writeToObjectOutputStream(ObjectOutputStream oos, List listToWrite) throws IOException {
+    oos.writeObject(listToWrite.size());
+  }
+
+  int getSessionCountFromObjectInputStream(ObjectInputStream ois)
+      throws IOException, ClassNotFoundException {
+    return (Integer) ois.readObject();
+  }
+
```
jujoramos: Do we really need all these new methods to create, write, and other operations on low level java classes just for mocking purposes?. Looks "a bit overkilling" to me...

extensions/geode-modules-test/src/main/java/org/apache/geode/modules/session/catalina/DeltaSessionManagerJUnitTest.java
BenjaminPerryRoss: In other tests this was possible and I've replaced internal uses there with the public interfaces. For this test specifically, the 3 internal references are necessary. The product itself performs a cast of the retrieved Cache object to GemFireCacheImpl, and as a result passing a mock of Cache.class throws an exception since a mock of Cache.class can't be cast to an instance of GemFireCacheImpl. This also ties us to the use of InternalQueryService as the getQueryService() implementation of GemFireCacheImpl returns an InternalQueryService specifically, so in order to mock that call we have to use the internal object. The SelectResults on the other hand aren't being used for mocking, we're actually creating a real object to return. It seemed appropriate to use an actual implementation of the interface for this rather than creating one specifically for this test.

extensions/geode-modules/src/main/java/org/apache/geode/modules/session/catalina/DeltaSessionManager.java
BenjaminPerryRoss: The problem is all these input streams are being created, used, and discarded internally in the load/unload process. This means that if we want to verify any behavior of those objects (in addition to avoiding the NullPointerExceptions that come with trying to actually perform these filesystem operations in this unit test environment) we have to mock them, and as far as I could tell the best way to do that here was to create these new methods so that the spy could use them to return mocks instead of freshly constructed objects. The writeObject method is also necessary because the call it's making internally is final and thus can't be mocked. This method throws exceptions if run on a mocked output stream so we need to call it in a helper method to prevent execution during testing.

mhansonp:
@jhuynh1 @upthewaterspout Can you review this and merge it?

PR #4067, opened 2019-09-18
GEODE-6871: Disk free space info exposed via JMX

alb3rtobr:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/integrationTest/java/org/apache/geode/internal/cache/DiskStoreImplIntegrationTest.java
+++ b/geode-core/src/integrationTest/java/org/apache/geode/internal/cache/DiskStoreImplIntegrationTest.java
@@ -105,9 +105,10 @@ public class DiskStoreImplIntegrationTest {
   }
 
   @Test
-  public void givenDiskStoreIsCreatedWithMaxSize_thenGetDiskUsagePercentageShouldBeZero()
+  public void givenDiskStoreIsCreatedWithMaxSize_thenDiskPercentagesShouldBeZeroAnd100()
       throws Exception {
 
+    final int ALLOWED_MARGIN = 1;
     File[] diskDirs = new File[2];
     diskDirs[0] = temporaryDirectory.newFolder("dir1");
     diskDirs[1] = temporaryDirectory.newFolder("dir2");
@@ -121,13 +122,16 @@ public class DiskStoreImplIntegrationTest {
     DiskStore diskStore = cache.findDiskStore(DISK_STORE_NAME);
 
     assertEquals(0, ((DiskStoreImpl) diskStore).getDiskUsagePercentage(), 0);
+    assertEquals(100, ((DiskStoreImpl) diskStore).getDiskFreePercentage(),
+        ALLOWED_MARGIN);
   }
 
   @Test
-  public void givenDiskStoreIsCreatedWithMaxSize_whenDataIsStored_thenGetDiskUsagePercentageShouldBeIncreased()
+  public void givenDiskStoreIsCreatedWithMaxSize_whenDataIsStored_thenDiskPercentagesShouldBeModified()
       throws Exception {
 
     final int ESTIMATED_USAGE_PERCENTAGE = 7; // Estimated disk percentage for NUM_ENTRIES
+    final int ESTIMATED_FREE_PERCENTAGE = 100 - ESTIMATED_USAGE_PERCENTAGE;
     final int ALLOWED_MARGIN = 2;
 
     File[] diskDirs = new File[2];
@@ -150,10 +154,12 @@ public class DiskStoreImplIntegrationTest {
 
     assertEquals(ESTIMATED_USAGE_PERCENTAGE, ((DiskStoreImpl) diskStore).getDiskUsagePercentage(),
         ALLOWED_MARGIN);
+    assertEquals(ESTIMATED_FREE_PERCENTAGE, ((DiskStoreImpl) diskStore).getDiskFreePercentage(),
+        ALLOWED_MARGIN);
   }
 
   @Test
-  public void givenDiskStoreIsCreatedWithDefaultSize_thenGetDiskUsagePercentageShouldNotBeAvailable()
+  public void givenDiskStoreIsCreatedWithDefaultSize_thenDiskPercentagesShouldNotBeAvailable()
       throws Exception {
 
     File[] diskDirs = new File[2];
@@ -166,10 +172,12 @@ public class DiskStoreImplIntegrationTest {
 
     assertEquals(ManagementConstants.NOT_AVAILABLE_FLOAT,
         ((DiskStoreImpl) diskStore).getDiskUsagePercentage(), 0);
+    assertEquals(ManagementConstants.NOT_AVAILABLE_FLOAT,
+        ((DiskStoreImpl) diskStore).getDiskFreePercentage(), 0);
   }
 
   @Test
-  public void givenDiskStoreIsCreatedWithDefaultSize_whenDataIsStored_thenGetDiskUsagePercentageShouldNotBeAvailable()
+  public void givenDiskStoreIsCreatedWithDefaultSize_whenDataIsStored_thenDiskPercentagesShouldNotBeAvailable()
       throws Exception {
 
     File[] diskDirs = new File[2];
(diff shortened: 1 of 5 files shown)


geode-core/src/main/java/org/apache/geode/management/DiskStoreMXBean.java
```diff
@@ -196,6 +196,11 @@
    */
   float getDiskUsagePercentage();
 
+  /**
+   * Returns the free percentage of the disk storage
+   */
+  float getDiskFreePercentage();
```
jujoramos: This is a public interface, even though I rarely see people manually implementing it, the addition of a new method would break backward compatibility... would it make sense to add a `default` implementation?.

geode-core/src/integrationTest/java/org/apache/geode/internal/cache/DiskStoreImplIntegrationTest.java
```diff
@@ -105,9 +105,10 @@ public void queueSizeStatIncrementedAfterAsyncFlush() throws Exception {
   }
 
   @Test
-  public void givenDiskStoreIsCreatedWithMaxSize_thenGetDiskUsagePercentageShouldBeZero()
+  public void givenDiskStoreIsCreatedWithMaxSize_thenDiskPercentagesShouldBeZeroAnd100()
```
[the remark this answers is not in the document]
alb3rtobr: As this test uses given-when-then style, I find its easy to understand what the test is doing by separating the statements, but I understand its not the standard in all the tests, so I can change it.

geode-core/src/main/java/org/apache/geode/management/DiskStoreMXBean.java
alb3rtobr: This PR is based on GEODE-5222, and in that PR it was considered ok to modify the interface to add "getDiskUsagePercentage" without default implementation, I suppose it was due to its not usually implemented as you said.

jujoramos: I'm not sure why we didn't consider "worth it" adding a `default` implementation for `getDiskUsagePercentage` in [GEODE-5222](https://issues.apache.org/jira/browse/GEODE-5222), I'm tagging the original reviewers as they might have more context than me on this one.
Either way, unless there's a planned major version release in the near future, I just don't like the idea of breaking the backward compatibility in our public classes... even if they are rarely implemented by our users, we can't tell for sure.

@moleske @kirklund @aaronlindsey

dschneider-pivotal: @jujoramos Does the addition of a new method on an interface break old code that is already compiled against the old interface? When you say it "breaks backward compatibility" what do you mean? Is it that the old class files will no longer run with the new product jar? Or is it that old code will no longer compile? In the past we treated method additions as compatible, but removing methods and changing existing methods as incompatible.

geode-core/src/integrationTest/java/org/apache/geode/internal/cache/DiskStoreImplIntegrationTest.java
```diff
@@ -121,13 +122,16 @@ public void givenDiskStoreIsCreatedWithMaxSize_thenGetDiskUsagePercentageShouldB
     DiskStore diskStore = cache.findDiskStore(DISK_STORE_NAME);
 
     assertEquals(0, ((DiskStoreImpl) diskStore).getDiskUsagePercentage(), 0);
+    assertEquals(100, ((DiskStoreImpl) diskStore).getDiskFreePercentage(),
+        ALLOWED_MARGIN);
   }
 
   @Test
-  public void givenDiskStoreIsCreatedWithMaxSize_whenDataIsStored_thenGetDiskUsagePercentageShouldBeIncreased()
+  public void givenDiskStoreIsCreatedWithMaxSize_whenDataIsStored_thenDiskPercentagesShouldBeModified()
```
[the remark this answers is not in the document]
dschneider-pivotal: I think this mix of camel case and underscores was intentional to help highlight the three sections (given, when, then). @dhemery may have some helpful input on this.

geode-core/src/main/java/org/apache/geode/management/DiskStoreMXBean.java
```diff
@@ -196,6 +196,11 @@
    */
   float getDiskUsagePercentage();
 
+  /**
+   * Returns the free percentage of the disk storage
```
dschneider-pivotal: Add a "@since geode ..." tag.
Also, should you describe here when it will return -1?
I know you inherited "of the disk storage" from getDiskUsagePercentage but I'm not sure what that means. I think this mxbean represents just one disk store. If that is correct, then if these methods said "of this disk store" I think it would be clearer.

geode-core/src/main/java/org/apache/geode/management/DiskStoreMXBean.java
dschneider-pivotal: @jujoramos I like the idea of adding a default to interface methods we add to existing external interfaces.

jujoramos: @dschneider-pivotal: I'm not sure about what happens when code compiled against the old interface is executed using the new one, haven't tried yet.
That said, user code implementing the old version of the interface will certainly not compile if we don't provide a `default` implementation, that already sounds like an API change that is no backward compatible to me (please correct me if I'm wrong here).

alb3rtobr: I have added default implementation for both methods. I suppose the default implementation of `getDiskUsagePercentage` should be included in 1.10 before release, right?

dschneider-pivotal: Geode exposes much of the external apis on interfaces. Many of those interfaces are not meant to be implemented by geode users. Instead they are used to hide the internal geode implementation. For example Region is an interface that a geode user never needs to implement. So in the past as we added new features to geode adding a method to one of these interfaces was considered not breaking backward compatibility. But in the past we didn't even have the option of "default" methods on interfaces. The external geode api also has interfaces that only exist for geode users to implement. An example is CacheListener. It seems to me like a good rule going forward that if we add a method on an existing external interface that we use "default" on it to be safe. It can be unclear if an interface is meant to be implemented, and it is always on option and a geode user might implement Region for testing purposes or to have their own wrapper around the actual geode Region.

aaronlindsey: > I'm not sure why we didn't consider "worth it" adding a `default` implementation for `getDiskUsagePercentage` in [GEODE-5222](https://issues.apache.org/jira/browse/GEODE-5222), I'm tagging the original reviewers as they might have more context than me on this one.

I think it's a good idea to add a default implementation for both methods. At the time I was very new to this project so I didn't consider the possibility of a breaking change. I'm not sure whether this would be considered a "critical" fix to make it in 1.10. I suppose this might be a good question for the dev list.

geode-core/src/integrationTest/java/org/apache/geode/internal/cache/DiskStoreImplIntegrationTest.java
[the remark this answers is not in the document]
aaronlindsey: I think the underscores help make the test name more readable in this case (because of the given/when/then style and the long length of the name).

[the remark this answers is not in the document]
alb3rtobr: I wrote these tests, I used the underscores for distinguish the GIVEN-WHEN-THEN sections. After first @jujoramos I removed them on a new commit for this PR, although I think they make the test easy to understand. Should I remove them or keep them?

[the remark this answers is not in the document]
aaronlindsey: > Should I remove them or keep them?

As a general rule, I'd say adhere to the existing style. However, in Geode we don't seem to have a standard style for test names. I personally prefer the style with the underscores, but I'd be fine with it either way.

alb3rtobr:
> Seems alright to me, I appreciate the descriptive test names!

Its better to use descriptive names than comments explaining the test case :) Thanks!

PR #4071, opened 2019-09-19
GEODE-7165: Remove networking dependency from GatewayReceiverImplTest

kirklund:
The test was failing on machines with ports 2000 or 2001 in use.

This change replaces GatewayReceiverImpl's dependency on AvailablePort
with functions that are now injected by the unit tests.

Specifically, this test was failing for @dickcav and @onichols-pivotal.

### DIFF
--- a/geode-wan/src/test/java/org/apache/geode/internal/cache/wan/GatewayReceiverImplTest.java
+++ b/geode-wan/src/test/java/org/apache/geode/internal/cache/wan/GatewayReceiverImplTest.java
@@ -62,7 +62,7 @@ public class GatewayReceiverImplTest {
   @Test
   public void getHostOnUnstartedGatewayShouldReturnLocalHostName() throws UnknownHostException {
     GatewayReceiver gateway = new GatewayReceiverImpl(cache, 2000, 2001, 5, 100, null, null, null,
-        true, port -> true, portRange -> 2000);
+        true, true, 2000);
 
     assertThat(gateway.getHost()).isEqualTo(SocketCreator.getLocalHost().getHostName());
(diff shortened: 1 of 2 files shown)


geode-wan/src/test/java/org/apache/geode/internal/cache/wan/GatewayReceiverImplTest.java
```diff
@@ -62,7 +62,8 @@ public void setUp() {
   @Test
   public void getHostOnUnstartedGatewayShouldReturnLocalHostName() throws UnknownHostException {
     GatewayReceiver gateway =
-        new GatewayReceiverImpl(cache, 2000, 2001, 5, 100, null, null, null, true);
+        new GatewayReceiverImpl(cache, 2000, 2001, 5, 100, null, null, null, true, a -> true,
```
gesterzhou: Can you change the "a->" to a better name?

kirklund: `a -> true` has been renamed to `port -> true`

`a -> 2000` has been renamed to `portRange -> 2000`

kirklund: Since the test doesn't need or use the input arguments, I changed it a 2nd time to remove them altogether.

kirklund:
Latest changes:
* Add new constructor for use in test that accepts function results since the test has no use for the input arguments.
* Remove VisibleForTesting annotation from PortRange.

PR #4073, opened 2019-09-19
GEODE-7219 BufferUnderflowException in PutReplyMessage deserialization

bschuchardt:
VersionTag serialization was being affected by concurrent modification
of its memberId/previousMemberId fields, causing the HAS_PREVIOUS_MEMBER_ID
flag bit to be set and the DUPLICATE_MEMBER_IDS flag to _not_ be set.
It then went on to perform the same checks later in toData() and make
different decisions, winding up by not serializing the previousMemberId
field because it was then == to the memberId field.

1) decide on what the flags are going to be in toData and do not perform
the same calculations again.

2) use equals() when seeing if memberId and previousMemberId are equal.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/versions/VersionTag.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/versions/VersionTag.java
@@ -17,6 +17,8 @@ package org.apache.geode.internal.cache.versions;
 import java.io.DataInput;
 import java.io.DataOutput;
 import java.io.IOException;
+import java.nio.BufferUnderflowException;
+import java.util.Objects;
 import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 
 import org.apache.logging.log4j.Logger;
@@ -29,6 +31,7 @@ import org.apache.geode.internal.logging.log4j.LogMarker;
 import org.apache.geode.internal.serialization.DataSerializableFixedID;
 import org.apache.geode.internal.serialization.DeserializationContext;
 import org.apache.geode.internal.serialization.SerializationContext;
+import org.apache.geode.internal.serialization.Version;
 import org.apache.geode.internal.size.ReflectionSingleObjectSizer;
 
 /**
@@ -60,11 +63,11 @@ public abstract class VersionTag<T extends VersionSource>
 
 
   // flags for serialization
-  private static final int HAS_MEMBER_ID = 0x01;
-  private static final int HAS_PREVIOUS_MEMBER_ID = 0x02;
-  private static final int VERSION_TWO_BYTES = 0x04;
-  private static final int DUPLICATE_MEMBER_IDS = 0x08;
-  private static final int HAS_RVV_HIGH_BYTE = 0x10;
+  static final int HAS_MEMBER_ID = 0x01;
+  static final int HAS_PREVIOUS_MEMBER_ID = 0x02;
+  static final int VERSION_TWO_BYTES = 0x04;
+  static final int DUPLICATE_MEMBER_IDS = 0x08;
+  static final int HAS_RVV_HIGH_BYTE = 0x10;
 
   private static final int BITS_POSDUP = 0x01;
   private static final int BITS_RECORDED = 0x02; // has the rvv recorded this?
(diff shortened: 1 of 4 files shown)


geode-core/src/main/java/org/apache/geode/internal/cache/versions/VersionTag.java
```diff
@@ -350,10 +353,13 @@ public void toData(DataOutput out, boolean includeMember) throws IOException {
     if (this.memberID != null && includeMember) {
       flags |= HAS_MEMBER_ID;
     }
-    if (this.previousMemberID != null) {
+    boolean writePreviousMemberID = false;
```
Bill: I'm curious why you chose to introduce a new variable rather than referencing the `DUPLICATE_MEMBER_IDS` bit in `flags`?

geode-core/src/test/java/org/apache/geode/internal/cache/versions/AbstractVersionTagTestBase.java
```diff
@@ -31,6 +64,122 @@ public void setup() {
     this.vt = createVersionTag();
   }
 
+  int getRandomUnusedInt() {
+    int unusedInt;
+    do {
+      unusedInt = random.nextInt(60000);
+    } while (usedInts.contains(unusedInt));
```
Bill: this method has failed to add the new `unusedInt` value to `usedInts` before returning

geode-core/src/test/java/org/apache/geode/internal/cache/versions/AbstractVersionTagTestBase.java
```diff
@@ -31,6 +64,122 @@ public void setup() {
...
+          // previousMemberID again to see if it's the same as the memberID.
+          spy.setPreviousMemberID(spy.getMemberID());
+        }
+        return null;
+      }
+    };
+    doAnswer(myAnswer).when(dataOutput).writeShort(any(Integer.class));
+    spy.toData(dataOutput, true);
+    // verify that we only wrote the
+    verify(spy, times(2)).writeMember(isA(VersionSource.class), isA(DataOutput.class));
```
Bill: From the method name `testConcurrent…` I get the impression that this test is meant to verify that the bug (GEODE-7219) is fixed. I went to what is now line 385 of `VersionTag.java` and reverted what seems to be the primary fix, and then changed this spy to `times(1)` and this test passes (I expected it to fail). I ran `VMVersionTagTest` and `DiskVersionTagTest`—the two subclasses, and they both pass.

geode-core/src/main/java/org/apache/geode/internal/cache/versions/VersionTag.java
bschuchardt: We discussed not duplicating the checks made earlier and I was deeply unsatisfied with redoing checks on "flags".  Just being OCD I guess.

geode-core/src/test/java/org/apache/geode/internal/cache/versions/AbstractVersionTagTestBase.java
bschuchardt: oooh - I'll fix that.  I hate flaky tests

geode-core/src/test/java/org/apache/geode/internal/cache/versions/AbstractVersionTagTestBase.java
bschuchardt: That bit is complicated by the other change I made - to use Objects.equals() instead of "==" for equality checks.

bschuchardt: If you revert all of the changes to VersionTag.toData I think you'll find that the test fails with the method only writing one member ID.  That's the error that we're actually seeing cause the BufferUnderflowExceptions.  You'll also see the other new test "testSerializationWritesNoMemberID" failing because the old logic was writing out the previousMemberID even though it was being told explicitly to _not_ write any member IDs.

Bill: With changes reverted I see one test failing: `testBufferUnderflowFromOldVersionIsIgnored()`. It fails with `BufferUnderflowException` ✓

geode-core/src/main/java/org/apache/geode/internal/cache/versions/VersionTag.java
Bill: ok

geode-core/src/test/java/org/apache/geode/internal/cache/versions/AbstractVersionTagTestBase.java
bschuchardt: All three new tests should be failing if you revert all changes in toData

Bill: confirmed

bschuchardt: bill told me in a pm that those other two tests also failed for him

Bill:
Looks good now @bschuchardt

PR #4074, opened 2019-09-19
GEODE-7220 NPE in the logs on reconnect

bschuchardt:
Use reflection to poke a new Receiver into JGroups to avoid
synchronization problems in JGroups code.

No new tests were needed.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/distributed/internal/membership/gms/messenger/JGroupsMessenger.java
+++ b/geode-core/src/main/java/org/apache/geode/distributed/internal/membership/gms/messenger/JGroupsMessenger.java
@@ -184,9 +184,9 @@ public class JGroupsMessenger implements Messenger {
 
   public static void setChannelReceiver(JChannel channel, Receiver r) {
     try {
-      // Channel.setReceiver() won't set a new receiver unless the field is null but that
-      // causes NPEs in receiver threads because use of the the field isn't synchronized.
-      // Attempt to use reflection to avoid having the field be null. See GEODE-7220
+      // Channel.setReceiver() will issue a warning if we try to set a new receiver
+      // and the channel already has one. Rather than set the receiver to null &
+      // then establish a new one we use reflection to set the channel receiver. See GEODE-7220
       Field receiver = Channel.class.getDeclaredField("receiver");
       receiver.setAccessible(true);
       receiver.set(channel, r);
--- a/geode-core/src/main/java/org/apache/geode/distributed/internal/membership/gms/messenger/JGroupsMessenger.java
+++ b/geode-core/src/main/java/org/apache/geode/distributed/internal/membership/gms/messenger/JGroupsMessenger.java
@@ -49,11 +49,13 @@ import java.util.stream.Collectors;
 import it.unimi.dsi.fastutil.ints.Int2ObjectOpenHashMap;
 import org.apache.logging.log4j.Logger;
 import org.jgroups.Address;
+import org.jgroups.Channel;
 import org.jgroups.Event;
 import org.jgroups.JChannel;
 import org.jgroups.Message;
 import org.jgroups.Message.Flag;
 import org.jgroups.Message.TransientFlag;
+import org.jgroups.Receiver;
 import org.jgroups.ReceiverAdapter;
 import org.jgroups.View;
 import org.jgroups.ViewId;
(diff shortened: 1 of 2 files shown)


geode-core/src/main/java/org/apache/geode/distributed/internal/membership/gms/messenger/JGroupsMessenger.java
```diff
@@ -179,6 +182,19 @@
    */
   private JGroupsReceiver jgroupsReceiver;
 
+  public static void setChannelReceiver(JChannel channel, Receiver r) {
+    try {
+      // Channel.setReceiver() won't set a new receiver unless the field is null but that
```
Bill: Looking at `Channel.setReceiver()`, this comment is false. Dig:

```java
    public void setReceiver(Receiver r) {
        if(receiver != null && r != null)
            log.warn("%s: receiver already set", getAddress());
        receiver=r;
    }
```

geode-core/src/main/java/org/apache/geode/distributed/internal/membership/gms/messenger/JGroupsMessenger.java
```diff
@@ -179,6 +182,19 @@
...
+      // causes NPEs in receiver threads because use of the the field isn't synchronized.
+      // Attempt to use reflection to avoid having the field be null. See GEODE-7220
+      Field receiver = Channel.class.getDeclaredField("receiver");
+      receiver.setAccessible(true);
+      receiver.set(channel, r);
```
Bill: I don't understand how this change addresses the issue (of a thread getting an NPE). The field is neither `volatile` nor are accesses to it `synchronized`. Threads that have read a `null` value for this field may (in general) continue to see a `null` value for it, after this block completes.

geode-core/src/main/java/org/apache/geode/distributed/internal/membership/gms/messenger/JGroupsMessenger.java
bschuchardt: I'll fix that comment.

geode-core/src/main/java/org/apache/geode/distributed/internal/membership/gms/messenger/JGroupsMessenger.java
bschuchardt: The state of the "receiver" field is established prior to starting the channel's reader threads.  What I'm trying to avoid is setting it to null again after that happens and also avoid a "warning" message, which will trigger an operator alert.
That we're hitting NPEs due to this field being modified with setReceiver(null) tells me that synchronization is happening somewhere BUT the jgroups code is using an anti-pattern of
reading the state of "receiver" in one method and then assuming that it hasn't changed state in a method that it invokes right after that read.

if (receiver != null) notifyReciever()
.
.
.
notifyReceiver() {
  receiver.receive(message); // NPE here
}

bschuchardt: ... but I can see that there are no synchronizations between the two uses of "receiver", so we're just seeing the new state by chance.  That will be okay as long as it isn't null.  Both quorum-checker and jgroupsmessenger can respond to and consume "ping-pong" messages and startup retry logic can handle any other lost messages during the transition from one to the other.

Bill: I can see how this avoids the warning message generated in `setReceiver(…)`. I don't see how this code avoids setting the field to `null` after starting the reader threads.

bschuchardt: JGroups threads will never see a null "receiver" with this code change because we establish a non-null "receiver" before starting JGroups threads.  The Thread.start() method is synchronized.  Thereafter we never set it to null again.

Bill: ah is that due to the change in `GMSQuorumChecker.java`? (I hadn't seen that change, initially)

geode-core/src/main/java/org/apache/geode/distributed/internal/membership/gms/messenger/JGroupsMessenger.java
Bill: ok

Bill:
Looks good @bschuchardt

PR #4075, opened 2019-09-19
GEODE-6964: Move geode log4j core classes to geode-log4j

kirklund:
1) [GEODE-6964: Move geode log4j core classes to geode-log4j](https://github.com/apache/geode/pull/4075/commits/9ffdd321f2471795ba7d6dc8b512852352b59ad6)
```
commit 9ffdd321f2471795ba7d6dc8b512852352b59ad6
Author: Kirk Lund <klund@apache.org>
Date:   Thu Aug 8 15:17:54 2019 -0700

    GEODE-6964: Move geode log4j core classes to geode-log4j
    
    Introduce new Logging and Alerting SPIs. Extract all log4j-core code to
    geode-log4j module.
    
    The geode-core module no longer contains log4j2.xml and no longer has a
    dependency on log4j-core.
    
    All code that uses log4j-core has moved to the new module geode-log4j.
    The log4j2.xml for Geode now lives in geode-log4j as well. These
    changes ensure that users have better control over logging including
    which backend to use. This should improve user experience when using
    Spring Boot.
    
```
2)  [GEODE-7152: Send alert messages using executor](https://github.com/apache/geode/pull/4075/commits/24a4b5ec01a343d90ab5a6515fe011d3d232485c)
```
commit 24a4b5ec01a343d90ab5a6515fe011d3d232485c
Author: Kirk Lund <klund@apache.org>
Date:   Tue Sep 17 10:39:05 2019 -0700

    GEODE-7152: Send alert messages using executor
    
    AlertAppender uses a ThreadLocal to prevent recursive calls from
    actually doing anything. However, a recent upgrade to our log4j
    dependencies seems to have changed the behavior such that log4j refuses
    to invoke doAppend if the thread is currently handling a sendAlert
    initiated from doAppend. To fix this bug, sendAlert must be async.
```


geode-core/src/main/java/org/apache/geode/internal/util/ListCollectingServiceLoader.java
```diff
@@ -23,17 +21,17 @@
...
 import org.apache.geode.annotations.VisibleForTesting;
-import org.apache.geode.internal.logging.LogService;
 
 /**
  * Implements {@link CollectingServiceLoader} by returning a {@link List} of all currently loadable
  * implementations of the given service interface.
  */
 public class ListCollectingServiceLoader<S> implements CollectingServiceLoader<S> {
-  private static final Logger logger = LogService.getLogger();
+  private static final Logger logger = LogManager.getLogger();
```
mcmellawatt: This is more of a question than a concern, but when do we use LogManager vs LogService to get the logger?  I see a mix of both in this PR.  Does it depend on the context in some way?

kirklund: LogService uses ListCollectingServiceLoader to load the Provider impl for LogService. This means that LogService isn't ready when ListCollectingServiceLoader is used. If I leave it as LogService then it goes into recursion.

Basically, there are some classes that are used to initialize logging (such as ListCollectingServiceLoader) which need to either not log themselves or they have to use non-Geode logging. 

Classes such as ListCollectingServiceLoader should maybe move to geode-common and then that module must never have a dependency on geode-core or geode-logging. We might want to discuss that further...

kirklund:
The only precheckin failures were AnalyzeSerializablesJUnitTest and AnalyzeManagementSerializablesJUnitTest because the packages for the alerting/logging classes were no longer correct in excludedClasses.txt.

I just pushed a fix for this.

pivotal-jbarrett:
> I notice there are a lot of "exclude module 'geode-core'" lines sprinkled through the build files - this seems like a smell that we are doing something strange. Do we really need any of these?

I think these are left over from issues with "old-versions" from a really old version of the Gradle files. I think they could be cleaned up now but nobody has taken the time to look into it.

pivotal-jbarrett:
> One minor change I'd like to see before merging this - can we make log4j-api an implementation dependency (not API). Nothing in the API is using log4j as it stands with these changes.

If `LogManager` is still in `geode-core` then `log4j-api` is an `api` scoped module. Downstream modules that use `LogManager` need `log4j-api` on the compile class path, which is what `api` scope will do.

upthewaterspout:
> I think these are left over from issues with "old-versions" from a really old version of the Gradle files. I think they could be cleaned up now but nobody has taken the time to look into it.

I'm talking about new excludes that are added by this commit. For example, this change in geode-dunit. Maybe this is just copying a bad pattern in our gradle files?
```
compile(project(':geode-core'))
+  implementation(project(':geode-log4j')) {
+    exclude module: 'geode-core'
+  }

```

pivotal-jbarrett:
> > I think these are left over from issues with "old-versions" from a really old version of the Gradle files. I think they could be cleaned up now but nobody has taken the time to look into it.
> 
> I'm talking about new excludes that are added by this commit. For example, this change in geode-dunit. Maybe this is just copying a bad pattern in our gradle files?

I think you're right that it's a copy of a previous pattern. I am pretty sure it is safe to cleanup now. There is some danger in the rolling upgrade tests. I would suggest a separate PR to just clean up the dependency mess. There are other things like this we could clean up. Also converting to Java Library plugin all over should clean up some messes.

upthewaterspout:
> If `LogManager` is still in `geode-core` then `log4j-api` is an `api` scoped module. Downstream modules that use `LogManager` need `log4j-api` on the compile class path, which is what `api` scope will do.

log4j-api used to be an implementation dependency. I think @kirklund originally flipped it to API because he was adding new public API that depended on log4j-api, and didn't flip it back to implementation when he made that API internal again.

pivotal-jbarrett:
> log4j-api used to be an implementation dependency. I think @kirklund originally flipped it to API because he was adding new public API that depended on log4j-api, and didn't flip it back to implementation when he made that API internal again.

👍

I can support that since originally it was `implementation` scoped and internal downstream modules already had to explicitly include `log4j-api` this should be a no-op change. When we do get to the point of spinning out a `geode-logging` module that would be the appropriate time to make it `api` scope.

kirklund:
I flipped log4j-api from implementation to api when Robert was pairing with me. He explained what it meant and because the other modules are all getting log4j-api Loggers from LogService which lives in geode-core, we determined that it was supposed to be api. Even though LogService still lives in an internal package, it does still meet the definition of an api dependency instead of implementation.

I think the problem here is that we're trying to saying it's an implementation dependency for users of Geode, but it's an api dependency between Geode modules.

kirklund:
I don't want to risk this PR with too many additional changes that I'm unsure of, so I've filed https://github.com/apache/geode/pull/4079 to test this PR with the geode-core exclusions removed for geode-log4j dependency statements in the gradle files.

kirklund:
I filed https://github.com/apache/geode/pull/4080 to test this PR with geode-log4j changed from api dependency to implementation dependency in geode-core.

kirklund:
GREEN precheckin results

kirklund:
I changed the log4j-api dependency in geode-core from `api` back to `implementation` and updated the expected pom files.

I'm having too much trouble with removing the `exclude geode-core` clauses from geode-log4j dependency declaration, so I'm going to leave that change for someone else to followup on. There are many other exclusions like this that would also need to be removed anyway.

PR #3178, opened 2019-02-08
GEODE-6380: Add static analysis for mutable static fields

upthewaterspout:
Add a pmd rule to look for any static fields that are marked as static but not final, or static final but have values that can be modified (like a HashMap). This will allow us to eventually eliminate and enforce not having mutable static fields.

Adding a few annotations to categorize our static fields:

- Immutable - mark classes or fields that are really immutable as immutable. The pmd rule can ignore these.
- MakeImmutable - Temporary annotation to mark fields that aren't immutable but should be. For example a static final HashMap that we never change.
- MakeNonStatic - Temporary annotation mark fields that are static that we are allowing to be modified at runtime. These fields should be turned into instance fields or cache.
- MutableForTests - Temporary annotation to mark fields that should be immutable, but are not to allow setting them in tests. These also should eventually be turned into instance fields.

Annotating fields within geode-core with these annotations.

PR #4078, opened 2019-09-20
GEODE-7226: Fix StaticFieldsMustBeImmutable docs

jujoramos:
- Fixed documentation for the 'StaticFieldsMustBeImmutable' rule so 
  it matches the actual rule's behavior.

### For all changes:

### Note:

### DIFF
--- a/static-analysis/pmd-rules/src/main/java/org/apache/geode/pmd/Annotations.java
+++ b/static-analysis/pmd-rules/src/main/java/org/apache/geode/pmd/Annotations.java
@@ -14,13 +14,19 @@
  */
 package org.apache.geode.pmd;
 
+import net.sourceforge.pmd.lang.java.ast.ASTClassOrInterfaceDeclaration;
 import net.sourceforge.pmd.lang.java.ast.ASTFieldDeclaration;
 
 public class Annotations {
+
   static boolean hasDocumentedAnnotation(ASTFieldDeclaration node) {
     return node.isAnnotationPresent("org.apache.geode.annotations.Immutable") ||
         node.isAnnotationPresent("org.apache.geode.annotations.internal.MakeNotStatic") ||
         node.isAnnotationPresent("org.apache.geode.annotations.internal.MakeImmutable") ||
         node.isAnnotationPresent("org.apache.geode.annotations.internal.MutableForTesting");
+  }
+
+  static boolean hasInmutableAnnotation(ASTClassOrInterfaceDeclaration node) {
+    return node.isAnnotationPresent("org.apache.geode.annotations.Immutable");
   }
 }
--- a/static-analysis/pmd-rules/src/main/java/org/apache/geode/pmd/Annotations.java
+++ b/static-analysis/pmd-rules/src/main/java/org/apache/geode/pmd/Annotations.java
@@ -14,19 +14,13 @@
  */
 package org.apache.geode.pmd;
 
-import net.sourceforge.pmd.lang.java.ast.ASTClassOrInterfaceDeclaration;
 import net.sourceforge.pmd.lang.java.ast.ASTFieldDeclaration;
 
 public class Annotations {
-
   static boolean hasDocumentedAnnotation(ASTFieldDeclaration node) {
     return node.isAnnotationPresent("org.apache.geode.annotations.Immutable") ||
         node.isAnnotationPresent("org.apache.geode.annotations.internal.MakeNotStatic") ||
         node.isAnnotationPresent("org.apache.geode.annotations.internal.MakeImmutable") ||
         node.isAnnotationPresent("org.apache.geode.annotations.internal.MutableForTesting");
-  }
-
-  static boolean hasInmutableAnnotation(ASTClassOrInterfaceDeclaration node) {
-    return node.isAnnotationPresent("org.apache.geode.annotations.Immutable");
   }
 }
--- a/static-analysis/pmd-rules/src/main/java/org/apache/geode/pmd/StaticFieldsMustBeImmutable.java
+++ b/static-analysis/pmd-rules/src/main/java/org/apache/geode/pmd/StaticFieldsMustBeImmutable.java
@@ -15,7 +15,6 @@
 package org.apache.geode.pmd;
 
 import static org.apache.geode.pmd.Annotations.hasDocumentedAnnotation;
-import static org.apache.geode.pmd.Annotations.hasInmutableAnnotation;
 
 import java.io.File;
 import java.math.BigInteger;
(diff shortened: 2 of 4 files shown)


jujoramos:
Hey @upthewaterspout,
Yes, the idea is that, if you annotate a class with `@Immutable`, then all fields within **that single class** will be ignored by the rule. This is particularly helpful for classes on which there are multiple effectively immutable fields (like `org.apache.geode.internal.serialization.Version`).

demery-pivotal:
I'm not happy about this change. I'm concerned that PMD will no longer be able to inform us if we accidentally make a field mutable, or if we accidentally omit `final`.

jujoramos:
@demery-pivotal: I hear and - at some level - agree with you concern... however, this is still a bug form my perspective as the annotation and PMD analysis is not working as it should have worked since we introduced this to Geode back in February. See the [discussion](https://markmail.org/message/ekuvv24orjha64lu) and the [original's PR description](https://github.com/apache/geode/pull/3178) for further details.
I expect the PMD analysis to totally ignore the static "mutable" (not really) fields within my class if I add the `@Immutable` annotation to the class itself, by adding the annotation I'm explicitly stating that the entire class is immutable and it's my responsibility to make sure this stands truth. We can certainly make mistakes, of course, but that's when the reviewers from the community should point them out while analysing the PR.
Just my 2 cents, please let me know your thoughts.

onichols-pivotal:
A class-level @Immutable annotation would be useful iff PMD could leverage it to enforce that all fields are immutable or effectively immutable or annotated @Immutable. 

As it is, this PR seems to be proposing to de-automate a check currently performed by PMD, instead expecting "reviewers from the community should point them out while analysing the PR."  This seems like a step backwards.  Why is it such a bad thing that fields be analyzed and annotated individually?

jujoramos:
@onichols-pivotal: I'm not saying it's a bad thing (a bit verbose sometimes, yes, not a bad thing per se), I'm saying that the analysis **_is not working as it's intended, neither as it's documented right now_**, so it's a bug that needs to be fixed in my opinion.
To give you some more context: I opened a PR the other day and one of my classes had multiple effectively immutable fields so I annotated it directly with `@Immutable`, as described [here](https://github.com/apache/geode/blob/develop/static-analysis/pmd-rules/src/main/resources/org/apache/geode/pmd/staticimmutable.xml#L35-L36): _**you can annotate that class or the field**_. The build CI, however, failed with `All static final fields must be immutable objects. No mutable static state is allowed`; it wasn't obvious why the failure was happening until I checked the source code for `StaticFieldsMustBeImmutable` class and found the "bug"...
Sure, I can go ahead and change the `PR` to just modify the rule documentation/description of the rule so the user knows that only the field level annotation is supported but, considering that this was [discussed and approved](https://markmail.org/message/ekuvv24orjha64lu) through the dev list, I still believe it's better to change rule to behave as it was originally proposed.

onichols-pivotal:
I agree that the docs should match the behavior.

If I were a new developer and saw @Immutable on a class, I would assume[1] it was meant to *enforce* that *all* (especially non-static) members were a) final and b) references to other @Immutable classes (deeply immutable), similar to how other popular @Immutable tags work [2][3].  I would be surprised to discover it actually enforces nothing (actually the reverse, it is a suppression), and very surprised that it actually applies only shallowly and only to statics, especially since its javadoc comment claims it is "broadly similar to the Java Concurrency in Practice annotation".

Static members are not generally thought of as part of the class, just namespaced global variables/constants.  I would much prefer that the only way to mark _static_ references as immutable is one-at-a-time.

The discuss thread from February seems far from settled on how our version of @Immutable should behave at the class level, as the discussion was really centered elsewhere.

In summary, the concept of an _immutable reference_ and an _immutable class_ are two very different things; we really should not be using the same annotation for both.  We should never have overloaded @Immutable, as it has precedent as a class-level-only annotation; we should have picked a different name for our custom Geode annotation for marking effectively-immutable _references_, e.g. @EffectivelyImmutable.

That said, annotations like @Immutable do not actually prescribe how any tooling might or might not use them.  So, as long as classes being marked with @Immutable do in fact follow the generally-accepted semantics of an immutable class, then I guess also using it as a bulk "ignore" flag to work around PMD shortcomings could be tenable for now.

[1] https://stackoverflow.com/questions/27101978/what-is-the-advantage-of-annotating-an-immutable-java-class-with-immutable
[2] https://errorprone.info/bugpattern/Immutable
[3] http://jcip.net/annotations/doc/net/jcip/annotations/Immutable.html

jujoramos:
@onichols-pivotal: agreed, and the detailed response is really appreciated 👍.
That said, just one more thing: according to the [original PR description](https://github.com/apache/geode/pull/3178):

```
Immutable - mark classes or fields that are really immutable as immutable. The pmd rule can ignore these.
```

The above (plus the fact that my class was effectively immutable and annotated with the `@Immutable` annotation) is what made me think I was actually hitting a bug in our custom rule, that's why I opened the ticket and made the changes. Deciding whether we should rename and re-think our current annotations should be addressed through another ticket, and probably after some discussion within the dev list.
If there's not enough consensus to merge this PR I'll just change the rule documentation without touching the actual implementation, at the end of the day it won't kill anyone to annotate multiple fields instead of the class 🤷‍♂.

@upthewaterspout @Bill : thoughts?.

Bill:
IMO it is not ok for the `StaticFieldsMustBeImmutable` rule to ignore classes with the `@Immutable` annotation.

The only purpose I can see, of an `@Immutable` anno, would be to express developer intent similarly to the way `@Override` does. For the expression of intent to be most useful, it would be enforced somehow, e.g. if I mark a class as `@Immutable` but that class has a non-final field, I should get a build-time error.

Mutability of static fields is an issue orthogonal to mutability of the class that contains the fields. Whether or not a class (i.e. its instances) are mutable, it should be viewed as a design bug, if that class has any non-final static field, or any final static field referencing a mutable object type.

So I'm a -1 on this PR as it stands.

upthewaterspout:
I guess I'm in favor of not merging this PR.

Originally I was hoping we could just use `@Immutable` like the JCP annotation - putting it on a class would mean instances of the class are immutable. Then PMD could enforce that static final fields only to refer to `@Immutable` classes. So the fact that you can put `@Immutable` on a static field is really more of a workaround to the limitations of PMD, and maybe the fact that some static fields don't refer to immutable classes that we control.

I agree there is probably more value in using `@Immutable` on a class as documentation that the *instances* of the class are immutable.

jujoramos:
@upthewaterspout @Bill @onichols-pivotal @demery-pivotal 

Thanks all for the feedback, instead of modifying the rule's implementation I'll just update its documentation to prevent confusions when the PMD task fails.
I won't open a new PR though, will just update this one instead as the discussion thread is pretty good and might be useful in the future when looking at the ticket 👍.
Cheers.

Bill:
Good one Juan.

PR #4080, opened 2019-09-20
TEST: geode-log4j with impl dependency in geode-core instead of api

kirklund:
Do not review and do not merge.


kirklund:
Build failures include checkPom. This can be updated and fixed with:
```
./gradlew :geode-dunit:updateExpectedPom
```

PR #4081, opened 2019-09-20
GEODE-1074 Create geode-gfsh project.

rhoughton-pivot:

### For all changes:

- [X] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)?

### Note:

### DIFF
--- a/boms/geode-all-bom/src/test/resources/expected-pom.xml
+++ b/boms/geode-all-bom/src/test/resources/expected-pom.xml
@@ -202,7 +202,7 @@
       <dependency>
         <groupId>io.github.classgraph</groupId>
         <artifactId>classgraph</artifactId>
-        <version>4.0.6</version>
+        <version>4.8.47</version>
         <scope>compile</scope>
       </dependency>
       <dependency>
--- a/buildSrc/src/main/groovy/org/apache/geode/gradle/plugins/DependencyConstraints.groovy
+++ b/buildSrc/src/main/groovy/org/apache/geode/gradle/plugins/DependencyConstraints.groovy
@@ -111,7 +111,7 @@ class DependencyConstraints implements Plugin<Project> {
         api(group: 'commons-logging', name: 'commons-logging', version: '1.2')
         api(group: 'commons-modeler', name: 'commons-modeler', version: '2.0.1')
         api(group: 'commons-validator', name: 'commons-validator', version: get('commons-validator.version'))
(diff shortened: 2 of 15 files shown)


pivotal-jbarrett:
It does look like it removes the manifest jar. We probably need to keep that around until the next major as "api".

rhoughton-pivot:
@upthewaterspout Currently this is a fat-jar of all dependencies. The up-side is that it would give a single artifact for download that would allow CLI interface with a running cluster. I was literally testing with `java -jar geode-gfsh.jar` and getting good results

rhoughton-pivot:
@pivotal-jbarrett Note, this is only a draft so far

jinmeiliao:
if we do decide to do this. Core needs to depend on this gfsh.jar. Or we can move all the commands  to a module, create commands.jar, and have geode-core and geode-gfsh both depend on that.

jdeppe-pivotal:
You'll probably discover this when looking at the test failures, but the default build doesn't build `lib/geode-gfsh-x.y.z.jar`

jdeppe-pivotal:
@jinmeiliao I'm not sure that `geode-core` needs to depend on this but it is a good idea to pull all the gfsh commands into their own module.

rhoughton-pivot:
@jdeppe-pivotal I'll modify the build ordering to make sure that `:geode-gfsh:bootJar` is on the default `build` list.

jdeppe-pivotal:
Zip files are nice because the directory info is appended at the **end** of the file which allows for interesting things like prepending an `exe` to the zip. Although this current solution gets to the point of allowing a user to simply do `java -jar geode-gfsh.jar` we could take it one step further and prepend a small exe which essentially does what the `gfsh` script  does now. So it would truly be a single executable. Yes, that would mean separate binaries for each OS.

For inspiration: https://github.com/GeertJohan/go.rice

jdeppe-pivotal:
In case anyone keeps working on this, I did a bit more work to allow launching locators and servers and running embedded webapps: https://github.com/jdeppe-pivotal/geode/commit/e56aa2c2aeaeda4bbe1e78d69004fb549590020e

jmelchio:
@rhoughton-pivot given that we now have a `geode-gfsh` module should this be closed?

metatype:
Closing this out in favor of doing an RFC.  We can still refer to the ideas proposed in this PR.

PR #4083, opened 2019-09-23
GEODE-6897: revert accidental change of apidoc's link for dev rest api.

jinmeiliao:
we should not change the api docs link for dev rest api.

### DIFF
--- a/geode-web-api/src/main/java/org/apache/geode/rest/internal/web/security/RestSecurityConfiguration.java
+++ b/geode-web-api/src/main/java/org/apache/geode/rest/internal/web/security/RestSecurityConfiguration.java
@@ -50,7 +50,7 @@ public class RestSecurityConfiguration extends WebSecurityConfigurerAdapter {
   protected void configure(HttpSecurity http) throws Exception {
     http.sessionManagement().sessionCreationPolicy(SessionCreationPolicy.STATELESS).and()
         .authorizeRequests()
-        .antMatchers("/ping", "/docs/**", "/swagger-ui.html", "/experimental/api-docs/**",
+        .antMatchers("/ping", "/docs/**", "/swagger-ui.html", "/v1/api-docs/**",
             "/webjars/springfox-swagger-ui/**", "/swagger-resources/**")
         .permitAll().anyRequest().authenticated().and().csrf().disable();
 
--- a/geode-web-api/src/main/resources/swagger.properties
+++ b/geode-web-api/src/main/resources/swagger.properties
@@ -12,5 +12,4 @@
 # or implied. See the License for the specific language governing permissions and limitations under
 # the License.
 #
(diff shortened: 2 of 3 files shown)


jinmeiliao:
@jmelchio

onichols-pivotal:
Does this change need to go into 1.10 as well?

onichols-pivotal:
What happens when we go to v3?  If this is not supposed to be a versioned link, how is v2 any better?

jinmeiliao:
In fact, since dev rest is always under /geode/v1, I prefer to have the dev rest api also be under /geode/v1/api-docs, instead of geode/v2/api-docs. The original swagger api docs for dev rest is in fact under /geode/v1/api-docs, but when Kevin upgraded swagger to use the version 2 of api-docs, he accidentally changed v1 to v2. If we want to roll back to what it was before all the mis-haps, the dev rest api-docs link should remain as "/geode/v1/api-docs"

geode-web-api/src/main/java/org/apache/geode/rest/internal/web/security/RestSecurityConfiguration.java
```diff
@@ -50,7 +50,7 @@ public AuthenticationManager authenticationManagerBean() throws Exception {
   protected void configure(HttpSecurity http) throws Exception {
     http.sessionManagement().sessionCreationPolicy(SessionCreationPolicy.STATELESS).and()
         .authorizeRequests()
-        .antMatchers("/ping", "/docs/**", "/swagger-ui.html", "/experimental/api-docs/**",
+        .antMatchers("/ping", "/docs/**", "/swagger-ui.html", "/v1/api-docs/**",
```
dschneider-pivotal: Does this need to be versioned? Should it be more like the "docs" one and just be "/api-docs/**"?

geode-web-api/src/main/resources/swagger.properties
```diff
@@ -12,5 +12,4 @@
 # or implied. See the License for the specific language governing permissions and limitations under
 # the License.
 #
-springfox.documentation.swagger.v2.path=/experimental/api-docs
-springfox.documentation.swagger.v1.path=/v1/api-docs
\ No newline at end of file
+springfox.documentation.swagger.v2.path=/v1/api-docs
```
dschneider-pivotal: Why is the variable name still "v2" since the value has been changed to "v1"?

jinmeiliao: because swagger library is updated, our swagger version is v2.

geode-web-api/src/main/java/org/apache/geode/rest/internal/web/security/RestSecurityConfiguration.java
jinmeiliao: there is no difference for now, since we only have one version of dev rest.

PR #4086, opened 2019-09-24
GEODE-6807: cache adviseUpdate and adviseAllEventsOrCached

mivanac:

### For all changes:

- [] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)?

### Note:

### DIFF
--- a/geode-core/src/integrationTest/java/org/apache/geode/internal/cache/CacheDistributionAdvisorConcurrentTest.java
+++ b/geode-core/src/integrationTest/java/org/apache/geode/internal/cache/CacheDistributionAdvisorConcurrentTest.java
@@ -61,9 +61,9 @@ public class CacheDistributionAdvisorConcurrentTest {
 
   }
 
+  @Test
   public void getAdviseUpdateForConcurrentUpdateShouldSucceed(
       ParallelExecutor executor) throws Exception {
-
 
     EntryEventImpl event = new EntryEventImpl();
     event.setNewValue(null);
--- a/geode-core/src/integrationTest/java/org/apache/geode/internal/cache/CacheDistributionAdvisorConcurrentTest.java
+++ b/geode-core/src/integrationTest/java/org/apache/geode/internal/cache/CacheDistributionAdvisorConcurrentTest.java
@@ -1,0 +1,105 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.apache.geode.internal.cache;
+
+
+import static org.apache.geode.test.concurrency.Utilities.availableProcessors;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import org.apache.geode.CancelCriterion;
+import org.apache.geode.cache.Operation;
+import org.apache.geode.distributed.internal.DistributionAdvisor;
+import org.apache.geode.distributed.internal.DistributionManager;
+import org.apache.geode.distributed.internal.membership.InternalDistributedMember;
+import org.apache.geode.distributed.internal.membership.MemberAttributes;
+import org.apache.geode.internal.cache.CacheDistributionAdvisor.CacheProfile;
+import org.apache.geode.test.concurrency.ConcurrentTestRunner;
+import org.apache.geode.test.concurrency.ParallelExecutor;
+
+
+@RunWith(ConcurrentTestRunner.class)
+public class CacheDistributionAdvisorConcurrentTest {
+  private final int count = availableProcessors() * 2;
+
+  @Test
+  public void getAdviseAllEventsOrCachedForConcurrentUpdateShouldSucceed(
+      ParallelExecutor executor) throws Exception {
+
+    DistributionAdvisor advisor = createCacheDistributionAdvisor();
+    CacheProfile profile = createCacheProfile();
+    advisor.putProfile(profile, true);
+
+    executor.inParallel(() -> {
+      ((CacheDistributionAdvisor) advisor).adviseAllEventsOrCached();
+    }, count);
+    executor.execute();
+
+    assertTrue(((CacheDistributionAdvisor) advisor).adviseAllEventsOrCached()
+        .contains(profile.getDistributedMember()));
+    assertEquals(((CacheDistributionAdvisor) advisor).adviseAllEventsOrCached().size(), 1);
+
+  }
+
+  public void getAdviseUpdateForConcurrentUpdateShouldSucceed(
+      ParallelExecutor executor) throws Exception {
+
+
+    EntryEventImpl event = new EntryEventImpl();
+    event.setNewValue(null);
+    event.setOperation(Operation.CREATE);
+
+    DistributionAdvisor advisor = createCacheDistributionAdvisor();
+    CacheProfile profile = createCacheProfile();
+    advisor.putProfile(profile, true);
+
+    executor.inParallel(() -> {
+      ((CacheDistributionAdvisor) advisor).adviseUpdate(event);
+    }, count);
+    executor.execute();
+
+    assertTrue(((CacheDistributionAdvisor) advisor).adviseAllEventsOrCached()
+        .contains(profile.getDistributedMember()));
+    assertEquals(((CacheDistributionAdvisor) advisor).adviseAllEventsOrCached().size(), 1);
+
+  }
+
+  private DistributionAdvisor createCacheDistributionAdvisor() {
+    CacheDistributionAdvisee advisee = mock(CacheDistributionAdvisee.class);
+    CancelCriterion cancelCriterion = mock(CancelCriterion.class);
+    when(advisee.getCancelCriterion()).thenReturn(cancelCriterion);
+    DistributionManager distributionManager = mock(DistributionManager.class);
+    when(advisee.getDistributionManager()).thenReturn(distributionManager);
+    CacheDistributionAdvisor result =
+        CacheDistributionAdvisor.createCacheDistributionAdvisor(advisee);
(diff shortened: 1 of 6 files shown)


geode-core/src/main/java/org/apache/geode/distributed/internal/DistributionAdvisor.java
```diff
@@ -1314,7 +1320,7 @@ public Set adviseProfileRemove() {
   private synchronized boolean basicAddProfile(Profile p) {
     // don't add more than once, but replace existing profile
     // try {
-
+    profilesVersion++;
```
pivotal-jbarrett: You are safe here because the method is synchronized but `volatile` increment operations are not atomic.

geode-core/src/main/java/org/apache/geode/internal/cache/CacheDistributionAdvisor.java
```diff
@@ -145,14 +153,22 @@ public String toString() {
...
+        || excludeInRecovery != prevExcludeInRecovery) {
+      adviseSetforAllEvents = Collections.unmodifiableSet(adviseFilter(profile -> {
+        assert profile instanceof CacheProfile;
+        CacheProfile cp = (CacheProfile) profile;
+        if (excludeInRecovery && cp.inRecovery) {
+          return false;
+        }
+        return cp.cachedOrAllEventsWithListener();
+      }));
+      adviseAllEventsVersion = profilesVersion;
```
pivotal-jbarrett: Between the above check and this update the `profileVersion` can be concurrently updated by calls to `basicAdd/Remove` methods.

geode-core/src/main/java/org/apache/geode/internal/cache/CacheDistributionAdvisor.java
```diff
@@ -167,13 +183,18 @@ Set adviseUpdate(final EntryEventImpl event) throws IllegalStateException {
...
-      });
+      if (adviseUpdateVersion != profilesVersion || adviseUpdateVersion == 0) {
+        adviseSetforUpdate = Collections.unmodifiableSet(adviseFilter(profile -> {
+          assert profile instanceof CacheProfile;
+          CacheProfile cp = (CacheProfile) profile;
+          DataPolicy dp = cp.dataPolicy;
+          return dp.withReplication()
+              || (cp.allEvents() && (dp.withStorage() || cp.hasCacheListener));
+        }));
+        adviseUpdateVersion = profilesVersion;
```
pivotal-jbarrett: Similar race condition on version update.

geode-core/src/main/java/org/apache/geode/internal/cache/DistributedCacheOperation.java
```diff
@@ -345,7 +345,8 @@ protected void _distribute() {
 
     try {
       // Recipients with CacheOp
-      Set<InternalDistributedMember> recipients = getRecipients();
+      Set<InternalDistributedMember> recipients = new HashSet<>();
```
pivotal-jbarrett: Why are you copying the recipient set?
Also, if copying is desired use `new HashSet<>(getRecipients())`, which results in a properly sized `HashSet` on allocation.

mivanac: getRecipients() returns unmodified Set, and code (in _distribute method) few lines after, isperforming removeAll on returned Set.

geode-core/src/main/java/org/apache/geode/internal/cache/CacheDistributionAdvisor.java
```diff
@@ -145,14 +153,24 @@ public String toString() {
...
-      CacheProfile cp = (CacheProfile) profile;
-      if (excludeInRecovery && cp.inRecovery) {
-        return false;
-      }
-      return cp.cachedOrAllEventsWithListener();
-    });
+
+    // minimize volatile reads by copying ref to local var
+    long tempProfilesVersion = profilesVersion; // volatile read
+    if (adviseAllEventsVersion != tempProfilesVersion || adviseAllEventsVersion == 0
```
dschneider-pivotal: Instead of having the cached value recomputed based on the state of the boolean excludeInRecovery, would it be better just to support two cached values: one for when excludeInRecovery is false and one for when it is true? I worry that if we have a pattern of asking for both flavors then they will keep throwing away the previous cached value and we end up with no caching.

mivanac: Currently excludeInRecovery is set to true, in both places from where it is called.

dschneider-pivotal: In that case I think we should remove this parameter since it is an unused feature that is complicating the new implementation.

geode-core/src/main/java/org/apache/geode/internal/cache/CacheDistributionAdvisor.java
```diff
@@ -142,17 +149,25 @@ public String toString() {
    *
    * @param excludeInRecovery if true then members in recovery are excluded
```
dschneider-pivotal: Update this comment by removing the @param and saying something like "members that are currently in recovery will not be in the returned set.

geode-core/src/main/java/org/apache/geode/internal/cache/CacheDistributionAdvisor.java
```diff
@@ -142,17 +149,25 @@ public String toString() {
...
-      CacheProfile cp = (CacheProfile) profile;
-      if (excludeInRecovery && cp.inRecovery) {
-        return false;
-      }
-      return cp.cachedOrAllEventsWithListener();
-    });
+
+    // minimize volatile reads by copying ref to local var
+    long tempProfilesVersion = profilesVersion; // volatile read
+    if (adviseAllEventsVersion != tempProfilesVersion) {
```
dschneider-pivotal: I think you may have a java memory model problem with the boolean and Set that maintain the cached state. Neither of these are volatile but the only code that updates them on read a volatile (profilesVersion). You could have two threads that call adviseAllEventOrCached. Since neither of those threads wrote a volatile (profilesVersion) I don't think we have any JMM rules saying we will see the latest values of adviseAllEventsVersion and adviseSetforAllEvents. At first I thought this would be okay if both threads just end up computing the same result and the last one to do so would end up storing its value in the cache. But I think we might have a couple of problems. 1. we are setting two fields and it would be possible for one thread to have a later profilesVersion but then the order they update the two cached fields is not safe so the could win on one and lose on the other. So now the cached state is inconsistent with itself. 2. A thread could see the new boolean and think its adviseSetForAllEvents is up to date but when it reads that field it reads the old stale state. According to my reading of the JMM if you made "adviseAllEventsVersion" also volatile then it would address both of these issues. Another way to code for these issues is to store the cached state in a single AtomicReference whose value is a class with two final fields: a boolean and a set. I think doing it with an AtomicReference is clearer so I would recommend that.

geode-core/src/main/java/org/apache/geode/internal/cache/CacheDistributionAdvisor.java
```diff
@@ -142,17 +149,25 @@ public String toString() {
...
+      adviseSetforAllEvents = Collections.unmodifiableSet(adviseFilter(profile -> {
+        assert profile instanceof CacheProfile;
```
dschneider-pivotal: This new code on CacheDistributionAdvisor needs tests.
Add unit tests and since this code has some tricky concurrency issues add a ConcurrentTestRunner.
The ConcurrentTestRunner has been good at reproducing concurrency bugs.
See MaxLongGaugeConcurrentTest for an example of a simple ConcurrentTestRunner.

geode-core/src/main/java/org/apache/geode/internal/cache/CacheDistributionAdvisor.java
pivotal-jbarrett: I suggested a similar issue and solution in my overall review comment above.

mivanac: Solution updated according to your proposal.

mivanac: Test, are failing, so I will revert this changes

geode-core/src/main/java/org/apache/geode/internal/cache/CacheDistributionAdvisor.java
```diff
@@ -162,18 +181,31 @@ public String toString() {
...
+
+      // minimize volatile reads by copying ref to local var
+      long tempProfilesVersion = profilesVersion; // volatile read
+
+      if (adviseUpdateVersion != tempProfilesVersion) {
+        synchronized (adviseSetforUpdate) {
+          if (adviseUpdateVersion != tempProfilesVersion) {
+
+            adviseSetforUpdate = Collections.unmodifiableSet(adviseFilter(profile -> {
+              assert profile instanceof CacheProfile;
```
pivotal-jbarrett: Please remove the `assert`. This code does not get executed because we don't run with asserts enabled. There is a thread from about 2 years go where we resolved to remove and not use any `assert` statements in the source.

dschneider-pivotal:
@mivanac don't do force pushes on pull requests

geode-core/src/main/java/org/apache/geode/internal/cache/CacheDistributionAdvisor.java
dschneider-pivotal: The code a currently see looks reasonable (with the additional volatile and added sync). It looks correct. But you still need to add the requests unit and ConcurrentTestRunner tests

mhansonp:
@dschneider-pivotal @pivotal-jbarrett  Are you satisfied with the changes? the failure is a known one. It has been fixed, @mivanac  just needs to rebase and push or we rebase and merge. Thoughts?

pivotal-jbarrett:
@mhansonp I have already approved these changes.

mhansonp:
> @mhansonp I have already approved these changes.

Thanks.

mivanac:
UT and update inRecovery logic added.

mivanac:
Force pushed due to, not able to normal push
([rejected]              feature/GEODE-6807 -> feature/GEODE-6807 (non-fast-forward))

mivanac:
This is first part of work, to update solution and add UT.
I will add also ConcurrentTestRunner based test.

dschneider-pivotal:
It looks good to me now except for the need for the concurrent test

geode-core/src/integrationTest/java/org/apache/geode/internal/cache/CacheDistributionAdvisorConcurrentTest.java
```diff
@@ -0,0 +1,105 @@
...
+    }, count);
+    executor.execute();
+
+    assertTrue(((CacheDistributionAdvisor) advisor).adviseAllEventsOrCached()
+        .contains(profile.getDistributedMember()));
+    assertEquals(((CacheDistributionAdvisor) advisor).adviseAllEventsOrCached().size(), 1);
+
+  }
+
+  public void getAdviseUpdateForConcurrentUpdateShouldSucceed(
```
dschneider-pivotal: Should this method also have @Test?

mivanac:
Thanks for comment.

PR #4087, opened 2019-09-24
GEODE-7216 include a timestamp in stack trace logs

mkevo:

### For all changes:

### Note:


geode-core/src/main/java/org/apache/geode/management/internal/cli/commands/ExportStackTraceCommand.java
```diff
@@ -126,7 +128,9 @@ public String writeStacksToFile(Map<String, byte[]> dumps, String fileName) thro
       ps = new PrintWriter(os);
 
       for (Map.Entry<String, byte[]> entry : dumps.entrySet()) {
-        ps.append(STACK_TRACE_FOR_MEMBER).append(entry.getKey()).append(" ***")
+        ps.append(STACK_TRACE_FOR_MEMBER).append(entry.getKey()).append(" at ")
+            .append(new SimpleDateFormat("yyyy/MM/dd HH:mm:ss.SS").format(new Date()))
```
jujoramos: This is actually using the timestamp from the JVM on which the `gfsh` command was executed after gathering the results from all members, **it is not the actual timestamp at which the thread dump was taken on each member**. Shouldn't the timestamp be obtained directly on `GetStackTracesFunction` and be part of `StackTracesPerMember`?.

geode-core/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/ExportStackTraceCommandDUnitTest.java
```diff
@@ -97,4 +104,40 @@ public void exportStackTrace_with_file() {
     // delete this file afterwards so that we won't pollute the other tests in this class
     files[0].delete();
   }
+
+  @Test
+  public void exportStackTrace_check_file_content() throws IOException {
```
jujoramos: Can you change the method name to reflect what the test is doing and try to use _camelCase_ for the name instead of underscores?.
**_PD:_** I realise that other methods within the class use underscore, but I'd prefer us to use accepted formats when adding new code.

geode-core/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/ExportStackTraceCommandDUnitTest.java
```diff
@@ -97,4 +104,40 @@ public void exportStackTrace_with_file() {
...
+    gfsh.executeAndAssertThat("export stack-traces --file=" + stackTraceFile.getAbsolutePath())
+        .statusIsSuccess().containsOutput("stack-trace(s) exported to file");
+
+    Date date = new Date();
+    String possibleMsg = stackTraceMsg + new SimpleDateFormat("yyyy/MM/dd HH:mm:ss").format(date);
+    date.setSeconds(date.getSeconds() - 1);
+    String anotherPossibleMsg =
+        stackTraceMsg + new SimpleDateFormat("yyyy/MM/dd HH:mm:ss").format(date);
+
+    List<String> expectedStrings = Arrays.asList(possibleMsg, anotherPossibleMsg);
```
jujoramos: This will probably introduce flakiness in the tests as there's no guarantee that the `Date` created in the test will match (or be 1 second apart) from the one used in the actual command.
You'd need to use another approach to verify the changes; maybe check that the stack trace for each member **_contains_** a date and that the date follows the pattern you use within the command (`yyyy/MM/dd HH:mm:ss`)?.

geode-core/src/main/java/org/apache/geode/management/internal/cli/domain/StackTracesPerMember.java
```diff
@@ -19,16 +19,16 @@
 public class StackTracesPerMember implements Serializable {
 
   private static final long serialVersionUID = 1L;
-  private final String memberNameOrId;
+  private final String memberNameOrIdWithTimestamp;
```
jujoramos: Can we split these concepts (`memberId` and `timestamp`) in two different attributes?.
I don't see any good reason about why they should just be one single attributed created by a String concatenation (plus the `at` word). This logic is basically part of the "view" or representation and, as such, it should be handled directly in the gfsh command.

geode-core/src/main/java/org/apache/geode/management/internal/cli/functions/GetStackTracesFunction.java
```diff
@@ -34,8 +37,10 @@ public void execute(FunctionContext context) {
       if (memberNameOrId == null) {
         memberNameOrId = cache.getDistributedSystem().getDistributedMember().getId();
       }
+      String timestamp = new SimpleDateFormat("yyyy/MM/dd HH:mm:ss.SS").format(new Date());
```
jujoramos: Last request from my side, promise ✋: the `StackTracesPerMember` domain bean should have the full `Date` object, and the actual `Date` representation should be rendered in the "view" layer directly, so I'd move this part to the actual gfsh command as well.

mkevo: Hi @jujoramos , do you mean to move this line outside of the try/catch block or to the `ExportStackTraceCommand` before _for_ loop? Or to have `new Date()` in the `StackTracesPerMember` and formatting it in the `ExportStackTraceCommand`?

jujoramos: Hey @mkevo,

I mean that the `timestamp` attribute within the `StackTracesPerMember` class should be of type `Date`, not `String`. The actual formatting should happen on the `ExportStackTraceCommand`.

aaronlindsey:
Did you consider using `java.time.*` classes instead of `java.util.Date`? I think the best practice is to use the former unless there is a reason why only `java.util.Date` will do. Jon Skeet has an [interesting article](https://codeblog.jonskeet.uk/2017/04/23/all-about-java-util-date/) about the difference.

If you do switch to a `java.time.*` class, you'll have to be careful about time zones. By default `java.util.Date` uses the system-local time zone which in this case happens to be the same as the time zone in the logs so it works out by happenstance. But if you switch to `java.time.Instant` (basically the analogue to Date), that class defaults to UTC. You'd have to make sure the string version which is output by the command matches the time zone in the logs.

mkevo:
Thanks @aaronlindsey! Great article! I will change it.

geode-core/src/main/java/org/apache/geode/management/internal/cli/functions/GetStackTracesFunction.java
```diff
@@ -36,7 +36,7 @@ public void execute(FunctionContext context) {
       if (memberNameOrId == null) {
         memberNameOrId = cache.getDistributedSystem().getDistributedMember().getId();
       }
-      Date timestamp = new Date();
+      Instant timestamp = Instant.now();
```
jujoramos: This variable is redundant and not needed, you can just use `Instant.now()` when constructing the `StackTracesPerMember` instance.

geode-core/src/main/java/org/apache/geode/management/internal/cli/commands/ExportStackTraceCommand.java
```diff
@@ -50,6 +53,10 @@
   public static final String STACK_TRACE_FOR_MEMBER = "*** Stack-trace for member ";
   private final GetStackTracesFunction getStackTracesFunction = new GetStackTracesFunction();
 
+  DateTimeFormatter formatter =
+      DateTimeFormatter.ofPattern("yyyy/MM/dd HH:mm:ss.SSS", Locale.UK)
```
jujoramos: Why are you hardcoding the `Locale` to `UK` here?.
I think you should use `Locale.getDefault()` or [`DateTimeFormatter.ofPattern(String pattern)`](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html#ofPattern-java.lang.String-) directly.

geode-core/src/main/java/org/apache/geode/management/internal/cli/commands/ExportStackTraceCommand.java
```diff
@@ -50,6 +53,10 @@
   public static final String STACK_TRACE_FOR_MEMBER = "*** Stack-trace for member ";
   private final GetStackTracesFunction getStackTracesFunction = new GetStackTracesFunction();
 
+  DateTimeFormatter formatter =
```
jujoramos: Small Nitpick: now that you're using `DateTimeFormatter` (immutable and thread safe) instead of `SimpleDateFormat`, you might want to create the formatter instance just once and re-use it instead of creating it every time the command is executed.

geode-core/src/main/java/org/apache/geode/management/internal/cli/commands/ExportStackTraceCommand.java
mkevo: Yes we can set it getDefault, I didn't see it. I just pick UK to be consistent with timeStamp in log file. I will change it. Thank for  the comment!

geode-core/src/main/java/org/apache/geode/management/internal/cli/commands/ExportStackTraceCommand.java
```diff
@@ -50,6 +53,10 @@
   public static final String STACK_TRACE_FOR_MEMBER = "*** Stack-trace for member ";
   private final GetStackTracesFunction getStackTracesFunction = new GetStackTracesFunction();
 
+  private static final DateTimeFormatter formatter =
+      DateTimeFormatter.ofPattern("yyyy/MM/dd HH:mm:ss.SSS", Locale.getDefault())
+          .withZone(ZoneId.systemDefault());
```
jujoramos: The `Locale.getDefault()` parameter and the further customisation through `.withZone(ZoneId.systemDefault())` are not needed, you can achieve the same by just executing`DateTimeFormatter.ofPattern("yyyy/MM/dd HH:mm:ss.SSS")`.

mkevo: It is possible to go without `Locale.getDefault()`, but we need to add `.withZone(ZoneId.systemDefault())`. To format an `Instant` a time-zone is required. Without a time-zone, the formatter does not know how to convert the instant to human date-time fields, and therefore throws an exception.

geode-core/src/main/resources/org/apache/geode/internal/sanctioned-geode-core-serializables.txt
```diff
@@ -470,7 +470,7 @@ org/apache/geode/management/internal/cli/domain/RegionAttributesInfo,true,336184
...
 org/apache/geode/management/internal/cli/domain/RegionDescriptionPerMember,true,1,hostingMember:java/lang/String,isAccessor:boolean,name:java/lang/String,regionAttributesInfo:org/apache/geode/management/internal/cli/domain/RegionAttributesInfo,size:int
 org/apache/geode/management/internal/cli/domain/RegionInformation,true,1,dataPolicy:org/apache/geode/cache/DataPolicy,isRoot:boolean,name:java/lang/String,parentRegion:java/lang/String,path:java/lang/String,scope:org/apache/geode/cache/Scope,subRegionInformationSet:java/util/Set
-org/apache/geode/management/internal/cli/domain/StackTracesPerMember,true,1,memberNameOrId:java/lang/String,stackTraces:byte[]
+org/apache/geode/management/internal/cli/domain/StackTracesPerMember,true,1,memberNameOrId:java/lang/String,stackTraces:byte[],timestamp:java/time/Instant
```
aaronlindsey: Would `StackTracesPerMember` ever be deserialized from an older version that didn't include the timestamp field? I'm just asking — I don't actually know. Maybe @jujoramos would know. If this could happen, then the timestamp field would be null and the `formatter.format(...)` line in `ExportStackTraceCommand` could throw a NPE.

jujoramos: @aaronlindsey: good catch, I haven't thought about it but my guess is that it could happen, yes. I'm not a 100% sure, though.

mkevo: Do we need to use `try/catch` on formatter or override `toData/fromData` in `StackTracesPerMember`?

aaronlindsey: I would do something like the following: Add package private method in `ExportStackTraceCommand` which handles generating the header message. Then you can write a unit test for this more easily. I don't see an easy way to write an integeration (D-unit) test for this case.

```java
  @VisibleForTesting
  String getHeaderMessage(StackTracesPerMember stackTracesPerMember) {
    String headerMessage = stackTracesPerMember.getMemberNameOrId();

    Instant timestamp = stackTracesPerMember.getTimestamp();
    if (timestamp != null) {
      headerMessage += " at " + formatter.format(timestamp);
    }

    return headerMessage;
  }
``` 

The issue with overriding `readObject` in `StackTracesPerMember` is that you wouldn't know which value to assign to the `timestamp` field when the object is deserialized.

geode-core/src/test/java/org/apache/geode/management/internal/cli/commands/ExportStackTraceCommandTest.java
```diff
@@ -66,4 +70,14 @@ public void abortIfFileExists() throws IOException {
...
+
+  @Test
+  public void getHeaderMessage() throws IOException {
+    StackTracesPerMember stackTracePerMember =
+        new StackTracesPerMember("server", Instant.now(),
+            OSProcess.zipStacks());
+    String headerMessage = command.getHeaderMessage(stackTracePerMember);
+
+    assertThat(headerMessage).isNotNull();
+  }
```
aaronlindsey: Thanks for addressing my feedback about the possibility of the timestamp field being null when deserializing a `StackTracesPerMember`. However, I think the test would be more effective if you split it out into a couple of more specific scenarios. For example:
1. When I pass a value of `null` as the timestamp for `StackTracesPerMember`, I expect that the `headerMessage` to be exactly "server". 
2. When I pass a specific value like "2000/01/02 03:04:05.006" as the timestamp for `StackTracesPerMember`, I expect the `headerMessage` to be exactly "server at 2000/01/02 03:04:05.006".

mkevo: Thanks for comments! I added new test with null as the timestamp and modified existing to check if `headerMessage` has specific format as expected in output of this command.

aaronlindsey: Thanks! It looks good.

mivanac:
Since this changes are approved, I will merge this PR.

PR #4089, opened 2019-09-24
GEODE-7237: Fix race condition in GfshRule

kirklund:
Creation of geode-log4j widened a race condition in GfshRule in which
the process being forked would start printing output before the
ProcessLogger had been created and started in order to intercept that
output.

The widening of that race condition resulted in CI failures of
ConnectCommandAcceptanceTest.

ProcessLogger does not appear to have any real need for manipulating
log4j-core which is what widens the race condition far enough to fail
in CI.

Removing the manipulation of log4j-core from ProcessLogger reduces
the race condition for consistent passing. The window itself is still
there but without resorting to something like MainLauncher [1], there's
no way to completely close the window.

[1] geode-junit: org.apache.geode.test.process.MainLauncher

If ConnectCommandAcceptanceTest or other tests using GfshRule fail
after this commit, then we'll need to use something like MainLauncher.

Please review: @aaronlindsey

### DIFF
--- a/geode-assembly/src/acceptanceTest/java/org/apache/geode/management/internal/cli/commands/ConnectCommandAcceptanceTest.java
+++ b/geode-assembly/src/acceptanceTest/java/org/apache/geode/management/internal/cli/commands/ConnectCommandAcceptanceTest.java
@@ -40,21 +40,32 @@ public class ConnectCommandAcceptanceTest {
   public GfshRule gfshDefault = new GfshRule();
 
   @Test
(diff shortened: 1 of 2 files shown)


kirklund:
Concourse errors:
{noformat}
ERROR: (gcloud.compute.instances.create) Could not fetch resource:
 - The resource 'projects/apachegeode-ci/zones/us-central1-a/instances/heavy-lifter-cfc45a71-ed86-5774-99bc-451122fbee1f' already exists
{noformat}

PR #4095, opened 2019-09-26
GEODE-7123: Add create disk-store command option, stage-configuration

pdxrunner:
The new option, --stage-configuration, to the create disk-store command
provides the capability to define a disk-store in the cluster configuration
prior to any server members have joined the cluster. Using this option
will only update the cluster-configuration, disk-stores will not be
created on any current server members.

Servers joining the cluster (or restarting) will then inherit the new
disk-store configuration and create the appropriate directory and
disk-store files.

This is useful for persisting the PDX types to a disk-store other than
DEFAULT. When starting a cluster, the disk-store for pdx can be added to
the configuration after the locators are started and before any servers
are started. Next, PDX persistence is then configured to use the disk-store,
and finally the servers are started.

Authored-by: Ken Howe <khowe@pivotal.io>

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/management/internal/cli/commands/CreateDiskStoreCommand.java
+++ b/geode-core/src/main/java/org/apache/geode/management/internal/cli/commands/CreateDiskStoreCommand.java
@@ -44,7 +44,6 @@ import org.apache.geode.management.internal.cli.functions.CliFunctionResult;
 import org.apache.geode.management.internal.cli.functions.CreateDiskStoreFunction;
 import org.apache.geode.management.internal.cli.functions.ListDiskStoresFunction;
 import org.apache.geode.management.internal.cli.i18n.CliStrings;
-import org.apache.geode.management.internal.cli.result.model.InfoResultModel;
 import org.apache.geode.management.internal.cli.result.model.ResultModel;
 import org.apache.geode.management.internal.security.ResourceOperation;
 import org.apache.geode.security.ResourcePermission;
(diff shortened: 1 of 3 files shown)


geode-core/src/main/java/org/apache/geode/management/internal/cli/commands/CreateDiskStoreCommand.java
```diff
@@ -122,7 +126,11 @@ public ResultModel createDiskStore(
     Set<DistributedMember> targetMembers = findMembers(groups, null);
 
     if (targetMembers.isEmpty()) {
-      return ResultModel.createError(CliStrings.NO_MEMBERS_FOUND_MESSAGE);
+      if (!stageDiskStoreConfig) {
```
jinmeiliao: If this "allowing configuration to be created before any server is started" feature is needed, we should be bold about changing this old behavior of returning an error when no server found. No matter this flag is there or not, we should just go ahead and create the configuration in cc.

metatype:
@pdxrunner Could you accomplish the same thing w/o adding a new option?  Happens if you invoke `create disk-store` without any servers online?

How does the user know that configuration changes have been `staged` but not `applied`?  I think this introduces a new state into the system that the user needs to be aware of.  Should we warn the user:  "You have made configuration changes that will not be applied until the cluster members have been restarted."

And then:  How does a user know which members have the change applied vs staged?

dschneider-pivotal:
@metatype currently the create disk-store command with no servers online will fail. I never liked that behavior. I think the rational for it was to make sure the create disk-store would actually work on a real server instead of just putting the info in cluster config and then finding out later when you start a server that some of the disk-store config does not work.
The warning you suggest is much like one we have on the configure pdx command.

dschneider-pivotal:
@pdxrunner in the use case you describe you say the new option is to allow a disk-store to be created "prior to any server members have joined the cluster". Is this all that is needed? Or do you also need to support creating a disk-store, just in cluster config, when servers have already joined?

pdxrunner:
> @pdxrunner in the use case you describe you say the new option is to allow a disk-store to be created "prior to any server members have joined the cluster". Is this all that is needed? Or do you also need to support creating a disk-store, just in cluster config, when servers have already joined?

No, I don't see a situation where that would be needed. The end result is to have the disk-store created on all servers. After discussing with @metatype  and @jinmeiliao I will rework this PR to provide the behavior that creating a disk-store when there are no servers will not error and will create the configuration.

pdxrunner:
> What about documenting this new gfsh option in the User Guide?
> If you don't plan to include doc additions in this PR, then please create a subtask in the JIRA system for the doc work so it doesn't get overlooked.

As of the second commit, there's no change to the command options. I don't think any changes are needed in the docs that I reviewed.

pdxrunner:
Closing this PR because the precheckin test failures show that the case of {{create disk-store --group=xxxx ...}} when there are no members in group `xxxx` is not being handled. I will re-work validation of disk-store attributes and submit a new PR
{code}
    // TODO: This condition is not correctly handling case when --groups option is specified
    if (targetMembers.isEmpty()) {
      System.out.println("DEBUG: targetMembers is empty");
     . . .
    } else {

{code}

PR #4097, opened 2019-09-26
GEODE-7249: Disable GfshInitFileIntegrationTest

kirklund:
GfshInitFileIntegrationTest fails intermittently and currently uses
reflection to alter values of static variables in Gfsh.

It needs to be disabled in CI until someone has the time and desire to
rewrite this test.


kirklund:
I merged this to develop early because it's failing for everyone.

PR #4099, opened 2019-09-27
GEODE-7248: update flaky test

mivanac:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/persistence/PersistentPartitionedRegionWithRedundancyDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/persistence/PersistentPartitionedRegionWithRedundancyDUnitTest.java
@@ -144,6 +144,7 @@ public class PersistentPartitionedRegionWithRedundancyDUnitTest implements Seria
   @Test
   public void testGetDataDelayDueToRecoveryAfterServerShutdown() throws Exception {
     int numBuckets = 5000;
+    System.setProperty("gemfire.disk.recoverValuesSync", "true");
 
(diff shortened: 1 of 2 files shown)


mivanac:
update

aaronlindsey:
Is there any way to make this test deterministic? I see that the failure condition is based on the actual latency of a get operation on a partitioned region while a server is recovering. Is there any 
deterministic technique you could use to accomplish the same purpose of making sure that a proxied get operation does not go to a server that is recovering?

mivanac:
Solution updated according to comment.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/persistence/PersistentPartitionedRegionWithRedundancyDUnitTest.java
```diff
@@ -143,14 +140,15 @@ public void recoversFromDisk() throws Exception {
    */
   @Test
   public void testGetDataDelayDueToRecoveryAfterServerShutdown() throws Exception {
-    int numBuckets = 10000;
+    int numEntries = 10000;
+    System.setProperty("gemfire.disk.recoverValuesSync", "true");
```
aaronlindsey: I'm not sure this is required — It seems to pass on my machine with and without this line. Setting system properties can be problematic because it will affect how other tests run. If you do absolutely need to set a system property here, then you should find a way to set it back to the original value after the test has completed.

mivanac: Removed setting of system property, since it is no more needed (modified TC).

PR #4102, opened 2019-09-30
GEODE-7255: Pickup Jackson CVE fix

mhansonp:
Need to pick up a fix for a CVE in Jackson. Hence rolling the version...


jmelchio:
You need to make sure that the files `assembly-context.txt` and `dependency-classpath.txt` in `geode-assembly/src/integrationTest/resources` are updated with the new version as well.

jmelchio:
This should also resolve `CVE-2019-12814` and `CVE-2019-12384` from `GEODE-7264`

mhansonp:
> You need to make sure that the files `assembly-context.txt` and `dependency-classpath.txt` in `geode-assembly/src/integrationTest/resources` are updated with the new version as well.

Done. Thanks for your guidance.

mhansonp:
Squashed to force a new commit.

PR #4108, opened 2019-10-01
GEODE-6945: remove deprecated types in v2 RegionType

jinmeiliao:
* experimental api should not have deprecated apis.


jinmeiliao:
@jmelchio

jinmeiliao:
it will be either "LEGACY" or one of the supported types in v2 with redundancy level set.

PR #4110, opened 2019-10-02
GEODE-7258: Pulling the failing test while fixing

mhansonp:
GEODE-7258: Pulling a test while it is fixed.


geode-lucene/src/upgradeTest/java/org/apache/geode/cache/lucene/RollingUpgradeQueryReturnsCorrectResultsAfterClientAndServersAreRolledOverAllBucketsCreated.java
```diff
@@ -31,7 +29,7 @@
 public class RollingUpgradeQueryReturnsCorrectResultsAfterClientAndServersAreRolledOverAllBucketsCreated
     extends LuceneSearchWithRollingUpgradeDUnit {
 
-  @Test
+  // @Test
```
kirklund: Actually, add `@Ignore` and just leave `@Test` in place. You co do something like:
```
@Ignore("Disabled until GEODE-7258 is fixed")
```
And then add a note to GEODE-7258 to re-enable this test.

mhansonp: Done.

PR #4111, opened 2019-10-02
GEODE-7227: ClassName now has a default constructor allowing it to be serialized

dschneider-pivotal:
Changed PdxManagementTest to include a ClassName attribute which caused the test to fail.
Added default constructor to ClassName so that it can be serialized and test now passes.

### For all changes:

### Note:

### DIFF
--- a/geode-management/src/main/java/org/apache/geode/management/configuration/ClassName.java
+++ b/geode-management/src/main/java/org/apache/geode/management/configuration/ClassName.java
@@ -43,6 +43,11 @@ public class ClassName implements Serializable {
   public static ClassName EMPTY = new ClassName("");
 
   /**
+   * Default constructor used for serialization.
+   */
+  public ClassName() {}
+
+  /**
    * Object to be instantiated using the empty param constructor of the className
    *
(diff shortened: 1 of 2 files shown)


dschneider-pivotal:
@jmelchio

geode-management/src/main/java/org/apache/geode/management/configuration/ClassName.java
```diff
@@ -42,6 +42,11 @@
   // e.g. alter region --name=regionA --cache-loader=''
   public static ClassName EMPTY = new ClassName("");
 
+  /**
+   * Default constructor used for serialization.
+   */
+  public ClassName() {}
```
jinmeiliao: so, we can add this without adding setters?

dschneider-pivotal: We have other constructors if they want to specify a class name. And the javadocs on this one states it is for serialization. I would not be against us changing the ClassName impl to be more like our other classes (which user settors) but that would not be part of this pull request.

PR #4116, opened 2019-10-02
GEODE-6988: Implement JavaBeanAccessorMethodAuthorizer

DonalEvans:
- Made the class final, immutable and thread safe.
- Added comprehensive javadocs to the class and its methods.
- Added several unit tests for the class and all public methods.

### For all changes:

### Note:


geode-core/src/main/java/org/apache/geode/cache/query/security/JavaBeanAccessorMethodAuthorizer.java
```diff
@@ -0,0 +1,153 @@
...
+ * <p/>
+ *
+ * The {@code Region Entry Modification} security risk still exists: users with the
+ * {@code DATA:READ:RegionName} privilege will be able to execute ANY method (even mutators) on the
+ * objects stored within the region and on instances used as bind parameters of the OQL, so this
+ * authorizer implementation must be used with extreme care.
+ * <p/>
+ *
+ * Usage of this authorizer implementation is only recommended for secured clusters on which only
+ * trusted users and applications have access to the OQL engine. It might also be used on clusters
```
jujoramos: The sentence contradicts what we have in the [RFC](https://cwiki.apache.org/confluence/display/GEODE/OQL+Method+Invocation+Security#OQLMethodInvocationSecurity-JavaBeanAccessorMethodAuthorizer). This authorizer is actually intended/recommended to be used on secured clusters on which the operator has fully confidence in that only objects following the _JavaBean specification 1.01_ are stored into regions and used as OQL bind parameters.

geode-core/src/main/java/org/apache/geode/cache/query/security/JavaBeanAccessorMethodAuthorizer.java
```diff
@@ -0,0 +1,153 @@
...
+ * on which the entries stored are immutable.
+ * <p/>
+ *
+ * @see org.apache.geode.cache.Cache
+ * @see org.apache.geode.cache.query.security.MethodInvocationAuthorizer
+ * @see org.apache.geode.cache.query.security.RestrictedMethodAuthorizer
+ */
+
+public class JavaBeanAccessorMethodAuthorizer implements MethodInvocationAuthorizer {
```
jujoramos: Make the class `final` so it can't be inherited.

jujoramos:
@DonalEvans: in order to avoid the `AssemblyContentsIntegrationTest` failure, you'll need to add `javadoc/org/apache/geode/cache/query/security/JavaBeanAccessorMethodAuthorizer.html` to the `geode/geode-assembly/src/integrationTest/resources/assembly_content.txt` file (around line 495).

geode-core/src/main/java/org/apache/geode/cache/query/security/JavaBeanAccessorMethodAuthorizer.java
```diff
@@ -0,0 +1,153 @@
...
+ * <p/>
+ *
+ * Some known dangerous methods, like {@link Object#getClass()}, are also rejected by this
+ * authorizer implementation (see
+ * {@link RestrictedMethodAuthorizer#isKnownDangerousMethod(Method, Object)}).
+ * <p/>
+ *
+ * This authorizer implementation addresses only three of the four known security risks:
+ * {@code Java Reflection}, {@code Cache Modification} and {@code Region Modification}.
+ * <p/>
```
jujoramos: I'd also add a comment stating that, when used correctly (domain model follows the `JavaBean specification 1.01`), this authorizer address the four security risks. If there are accessors method that can actually modify the object instance, then only 3 out of 4 risks are addressed.

geode-core/src/main/java/org/apache/geode/cache/query/security/JavaBeanAccessorMethodAuthorizer.java
```diff
@@ -0,0 +1,153 @@
...
+    boolean matches = false;
+
+    if ((methodName.startsWith("get") || methodName.startsWith("is"))) {
+      Iterator<String> iterator = this.allowedPackages.iterator();
+      while (iterator.hasNext() && !matches) {
+        matches = iterator.next().startsWith(packageName);
+      }
+    }
+    // Return true if we found a match, otherwise, delegate to the default authorizer.
+    return matches || restrictedMethodAuthorizer.authorize(method, target);
```
jujoramos: I've just thought about this scenario while implementing the `RegExMethodAuthorizer`...
What if the user sets `org.apache.geode` as one of the allowed packages for `JavaBeanAccessorMethodAuthorizer`, `matches` will be `true` at this point **BUT** the actual method is not considered safe by Geode (`restrictedMethodAuthorizer.isAllowedGeodeMethod(method, target) = false`)?. With the current implementation the actual return value would be `true`, but the actual result should be `false`... we should either reject any package that starts with `org.apache.geode` during the initialization of the authorizer, or make sure we call `restrictedMethodAuthorizer.isAllowedGeodeMethod(method, target)` when needed.

geode-core/src/main/java/org/apache/geode/cache/query/security/JavaBeanAccessorMethodAuthorizer.java
```diff
@@ -0,0 +1,153 @@
...
+ * This authorizer implementation addresses only three of the four known security risks:
+ * {@code Java Reflection}, {@code Cache Modification} and {@code Region Modification}.
+ * <p/>
+ *
+ * The {@code Region Entry Modification} security risk still exists: users with the
+ * {@code DATA:READ:RegionName} privilege will be able to execute ANY method (even mutators) on the
+ * objects stored within the region and on instances used as bind parameters of the OQL, so this
+ * authorizer implementation must be used with extreme care.
+ * <p/>
+ *
```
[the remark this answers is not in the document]
DonalEvans: Should be fixed in the latest commit

geode-core/src/main/java/org/apache/geode/cache/query/security/JavaBeanAccessorMethodAuthorizer.java
DonalEvans: Should be fixed in the latest commit

geode-core/src/main/java/org/apache/geode/cache/query/security/JavaBeanAccessorMethodAuthorizer.java
DonalEvans: Done

geode-core/src/main/java/org/apache/geode/cache/query/security/JavaBeanAccessorMethodAuthorizer.java
DonalEvans: Should be fixed in the latest commit

geode-core/src/main/java/org/apache/geode/cache/query/security/JavaBeanAccessorMethodAuthorizer.java
DonalEvans: Added changes to address this in the latest commit

geode-core/src/main/java/org/apache/geode/cache/query/security/JavaBeanAccessorMethodAuthorizer.java
```diff
@@ -0,0 +1,173 @@
...
+   *
+   * @see org.apache.geode.cache.query.security.MethodInvocationAuthorizer
+   */
+  @Override
+  public boolean authorize(Method method, Object target) {
+
+    // First ensure that the given method exists for the given target
+    if (!Arrays.asList(target.getClass().getMethods()).contains(method)) {
+      return false;
+    }
```
jujoramos: Even though is not a bad idea to have this, is not needed and will just incur into a waste time during the query execution. The query engine is responsible for invoking the `MethodInvocationAuthorizer` and, internally, it always gets the actual `method` instance from the actual `target` object through the usage of reflection, so the condition will always evaluate to `true`.

geode-core/src/main/java/org/apache/geode/cache/query/security/JavaBeanAccessorMethodAuthorizer.java
```diff
@@ -0,0 +1,173 @@
...
+    Objects.requireNonNull(allowedPackages, NULL_PACKAGE_MESSAGE);
+    this.restrictedMethodAuthorizer = restrictedMethodAuthorizer;
+    this.allowedPackages = Collections.unmodifiableSet(allowedPackages);
+  }
+
+  /**
+   * Executes the authorization logic to determine whether the {@code method} is allowed to be
+   * executed on the {@code target} object instance.
+   * <p/>
+   *
```
jujoramos: I'd add the following paragraph to the javadoc, mainly to keep consistency with the other authorizers and to make sure the user is aware of this fact:

```
* If the {@code target} object is an instance of {@link Region}, this methods also ensures that
* the user has the {@code DATA:READ} permission granted for the target {@link Region}.
```

geode-core/src/main/java/org/apache/geode/cache/query/security/JavaBeanAccessorMethodAuthorizer.java
```diff
@@ -0,0 +1,173 @@
...
+    if (packageName.startsWith(GEODE_BASE_PACKAGE)) {
+      return restrictedMethodAuthorizer.isAllowedGeodeMethod(method, target);
+    }
+
+    String methodName = method.getName();
+
+    // Return true if the the object belongs to an allowed package and the method starts with
+    // exactly 'get' or 'is' followed by a non-lowercase character (to prevent matching with
+    // methods that might have names like 'getawayDate' or 'islandName' etc.)
+    if (Pattern.matches("^(get|is)($|[^A-Z])+", methodName)) {
```
jujoramos: Great catch regarding non accessor methods starting with `is|get`!!.
One minor change, though, `Pattern.matches` internally compiles the regular expression every single time, resulting in a waste of processing time while evaluating the OQL:

```
public static boolean matches(String regex, CharSequence input) {
  Pattern p = Pattern.compile(regex);
  Matcher m = p.matcher(input);
  return m.matches();
}
```

Instead of doing that, I'd suggest to store and reuse the compiled pattern within a `private final static` field, something like below:
```
Pattern pattern = Pattern.compile("^(get|is)($|[^A-Z])+");
pattern.matcher(methodName).matches();
```

PR #4123, opened 2019-10-04
GEODE-6987: Implement RegExMethodAuthorizer

jujoramos:
- Made the class final, immutable and thread safe.
- Added comprehensive javadocs to the class and its methods.
- Added several unit tests for the class and all public methods.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/cache/query/security/RegExMethodAuthorizer.java
+++ b/geode-core/src/main/java/org/apache/geode/cache/query/security/RegExMethodAuthorizer.java
@@ -1,0 +1,167 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.apache.geode.cache.query.security;
+
+import java.lang.reflect.Method;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Objects;
+import java.util.Properties;
+import java.util.Set;
+import java.util.regex.Pattern;
+
+import org.apache.geode.cache.Cache;
+import org.apache.geode.cache.Declarable;
+import org.apache.geode.cache.Region;
+
+/**
+ * An immutable and thread-safe {@link MethodInvocationAuthorizer} that only allows the execution of
+ * those methods matching the configured regular expression.
+ * <p/>
+ *
+ * Some known dangerous methods, like {@link Object#getClass()}, are also rejected by this
+ * authorizer implementation, no matter whether the method matches the configured regular
+ * expressions
+ * or not (see {@link RestrictedMethodAuthorizer#isKnownDangerousMethod(Method, Object)}).
+ * <p/>
+ *
+ * When correctly configured, this authorizer implementation addresses the four known security
+ * risks: {@code Java Reflection}, {@code Cache Modification}, {@code Region Modification} and
+ * {@code Region Entry Modification}.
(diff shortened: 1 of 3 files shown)


jujoramos:
@DonalEvans: since you're also implementing another autorizer, can I have your review on this one please?.

geode-core/src/main/java/org/apache/geode/cache/query/security/RegExMethodAuthorizer.java
```diff
@@ -0,0 +1,167 @@
...
+      return restrictedMethodAuthorizer.isAllowedGeodeMethod(method, target);
+    }
+
+    // Compare fully qualified method name against compiled expressions.
+    String fullyQualifiedMethodName = target.getClass().getName() + "." + method.getName();
+    boolean match = compiledPatterns.stream()
+        .anyMatch(pattern -> pattern.matcher(fullyQualifiedMethodName).matches());
+
+    // Return if there was a match, otherwise delegate to the default authorizer.
+    return match || restrictedMethodAuthorizer.authorize(method, target);
```
DonalEvans: Just a small concern, for clarity it might be better to have each criteria that might be met (dangerous method, geode method, matching regex and default authorizer) have a separate return statement, rather than combining the last two into an or statement. Something like:

```
    if(compiledPatterns.stream().anyMatch(pattern -> pattern.matcher(fullyQualifiedMethodName).matches())) {
        return true
    }
    
    return restrictedMethodAuthorizer.authorize(method, target);
```

geode-core/src/test/java/org/apache/geode/cache/query/security/RegExMethodAuthorizerTest.java
```diff
@@ -0,0 +1,289 @@
...
+
+    assertThatThrownBy(
+        () -> new RegExMethodAuthorizer(new RestrictedMethodAuthorizer(mockCache), null))
+            .isInstanceOf(NullPointerException.class)
+            .hasMessage(NULL_REGULAR_EXPRESSIONS_MESSAGE);
+  }
+
+  @Test
+  public void allowedPatternsShouldBeAccessibleBuNotModifiable() {
+    assertThatThrownBy(() -> geodeRegExMatchingMethodAuthorizer.getAllowedPatterns().remove(".*"))
```
DonalEvans: The pattern attempting to be removed doesn't appear to match the pattern that was used to initialize the Authorizer on line 76. Perhaps it should be extracted to a field if it's going to be referenced in this method too.

geode-core/src/main/java/org/apache/geode/cache/query/security/RegExMethodAuthorizer.java
jujoramos: I'll change it, thanks!.

geode-core/src/test/java/org/apache/geode/cache/query/security/RegExMethodAuthorizerTest.java
jujoramos: The method is just verifying that the original set can no be modified (`remove` throws an `UnsupportedOperationException`, no matter whether the collection contains the element or not). Trying to remove something that doesn't even exist might confuse the reader as you say, though, so I'll change the method to try to to add something to the set instead of removing it (the result will be the same anyway -> the collection can't be modified).

geode-core/src/main/java/org/apache/geode/cache/query/security/RegExMethodAuthorizer.java
```diff
@@ -0,0 +1,169 @@
...
+   * @param target the {@link Object} on which the {@link Method} will be executed.
+   * @return {@code true} if the {@code method} can be executed on on the {@code target} instance,
+   *         {@code false} otherwise.
+   *
+   * @see org.apache.geode.cache.query.security.MethodInvocationAuthorizer
+   */
+  @Override
+  public boolean authorize(Method method, Object target) {
+    // Return false for known dangerous methods.
+    if (restrictedMethodAuthorizer.isKnownDangerousMethod(method, target)) {
```
nabarunnag: Do you feel like changing the wording "dangerous" to something like isAlwaysRestrictedMethod() or isPermanentlyRestrictedMethod() will be more informative?

jujoramos: Hey @nabarunnag, thanks for reviewing this 👍 .
Regarding the method name change, it's of course doable but not as part of this PR: `isKnownDangerousMethod` belongs to the `RestrictedMethodAuthorizer` class and it's also used by the other 3 authorizers, so it's better to create a new GEODE JIRA for that.

PR #4124, opened 2019-10-04
GEODE-3780: suspected member is never watched again after passing fina…

bschuchardt:
…l check

This restores our original behavior (pre-1.0) behavior of performing a
final check on a member if UDP communications with that member fail.

We now also send exonoration messages to all other members if a suspect is
cleared.  We need to do that because another member may have sent a
Suspect message that was ignored because the suspect was already
undergoing a final check.

I also noticed that our tcp/ip final check loop was performing more than
one check in many cases because the nanosecond clock has a coarse
granularity.  A socket so-timeout based on the millisecond clock was
timing out but the nanosecond clock didn't line up with that timeout and
caused the "for" loop to make another attempt.  I changed that loop to
convert the nanosecond clock value to milliseconds.

### For all changes:

### Note:


echobravopapa:
looks like some JDK11 test failures...

PR #4128, opened 2019-10-04
GEODE-7237: Use invalid host name in ConnectCommandAcceptanceTest

kirklund:
The test invalidHostname continues to fail intermittently. In the
failure, something fails to throw UnknownHostException. Having
spaces in the "invalid host name" should guarantee that it always
throws UnknownHostException now.

Also, remove the double negative from the assumeTrue by flipping it
to assumeFalse. Move the comment into the assumeFalse as a message.

Please review: @aaronlindsey


kirklund:
Timeout exceeded on DistributedTestOpenJDK11.

kirklund:
UpgradeTestOpenJDK11 failed in WANRollingUpgradeSecondaryEventsNotReprocessedAfterOldSiteMemberFailover with BindException. This is unrelated to the PR.

PR #4130, opened 2019-10-08
GEODE-7131: improve HATEOAS discoverability

onichols-pivotal:
We have lots of endpoints of the form `http://locator:7070/management/v1/something`. If you don't know the right "something" or you wonder what other somethings might be available, a user might try to discover them by shortening a known url to `http://locator:7070/management/v1/` in hopes of browsing other available APIs (for example, this works on GitHub: `curl https://api.github.com/` return a list of all top-level api endpoints).

Does this seem like a desirable feature for Geode's Management REST API?  If not, I'm fine with a `-1` for this PR.

Before:
```
curl http://locator:7070/management/v1/
404 Not Found
```

After:
```
curl http://locator:7070/management/v1/
{
  "statusCode": "OK",
  "links": {
    "swagger": "http://localhost:7070/management/swagger-ui.html",
    "docs": "https://geode.apache.org/docs",
    "wiki": "https://cwiki.apache.org/confluence/display/pages/viewpage.action?pageId=115511910",
    "get gateway-receiver": "http://localhost:7070/management/v1/gateways/receivers/{id}",
    "create gateway-receiver": "http://localhost:7070/management/v1/gateways/receivers",
    "list gateway-receivers": "http://localhost:7070/management/v1/gateways/receivers",
    "get member": "http://localhost:7070/management/v1/members/{id}",
    "list members": "http://localhost:7070/management/v1/members",
    "configure pdx": "http://localhost:7070/management/v1/configurations/pdx",
    "ping": "http://localhost:7070/management/v1/ping",
    "list rebalances": "http://localhost:7070/management/v1/operations/rebalances",
    "check rebalance": "http://localhost:7070/management/v1/operations/rebalances/{id}",
    "start rebalance": "http://localhost:7070/management/v1/operations/rebalances",
    "get region": "http://localhost:7070/management/v1/regions/{id}",
    "get index": "http://localhost:7070/management/v1/regions/{regionName}/indexes/{id}",
    "create region": "http://localhost:7070/management/v1/regions",
    "list regions": "http://localhost:7070/management/v1/regions",
    "list region indexes": "http://localhost:7070/management/v1/regions/{regionName}/indexes",
    "delete region": "http://localhost:7070/management/v1/regions/{id}",
    "list indexes": "http://localhost:7070/management/v1/indexes",
    "api root": "http://localhost:7070/management/v1/"
  }
}
```


dschneider-pivotal:
The invalidHostname failure in AcceptanceTestOpenJDK11 is the know issue: GEODE-7237

dschneider-pivotal:
The IntegrationTestOpenJDK11 failure was caused by Links being moved to an internal package. So we need to fix that test.

geode-management/src/main/java/org/apache/geode/management/api/RestfulEndpoint.java
```diff
@@ -0,0 +1,84 @@
...
+   * Returns the full URI path that uniquely identifies this instance.
+   * If the id is null, then null is returned.
+   *
+   * <p>
+   * Note that the result does not include the prefix: <code>http://hostname:port</code>
+   *
+   * @return {@link #URI_CONTEXT} + {@link #URI_VERSION} + {@link #getIdentityEndpoint()}
+   */
+  @JsonIgnore()
+  default String getUri() {
```
jinmeiliao: we should get rid of this method, since there is only one usage of it in `Links` and we can add the URI_CONTEXT and URI_VERSION like we do in other methods there.

geode-management/src/main/java/org/apache/geode/management/api/ClusterManagementResult.java
```diff
@@ -131,15 +148,35 @@ public String getStatusMessage() {
...
    * via REST
    */
+  @JsonIgnore
   public String getUri() {
-    return uri;
+    return (links == null) ? null : links.get(Links.SELF);
+  }
+
+  /**
+   * for internal use only
```
jinmeiliao: I prefer we would not have so many "for internal use only" methods. I can understand we would have a few for json serialization or  convenience method. For this story, do we need to have a top level links?

geode-management/src/main/java/org/apache/geode/management/internal/Links.java
```diff
@@ -0,0 +1,61 @@
...
+    links.putIfAbsent(API_ROOT, URI_CONTEXT + URI_VERSION + "/");
+  }
+
+  /**
+   * Enriches a {@link LinkedHashMap} with URI links to documentation that assists in the
+   * understanding of Geode management API
+   *
+   * The keys in the map are {@link #SWAGGER}, {@link #DOCS} and {@link #WIKI}
+   */
+  public static LinkedHashMap<String, String> rootLinks() {
```
jinmeiliao: since this is not part of the story, let's not complicate this story for now.

geode-web-management/src/main/java/org/apache/geode/management/internal/rest/controllers/ManagementControllerAdvice.java
```diff
@@ -91,6 +102,30 @@ public static String removeClassFromJsonText(String json) {
...
+      if (pos >= 0) {
+        // for MockMVC tests the URI_CONTEXT will be absent, but tests expect it so fake it
+        return requestUrl.substring(0, pos) + URI_CONTEXT;
+      } else {
+        return null;
+      }
+    }
+  }
+
+  static String qualifyHrefsInJsonText(String json, String requestUrl) {
```
jinmeiliao: if we are manipulating the final links here, maybe we should have all the apis return the links only with the URI_VERSION + uri, and in here add the host and context.

geode-web-management/src/main/java/org/apache/geode/management/internal/rest/controllers/RegionManagementController.java
```diff
@@ -165,6 +167,37 @@ public ClusterManagementResult deleteRegion(
...
+  private static void addBiDirectionalUpLink(Map<String, String> links) {
+    String parent = links.get(SELF).replaceFirst("/[^/]*/[^/*]*$", "");
+    String type = parent.replaceFirst("s/[^/]*$", "").replaceFirst(".*/", "");
+    links.put(type, parent);
+  }
+
+  private static ClusterManagementListResult<Index, RuntimeInfo> addUplinks(
+      ClusterManagementListResult<Index, RuntimeInfo> list) {
+    list.getResult().forEach(r -> addBiDirectionalUpLink(r.getLinks()));
+    return list;
```
jinmeiliao: these kind of operations somehow don't feel right. I thought it should be some interfaces that we just need to implement in order to add the links, not manipulate each and every end point. Have we explored other options to do this?

geode-web-management/src/main/java/org/apache/geode/management/internal/rest/controllers/RootManagementController.java
```diff
@@ -0,0 +1,79 @@
...
+import org.springframework.web.bind.annotation.RestController;
+import org.springframework.web.servlet.mvc.method.annotation.RequestMappingHandlerMapping;
+
+import org.apache.geode.internal.logging.LogService;
+import org.apache.geode.management.api.ClusterManagementResult;
+import org.apache.geode.management.internal.Links;
+
+@RestController("api-root")
+@RequestMapping(URI_VERSION)
+public class RootManagementController extends AbstractManagementController {
```
jinmeiliao: In order to not complicate this PR, let's rip out all the root links first.

geode-management/src/main/java/org/apache/geode/management/configuration/AbstractConfiguration.java
```diff
@@ -86,62 +84,9 @@ public String getGroup() {
...
-
-  /**
-   * Returns the portion of the URI, after {@link #URI_VERSION}, that specifies the type of this
-   * instance.
-   * It is possible that more than once instance of this type can exist.
-   *
-   * @return the portion of the URI that identifies the type of this instance
-   */
   @JsonIgnore
   public abstract String getEndpoint();
```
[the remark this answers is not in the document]
onichols-pivotal: Ideally yes, but we found that it doesn't compile if this line is remove.  Seems to be a bug/limitation of the Java 8 compiler...

geode-management/src/main/java/org/apache/geode/management/api/RestfulEndpoint.java
onichols-pivotal: now that it is not required in Json output, that seems reasonable

geode-management/src/main/java/org/apache/geode/management/api/ClusterManagementResult.java
onichols-pivotal: these seem like 2 unrelated issues.  public internal methods can be solved by changing our package structure or by moving to Jigsaw modules, otherwise I don't see this as any worse than countless other places.  If you want this split into multiple PRs, that's fine but should have nothing to do with internal method scoping...

geode-management/src/main/java/org/apache/geode/management/internal/Links.java
onichols-pivotal: So you want this split into multiple PRs?

geode-web-management/src/main/java/org/apache/geode/management/internal/rest/controllers/ManagementControllerAdvice.java
onichols-pivotal: it doesn't leave as much to match on, but might be possible to do it that way

geode-web-management/src/main/java/org/apache/geode/management/internal/rest/controllers/RegionManagementController.java
onichols-pivotal: do you want to rework this as part of this PR or a subsequent PR?

geode-web-management/src/main/java/org/apache/geode/management/internal/rest/controllers/RootManagementController.java
onichols-pivotal: when you say "rip out" do you mean split into two PRs, or do you mean get rid of forever?

jinmeiliao: I mean if we don't have this root link requirement, the implementation might be easier. But if you didn't think it would make things easier, you can leave them here.

geode-web-management/src/main/java/org/apache/geode/management/internal/rest/controllers/RegionManagementController.java
jinmeiliao: what is not clear to me is where do we decide if we want to add the index link to the region result?

geode-management/src/main/java/org/apache/geode/management/configuration/AbstractConfiguration.java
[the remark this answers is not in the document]
onichols-pivotal: PR was split so this is no longer relevant

geode-management/src/main/java/org/apache/geode/management/api/RestfulEndpoint.java
onichols-pivotal: PR was split so this is no longer relevant

geode-management/src/main/java/org/apache/geode/management/api/ClusterManagementResult.java
onichols-pivotal: PR was split so this is no longer relevant

geode-management/src/main/java/org/apache/geode/management/internal/Links.java
onichols-pivotal: PR was split and Links was implemented in the other PR instead of this one

geode-web-management/src/main/java/org/apache/geode/management/internal/rest/controllers/ManagementControllerAdvice.java
onichols-pivotal: PR was split so this is no longer relevant

geode-web-management/src/main/java/org/apache/geode/management/internal/rest/controllers/RegionManagementController.java
onichols-pivotal: PR was split so this is no longer relevant

geode-web-management/src/main/java/org/apache/geode/management/internal/rest/controllers/RootManagementController.java
onichols-pivotal: PR was split, so this PR is now *just* the root controller

onichols-pivotal:
The only purpose of HATEOAS is to make endpoint discovery easier (for a human, generally, although the dream was for automated discovery).  It's been suggested to me that a simpler way to accomplish that might just be to redirect `/` to `/docs`.  That should work just as well for humans, but then why are we doing any HATEOAS at all...?

onichols-pivotal:
It's also been suggested that `/` may require authentication -- only `/docs` and a few select paths are explicitly configured as public.  Before accepting this PR, it would likely be a good idea to make `/` public...

onichols-pivotal:
Given the lack of interest for this feature, I am closing this PR without merging.

PR #4132, opened 2019-10-08
GEODE-7279: Warning log for deprecated system value should be shown once

DonalEvans:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/DefaultQueryServiceDeprecationTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/DefaultQueryServiceDeprecationTest.java
@@ -1,0 +1,74 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.apache.geode.cache.query.dunit;
+
+import static org.apache.geode.distributed.internal.DistributionConfig.GEMFIRE_PREFIX;
+
+import java.io.File;
+import java.io.Serializable;
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+
+import org.apache.geode.cache.query.internal.DefaultQueryService;
+import org.apache.geode.test.assertj.LogFileAssert;
+import org.apache.geode.test.dunit.rules.ClusterStartupRule;
+import org.apache.geode.test.dunit.rules.MemberVM;
+
+public class DefaultQueryServiceDeprecationTest implements Serializable {
+  MemberVM locator;
+
+  @Rule
+  public ClusterStartupRule cluster = new ClusterStartupRule();
+
+  @Before
+  public void startUp() {
+    locator = cluster.startLocatorVM(0);
+  }
+
+  @After
+  public void shutDown() {
+    cluster.stop(0);
+  }
+
+  @Test
+  public void warningMessageIsOnlyLoggedOnceWhenDeprecatedPropertyUsed() {
+    MemberVM server = cluster.startServerVM(1,
+        x -> x.withConnectionToLocator(locator.getPort()).withLogFile().withSystemProperty(
+            GEMFIRE_PREFIX + "QueryService.allowUntrustedMethodInvocation", "true"));
+
+    server.invoke(() -> {
+      File logFile = new File(server.getName() + ".log");
+      ClusterStartupRule.getCache().getQueryService();
+      ClusterStartupRule.getCache().getQueryService();
+      LogFileAssert.assertThat(logFile).containsOnlyOnce(DefaultQueryService.DEPRECIATION_WARNING);
+    });
+    server.getVM().bounce();
+  }
+
+  @Test
+  public void warningMessageIsNotLoggedWhenDeprecatedPropertyIsNotUsed() {
+    MemberVM server =
+        cluster.startServerVM(1, x -> x.withConnectionToLocator(locator.getPort()).withLogFile());
+    server.invoke(() -> {
+      File logFile = new File(server.getName() + ".log");
+      ClusterStartupRule.getCache().getQueryService();
+      LogFileAssert.assertThat(logFile).doesNotContain(DefaultQueryService.DEPRECIATION_WARNING);
+    });
+    server.getVM().bounce();
+  }
+}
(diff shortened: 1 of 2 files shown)


geode-core/src/main/java/org/apache/geode/cache/query/internal/DefaultQueryService.java
```diff
@@ -104,10 +104,19 @@
...
+  public static final boolean ALLOW_UNTRUSTED_METHOD_INVOCATION;
+
+  static {
+    String deprecatedValue =
+        System.getProperty(GEMFIRE_PREFIX + "QueryService.allowUntrustedMethodInvocation");
+    if (deprecatedValue != null) {
+      logger.warn("The property " + GEMFIRE_PREFIX +
+          "QueryService.allowUntrustedMethodInvocation is deprecated. To provide the same " +
+          "functionality, please use the UnrestrictedMethodAuthorizer implementation of" +
+          "MethodInvocationAuthorizer");
```
jujoramos: There's a missing space between `of` and `MethodInvocationAuthorizer` here.
Also, and sorry for nitpicking here, I think it would be better if the message reads something like _X is deprecated, please use Y instead_.

DonalEvans: Done

geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/DefaultQueryServiceDeprecationTest.java
```diff
@@ -0,0 +1,74 @@
...
+
+  @Before
+  public void startUp() {
+    locator = cluster.startLocatorVM(0);
+  }
+
+  @After
+  public void shutDown() {
+    cluster.stop(0);
+  }
```
jujoramos: I don't think this is needed, the `ClusterStartupRule` already stops every started member and restores the system properties after each test method. See [here](https://github.com/apache/geode/blob/develop/geode-dunit/src/main/java/org/apache/geode/test/dunit/rules/ClusterStartupRule.java#L159-L186) for more details.

geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/DefaultQueryServiceDeprecationTest.java
```diff
@@ -0,0 +1,74 @@
...
+        x -> x.withConnectionToLocator(locator.getPort()).withLogFile().withSystemProperty(
+            GEMFIRE_PREFIX + "QueryService.allowUntrustedMethodInvocation", "true"));
+
+    server.invoke(() -> {
+      File logFile = new File(server.getName() + ".log");
+      ClusterStartupRule.getCache().getQueryService();
+      ClusterStartupRule.getCache().getQueryService();
+      LogFileAssert.assertThat(logFile).containsOnlyOnce(DefaultQueryService.DEPRECIATION_WARNING);
+    });
+    server.getVM().bounce();
```
jujoramos: Just out of curiosity, why do we need to bounce the server VM?.

geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/DefaultQueryServiceDeprecationTest.java
```diff
@@ -0,0 +1,74 @@
...
+  public void warningMessageIsOnlyLoggedOnceWhenDeprecatedPropertyUsed() {
+    MemberVM server = cluster.startServerVM(1,
+        x -> x.withConnectionToLocator(locator.getPort()).withLogFile().withSystemProperty(
+            GEMFIRE_PREFIX + "QueryService.allowUntrustedMethodInvocation", "true"));
+
+    server.invoke(() -> {
+      File logFile = new File(server.getName() + ".log");
+      ClusterStartupRule.getCache().getQueryService();
+      ClusterStartupRule.getCache().getQueryService();
+      LogFileAssert.assertThat(logFile).containsOnlyOnce(DefaultQueryService.DEPRECIATION_WARNING);
```
jujoramos: Extra nitpick: wouldn't be better to use the `TemporaryFolder` rule here instead and, while starting the server, set your custom log file directly instead of relying entirely on the `ClusterStartupRule` (if the default log file name used by the rule changes, the test will break)?.
Something like below:

```
    File logFile = folderRule.newFile("myCustomLogFile.log");

    MemberVM server = cluster.startServerVM(1,
        x -> x.withConnectionToLocator(locator.getPort())
            .withSystemProperty(GEMFIRE_PREFIX + "QueryService.allowUntrustedMethodInvocation", "true")
            .withProperty("log-file", logFile.getAbsolutePath())
```

geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/DefaultQueryServiceDeprecationTest.java
DonalEvans: If the VM isn't bounced, then on repeated test runs (during StressNewTest) the `DefaultQueryService` isn't re-initialized, causing the static block that logs the deprecation message not to be triggered and the test to fail due to not finding the deprecation warning in the logs.

geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/DefaultQueryServiceDeprecationTest.java
DonalEvans: Done

geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/DefaultQueryServiceDeprecationTest.java
DonalEvans: Done

geode-core/src/main/java/org/apache/geode/cache/query/internal/DefaultQueryService.java
```diff
@@ -104,10 +104,21 @@
    * implementation of MethodInvocationAuthorizer, which provides the same functionality.
    */
   @Deprecated
-  public static final boolean ALLOW_UNTRUSTED_METHOD_INVOCATION = Boolean.getBoolean(
-      GEMFIRE_PREFIX + "QueryService.allowUntrustedMethodInvocation");
-
-  private boolean deprecatedWarningHasBeenShown = false;
+  public static final boolean ALLOW_UNTRUSTED_METHOD_INVOCATION;
+
+  public static final String DEPRECIATION_WARNING = "The property " + GEMFIRE_PREFIX +
```
jujoramos: Small correction: the word should read `DEPRECATION` instead of `DEPRECIATION`.

geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/DefaultQueryServiceDeprecationTest.java
jujoramos: 👍

geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/DefaultQueryServiceDeprecationTest.java
```diff
@@ -0,0 +1,79 @@
...
+import org.junit.rules.TemporaryFolder;
+
+import org.apache.geode.cache.query.internal.DefaultQueryService;
+import org.apache.geode.test.assertj.LogFileAssert;
+import org.apache.geode.test.dunit.rules.ClusterStartupRule;
+import org.apache.geode.test.dunit.rules.MemberVM;
+
+public class DefaultQueryServiceDeprecationTest implements Serializable {
+
+  MemberVM locator;
```
jujoramos: Can you make this `private` (no point in keeping it accessible if there's no class inheriting from this one, and also reduces IDE warnings)?.

geode-core/src/main/java/org/apache/geode/cache/query/internal/DefaultQueryService.java
DonalEvans: Done

geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/DefaultQueryServiceDeprecationTest.java
DonalEvans: Done

PR #4135, opened 2019-10-08
GEODE-7184: Add function execution timers

aaronlindsey:
- Add micrometer timers for recording function executions
- Refactor FunctionStats and FunctionServiceStats

---

### For all changes:

- [ ] ~~If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)?~~

### Note:


geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/command/ExecuteFunction66.java
```diff
@@ -228,6 +229,7 @@ public void cmdExecute(final Message clientMessage, final ServerConnection serve
       int earlierClientReadTimeout = handshake.getClientReadTimeout();
       handshake.setClientReadTimeout(functionTimeout);
 
+      long statStartTime = stats.getTime();
```
demery-pivotal: It looks as if we're getting the start time only so that this method can update failure stats when `executeFunctionOnGroups()` and `executeFunctionLocally()` decide not to catch and handle exceptions themselves. Is there a way to push all of the exception handling into those methods, so that this method doesn't have to worry about updating the execution failure stats at all?

geode-core/src/main/java/org/apache/geode/internal/cache/MemberFunctionStreamingMessage.java
```diff
@@ -188,7 +190,7 @@ protected void process(final ClusterDistributionManager dm) {
       FunctionContextImpl context = new MultiRegionFunctionContextImpl(cache,
           this.functionObject.getId(), this.args, resultSender, regions, isReExecute);
 
-      long start = stats.startTime();
+      start = stats.getTime();
       stats.startFunctionExecution(this.functionObject.hasResult());
```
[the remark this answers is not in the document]
demery-pivotal: I don't like the inconsistent timings either.

Do we want to count the "result sender prep" work as part of function execution? If so, let's _always_ include the duration in the timer. That is, let's always get the start time and call `startFunctionExecution()` outside the `try`.

If we _don't_ want to count that prep work as part of function execution, it seems to me we also don't want to count prep failures as function execution failures.

I can think of two ways to avoid counting prep failures as function execution failures.

First:
- `boolean startedExecution = false;` before the `try` block.
- startedExecution = true` just before or after `stats.startFunctionExecution(…)`.
- Then in each catch block, record the failure only if `startedExecution` is `true`.

Second:
- Put the result sender prep work in one try/catch block, which never reports execution start and end.
- Put the actual function execution in a second try/catch block, which always reports execution start and end.

The wisdom of this second approach depends on how many catch blocks are needed for exceptions thrown by the result sender prep work.

geode-core/src/main/java/org/apache/geode/internal/cache/execute/metrics/FunctionServiceStats.java
```diff
@@ -382,22 +382,18 @@ public void startFunctionExecution(boolean haveResult) {
...
-   * @param start The start of the functionExecution (which is decremented from the current time to
-   *        determine the function Execution processing time).
+   * @param elapsed The elapsed time for function execution in nanoseconds
    * @param haveResult haveResult=true then update the _functionExecutionHasResultRunningId and
    *        _functionExecutionHasResultCompleteProcessingTimeId
    */
-  public void endFunctionExecution(long start, boolean haveResult) {
-    long ts = getTime();
-
+  public void endFunctionExecution(long elapsed, boolean haveResult) {
```
jhuynh1: Just curious, I think I saw some endFunctionExecution() method invocations that passed in start time.  Is that a different method signature?

geode-core/src/main/java/org/apache/geode/internal/cache/MemberFunctionStreamingMessage.java
```diff
@@ -188,7 +190,7 @@ protected void process(final ClusterDistributionManager dm) {
       FunctionContextImpl context = new MultiRegionFunctionContextImpl(cache,
           this.functionObject.getId(), this.args, resultSender, regions, isReExecute);
 
-      long start = stats.startTime();
+      start = stats.getTime();
```
jhuynh1: I think i'm just confused, but I think you are suggesting we can remove this getTime() call here as long as we are ok with adding the context creation etc to the total execution time?  I think I would agree.

geode-core/src/main/java/org/apache/geode/internal/cache/execute/metrics/FunctionServiceStats.java
aaronlindsey: `FunctionStats.endFunctionExecution()` takes a start time while `FunctionServiceStats.endFunctionExecution()` takes an elapsed time. The FunctionServiceStats method is only called from FunctionStats, so I think the invocations you saw were probably `FunctionStats.endFunctionExecution()`. I can see how this is confusing since they both have the same name and signature.

geode-core/src/main/java/org/apache/geode/internal/cache/MemberFunctionStreamingMessage.java
aaronlindsey: Yes, that was one of the things I/we suggested. Do you know if users would care about this change to the total execution time?

jhuynh1: Perhaps we can put a post on user and dev lists explaining the change and see if anyone objects?  I can't think of a reason other than maybe some performance test complaining somewhere (assuming it used stats).

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/command/ExecuteFunction66.java
aaronlindsey: I like the idea, but in my opinion this whole method and especially `executeFunctionLocally()` needs to be rewritten because there appear to be multiple problems with the way stats are being recorded. I can't think of a way to fix this issue without making significant changes to these methods which seems out of scope for this PR.

geode-core/src/main/java/org/apache/geode/internal/cache/execute/metrics/FunctionServiceStats.java
aaronlindsey: I renamed `FunctionServiceStats.endFunctionExecution()` to `FunctionServiceStats.endFunctionExecutionWithElapsedTime()` to make it more apparent that it takes an elapsed time as an argument.

geode-core/src/main/java/org/apache/geode/internal/cache/MemberFunctionStreamingMessage.java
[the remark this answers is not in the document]
aaronlindsey: I removed the second call to `getTime()` so now successful executions will include the time to create the context, etc in the total time. This is something I will ask about on the dev/user list.

To avoid counting prep failures as execution failures, I went with @demery-pivotal's first suggestion above.

geode-core/src/main/java/org/apache/geode/internal/cache/execute/ServerRegionFunctionExecutor.java
```diff
@@ -196,20 +198,21 @@ protected ResultCollector executeFunction(final String functionId, boolean resul
...
-    FunctionStats stats = FunctionStats.getFunctionStats(function.getId(), region.getSystem());
+    FunctionStats stats =
+        FunctionStatsManager.getFunctionStats(function.getId(), region.getSystem());
+    long start = stats.getTime();
+    stats.startFunctionExecution(true);
     try {
       validateExecution(function, null);
-      long start = stats.startTime();
-      stats.startFunctionExecution(true);
       srp.executeFunction(function, this, collector, hasResult,
```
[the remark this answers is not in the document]
aaronlindsey: I am going to ask about this on the dev/user list.

geode-core/src/main/java/org/apache/geode/internal/cache/MemberFunctionStreamingMessage.java
aaronlindsey: I am going to check with the user/dev list.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/command/ExecuteFunction66.java
demery-pivotal: > multiple problems with the way stats are being recorded

Do you mean that the stats have incorrect values?

If so, do our new meters also have those problems? If so, I think it's in scope. Or we need to renegotiate the scope.

If the meters are correct, then I agree that fixing problems in the stats is out of scope.

aaronlindsey: > multiple problems with the way stats are being recorded

For example, `executeFunctionLocally()` will schedule a function execution on an executor and then immediately call `endFunctionExecution()` even though the scheduled function execution has not necessarily been run at this point. To me this looks wrong, but I haven't been able to test this case specifically.

> Do you mean that the stats have incorrect values?

We haven't tested this code path specifically, but we have tested all of the scenarios in the JIRA ticket and those all have the correct values.

aaronlindsey: We dug into this method and it turned out that `executeFunctionLocally()` gets called when executing a function using `FunctionService.onServer()` or `FunctionService.onServers()`. In the case where the function does not return a result (`hasResult()` returns `false`), the recorded execution time will be incorrect because the code does not wait for the execution to complete before recording that the execution finished.

We fixed this issue in the latest commits and added additional acceptance tests to cover scenarios where we execute a function that does not return a result.

aaronlindsey:
> ~Please don't merge this yet — it seems there might be a case where this doesn't record the correct times for function executions without a result.~

I believe this issue is fixed. See [comment](https://github.com/apache/geode/pull/4135#discussion_r336275403).

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/command/ExecuteFunction66.java
aaronlindsey: I believe this issue is resolved, but please review the latest changes and add a review comment on the latest changes if you find any issue.

geode-assembly/src/acceptanceTest/java/org/apache/geode/metrics/function/executions/FunctionExecutionsTimerLonerTest.java
```diff
@@ -0,0 +1,364 @@
...
+    stopServer();
+  }
+
+  @Test
+  public void timersNotRegisteredIfFunctionDeployedButNotExecuted() {
+    startServerWithStatsEnabled();
+
+    deployFunction(FunctionToTimeWithResult.class);
+
+    Assertions.assertThat(getExecutionsTimerValuesFor(FunctionToTimeWithResult.ID))
```
kirklund: I recommend we use static import for `assertThat`

geode-core/src/main/java/org/apache/geode/internal/cache/execute/ServerFunctionExecutor.java
```diff
@@ -146,11 +148,11 @@ protected ResultCollector executeFunction(final Function function, long timeout,
 
   private ResultCollector executeOnServer(Function function, ResultCollector rc, byte hasResult,
       int timeoutMs) {
-    FunctionStats stats = FunctionStats.getFunctionStats(function.getId());
+    FunctionStats stats = FunctionStatsManager.getFunctionStats(function.getId());
+    long start = stats.getTime();
+    stats.startFunctionExecution(true);
```
kirklund: Some of the other stats classes combine getTime with the start method:
```
long start = stats.startFunctionExecution(true);
```
This would help streamline these lines some.

geode-assembly/src/acceptanceTest/java/org/apache/geode/metrics/function/executions/FunctionExecutionsTimerLonerTest.java
aaronlindsey: Done.

geode-core/src/main/java/org/apache/geode/internal/cache/execute/ServerFunctionExecutor.java
aaronlindsey: I removed `getTime()` in favor of returning the start time from `startFunctionExecution()`.

PR #4150, opened 2019-10-11
GEODE-7189: Disable default server in startWaitsForStartupTasksToComp…

kirklund:
GEODE-7189: Disable default server in startWaitsForStartupTasksToComplete

* setDisableDefaultServer true to skip AvailablePort check
* Move startupTasks and when to top of test method

Please review @aaronlindsey

### DIFF
--- a/geode-core/src/test/java/org/apache/geode/distributed/ServerLauncherTest.java
+++ b/geode-core/src/test/java/org/apache/geode/distributed/ServerLauncherTest.java
@@ -435,7 +435,7 @@ public class ServerLauncherTest {
 
   private CompletableFuture<Void> completedExceptionallyFuture() {
     CompletableFuture<Void> completedExceptionallyFuture = new CompletableFuture<>();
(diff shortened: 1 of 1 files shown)


geode-core/src/test/java/org/apache/geode/distributed/ServerLauncherTest.java
```diff
@@ -435,7 +435,7 @@ public void startWaitsForStartupTasksToComplete() {
 
   private CompletableFuture<Void> completedExceptionallyFuture() {
     CompletableFuture<Void> completedExceptionallyFuture = new CompletableFuture<>();
-    completedExceptionallyFuture.completeExceptionally(new RuntimeException());
+    completedExceptionallyFuture.completeExceptionally(new RuntimeException(""));
```
kirklund: This line change is unintentional

kirklund: Reverted this line

PR #4151, opened 2019-10-11
GEODE-7289: Use random port in StandaloneClientManagementAPIAcceptanc…

kirklund:
…eTest

The test is already using random ports for the locator port and http
service port. Use a random port for jmx manager port as well.

Please review @aaronlindsey


kirklund:
Failure in DistributedTestOpenJDK11 is PersistentPartitionedRegionWithRedundancyDUnitTest which is unrelated. Looks like that is a recently added flaky test.

PR #4153, opened 2019-10-11
GEODE-7237: Fix race condition in GfshRule

aaronlindsey:
`GfshRule.execute()` now blocks while the ProcessLogger is still collecting output from the GFSH process. This fixes a race condition that causes `ConnectCommandAcceptanceTest.invalidHostname()` to fail intermittently.


jdeppe-pivotal:
From my evaluation it appears that the problem is between the time that the `gfsh` process has launched and the test calls `getOutputText()`. At this point the `ProcessLogger` has not had enough time to process all the output from the gfsh process. I'd suggest augmenting `GfshExecution.awaitTermination()` to also wait on `ProcessLogger` to 'complete' - that would imply to wait until its `StreamGobbler` threads have stopped.

kirklund:
Thanks for investigating this @jdeppe-pivotal!

aaronlindsey:
Thanks @jdeppe-pivotal! I re-worked this PR per your suggestion. Now `GfshExecution.awaitTermination()` will also wait for `ProcessLogger` to finish collecting output from the process.

aaronlindsey:
Looks like this is breaking some acceptance tests... will fix.

PR #4154, opened 2019-10-14
GEODE-7259: Document REST API limitation on function execution params

alb3rtobr:


geode-docs/rest_apps/post_execute_functions.html.md.erb
```diff
@@ -39,7 +39,7 @@ Execute <%=vars.product_name%> function on entire cluster or on a specified regi
     http://serverURL/functions/SampleFunction?onRegion=TestPartitionRegion&filter=key1,key2,key3
     ```
 
--   Any *function arguments* are passed in the request body in JSON format. The content of the arguments can depend on how the function is defined. Use @type to declare argument types and @value for specifying a scalar value. An example set of arguments:
+-   Any *function arguments* are passed in the request body in JSON format. The content of the arguments can depend on how the function is defined. Use @type to declare argument types and @value for specifying a scalar value. Collection of objects cannot be specified as parameters. So if a function receives a collection of objects as one of its arguments, it cannot be called using <%=vars.product_name%> REST API. An example set of arguments:
```
jujoramos: Both inner collections (`[]` as JSON) and inner objects (`{}` as JSON) are not supported as part of the parameters.

alb3rtobr: What I wrote "collection of objects" I was thinking about Java objects. What about this?

Any function arguments are passed in the request body in JSON format. The content of the arguments can depend on how the function is defined. Each function argument has to be written as a JSON object (sourrounded by { }) and using "@ type" to declare its type. "@ value" can be used to specify a scalar value. Inner JSON objects ({ }) or JSON collections ([ ]) are not accepted.
Collections of Java objects cannot be specified as parameters in JSON format. So if a function receives a collection of objects as one of its arguments (e.g. List), it cannot be called using <%=vars.product_name%> REST API.

jujoramos: @alb3rtobr: that's better. What about adding a simple example of unsupported JSON documents?, that will make things even more clear to the reader.

alb3rtobr: Ok, Im on it

davebarnes97:
@alb3rtobr I think this PR is ready to merge, but I'd like to make sure you have a chance to weigh in on the changes some of us have made. What do you think - shall we merge this now?

alb3rtobr:
The changes look good to me, thanks for the corrections

PR #4157, opened 2019-10-14
GEODE-7278: convert transport filters

dschneider-pivotal:
Added unit test for GatewayReceiverConverter.
GatewayReceiverConverter now handles the transport filters.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/management/internal/configuration/converters/GatewayReceiverConverter.java
+++ b/geode-core/src/main/java/org/apache/geode/management/internal/configuration/converters/GatewayReceiverConverter.java
@@ -68,11 +68,13 @@ public class GatewayReceiverConverter
     receiver.setManualStart(configObject.isManualStart());
     receiver.setMaximumTimeBetweenPings(intToString(configObject.getMaximumTimeBetweenPings()));
     receiver.setSocketBufferSize(intToString(configObject.getSocketBufferSize()));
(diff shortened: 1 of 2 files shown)


dschneider-pivotal:
@jmelchio

geode-core/src/main/java/org/apache/geode/management/internal/configuration/converters/GatewayReceiverConverter.java
```diff
@@ -30,6 +37,26 @@ protected GatewayReceiver fromNonNullXmlObject(GatewayReceiverConfig xmlObject)
     receiver.setMaximumTimeBetweenPings(stringToInt(xmlObject.getMaximumTimeBetweenPings()));
     receiver.setSocketBufferSize(stringToInt(xmlObject.getSocketBufferSize()));
     receiver.setStartPort(stringToInt(xmlObject.getStartPort()));
+    List<ClassName> configFilters = new ArrayList<>();
```
jinmeiliao: use a ClassNameConverter to convert between ClassName and Declarable. See `PdxConverter` as an example.

dschneider-pivotal: Good idea. One concern a had about the ClassNameConverter logic is that it does not support all of the xml logic. For the parameters it only supports "string" values and not "declarable" values. Is it okay to just not handle the "declarable" parameter values?

PR #4158, opened 2019-10-14
GEODE-7131: add Links object to replace RestfulEndPoint

jinmeiliao:
Co-authored-by: Owen Nichols <onichols@pivotal.io>

* add capability to add additional links to the configuration obejct

### For all changes:

### Note:


jinmeiliao:
@jmelchio

geode-management/src/main/java/org/apache/geode/management/api/Links.java
```diff
@@ -0,0 +1,71 @@
...
+package org.apache.geode.management.api;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import com.fasterxml.jackson.annotation.JsonAnyGetter;
+import com.fasterxml.jackson.annotation.JsonAnySetter;
+import org.apache.commons.lang3.StringUtils;
+
+public class Links {
```
dschneider-pivotal: Add javadocs for this class. It is an external api. I find the class confusing. I expected it to just be a map but instead it has self, list, and others. It is also not clear how the constants on this class are related to links.
It would be best for this to be an interface so that it could be read only.
We really don't intend for users to be calling any of these settors or constructors.
If we can't make it an interface then get rid of any of the settors that we can. For example get rid of addLink. We don't need it for our implementation. We can instead just call getOthers().put(key, url)

dschneider-pivotal: Did you consider just having Links extend Map<String,String>?

geode-core/src/main/java/org/apache/geode/management/internal/configuration/converters/RegionConverter.java
```diff
@@ -75,6 +75,8 @@ protected Region fromNonNullXmlObject(RegionConfig xmlObject) {
         region.setExpirations(expirations);
       }
     }
+
+    List<RegionConfig.Index> indexes = xmlObject.getIndexes();
```
dschneider-pivotal: What is the point of this addition? It looks like "indexes" is never used

geode-management/src/main/java/org/apache/geode/management/api/Links.java
jinmeiliao: moved this to internal package. User doesn't need to use this class.

jinmeiliao: now that we have @JsonAnyGetter annotation, we can serialize the content of the map directly.

geode-core/src/main/java/org/apache/geode/management/internal/configuration/converters/RegionConverter.java
jinmeiliao: oh, added by accident. Good catch!

geode-management/src/main/java/org/apache/geode/management/api/ClusterManagementResult.java
```diff
@@ -127,19 +130,11 @@ public String getStatusMessage() {
...
-  public String getUri() {
-    return uri;
-  }
-
   /**
    * for internal use only
    */
-  public void setUri(String uri) {
-    this.uri = uri;
+  public void setLinks(Links links) {
```
dschneider-pivotal: This is a problem. You can't have "internal" classes external APIs.

geode-management/src/main/java/org/apache/geode/management/api/ConfigurationResult.java
```diff
@@ -73,4 +76,9 @@ public void setConfiguration(T configuration) {
   public void setRuntimeInfo(List<R> runtimeInfo) {
     this.runtimeInfo = runtimeInfo;
   }
+
+  @JsonProperty(access = JsonProperty.Access.READ_ONLY)
+  public Links getLinks() {
```
dschneider-pivotal: another internal "Links" exposed an a non-internal api

geode-management/src/main/java/org/apache/geode/management/configuration/AbstractConfiguration.java
```diff
@@ -86,61 +85,11 @@ public String getGroup() {
...
-   * @return {@link #URI_CONTEXT} + {@link #URI_VERSION} + {@link #getIdentityEndpoint()}
-   */
-  @JsonProperty(access = JsonProperty.Access.READ_ONLY)
-  public String getUri() {
-    if (getIdentityEndpoint() == null) {
-      return null;
-    }
-    return URI_CONTEXT + URI_VERSION + getIdentityEndpoint();
-  }
+  public abstract Links getLinks();
```
dschneider-pivotal: Another exposure of Links on a non-external api

geode-management/src/main/java/org/apache/geode/management/api/ClusterManagementResult.java
onichols-pivotal: After some review, we concluded that Links just needs to be public, as all Configuration objects are required to implement `abstract Links getLinks()`.  We found low value in separating out an interface, as this is a pretty simple data class.  Even if we make it immutable in a future refactoring, there would still be no need for an interface.

jinmeiliao:
we decide to move Links back to the public api because extension developers will need a concrete class to create the Links object when they implement AbstractConfiguration.getLinks() call.

PR #4159, opened 2019-10-14
GEODE:7266: improve ClassName javadocs

dschneider-pivotal:
cleaned up javadocs on ClassName
made isClassNameValid private

### For all changes:

### Note:

### DIFF
--- a/geode-management/src/main/java/org/apache/geode/management/configuration/ClassName.java
+++ b/geode-management/src/main/java/org/apache/geode/management/configuration/ClassName.java
@@ -24,6 +24,7 @@ import java.util.regex.Pattern;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.apache.commons.lang3.StringUtils;
 
+import org.apache.geode.management.internal.ManagementHelper;
 import org.apache.geode.util.internal.GeodeJsonMapper;
 
 /**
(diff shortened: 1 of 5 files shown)


dschneider-pivotal:
@jmelchio

geode-management/src/main/java/org/apache/geode/management/configuration/ClassName.java
```diff
@@ -76,20 +80,31 @@ public ClassName(String className, String jsonInitProperties) {
...
 
+  /**
+   * @return the properties that will be used when an instance is created.
+   */
   public Properties getInitProperties() {
     return initProperties;
   }
 
-  public static boolean isClassNameValid(String fqcn) {
+  private static boolean isClassNameValid(String fqcn) {
```
jinmeiliao: why we make this private and then repeat the logic in another place? Would it be better to make this public so that other people don't have to repeat the effort?

dschneider-pivotal: The reason for making it private on this class is that this is an external class and our supported APIs do not need expose this functionality.
I will find a common "internal" class that can host this functionality.

dschneider-pivotal: I added a ManagementHelper class that is internal and has isClassNameValid. It should be available on both the client and the server.

PR #4161, opened 2019-10-15
GEODE-6990: Add xml persistence for OQL authorization

BenjaminPerryRoss:

### For all changes:

### Note:


geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/QueryServiceXmlDUnitTest.java
```diff
@@ -0,0 +1,197 @@
...
+import org.apache.geode.cache.util.TestMethodAuthorizer;
+import org.apache.geode.examples.SimpleSecurityManager;
+import org.apache.geode.test.compiler.ClassBuilder;
+import org.apache.geode.test.dunit.rules.ClusterStartupRule;
+import org.apache.geode.test.dunit.rules.MemberVM;
+import org.apache.geode.test.junit.rules.LocatorStarterRule;
+import org.apache.geode.test.junit.rules.serializable.SerializableTemporaryFolder;
+
+@RunWith(JUnitParamsRunner.class)
+public class QueryServiceXmlDUnitTest {
```
jujoramos: Only one server VM is used during the entire test, I think is better to move this class to `integationTest` and use `ServerStarterRule` instead of having it under `distributedTest` with `ClusterStartupRule`.

geode-core/src/main/java/org/apache/geode/internal/cache/GemFireCacheImpl.java
```diff
@@ -4203,7 +4204,11 @@ public JSONFormatter getJsonFormatter() {
 
   @Override
   public QueryService getLocalQueryService() {
-    return new DefaultQueryService(this);
+    try {
+      return new DefaultQueryService(this);
+    } catch (ServiceNotFoundException e) {
+      throw new IllegalStateException(e.getMessage(), e.getCause());
+    }
```
jujoramos: See my comments in the `DefaultQueryService` constructor regarding the new exceptions we are throwing.

geode-core/src/test/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceImplTest.java
```diff
@@ -0,0 +1,219 @@
...
+
+    configService.init(mockCache);
+    assertThat(configService.getMethodAuthorizer()).isInstanceOf(RestrictedMethodAuthorizer.class);
+    configService.updateMethodAuthorizer(mockCache, TestMethodAuthorizer.class.getName(),
+        new HashSet<>());
+    assertThat(configService.getMethodAuthorizer()).isInstanceOf(RestrictedMethodAuthorizer.class);
+    verify(configService).logError(any(String.class));
+  }
+
+  private Object[] getMethodAuthorizerClasses() {
```
jujoramos: Just a question, not a requirement, from an obsessive compulsive type of person... we know this method is used to provide test parameters, can we add @SuppressWarnings("unused") to it and prevent those annoying yellow warning marks from the IDE? :-).

geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/QueryServiceXmlDUnitTest.java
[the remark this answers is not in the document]
DonalEvans: Done!

geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/QueryServiceXmlDUnitTest.java
```diff
@@ -0,0 +1,197 @@
...
+
+  private String getFilePath(String fileName) throws IOException {
+    URL url = getClass().getResource(fileName);
+    File cacheXmlFile = this.temporaryFolder.newFile(fileName);
+    FileUtils.copyURLToFile(url, cacheXmlFile);
+
+    return cacheXmlFile.getAbsolutePath();
+  }
+
+  private Object[] authorizerXmlAndClassParams() {
```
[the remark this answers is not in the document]
DonalEvans: Done!

geode-core/src/distributedTest/java/org/apache/geode/cache/util/TestMethodAuthorizer.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.Set;
+
+import org.apache.geode.cache.Cache;
+import org.apache.geode.cache.query.security.MethodInvocationAuthorizer;
+
+/**
+ * If this class is modified, the changes should also be applied to TestMethodAuthorizer.txt to keep
+ * the two files identical
+ */
+public class TestMethodAuthorizer implements MethodInvocationAuthorizer {
```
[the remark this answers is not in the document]
DonalEvans: It is necessary, yes. The different test packages don't have access to each other, so tests in the `integrationTest` package don't have access to the `TestMethodAuthorizer` in the `test` package and vice versa.

geode-core/src/integrationTest/java/org/apache/geode/cache/query/DefaultQueryServiceDeprecationTest.java
```diff
@@ -12,68 +12,59 @@
...
 
 public class DefaultQueryServiceDeprecationTest implements Serializable {
 
-  private MemberVM locator;
-
   @ClassRule
   public static TemporaryFolder folderRule = new TemporaryFolder();
 
   @Rule
   public ClusterStartupRule cluster = new ClusterStartupRule();
```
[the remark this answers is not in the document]
DonalEvans: Done!

geode-core/src/main/java/org/apache/geode/cache/query/internal/DefaultQueryService.java
```diff
@@ -131,19 +110,21 @@
...
+    QueryConfigurationService queryConfigurationService =
+        cache.getService(QueryConfigurationService.class);
+    if (queryConfigurationService != null) {
+      this.methodInvocationAuthorizer = queryConfigurationService.getMethodAuthorizer();
     } else {
-      this.methodInvocationAuthorizer = new RestrictedMethodAuthorizer(cache);
+      throw new ServiceNotFoundException("QueryService creation failed. The service "
+          + QueryConfigurationService.class.getName() + " was not found.");
     }
   }
```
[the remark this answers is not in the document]
DonalEvans: I think you're right that it's not possible for the `QueryConfigurationService` to be null at this point, unless the `ServiceLoader` somehow fails to load it from the classpath. Other uses of `cache.getService()` don't seem to check if the returned object is null, so I'll just remove the check.

geode-core/src/main/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceImpl.java
```diff
@@ -0,0 +1,180 @@
...
+      "Exception while updating MethodInvocationAuthorizer: ";
+  public static final String NULL_CLASS_NAME =
+      "Null class name found for MethodInvocationAuthorizer. ";
+  public static final String NO_CLASS_FOUND =
+      "No MethodInvocationAuthorizer class found with name ";
+  public static final String NO_VALID_CONSTRUCTOR =
+      "No valid MethodInvocationAuthorizer constructor available. ";
+  public static final String INSTANTIATION_ERROR =
+      "Error occurred while instantiating MethodInvocationAuthorizer. ";
+  public static final String AUTHORIZER_NOT_UPDATED = "The authorizer was not updated.";
```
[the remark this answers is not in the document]
DonalEvans: Done

geode-core/src/main/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceImpl.java
```diff
@@ -0,0 +1,180 @@
...
+  @Deprecated
+  public final boolean ALLOW_UNTRUSTED_METHOD_INVOCATION;
+
+  public static final String DEPRECATION_WARNING = "The property " + GEMFIRE_PREFIX +
+      "QueryService.allowUntrustedMethodInvocation is deprecated. " +
+      "Please use the UnrestrictedMethodAuthorizer implementation of MethodInvocationAuthorizer " +
+      "instead";
+
+  public QueryConfigurationServiceImpl() {
+    ALLOW_UNTRUSTED_METHOD_INVOCATION = Boolean.valueOf(
```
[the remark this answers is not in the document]
DonalEvans: Done

geode-core/src/main/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceImpl.java
```diff
@@ -0,0 +1,180 @@
...
+    } else {
+      logError(UPDATE_ERROR_MESSAGE + INSTANTIATION_ERROR + AUTHORIZER_NOT_UPDATED + " Reason: "
+          + e.toString());
+    }
+  }
+
+  // For testing
+  void logWarning(String message) {
+    logger.warn(message);
+  }
```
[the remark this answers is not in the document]
DonalEvans: Thanks for catching this. It was used in a planned test that we didn't end up including as the behaviour was already being tested elsewhere.

geode-core/src/test/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceImplTest.java
DonalEvans: Done

geode-core/src/main/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceImpl.java
```diff
@@ -0,0 +1,176 @@
...
+        Class<MethodInvocationAuthorizer> userClass =
+            (Class<MethodInvocationAuthorizer>) ClassPathLoader.getLatest().forName(className);
+        Constructor<MethodInvocationAuthorizer> userConstructor =
+            userClass.getDeclaredConstructor(Cache.class, Set.class);
+        this.authorizer = userConstructor.newInstance(cache, parameters);
+      } catch (Exception e) {
+        logErrorMessage(className, e);
+      }
+    }
+  }
```
jujoramos: Random thought: should we use a lock or something to synchronize these methods?, what happens if multiple threads try to update/read the authorizer at the same time?.

geode-core/src/integrationTest/java/org/apache/geode/cache/query/internal/QueryServiceXmlIntegrationTest.java
```diff
@@ -0,0 +1,173 @@
...
+
+    assertThat(authorizer).isInstanceOf(TestMethodAuthorizer.class);
+
+    String testString = "testString";
+    Method allowedMethod = String.class.getMethod("toString");
+    assertThat(authorizer.authorize(allowedMethod, testString)).isTrue();
+
+    Method disallowedMethod = String.class.getMethod("isEmpty");
+    assertThat(authorizer.authorize(disallowedMethod, testString)).isFalse();
+  }
```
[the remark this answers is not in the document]
DonalEvans: You're right, the text file was necessary when the test was using ClusterStartupRule, but now that it's not running in a separate VM, it's no longer needed.

geode-core/src/integrationTest/java/org/apache/geode/cache/query/internal/DefaultQuerySecurityIntegrationTest.java
```diff
@@ -76,7 +78,7 @@ private void createAndPopulateRegion(String name, RegionShortcut shortcut, int e
   }
 
   @Before
-  public void setUp() {
+  public void setUp() throws ServiceNotFoundException {
```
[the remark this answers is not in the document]
DonalEvans: > Random question: what happens if multiple servers have different authorizers configured within the `cache.xml` file when they start up?. Should we add some safe guard to prevent this, as we do to avoid multiple members having the same region with different attributes?, something similar?, nothing at all?.

If I recall correctly, there are other elements of `cache.xml` that also do not check to confirm that all servers have the same settings, so while it's possible that a user could create a situation where different servers have different authorizers configured, it's not something I think we need to check here. Using cache.xml at all is generally discouraged at this point, and the main expected use case will be configuration through gfsh and cluster config, so this should not present a major risk.

geode-core/src/integrationTest/resources/org/apache/geode/cache/query/internal/QueryServiceWithJavaBeanAuthorizer.xml
```diff
@@ -0,0 +1,32 @@
...
+  limitations under the License.
+-->
+<cache
+        xmlns="http://geode.apache.org/schema/cache"
+        xmlns:query-service="http://geode.apache.org/schema/query-service"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://geode.apache.org/schema/cache
+                            http://geode.apache.org/schema/cache/cache-1.0.xsd
+                            http://geode.apache.org/schema/query-service
+                            http://geode.apache.org/schema/query-service/query-service-1.0.xsd"
```
jhuynh1: should we change the xsd and related variables to query-config-service?

geode-core/src/main/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceImpl.java
```diff
@@ -0,0 +1,176 @@
...
+    if (isSecurityDisabled((InternalCache) cache) || ALLOW_UNTRUSTED_METHOD_INVOCATION) {
+      return;
+    }
+
+    if (className == null) {
+      logError(UPDATE_ERROR_MESSAGE + NULL_CLASS_NAME + AUTHORIZER_NOT_UPDATED);
+      return;
+    }
+
+    if (className.equals(RestrictedMethodAuthorizer.class.getSimpleName())) {
```
jhuynh1: is there a reason why we use simple name instead of the entire path?  If we use simple name, we probably need to document that these are 'reserved' in that if a customer tries to create their own authorizers with the same name, they will end up getting our prebuilt ones?

geode-core/src/main/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceImpl.java
jhuynh1: we probably need to solve this on the distributed level as well.

geode-core/src/main/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceImpl.java
```diff
@@ -0,0 +1,176 @@
...
+    } else {
+      logError(UPDATE_ERROR_MESSAGE + INSTANTIATION_ERROR + AUTHORIZER_NOT_UPDATED + " Reason: "
+          + e.toString());
+    }
+  }
+
+  // For testing
+  void logError(String message) {
+    logger.error(message);
+  }
```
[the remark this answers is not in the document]
DonalEvans: By the time the exception is thrown, the XML has already been entirely parsed without any problems and the cache is in the process of being configured. Even if an exception is thrown here, the `QueryConfigurationServiceImpl` will still be available and have the default `RestrictedMethodAuthorizer` configured for use. Perhaps this is grounds for aborting startup, I don't know how other cache elements handle similar issues.

+1 to printing out the stack trace though. Making errors more visible seems like it's rarely a bad thing.

geode-core/src/main/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceImpl.java
DonalEvans: That's a valid concern. Adding some kind of synchronization on the class and/or some distributed lock to prevent the possible future case of multiple gfsh commands leading to a mismatch in authorizer configuration would be a good idea.

DonalEvans:
> Do we have a test for testing invalid authorizers and verifying what the cache ends up installing?

The `QueryConfigurationServiceImplTest` covers that case, as it verifies that the authorizer is not updated if an invalid authorizer is specified. An invalid authorizer here is defined as something that either doesn't inherit from `MethodInvocationAuthorizer`, doesn't have a public constructor that takes a `Cache` and `Set<String>` or that fails to be instantiated.

geode-core/src/main/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceImpl.java
DonalEvans: We can definitely switch to using the full name for the out-of-the-box authorizers. The choice to use simple names was just one of convenience, to prevent the user having to type the whole thing out, but it definitely saves confusion to use the same pattern in all cases.

geode-core/src/integrationTest/resources/org/apache/geode/cache/query/internal/QueryServiceWithJavaBeanAuthorizer.xml
DonalEvans: Done

DonalEvans:
I had to rebase the branch to fix merge conflicts, but not all feedback/requests for changes have yet been applied.

geode-core/src/main/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceImpl.java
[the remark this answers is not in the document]
BenjaminPerryRoss: I would be hesitant to abort the startup of the cluster entirely because this can put people in situations where their whole cluster is held hostage by a feature they're not particularly invested in. Since our default settings are to restrict access to methods I would think that a cluster starting with the wrong authorizer due to a bug or user error could be fixed by using our gfsh command (or other methods in the future) to set the authorizer to the one they intended is preferable to a cluster failing to start. In the past this is how I've seen errors like this handled in cases such as the datasources loaded from cluster config which were used by the JDBC connector, and although this is a security related setting we default to high security rather than low security so I don't see that being an issue.

geode-core/src/main/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceImpl.java
BenjaminPerryRoss: We looked into adding a Dlock but realized that putting a Dlock in this code would only synchronize the actual assignment of the authorizer on each member. What we want to do is protect multiple processes attempting to assign this authorizer across the cluster from running over each other. To do that we need to add a lock to the places which trigger this update to be called across the distributed system. Since we can't block the Cache Configuration process the only place we could add such a lock would be in the gfsh command which will be added in the following story for this feature.

geode-core/src/integrationTest/java/org/apache/geode/cache/query/internal/DefaultQuerySecurityIntegrationTest.java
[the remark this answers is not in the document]
jujoramos: Fair enough, I added a [comment](https://issues.apache.org/jira/browse/GEODE-6994?focusedCommentId=16961921&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16961921) within the documentation ticket to make sure this gets captured.

geode-core/src/main/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceImpl.java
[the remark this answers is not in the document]
jujoramos: @BenjaminPerryRoss: I don't agree on this one, if I configure something wrong and the `cache.xml` can't be parsed altogether, I expect the member startup to halt right away and throw an exception so I know I did something wrong. If we continue with the startup and use the default restricted authorizer, it will be harder for our users later in time to debug what's wrong (_hey, why is this method being rejected if the authorizer I've provided is supposed to allow it??_ ). We also need to consider that the user executing queries (developer) might not the same one as the one configuring the cluster in the first place (operator).
That said, throwing an exception and stopping the member from starting up is exactly what we do when a user makes a mistake configuring other aspects of the cache, like a `CacheListener`, a `CacheLoader`, a `CacheWriter`, etc..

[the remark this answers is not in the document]
BenjaminPerryRoss: Sorry, missed this comment. While Region Attributes like CacheListener, CacheWriter, etc do cause a failure if they're misconfigured or some other problem occurs during configuration there are also features which only log errors like JNDIBindings. My understanding was that whether or not a configuration halts the start up of a member or not is related to the importance of the object being configured. I don't feel too strongly about this specific object although my original inclination was definitely that it wasn't critical enough to prevent startup so if you feel like it's important to change that then we can definitely do that. I mostly just want to make sure we're weighing the value of the Method Authorizer coming up with configuration vs the cost of failing to start a member in these situations.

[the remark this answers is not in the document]
BenjaminPerryRoss: I also wanted to clarify - if the cache.xml has invalid syntax the parsing will fail before it reaches our custom code. The only circumstance in which we log a warning is if the xml is defined in such a way that the xml is parsed correctly, but an exception is thrown while trying to build the QueryConfigurationService/MethodAuthorizer from the values parsed.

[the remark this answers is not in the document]
jujoramos: Hey @BenjaminPerryRoss , thanks for answering my concerns.
I'd still argue that it's quite important to totally halt the startup if we can't instantiate the class the customer has configured and expects to use. If we just log a warning message and continue as it nothing bad had happened, it'll be much harder for them later in time to discover what went wrong after a query failure due to an non authorised method... it could get even worse if they "lose" the initial section of the log due to rolling (I've seen it happening).
Maybe @nabarunnag or @jhuynh1 have some thoughts on this?.

DonalEvans:
> @BenjaminPerryRoss : I think there's still no definitive answer to [this discussion](https://github.com/apache/geode/pull/4161#discussion_r339060678) within the PR. Thoughts?.

This has now been addressed. Any exception encountered in the `updateMethodAuthorizer()` method during cache creation now results a `QueryConfigurationServiceException` being thrown and creation being aborted.

geode-core/src/main/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceImpl.java
```diff
@@ -0,0 +1,146 @@
...
+        this.authorizer = new JavaBeanAccessorMethodAuthorizer(cache, parameters);
+      } else if (className.equals(RegExMethodAuthorizer.class.getName())) {
+        this.authorizer = new RegExMethodAuthorizer(cache, parameters);
+      } else {
+        @SuppressWarnings("unchecked")
+        Class<MethodInvocationAuthorizer> userClass =
+            (Class<MethodInvocationAuthorizer>) ClassPathLoader.getLatest().forName(className);
+        Constructor<MethodInvocationAuthorizer> userConstructor =
+            userClass.getDeclaredConstructor(Cache.class, Set.class);
+        this.authorizer = userConstructor.newInstance(cache, parameters);
```
jujoramos: The above implies that, whenever our users want to provide their own `MethodInvocationAuthorizer`, the class must provide at least one constructor receiving a `Cache` and a `Set` as parameters, right?. What's the reasoning behind this change?, shouldn't we require a constructor with no parameters and invoke `Declarable.init()` afterwards as we do with other customisations provided by customers?.
Am I missing something?.

geode-core/src/main/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceImpl.java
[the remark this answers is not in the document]
jhuynh1: I think Juan's concerns have swayed my opinion to agree with making the user aware that a class is missing or configured improperly.  So at this point I would agree with Juan's concern.

geode-core/src/main/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceImpl.java
jhuynh1: good question, I think you mean initialize(Cache cache, Properties properties)?

geode-core/src/main/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceImpl.java
```diff
@@ -0,0 +1,146 @@
...
+
+  public QueryConfigurationServiceImpl() {
+    ALLOW_UNTRUSTED_METHOD_INVOCATION = Boolean.parseBoolean(
+        System.getProperty(GEMFIRE_PREFIX + "QueryService.allowUntrustedMethodInvocation"));
+  }
+
+  public static MethodInvocationAuthorizer getNoOpAuthorizer() {
+    // A no-op authorizer, allow method invocation
+    return ((Method m, Object t) -> true);
+  }
```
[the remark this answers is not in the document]
DonalEvans: Done!

geode-core/src/main/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceImpl.java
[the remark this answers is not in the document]
DonalEvans: This change has been applied in commit 6df1e47. Cluster startup is now aborted if an exception is thrown during method authorizer initialization.

geode-core/src/main/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceImpl.java
jujoramos: @jhuynh1 : you're right, the old `init` method is deprecated in favour of `initialize` 👍 .

BenjaminPerryRoss: After some investigation we discovered that implementing declarable would require significant changes to the rest of this feature and would also introduce other challenges around configuration that would need to be solved. We decided to change this constructor to an init method that mirrors the pattern of the Declarable interface without actually using the interface itself - this should make it easier for users to implement custom authorizers similar to how they implement other custom classes within Geode.

geode-core/src/main/java/org/apache/geode/cache/query/security/MethodInvocationAuthorizer.java
```diff
@@ -58,4 +59,6 @@
    *         {@code false} otherwise.
    */
   boolean authorize(Method method, Object target);
+
+  void initialize(Cache cache, Set<String> parameters);
```
jujoramos: Can you add `default` to the `initialize` method so actual implementations of the interface are not forced to manually implement the method if they don't need to?. Similar to what we do for the [Function](https://github.com/apache/geode/blob/rel/v1.10.0/geode-core/src/main/java/org/apache/geode/cache/execute/Function.java) interface.

geode-core/src/main/java/org/apache/geode/cache/query/security/RegExMethodAuthorizer.java
```diff
@@ -166,4 +166,7 @@ public boolean authorize(Method method, Object target) {
     // Delegate to the default authorizer.
     return restrictedMethodAuthorizer.authorize(method, target);
   }
+
+  @Override
+  public void initialize(Cache cache, Set<String> parameters) {}
```
jujoramos: Remove this once the interface method is declared as `default`.

geode-core/src/main/java/org/apache/geode/cache/query/security/RestrictedMethodAuthorizer.java
```diff
@@ -395,4 +395,7 @@ public boolean authorize(Method method, Object target) {
 
     return true;
   }
+
+  @Override
+  public void initialize(Cache cache, Set<String> parameters) {}
```
jujoramos: Remove this once the interface method is declared as `default`.

geode-core/src/main/java/org/apache/geode/cache/query/security/UnrestrictedMethodAuthorizer.java
```diff
@@ -117,4 +118,7 @@ public boolean authorize(Method method, Object target) {
     // Delegate to the default authorizer.
     return restrictedMethodAuthorizer.isAllowedGeodeMethod(method, target);
   }
+
+  @Override
+  public void initialize(Cache cache, Set<String> parameters) {}
```
jujoramos: Remove this once the interface method is declared as `default`.

geode-core/src/integrationTest/java/org/apache/geode/cache/query/internal/DefaultQuerySecurityIntegrationTest.java
```diff
@@ -265,6 +266,9 @@ public boolean authorize(Method method, Object target) {
 
       return true;
     }
+
+    @Override
+    public void initialize(Cache cache, Set<String> parameters) {}
```
jujoramos: Remove this once the interface method is declared as `default`.

geode-core/src/main/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceImpl.java
```diff
@@ -153,5 +161,8 @@ private boolean isSecurityDisabled(InternalCache cache) {
     public boolean authorize(Method method, Object target) {
       return true;
     }
+
+    @Override
+    public void initialize(Cache cache, Set<String> parameters) {}
```
jujoramos: Remove this once the interface method is declared as `default`.

geode-core/src/main/java/org/apache/geode/cache/query/security/MethodInvocationAuthorizer.java
DonalEvans: Done!

geode-core/src/main/java/org/apache/geode/cache/query/security/MethodInvocationAuthorizer.java
```diff
@@ -58,4 +59,6 @@
    *         {@code false} otherwise.
    */
   boolean authorize(Method method, Object target);
+
+  default void initialize(Cache cache, Set<String> parameters) {}
```
jhuynh1: Would you be able to add java doc here to explain that a public no arg constructor needs to be added to the implementing class and what this method is for?

DonalEvans:
> Actually I don't think we should make initlalize default method

@jhuynh1 It's possible for a user-defined authorizer to not need to implement the `initialize()` method, for example, if the configuration for the authorizer is fixed and does not need parameters in order to perform authorizations. For this reason, it would be preferable not to force users to implement the `initialize()` method by making it non-default.

geode-core/src/main/java/org/apache/geode/cache/query/security/MethodInvocationAuthorizer.java
DonalEvans: Done

geode-core/src/main/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceImpl.java
```diff
@@ -0,0 +1,165 @@
...
+import org.apache.geode.internal.ClassPathLoader;
+import org.apache.geode.internal.cache.CacheService;
+import org.apache.geode.internal.cache.InternalCache;
+import org.apache.geode.logging.internal.log4j.api.LogService;
+import org.apache.geode.management.internal.beans.CacheServiceMBeanBase;
+
+public class QueryConfigurationServiceImpl implements QueryConfigurationService {
+
+  private static final Logger logger = LogService.getLogger();
+  static final String UPDATE_ERROR_MESSAGE =
```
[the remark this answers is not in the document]
DonalEvans: This string will need to be visible for verification of the output of gfsh commands (the next story that is being worked on) so we'll keep it package-private here to avoid having to just switch it back immediately in the next PR.

geode-core/src/main/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceException.java
```diff
@@ -0,0 +1,37 @@
...
+package org.apache.geode.cache.query.internal;
+
+import org.apache.geode.cache.CacheException;
+
+public class QueryConfigurationServiceException extends CacheException {
+  private static final long serialVersionUID = -5122191060240009964L;
+
+  QueryConfigurationServiceException() {
+    super();
+  }
```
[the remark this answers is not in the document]
DonalEvans: Done

geode-core/src/main/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceException.java
```diff
@@ -0,0 +1,37 @@
...
+    super(s);
+  }
+
+  QueryConfigurationServiceException(String s, Throwable cause) {
+    super(s, cause);
+  }
+
+  QueryConfigurationServiceException(Throwable cause) {
+    super(cause);
+  }
```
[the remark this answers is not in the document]
DonalEvans: Done

geode-core/src/main/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceImpl.java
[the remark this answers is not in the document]
jujoramos: Good to know, thanks for the explanation! 👍

BenjaminPerryRoss:
> I've made some minor comments on the code and I have a bigger one, probably too late to implement but will throw it here anyway, though: the whole PR uses names like `query-config-service` for `XML` and `QueryConfigurationService*` for classes, which seems a bit confusing to me... wouldn't it be better to change this naming `query-service` / `query-service-config` and `QueryServiceConfiguration` instead as it was in the [original proposal](https://cwiki.apache.org/confluence/display/GEODE/OQL+Method+Invocation+Security#OQLMethodInvocationSecurity-ConfigurationOptions)?. I believe it better reflects the intent of the feature.

We wanted to avoid confusing between the QueryService (the previously existing service) and our new Service class which configures it. We also wanted to maintain the naming pattern used by other cache services, which end with the term 'Service'. We feel the most accurate name would be "QueryServiceConfigurationService" as it is a service which configures the query service, but truncated it to QueryConfigurationService as we felt the full name sounded a bit wordy and confusing. For now we plan to leave this as it is but if the naming becomes an issue we can always create a new PR to change it.

PR #4162, opened 2019-10-15
GEODE-7192 Update cluster config behavior in docs

karensmolermiller:


karensmolermiller:
Unrelated issues cased JDK11 tests to fail. Merging without rerunning those JDK11 tests.

PR #4169, opened 2019-10-17
GEODE-7283: Store MethodAuthorizer in Context

jujoramos:
The MethodInvocationAuthorizer should be unique per query and should
not be changed for a particular query while it's being executed so,
starting with this commit, it is directly stored as part of the
ExecutionContext. This also prevents the creation of multiple instances
of the DefaultQueryService and the configured authorizer itself while
the intermediate result objects are being traversed during the query
evaluation.

- Added tests.
- The AttributeDescriptor and MethodDispatch classes do not require
  MethodInvocationAuthorizer as a constructor parameter anymore, the
  authorizer is obtained from the context, and **only** when needed.
- The MethodDispatch class does not internally store and cache the
  MethodInvocationAuthorizer instance anymore, thus allowing Geode to
  change the authorizer in runtime without needing to re-load every
  already known method (expensive due to reflection).

### For all changes:

### Note:


jujoramos:

jujoramos:
CQ tests were failing because the `getQueryService()` method on client side (invoked when creating the new `ExecutionContext`) causes an implicit `DEFAULT` Pool initialization, which can not be avoided when using custom pools as the tests do.
The `ExecutionContext` now uses a no-op authorizer when created from a client cache: authorization and authentication logic happens only on server side anyways, and the authorizer will be obtained from the cluster configuration service directly instead of the query service
when OQL security is fully implemented (GEODE-6983).

PR #4170, opened 2019-10-17
GEODE-7273: Able to detect not colocated transaction

pivotal-eshu:
* Able to detect not colocated transaction if it is caused by the first operation on a
    replicate region and then on a partitioned region.
  * Make sure transaction host can detect this and throw TransactionDataNotColocatedException.
  * Transaction host will throw appropriate TransactionException based on the operations
    exectued.
  * Remote non host will rely on the TransactionException thrown from the tx host.

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/TXStateProxyImpl.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/TXStateProxyImpl.java
@@ -70,6 +70,8 @@ public class TXStateProxyImpl implements TXStateProxy {
    */
   private Map<Integer, Boolean> buckets = new HashMap<Integer, Boolean>();
 
+  private boolean firstOperationOnPartitionedRegion = false;
+
   protected volatile TXStateInterface realDeal;
 
   protected boolean inProgress = true;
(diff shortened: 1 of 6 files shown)


geode-core/src/main/java/org/apache/geode/internal/cache/TXStateProxyImpl.java
```diff
@@ -531,8 +591,13 @@ public Object findObject(KeyInfo key, LocalRegion r, boolean isCreate, boolean g
           disableCopyOnRead, preferCD, requestingClient, clientEvent, false);
       trackBucketForTx(key);
       return retVal;
-    } catch (TransactionDataRebalancedException | PrimaryBucketException re) {
-      throw getTransactionException(key, re);
+    } catch (TransactionDataRebalancedException transactionDataRebalancedException) {
```
jhuynh1: Would it make sense to pull all this handling into a common method?  I see the negatives of doing this as well but it would make updating this code in the future a lot easier

pivotal-eshu: The reason to catch separate exceptions is to avoid instance checks. l think it is preferable to avoid instance check in the code.

pivotal-eshu: Refactoring partial codes.

PR #4179, opened 2019-10-18
GEODE-7283: Store MethodAuthorizer in Context

jujoramos:
The MethodInvocationAuthorizer should be unique per query and should
not be changed for a particular query while it's being executed so,
starting with this commit, it is directly stored as part of the
ExecutionContext. This also prevents the creation of multiple instances
of the DefaultQueryService and the configured authorizer itself while
the intermediate result objects are being traversed during the query
evaluation.

- Added tests.
- Fixed minor warnings.
- Replaced usages of 'junit.Assert' by 'assertj'.
- The AttributeDescriptor and MethodDispatch classes do not require
  MethodInvocationAuthorizer as a constructor parameter anymore, the
  authorizer is obtained from the context, and **only** when needed.
- The MethodDispatch class does not internally store and cache the
  MethodInvocationAuthorizer instance anymore, thus allowing Geode to
  change the authorizer in runtime without needing to re-load every
  already known method (expensive due to reflection).
- The ExecutionContext uses a no-op authorizer when created from a
  client cache: authorization and authentication logic happens only on
  server side, and the authorizer will be obtained from the cluster
  configuration service directly instead of the query service when OQL
  security is fully implemented.

### For all changes:

### Note:


jujoramos:

geode-core/src/integrationTest/java/org/apache/geode/cache/query/internal/ExecutionContextIntegrationTest.java
```diff
@@ -0,0 +1,340 @@
...
+    List list1 = context.getCurrScopeDpndntItrsBasedOnSingleIndpndntItr(indItr);
+    assertThat(list1.size())
+        .as("The dependency set returned incorrect result with size =" + list1.size())
+        .isEqualTo(4);
+
+    assertThat(list1).isEqualTo(checkList);
+  }
+
+  @Test
+  public void testScopeIndex1() throws Exception {
```
jhuynh1: is there a more descriptive name for this test?

geode-core/src/integrationTest/java/org/apache/geode/cache/query/internal/ExecutionContextIntegrationTest.java
```diff
@@ -0,0 +1,340 @@
...
+
+    assertIteratorScope(itr);
+    query.evaluate(context);
+    runtimeItrs = context.getDependencySet(query, true);
+    itr = runtimeItrs.iterator();
+    assertIteratorScope(itr);
+  }
+
+  @Test
+  public void testMultiThreadedScopeIndex() {
```
jhuynh1: Is there a more descriptive name that describes what the desired behavior is?

geode-core/src/integrationTest/java/org/apache/geode/cache/query/internal/ExecutionContextIntegrationTest.java
```diff
@@ -0,0 +1,340 @@
...
+        .withProperty("security-password", "cluster")
+        .startServer();
+    ExecutionContext securedContext = new QueryExecutionContext(null, server.getCache());
+    MethodInvocationAuthorizer authorizer = securedContext.getMethodInvocationAuthorizer();
+    assertThat(authorizer).isNotNull();
+    assertThat(authorizer).isInstanceOf(RestrictedMethodAuthorizer.class);
+  }
+
+  @Test
+  public void testFunctionalAddToIndependentRuntimeItrMapWithoutIndex() throws Exception {
```
jhuynh1: what is this testing (more descriptive name?)  It looks like we are checking that the internal id is the same as the index internal id?

geode-core/src/integrationTest/java/org/apache/geode/cache/query/internal/ExecutionContextIntegrationTest.java
```diff
@@ -0,0 +1,340 @@
...
+    Iterator iter = list.iterator();
+    int i = 0;
+    while (iter.hasNext()) {
+      CompiledIteratorDef iterDef = (CompiledIteratorDef) iter.next();
+      i = computeEvaluateAndAssertIterator(context, i, iterDef);
+    }
+  }
+
+  @Test
+  public void testObtainingRegionPath() throws Exception {
```
jhuynh1: Is this confirming that the region path in the context matches after computing dependencies or in this case computingUltimateDependencies?  (Not knowing what the difference is at this time...)

jujoramos:
@DonalEvans: thanks for reviewing.
@jhuynh1: I didn't change the existing tests, just refactored the class to get rid of deprecated code and warnings, and added some new tests with meaningful names to test my changes. I'll dive into these particular tests to figure out what they're doing and change the names as requested, though, thanks.

PR #4184, opened 2019-10-18
GEODE-7312: modify the ThreadsMonitor to print the stack of a blocking…

bschuchardt:
… thread

Log a thread trace for a thread that's blocking a "stuck" thread.

This will help a lot when a user experiences hung operations.  Prior to
this change we needed to request thread dumps for servers and that was
usually not possible to obtain because the servers had already been
terminated & restarted.  This change puts the relevant thread dumps in
the server's log file, which is much easier for folks to gather after
the fact.

@yossireg 

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/monitoring/executor/AbstractExecutor.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/monitoring/executor/AbstractExecutor.java
@@ -92,7 +92,7 @@ public abstract class AbstractExecutor {
       writeThreadStack(thread, "Thread stack:", stringBuilder);
     }
 
-    if (thread.getLockOwnerName() != null) {
+    if (logThreadDetails && thread.getLockOwnerName() != null) {
       ThreadInfo lockOwnerThread = ManagementFactory.getThreadMXBean()
           .getThreadInfo(thread.getLockOwnerId(), THREAD_DUMP_DEPTH);
       if (lockOwnerThread != null) {
--- a/geode-core/src/main/java/org/apache/geode/internal/monitoring/executor/AbstractExecutor.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/monitoring/executor/AbstractExecutor.java
@@ -28,6 +28,7 @@ public abstract class AbstractExecutor {
 
   private static final int THREAD_DUMP_DEPTH = 40;
   private static final Logger logger = LogService.getLogger();
+  public static final String LOCK_OWNER_THREAD_STACK = "Lock owner thread stack";
   private long threadID;
   private String groupName;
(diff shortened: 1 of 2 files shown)


geode-core/src/main/java/org/apache/geode/internal/monitoring/executor/AbstractExecutor.java
```diff
@@ -47,53 +48,68 @@ public AbstractExecutor(ThreadsMonitoring tMonitoring, long threadID) {
 
   public void handleExpiry(long stuckTime) {
     this.incNumIterationsStuck();
-    logger.warn(handleLogMessage(stuckTime));
+    logger.warn(createThreadReport(stuckTime));
   }
 
-  private String handleLogMessage(long stuckTime) {
+  String createThreadReport(long stuckTime) {
```
echobravopapa: like the renames

geode-core/src/main/java/org/apache/geode/internal/monitoring/executor/AbstractExecutor.java
```diff
@@ -47,53 +48,68 @@ public AbstractExecutor(ThreadsMonitoring tMonitoring, long threadID) {
...
+      if (lockOwnerThread != null) {
+        writeThreadStack(lockOwnerThread, LOCK_OWNER_THREAD_STACK, stringBuilder);
       }
     }
 
-    return strb.toString();
+    return stringBuilder.toString();
+  }
+
+  private void writeThreadStack(ThreadInfo thread, String header, StringBuilder strb) {
```
echobravopapa: nice method extraction

geode-core/src/main/java/org/apache/geode/internal/monitoring/executor/AbstractExecutor.java
```diff
@@ -47,53 +48,68 @@ public AbstractExecutor(ThreadsMonitoring tMonitoring, long threadID) {
...
 
     if (logThreadDetails) {
-      strb.append("Thread Stack:").append(System.lineSeparator());
-      for (int i = 0; i < thread.getStackTrace().length; i++) {
-        String row = thread.getStackTrace()[i].toString();
-        strb.append(row).append(System.lineSeparator());
+      writeThreadStack(thread, "Thread stack:", stringBuilder);
+    }
+
+    if (logThreadDetails && thread.getLockOwnerName() != null) {
```
echobravopapa: what is the situation when these could/would be null?

bschuchardt: If there's no thread blocking the stuck thread there won't be a lockOwnerName

PR #4185, opened 2019-10-18
GEODE-7266: change className validation

jinmeiliao:

### For all changes:

### Note:


jinmeiliao:
@jmelchio

geode-core/src/test/java/org/apache/geode/management/internal/cli/domain/ClassNameTest.java
```diff
@@ -15,18 +15,22 @@
 
 package org.apache.geode.management.internal.cli.domain;
```
dschneider-pivotal: This unit test should be in the same package the the class it is testing. ClassName is not internal so this test should not be internal.

geode-management/src/main/java/org/apache/geode/management/configuration/ClassName.java
```diff
@@ -35,8 +36,8 @@
 public class ClassName implements Serializable {
```
dschneider-pivotal: Is this class only intended to be serialized with the spring ObjectMapper? If so does it need to implement Serializable? I don't think we ever intend to use standard jdk serialization but since it is an external api and since it implements Serializable, we should either remove Serializable (and the uid) from it, or we should test it and make sure standard jdk serialization works with it.

geode-management/src/main/java/org/apache/geode/management/configuration/ClassName.java
```diff
@@ -134,4 +135,12 @@ public boolean equals(Object o) {
         && this.getInitProperties().equals(that.getInitProperties());
   }
 
+  public static boolean isClassNameValid(String className) {
```
dschneider-pivotal: I'd prefer that this method be on an internal class (like the ManagementHelper class you removed) instead of exposing it to users on the external api.

geode-management/src/main/java/org/apache/geode/management/configuration/ClassName.java
jinmeiliao: I believe this class is used to be sent from locator to server, that's where it needs to be Serializable.

geode-management/src/main/java/org/apache/geode/management/configuration/ClassName.java
jinmeiliao: I don't see the harm of exposing it to users. I really don't like the general purpose XXXHelper class to hold these methods. It makes it hard to find where it is

PR #4186, opened 2019-10-18
GEODE-7258: The function retry logic is modified to handle exception

gesterzhou:
thrown, while trying to connect to a server thats shutdown/closed.

    Co-authored-by: Xiaojian Zhou <gzhou@pivotal.io>

### For all changes:

### Note:


geode-core/src/main/java/org/apache/geode/cache/client/internal/ExecuteFunctionOp.java
```diff
@@ -107,11 +106,8 @@ public static void execute(final PoolImpl pool,
 
         } catch (ServerConnectivityException se) {
 
-          if (maxRetryAttempts == PoolFactory.DEFAULT_RETRY_ATTEMPTS) {
-            // If the retryAttempt is set to default(-1). Try it on all servers once.
-            // Calculating number of servers when function is re-executed as it involves
-            // messaging locator.
-            maxRetryAttempts = pool.getConnectionSource().getAllServers().size() - 1;
+          if (maxRetryAttempts == -1) {
+            maxRetryAttempts = pool.calculateRetryAttempts(se);
```
jhuynh1: I didn't notice this before, but if we continue to get ServerConnectivityExceptions, we will continue to recalculate the number of maxRetryAttempts - so this maxRetryAttempts could potentially change during every loop?  is this expected/intended?  (the behavior didnt change in this diff, but just something to be aware of)

geode-core/src/main/java/org/apache/geode/cache/client/internal/ExecuteFunctionOp.java
```diff
@@ -83,8 +82,8 @@ public static void execute(final PoolImpl pool,
     } else {
 
       boolean reexecute = false;
+      int maxRetryAttempts = -1;
```
jhuynh1: This can be a bit confusing since it's(-1) the same value as DEFAULT_RETRY_ATTEMPTS.

Would it make sense to pull out a local variable such as "NEEDS_TO_CALCULATE_RETRY_ATTEMPTS = -2?"

geode-core/src/main/java/org/apache/geode/cache/client/internal/ExecuteRegionFunctionOp.java
```diff
@@ -67,17 +66,14 @@ private ExecuteRegionFunctionOp() {
...
-   * @param maxRetryAttempts Maximum number of retry attempts
    */
   static void execute(ExecutablePool pool,
       ResultCollector resultCollector,
-      int maxRetryAttempts, boolean isHA,
+      int retryAttempts, boolean isHA,
       ExecuteRegionFunctionOpImpl op, boolean isReexecute,
       Set<String> failedNodes) {
 
+    int maxRetryAttempts = retryAttempts > 0 ? retryAttempts : -1;
```
jhuynh1: Same comment as above, but perhaps we should use the same extracted variable as before?  -1 has multiple meanings at this point...

geode-core/src/main/java/org/apache/geode/cache/client/internal/PoolImpl.java
```diff
@@ -1581,4 +1582,27 @@ public int getPendingEventCount() {
...
+      // As calculating number of servers involves sending message to locator, it is
+      // done only when there is an exception.
+      if (cause instanceof ServerConnectivityException
+          && cause.getMessage().contains(ConnectionManagerImpl.BORROW_CONN_ERROR_MSG)) {
+        // The client was unable to establish a connection before sending the
+        // request.
+        maxRetryAttempts = getConnectionSource().getAllServers().size();
+      } else {
+        // The request was sent once.
+        maxRetryAttempts = getConnectionSource().getAllServers().size() - 1;
```
jhuynh1: Do we know that in all other cases we want -1.  Would it be safer to do size() for all cases and isolate the specific scenarios where we would want to do a -1?

geode-core/src/main/java/org/apache/geode/cache/client/internal/SingleHopClientExecutor.java
```diff
@@ -120,15 +118,8 @@ static int submitAllHA(List callableTasks, LocalRegion region, boolean isHA,
...
-            if (maxRetryAttempts == 0) {
-              maxRetryAttempts = retryAttemptsArg;
-            }
-
-            if (maxRetryAttempts == PoolFactory.DEFAULT_RETRY_ATTEMPTS) {
-              // If the retryAttempt is set to default(-1). Try it on all servers once.
-              // Calculating number of servers when function is re-executed as it involves
-              // messaging locator.
-              maxRetryAttempts = pool.getConnectionSource().getAllServers().size() - 1;
+            if (maxRetryAttempts == -1) {
```
jhuynh1: Same question about -1 as before

geode-core/src/main/java/org/apache/geode/cache/client/internal/PoolImpl.java
gesterzhou: There's a discussion on that. We can use size() -1 as default SCE, and identify each special case to not to decrease 1. Another alternative is: use size() as default SCE, only to use size()-1 for timeout error msg.

geode-core/src/main/java/org/apache/geode/cache/client/internal/ExecuteFunctionOp.java
gesterzhou: The retryAttempts are calculated for the first ServerConnectivityException; which changes the init value of maxRetryAttempts; in the following code, we check to see if its less than 0 and return from the call (breaking from the for loop).

geode-core/src/main/java/org/apache/geode/cache/client/internal/PoolImpl.java
```diff
@@ -1581,4 +1582,28 @@ public int getPendingEventCount() {
...
+
+  public int calculateRetryAttempts(Throwable cause) {
+
+    int maxRetryAttempts = getRetryAttempts();
+
+    if (maxRetryAttempts == PoolFactory.DEFAULT_RETRY_ATTEMPTS) {
+      // If the retryAttempt is set to default(-1). Try executing on all servers once.
+      // As calculating number of servers involves sending message to locator, it is
+      // done only when there is an exception.
+      if (cause instanceof ServerConnectivityException
```
jhuynh1: would it make sense to have the .size() be the default and the specific scenario we can detect, be the one that does the -1?

Meaning we error on the side of retrying 1 extra time instead of 1 less time?

geode-lucene/src/upgradeTest/java/org/apache/geode/cache/lucene/RollingUpgradeQueryReturnsCorrectResultsAfterClientAndServersAreRestartedFromCurrentVersion.java
```diff
@@ -0,0 +1,147 @@
...
+    Collection<Object[]> rval = new ArrayList<>();
+    rval.add(new Object[] {true, true});
+    rval.add(new Object[] {true, false});
+    rval.add(new Object[] {false, true});
+    rval.add(new Object[] {false, false});
+    return rval;
+  }
+
+  @Test
+  public void test() throws Exception {
```
jhuynh1: Probably rename this test to something more meaningful

PR #4188, opened 2019-10-19
GEODE-7319: Solution for flaky test

mivanac:

### For all changes:

### Note:


geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/CacheClientNotifier.java
```diff
@@ -162,6 +162,11 @@ public static CacheClientNotifier getInstance() {
     return ccnSingleton;
   }
 
+  @VisibleForTesting
+  public static void resetInstance() {
+    ccnSingleton = null;
+  }
+
```
jujoramos: You might want to double check on the dev list about this, but I think we want to get rid of these `@VisibleForTesting` stuff instead of introducing new ones. Maybe you can somehow refactor the class to make it more testable?.

mivanac: This method is introduced to solve flaky test. And I wanted to restrict its use only for testing.

mivanac: Problem is CacheClientNotifier.singletonHasClientProxies(), which returns true, although CacheClientNotifier instance is not allocated. So proposed solution implementing this new method, or we can set initial value of ccnSingleton to null.

PR #4191, opened 2019-10-21
do not include jackson jars in the war file

jinmeiliao:

### For all changes:

### Note:


jinmeiliao:
1. Versions of dependent jars can be different to what is required/used by Geode itself
   I may be wrong, but I thought the gradle version conflict resolution will always resolve to use the same version of jackson libs even if the web component is using a different version than the core.

1. Since we're publishing wars (and they can be referenced and used in Spring apps for example) they should definitely be self contained.
   but the wars are not self-contained even now, they definitely doesn't include the geode-core jars that it needs. Any Spring apps that includes this war will have to pull in geode-core.jar which will pull in jackson libs.

jdeppe-pivotal:
1. What I meant was just that the 'web' versions and 'core' versions *could* be different if needed. (Say we wanted the web components to use Spring 5, for example). Also, we should consider situations where components are not started/managed by `gfsh` but rather by something like *Spring Boot* where it's much easier for a user to adjust the versions of dependencies.
2. It is true that the web components do not include *geode*, however I don't think it's ever a good idea to implicitly depend on a transitive dependency of an immediate dependency. (Hopefully that makes sense). As much as possible, we should be isolating and encapsulating the dependencies of the web components.

PR #4201, opened 2019-10-22
GEODE-7326: Add cache gets timers

aaronlindsey:
- Add timers to record get operations from a Java client
- Tag each timer with region name and result (hit/miss)
- Close existing entries gauge when closing `RegionPerfStats`
- Improve micrometer assertj assertions

### DIFF
--- a/geode-assembly/src/acceptanceTest/java/org/apache/geode/metrics/CacheGetsTimerTest.java
+++ b/geode-assembly/src/acceptanceTest/java/org/apache/geode/metrics/CacheGetsTimerTest.java
@@ -1,0 +1,521 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.apache.geode.metrics;
+
+import static java.io.File.pathSeparatorChar;
+import static java.util.concurrent.TimeUnit.NANOSECONDS;
+import static java.util.stream.Collectors.toList;
+import static org.apache.geode.cache.execute.FunctionService.onServer;
+import static org.apache.geode.distributed.ConfigurationProperties.SECURITY_MANAGER;
+import static org.apache.geode.internal.AvailablePortHelper.getRandomAvailableTCPPorts;
+import static org.apache.geode.test.compiler.ClassBuilder.writeJarFromClasses;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.catchThrowable;
+
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.Serializable;
+import java.nio.file.Path;
+import java.util.Collection;
+import java.util.List;
+import java.util.Properties;
+
+import io.micrometer.core.instrument.Timer;
+import org.junit.After;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+
+import org.apache.geode.cache.Region;
+import org.apache.geode.cache.client.ClientCache;
+import org.apache.geode.cache.client.ClientCacheFactory;
+import org.apache.geode.cache.client.ClientRegionShortcut;
+import org.apache.geode.cache.execute.Function;
+import org.apache.geode.cache.execute.FunctionContext;
+import org.apache.geode.distributed.DistributedMember;
+import org.apache.geode.rules.ServiceJarRule;
+import org.apache.geode.security.AuthInitialize;
+import org.apache.geode.security.AuthenticationFailedException;
+import org.apache.geode.security.ResourcePermission;
+import org.apache.geode.test.junit.rules.gfsh.GfshRule;
+
+public class CacheGetsTimerTest {
+  private int locatorPort;
+  private ClientCache clientCache;
+  private Region<Object, Object> replicateRegion;
+  private Region<Object, Object> partitionRegion;
+
+  @Rule
+  public GfshRule gfshRule = new GfshRule();
+
+  @Rule
+  public TemporaryFolder temporaryFolder = new TemporaryFolder();
+
+  @Rule
+  public ServiceJarRule serviceJarRule = new ServiceJarRule();
+
+  @After
+  public void tearDown() {
+    if (clientCache != null) {
+      clientCache.close();
+    }
+
+    if (locatorPort != 0) {
+      String connectToLocatorCommand = "connect --locator=localhost[" + locatorPort + "]";
+      String shutdownCommand = "shutdown --include-locators";
+      gfshRule.execute(connectToLocatorCommand, shutdownCommand);
+    }
+  }
+
+  @Test
+  public void hitTimerRecordsCountAndTotalTime_ifGetsPerformedOnReplicateRegionWithExistingKey()
+      throws IOException {
+    startWithTimeStatisticsEnabled();
+
+    for (int i = 0; i < 5; i++) {
+      replicateRegion.put(i, i);
+      replicateRegion.get(i);
+    }
+
+    TimerValue hitTimerValue = hitTimerValueForRegion(replicateRegion);
+
+    assertThat(hitTimerValue.region)
+        .as("Cache gets hit timer region tag value")
+        .isEqualTo(replicateRegion.getName());
+
+    assertThat(hitTimerValue.result)
+        .as("Cache gets hit timer result tag value")
+        .isEqualTo("hit");
+
+    assertThat(hitTimerValue.count)
+        .as("Cache gets hit timer count for region " + replicateRegion.getName())
+        .isEqualTo(5);
+
+    assertThat(hitTimerValue.totalTime)
+        .as("Cache gets hit timer total time for region " + replicateRegion.getName())
+        .isGreaterThan(0);
+  }
+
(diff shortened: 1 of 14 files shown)


aaronlindsey:
Please review @kirklund @demery-pivotal @mhansonp

geode-junit/src/main/java/org/apache/geode/test/micrometer/AbstractMeterAssert.java
```diff
@@ -38,7 +38,7 @@
    */
   AbstractMeterAssert(M meter, Class<A> selfType) {
     super(meter, selfType);
-    meterId = meter.getId();
+    meterId = meter == null ? null : meter.getId();
```
mhansonp: A total nitpick but a set of parens would help readability.

demery-pivotal: I've fixed this in a different way in the metrics session PR. My original code was an unwarranted optimization (caching the ID). Better is to defer fetching the ID until an assertion evaluates it. Given that the first step in each assertion method is to assert that the actual meter is non-null, each assertion method can safely dereference the meter to get the ID.

geode-junit/src/main/java/org/apache/geode/test/micrometer/TimerAssert.java
```diff
@@ -49,6 +63,19 @@ public TimerAssert hasCount(Condition<? super Long> condition) {
...
+  /**
+   * Verifies that the timer's total time satisfies the given condition.
+   *
+   * @param timeUnit the time unit to which to convert the total time before evaluating
+   * @param totalTime the expected value of the timer's total time
+   * @return this assertion object
+   * @throws AssertionError if the timer is {@code null}
+   * @throws AssertionError if the condition rejects the timer's total time
+   */
+  public TimerAssert hasTotalTime(TimeUnit timeUnit, double totalTime) {
```
demery-pivotal: At some point (not now) we may want to add a tolerance parameter. Testing `double` for equality can cause trouble due to rounding.

geode-junit/src/main/java/org/apache/geode/test/micrometer/AbstractMeterAssert.java
aaronlindsey: It seems best to go with @demery-pivotal's changes. Whoever merges their PR last can just keep the changes from Dale's PR.

geode-assembly/src/acceptanceTest/java/org/apache/geode/metrics/CacheGetsTimerTest.java
```diff
@@ -0,0 +1,521 @@
...
+        .as("All timer values for region " + partitionRegion.getName())
+        .hasSize(2)
+        .anyMatch(tv -> tv.result.equals("hit"))
+        .anyMatch(tv -> tv.result.equals("miss"))
+        .allMatch(tv -> tv.count == 1, "All timers have count of one")
+        .allMatch(tv -> tv.totalTime == 0, "All timers have total time of zero");
+  }
+
+  @Test
+  public void timersDoNotRecord_ifNotAuthorized() throws IOException {
```
demery-pivotal: I would give this a more general name, because the responsibility to refrain from recording is not about authorization specifically, but about exceptions in general. Authorization is merely one example of that.

Consider: `...ifGetThrowsException()`

geode-core/src/main/java/org/apache/geode/internal/cache/RegionPerfStats.java
```diff
@@ -313,13 +336,17 @@ public void incConflatedEventsCount() {
...
+    long totalNanos = 0;
     if (clock.isEnabled()) {
-      long totalNanos = getTime() - start;
+      totalNanos = getTime() - start;
       stats.incLong(getTimeId, totalNanos);
     }
     stats.incLong(getsId, 1L);
     if (miss) {
       stats.incLong(missesId, 1L);
+      cacheGetsMissTimer.record(totalNanos, NANOSECONDS);
```
demery-pivotal: I thought we were going to measure these in Get70, to include all of the overhead, and to ensure that we are measuring only gets that originated outside the server. Has this changed?

aaronlindsey:
> I thought we were going to measure these in Get70, to include all of the overhead, and to ensure that we are measuring only gets that originated outside the server. Has this changed?

At first, we noticed that `Get70` calls into `LocalRegion.get()` which is already instrumented with `RegionPerfStats.endGet()`. We (perhaps naively) assumed that we could just use the existing instrumentation instead of adding a new method to `RegionPerfStats`.

When we looked more closely at what's calling `RegionPerfStats.endGet()`, it turns out that java client "get all" operations, GFSH gets, and REST API gets are all recorded if we use this method to instrument. While this may actually be desirable, we aren't sure what else might be recorded if we use this method. We don't want peer-to-peer get operations to be recorded, for example.

geode-core/src/main/java/org/apache/geode/internal/cache/RegionPerfStats.java
aaronlindsey: > I'm ok with counting gets from REST, GFSH, Functions, etc. Dale has a point about timing the overhead in Get70 but I don't feel strongly about it.

If we do decide to leave the instrumentation as-is, then we should probably add some additional test for "get all", GFSH, REST, etc.

aaronlindsey:
> I'm ok with counting gets from REST, GFSH, Functions, etc. Dale has a point about timing the overhead in Get70 but I don't feel strongly about it.

If we do decide to leave the instrumentation as-is, then we should probably add some additional test for "get all", GFSH, REST, etc.

aaronlindsey:
> I thought we were going to measure these in Get70, to include all of the overhead, and to ensure that we are measuring only gets that originated outside the server. Has this changed?

After discussing with @kirklund and @demery-pivotal, I will update this PR to perform measurements in `Get70` and update the timers' descriptions to clarify that they are only recording gets from a java or native client.

geode-assembly/src/acceptanceTest/java/org/apache/geode/metrics/CacheGetsTimerTest.java
aaronlindsey: Addressed in latest commit

geode-core/src/main/java/org/apache/geode/internal/cache/RegionPerfStats.java
aaronlindsey: Addressed in latest commit

geode-junit/src/main/java/org/apache/geode/test/micrometer/TimerAssert.java
```diff
@@ -32,6 +34,18 @@
     super(timer, TimerAssert.class);
   }
 
+  /**
+   * Verifies that the timer's count satisfies the given condition.
+   *
+   * @param count the expected value of the timer's count
+   * @return this assertion object
+   * @throws AssertionError if the timer is {@code null}
+   * @throws AssertionError if the condition rejects the timer's count
```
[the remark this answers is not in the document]
aaronlindsey: Addressed in latest commit

geode-junit/src/main/java/org/apache/geode/test/micrometer/TimerAssert.java
```diff
@@ -49,6 +63,19 @@ public TimerAssert hasCount(Condition<? super Long> condition) {
...
   }
 
+  /**
+   * Verifies that the timer's total time satisfies the given condition.
+   *
+   * @param timeUnit the time unit to which to convert the total time before evaluating
+   * @param totalTime the expected value of the timer's total time
+   * @return this assertion object
+   * @throws AssertionError if the timer is {@code null}
+   * @throws AssertionError if the condition rejects the timer's total time
```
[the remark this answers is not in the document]
aaronlindsey: Addressed in latest commit

geode-assembly/src/acceptanceTest/java/org/apache/geode/metrics/CacheGetsTimerTest.java
```diff
@@ -0,0 +1,527 @@
...
+        .hasSize(2)
+        .anyMatch(tv -> tv.result.equals("hit"))
+        .anyMatch(tv -> tv.result.equals("miss"))
+        .allMatch(tv -> tv.count == 1, "All timers have count of one")
+        .allMatch(tv -> tv.totalTime == 0, "All timers have total time of zero");
+  }
+
+  @Test
+  public void timersDoNotRecord_ifGetThrows() throws IOException {
+    givenClusterWithSecurityThatDeniesAllGetOperations();
```
demery-pivotal: You ended up with a spectacular name here!

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/command/Get70.java
```diff
@@ -306,12 +308,8 @@ public Entry getValueAndIsObject(Region region, Object key, Object callbackArg,
...
       isObject = false;
     }
-    Entry result = new Entry();
-    result.value = data;
-    result.isObject = isObject;
-    result.keyNotPresent = !wasInvalid && (data == null || data == Token.TOMBSTONE);
-    result.versionTag = versionTag;
-    return result;
+    boolean keyNotPresent = !wasInvalid && (data == null || data == Token.TOMBSTONE);
+    return new Entry(data, isObject, keyNotPresent, versionTag);
```
demery-pivotal: Nice!

geode-core/src/main/java/org/apache/geode/internal/cache/RegionPerfStats.java
```diff
@@ -324,6 +347,16 @@ public void endGet(long start, boolean miss) {
     cachePerfStats.endGet(start, miss);
   }
 
+  @Override
+  public void endGetForClient(long start, boolean miss) {
+    long totalNanos = clock.isEnabled() ? getTime() - start : 0;
```
demery-pivotal: I am sad that we are observing clock disablement here, but I understand that changing it would require too much work for this PR.

aaronlindsey: One option I considered was always calling `clock.getTime()` instead of only calling it if the clock is enabled. This would work for now because the implementation returns zero if the clock is disabled. However, the StatisticsClock API doesn't specify that implementations of getTime() will always return zero if the clock is disabled. I suppose we could change the API since it's internal...

PR #4202, opened 2019-10-22
GEODE-5194: allow gfsh to connect if serialization version matches

onichols-pivotal:
This builds on the previous work that relaxed the version check to
ignore patch version changes.  Now even differing minor versions may be
compatible (i.e. no changes to gfsh commands or options).  Serialization
version is not a perfect proxy for when gfsh commands have changed, but
this is at least an improvement over the previous check...

Co-authored-by: Jinmei Liao <jiliao@pivotal.io>


geode-core/src/main/java/org/apache/geode/management/internal/cli/commands/ConnectCommand.java
```diff
@@ -161,17 +161,21 @@ public ResultModel connect(
...
+      String remoteGeodeVersion = invoker.getRemoteGeodeVersion();
+      if (hasSameMajorMinor(gfshGeodeVersion, remoteGeodeVersion)) {
         return result;
       }
     } catch (Exception e) {
-      gfsh.logInfo("failed to get the the remote version.", e);
+      // we failed to get the remote geode version, now try to get the remote product version for
+      // error message
+      try {
+        remoteVersion = invoker.getRemoteVersion();
```
dschneider-pivotal: I don't understand why we do this. Since we are now comparing the remote geode version, shouldn't we also report it in the error message?

onichols-pivotal: It is possible that, for example, there may be no changes in gfsh or serialization protocol between 1.10 and 1.11, so it would be safe for the 1.10 gfsh to talk to the 1.11 cluster or the other way around.  The error message is only printed if truly incompatible, in which case it makes more sense to show the actual product version, not the serialization version level (despite being named after Geode versions, if there were actually no serialization changes in 1.11, the serialization version would remain 1.10).  Would it be clearer to rename `geodeVersion` to `geodeSerializationVersion`?

onichols-pivotal:
> So we don't need to change the place where gfsh-dependency.jar is created? where do we add the geode-version.jar dependency to gfsh?

gfshDependencies is created from geode-all-bom with a few specific exclusions, so it picks up everything else automatically including geode-serialization.

geode-core/src/main/java/org/apache/geode/management/internal/cli/commands/ConnectCommand.java
onichols-pivotal: I changed `geodeVersion` to `geodeSerializationVersion`

PR #4205, opened 2019-10-23
GEODE-6971: CqAttributesFactory to offer fluid API

albertogpz:

### For all changes:

### Note:


albertogpz:
The current solution breaks the backward compatibility of the API as it can be observed in the CI failing tests (ClientAuthorizationCQDUnitTest).
Either the proposal should be revisited or a new solution must be devised.

jhuynh1:
I have a patch that I submitted a PR onto your branch.  I think it should get this diff past the test issues

albertogpz:
> I have a patch that I submitted a PR onto your branch. I think it should get this diff past the test issues

I cannot see the patch. Where should I find it?

albertogpz:
> > I have a patch that I submitted a PR onto your branch. I think it should get this diff past the test issues
> 
> Thanks a lot!

I got the patch as you provided it but the backward compatibility test cases are still failing.

jhuynh1:
Looks like I missed another location, I opened another PR, it's a similar change just at a different location and different listener.  There may be others I missed...

jhuynh1:
My concern at the moment is... what if a client app has a function compiled against the old cq api.  Would they have to recompile against the new source?  If so, that would be forcing the client app to upgrade which kind of defeats the purpose of being backwards compatible.

jhuynh1:
The tests are compiled with the new method signature and the test will execute the test code on the old servers which had the old method signature.  The fix was to use reflection to grab the correct method/method signature and do it at runtime instead of compile time.

Two potential problems:
1.) The problem now is if a client application that is compiled against the old signature, attempts to run the compiled code on a new server.. 

2.) If this change is just client side changes, then this would only be an issue if they update the geode jar without recompiling the client...

albertogpz:
> The tests are compiled with the new method signature and the test will execute the test code on the old servers which had the old method signature. The fix was to use reflection to grab the correct method/method signature and do it at runtime instead of compile time.
> 
> Two potential problems:
> 1.) The problem now is if a client application that is compiled against the old signature, attempts to run the compiled code on a new server..
> 
> 2.) If this change is just client side changes, then this would only be an issue if they update the geode jar without recompiling the client...

The way I see it (please, correct me if I am wrong) is that the method modified is only run on the client.
If this is so, the test is doing something strange: It compiles the client with the new library but then runs the client with the old one. That's why you needed to use reflection to get the right version of the method.
I think clients should be expected to be recompiled if the geode library is updated.

jhuynh1:
I think you understand how this specific test is running.
I am not sure if we can expect a user to recompile a client whenever a server gets upgraded (which is why we have backwards compatible clients)
In this specific case, if the api is only used on client side, then this change is probably a safe change.  Maybe we can ask the user list to see if anyone used the cq attributes in a function on the server.

jhuynh1:
@upthewaterspout we've historically had issues modifying apis and compiled application code. 
I am not sure if this brings back any memories or if you know whether this change definitely will break a certain use case... I would think cq's are primarily created on clients and none of that code would be sent to the server...

jhuynh1:
I'm adding a few reviewers who might have some additional insight/opinion on this area

upthewaterspout:
I think we have previously decided that breaking binary compatibility counts as breaking backwards compatibility. 

Here's a discussion thread where this came up previously:
https://lists.apache.org/thread.html/ebdeabc24ce0ff418427021dfbbd45aceb51939f25de4aba7bb5c97b@%3Cdev.geode.apache.org%3E

I think maintaining binary compatibility is important because the person upgrading geode may not always be able to recompile the source - for example when using a third party library based on geode.

albertogpz:
So it seems that it is not worth to implement this change. Right?
Should we close the PR and the JIRA with the conclusions?

jhuynh1:
Thanks Dan!
This probably should get into Geode 2.0 (whenever that happens)...
but for now I guess we can close the PR and update the jira...  Thanks Alberto!

onichols-pivotal:
closing as per above discussion

PR #4210, opened 2019-10-23
GEODE-7341: Provide a way to avoid memory lock if over committed

pivotal-eshu:


geode-core/src/main/java/org/apache/geode/distributed/internal/InternalDistributedSystem.java
```diff
@@ -817,6 +807,34 @@ private void initialize(SecurityManager securityManager, PostProcessor postProce
     reconnectAttemptCounter.set(0);
   }
 
+  void lockMemory(long avail, long size) {
+    if (avail < size) {
+      if (avoidMemoryLockWhenOvercommitted) {
+        logger.warn(
+            "System memory appears to be over committed by {} bytes. Avoid memory lock.",
```
jhuynh1: Wouldn't this experience similar issues to locking memory, but just without locking memory?  Meaning, should the two messages be the same?

geode-core/src/main/java/org/apache/geode/distributed/internal/InternalDistributedSystem.java
```diff
@@ -817,6 +807,34 @@ private void initialize(SecurityManager securityManager, PostProcessor postProce
...
+                "Insufficient free memory (%s) when attempting to lock %s bytes.  Either reduce the amount of heap or off-heap memory requested or free up additional system memory.  You may also specify -Dgemfire.Cache.ALLOW_MEMORY_OVERCOMMIT=true on the command-line to override the constraint check.",
+                avail, size));
+      }
+    } else {
+      lockMemory();
+    }
+  }
+
+  void lockMemory() {
+    logger.info("Locking memory. This may take a while...");
```
jhuynh1: Are these logs intended?

geode-core/src/test/java/org/apache/geode/distributed/internal/InternalDistributedSystemTest.java
```diff
@@ -0,0 +1,95 @@
...
+        spy(new InternalDistributedSystem.Builder(new Properties()).build());
+
+    system.lockMemory(100, 200);
+
+    verify(system, never()).lockMemory();
+  }
+
+
+  @Test
+  public void lockMemoryThrowsIfMemoryOverCommit() {
```
jhuynh1: Why didn't this need to set the system property for ALLOW_MEMORY_LOCK_WHEN_OVERCOMMITTED?

geode-core/src/main/java/org/apache/geode/distributed/internal/InternalDistributedSystem.java
```diff
@@ -172,6 +167,16 @@
...
   private final FunctionStatsManager functionStatsManager;
+  /**
+   * True if the user is allowed lock when memory resources appear to be overcommitted.
+   */
+  private final boolean allowMemoryLockWhenOvercommitted =
+      Boolean.getBoolean(GEMFIRE_PREFIX + "Cache.ALLOW_MEMORY_OVERCOMMIT");
+  /**
+   * True if memory lock is avoided when memory resources appear to be overcommitted.
+   */
+  private final boolean avoidMemoryLockWhenOvercommitted =
```
jhuynh1: Why would a user ever set both to TRUE?  It seems like one is saying lock memory when over committed and the other says don't lock memory when over committed...

geode-core/src/main/java/org/apache/geode/distributed/internal/InternalDistributedSystem.java
pivotal-eshu: I believe so. It tells user that it may take time during cache creation.

geode-core/src/main/java/org/apache/geode/distributed/internal/InternalDistributedSystem.java
pivotal-eshu: Yes, user should not set both system properties in a cache. These two system properties mean different things. User should only set one of them. However, if the user sets both, Geode will not lock memory to avoid system experience unexpected consequences by locking memory without enough memory.

geode-core/src/main/java/org/apache/geode/distributed/internal/InternalDistributedSystem.java
pivotal-eshu: No it would not experience the issue if locking memory. Without locking memory, memory can be paged in and out -- yes, it may have performance impact but should not experience instability issue, etc.

geode-core/src/test/java/org/apache/geode/distributed/internal/InternalDistributedSystemTest.java
pivotal-eshu: This will make sure users know that the system does not have enough memory to lock the memory as they desire. They need to either run their apps on different boxes with enough resources to be able to lock memory, or they have to set the system property with the consequences of instability by setting the system property. User has to set the system property.

pivotal-eshu:
Tests use RestoreSystemProperties rule to reset the system properties.

This should not cause side effects if not lock memory using the new system property introduced, compared to the unknown side effect if using the original ALLOW_MEMORY_OVERCOMMIT system property setting.

geode-core/src/main/java/org/apache/geode/distributed/internal/InternalDistributedSystem.java
```diff
@@ -817,6 +807,34 @@ private void initialize(SecurityManager securityManager, PostProcessor postProce
...
+            size - avail);
+      } else if (allowMemoryLockWhenOvercommitted) {
+        logger.warn(
+            "System memory appears to be over committed by {} bytes.  You may experience instability, performance issues, or terminated processes due to the Linux OOM killer.",
+            size - avail);
+        lockMemory();
+      } else {
+        throw new IllegalStateException(
+            String.format(
+                "Insufficient free memory (%s) when attempting to lock %s bytes.  Either reduce the amount of heap or off-heap memory requested or free up additional system memory.  You may also specify -Dgemfire.Cache.ALLOW_MEMORY_OVERCOMMIT=true on the command-line to override the constraint check.",
```
dschneider-pivotal: I think it would be better to have the system property name defined in a symbolic static final String and reference that here (and in the code that reads it) instead of hardcoding it in this string literal

geode-core/src/main/java/org/apache/geode/distributed/internal/InternalDistributedSystem.java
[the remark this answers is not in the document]
dschneider-pivotal: You might also want to add "because -D... was set to true

geode-core/src/main/java/org/apache/geode/distributed/internal/InternalDistributedSystem.java
pivotal-eshu: fixed.

PR #4222, opened 2019-10-24
Revert "GEODE-7326: Add cache gets timers"

mhansonp:
Reverts apache/geode#4201


mhansonp:
I accidentally merged this before a test was finished running and it is a problem.

PR #4203, opened 2019-10-23
GEODE-7171: Encapsulate metrics session

demery-pivotal:
This PR gathers metrics session responsibilities, which have become scattered across too many core Geode classes, and encapsulates them into a small set of closely related classes.

Changes to core Geode classes:

- GemFireCacheImpl no longer holds a meter registry or the set of "user"
  registries added by the cache builder.

- InternalCacheBuilder no longer constructs objects on behalf of the
  metrics session.  Instead, it gathers meter registries and other
  details into a MetricsService.Builder, which it passes to the
  InternalDistributedSystem.Builder.

- InternalDistributedSystem no longer knows about client meter
  registries. Instead, it holds a MetricsService, which remembers its
  builder.  During reconnect, the InternalDistributedSystem retrieves
  the builder from its metrics service, and uses the builder to build
  a similar metrics service in the reconnected system.

New classes and interfaces (org.apache.geode.metrics.internal):

- StandardMeterBinder
  - Binds standard meter binders to the meter registry.
  - close() closes the closeable binders, freeing their resources.

- MetricsService interface
  - Extends MetricsSession.
  - Adds methods to start and stop the service, and to access the
    service's meter registry.
  - Adds a method that returns the builder that built the metrics
    service, so that reconnect can use the same builder to build the
    metrics service for the reconnected system.

- InternalDistributedSystemMetricsService
  - Implements MetricsService.
  - Creates and retains a composite meter registry to manage meters.
  - Configures the composite composite registry:
    - Adds common tags that describe the system
    - Adds client meter registries to the composite.
    - Applies a StandardMeterBinder to add standard meters to the
      registry.
  - Discovers, starts, and stops metrics publishing services.
  - Adds and removes sub-registries as instructed by publishing
    services.
  - When stopped, stops each publishing service, closes all closeable
    meter binders, removes all sub-registries, and closes its composite
    meter registry.

- MetricsService.Builder interface

- InternalDistriburtedSystemMetricsService.Builder
  - Implements MetricsService.Builder
  - Accumulates and retains constructor parameters for the eventual
    InternalDistributedSystemMetricsService.
  - In particular, maintains a collection of meter registries added by
    the cache builder. It calls these 'persistent' registries, because
    they will persist across reconnects.

Authored-by: Dale Emery <demery@pivotal.io>

Thank you for submitting a contribution to Apache Geode.

In order to streamline the review of the contribution we ask you
to ensure the following steps have been taken:

### For all changes:
[...]

PR #4223, opened 2019-10-24
GEODE-7326: Add cache gets timers

aaronlindsey:
- Add timers to record get operations from a Java client
- Tag each timer with region name and result (hit/miss)
- Close existing entries gauge when closing `RegionPerfStats`
- Improve micrometer assertj assertions
- Make public fields immutable on `Get70.Entry`

The previous PR (#4201) was accidentally merged before all of the tests were passing. It was subsequently reverted. The first commit re-applies the changes from the original PR.

The second commit fixes an issue where `Get70.cmdExecute()` throws a `NullPointerException` when `InternalRegion.getRegionPerfStats()` returns null. The latter method returns null in the case where the `InternalRegion` is a `PartitionedRegion` and its data store has not been initialized:
```
  @Override
  public CachePerfStats getRegionPerfStats() {
    PartitionedRegionDataStore ds = getDataStore();
    CachePerfStats result = null;
    if (ds != null) {
      // If we don't have a data store (we are an accessor)
      // then we do not have per region stats.
      // This is not good. We should still have stats even for accessors.
      result = ds.getCachePerfStats(); // fixes 46692
    }
    return result;
  }
```


aaronlindsey:
Please review @demery-pivotal @kirklund @mhansonp

aaronlindsey:
The conflict in `AbstractMeterAssert.java` may be resolved by taking their changes and discarding mine. PR #4203 fixed the same issue in a better way (i.e. handing the case where the meter is null).

When squashing and merging, please use the following commit message:
```
GEODE-7326: Add cache gets timers

- Add timers to record get operations from a Java client
- Tag each timer with region name and result (hit/miss)
- Close existing entries gauge when closing RegionPerfStats
- Improve micrometer assertj assertions
- Make public fields immutable on Get70.Entry

```

PR #4225, opened 2019-10-25
GEODE-7334: do not load jackson classes from system class loader

jinmeiliao:
* also fix dev rest api to honor json annotation loaded by the system classloader

### For all changes:

### Note:

### DIFF
--- a/geode-http-service/src/main/java/org/apache/geode/internal/cache/InternalHttpService.java
+++ b/geode-http-service/src/main/java/org/apache/geode/internal/cache/InternalHttpService.java
@@ -181,6 +181,12 @@ public class InternalHttpService implements HttpService {
     webapp.setContextPath(webAppContext);
     webapp.setWar(warFilePath.toString());
     webapp.setParentLoaderPriority(false);
+
+    // GEODE-7334: load all jackson classes from war file except jackson annotations
+    webapp.getSystemClasspathPattern().add("com.fasterxml.jackson.annotation.");
+    webapp.getServerClasspathPattern().add("com.fasterxml.jackson.",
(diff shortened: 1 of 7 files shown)


geode-http-service/src/main/java/org/apache/geode/internal/cache/InternalHttpService.java
```diff
@@ -181,6 +181,9 @@ public synchronized void addWebApplication(String webAppContext, Path warFilePat
     webapp.setContextPath(webAppContext);
     webapp.setWar(warFilePath.toString());
     webapp.setParentLoaderPriority(false);
+
+    // GEODE-7334: do not load jodaModule even if it's present in the system classloader
+    webapp.getServerClasspathPattern().add("com.fasterxml.jackson.datatype.joda.JodaModule");
```
jdeppe-pivotal: I think this pattern restriction should be made more general to cover all of `com.fasterxml.jackson` as well as all of `org.springframework`. It's not just `JodaModule` that would cause this problem. Have a look at this: https://github.com/spring-projects/spring-framework/blob/809009e6c4d8653ea18e8fe9ae590242871711c4/spring-web/src/main/java/org/springframework/http/converter/json/Jackson2ObjectMapperBuilder.java#L772-L822.

The general approach should be to have the webapp be self-contained as much as possible. These classpath patterns should probably only exist as failsafes.

kirklund:
What's `JodaMudle`? Did you mean to instead use `JodaTime` in the commit message?

jinmeiliao:
reworked the commit message and rebase to develop

PR #4228, opened 2019-10-25
GEODE-7355: Refactored dependency on DistributionStats to use a Funct…

echobravopapa:
…ionalInterface parameter.

- Also created ArchUnit test for geode-logging

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/distributed/internal/PooledExecutorWithDMStats.java
+++ b/geode-core/src/main/java/org/apache/geode/distributed/internal/PooledExecutorWithDMStats.java
@@ -142,14 +142,15 @@ public class PooledExecutorWithDMStats extends ThreadPoolExecutor {
    * Sets timeout to IDLE_THREAD_TIMEOUT
    */
   public PooledExecutorWithDMStats(BlockingQueue<Runnable> q, int poolSize, PoolStatHelper stats,
-      ThreadFactory tf, ThreadsMonitoring tMonitoring) {
+      ThreadFactory tf, ThreadsMonitoring tMonitoring,
+      String systemPropertyPrefix) {
     /**
      * How long an idle thread will wait, in milliseconds, before it is removed from its thread
      * pool. Default is (30000 * 60) ms (30 minutes). It is not static so it can be set at runtime
      * and pick up different values.
      */
     this(q, poolSize, stats, tf,
-        Integer.getInteger(DistributionConfig.GEMFIRE_PREFIX + "IDLE_THREAD_TIMEOUT", 30000 * 60)
+        Integer.getInteger(systemPropertyPrefix + "IDLE_THREAD_TIMEOUT", 30000 * 60)
             .intValue(),
         tMonitoring);
   }
@@ -158,8 +159,8 @@ public class PooledExecutorWithDMStats extends ThreadPoolExecutor {
    * Default timeout with no stats.
    */
   public PooledExecutorWithDMStats(BlockingQueue<Runnable> q, int poolSize, ThreadFactory tf,
-      ThreadsMonitoring tMonitoring) {
-    this(q, poolSize, null/* no stats */, tf, tMonitoring);
+      ThreadsMonitoring tMonitoring, String systemPropertyPrefix) {
+    this(q, poolSize, null/* no stats */, tf, tMonitoring, systemPropertyPrefix);
   }
 
   @Override
--- a/geode-core/src/main/java/org/apache/geode/distributed/internal/tcpserver/TcpServer.java
+++ b/geode-core/src/main/java/org/apache/geode/distributed/internal/tcpserver/TcpServer.java
@@ -133,7 +133,7 @@ public class TcpServer {
 
   /*
    * We define this interface rather than using Supplier<Long> so as to avoid
-   * constructing a Long object ever time we need nano time.
+   * constructing a Long object every time we need nano time.
    */
   @FunctionalInterface
   public interface NanoTimeSupplier {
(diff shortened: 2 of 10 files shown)


geode-core/src/main/java/org/apache/geode/distributed/internal/tcpserver/TcpServer.java
```diff
@@ -126,13 +125,24 @@
...
-
   private ExecutorService executor;
   private final String threadName;
   private volatile Thread serverThread;
 
   protected SocketCreator socketCreator;
 
+  /*
+   * We define this interface rather than using Supplier<Long> so as to avoid
+   * constructing a Long object ever time we need nano time.
```
Bill: typo "ever" -> "every"

geode-core/src/main/java/org/apache/geode/distributed/internal/tcpserver/TcpServer.java
```diff
@@ -457,7 +471,7 @@ private void processOneConnection(Socket socket, long startTime, DataInputStream
 
       handler.endRequest(request, startTime);
 
-      startTime = DistributionStats.getStatTime();
+      startTime = nanoTimeSupplier.get();
```
Bill: It's wrong to modify a method parameter (`startTime` in this case).

recommend making the method parameter `final` and defining a separate `final` variable on this line e.g. `startTime2`

geode-core/src/test/java/org/apache/geode/distributed/internal/tcpserver/TcpServerDependenciesTest.java
```diff
@@ -77,7 +76,6 @@
               .or(type(SSLConfigurationFactory.class))
 
               // TODO - stats
-              .or(type(DistributionStats.class))
```
Bill: yay!

geode-core/src/main/java/org/apache/geode/distributed/internal/PooledExecutorWithDMStats.java
```diff
@@ -158,8 +159,8 @@ public PooledExecutorWithDMStats(BlockingQueue<Runnable> q, int poolSize, PoolSt
    * Default timeout with no stats.
    */
   public PooledExecutorWithDMStats(BlockingQueue<Runnable> q, int poolSize, ThreadFactory tf,
-      ThreadsMonitoring tMonitoring) {
-    this(q, poolSize, null/* no stats */, tf, tMonitoring);
+      ThreadsMonitoring tMonitoring, String systemPropertyPrefix) {
```
upthewaterspout: I'm not sure I see the reason for this change - doesn't look like this class moved out of core?

geode-core/src/main/java/org/apache/geode/distributed/internal/tcpserver/TcpServer.java
```diff
@@ -468,7 +474,7 @@ private void processOneConnection(Socket socket, long startTime, DataInputStream
         output.flush();
       }
 
-      handler.endResponse(request, startTime);
+      handler.endResponse(request, startTime2);
```
upthewaterspout: I'm ok with the LongSupplier approach - but I wonder if it would have been better to somehow associate the time supplier with the handler (eg, maybe Handler.startResponse() or something). We're getting the time from one place but then handing it off to another.

geode-core/src/main/java/org/apache/geode/distributed/internal/PooledExecutorWithDMStats.java
echobravopapa: It stems from another effort to break deps on CoreLoggingExecutors, I'd like to leave the change since it now uses `SystemPropertyHelper.GEMFIRE_PREFIX` which seems the more appropriate constant to be used...

geode-core/src/main/java/org/apache/geode/distributed/internal/tcpserver/TcpServer.java
echobravopapa: Maybe that's a good idea for some future work... right now the call to `endResponse` results in operations that could be a bit more clear... e.g.
in `LocatorStats` which is one of the places a call to `handler.endResponse` might end up

`public void endLocatorResponse(long startTime) { `
calls the ubiquitous `getStatTime`
    `long took = DistributionStats.getStatTime() - startTime;`

Bill: This makes sense @upthewaterspout. Oddly though, in `TcpServer` the start time does not typically coincide with `TcpHandler.processRequest()`. The start time is set earlier. So, as you allude, we'd need to add a method to capture the start time.

If we add only a `startResponse()` method, then that wouldn't make sense for starting a request. If we added two methods: `startRequest()`, `startResponse()` then those impls could capture a start time, but then the possibility of bugs arise: to wit, what happens if a caller starts a request, then ends a response or vice versa.

Bill: Maybe we could add a generic `startXX()` method. Then the ugly problem is implementing that on all the implementation classes:

image

Bill: Looking at the various implementations of e.g. `endRequest()` in that hierarchy, there would be value in eliminating the `startTime` parameter since the vast majority of implementations of that method ignore that parameter! Only `GMSLocatorAdapter()` uses that parameter (which it forwards to stats).

PR #4236, opened 2019-10-28
add credential type constraint

jinmeiliao:

### For all changes:

### Note:


geode-core/src/main/java/org/apache/geode/management/internal/ManagementAgent.java
```diff
@@ -349,8 +349,12 @@ private void configureAndStart() throws IOException {
     // Retrieve the PlatformMBeanServer.
     MBeanServer mbs = ManagementFactory.getPlatformMBeanServer();
 
-    // Environment map. why is this declared as HashMap?
+    // Environment map
     final HashMap<String, Object> env = new HashMap<>();
+    env.put("jmx.remote.rmi.server.credential.types", new String[] {
+        String[].class.getName(),
+        String.class.getName(),
+        Properties.class.getName()});
```
dschneider-pivotal: A comment would be helpful explaining why these classes are permitted

geode-core/src/main/java/org/apache/geode/management/internal/ManagementAgent.java
```diff
@@ -396,12 +402,12 @@ public JMXServiceURL getAddress() {
 
           @Override
           public synchronized void start() throws IOException {
+            super.start();
```
dschneider-pivotal: Why did this method change? It used to bind first and then called super.start. Now it calls start first and then does a bind which does not seem correct.

PR #4237, opened 2019-10-28
GEODE-7370: Upgrade ClassGraph to 4.8.52

kirklund:
Upgrading ClassGraph from 4.0.6 to 4.8.52 fixes a memory leak that
appeared after introducing geode-log4j even though logging does not
use ClassGraph.


kirklund:
These tests need to be updated for the dependency upgrade:
* geode-assembly:integrationTest:GeodeDependencyJarIntegrationTest
* geode-assembly:integrationTest:AssemblyContentsIntegrationTest

PR #4253, opened 2019-10-30
GEODE-7369: Deprecate SystemFailure Class

Bill:
[GEODE-7369](https://issues.apache.org/jira/browse/GEODE-7369)

Deprecate SystemFailure class.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/SystemFailure.java
+++ b/geode-core/src/main/java/org/apache/geode/SystemFailure.java
@@ -24,6 +24,12 @@ import org.apache.geode.internal.cache.GemFireCacheImpl;
 import org.apache.geode.logging.internal.executors.LoggingThread;
 
 /**
+ *
+ * @deprecated since Geode 1.11 since it is potentially counterproductive to try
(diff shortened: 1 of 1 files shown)


geode-core/src/main/java/org/apache/geode/SystemFailure.java
```diff
@@ -142,7 +143,14 @@
  * </pre>
  *
  * @since GemFire 5.1
+ *
+ * @deprecated since Geode 1.11 since it is potentially counterproductive to try
```
upthewaterspout: "since Geode 1.11 since" doesn't quite parse. Maybe "since Geode 1.11. It is potentially ...."

Maybe we could recommend people use -XX:+ExitOnOutOfMemoryError if they are starting geode processes without using gfsh? We already set that (actually the older option of -XX:OnOutOfMemoryError=kill -KILL %p) from gfsh start server.

Bill: Improved the wording. Didn't add the recommendation about the JVM switch.

rhoughton-pivot:
Ah, I was too late!

PR #4257, opened 2019-10-30
GEODE-7384: Members's old Persistent Member Id should be removed

pivotal-eshu:
* Always remove old Persistent Member Id if persisting its new Id.

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/persistence/PersistenceAdvisorImpl.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/persistence/PersistenceAdvisorImpl.java
@@ -646,7 +646,7 @@ public class PersistenceAdvisorImpl implements InternalPersistenceAdvisor {
     }
   }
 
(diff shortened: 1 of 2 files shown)


geode-core/src/main/java/org/apache/geode/internal/cache/persistence/PersistenceAdvisorImpl.java
```diff
@@ -758,9 +758,7 @@ public void prepareNewMember(InternalDistributedMember sender, PersistentMemberI
 
       // The oldId and newId could be the same if the member is retrying a GII. See bug #42051
       if (oldId != null && !oldId.equals(newId)) {
-        if (initialized) {
-          memberRemoved(oldId, false);
-        }
+        memberRemoved(oldId, false);
```
agingade: I am not sure implication of not to wait for initialization; but looking at the code above this line; where a newId is added without checking for initialization; this seems to be fine.

PR #4270, opened 2019-11-01
GEODE-7400: Prevent RejectedExecutionException in FederatingManager

aaronlindsey:
Commit f0c96db changed the `FederatingManager` class so that it reuses the same `ExecutorService` between restarts. After that change, if we start the manager after previously starting and stopping it, we get `RejectedExecutionException` because it tries to invoke a task on the same `ExecutorService` which has been shut down.

This PR changes the `FederatingManager` so that it invokes a supplier to get a new `ExecutorService` each time it is started to prevent the `RejectedExecutionException`.

### DIFF
--- a/geode-core/src/integrationTest/java/org/apache/geode/management/internal/FederatingManagerConcurrencyIntegrationTest.java
+++ b/geode-core/src/integrationTest/java/org/apache/geode/management/internal/FederatingManagerConcurrencyIntegrationTest.java
@@ -24,6 +24,7 @@ import static org.mockito.Mockito.when;
 
 import java.net.UnknownHostException;
 import java.util.concurrent.ExecutorService;
+import java.util.function.Supplier;
 
 import org.junit.After;
 import org.junit.Before;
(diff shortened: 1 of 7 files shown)


geode-core/src/integrationTest/java/org/apache/geode/management/internal/FederatingManagerConcurrencyIntegrationTest.java
```diff
@@ -0,0 +1,111 @@
...
+    federatingManager.startManager();
+  }
+
+  @After
+  public void tearDown() {
+    cache.close();
+  }
+
+  @Test
+  public void testFederatingManagerConcurrency() throws UnknownHostException {
```
jhuynh1: Sorry, I am unable to figure out how this is doing operations concurrently, would you be able to explain?  Also, what are we asserting, that the manager can add all the members without exception?

aaronlindsey: @jhuynh1 This was an existing test... We didn't modify this class except for renaming it and changing the method signature in `FederatingManagerFactoryWithMockMessenger`. If you look at just the first commit, you will be able to tell. Sorry for the confusion — that's just how git handles renames and is why we did the first commit separately.

aaronlindsey: The concurrency comes into play because each invocation of `FederatingManager.addMember()` schedules a GII task to be run asynchronously on a thread pool. This test is scheduling a hundred of them.

PR #4276, opened 2019-11-04
GEODE-7403: Test LogConsumer with log levels in all VM types

kirklund:
Adds new tests for all log levels within all DUnit VM types to verify
that the DUnit grep for suspect strings mechanism works as expected.

I'm seeing the locator VM log a stack trace for "Context initialization
failed" with a root cause of ClassNotFoundException LogService.
For some reason, the dunit grep for suspect strings is ignoring it.

This PR adds unit and dunit tests to ensure that the string itself can be
matched by LogConsumer.

GEODE-7403 is closed, but this test is still good to have.


kirklund:
Adding more tests

PR #4291, opened 2019-11-07
GEODE-7337: Adding DescribeQueryService Command

BenjaminPerryRoss:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/cache/query/security/RegExMethodAuthorizer.java
+++ b/geode-core/src/main/java/org/apache/geode/cache/query/security/RegExMethodAuthorizer.java
@@ -85,7 +85,8 @@ public final class RegExMethodAuthorizer implements MethodInvocationAuthorizer {
    *
    * @return an unmodifiable view of the regular expressions used to configure this authorizer.
    */
-  public Set<String> getAllowedPatterns() {
+  @Override
+  public Set<String> getParameters() {
     return allowedPatterns;
   }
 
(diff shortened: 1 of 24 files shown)


geode-core/src/main/java/org/apache/geode/cache/query/security/RegExMethodAuthorizer.java
```diff
@@ -85,7 +85,8 @@
    *
    * @return an unmodifiable view of the regular expressions used to configure this authorizer.
    */
-  public Set<String> getAllowedPatterns() {
+  @Override
+  public Set<String> getParameters() {
```
jhuynh1: Not quite related to this diff... but the javadoc mentions that the returned view/set is unmodifiable.  Should this be wrapped into an actual unmodifiable collection instead of the real set?  I guess the same question would be for all of the getParameter methods.  Or we modify the javaDoc.

geode-core/src/main/java/org/apache/geode/management/internal/cli/functions/AlterQueryServiceFunction.java
```diff
@@ -65,6 +65,7 @@ public CliFunctionResult executeFunction(FunctionContext<Object[]> context) {
         ((InternalCache) cache)
             .getService(org.apache.geode.cache.query.internal.QueryConfigurationService.class)
             .updateMethodAuthorizer(cache, authorizerName, parameterSet);
+        System.out.println("DONAL: Updated authorizer to " + authorizerName);
```
jhuynh1: Remove comment ? :-)

DonalEvans: Done

geode-core/src/main/java/org/apache/geode/cache/query/security/RegExMethodAuthorizer.java
DonalEvans: When the `allowedPatterns` set (and the `allowedPackages` set in the JavaBeanAccessorMethodAuthorizer) are assigned, it's as a `Collections.UnmodifiableSet` object, and there are unit tests to confirm that the output of these methods are unmodifiable.

PR #4294, opened 2019-11-07
GEODE-7229: adding autoSerializer to pdx and removing the persistent attribute

dschneider-pivotal:
The Pdx configuration class no longer has the persistent attribute and now has a new AutoSerializer attribute.
The persistent attribute was not needed because it is implied by having a non-null diskStoreName.
The autoSerializer attribute allow the ReflectionBasedAutoSerializer to be used more easily.
Before it would look something like this in json: pdxSerializer:{className:"org.apache.geode.pdx.ReflectionBasedAutoSerializer", initProperties:{"classes", "pat1,pat2"}}
Now in json: autoSerializer: {patterns:["pat1", "pat2"]}


dschneider-pivotal:
@jmelchio

geode-management/src/main/java/org/apache/geode/management/configuration/Pdx.java
```diff
@@ -31,8 +32,8 @@
 
   private Boolean readSerialized;
   private ClassName pdxSerializer;
+  private AutoSerializer autoSerializer;
   private Boolean ignoreUnreadFields;
-  private Boolean persistent;
```
jinmeiliao: getting rid of "persistent" is good, but makes it harder for user to use "DEFAULT"  diskstore. Let's add a method called `setPersistent()`, which set the diskstore name to "DEFAULT"

geode-management/src/main/java/org/apache/geode/management/configuration/Pdx.java
```diff
@@ -31,8 +32,8 @@
 
   private Boolean readSerialized;
   private ClassName pdxSerializer;
+  private AutoSerializer autoSerializer;
```
jinmeiliao: I believe we don't need to save this autoSerializer, all it does is to set a default pdx serializer, it can just directly modify the pdxSerializer variable. But we do need am internal flag indicating that setAutoSerializer method is called, and when user call setPdxSerializer, we should check that flag and report error if that's true. This would get rid of autoserilizaer conversion/validation code.

geode-management/src/main/java/org/apache/geode/management/configuration/Pdx.java
dschneider-pivotal: In this particular case, we should not make it easy to use the default disk store. The PCC team, for example, recently went through a difficult upgrade to switch pdx from the default disk store to a dedicated pdx disk store. Having the pdx types stores in the same disk store as the region data can cause issues down the road. So this particular change I think is good because it forces a user to explicitly say "I want my pdx to use the DEFAULT disk store".

If we decide to make it easy to use the default disk store for pdx then I'm not sure I like doing it by adding a setPersistent() method. This would only be of help to users of the java management api; not the rest management api users. I think we want to keep a close correspondence between the java and rest management apis. 

I also don't like us having methods that look like a settor that have no corresponding attribute. Users expect if they call setPersistent and then print out the config to see something like "persistent=true" not "diskStoreName=DEFAULT".

geode-management/src/main/java/org/apache/geode/management/configuration/Pdx.java
dschneider-pivotal: I think we do need this make it easy to configure a ReflectionBasedAutoSerializer from JSON. If all we have is a "pdxSerializer" then in JSON the user need to specify the ReflectionBasedAutoSerializer as a ClassName. See the initial comment on this PR for a "before and after" of what the JSON looks like for this. I think I may be missing something so lets discuss this further.

jinmeiliao: yes, you can still do that with this. I can pair with you

geode-management/src/main/java/org/apache/geode/management/configuration/Pdx.java
jinmeiliao: after discussion with Gang, we agreed that it would be a good idea to forcing user to specify a diskstore name since we don't want them to use "DEFAULT" without justifying it.

geode-management/src/main/java/org/apache/geode/management/configuration/Pdx.java
jinmeiliao: Since we want to show back the user exactly the way user configures it, then we still need the converter. Darrel is going to move the validation of mutual exclusion to the configuration objects

PR #4305, opened 2019-11-08
Increment stress-new-test from 2h15m to 6h to accomodate long tests

rhoughton-pivot:
We have had issues with UpgradeTest changes blowing out the Concourse timeout in StressNewTest. Bump the timeout limit.


ci/pipelines/shared/jinja.variables.yml
```diff
@@ -122,7 +122,7 @@ tests:
   CALL_STACK_TIMEOUT: '7200'
   CPUS: '96'
   DUNIT_PARALLEL_FORKS: '24'
-  EXECUTE_TEST_TIMEOUT: 2h15m
+  EXECUTE_TEST_TIMEOUT: 6h
```
onichols-pivotal: please also increase the call stack timeout.  it should always be set 15 minutes less than the new timeout, e.g. 20700

PR #4306, opened 2019-11-09
GEODE-7431: limit Benchmarks to three-at-a-time to stay within quota

onichols-pivotal:
For example, https://concourse.apachegeode-ci.info/teams/main/pipelines/apache-develop-main/jobs/Benchmark/builds/677 failed because there were already 3 benchmarks jobs running at the time.

Since the quota is 40 and each benchmark job uses 12, this PR restricts `max_in_flight` to `3`

### DIFF
--- a/ci/pipelines/geode-build/jinja.template.yml
+++ b/ci/pipelines/geode-build/jinja.template.yml
@@ -406,6 +406,7 @@ jobs:
 
 - name: Benchmark
   public: true
+  max_in_flight: 3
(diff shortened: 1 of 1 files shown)


pivotal-jbarrett:
We could just bump the limit too.

onichols-pivotal:
I don't have that power, but if you do, I will be happy to amend this PR to correspond to 1/12 of the new quota.

PR #4310, opened 2019-11-11
GEODE-7414: Added Client Hello extension

mivanac:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/net/SocketCreator.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/net/SocketCreator.java
@@ -1078,15 +1078,15 @@ public class SocketCreator {
       SSLParameters modifiedParams =
           checkAndEnableHostnameValidation(sslSocket.getSSLParameters());
 
-      String sysId = "";
+      String serverSSLNameExtension = "";
       InternalDistributedSystem system = InternalDistributedSystem.getAnyInstance();
       if (system != null) {
-        sysId = system.getConfig().getClientHelloExtension();
+        serverSSLNameExtension = system.getConfig().getSSLServerNameExtension();
       }
 
-      if (sysId != "") {
+      if (serverSSLNameExtension != "") {
         List<SNIServerName> serverNames = new ArrayList<>(1);
-        SNIHostName serverName = new SNIHostName(sysId);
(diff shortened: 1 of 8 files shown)


geode-core/src/main/java/org/apache/geode/internal/net/SocketCreator.java
```diff
@@ -1076,6 +1080,20 @@ private void configureClientSSLSocket(Socket socket, int timeout) throws IOExcep
 
       SSLParameters modifiedParams =
           checkAndEnableHostnameValidation(sslSocket.getSSLParameters());
+
+      String sysId = "";
+      InternalDistributedSystem system = InternalDistributedSystem.getAnyInstance();
+      if (system != null) {
+        sysId = system.getConfig().getClientHelloExtension();
```
jinmeiliao: 1. If this property is ssl related, it should start with "ssl-"
1. I don't see the connection between the 'client-hello-extension' and 'sysId', probably a better name for this property?
1. adding a config property probably would need a discussion thread on the dev list to gather feedback first.

mivanac:
This is only solution with updated config parameter. Start working on Jens proposal.

mivanac:
Idea is that ssl-server-name-extension will contain generic string value which will be passed as Server Name extension in Client Hello message.  Could you clarify your idea. What will be content of ssl-parameter-extension parameter?

jdeppe-pivotal:
If the user wishes to manipulate `SSLParameter`s (say in order to add their own SNI value) they would need to implement an interface something like:
```
public interface SSLParameterExtension {
  SSLParameter modify(SSLParameter, SSLConfig);
}
```
On starting Geode, they would provide the following parameter:
```
ssl-parameter-extension=com.example.ssl.MySSLParams.class
```
When the `SSLConfig` class is created it would also create (and hold a reference to) an instance of this class. At the point where SSLParameters need to be adjusted, this class' `modify` method could be called in order to perform whatever changes the user wants.

Depending on your particular use case you may need to also have the interface provide an `init()` method with whatever parameters are necessary for `modify` to do what you want - perhaps `Cache` or `DistributionConfig` - something fairly general so that additional use cases would likely also be satisfied in the future.

PR #4317, opened 2019-11-12
GEODE-7442: Fix RegionAttributesCreation

jujoramos:
- Do not automatically override the 'off-heap' property when
  configuring the 'PartitionAttributes' declaratively.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/integrationTest/java/org/apache/geode/internal/cache/xmlcache/CacheXmlGeneratorIntegrationTest.java
+++ b/geode-core/src/integrationTest/java/org/apache/geode/internal/cache/xmlcache/CacheXmlGeneratorIntegrationTest.java
@@ -1,0 +1,95 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.apache.geode.internal.cache.xmlcache;
+
+import static org.apache.geode.distributed.ConfigurationProperties.OFF_HEAP_MEMORY_SIZE;
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.io.File;
+import java.io.FileWriter;
+import java.io.PrintWriter;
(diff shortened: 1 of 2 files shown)


geode-core/src/integrationTest/java/org/apache/geode/internal/cache/xmlcache/CacheXmlGeneratorIntegrationTest.java
```diff
@@ -0,0 +1,95 @@
...
+        .withProperty("cache-xml-file", cacheXmlFile.getAbsolutePath()).startServer();
+
+    Cache cache = serverStarterRule.getCache();
+    assertThat(cache).isNotNull();
+    Region<Object, Object> region = cache.getRegion(testName.getMethodName());
+    assertThat(region).isNotNull();
+    assertThat(region.getAttributes().getOffHeap()).isTrue();
+    PartitionAttributes parsedAttributes = region.getAttributes().getPartitionAttributes();
+    assertThat(parsedAttributes.getLocalMaxMemory())
+        .isEqualTo(parsedAttributes.getLocalMaxMemory());
```
BenjaminPerryRoss: I might be misunderstanding the purpose of these assertions, but should these assertions be comparing the values stored in "parsedAttributes" (the values we retrieve from the region) to the values in "partitionedAttributes" (the attributes initialized as a class field at the beginning of the test)? At the moment it looks like we're comparing the values in "parsedAttributes" to themselves which should always be true.

agingade: +1 looks like it has to compare partitionAttributes and parsedAttributes

geode-core/src/main/java/org/apache/geode/internal/cache/xmlcache/RegionAttributesCreation.java
```diff
@@ -1576,7 +1576,6 @@ public void setPartitionAttributes(PartitionAttributes partitionAttr) {
       }
 
       setHasPartitionAttributes(true);
-      ((PartitionAttributesImpl) partitionAttributes).setOffHeap(offHeap);
```
agingade: Don't know much about this code; but looking at the new test added; it looks like the offHeap property is getting set in some other place.

geode-core/src/integrationTest/java/org/apache/geode/internal/cache/xmlcache/CacheXmlGeneratorIntegrationTest.java
jujoramos: Great catch!, thank you both!.

geode-core/src/main/java/org/apache/geode/internal/cache/xmlcache/RegionAttributesCreation.java
jujoramos: @agingade : yes, `offHeap` can't be effectively set until the member is up and running and we have a distributed system created. I've wrongly introduced this line here while working on [GEODE-7162](https://issues.apache.org/jira/browse/GEODE-7162).

PR #4318, opened 2019-11-12
GEODE-7428: change log level in management API

jmelchio:
Simple change to log level in management API to avoid logs being too chatty

### For all changes:

### Note:

### DIFF
--- a/geode-web-management/src/main/java/org/apache/geode/management/internal/rest/ManagementLoggingFilter.java
+++ b/geode-web-management/src/main/java/org/apache/geode/management/internal/rest/ManagementLoggingFilter.java
@@ -37,6 +37,11 @@ public class ManagementLoggingFilter extends OncePerRequestFilter {
   protected void doFilterInternal(HttpServletRequest request, HttpServletResponse response,
       FilterChain filterChain) throws ServletException, IOException {
 
+    if (!logger.isDebugEnabled()) {
+      filterChain.doFilter(request, response);
+      return;
+    }
+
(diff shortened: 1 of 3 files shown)


jmelchio:
@jinmeiliao @onichols-pivotal @dschneider-pivotal @dhemery

geode-web-management/src/main/java/org/apache/geode/management/internal/rest/ManagementLoggingFilter.java
```diff
@@ -31,21 +31,12 @@
...
 
   @Override
   protected void doFilterInternal(HttpServletRequest request, HttpServletResponse response,
       FilterChain filterChain) throws ServletException, IOException {
 
-    if (!ENABLE_REQUEST_LOGGING) {
-      filterChain.doFilter(request, response);
-      return;
-    }
-
```
onichols-pivotal: This change seems to have an adverse performance impact.  Even if debug logging is not enabled, we will now always wrap and capture the request and response.  Before, the wrapping and capturing was able to be skipped.  Why not keep the same logic as before, just change the test to from system-property check to isDebugEnabled check?

jmelchio: Yes, that's a good point, I'll make that change.

onichols-pivotal:
looks like ManagementRequestLoggingDistributedTest needs to be updated

PR #4320, opened 2019-11-13
GEODE-7409: get latest PDX configuration of cluster in REST API for M…

jinmeiliao:
…anagement

Co-authored-by: Joris Melchior <jmelchior@pivotal.io>

### For all changes:

### Note:


geode-core/src/main/java/org/apache/geode/management/internal/configuration/realizers/ConfigurationRealizer.java
```diff
@@ -29,15 +29,25 @@
  */
 @Experimental
 public interface ConfigurationRealizer<T extends AbstractConfiguration<R>, R extends RuntimeInfo> {
-  RealizationResult create(T config, InternalCache cache);
+  default RealizationResult create(T config, InternalCache cache) {
```
dschneider-pivotal: Having most all of the methods on an interface have default implementations that throw an exception seems like an anti-pattern. The default impl of isReadOnly makes sense but I think the old create/update/delete was better when each implementer of the interface was forced to deal with how they would implement those methods. Now it seems like it is really easy to implement ConfigurationRealizer but it will blow up at runtime.

geode-management/src/main/java/org/apache/geode/management/runtime/PdxInfo.java
```diff
@@ -0,0 +1,67 @@
...
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package org.apache.geode.management.runtime;
+
+import org.apache.geode.annotations.Experimental;
+
+@Experimental
+public class PdxInfo extends RuntimeInfo {
```
dschneider-pivotal: I think PdxInfo should be in sync with the configuration Pdx class.
So they should both name "diskStore"/"diskStoreName" the same.
PdxInfo should not have a "persistent" field.
PdxInfo should have "autoSerializer" and "pdxSerializer" instead of just "serializerClassName".
I think this is important for both the java apis and rest.

geode-web-management/src/main/java/org/apache/geode/management/internal/rest/ManagementLoggingFilter.java
```diff
@@ -37,10 +37,10 @@
   protected void doFilterInternal(HttpServletRequest request, HttpServletResponse response,
       FilterChain filterChain) throws ServletException, IOException {
 
-    if (!logger.isDebugEnabled()) {
-      filterChain.doFilter(request, response);
-      return;
-    }
+    // if (!logger.isDebugEnabled()) {
```
dschneider-pivotal: remove this dead code?

geode-web-management/src/main/java/org/apache/geode/management/internal/rest/ManagementLoggingFilter.java
```diff
@@ -70,7 +70,7 @@ private void logRequest(HttpServletRequest request, ContentCachingRequestWrapper
     }
     String payload = getContentAsString(wrappedRequest.getContentAsByteArray(),
         wrappedRequest.getCharacterEncoding());
-    logger.debug(String.format(requestPattern, request.getMethod(), requestUrl,
+    logger.info(String.format(requestPattern, request.getMethod(), requestUrl,
```
dschneider-pivotal: Why was this changed back to info?

jinmeiliao: sorry, this was not meant to be checked in

geode-management/src/main/java/org/apache/geode/management/runtime/PdxInfo.java
jinmeiliao: Agree that they should have consistent names. But for persistent field, because there could be a case when a server can have a diskstore, but not persistent, in reality, we should report that case on a member. This won't happen in configuration, but could happen on a server

dschneider-pivotal: When that happens (i.e. persistent is false and diskStore != null) then the product ignores the diskStore. So if we said the PdxInfo will show the diskStore used by pdx then in this case it would be correct to return null since pdx is not using a disk store. I think this is reasonable and allows us not to expose the old way of configuring it with two attributes.

jinmeiliao: Ok, updated the realizer to include this logic

geode-core/src/main/java/org/apache/geode/management/internal/configuration/realizers/ConfigurationRealizer.java
jinmeiliao: used a ReadonlyRealizer to do this.

geode-core/src/main/java/org/apache/geode/management/internal/configuration/realizers/PdxRealizer.java
```diff
@@ -25,12 +25,12 @@
...
     PdxInfo info = new PdxInfo();
     info.setReadSerialized(cache.getPdxReadSerialized());
-    info.setDiskStore(cache.getPdxDiskStore());
+    info.setDiskStoreName(cache.getPdxDiskStore());
     info.setIgnoreUnreadFields(cache.getPdxIgnoreUnreadFields());
     info.setPersistent(cache.getPdxPersistent());
     PdxSerializer pdxSerializer = cache.getPdxSerializer();
     if (pdxSerializer != null) {
-      info.setSerializerClassName(pdxSerializer.getClass().getName());
+      info.setPdxSerializer(pdxSerializer.getClass().getName());
```
dschneider-pivotal: If "pdxSerializer" is an instanceof ReflectionBasedAutoSerializer then it should show up in PdxInfo as an AutoSerializer. You can get the patterns and portable values by calling "getConfig()" on the ReflectionBasedAutoSerializer.
The other thing that is missing here is the initParameters when it is not a ReflectionBasedAutoSerializer. You can not get these given a PdxSerializer. But they are easy to get on the locator from the cluster config. On the locator are you actually showing the "Pdx" instance under configuration and "PdxInfo" is only used on non-locators? The bad thing about not showing the initParameters is that it might be the one thing that changed that has not yet taken effect on a server but you would not be able to tell. Would it be possible for  this command to display something about each member whose pdx config is outdated and needs a restart?

dschneider-pivotal: Since the cluster config pdx will be show in the result under configuration, I'm okay with the PdxInfo have less information.

PR #4324, opened 2019-11-13
GEODE-6991: Create alter query-service gfsh command

DonalEvans:

### For all changes:

### Note:


geode-core/src/main/java/org/apache/geode/management/internal/cli/commands/AlterQueryServiceCommand.java
```diff
@@ -0,0 +1,209 @@
...
+  static final String SINGLE_PARAM_AND_PARAMETERS_SPECIFIED_MESSAGE =
+      "Only one of '--" + AUTHORIZER_PARAMETERS + "' and '--" + SINGLE_AUTHORIZER_PARAMETER
+          + "' options can be specified.";
+  static final String NO_ARGUMENTS_MESSAGE =
+      "No arguments were provided. No changes have been applied.";
+  static final String NO_MEMBERS_FOUND_MESSAGE = "No members found.";
+  static final String SECURITY_NOT_ENABLED_MESSAGE =
+      "Integrated security is not enabled for this distributed system. Updating the method authorizer requires integrated security to be enabled.";
+  public static final String PARTIAL_FAILURE_MESSAGE =
+      "In the event of a partial failure of this command, re-running the command or restarting failing members should restore consistency.";
```
jujoramos: Some constants are not used outside of the package, can you make them `private`?.

geode-core/src/main/java/org/apache/geode/management/internal/cli/commands/AlterQueryServiceCommand.java
```diff
@@ -0,0 +1,209 @@
...
+    QueryConfigService queryConfigService = getQueryConfigService();
+    Set<String> parametersSet = null;
+    if (methodAuthorizerName != null) {
+      if (!isSecurityEnabled()) {
+        return ResultModel.createError(SECURITY_NOT_ENABLED_MESSAGE);
+      }
+      if (Boolean.parseBoolean(
+          System.getProperty(GEMFIRE_PREFIX + "QueryService.allowUntrustedMethodInvocation"))) {
+        return ResultModel.createError(DEPRECATED_PROPERTY_ERROR);
+      }
```
jujoramos: According to my understanding this condition will always be `false`: the property `QueryService.allowUntrustedMethodInvocation` is set on servers only, not on the JVM on which the `gfsh` command is executed, so you can safely remove the whole code block.

geode-core/src/main/java/org/apache/geode/management/internal/cli/commands/AlterQueryServiceCommand.java
```diff
@@ -0,0 +1,209 @@
...
+      interceptor = "org.apache.geode.management.internal.cli.commands.AlterQueryServiceCommand$Interceptor")
+  @ResourceOperation(resource = ResourcePermission.Resource.CLUSTER,
+      operation = ResourcePermission.Operation.MANAGE)
+  public ResultModel execute(
+      @CliOption(key = METHOD_AUTHORIZER_NAME,
+          help = METHOD_AUTHORIZER_NAME_HELP) String methodAuthorizerName,
+      @CliOption(key = AUTHORIZER_PARAMETERS,
+          help = AUTHORIZER_PARAMETERS_HELP) String[] authorizerParameters,
+      @CliOption(key = SINGLE_AUTHORIZER_PARAMETER,
+          help = SINGLE_AUTHORIZER_PARAMETER_HELP) String singleAuthorizerParameter) {
```
jujoramos: We end up using a `Set<String>` internally, so I'd merge these two properties into just one as it might confuse users.

geode-core/src/main/java/org/apache/geode/management/internal/cli/commands/AlterQueryServiceCommand.java
```diff
@@ -0,0 +1,209 @@
...
+      List<CliFunctionResult> functionResults =
+          executeAndGetFunctionResult(new AlterQueryServiceFunction(), args,
+              targetMembers);
+      String footer = null;
+      for (CliFunctionResult cliResult : functionResults) {
+        if (!cliResult.isSuccessful()) {
+          footer = PARTIAL_FAILURE_MESSAGE;
+          break;
+        }
+      }
```
jujoramos: What if the update fails on all members?, we shouldn't show the `PARTIAL_FAILURE_MESSAGE`, right?.

geode-core/src/main/java/org/apache/geode/management/internal/cli/commands/AlterQueryServiceCommand.java
```diff
@@ -0,0 +1,209 @@
...
+      result = ResultModel.createInfo(NO_MEMBERS_FOUND_MESSAGE);
+    }
+    result.setConfigObject(queryConfigService);
+    return result;
+  }
+
+  boolean isSecurityEnabled() {
+    return ((InternalCache) CacheFactory.getAnyInstance()).getSecurityService()
+        .isIntegratedSecurity();
+  }
```
jujoramos: I don't know this part of the code so the question might make no sense at all: here we're checking wether security is enabled for the JVM on which the gfsh command is running, shouldn't we check wether security is enabled at the cluster level (locator) instead?.

jinmeiliao: I am not sure how this methodAuthorizer is related to the securityService, but commands are executed on jmxManager which is usually a locator or a server with jmx-manager enabled (in the case where is no locator involved). When security service is enabled, it should be enabled on all members (locators and servers).

geode-core/src/main/java/org/apache/geode/management/internal/cli/commands/AlterQueryServiceCommand.java
DonalEvans: It's necessary to provide an option that is parsed as a single String (as opposed to a String array) because if a user specifies `AUTHORIZER_PARAMETERS` with an argument that includes a comma as part of one of the parameters (for example, in a regex using the X{n,m} quantifier) then gfsh will split that argument on the comma as part of its parsing. That being said, if there's a way to ensure that user input will only be split on the "correct" commas then having only one option would definitely be preferable.

geode-core/src/main/java/org/apache/geode/management/internal/cli/commands/AlterQueryServiceCommand.java
DonalEvans: Done!

geode-core/src/main/java/org/apache/geode/management/internal/cli/commands/AlterQueryServiceCommand.java
DonalEvans: You're right, this check should be moved to the `AlterQueryServiceFunction` class.

geode-core/src/main/java/org/apache/geode/management/internal/cli/commands/AlterQueryServiceCommand.java
DonalEvans: You're right, we can change this to only be shown in the event of a partial failure.

geode-core/src/main/java/org/apache/geode/management/internal/cli/commands/AlterQueryServiceCommand.java
jhuynh1: Is this a correct statement - There is no way for a user to supply multiple regex's with commas in each regex.  This solution only allows for one regex with commas?

geode-core/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/AlterQueryServiceCommandWithSecurityDUnitTest.java
```diff
@@ -0,0 +1,256 @@
...
+        .hasTableSection().hasRowSize(serversToStart);
+
+    verifyCurrentAuthorizerClass(regExAuthorizerClass);
+    servers.forEach(server -> server.invoke(() -> {
+      RegExMethodAuthorizer methodAuthorizer = (RegExMethodAuthorizer) Objects
+          .requireNonNull(ClusterStartupRule.getCache()).getService(QueryConfigurationService.class)
+          .getMethodAuthorizer();
+
+      assertThat(methodAuthorizer.getAllowedPatterns().size()).isEqualTo(1);
+      assertThat(methodAuthorizer.getAllowedPatterns()).containsOnly(REGEX_AUTHORIZER_PARAMETER);
```
jhuynh1: I feel at this point, this could be broken into a seperate test as we are now testing that the pattern and the Regex specific authorizor

geode-core/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/AlterQueryServiceCommandWithSecurityDUnitTest.java
```diff
@@ -0,0 +1,256 @@
...
+          .requireNonNull(ClusterStartupRule.getCache()).getService(QueryConfigurationService.class)
+          .getMethodAuthorizer();
+
+      Set<String> expectedParameters =
+          new HashSet<>(Arrays.asList(JAVA_BEAN_AUTHORIZER_PARAMS.split(",")));
+      assertThat(methodAuthorizer.getAllowedPackages()).isEqualTo(expectedParameters);
+    }));
+
+    authorizerName = regExAuthorizerClass.getName();
+    gfsh.executeAndAssertThat(COMMAND_NAME + " --" + METHOD_AUTHORIZER_NAME + "=" + authorizerName
```
jhuynh1: Maybe break into a separate helper method since we do the same 3-ish steps multiple times in this test.

geode-core/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/AlterQueryServiceCommandWithSecurityDUnitTest.java
```diff
@@ -0,0 +1,256 @@
...
+        .hasTableSection().hasRowSize(serversToStart);
+
+    verifyCurrentAuthorizerClass(authorizerClass);
+    servers.forEach(server -> server.invoke(() -> {
+      TestMethodAuthorizer methodAuthorizer = (TestMethodAuthorizer) Objects
+          .requireNonNull(ClusterStartupRule.getCache()).getService(QueryConfigurationService.class)
+          .getMethodAuthorizer();
+
+      Set<String> expectedParameters =
+          new HashSet<>(Arrays.asList(USER_AUTHORIZER_PARAMETERS.split(",")));
```
jhuynh1: This might be able to be broken up into a test to validate that the parameters are split correctly.

geode-core/src/main/java/org/apache/geode/management/internal/cli/commands/AlterQueryServiceCommand.java
```diff
@@ -0,0 +1,209 @@
...
+      String footer = null;
+      for (CliFunctionResult cliResult : functionResults) {
+        if (!cliResult.isSuccessful()) {
+          footer = PARTIAL_FAILURE_MESSAGE;
+          break;
+        }
+      }
+      result =
+          ResultModel.createMemberStatusResult(functionResults, null, footer, false, true);
+    } else {
```
jhuynh1: Personal preference, maybe flip the if to be if (targetMembers.size() == 0) { //short block of code; } else {//big block of code}

geode-core/src/main/java/org/apache/geode/management/internal/cli/commands/AlterQueryServiceCommand.java
```diff
@@ -0,0 +1,209 @@
...
+
+  void populateMethodAuthorizer(String methodAuthorizerName, Set<String> parameterSet,
+      QueryConfigService queryConfigService) {
+    QueryConfigService.MethodAuthorizer methodAuthorizer =
+        new QueryConfigService.MethodAuthorizer();
+    methodAuthorizer.setClassName(methodAuthorizerName);
+
+    if (parameterSet != null && parameterSet.size() != 0) {
+      List<QueryConfigService.MethodAuthorizer.Parameter> parameters = new ArrayList<>();
+      for (String param : parameterSet) {
```
jhuynh1: this is super nitpicky and silly but there is param, parameter and parameters.  Maybe change param to value?  Just took a bit for me to see where param was used with parameter is param is a parameter... kinda..

geode-core/src/main/java/org/apache/geode/management/internal/cli/commands/AlterQueryServiceCommand.java
DonalEvans: If cluster config is enabled then checking the locator is sufficient, but if cluster config is not enabled then it's possible for a locator to no have a security service but the servers that connect to it to have one, leading to incorrect behaviour. This check has been moved to the `AlterQueryServiceFunction` class to ensure correct behaviour.

geode-core/src/main/java/org/apache/geode/management/internal/cli/commands/AlterQueryServiceCommand.java
DonalEvans: @jhuynh1 That is correct. If there is a non-splitting comma in the argument, then the `authorizer-parameter` option must be specified, otherwise gfsh will treat that comma as a point to split the input string on.

geode-core/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/AlterQueryServiceCommandWithSecurityDUnitTest.java
DonalEvans: Done!

geode-core/src/main/java/org/apache/geode/management/internal/cli/commands/AlterQueryServiceCommand.java
DonalEvans: Done!

geode-core/src/main/java/org/apache/geode/management/internal/cli/commands/AlterQueryServiceCommand.java
DonalEvans: Done!

geode-core/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/AlterQueryServiceCommandWithSecurityDUnitTest.java
DonalEvans: These lines are just checking that the value for `authorizer-parameter` specified via the gfsh command is correctly set on the method authorizer that is present on the servers. It's not testing the behaviour or functionality of that method authorizer at all.

geode-core/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/AlterQueryServiceCommandWithSecurityDUnitTest.java
DonalEvans: The parameters are split automatically by gfsh when it parses the `authorizer-parameters` argument. We split the input string here to generate the expected contents of the set of strings that is stored in the method authorizer rather than manually creating the set so that they match.

geode-core/src/main/java/org/apache/geode/management/internal/cli/commands/AlterQueryServiceCommand.java
jujoramos: Is there any option to escape characters within a `gfsh` command parameter?.
I'd certainly prefer the above instead of having two different parameters for the same thing with just an `s` to differentiate them, it's counter intuitive and will probably cause issues for our users. If there's no way of achieving it, however, I suggest using two totally different and very verbose names for the parameters so users can tell the difference right away.

DonalEvans: We decided that the best solution to this issue is to split the gfsh input parameters on semicolons rather than commas using `optionContext = "splittingRegex=;"` in the `@CliOption` annotation for `AUTHORIZER_PARAMETERS`. This allows multiple regexes to be specified, even if they contain commas, and prevents the need for the corner-case `SINGLE_AUTHORIZER_PARAMETER` option.

geode-core/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/AlterQueryServiceCommandWithSecurityDUnitTest.java
DonalEvans: A new test has been added to specifically verify that parameters are correctly set

geode-core/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/AlterQueryServiceCommandWithSecurityDUnitTest.java
DonalEvans: A new test has been added to specifically verify that parameters are correctly set

PR #4330, opened 2019-11-14
GEODE-7450 SSL peerAppDataBuffer expansion needs work

bschuchardt:
The expansion logic for the decrypted buffer in NioSslEngine was
expanding the buffer to twice its old capacity and wouldn't go
beyond that.  Some ciphers will compress data a lot more than that, so
we need to continually increase available space in the decryption buffer
if the SslEngine reports a BUFFER_OVERFLOW status.  The changes in
this commit do that by always doubling the available space in the
decryption buffer in response to a BUFFER_OVERFLOW.

Along with that I found it unnecessary to take preemptive action to
increase the size of the buffer and deleted that code.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/net/NioSslEngine.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/net/NioSslEngine.java
@@ -267,14 +267,17 @@ public class NioSslEngine implements NioFilter {
     // transmit large payloads and we may have read a partial chunk
     // during the previous unwrap
 
-    // it's better to be pro-active about avoiding buffer overflows
-    expandPeerAppData(wrappedBuffer);
     peerAppData.limit(peerAppData.capacity());
     while (wrappedBuffer.hasRemaining()) {
       SSLEngineResult unwrapResult = engine.unwrap(wrappedBuffer, peerAppData);
       switch (unwrapResult.getStatus()) {
         case BUFFER_OVERFLOW:
-          expandPeerAppData(wrappedBuffer);
+          // buffer overflow expand and try again - double the available decryption space
+          int newCapacity =
+              (peerAppData.capacity() - peerAppData.position()) * 2 + peerAppData.position();
+          peerAppData =
+              bufferPool.expandWriteBufferIfNeeded(TRACKED_RECEIVER, peerAppData, newCapacity);
+          peerAppData.limit(peerAppData.capacity());
           break;
         case BUFFER_UNDERFLOW:
           // partial data - need to read more. When this happens the SSLEngine will not have
@@ -289,14 +292,6 @@ public class NioSslEngine implements NioFilter {
     }
     wrappedBuffer.clear();
     return peerAppData;
-  }
-
-  void expandPeerAppData(ByteBuffer wrappedBuffer) {
-    if (peerAppData.capacity() - peerAppData.position() < 2 * wrappedBuffer.remaining()) {
-      peerAppData =
-          bufferPool.expandWriteBufferIfNeeded(TRACKED_RECEIVER, peerAppData,
-              expandedCapacity(wrappedBuffer, peerAppData));
-    }
   }
 
   @Override
@@ -321,9 +316,6 @@ public class NioSslEngine implements NioFilter {
         peerAppData.compact();
         peerAppData.flip();
       }
-    } else {
-      peerAppData =
-          bufferPool.expandReadBufferIfNeeded(TRACKED_RECEIVER, peerAppData, bytes);
     }
 
     while (peerAppData.remaining() < bytes) {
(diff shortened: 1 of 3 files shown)


geode-core/src/test/java/org/apache/geode/internal/net/NioSslEngineTest.java
```diff
@@ -265,16 +268,22 @@ public void unwrapWithBufferOverflow() throws Exception {
     spyNioSslEngine.engine = testEngine;
 
     testEngine.addReturnResult(
-        new SSLEngineResult(BUFFER_OVERFLOW, NEED_UNWRAP, netBytes.length, netBytes.length),
+        new SSLEngineResult(BUFFER_OVERFLOW, NEED_UNWRAP, 0, 0), // results in 30,000 byte buffer
```
echobravopapa: Would it be useful to have incremental assertions for each of these return results being added to the engine?

geode-core/src/test/java/org/apache/geode/internal/net/NioSslEngineTest.java
```diff
@@ -265,16 +268,22 @@ public void unwrapWithBufferOverflow() throws Exception {
...
+    expectedCapacity =
+        2 * (expectedCapacity - originalPeerAppDataPosition) + originalPeerAppDataPosition;
+    expectedCapacity =
+        2 * (expectedCapacity - originalPeerAppDataPosition) + originalPeerAppDataPosition;
     ByteBuffer unwrappedBuffer = spyNioSslEngine.unwrap(wrappedData);
     unwrappedBuffer.flip();
-
-    verify(spyNioSslEngine, times(2)).expandPeerAppData(any(ByteBuffer.class));
-    assertThat(unwrappedBuffer).isEqualTo(ByteBuffer.wrap(netBytes));
+    assertThat(unwrappedBuffer.capacity()).isEqualTo(expectedCapacity);
```
echobravopapa: i.e. we only check the capacity at the end, but above buffer grows from 30K to 90K bytes...

geode-core/src/test/java/org/apache/geode/internal/net/NioSslEngineTest.java
bschuchardt: I think you're supposed to do the work and then do assertions.  I'd have to break it into multiple tests to do that.  But why?  If the final result is wrong the intermediate steps failed.  It would only help with debugging, I think.

geode-core/src/test/java/org/apache/geode/internal/net/NioSslEngineTest.java
bschuchardt: yep

geode-core/src/main/java/org/apache/geode/internal/net/NioSslEngine.java
```diff
@@ -267,14 +267,17 @@ public synchronized ByteBuffer unwrap(ByteBuffer wrappedBuffer) throws IOExcepti
...
-    expandPeerAppData(wrappedBuffer);
     peerAppData.limit(peerAppData.capacity());
     while (wrappedBuffer.hasRemaining()) {
       SSLEngineResult unwrapResult = engine.unwrap(wrappedBuffer, peerAppData);
       switch (unwrapResult.getStatus()) {
         case BUFFER_OVERFLOW:
-          expandPeerAppData(wrappedBuffer);
+          // buffer overflow expand and try again - double the available decryption space
+          int newCapacity =
+              (peerAppData.capacity() - peerAppData.position()) * 2 + peerAppData.position();
```
dschneider-pivotal: I was wondering why you didn't just do this: newCapacity = peerAppData.capacity() * 2?
I was concerned that if position is close to capacity then this will cause it to grow very slowly.
If it is possible for position to be equal to capacity then this would never grow it and be an infinite loop.
I can't tell if either of these are possible.

bschuchardt: Hmm, I think you're right Darrel.  I had a minimum of 3/2 growth of the buffer's capacity but took that out, but if the position is near the buffer's capacity (or equal!) that could cause a problem.  I'll re-add that and write a test.

bschuchardt: int newCapacity =
              (peerAppData.capacity() - peerAppData.position()) * 2 + peerAppData.position();
          newCapacity = Math.max(newCapacity, peerAppData.capacity() / 2 * 3);
          peerAppData =
              bufferPool.expandWriteBufferIfNeeded(TRACKED_RECEIVER, peerAppData, newCapacity);

PR #4332, opened 2019-11-15
GEODE-7252: Create IncrementalBackupIntegrationTest

kirklund:
* Extract integration tests from IncrementalBackupDistributedTest
* Use TemporaryFolder for deploy directory
* Reduce unnecessary VM invocations in the DistributedTest

I also touched up some product code:
• Improved toString on ClassPathLoader and JarDeployer for debugging
• Removed an unnecessary cast of DistributionManager to ClusterDistributionManager (prevented testing the behavior with a loner)
• Replaced public static variable in TransformUtils with a public static getter

Verify that [6d3f82](https://github.com/apache/geode/commit/6d3f82) fixes GEODE-7252, GEODE-7293 and GEODE-7310.


kirklund:
@mhansonp That's correct. All but one test in the DistributedTest used a single JVM so I extracted those to create the new IntegrationTest.

PR #4343, opened 2019-11-18
GEODE-7436: Deploy jar using semantic versioning scheme

jinmeiliao:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/DeployedJar.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/DeployedJar.java
@@ -61,7 +61,7 @@ public class DeployedJar {
   private static final MessageDigest messageDigest = getMessageDigest();
   private static final Pattern PATTERN_SLASH = Pattern.compile("/");
 
-  private final String jarName;
+  private final String artifactId;
   private final File file;
   private final byte[] md5hash;
   private final Collection<Function> registeredFunctions = new ArrayList<>();
@@ -86,12 +86,12 @@ public class DeployedJar {
   /**
    * Writes the given jarBytes to versionedJarFile
    */
-  public DeployedJar(File versionedJarFile, final String jarName) {
-    Assert.assertTrue(jarName != null, "jarName cannot be null");
+  public DeployedJar(File versionedJarFile, final String artifactId) {
+    Assert.assertTrue(artifactId != null, "artifactId cannot be null");
     Assert.assertTrue(versionedJarFile != null, "versionedJarFile cannot be null");
 
     this.file = versionedJarFile;
-    this.jarName = jarName;
+    this.artifactId = artifactId;
 
     if (!hasValidJarContent(versionedJarFile)) {
       throw new IllegalArgumentException(
@@ -366,10 +366,13 @@ public class DeployedJar {
   }
 
   /**
-   * @return the unversioned name of this jar file, e.g. myJar.jar
+   * Get this jar's artifact ID, which is the part of the jar file name that precedes the version
+   * information.
+   *
+   * @return the artifact ID for this jar
    */
-  public String getJarName() {
-    return this.jarName;
+  public String getArtifactId() {
+    return this.artifactId;
   }
 
   public String getFileName() {
@@ -402,7 +405,7 @@ public class DeployedJar {
   public int hashCode() {
     final int prime = 31;
     int result = 1;
-    result = prime * result + (this.jarName == null ? 0 : this.jarName.hashCode());
+    result = prime * result + (this.artifactId == null ? 0 : this.artifactId.hashCode());
     return result;
   }
 
@@ -418,11 +421,11 @@ public class DeployedJar {
       return false;
     }
     DeployedJar other = (DeployedJar) obj;
-    if (this.jarName == null) {
-      if (other.jarName != null) {
+    if (this.artifactId == null) {
+      if (other.artifactId != null) {
         return false;
       }
-    } else if (!this.jarName.equals(other.jarName)) {
+    } else if (!this.artifactId.equals(other.artifactId)) {
       return false;
     }
     return true;
@@ -432,7 +435,7 @@ public class DeployedJar {
   public String toString() {
     final StringBuilder sb = new StringBuilder(getClass().getName());
     sb.append('@').append(System.identityHashCode(this)).append('{');
-    sb.append("jarName=").append(this.jarName);
+    sb.append("artifactId=").append(this.artifactId);
     sb.append(",file=").append(this.file.getAbsolutePath());
     sb.append(",md5hash=").append(toHex(this.md5hash));
     sb.append(",version=").append(this.getVersion());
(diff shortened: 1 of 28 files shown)


jinmeiliao:
@jmelchio

geode-core/src/main/java/org/apache/geode/internal/JarDeployer.java
```diff
@@ -50,17 +52,21 @@
...
-  public static final String JAR_PREFIX_FOR_REGEX = "";
+  // The pound version scheme predates the sequenced version scheme
+  private static final Pattern POUND_VERSION_SCHEME =
+      Pattern.compile("^vf\\.gf#(?<artifact>.*)\\.jar#(?<version>\\d+)$");
+  // The sequenced version scheme predates the semantic version scheme. We use this scheme when
+  // the file name has no semantic version
+  private static final Pattern SEQUENCED_VERSION_SCHEME =
+      Pattern.compile("(?<artifact>..*)\\.v(?<version>\\d++).jar$");
+  // We use this scheme when we detect that the file has semantic version information
+  private static final Pattern SEMANTIC_VERSION_SCHEME =
```
onichols-pivotal: Please consider using the official semver regex (the "one with numbered capture groups") from https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string

jinmeiliao: this is nice, but we don't really care about major.minor.patch.tags groups, all we care is the part before the version string.

onichols-pivotal: The version string IS the major.minor.patch.tags; we aren't guaranteed to get "the part before" correctly if we aren't matching the version part correctly...
For example, `javax.inject-2.4.0-b34.jar` or `spring-oxm-4.3.18.RELEASE.jar` or `hive-cli-1.2.1+spark2.jar` all have valid semantic versions.  Is it reasonable that Geode should be able to recognize the artifact ID correctly in these cases?  Why "re-invent" the semver regex when there is already an official published regex?

However, if the intent is explicitly to do something that differs from the semver standard, an explicit comment describing the versioning scheme _we_ recognize would be helpful otherwise the constant name `SEMANTIC_VERSION_SCHEME` might give the wrong impression.

geode-core/src/main/java/org/apache/geode/internal/DeployedJar.java
```diff
@@ -366,14 +368,23 @@ private Function newFunction(final Class<Function> clazz, final boolean errorOnN
...
-  public String getFileName() {
-    return this.file.getName();
+  /**
+   * @return the filename as user deployed, i.e remove the sequence number
+   */
+  public String getDeployedFileName() {
+    String fileName = this.file.getName();
+    Matcher matcher = SEQUENCED_VERSION_SCHEME.matcher(fileName);
+    matcher.find();
+    return matcher.group("artifact") + ".jar";
```
dschneider-pivotal: I find the group name "artifact" confusing here because this is not the same as what getArtifactId returns. In this case it is just the sequence number we remove. Would a better name for this group be "deployed"?

geode-core/src/main/java/org/apache/geode/internal/JarDeployer.java
dschneider-pivotal: Since semver always has three numeric fields and we want to support fewer than that I agree that it is confusing to name this SEMANTIC_VERSION_SCHEME. What would be a better name? EXPLICIT_VERSION_SCHEME? USER_VERSION_SCHEME?

geode-core/src/main/java/org/apache/geode/internal/JarDeployer.java
```diff
@@ -50,17 +48,21 @@
 public class JarDeployer implements Serializable {
   private static final long serialVersionUID = 1L;
   private static final Logger logger = LogService.getLogger();
-  public static final String JAR_PREFIX_FOR_REGEX = "";
+  // The pound version scheme predates the sequenced version scheme
+  private static final Pattern POUND_VERSION_SCHEME =
+      Pattern.compile("^vf\\.gf#(?<artifact>.*)\\.jar#(?<version>\\d+)$");
+  // The sequenced version scheme predates the semantic version scheme. We use this scheme when
```
dschneider-pivotal: These javadocs are now wrong. We always use this scheme even if it has a semantic version

geode-core/src/main/java/org/apache/geode/internal/JarDeployer.java
```diff
@@ -158,61 +156,55 @@ protected File getNextVersionedJarFile(String unversionedJarName) {
...
   /**
-   * Find all versions of the JAR file that are currently on disk and return them sorted from newest
-   * (highest version) to oldest
+   * get the artifact id from the existing files on the server. This will skip files that
+   * do not have sequence id or semantic version appended to them.
    *
-   * @param unversionedJarName Name of the JAR file that we want old versions of
-   * @return Sorted array of files that are older versions of the given JAR
+   * @param versionedJarFileName: the file names that exists on the server, could be file with
+   *        sequence numbers abc.v1.jar, or file with semantic version abc-1.0.0.jar
```
dschneider-pivotal: these javadocs are wrong. "abc-1.0.0.jar" should now also have a sequence number.

geode-core/src/main/java/org/apache/geode/internal/JarDeployer.java
```diff
@@ -158,61 +156,55 @@ protected File getNextVersionedJarFile(String unversionedJarName) {
...
+   * @return the artifact id, which could be optional for other forms of jar names
```
dschneider-pivotal: javadoc that it will return null if the input does not have a sequence id

geode-core/src/main/java/org/apache/geode/internal/JarDeployer.java
```diff
@@ -158,61 +156,55 @@ protected File getNextVersionedJarFile(String unversionedJarName) {
...
 
   /**
-   * Find all versions of the JAR file that are currently on disk and return them sorted from newest
-   * (highest version) to oldest
+   * get the artifact id from the existing files on the server. This will skip files that
+   * do not have sequence id or semantic version appended to them.
    *
-   * @param unversionedJarName Name of the JAR file that we want old versions of
-   * @return Sorted array of files that are older versions of the given JAR
+   * @param versionedJarFileName: the file names that exists on the server, could be file with
```
dschneider-pivotal: the ":" in this @param is causing the LGTM warning

geode-core/src/main/java/org/apache/geode/internal/JarDeployer.java
```diff
@@ -158,61 +156,55 @@ protected File getNextVersionedJarFile(String unversionedJarName) {
...
+        .findFirst().orElse(null);
   }
 
-  protected String removeJarExtension(String jarName) {
-    if (jarName != null && jarName.endsWith(".jar")) {
-      return jarName.replaceAll("\\.jar$", "");
+  /**
+   * get the artifact id from the files deployed by the user
+   *
+   * @param deployedJarFileName: the filename that's deployed by the user. could be in the form of
```
dschneider-pivotal: the ":" on this @param is causing a LGTM warning

PR #4348, opened 2019-11-19
GEODE-6616: fix testClientDynamicallyDropsStoppedLocator

mhansonp:
fixed an issue where the test was not waiting for getOnlineLocators' value
to become 1. It can take a moment to update.

### DIFF
--- a/geode-core/src/distributedTest/java/org/apache/geode/cache/client/internal/AutoConnectionSourceDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/cache/client/internal/AutoConnectionSourceDUnitTest.java
@@ -22,7 +22,6 @@ import static org.assertj.core.api.Assertions.assertThatCode;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.fail;
 
-import java.io.IOException;
 import java.io.Serializable;
 import java.net.InetSocketAddress;
 import java.util.Arrays;
@@ -217,15 +216,8 @@ public class AutoConnectionSourceDUnitTest extends LocatorTestBase {
       await().until(() -> pool.getOnlineLocators().size() == 1);
     });
 
-
-
-    int serverPort = serverVM.invoke("Start BridgeServer", () -> {
-      try {
-        return startBridgeServer(null, getLocatorString(hostName, locator1Port));
-      } catch (IOException e) {
-        return -1;
-      }
-    });
+    int serverPort = serverVM.invoke("Start BridgeServer",
+        () -> startBridgeServer(null, getLocatorString(hostName, locator1Port)));
     assertThat(serverPort).isGreaterThan(0);
 
     putAndWaitForSuccess(clientVM, REGION_NAME, KEY, VALUE);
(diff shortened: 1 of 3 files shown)


geode-core/src/distributedTest/java/org/apache/geode/cache/client/internal/AutoConnectionSourceDUnitTest.java
```diff
@@ -199,19 +212,31 @@ public void testClientDynamicallyDropsStoppedLocator() throws Exception {
 
     // stop one of the locators and ensure that the client can find and use a server
     locator0VM.invoke("Stop Locator", this::stopLocator);
+    clientVM.invoke(() -> {
```
mhansonp: This is the fix.

geode-core/src/distributedTest/java/org/apache/geode/cache/client/internal/AutoConnectionSourceDUnitTest.java
```diff
@@ -199,19 +212,31 @@ public void testClientDynamicallyDropsStoppedLocator() throws Exception {
...
+    });
 
-    serverVM.invoke("Start BridgeServer",
-        () -> startBridgeServer(null, getLocatorString(hostName, locator1Port)));
 
-    putAndWaitForSuccess(clientVM, REGION_NAME, "key", "value");
-    Assert.assertEquals("value", getInVM(serverVM, "key"));
+
+    int serverPort = serverVM.invoke("Start BridgeServer", () -> {
+      try {
```
kirklund: Recommend deleting the try-catch block entirely. Add `throws IOException` to the test method signature, and then you won't need the assertion on line 229 either.

geode-core/src/distributedTest/java/org/apache/geode/cache/client/internal/AutoConnectionSourceDUnitTest.java
```diff
@@ -455,71 +478,78 @@ protected void putInVM(VM vm, final String regionName, final Serializable key,
...
+        for (ServerLocation sl : endpoints) {
+          actualEndpointPorts.add(sl.getPort());
         }
         assertEquals(expectedEndpointPorts, actualEndpointPorts);
       });
     });
   }
 
-  protected void checkLocators(VM vm, final InetSocketAddress[] expectedInitial,
+  private Boolean verifyLocatorsMatched(InetSocketAddress[] expected,
```
kirklund: It's better to just use primitives instead of wrappers:
```
private boolean verifyLocatorsMatched(...
```

geode-core/src/main/java/org/apache/geode/cache/client/internal/AutoConnectionSourceImpl.java
```diff
@@ -249,8 +249,8 @@ protected void updateLocatorInLocatorList(HostAddress locator) {
         }
       }
 
-      logger.info("updateLocatorInLocatorList locator list from: {} to {}",
-          locatorList.getLocators(), newLocatorsList);
+      // logger.debug("updateLocatorInLocatorList locator list from: {} to {}",
```
kirklund: Delete the commented out lines.

mhansonp: I actually meant to put that line back...

PR #4353, opened 2019-11-21
GEODE-7486: Fix IndexManager

jujoramos:
- Added unit and integration tests for the IndexManager class.
- Fixed IndexManager to mark indexes as invalid whenever there's an
  exception while adding or removing a mapping from the index.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/cache/query/internal/index/IndexManager.java
+++ b/geode-core/src/main/java/org/apache/geode/cache/query/internal/index/IndexManager.java
@@ -1155,8 +1155,9 @@ public class IndexManager {
     } catch (Exception exception) {
       index.markValid(false);
       setPRIndexAsInvalid((AbstractIndex) index);
-      logger.warn("Error adding an index mapping, index might be corrupt: "
-          + ((AbstractIndex) index).indexName + ".", exception);
+      logger.warn(String.format(
+          "Updating the Index %s failed. The index is corrupted and marked as invalid.",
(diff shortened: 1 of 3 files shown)


geode-core/src/main/java/org/apache/geode/cache/query/internal/index/IndexManager.java
```diff
@@ -1149,14 +1149,25 @@ private void processAction(RegionEntry entry, int action, int opCode) throws Que
...
-  private void addIndexMapping(RegionEntry entry, IndexProtocol index) throws IMQException {
+  void addIndexMapping(RegionEntry entry, IndexProtocol index) {
     try {
       index.addIndexMapping(entry);
     } catch (Exception exception) {
       index.markValid(false);
       setPRIndexAsInvalid((AbstractIndex) index);
-      logger.warn("Put operation for the entry corrupted the index : "
-          + ((AbstractIndex) index).indexName + " with the exception : \n " + exception);
+      logger.warn("Error adding an index mapping, index might be corrupt: "
```
agingade: If it needs to logged as error; how about using logger.error()?
Also changing it to: 
"Updating the Index (name:) " + indexName + " on region " + regionName + " failed. The index is corrupted and marked as invalid."

Reason being, customer may not know about "index mapping".

jhuynh1: I think warning would be fine (opinion) but I like the message Anil suggested.  Same comment for the remove

PR #4364, opened 2019-11-22
GEODE-7344: DSFID implements BasicSerializable

Bill:
[GEODE-7344](https://issues.apache.org/jira/browse/GEODE-7344)

As part of eliminating `TcpServer` dependency on `geode-core`, this ticket removes `TcpServer` dependencies on `DataSerializable`.

The way we did that was to change the 6 message types used to talk to `TcpServer` so instead of deriving from `DataSerializable`—they now derive from a new `BasicSerializable` interface.

That interface is made a base for `DataSerializableFixedID`. Then we modified `geode-serialization` so it handles any `BasicSerializable` i.e. both the base type and the derived (`DataSerializableFixedID`) type. This entailed a few changes up in `geode-core`: primarily in `InternalDataSerializer`.

The six message types that are now `BasicSerializable` are:

* `VersionRequest`/`Response`
* `InfoRequest`/`Response`
* `ShutdownRequest`/`Response`

With these changes, the on-the-wire format of the new `BasicSerializable` is the same as the old `DataSerializable`. This lets a new Geode and an old Geode communicate.

With this PR, the `TcpServer` package no longer depends on anything in `geode-core` and `TcpServer` and friends (in `org.apache.geode.distributed.internal.tcpserver`)  are ready to move to their very own new module (in a future PR).

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/InternalDataSerializer.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/InternalDataSerializer.java
@@ -78,7 +78,6 @@ import org.apache.geode.GemFireConfigException;
 import org.apache.geode.GemFireIOException;
 import org.apache.geode.GemFireRethrowable;
 import org.apache.geode.Instantiator;
-import org.apache.geode.InternalGemFireError;
 import org.apache.geode.SerializationException;
 import org.apache.geode.SystemFailure;
 import org.apache.geode.ToDataException;
@@ -107,6 +106,7 @@ import org.apache.geode.internal.cache.tier.sockets.OldClientSupportService;
 import org.apache.geode.internal.cache.tier.sockets.Part;
 import org.apache.geode.internal.lang.ClassUtils;
 import org.apache.geode.internal.logging.log4j.LogMarker;
+import org.apache.geode.internal.serialization.BasicSerializable;
 import org.apache.geode.internal.serialization.DSCODE;
 import org.apache.geode.internal.serialization.DSFIDSerializer;
 import org.apache.geode.internal.serialization.DSFIDSerializerFactory;
@@ -254,10 +254,6 @@ public abstract class InternalDataSerializer extends DataSerializer {
       LOAD_CLASS_EACH_TIME ? null : new CopyOnWriteHashMap<>();
   private static final Object cacheAccessLock = new Object();
 
-  private static final String PRE_GEODE_100_TCPSERVER_PACKAGE =
-      "com.gemstone.org.jgroups.stack.tcpserver";
-  private static final String POST_GEODE_100_TCPSERVER_PACKAGE =
-      "org.apache.geode.distributed.internal.tcpserver";
   private static final String POST_GEODE_190_SERVER_CQIMPL =
       "org.apache.geode.cache.query.cq.internal.ServerCQImpl";
   private static final String PRE_GEODE_190_SERVER_CQIMPL =
@@ -321,15 +317,14 @@ public abstract class InternalDataSerializer extends DataSerializer {
    * For backward compatibility we must swizzle the package of some classes that had to be moved
    * when GemFire was open- sourced. This preserves backward-compatibility.
    *
-   * @param name the fully qualified class name
+   * @param nameArg the fully qualified class name
    * @return the name of the class in this implementation
    */
-  public static String processIncomingClassName(String name) {
-    // TCPServer classes are used before a cache exists and support for old clients has been
-    // initialized
-    if (name.startsWith(PRE_GEODE_100_TCPSERVER_PACKAGE)) {
-      return POST_GEODE_100_TCPSERVER_PACKAGE
-          + name.substring(PRE_GEODE_100_TCPSERVER_PACKAGE.length());
+  public static String processIncomingClassName(String nameArg) {
+    final String name = StaticSerialization.processIncomingClassName(nameArg);
+    // using identity comparison on purpose because we are on the hot path
+    if (name != nameArg) {
+      return name;
     }
     if (name.equals(PRE_GEODE_190_SERVER_CQIMPL)) {
       return POST_GEODE_190_SERVER_CQIMPL;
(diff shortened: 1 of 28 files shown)


geode-core/src/main/java/org/apache/geode/internal/InternalDataSerializer.java
```diff
@@ -2094,8 +2030,8 @@ public static void basicWriteObject(Object o, DataOutput out, boolean ensurePdxC
         out.writeByte(DSCODE.DATA_SERIALIZABLE.toByte());
         DataSerializer.writeClass(c, out);
       }
-      DataSerializable ds = (DataSerializable) o;
-      invokeToData(ds, out);
+      final DataSerializable bs = (DataSerializable) o;
```
dschneider-pivotal: why rename this to "bs" since it is still a DataSerializable?

geode-serialization/src/main/java/org/apache/geode/internal/serialization/BasicSerializable.java
```diff
@@ -0,0 +1,42 @@
...
+ * the License.
+ */
+package org.apache.geode.internal.serialization;
+
+import java.io.DataInput;
+import java.io.DataOutput;
+import java.io.IOException;
+
+/**
+ * This interface supports toData/fromData serialization. DataSerializableFixedID is almost always
```
dschneider-pivotal: Should you explain when this interface should be used?

geode-serialization/src/main/java/org/apache/geode/internal/serialization/BasicSerializable.java
```diff
@@ -0,0 +1,42 @@
...
+ * preferable to this interface for new work.
+ */
+public interface BasicSerializable {
+
+  /**
+   * Writes the state of this object as primitive data to the given <code>DataOutput</code>.
```
dschneider-pivotal: change DataOutput to context?

geode-serialization/src/main/java/org/apache/geode/internal/serialization/BasicSerializable.java
```diff
@@ -0,0 +1,42 @@
...
+   *
+   * @throws IOException A problem occurs while writing to <code>out</code>
+   */
+  default void toData(DataOutput out, SerializationContext context) throws IOException {};
+
+  /**
+   * Reads the state of this object as primitive data from the given <code>DataInput</code>.
```
dschneider-pivotal: change DataInput to context?

geode-serialization/src/main/java/org/apache/geode/internal/serialization/BasicSerializable.java
```diff
@@ -0,0 +1,42 @@
...
+ * preferable to this interface for new work.
+ */
+public interface BasicSerializable {
+
+  /**
+   * Writes the state of this object as primitive data to the given <code>DataOutput</code>.
+   *
+   * @throws IOException A problem occurs while writing to <code>out</code>
+   */
+  default void toData(DataOutput out, SerializationContext context) throws IOException {};
```
dschneider-pivotal: I don't like "default" on these methods. It seems better, in this case, to insist that implementors of this interface implement toData and fromData.

Bill:
Hydra tests passed:

* PDX tests: http://hydradb-sandbox.gemfire.pivotal.io:9090/hdb/run/13291 (1 auto-diagnosed failure)
* Cluster config tests: http://hydradb-sandbox.gemfire.pivotal.io:9090/hdb/run/13294 (all passed)

geode-serialization/src/main/java/org/apache/geode/internal/serialization/DSFIDSerializerImpl.java
```diff
@@ -111,19 +117,34 @@ public void writeDSFIDHeader(int dsfid, DataOutput out) throws IOException {
   }
 
   @Override
-  public void writeDSFID(DataSerializableFixedID o, DataOutput out) throws IOException {
-    if (o == null) {
-      out.writeByte(DSCODE.NULL.toByte());
+  public void write(final BasicSerializable bs, final DataOutput out) throws IOException {
+    if (writeMetaData(bs, out)) {
```
dschneider-pivotal: I think this code might be better if you changed writeMetaData to return void and just had it write the meta data.
Then after you call it you could conditionalize the invokeToData call with "if (bs != null)".

geode-core/src/main/java/org/apache/geode/internal/InternalDataSerializer.java
Bill: Actually we don't need that cast at all since `invokeToData()` takes an `Object`. Fixing…

geode-serialization/src/main/java/org/apache/geode/internal/serialization/BasicSerializable.java
Bill: agreed. Fixing…

geode-serialization/src/main/java/org/apache/geode/internal/serialization/BasicSerializable.java
Bill: I don't think so.

geode-serialization/src/main/java/org/apache/geode/internal/serialization/BasicSerializable.java
Bill: I don't think so.

geode-serialization/src/main/java/org/apache/geode/internal/serialization/BasicSerializable.java
Bill: why does that seem better @dschneider-pivotal ?

geode-serialization/src/main/java/org/apache/geode/internal/serialization/DSFIDSerializerImpl.java
Bill: yes! changing…

geode-serialization/src/main/java/org/apache/geode/internal/serialization/BasicSerializable.java
dschneider-pivotal: The defaults make it too easy to implement this interface and have it serialize nothing. By forcing the implementor to implement toData and fromData they get a chance to consider if their class has any data that needs to be serialized/deserialized. I don't think you have any backwards compatibility to be concerned about so you don't need to use default for that reason. And you don't actually have any interesting "default" implementation here (just empty methods).

geode-serialization/src/main/java/org/apache/geode/internal/serialization/BasicSerializable.java
dschneider-pivotal: You are right. I'm not sure why I thought it was the "context"

geode-serialization/src/main/java/org/apache/geode/internal/serialization/BasicSerializable.java
Bill: I understand your point. OTOH we (now) clearly document `BasicSerializable` as an interface that is solving a particular legacy problem and we clearly say not to use it for new work. Inasmuch as only 6 classes implement this interface and only 2/6 need to override the (no-op) behavior, I'm gonna leave the `default` impls in.

Bill:
`DistributedTestOpenJDK11` latest CI run failed but I believe it is due to a CI infrastructure problem. Previous CI run of that test was successful and I made no big changes after that. Merging.

PR #4366, opened 2019-11-23
GEODE-7478: Register interest in appropriate cases.

pivotal-eshu:
* register intesest only when local cache is enabled for Tomcat
 * add register interest in other app server when local cache is enabled.
 * add cache client test cases for tomcat cases.
 * modify test to allow gets mutliple times on containers when local cache is enabled
   on client cache to verify the fix.

### DIFF
--- a/extensions/geode-modules-session-internal/src/main/java/org/apache/geode/modules/session/internal/common/ClientServerSessionCache.java
+++ b/extensions/geode-modules-session-internal/src/main/java/org/apache/geode/modules/session/internal/common/ClientServerSessionCache.java
@@ -23,7 +23,9 @@ import javax.servlet.http.HttpSession;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import org.apache.geode.cache.DataPolicy;
 import org.apache.geode.cache.GemFireCache;
+import org.apache.geode.cache.InterestResultPolicy;
 import org.apache.geode.cache.Region;
 import org.apache.geode.cache.RegionShortcut;
 import org.apache.geode.cache.client.ClientCache;
@@ -124,6 +126,10 @@ public class ClientServerSessionCache extends AbstractSessionCache {
       LOG.debug("Created session region: " + this.sessionRegion);
     } else {
       LOG.debug("Retrieved session region: " + this.sessionRegion);
+    }
+
+    if (sessionRegion.getAttributes().getDataPolicy() != DataPolicy.EMPTY) {
+      sessionRegion.registerInterestForAllKeys(InterestResultPolicy.KEYS);
     }
   }
 
--- a/extensions/geode-modules-session-internal/src/main/java/org/apache/geode/modules/session/internal/common/ClientServerSessionCache.java
+++ b/extensions/geode-modules-session-internal/src/main/java/org/apache/geode/modules/session/internal/common/ClientServerSessionCache.java
@@ -23,7 +23,6 @@ import javax.servlet.http.HttpSession;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import org.apache.geode.cache.DataPolicy;
 import org.apache.geode.cache.GemFireCache;
 import org.apache.geode.cache.InterestResultPolicy;
 import org.apache.geode.cache.Region;
@@ -126,9 +125,8 @@ public class ClientServerSessionCache extends AbstractSessionCache {
       LOG.debug("Created session region: " + this.sessionRegion);
     } else {
       LOG.debug("Retrieved session region: " + this.sessionRegion);
-    }
 
-    if (sessionRegion.getAttributes().getDataPolicy() != DataPolicy.EMPTY) {
+      // Register interest in case users provide their own client cache region
       sessionRegion.registerInterestForAllKeys(InterestResultPolicy.KEYS);
     }
   }
@@ -175,6 +173,9 @@ public class ClientServerSessionCache extends AbstractSessionCache {
     }
 
     // Create the region
-    return factory.create(regionName);
+    Region region = factory.create(regionName);
+
+    region.registerInterestForAllKeys(InterestResultPolicy.KEYS);
+    return region;
   }
 }
--- a/extensions/geode-modules/src/main/java/org/apache/geode/modules/session/catalina/ClientServerSessionCache.java
+++ b/extensions/geode-modules/src/main/java/org/apache/geode/modules/session/catalina/ClientServerSessionCache.java
@@ -190,9 +190,8 @@ public class ClientServerSessionCache extends AbstractSessionCache {
         sessionRegion.getAttributesMutator().addCacheListener(new SessionExpirationCacheListener());
       }
 
-      // This is true for PROXY regions
-      if (sessionRegion.getAttributes().getDataPolicy() == DataPolicy.EMPTY) {
-        sessionRegion.registerInterest("ALL_KEYS", InterestResultPolicy.KEYS);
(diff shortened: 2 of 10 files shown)


extensions/geode-modules-session-internal/src/main/java/org/apache/geode/modules/session/internal/common/ClientServerSessionCache.java
```diff
@@ -125,6 +127,10 @@ private void createOrRetrieveRegion() {
     } else {
       LOG.debug("Retrieved session region: " + this.sessionRegion);
     }
+
+    if (sessionRegion.getAttributes().getDataPolicy() != DataPolicy.EMPTY) {
```
jhuynh1: This method may call createLocalSessionRegion().  Which it looks like this same interest registration code is duplicated into that method?

extensions/geode-modules/src/main/java/org/apache/geode/modules/session/catalina/ClientServerSessionCache.java
```diff
@@ -190,9 +190,8 @@ protected void createOrRetrieveRegion() {
         sessionRegion.getAttributesMutator().addCacheListener(new SessionExpirationCacheListener());
       }
 
-      // This is true for PROXY regions
-      if (sessionRegion.getAttributes().getDataPolicy() == DataPolicy.EMPTY) {
-        sessionRegion.registerInterest("ALL_KEYS", InterestResultPolicy.KEYS);
+      if (sessionRegion.getAttributes().getDataPolicy() != DataPolicy.EMPTY) {
```
jhuynh1: We've flipped the comparison here... is this intentional?  What happens in the EMPTY case?

extensions/geode-modules/src/main/java/org/apache/geode/modules/session/catalina/ClientServerSessionCache.java
```diff
@@ -247,12 +246,9 @@ void createSessionRegionOnServers() {
...
     Region region = factory.create(getSessionManager().getRegionName());
 
-    /*
-     * If we're using an empty client region, we register interest so that expired sessions are
-     * destroyed correctly.
-     */
-    if (!getSessionManager().getEnableLocalCache()) {
-      region.registerInterest("ALL_KEYS", InterestResultPolicy.KEYS);
+    // register interest to get updates from server if local cache is enabled
+    if (getSessionManager().getEnableLocalCache()) {
```
jhuynh1: Again, we flipped the logic.  Will this cause unexpected behavior changes?

extensions/geode-modules/src/main/java/org/apache/geode/modules/session/catalina/ClientServerSessionCache.java
agingade: It looks like the registration is required for:
1. To handle destroyed session on client (For empty regions) and 
2. Replication (for caching-proxy region) when the application fails-over between the two gemfire clients managing the session. Is this supported/requirement?

extensions/geode-modules-session-internal/src/main/java/org/apache/geode/modules/session/internal/common/ClientServerSessionCache.java
jhuynh1: Sure but a few lines up we call createLocalSessionRegion().  If that gets called, we end up registering interest in that method also.  Just wanted to know why we do that or if that was intended...

extensions/geode-modules/src/main/java/org/apache/geode/modules/session/catalina/ClientServerSessionCache.java
jhuynh1: In either case, why did we do == EMPTY before and != EMPTY now?  Why not just do it for all cases?  Since we flipped the logic, was it completely backwards/wrong before?

extensions/geode-modules-session-internal/src/main/java/org/apache/geode/modules/session/internal/common/ClientServerSessionCache.java
pivotal-eshu: createLocalSessionRegion() in common/ClientServerSessionCache does not register interest. 

However, I will add register interest code in createLocalSessionRegion() and only register interest if is user defined the cache region.

extensions/geode-modules/src/main/java/org/apache/geode/modules/session/catalina/ClientServerSessionCache.java
pivotal-eshu: Yes, needed to register interest for both cases - still needs the callback events for empty proxy region as well.

extensions/geode-modules/src/main/java/org/apache/geode/modules/session/catalina/ClientServerSessionCache.java
pivotal-eshu: Yes, we need both.

extensions/geode-modules-session-internal/src/main/java/org/apache/geode/modules/session/internal/common/ClientServerSessionCache.java
```diff
@@ -124,6 +125,9 @@ private void createOrRetrieveRegion() {
       LOG.debug("Created session region: " + this.sessionRegion);
     } else {
       LOG.debug("Retrieved session region: " + this.sessionRegion);
+
+      // Register interest in case users provide their own client cache region
+      sessionRegion.registerInterestForAllKeys(InterestResultPolicy.KEYS);
```
agingade: We could registerInterest in a common place for both user and internally created session regions.

extensions/geode-modules/src/main/java/org/apache/geode/modules/session/catalina/ClientServerSessionCache.java
```diff
@@ -225,6 +221,17 @@ void createSessionRegionOnServers() {
     }
   }
 
+  Region<String, HttpSession> createLocalSessionRegionWithRegisterInterest() {
+    Region<String, HttpSession> region = createLocalSessionRegion();
+
+    // register interest are needed for proxy or caching-proxy client:
+    // to get updates from server if local cache is enabled;
+    // to get callbacks for listener invocation for proxy client.
+    region.registerInterestForAllKeys(InterestResultPolicy.KEYS);
```
agingade: We could registerInterest in a common place for both user and internally created session regions.

extensions/geode-modules/src/main/java/org/apache/geode/modules/session/catalina/ClientServerSessionCache.java
```diff
@@ -225,6 +221,17 @@ void createSessionRegionOnServers() {
     }
   }
 
+  Region<String, HttpSession> createLocalSessionRegionWithRegisterInterest() {
+    Region<String, HttpSession> region = createLocalSessionRegion();
+
+    // register interest are needed for proxy or caching-proxy client:
```
agingade: It will be nice to call out the requirements separately.
In case of proxy session region Interest registration is done to get the destroy callbacks and for caching-proxy to replicate the data across clients.

extensions/geode-modules-session-internal/src/main/java/org/apache/geode/modules/session/internal/common/ClientServerSessionCache.java
pivotal-eshu: I think you are right - but this follows the pattern in catlina/ClientSessionCache. 
Could do refactoring at another time.

PR #4369, opened 2019-11-25
GEODE-7487: Update Running CQ Context

jujoramos:
- Added unit and integration tests.
- Implemented method to invalidate the cache used by CQs.
- Updated the context implementation to change the internal
  MethodInvocationAuthorizer used whenever the CQ resets the
  ExecutionContext.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/cache/query/internal/ExecutionContext.java
+++ b/geode-core/src/main/java/org/apache/geode/cache/query/internal/ExecutionContext.java
@@ -102,13 +102,19 @@ public class ExecutionContext {
   static final ThreadLocal<AtomicBoolean> isCanceled =
       ThreadLocal.withInitial(AtomicBoolean::new);
 
-  // Authorizer to use during the query execution.
-  private final MethodInvocationAuthorizer methodInvocationAuthorizer;
+  /**
+   * Authorizer to use during query execution.
+   * It can not be changed for queries in flight, but it should be modifiable for running CQs,
+   * that's the only reason why this field is not marked as {@code final}.
+   *
+   * @see ExecutionContext#reset()
+   */
+  private MethodInvocationAuthorizer methodInvocationAuthorizer;
+  private final QueryConfigurationService queryConfigurationService;
 
   /**
    * Returns the {@link MethodInvocationAuthorizer} that will be used, if needed, during the
-   * execution
-   * of the query associated with this context.
+   * execution of the query associated with this context.
    *
    * @return the {@link MethodInvocationAuthorizer} that will be used, if needed, during the
    *         execution of the query associated with this context.
@@ -127,13 +133,8 @@ public class ExecutionContext {
     this.bindArguments = bindArguments;
     this.cache = cache;
     this.cancelationTask = Optional.empty();
-
-    // Authorization & authentication logic happens on server side only.
-    if (cache.isClient()) {
-      this.methodInvocationAuthorizer = QueryConfigurationServiceImpl.getNoOpAuthorizer();
-    } else {
-      this.methodInvocationAuthorizer = cache.getQueryService().getMethodInvocationAuthorizer();
-    }
+    this.queryConfigurationService = cache.getService(QueryConfigurationService.class);
+    this.methodInvocationAuthorizer = queryConfigurationService.getMethodAuthorizer();
   }
 
   Optional<ScheduledFuture> getCancelationTask() {
@@ -585,6 +586,8 @@ public class ExecutionContext {
    */
   public void reset() {
     scopes.clear();
+    // Make sure we use the most up to date authorizer in CQs.
+    methodInvocationAuthorizer = queryConfigurationService.getMethodAuthorizer();
   }
 
   public BucketRegion getBucketRegion() {
--- a/geode-core/src/main/java/org/apache/geode/cache/query/internal/ExecutionContext.java
+++ b/geode-core/src/main/java/org/apache/geode/cache/query/internal/ExecutionContext.java
@@ -130,8 +130,8 @@ public class ExecutionContext {
    * @see org.apache.geode.cache.Region#query
    */
   public ExecutionContext(Object[] bindArguments, InternalCache cache) {
-    this.bindArguments = bindArguments;
     this.cache = cache;
+    this.bindArguments = bindArguments;
     this.cancelationTask = Optional.empty();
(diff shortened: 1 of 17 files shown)


geode-core/src/main/java/org/apache/geode/cache/query/internal/cq/ServerCQ.java
```diff
@@ -53,6 +53,13 @@ void registerCq(ClientProxyMembershipID p_clientProxyId, CacheClientNotifier p_c
    */
   void removeFromCqResultKeys(Object key, boolean isTokenMode);
 
+  /**
+   * Invalidates the internal cache containing the keys that are part of the CQ query results.
+   * Once this method finishes, the CQ engine will not apply the internal optimization for already
+   * seen keys anymore, not until the cache is manually rebuilt.
+   */
+  void invalidateCqResultKeys();
```
agingade: I am not seeing this method usage in product source code; I see it in the test code...Hope I am not missing anything.

geode-core/src/main/java/org/apache/geode/cache/query/internal/ExecutionContext.java
```diff
@@ -127,13 +133,8 @@ public ExecutionContext(Object[] bindArguments, InternalCache cache) {
...
     this.cache = cache;
     this.cancelationTask = Optional.empty();
-
-    // Authorization & authentication logic happens on server side only.
-    if (cache.isClient()) {
-      this.methodInvocationAuthorizer = QueryConfigurationServiceImpl.getNoOpAuthorizer();
-    } else {
-      this.methodInvocationAuthorizer = cache.getQueryService().getMethodInvocationAuthorizer();
-    }
+    this.queryConfigurationService = cache.getService(QueryConfigurationService.class);
```
agingade: Can this be null. What if user overrides the QueryConfigurationService provided with product.

geode-core/src/main/java/org/apache/geode/cache/query/internal/cq/ServerCQ.java
jujoramos: You haven't missed anything 👍, right now the method is only used by tests.
It will be officially used within the product code as part of [GEODE-7497](https://issues.apache.org/jira/browse/GEODE-7497), that is, invalidate the CQs cache whenever we update the `MethodInvocationAuthorizer`.

geode-core/src/main/java/org/apache/geode/cache/query/internal/ExecutionContext.java
jujoramos: I don't think this service should be configurable/exchangeable by the user, meaning that "it's okay to fail" if the user mess around with it. I'm not completely sure, though, do you any extra thoughts about it?.

jujoramos: We're not checking for the nullity of the `QueryConfigurationService` anywhere in the code, BTW, and maybe we should invest some time on investigating it... but I think that should belong to another PR and/or ticket, anyway, just my two cents 👍

jhuynh1: I think Anil has a good point, is there any chance the QueryConfigurationService is null (not set due to not being used?)  If no, then I think the null checks shouldn't be added.  If there is a chance for it to be null, maybe we should install a no-op service instead, so we can avoid null checks?

jujoramos: @agingade @jhuynh1: done!

agingade: The "MethodInvocationAuthorizer" returned by the "QueryConfigurationService" could be null as its implemented by users. This is may never happen; but we currently don't check for it.

jujoramos:
@agingade 
I don't think we should go that far with the null checks. In fact, I don't even think we should check for the nullity of the `QueryConfigurationService` in the first place as it should always be created by default [here](https://github.com/apache/geode/blob/develop/geode-core/src/main/java/org/apache/geode/internal/cache/GemFireCacheImpl.java#L1264-L1276).
In order to have more control about which methods should be allowed during an OQL execution, users must implement the `MethodInvocationAuthorizer` interface and use the configuration options we provide; they should never implement the `QueryConfigurationService` directly as it has nothing to do with the actual authorization, its purpose is to control how the query engine is configured. The `QueryConfigurationService` interface is internal and was never meant to be implemented by users at all. They can dig in the code, find the interface, implement it, and manage themselves to override the default service implementation, sure, but that would require some work and access to the actual file system on which Geode is running to "hack" the default service declaration within the `META-INF/services` directory, in which case we can't do nothing about it either.
I'm happy to discuss more about this, I see the following options moving forward:

- Leave things as they are, checking only the nullity of the `QueryConfigurationService` itself to make sure users don't put themselves into a bad situation by manually messing around with the classpath.
- Revert my [last commit on this PR](https://github.com/apache/geode/pull/4369/commits/def003221edaefa1faf606bafe70c6bc37c687cc) and leave things as they originally were, that is, don't check for the nullity of `QueryConfigurationService` nor for `QueryConfigurationService.getMethodInvocationAuthorizer()`. This assumes that our users will never implement `QueryConfigurationService`, which is expected as this interface is internal.
- Make `QueryConfigurationService` an internal service, not externally configurable through the `ServiceLoader` mechanism. This would guarantee that only our implementation is used, and avoid returning `null`.
- A mix of the above 3 options: leave the `QueryConfigurationService` externally configurable through the `ServiceLoader` mechanism but also make it a "special service" in the sense that, [during cache initialization](https://github.com/apache/geode/blob/develop/geode-core/src/main/java/org/apache/geode/internal/cache/GemFireCacheImpl.java#L1264-L1276), we throw an exception whenever either the service instance or the returned `MethodInvocationAuthorizer` by the instance are `null`.

@DonalEvans @BenjaminPerryRoss : I might be missing something here, thoughts?.

onichols-pivotal:
As a courtesy, please avoid force-push after a PR has begun receiving comments and reviews.  For example, it would be nice to see the delta that was pushed after this received an approving review.

jujoramos:
@onichols-pivotal 
I always try to avoid force pushing changes to a `PR`, this time I had no other option as I had to literally "revert" my last two commits within this `PR` prior to the pushing the new changes. Thanks for the heads up, though.

PR #4372, opened 2019-11-25
GEODE-4187: Fix PersistentColocatedPartitionedRegionDistributedTest

kirklund:
Flakiness was fixed by:
* Clean up messy uses of AsyncInvocation
* Replace uses of Thread.sleep with Mockito.verify using timeout()
* Use GeodeAwaitility timeout for all timeouts

High level changes to tests and product code:
* Rename PersistentColocatedPartitionedRegionDUnitTest to 
  PersistentColocatedPartitionedRegionDistributedTest
* Extract single-JVM tests to colocation integration tests
  (ModifyColocationIntegrationTest, PersistentColocationIntegrationTest)
* Move ColocationLogger to internal.cache.partitioned.colocation package
* Extract factory and interfaces from ColocationLogger
* Create unit tests for ColocationLogger and factory
* Pass ColocationLoggerFactory into PartitionedRegion constructor
* Use ColocationLogger via interfaces instead of implementations
* Clean up PartitionedRegionTest (unit test)
* Clean up colocation IllegalStateException message
* Add getLogger(Class) to LogService

Update PersistentColocatedPartitionedRegionDistributedTest:
* Make the code readable and understandable
* Update to non-deprecated APIs
* Inline super-class and lots of methods
* Use TemporaryFolder Rule for all files
* Use RestoreSystemProperties Rule instead of finally blocks
* Replace MockAppender with cleaner inlined use of Mockito
* Clean up usages of AsyncInvocation
* Replace messy try-catch blocks with AssertJ catchThrowable
* Throw failures and unexpected exceptions instead of catching them to invoke fail
* Use descriptive variable and method names
* Replace use of Admin API for shutdown with Cache.close()
* Use lambdas instead of SerializableRunnable variables
* Use lambdas instead of lots of tiny inner-classes
* Use tearDown instead of try-finally blocks
* Reimplement broken waits with Awaitility


geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
```diff
@@ -450,6 +452,8 @@ public void setShutDownAllStatus(int newStatus) {
 
   private final PartitionedRegion colocatedWithRegion;
 
+  private final ColocationLoggerFactory colocationLoggerFactory;
+
   private ColocationLogger missingColocatedRegionLogger;
```
kirklund: It would be more correct if this field were changed to `volatile`

kirklund:
ModifyColocationIntegrationTest is flaky because of the hardcoded 2-second sleeps. These need to be replaced by some more deterministic approach to detecting when the Cache and System are completely closed.

kirklund:
ModifyColocationIntegrationTest is fixed.

kirklund:
DistributedTestOpenJDK11 PASSED but seems to have hung during archival.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
igodwin: I agree.

kirklund:
DistributedTestOpenJDK11 had one failure in ClientDataAuthorizationUsingLegacySecurityWithFailoverDUnitTest involving client-server protocol -- unrelated to this PR.

PR #4375, opened 2019-11-25
Do Not Review: GEODE-6374 Changes to wait and single sourcing code.

mhansonp:
- Fixed some waits that were not as they needed to be for deterministic behavior
- Cleaning out AttributeFactory usages
  To do this, refactoring some helper methods was required
- Cleanup of RegionFactory usages
  There were some usages of new RegionFactory<>() which is deprecated

### DIFF
--- a/geode-core/src/distributedTest/java/org/apache/geode/cache30/DistributedNoAckRegionDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/cache30/DistributedNoAckRegionDUnitTest.java
@@ -30,6 +30,7 @@ import org.apache.geode.cache.CacheException;
 import org.apache.geode.cache.DataPolicy;
 import org.apache.geode.cache.Region;
 import org.apache.geode.cache.RegionAttributes;
+import org.apache.geode.cache.RegionFactory;
 import org.apache.geode.cache.Scope;
 import org.apache.geode.distributed.internal.membership.InternalDistributedMember;
 import org.apache.geode.internal.cache.DistributedRegion;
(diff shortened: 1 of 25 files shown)


geode-dunit/src/main/java/org/apache/geode/cache30/RegionTestCase.java
```diff
@@ -2095,33 +2026,31 @@ public void testCustomEntryTtl1() {
...
+      region = createRegion(name, regionFactory);
     } finally {
       System.getProperties().remove(LocalRegion.EXPIRY_MS_PROPERTY);
     }
 
     // Random values should not expire
     region.put(key1, value);
-    Wait.pause(timeout * 2);
-    assert (region.get(key1).equals(value));
+    await().atLeast(timeout * 2, MILLISECONDS).until(() -> region.get(key1).equals(value));
```
mhansonp: Important change

geode-core/src/distributedTest/java/org/apache/geode/cache30/DistributedNoAckRegionDUnitTest.java
```diff
@@ -220,17 +209,14 @@ public void run2() throws CacheException {
...
-        final byte[] value = new byte[valueSize];
+        Region<Object, Object> region = getRootRegion().getSubregion(name);
+        final byte[] value = new byte[VALUE_SIZE_VM];
         Arrays.fill(value, (byte) 0x42);
-        for (int i = 0; i < numEntries; i++) {
-          if ((i % 1000) == 0) {
-            LogWriterUtils.getLogWriter().info("testBug30705: initial put #" + i);
+        for (int i = 0; i < NUM_ENTRIES_VM; i++) {
+          if ((i % 1000) != 0) {
+            region.put(i, value);
```
jhuynh1: I don't think this is the same as before.  Before it would still do the put even if the mod is 0.  It would just print out the log line if it happened to be the first.  I'm surprised if we don't get any test failures from this change.

mhansonp: So you were right. I swapped the logic incorrectly.  The reason the test didn't fail is that it is ignored. I have swapped the logic back.

mhansonp: The test was also no longer needed, so I deleted it.

mhansonp:
Closing while I work on issues found during PR tests.

mhansonp:
Don't want PR builds for the moment.

PR #4379, opened 2019-11-26
GEODE-7485: REST URLs accept dots in trailing parameters

demery-pivotal:
By default, when Spring parses a request mapping parameter at the end of a URL, it treats the last dot in the paramater value to be the start of a file extension, and discards the 'extension" before passing the parameter to the handler method.

Several of our configurable items allow dots in their IDs. And several of our REST request mappings place those IDs at the end of the URL. In those cases, Spring's default behavior truncates the ID at the last dot, and passes the truncated ID to the request handler. As a result, the request fails.                                                

This commit configure all current management REST API request mappings to allow dots in their final parameters.

Co-authored-by: Jinmei Liao <jiliao@pivotal.io>

### For all changes:

### DIFF
--- a/geode-web-management/src/integrationTest/java/org/apache/geode/management/internal/rest/GatewayManagementControllerSpringTest.java
+++ b/geode-web-management/src/integrationTest/java/org/apache/geode/management/internal/rest/GatewayManagementControllerSpringTest.java
@@ -1,0 +1,86 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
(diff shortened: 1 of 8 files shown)


geode-assembly/src/distributedTest/java/org/apache/geode/management/internal/rest/ListIndexManagementDUnitTest.java
```diff
@@ -70,7 +76,8 @@ public static void beforeclass() throws Exception {
         "create index --name=index1 --type=key --expression=id --region=/region1")
         .statusIsSuccess();
     gfsh.executeAndAssertThat(
-        "create index --name=index2 --type=key --expression=key --region=/region1")
+        "create index --name=" + INDEX_WITH_DOT_IN_NAME
```
jinmeiliao: I found an easier way to test this, basically we only need to intercept what's been received by the controller, we don't really need to start a real locator/server for that.

We can piggybag on `JsonSerializationTest`, add another test method there:


@Test
  public void nameWithDot() throws Exception {
    String url = "/v1/regions/region.name/indexes/index.name";
    context.perform(get(url)).andExpect(status().isOk());
    ArgumentCaptor<Index> argumentCaptor = ArgumentCaptor.forClass(Index.class);
    verify(cms).get(argumentCaptor.capture());
    Index index = argumentCaptor.getValue();
    assertThat(index.getRegionName()).isEqualTo("region.name");
    assertThat(index.getId()).isEqualTo("index.name");
  }

demery-pivotal: Yes, I would much prefer to test this closer to the controller!

PR #4382, opened 2019-11-27
GEODE-7436: Deploy jar using semantic versioning scheme 

jinmeiliao:

### For all changes:

### Note:


jinmeiliao:
@jmelchio

geode-core/src/integrationTest/java/org/apache/geode/internal/JarDeployerIntegrationTest.java
```diff
@@ -168,7 +169,12 @@ public void undeploy() throws Exception {
         .isInstanceOf(IllegalArgumentException.class);
```
demery-pivotal: It's not clear in the test what makes this argument illegal. I would prefer to put this in a separate test, and give the test a name that describes why the argument is illegal.

geode-core/src/integrationTest/java/org/apache/geode/internal/ClassPathLoaderIntegrationTest.java
```diff
@@ -554,22 +554,11 @@ protected File getTempFile() {
...
-      return tempFile2;
-    }
-  }
-
   private File createJarWithClass(String fileName, String className) throws IOException {
     String stringBuilder = "package integration.parent;" + "public class " + className + " {}";
     File jarFile = new File(temporaryFolder.getRoot(), fileName);
     if (jarFile.exists()) {
-      jarFile.delete();
+      jarFile = new File(temporaryFolder.newFolder(), fileName);
```
demery-pivotal: I think the root of the problem is that the tests use the same dir for staging as for deploying. I would prefer that the tests _always_ use separate dirs for staging and deploying.

PR #4385, opened 2019-11-28
GEODE-7497: Check CQs prior to change authorizer

jujoramos:
The QueryConfigurationService now requires an extra flag to determine,
in the presence of continuous queries, whether to update the configured
MethodInvocationAuthorizer and invalidate the CQ's internal cache or
throw an exception and abort the update.

- Fixed minor warnings.
- Added unit and distributed tests.
- Updated docs for alter query-service command.
- Added 'forceUpdate' flag to 'AlterQueryServiceFunction' and
  'AlterQueryServiceCommand'.
- Fixed 'AlterQueryServiceCommand' to always require the
  'method-authorizer' parameter.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/AlterQueryServiceCommandWithSecurityDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/AlterQueryServiceCommandWithSecurityDUnitTest.java
@@ -100,7 +100,8 @@ public class AlterQueryServiceCommandWithSecurityDUnitTest {
     return commandBuilder.toString();
   }
 
-  private void executeAndAssertGfshCommandIsSuccessful(Class authorizerClass,
+  private void executeAndAssertGfshCommandIsSuccessful(
+      Class<? extends MethodInvocationAuthorizer> authorizerClass,
       String authorizerParameters) {
     gfsh.executeAndAssertThat(buildCommand(authorizerClass.getName(), authorizerParameters))
         .statusIsSuccess().hasTableSection().hasRowSize(servers.size());
@@ -123,7 +124,8 @@ public class AlterQueryServiceCommandWithSecurityDUnitTest {
     });
   }
 
-  private void verifyCurrentAuthorizerClass(Class authorizerClass) {
+  private void verifyCurrentAuthorizerClass(
+      Class<? extends MethodInvocationAuthorizer> authorizerClass) {
     servers.forEach(server -> server.invoke(() -> {
       MethodInvocationAuthorizer methodAuthorizer =
           Objects.requireNonNull(ClusterStartupRule.getCache())
@@ -132,7 +134,7 @@ public class AlterQueryServiceCommandWithSecurityDUnitTest {
     }));
   }
 
-  private String getFilePath() throws IOException {
+  private String getTestMethodAuthorizerFilePath() throws IOException {
     URL url = getClass().getResource(TEST_AUTHORIZER_TXT);
     File textFile = this.tempFolder.newFile(TEST_AUTHORIZER_TXT);
     FileUtils.copyURLToFile(url, textFile);
@@ -186,7 +188,8 @@ public class AlterQueryServiceCommandWithSecurityDUnitTest {
     verifyCurrentAuthorizerClass(DEFAULT_AUTHORIZER_CLASS);
     Class<TestMethodAuthorizer> authorizerClass = TestMethodAuthorizer.class;
     String authorizerName = authorizerClass.getName();
-    String classContent = new String(Files.readAllBytes(Paths.get(getFilePath())));
+    String classContent =
+        new String(Files.readAllBytes(Paths.get(getTestMethodAuthorizerFilePath())));
     String jarFileName = "testJar.jar";
     File jarFile = tempFolder.newFile(jarFileName);
     new ClassBuilder().writeJarFromContent(authorizerName, classContent, jarFile);
--- a/geode-core/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/AlterQueryServiceCommandWithSecurityDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/AlterQueryServiceCommandWithSecurityDUnitTest.java
@@ -15,9 +15,9 @@
 
 package org.apache.geode.management.internal.cli.commands;
 
+import static org.apache.geode.management.internal.cli.commands.AlterQueryServiceCommand.AUTHORIZER_NAME;
 import static org.apache.geode.management.internal.cli.commands.AlterQueryServiceCommand.AUTHORIZER_PARAMETERS;
 import static org.apache.geode.management.internal.cli.commands.AlterQueryServiceCommand.COMMAND_NAME;
-import static org.apache.geode.management.internal.cli.commands.AlterQueryServiceCommand.METHOD_AUTHORIZER_NAME;
 import static org.apache.geode.management.internal.cli.commands.AlterQueryServiceCommand.NO_MEMBERS_FOUND_MESSAGE;
 import static org.apache.geode.management.internal.cli.commands.AlterQueryServiceCommand.SPLITTING_REGEX;
 import static org.assertj.core.api.Assertions.assertThat;
@@ -61,14 +61,17 @@ import org.apache.geode.test.junit.rules.VMProvider;
 public class AlterQueryServiceCommandWithSecurityDUnitTest {
   private static final Class<RestrictedMethodAuthorizer> DEFAULT_AUTHORIZER_CLASS =
       RestrictedMethodAuthorizer.class;
-  private static final String JAVA_BEAN_AUTHORIZER_PARAMS = "java.lang;java.util";
-  // A regex containing a comma to confirm that the '--authorizer-parameter' option is parsed
-  // correctly
+  private static final String TEST_AUTHORIZER_TXT = "TestMethodAuthorizer.txt";
+  private static final String USER_AUTHORIZER_PARAMETERS = "param1;param2;param3";
+  private static final String JAVA_BEAN_AUTHORIZER_PARAMETERS = "java.lang;java.util";
+  // A regex containing a comma to confirm that '--authorizer-parameter' is correctly parsed
   private static final String REGEX_AUTHORIZER_PARAMETERS =
       "^java.util.List..{4,8}$;^java.util.Set..{4,8}$";
-  private static final String USER_AUTHORIZER_PARAMETERS = "param1;param2;param3";
-  private static final String TEST_AUTHORIZER_TXT = "TestMethodAuthorizer.txt";
-  private final int serversToStart = 2;
+  private MemberVM locator;
+  private List<MemberVM> servers = new ArrayList<>();
+
+  @Rule
+  public GfshCommandRule gfsh = new GfshCommandRule();
 
   @Rule
   public TemporaryFolder tempFolder = new TemporaryFolder();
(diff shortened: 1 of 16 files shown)


jujoramos:

geode-core/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/AlterQueryServiceCommandWithSecurityDUnitTest.java
```diff
@@ -61,81 +61,119 @@
...
+  private void verifyCurrentAuthorizerClass(Class authorizerClass) {
+    servers.forEach(server -> server.invoke(() -> {
+      MethodInvocationAuthorizer methodAuthorizer =
+          Objects.requireNonNull(ClusterStartupRule.getCache())
+              .getService(QueryConfigurationService.class).getMethodAuthorizer();
+      assertThat(methodAuthorizer.getClass()).isEqualTo(authorizerClass);
+    }));
+  }
+
+  private String getFilePath() throws IOException {
```
DonalEvans: Since this method now always gets the file path for the `TestMethodAuthorizer.txt` file, it might be better to rename it something like "getTestMethodAuthorizerFilePath" to make it clearer what it's doing when called.

geode-core/src/test/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceImplTest.java
```diff
@@ -43,49 +54,70 @@
...
+  private void setAllowUntrustedMethodInvocationSystemProperty() {
+    System.setProperty(ALLOW_UNTRUSTED_METHOD_INVOCATION_SYSTEM_PROPERTY, "true");
   }
 
   @Test
   public void initThrowsExceptionWhenCacheIsNull() {
     assertThatThrownBy(() -> configService.init(null))
-        .isInstanceOf(IllegalArgumentException.class);
+        .isInstanceOf(IllegalArgumentException.class)
+        .hasMessage("Cache must not be null");
```
DonalEvans: If this verifcation of the message is being done, then the string used to generate the message in `QueryConfigurationServiceImpl` should be extracted to a constant and referenced here.

geode-core/src/test/java/org/apache/geode/management/internal/cli/commands/AlterQueryServiceCommandTest.java
```diff
@@ -50,163 +50,191 @@
...
+    }
+
+    return commandBuilder.toString();
   }
 
   @Test
-  public void commandReturnsCorrectMessageWithNoOptions() {
-    gfsh.executeAndAssertThat(command, COMMAND_NAME).statusIsError()
-        .containsOutput(NO_ARGUMENTS_MESSAGE);
+  public void commandParsesForceUpdateCorrectly() {
```
DonalEvans: Is this test needed here? It seems to be testing the functionality of the gfsh parser rather than the contents of the `AltherQueryServiceCommand` class.

geode-core/src/main/java/org/apache/geode/management/internal/cli/commands/AlterQueryServiceCommand.java
```diff
@@ -31,78 +29,70 @@
...
       operation = ResourcePermission.Operation.MANAGE)
   public ResultModel execute(
-      @CliOption(key = METHOD_AUTHORIZER_NAME,
-          help = METHOD_AUTHORIZER_NAME_HELP) String methodAuthorizerName,
-      @CliOption(key = AUTHORIZER_PARAMETERS,
-          help = AUTHORIZER_PARAMETERS_HELP,
-          optionContext = "splittingRegex=" + SPLITTING_REGEX) String[] authorizerParameters) {
-    ResultModel result;
+      @CliOption(key = AUTHORIZER_NAME, help = METHOD_AUTHORIZER_NAME_HELP,
+          mandatory = true) String methodAuthorizerName,
```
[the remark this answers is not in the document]
jujoramos: Thanks for the explanation, it makes total sense!. I'd leave as it is right now, though, and make it optional once we actually need it.

geode-core/src/main/java/org/apache/geode/cache/query/internal/QueryConfigurationService.java
```diff
@@ -24,12 +24,41 @@
...
+   */
   MethodInvocationAuthorizer getMethodAuthorizer();
 
-  void updateMethodAuthorizer(Cache cache, QueryMethodAuthorizerCreation creation)
-      throws ClassNotFoundException;
-
-  void updateMethodAuthorizer(Cache cache, String className, Set<String> parameters)
-      throws ClassNotFoundException, QueryConfigurationServiceException;
+  /**
+   * Updates the configured {@link MethodInvocationAuthorizer}.
```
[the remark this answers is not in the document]
jujoramos: Good point, thanks.
I've updated the javadoc and also deleted the `forceUpdate` parameter from this method, it should always be `true` whenever we're creating the cache from scratch.

geode-core/src/main/java/org/apache/geode/cache/query/internal/QueryConfigurationService.java
```diff
@@ -24,12 +24,41 @@
...
+   *
+   * @param cache the cache on which the {@link MethodInvocationAuthorizer} instance will be
+   *        configured.
+   * @param forceUpdate {@code true} to apply the configuration change even when there are
+   *        continuous queries running, {@code false} otherwise.
+   * @param className the fully qualified name of the class that will be created and configured as
+   *        the new {@link MethodInvocationAuthorizer}.
+   * @param parameters the list of parameters that will be used to initialize the new
```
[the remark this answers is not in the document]
jujoramos: Done!.

geode-core/src/main/java/org/apache/geode/management/internal/cli/commands/AlterQueryServiceCommand.java
```diff
@@ -31,78 +29,70 @@
...
-      "A semicolon separated list of all parameter values for the specified method authorizer. Requires '--"
-          + METHOD_AUTHORIZER_NAME + "' option to be used";
-  static final String PARAMETERS_WITHOUT_AUTHORIZER_MESSAGE = "The '--" + AUTHORIZER_PARAMETERS
-      + "' option requires '--" + METHOD_AUTHORIZER_NAME + "' to be specified";
-  static final String NO_ARGUMENTS_MESSAGE =
-      "No arguments were provided. No changes have been applied.";
+      "A semicolon separated list of all parameter values for the specified method authorizer.";
+  static final String FORCE_UPDATE = "force-update";
+  private static final String FORCE_PARAMETER_HELP =
+      "Flag to indicate whether to forcibly update the currently configured authorizer, even when there are continuous queries registered (use with caution)";
```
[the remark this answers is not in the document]
jujoramos: I've changed the parameter name, thanks for pointing that out.
I'm not sure about adding the part about invalidating CQs, though, it might be too much for actually being "helpful" to a user (maybe they don't even have CQs at all, adding confusion instead of help). Also, we only invalidate the internal cache used by CQs (internal mechanism totally unknown to our users), I wouldn't know how to phrase this warning without going into much detail and confuse even more the operator...
The user always has the option to go and check the documentation if needed, and the `use with caution` part should be "enough" to make them think twice before using the flag.
Please let me know your thoughts around this.

geode-core/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/AlterQueryServiceCommandWithSecurityDUnitTest.java
jujoramos: Sounds good, done!.

geode-core/src/test/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceImplTest.java
jujoramos: Done!.

geode-core/src/test/java/org/apache/geode/management/internal/cli/commands/AlterQueryServiceCommandTest.java
jujoramos: I wrote the test **before** actually adding the new parameter `forceUpdate` with its associated constraints (`specifiedDefaultValue = "true", unspecifiedDefaultValue = "false"`), so I'm actually testing that what I did was right.
Long story short, I used TDD for this one and I'd like to keep these tests.

geode-core/src/main/java/org/apache/geode/management/internal/cli/commands/AlterQueryServiceCommand.java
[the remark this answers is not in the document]
DonalEvans: I think you're right that if the cache that's invalidated is entirely internal then we don't need to tell users about it. At that point it's just making things more confusing.

geode-core/src/main/java/org/apache/geode/management/internal/cli/commands/AlterQueryServiceCommand.java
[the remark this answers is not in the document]
DonalEvans: I think you're probably right. We don't even know at this point if we do plan to add other persistence to the `QueryConfigurationService`, so we can cross that bridge when we come to it.

geode-core/src/test/java/org/apache/geode/management/internal/cli/commands/AlterQueryServiceCommandTest.java
DonalEvans: That makes sense.

geode-core/src/main/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceImpl.java
```diff
@@ -141,12 +156,18 @@ public void updateMethodAuthorizer(Cache cache, String className, Set<String> pa
...
+
         MethodInvocationAuthorizer tmpAuthorizer =
             (MethodInvocationAuthorizer) userClass.newInstance();
         tmpAuthorizer.initialize(cache, parameters);
         this.authorizer = tmpAuthorizer;
-
       }
+
+      // Invalidate cached CQ results, if any.
+      cqService.getAllCqs().forEach(cqQuery -> {
```
jhuynh1: Should this be pulled into a method?  It would be self documenting (no comment needed) if the method name was correctly chosen

jujoramos: Done!

jujoramos:
Thanks for reviewing @jhuynh1 
Please find my answers below, inline for an easier follow-up.

> I thought clearing the local cq cache was needed if we updated the cq context with the new authorizer.

The CQ will always pick up the latest configured authorizer, no matter wether the `forceUpdate` flag is set as `true` or `false`, this logic is part of the `reset` method invoked by CQs during the execution. See [`CqServiceImpl.evaluateQuery`](https://github.com/apache/geode/blob/develop/geode-cq/src/main/java/org/apache/geode/cache/query/cq/internal/CqServiceImpl.java#L1604-L1607) and [`ExecutionContext.reset`](https://github.com/apache/geode/blob/develop/geode-core/src/main/java/org/apache/geode/cache/query/internal/ExecutionContext.java#L587-L591).

> If they have the old authorizer, why do we need to clear the cache?

If we don't clear the cache, the CQ won't be evaluated for those keys that are part of the cache, which can result in invoking `onEvent` instead of `onError` for the associated `CqListener`.

---

As an example, let's say the user has the CQ  `SELECT * from /testRegion object WHERE object.getId() > -1` running in the cluster, the configured `MethodInvocationAuthorizer` allows `getId` to be invoked. Some time passes and the CQ internal cache builds up to improve performance.
At that point the user decides to install a new `MethodInvocationAuthorizer` that doesn't allow `getId`. If we don't invalidate the internal CQ cache, the engine will invoke `onError` for events **that are not part of the internal cache** (the CQ will be executed) and `onEvent` for those events that **that are part of the internal cache** (the CQ won't be executed), which is not consistent.
Long story short, by invalidating the internal cache we make sure the CQ will be re-executed for every event  no matter what, and the execution itself will result in error if the newly configured authorizer doesn't allow the execution of methods present in the CQ expression.
Hope it makes sense now, please let me know your thoughts.
Cheers.

PR #4386, opened 2019-11-28
GEODE-7504: region eviction support

jmelchio:

### For all changes:

### Note:


jmelchio:
@jinmeiliao @dschneider-pivotal @demery-pivotal

geode-management/src/test/java/org/apache/geode/management/configuration/RegionTest.java
```diff
@@ -64,6 +64,72 @@ public void correctJson() throws Exception {
...
+    mapper.readValue(json2, Region.class);
+  }
+
+  @Test
+  public void correctJsonWithEviction() throws Exception {
+    String evictJson = "{\n"
+        + "  \"name\": \"region1\",\n"
+        + "  \"type\": \"PARTITION\",\n"
+        + "  \"eviction\": {\n"
+        + "    \"type\": \"ENTRY_COUNT\",\n"
```
jinmeiliao: if the json string has no "type":"ENTRY_COUNT", would the type being set correctly?

geode-management/src/main/java/org/apache/geode/management/configuration/Region.java
```diff
@@ -213,4 +224,74 @@ public void setAction(ExpirationAction action) {
...
+    public void setAction(EvictionAction action) {
+      this.action = action;
+    }
+
+    public Integer getEntryCount() {
+      return entryCount;
+    }
+
+    public void setEntryCount(Integer entryCount) {
+      this.entryCount = entryCount;
```
jinmeiliao: I thought setEntryCount would also set the type implicitly.

jmelchio: It does get set in the cluster configuration after conversion.

geode-management/src/test/java/org/apache/geode/management/configuration/RegionTest.java
jmelchio: Yes, upon conversion we assume that given that we have an entryCount value that the type is `ENTRY_COUNT` and we save it as such in the cluster configuration.

geode-management/src/main/java/org/apache/geode/management/configuration/Region.java
```diff
@@ -271,19 +268,27 @@ public void setAction(EvictionAction action) {
...
-      return entryCount;
+      if (type == EvictionType.ENTRY_COUNT) {
+        return limit;
+      }
+      return null;
     }
 
     public void setEntryCount(Integer entryCount) {
-      this.entryCount = entryCount;
+      type = EvictionType.ENTRY_COUNT;
```
dschneider-pivotal: I think I liked the way this was before this change. Now if a user sets both entryCount and memorySizeMb then the last one wins. If a users thinks they can set both (they might think we support both an entry count limit AND a memory size limit) then they would set both. Before this would have given them an error but now it just switches to the last one. And if they are using JSON then how do they even know which one is last? If they do this {entryCount=10; memorySizeInMb=100} they will not get an error anymore but instead it will be either one or the other. And if they also set the type to HEAP_PERCENTAGE then this may also get reset if these settors are invoked later. Before we just had each settor setting one attribute which I think is clearer to users.

jmelchio: Yes, that is an issue with this approach.

jinmeiliao: My issue with the previous approach is that it gives user too many opportunities to create an invalid eviction and it relies heavily on the validator to catch all the invalid cases.

jinmeiliao: I pushed in another commit to validate type been changed by setter.  Please check if that addressed Darrel's concern.

jmelchio: > My issue with the previous approach is that it gives user too many opportunities to create an invalid eviction and it relies heavily on the validator to catch all the invalid cases.

Given the JSON that a user can submit that does not really change with either approach. I think the point of the validator is that it does exactly that; catch the cases in which the user submits invalid requests.

jinmeiliao: I think the differences is the later approach would give user less opportunities to create an invalid object to begin with, whereas the the previous approach will let user create any combination of attributes and let validator to catch all the invalid cases.

geode-management/src/main/java/org/apache/geode/management/configuration/Region.java
```diff
@@ -213,4 +224,88 @@ public void setAction(ExpirationAction action) {
...
+      this.action = action;
+      this.limit = limit;
+      this.objectSizer = objectSizer;
+    }
+
+    public EvictionType getType() {
+      return type;
+    }
+
+    public void setType(EvictionType type) {
```
dschneider-pivotal: add javadocs describing when it will throw

geode-management/src/main/java/org/apache/geode/management/configuration/Region.java
```diff
@@ -213,4 +224,88 @@ public void setAction(ExpirationAction action) {
...
+    }
+
+    public Integer getEntryCount() {
+      if (type == EvictionType.ENTRY_COUNT) {
+        return limit;
+      }
+      return null;
+    }
+
+    public void setEntryCount(Integer entryCount) {
```
dschneider-pivotal: add javadocs describing when it will throw
also describe that it will also set the type to...l

geode-management/src/main/java/org/apache/geode/management/configuration/Region.java
```diff
@@ -213,4 +224,88 @@ public void setAction(ExpirationAction action) {
...
+    }
+
+    public Integer getMemorySizeMb() {
+      if (type == EvictionType.MEMORY_SIZE) {
+        return limit;
+      }
+      return null;
+    }
+
+    public void setMemorySizeMb(Integer memorySizeMb) {
```
dschneider-pivotal: add javadocs describing when it will throw
also describe that it will also set the type to...l

geode-management/src/main/java/org/apache/geode/management/configuration/Region.java
dschneider-pivotal: I like the changes you have made. It seems like the constructors on Eviction still allow for just as many illegal creations to be done. The settors now do more validation but the constructor that takes all the attributes does not do any validation. I'm not sure if it matters to the user if the validation happens on a method/constructor call vs. when they call ClusterManagementService.create.

PR #4387, opened 2019-11-29
GEODE-7458: Adding option in gfsh command "start gateway sender" to control clearing of existing queues

mivanac:
In this PR clearing of queue is added as option at starting of gateway sender.
Also behavior of parallel and serial gateway senders are now aligned.
Due to that, queue is no longer cleared at stopping of gateway sender,
since we now have option whether we want it to be cleared or restored, at restarting of gateway sender.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/cache/asyncqueue/internal/SerialAsyncEventQueueImpl.java
+++ b/geode-core/src/main/java/org/apache/geode/cache/asyncqueue/internal/SerialAsyncEventQueueImpl.java
@@ -193,8 +193,6 @@ public class SerialAsyncEventQueueImpl extends AbstractGatewaySender {
     InternalDistributedSystem system =
         (InternalDistributedSystem) this.cache.getDistributedSystem();
     system.handleResourceEvent(ResourceEvent.GATEWAYSENDER_STOP, this);
-
-    // this.eventProcessor = null;
   }
 
   @Override
--- a/geode-core/src/main/java/org/apache/geode/cache/asyncqueue/internal/SerialAsyncEventQueueImpl.java
+++ b/geode-core/src/main/java/org/apache/geode/cache/asyncqueue/internal/SerialAsyncEventQueueImpl.java
@@ -65,7 +65,7 @@ public class SerialAsyncEventQueueImpl extends AbstractGatewaySender {
   }
 
   @Override
-  public void startWithClean() {
+  public void startWithCleanQueue() {
     this.start(true);
   }
 
--- a/geode-core/src/main/java/org/apache/geode/cache/asyncqueue/internal/SerialAsyncEventQueueImpl.java
+++ b/geode-core/src/main/java/org/apache/geode/cache/asyncqueue/internal/SerialAsyncEventQueueImpl.java
@@ -86,7 +86,7 @@ public class SerialAsyncEventQueueImpl extends AbstractGatewaySender {
           getSenderAdvisor().makeSecondary();
         }
       }
-      eventProcessor = createEventProcessor();
+      eventProcessor = createEventProcessor(cleanQueues);
 
       if (startEventProcessorInPausedState) {
         pauseEvenIfProcessorStopped();
@@ -113,15 +113,15 @@ public class SerialAsyncEventQueueImpl extends AbstractGatewaySender {
     }
   }
 
-  protected AbstractGatewaySenderEventProcessor createEventProcessor() {
+  protected AbstractGatewaySenderEventProcessor createEventProcessor(boolean cleanQueues) {
     AbstractGatewaySenderEventProcessor eventProcessor;
     if (getDispatcherThreads() > 1) {
       eventProcessor = new ConcurrentSerialGatewaySenderEventProcessor(
-          SerialAsyncEventQueueImpl.this, getThreadMonitorObj());
+          SerialAsyncEventQueueImpl.this, getThreadMonitorObj(), cleanQueues);
     } else {
       eventProcessor =
           new SerialGatewaySenderEventProcessor(SerialAsyncEventQueueImpl.this, getId(),
-              getThreadMonitorObj());
+              getThreadMonitorObj(), cleanQueues);
     }
     return eventProcessor;
   }
--- a/geode-core/src/main/java/org/apache/geode/cache/asyncqueue/internal/SerialAsyncEventQueueImpl.java
+++ b/geode-core/src/main/java/org/apache/geode/cache/asyncqueue/internal/SerialAsyncEventQueueImpl.java
@@ -60,7 +60,16 @@ public class SerialAsyncEventQueueImpl extends AbstractGatewaySender {
   }
 
   @Override
-  public void start(boolean cleanQueues) {
+  public void start() {
+    this.start(false);
+  }
+
+  @Override
+  public void startWithClean() {
+    this.start(true);
+  }
+
+  private void start(boolean cleanQueues) {
     if (logger.isDebugEnabled()) {
       logger.debug("Starting gatewaySender : {}", this);
     }
--- a/geode-core/src/main/java/org/apache/geode/cache/asyncqueue/internal/SerialAsyncEventQueueImpl.java
+++ b/geode-core/src/main/java/org/apache/geode/cache/asyncqueue/internal/SerialAsyncEventQueueImpl.java
@@ -194,7 +194,7 @@ public class SerialAsyncEventQueueImpl extends AbstractGatewaySender {
         (InternalDistributedSystem) this.cache.getDistributedSystem();
     system.handleResourceEvent(ResourceEvent.GATEWAYSENDER_STOP, this);
 
-    this.eventProcessor = null;
+    // this.eventProcessor = null;
   }
 
   @Override
--- a/geode-core/src/main/java/org/apache/geode/cache/asyncqueue/internal/SerialAsyncEventQueueImpl.java
+++ b/geode-core/src/main/java/org/apache/geode/cache/asyncqueue/internal/SerialAsyncEventQueueImpl.java
@@ -60,7 +60,7 @@ public class SerialAsyncEventQueueImpl extends AbstractGatewaySender {
   }
 
   @Override
-  public void start() {
+  public void start(boolean cleanQueues) {
     if (logger.isDebugEnabled()) {
       logger.debug("Starting gatewaySender : {}", this);
     }
--- a/geode-core/src/main/java/org/apache/geode/cache/wan/GatewaySender.java
+++ b/geode-core/src/main/java/org/apache/geode/cache/wan/GatewaySender.java
@@ -178,11 +178,15 @@ public interface GatewaySender {
   /**
    * Starts this GatewaySender. Once the GatewaySender is running, its configuration cannot be
    * changed.
-   * In case, GatewaySender was previosly stopped, and some events wefe leftin queue, with option
-   * cleanQueues
-   * we choose whether to restore or discard previos queue. Default is to restore queue.
    */
-  void start(boolean cleanQueues);
+  void start();
+
+  /**
+   * Starts this GatewaySender and discards previous queue content. Once the GatewaySender is
+   * running,
+   * its configuration cannot be changed.
+   */
+  void startWithClean();
 
   /**
    * Stops this GatewaySender. The scope of this operation is the VM on which it is invoked. In case
--- a/geode-core/src/main/java/org/apache/geode/cache/wan/GatewaySender.java
+++ b/geode-core/src/main/java/org/apache/geode/cache/wan/GatewaySender.java
@@ -186,7 +186,7 @@ public interface GatewaySender {
    * running,
    * its configuration cannot be changed.
    */
-  void startWithClean();
+  void startWithCleanQueue();
 
   /**
    * Stops this GatewaySender. The scope of this operation is the VM on which it is invoked. In case
--- a/geode-core/src/main/java/org/apache/geode/cache/wan/GatewaySender.java
+++ b/geode-core/src/main/java/org/apache/geode/cache/wan/GatewaySender.java
@@ -178,8 +178,11 @@ public interface GatewaySender {
   /**
    * Starts this GatewaySender. Once the GatewaySender is running, its configuration cannot be
    * changed.
+   * In case, GatewaySender was previosly stopped, and some events wefe leftin queue, with option
+   * cleanQueues
+   * we choose whether to restore or discard previos queue. Default is to restore queue.
    */
-  void start();
+  void start(boolean cleanQueues);
 
   /**
    * Stops this GatewaySender. The scope of this operation is the VM on which it is invoked. In case
--- a/geode-core/src/main/java/org/apache/geode/cache/wan/GatewaySender.java
+++ b/geode-core/src/main/java/org/apache/geode/cache/wan/GatewaySender.java
@@ -183,8 +183,7 @@ public interface GatewaySender {
 
   /**
    * Starts this GatewaySender and discards previous queue content. Once the GatewaySender is
-   * running,
-   * its configuration cannot be changed.
+   * running, its configuration cannot be changed.
    */
   void startWithCleanQueue();
 
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/ColocationHelper.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/ColocationHelper.java
@@ -209,8 +209,6 @@ public class ColocationHelper {
       for (DiskStore diskStore : stores) {
         // Look at all of the partitioned regions.
 
-        boolean anyChildOnline = checkAnyChildOnline(diskStore, region);
-
         for (Map.Entry<String, PRPersistentConfig> entry : ((DiskStoreImpl) diskStore).getAllPRs()
             .entrySet()) {
 
@@ -224,7 +222,7 @@ public class ColocationHelper {
             if (childRegion == null) {
               // If the child region is offline, return true
               // unless it is a parallel queue that the user has removed.
-              if (!ignoreUnrecoveredQueue(region, childName, anyChildOnline)) {
+              if (!ignoreUnrecoveredQueue(region, childName)) {
                 region.addMissingColocatedRegionLogger(childName);
                 hasOfflineChildren = true;
               }
@@ -245,31 +243,7 @@ public class ColocationHelper {
     return hasOfflineChildren;
   }
 
-
-  private static boolean checkAnyChildOnline(DiskStore diskStore, PartitionedRegion region) {
-    // Check if any of child regions is ParallelGatewaySenderQueue and is online
-    InternalCache cache = region.getCache();
-
-    for (Map.Entry<String, PRPersistentConfig> entry : ((DiskStoreImpl) diskStore).getAllPRs()
-        .entrySet()) {
-
-      PRPersistentConfig config = entry.getValue();
-      String childName = entry.getKey();
-
-      // Check to see if they're colocated with this region.
-      if (region.getFullPath().equals(config.getColocatedWith())) {
-        PartitionedRegion childRegion = (PartitionedRegion) cache.getRegion(childName);
-        if (childRegion != null && ParallelGatewaySenderQueue.isParallelQueue(childName)) {
-          return true;
-        }
-      }
-    }
-    return false;
-  }
-
-
-  private static boolean ignoreUnrecoveredQueue(PartitionedRegion region, String childName,
-      boolean anyChildOnline) {
+  private static boolean ignoreUnrecoveredQueue(PartitionedRegion region, String childName) {
     // Hack for #50120 if the childRegion is an async queue, but we
     // no longer define the async queue, ignore it.
     if (!ParallelGatewaySenderQueue.isParallelQueue(childName)) {
@@ -277,10 +251,6 @@ public class ColocationHelper {
     }
 
     String senderId = ParallelGatewaySenderQueue.getSenderId(childName);
-
-    if (anyChildOnline && region.getParallelGatewaySenderIds().contains(senderId)) {
-      return true;
-    }
 
     if (!region.getAsyncEventQueueIds().contains(senderId)
         && !region.getParallelGatewaySenderIds().contains(senderId) && IGNORE_UNRECOVERED_QUEUE) {
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/ColocationHelper.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/ColocationHelper.java
@@ -208,6 +208,9 @@ public class ColocationHelper {
       // Look through all of the disk stores for offline colocated child regions
       for (DiskStore diskStore : stores) {
         // Look at all of the partitioned regions.
+
+        boolean anyChildOnline = checkAnyChildOnline(diskStore, region);
+
         for (Map.Entry<String, PRPersistentConfig> entry : ((DiskStoreImpl) diskStore).getAllPRs()
             .entrySet()) {
 
@@ -221,7 +224,7 @@ public class ColocationHelper {
             if (childRegion == null) {
               // If the child region is offline, return true
               // unless it is a parallel queue that the user has removed.
-              if (!ignoreUnrecoveredQueue(region, childName)) {
+              if (!ignoreUnrecoveredQueue(region, childName, anyChildOnline)) {
                 region.addMissingColocatedRegionLogger(childName);
                 hasOfflineChildren = true;
               }
@@ -243,7 +246,30 @@ public class ColocationHelper {
   }
 
 
-  private static boolean ignoreUnrecoveredQueue(PartitionedRegion region, String childName) {
+  private static boolean checkAnyChildOnline(DiskStore diskStore, PartitionedRegion region) {
+    // Check if any of child regions is ParallelGatewaySenderQueue and is online
+    InternalCache cache = region.getCache();
+
+    for (Map.Entry<String, PRPersistentConfig> entry : ((DiskStoreImpl) diskStore).getAllPRs()
+        .entrySet()) {
+
+      PRPersistentConfig config = entry.getValue();
+      String childName = entry.getKey();
+
+      // Check to see if they're colocated with this region.
+      if (region.getFullPath().equals(config.getColocatedWith())) {
+        PartitionedRegion childRegion = (PartitionedRegion) cache.getRegion(childName);
+        if (childRegion != null && ParallelGatewaySenderQueue.isParallelQueue(childName)) {
+          return true;
+        }
+      }
+    }
+    return false;
+  }
+
+
+  private static boolean ignoreUnrecoveredQueue(PartitionedRegion region, String childName,
+      boolean anyChildOnline) {
     // Hack for #50120 if the childRegion is an async queue, but we
     // no longer define the async queue, ignore it.
     if (!ParallelGatewaySenderQueue.isParallelQueue(childName)) {
@@ -252,7 +278,7 @@ public class ColocationHelper {
 
     String senderId = ParallelGatewaySenderQueue.getSenderId(childName);
 
-    if (region.getParallelGatewaySenderIds().contains(senderId)) {
+    if (anyChildOnline && region.getParallelGatewaySenderIds().contains(senderId)) {
       return true;
     }
 
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/ColocationHelper.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/ColocationHelper.java
@@ -251,6 +251,11 @@ public class ColocationHelper {
     }
 
     String senderId = ParallelGatewaySenderQueue.getSenderId(childName);
+
+    if (region.getParallelGatewaySenderIds().contains(senderId)) {
+      return true;
+    }
+
     if (!region.getAsyncEventQueueIds().contains(senderId)
         && !region.getParallelGatewaySenderIds().contains(senderId) && IGNORE_UNRECOVERED_QUEUE) {
       return true;
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ConcurrentParallelGatewaySenderQueue.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ConcurrentParallelGatewaySenderQueue.java
@@ -18,6 +18,7 @@ import java.util.List;
 import java.util.Set;
 import java.util.concurrent.BlockingQueue;
 
+import org.apache.geode.annotations.VisibleForTesting;
 import org.apache.geode.cache.CacheException;
 import org.apache.geode.cache.CacheListener;
 import org.apache.geode.cache.Region;
@@ -79,6 +80,12 @@ public class ConcurrentParallelGatewaySenderQueue implements RegionQueue {
 
   public Set<PartitionedRegion> getRegions() {
     return ((ParallelGatewaySenderQueue) (processors[0].getQueue())).getRegions();
+  }
+
+  // This is for UT purpose only
+  @VisibleForTesting
+  public boolean getCleanQueues() {
+    return ((ParallelGatewaySenderQueue) (processors[0].getQueue())).getCleanQueues();
   }
 
   @Override
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ConcurrentParallelGatewaySenderQueue.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ConcurrentParallelGatewaySenderQueue.java
@@ -82,7 +82,6 @@ public class ConcurrentParallelGatewaySenderQueue implements RegionQueue {
     return ((ParallelGatewaySenderQueue) (processors[0].getQueue())).getRegions();
   }
 
-  // This is for UT purpose only
   @VisibleForTesting
   public boolean getCleanQueues() {
     return ((ParallelGatewaySenderQueue) (processors[0].getQueue())).getCleanQueues();
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
@@ -543,7 +543,7 @@ public class ParallelGatewaySenderQueue implements RegionQueue {
         if (logger.isDebugEnabled()) {
           logger.debug("{}: Created queue region: {}", this, prQ);
         }
-        if ((prQ != null) && this.cleanQueues && sender.isPersistenceEnabled()) {
+        if ((prQ != null) && this.cleanQueues) {
           // now, clean up the shadowPR's buckets on this node (primary as well as
           // secondary) for a fresh start
           Set<BucketRegion> localBucketRegions = prQ.getDataStore().getAllLocalBucketRegions();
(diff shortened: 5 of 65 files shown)


geode-core/src/main/java/org/apache/geode/cache/wan/GatewaySender.java
```diff
@@ -178,8 +178,11 @@
   /**
    * Starts this GatewaySender. Once the GatewaySender is running, its configuration cannot be
    * changed.
+   * In case, GatewaySender was previosly stopped, and some events wefe leftin queue, with option
+   * cleanQueues
+   * we choose whether to restore or discard previos queue. Default is to restore queue.
    */
-  void start();
+  void start(boolean cleanQueues);
```
jujoramos: This is a public interface, adding this parameter here would break backward compatibility, which is not "allowed" unless we're releasing a major version of the product. "Common" users generally don't implement this interface at all, fair enough, but there are some frameworks (like `spring-data-geoge`) that do (see [here](https://github.com/spring-projects/spring-data-geode/blob/master/spring-data-geode/src/main/java/org/springframework/data/gemfire/wan/GatewaySenderWrapper.java#L34)), and they will be in trouble if we break the backward compatibility.

mivanac: Thanks for comment.

geode-core/src/main/java/org/apache/geode/management/internal/cli/i18n/CliStrings.java
```diff
@@ -2244,6 +2244,9 @@
   public static final String START_GATEWAYSENDER__MEMBER__HELP =
       "Name/Id of the member on which to start the Gateway Sender.";
 
+  public static final String START_GATEWAYSENDER__CLEAN_QUEUE = "cleanQueues";
+  public static final String START_GATEWAYSENDER__CLEAN_QUEUE__HELP =
+      "Option to clean existing queue at start of the Gateway Sender.";
```
alb3rtobr: Remember to add this option to the documentation of `start gateway-sender` command: https://geode.apache.org/docs/guide/110/tools_modules/gfsh/command-pages/start.html#topic_AB8BA3F42B9645A8BE9BD97CE2F839A8

(`geode/geode-docs/tools_modules/gfsh/command-pages/start.html.md.erb` file)

mivanac: Thanks

geode-core/src/main/java/org/apache/geode/cache/wan/GatewaySender.java
```diff
@@ -181,6 +181,13 @@
    */
   void start();
 
+  /**
+   * Starts this GatewaySender and discards previous queue content. Once the GatewaySender is
+   * running,
+   * its configuration cannot be changed.
+   */
+  void startWithClean();
```
jujoramos: Can we use something more meaningful as the method name?, `startWithCleanQueue`, `startWithEmptyQueue`, something else?.

jujoramos: Can you change the method name to something more meaningful?, `startWithCleanQueue`, `startWithEmptyQueue`, something similar?

geode-core/src/main/java/org/apache/geode/internal/cache/ColocationHelper.java
```diff
@@ -250,6 +250,11 @@ private static boolean ignoreUnrecoveredQueue(PartitionedRegion region, String c
     }
 
     String senderId = ParallelGatewaySenderQueue.getSenderId(childName);
+
+    if (region.getParallelGatewaySenderIds().contains(senderId)) {
+      return true;
+    }
+
```
jujoramos: What's the reasoning behind this change?. 
According to the next `if` statement that was already part of the code, we can return `true` if and only if the list of `ParallelGatewaySenderIds` or `AsyncEventQueueIds` do not contain `childName`:

```
if (!region.getAsyncEventQueueIds().contains(senderId) && !region.getParallelGatewaySenderIds().contains(senderId) && IGNORE_UNRECOVERED_QUEUE) {
  return true;
}
```

geode-core/src/main/java/org/apache/geode/management/GatewaySenderMXBean.java
```diff
@@ -185,6 +185,15 @@
...
   void start();
 
+  /**
+   * Starts this GatewaySender and cleans previous queue content.
+   * Once the GatewaySender is running its configuration cannot be changed.
+   *
+   */
+  @ResourceOperation(resource = Resource.CLUSTER, operation = Operation.MANAGE,
+      target = Target.GATEWAY)
+  void startWithClean();
```
jujoramos: Can you change the method name to something more meaningful?, `startWithCleanQueue`, `startWithEmptyQueue`, something similar?

geode-wan/src/main/java/org/apache/geode/internal/cache/wan/GatewaySenderFactoryImpl.java
```diff
@@ -254,6 +254,7 @@ public GatewaySender create(String id, int remoteDSId) {
         this.cache.addGatewaySender(sender);
 
         if (!this.attrs.isManualStart()) {
+          // set cleanQueues to default value
```
jujoramos: Remove the comment.

geode-wan/src/main/java/org/apache/geode/internal/cache/wan/GatewaySenderFactoryImpl.java
```diff
@@ -275,6 +276,7 @@ public GatewaySender create(String id, int remoteDSId) {
...
```
jujoramos: Remove the comment.

geode-wan/src/main/java/org/apache/geode/internal/cache/wan/GatewaySenderFactoryImpl.java
```diff
@@ -309,6 +311,7 @@ public GatewaySender create(String id) {
         sender = new ParallelGatewaySenderImpl(cache, statisticsClock, attrs);
         this.cache.addGatewaySender(sender);
         if (!this.attrs.isManualStart()) {
+          // set cleanQueues to default value
```
jujoramos: Remove the comment.

geode-wan/src/main/java/org/apache/geode/internal/cache/wan/GatewaySenderFactoryImpl.java
```diff
@@ -323,6 +326,7 @@ public GatewaySender create(String id) {
         sender = new SerialGatewaySenderImpl(cache, statisticsClock, attrs);
         this.cache.addGatewaySender(sender);
         if (!this.attrs.isManualStart()) {
+          // set cleanQueues to default value
```
jujoramos: Remove the comment.

mivanac: Updated

geode-wan/src/main/java/org/apache/geode/internal/cache/wan/GatewaySenderFactoryImpl.java
mivanac: Updated

geode-wan/src/main/java/org/apache/geode/internal/cache/wan/GatewaySenderFactoryImpl.java
mivanac: Updated

geode-wan/src/main/java/org/apache/geode/internal/cache/wan/GatewaySenderFactoryImpl.java
mivanac: Updated

geode-core/src/main/java/org/apache/geode/management/GatewaySenderMXBean.java
mivanac: Updated

geode-core/src/main/java/org/apache/geode/cache/wan/GatewaySender.java
mivanac: Updated

geode-core/src/main/java/org/apache/geode/internal/cache/ColocationHelper.java
mivanac: In scenario, when multi parallel GW senders are created on same region, and restarted, GW senders are hanging. I modified this part, and I will add test for it.

geode-core/src/main/java/org/apache/geode/management/internal/cli/i18n/CliStrings.java
davebarnes97: @mivanac Choice of camel-case for the gfsh option name (--cleanQueues) deviates from past convention. Camel-case is unreliable given different case-sensitivity conventions in different environments. Past practice is to delineate words using a hyphen. I strongly recommend changing the name to `--clean-queues` in accord with past practice.
Note: I also added a 'docs' component to the JIRA ticket, as docs are affected by this PR.

davebarnes97: @mivinac Your recent commit addressed my issue with the gfsh option name. Thank you!

mivanac:
Added tests for new functionality.

mivanac:
Most tests are added, except UT, since there are no existing for impacted classes.

DonalEvans:
> Most tests are added, except UT, since there are no existing for impacted classes.

I think that even if there are no existing unit tests for a class, you can still create a new test class and write tests to verify the specific changes you've made.

DonalEvans:
@mivanac I'm going on vacation for two weeks starting tomorrow, so I won't be able to actively review any further changes to this PR. I'm dismissing my existing request for changes so that it doesn't block the PR from being merged while I'm away, but I still do feel that unit tests for the affected methods should be added.

mivanac:
Added UT

mivanac:
@jujoramos are added tests OK with you?

jujoramos:
@mivanac I'm on vacations so I won't be able to work any further on this, I'll dismiss my review but please wait for at least one more review before merging the PR (@davebarnes97 approved the changes only from a docs standpoint).

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/parallel/ParallelWANPersistenceEnabledGatewaySenderDUnitTest.java
```diff
@@ -1599,6 +1599,197 @@ public void testRemoveGatewayFromPersistentRegionOnRestart() throws Throwable {
...
   }
 
+
+  /**
+   * Enable persistence for GatewaySender. Pause the sender and do some puts in local region. Stop
+   * GatewaySender.
+   * Then start GatewaySender. Check if the remote site receives all the events.
+   */
+  @Test
+  public void testPRWithGatewaySenderPersistenceEnabled_Restart_Gateway() {
```
aaronlindsey: Ideally all of the test names in this PR should succinctly describe 3 things: subject under test, specific condition under test, and what we expect to happen to the subject in this condition. These names are missing the last part. For example, something like: `persistentWanGateway_restartSender_expectAllEventsReceived` and `persistentWanGateway_restartSenderWithCleanQueues_expectNoEventsReceived` would be clearer about what is expected to happen.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialWANPersistenceEnabledGatewaySenderDUnitTest.java
```diff
@@ -508,4 +508,144 @@ public void testPersistentReplicatedRegionWithGatewaySenderPersistenceEnabled_Re
...
+
+    vm5.invoke(() -> WANTestBase.startSender("ln"));
+    LogWriterUtils.getLogWriter().info("Started the sender in vm 5");
+    try {
+      inv1.join();
+    } catch (InterruptedException e) {
+      fail("Got interrupted exception while waiting for startSender to finish.");
+    }
+
+    Wait.pause(5000);
```
aaronlindsey: Please remove this `Wait.pause(...)` and use `GeodeAwaitility` instead.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialWANPersistenceEnabledGatewaySenderDUnitTest.java
```diff
@@ -508,4 +508,144 @@ public void testPersistentReplicatedRegionWithGatewaySenderPersistenceEnabled_Re
...
+
+    vm5.invoke(() -> WANTestBase.startSenderwithCleanQueues("ln"));
+    LogWriterUtils.getLogWriter().info("Started the sender in vm 5");
+    try {
+      inv1.join();
+    } catch (InterruptedException e) {
+      fail("Got interrupted exception while waiting for startSender to finish.");
+    }
+
+    Wait.pause(5000);
```
aaronlindsey: Please remove this `Wait.pause(...)` and use `GeodeAwaitility` instead.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/wancommand/StartGatewaySenderCommandDUnitTest.java
```diff
@@ -335,6 +335,103 @@ public void testStartGatewaySender_Group_MissingSenderFromGroup() throws Excepti
     server3.invoke(() -> verifySenderState("ln", true, false));
   }
 
+  @Test
+  public void testStartGatewaySender_clean_queues() throws Exception {
```
aaronlindsey: I don't understand what these two tests are testing. I see that one cleans the queue and the other doesn't, but they have the same assertions on the expected behavior. Shouldn't the assertions be different since they do different things?

geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ConcurrentParallelGatewaySenderQueue.java
```diff
@@ -81,6 +82,12 @@ public PartitionedRegion getRegion(String fullpath) {
     return ((ParallelGatewaySenderQueue) (processors[0].getQueue())).getRegions();
   }
 
+  // This is for UT purpose only
```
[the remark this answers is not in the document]
mivanac: Thanks for comments.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/wancommand/StartGatewaySenderCommandDUnitTest.java
mivanac: Thist test is to verify, that new option "clean-queues" is supported, and to verify option values

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialWANPersistenceEnabledGatewaySenderDUnitTest.java
mivanac: updated

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/concurrent/ConcurrentParallelGatewaySenderOperation_2_DUnitTest.java
```diff
@@ -55,6 +56,7 @@ protected final void postSetUpWANTestBase() throws Exception {
     IgnoredException.addIgnoredException("Unexpected IOException");
   }
 
+  @Ignore("NotSupported")
```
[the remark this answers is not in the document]
mivanac: This test was failing, and according to documentation, is not supported.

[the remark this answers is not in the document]
aaronlindsey: I would suggest removing the test if it is no longer testing a valid use case.

mivanac:
Could someone help with UT. After last modification on PR (there was no update of UT), UT is reporting Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "Management Task", and eventually it timeouts. Thanks.

mivanac:
Locally on my own machine, everything is working fine.

geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
```diff
@@ -540,13 +561,23 @@ public void addShadowPartitionedRegionForUserPR(PartitionedRegion userPR) {
         if (logger.isDebugEnabled()) {
           logger.debug("{}: Created queue region: {}", this, prQ);
         }
+        if ((prQ != null) && this.cleanQueues && sender.isPersistenceEnabled()) {
+          // now, clean up the shadowPR's buckets on this node (primary as well as
+          // secondary) for a fresh start
+          Set<BucketRegion> localBucketRegions = prQ.getDataStore().getAllLocalBucketRegions();
```
agingade: I haven't looked at the complete code path...What if this member is joining a running cluster; with another member hosting primary queue (processing on-going events). And when a new member joins with clean  queue option; is the expectation is to remove local and remote events too or only local...The bucketRegion.clear() may be distributed...And may try to remove entries from the redundant queues...Do you have test for this scenario...

mivanac: Added Test case

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/concurrent/ConcurrentParallelGatewaySenderOperation_2_DUnitTest.java
```diff
@@ -55,70 +55,6 @@ protected final void postSetUpWANTestBase() throws Exception {
     IgnoredException.addIgnoredException("Unexpected IOException");
   }
 
-  @Test
```
jhuynh1: Where in the docs is this not supported?  I see a caution but I do know some people do use this feature (which is what lead to discovery and fix for the issue where secondary queues would continue to grow).

Whether or not it's supported, the test currently passes on develop, if this test is not passing with this diff, the diff negatively impacts the test in some way.  Do we have an understanding of why the test is hanging now?

geode-core/src/main/java/org/apache/geode/internal/cache/wan/serial/SerialGatewaySenderQueue.java
```diff
@@ -905,6 +915,10 @@ private void initializeRegion(AbstractGatewaySender sender, CacheListener listen
             new Object[] {this, this.regionName}),
             e);
       }
+      if ((this.region != null) && this.cleanQueues && this.enablePersistence) {
+        this.region.clear();
```
jhuynh1: Is this a distributed operation or only done locally?

geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
```diff
@@ -530,7 +549,9 @@ public void addShadowPartitionedRegionForUserPR(PartitionedRegion userPR) {
           addOverflowStatisticsToMBean(cache, prQ);
 
           // Wait for buckets to be recovered.
-          prQ.shadowPRWaitForBucketRecovery();
+          if (this.asyncEvent || sender.isPersistenceEnabled()) {
```
jhuynh1: Why are we only waiting for async or persistence when before it was in all cases?

geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
```diff
@@ -1590,7 +1621,13 @@ public void cleanUp() {
 
   @Override
   public void close() {
-    // Because of bug 49060 do not close the regions of a parallel queue
+    Region r = getRegion();
+    if (!this.asyncEvent && r != null && !r.isDestroyed()) {
+      try {
+        r.close();
+      } catch (RegionDestroyedException e) {
```
jhuynh1: We catch a region destroyed exception here, do we need the !r.isDestroyed() check in the if block then?

geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
```diff
@@ -1590,7 +1621,13 @@ public void cleanUp() {
 
   @Override
   public void close() {
```
jhuynh1: The modification to this close appears to have broken the test that was removed.  Since none of the new code calls close, what is the new implementation doing for us with the cleanQueues flags?

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/concurrent/ConcurrentParallelGatewaySenderOperation_2_DUnitTest.java
jhuynh1: It looks like the close() implementation causes this to fail.

geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
mivanac: Close is modified, cause inconsistent behavior between parallel and serial gw sender (what is reported as a fault). In case of parallel gw sender, without this modification, content of queue is lost when persistent GW sender is restarted.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/concurrent/ConcurrentParallelGatewaySenderOperation_2_DUnitTest.java
mivanac: If my understanding is correct, according to [https://geode.apache.org/docs/guide/12/topologies_and_comm/topology_concepts/multisite_overview.html](url) , in chapter Stopping Gateway Senders it is stated that:
> The API and gfsh command stops the parallel gateway sender in one member, which causes data loss because events to buckets in that member will be dropped by the stopped sender. The partitioned region does not failover in this scenario since the member is still running. Instead, to ensure that the remaining events are sent, shut down the entire member to ensure proper failover of partition region events.

jhuynh1: I think it's describing the effects of calling these api's directly (which one would assume would mean, not supported/recommended).  However I do know someone has used these api's directly in the past. 

In either case, the close() implementation is causing replication to the other wan site not to occur.  I have some other private tests running that show weird symptoms with this PR that I'll spend some time on.  I'll try to find out why this is not replicating across the wan with this PR and provide more detail.

mivanac: Thanks

geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
mivanac: Previously, this was executed only at first start of GW sender (since at stop region was not closed).

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/concurrent/ConcurrentParallelGatewaySenderOperation_2_DUnitTest.java
jhuynh1: Hi Mario, 
The close in combination with stopping gateways senders can cause odd issues, like PartitionedOfflineExceptions, RegionDestroyedExceptions or behavior like this test is exhibiting.  We have some internal applications that are running into these types of issues with this diff.  Other than our own internal applications I do know of some users that rely on being able to do this.  A change of this behavior should probably have a discuss on the dev list about current vs future behavior and ramifications. Thanks!

geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
mivanac: This is updated, according to your proposal. I t seems that closing of regain causes some additional, problems, like events in diskStore, will never be deleted. So parallel and serial GW sender queue are updated not to close region, at stoping of GW sender.

mivanac: Deleted TC is returned, to prove nothing is broken.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/concurrent/ConcurrentParallelGatewaySenderOperation_2_DUnitTest.java
mivanac: Solution with closing of region is removed, so there will be no odd issues, and weird symptoms.

geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
mivanac: Removed

geode-core/src/main/java/org/apache/geode/internal/cache/wan/serial/SerialGatewaySenderQueue.java
mivanac: Locally.

geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
mivanac: This part is delated, since we are not closing region at stop any more.

mivanac:
@jhuynh1 are new changes OK with you.

jhuynh1:
Hi Mario,
Sorry for the delay, I'm going to take a more thorough look as this diff is quite large.  Please bear with me and I'll hopefully have it reviewed in a couple more days (early next week?)

@agingade, you were also reviewing this at one point.  Perhaps you'd like to re-review?

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/concurrent/ConcurrentParallelGatewaySenderOperation_1_DUnitTest.java
```diff
@@ -649,7 +655,7 @@ public void testParallelPropagationSenderStartAfterStopOnAccessorNode() throws T
     vm7.invoke(() -> WANTestBase.stopSender("ln"));
 
     // SECOND RUN: do some of the puts after the senders are stopped
-    vm4.invoke(() -> WANTestBase.doPuts(getTestMethodName() + "_PR", 1000));
+    vm4.invoke(() -> WANTestBase.doPuts(getTestMethodName() + "_PR", 800));
```
jhuynh1: just curious why this needed to be changed to 800?

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialGatewaySenderOperationsDistributedTest.java
```diff
@@ -618,20 +618,20 @@ public void test_Bug44153_StopOneSender_StartAnotherSender_CheckQueueSize() {
...
     vm5.invoke(() -> doPuts(className + "_RR", 10, 110));
 
-    vm5.invoke(() -> validateQueueContents("ln", 100));
+    vm5.invoke(() -> validateQueueContents("ln", 110));
     vm5.invoke(() -> stopSender("ln"));
     vm5.invoke(() -> validateSenderStoppedState("ln"));
 
     vm4.invoke(() -> startSender("ln"));
-    vm4.invoke(() -> validateQueueContents("ln", 10));
+    vm4.invoke(() -> validateQueueContents("ln", 110));
```
jhuynh1: Is this because we are no longer clearing the queue when stopping a gateway sender?

geode-core/src/main/java/org/apache/geode/cache/asyncqueue/internal/SerialAsyncEventQueueImpl.java
```diff
@@ -185,7 +194,7 @@ public void stop() {
         (InternalDistributedSystem) this.cache.getDistributedSystem();
     system.handleResourceEvent(ResourceEvent.GATEWAYSENDER_STOP, this);
 
-    this.eventProcessor = null;
+    // this.eventProcessor = null;
```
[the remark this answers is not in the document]
mivanac: This is intended. I will remove line.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialGatewaySenderOperationsDistributedTest.java
mivanac: Yes

geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
```diff
@@ -1317,16 +1348,19 @@ private void addPeekedEvents(List<GatewaySenderEventImpl> batch, int batchSize)
...
-        final PartitionedRegion region = (PartitionedRegion) event.getRegion();
-        if (!region.getRegionAdvisor().isPrimaryForBucket(bucketId)) {
-          iterator.remove();
-          BucketRegionQueue brq = getBucketRegionQueueByBucketId(getRandomShadowPR(), bucketId);
-          if (brq != null) {
-            brq.pushKeyIntoQueue(event.getShadowKey());
+      Object[] helpArray = peekedEvents.toArray();
+      if (helpArray.length > 0) {
+
+        for (int i = helpArray.length - 1; i >= 0; i--) {
```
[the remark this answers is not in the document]
mivanac: But we are going through array, in backward direction, index 0 is valid index

geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
```diff
@@ -1317,16 +1348,19 @@ private void addPeekedEvents(List<GatewaySenderEventImpl> batch, int batchSize)
...
-          .hasNext();) {
-        GatewaySenderEventImpl event = iterator.next();
-        final int bucketId = event.getBucketId();
-        final PartitionedRegion region = (PartitionedRegion) event.getRegion();
-        if (!region.getRegionAdvisor().isPrimaryForBucket(bucketId)) {
-          iterator.remove();
-          BucketRegionQueue brq = getBucketRegionQueueByBucketId(getRandomShadowPR(), bucketId);
-          if (brq != null) {
-            brq.pushKeyIntoQueue(event.getShadowKey());
+      Object[] helpArray = peekedEvents.toArray();
```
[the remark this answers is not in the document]
mivanac: We need to to return events from peekedEvents to queue, from last till first.

geode-core/src/main/java/org/apache/geode/internal/cache/wan/serial/SerialGatewaySenderQueue.java
```diff
@@ -907,9 +910,17 @@ private void initializeRegion(AbstractGatewaySender sender, CacheListener listen
             new Object[] {this, this.regionName}),
             e);
       }
+      if ((this.region != null) && this.cleanQueues) {
```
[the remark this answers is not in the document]
mivanac: Yes. Thanks

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/concurrent/ConcurrentParallelGatewaySenderOperation_1_DUnitTest.java
mivanac: There is no need to change this. I will return old value.

geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
```diff
@@ -234,20 +245,28 @@ private Object deserialize(Object serializedBytes) {
...
     this.metaRegionFactory = metaRegionFactory;
 
+    this.cleanQueues = cleanQueues;
+
     this.index = idx;
     this.nDispatcher = nDispatcher;
     this.stats = sender.getStatistics();
     this.sender = sender;
 
+    if (this.sender.getId().contains(AsyncEventQueueImpl.ASYNC_EVENT_QUEUE_PREFIX)) {
```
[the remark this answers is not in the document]
mivanac: Could we trust isAsyncEventQueue(), because it depends on is any listener, which can be set for GatewaySenderFactoryImpl and AsyncEventQueueFactoryImpl.

geode-core/src/main/java/org/apache/geode/internal/cache/ColocationHelper.java
```diff
@@ -242,14 +245,42 @@ private static boolean hasOfflineColocatedChildRegions(PartitionedRegion region)
...
+      boolean anyChildOnline) {
     // Hack for #50120 if the childRegion is an async queue, but we
     // no longer define the async queue, ignore it.
     if (!ParallelGatewaySenderQueue.isParallelQueue(childName)) {
       return false;
     }
 
     String senderId = ParallelGatewaySenderQueue.getSenderId(childName);
+
+    if (anyChildOnline && region.getParallelGatewaySenderIds().contains(senderId)) {
```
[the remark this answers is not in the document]
mivanac: This is not needed any more. this was part of old solution.

jhuynh1:
Hi Mario,

I ran the previous sha through an app I have and am running into a few hangs.  I'll try to debug and see if I can reproduce in a DUnit.... I'll also try to take your latest diff (maybe that will help?)  
Please allow me some time to debug or reproduce as these hangs can be difficult.

mivanac:
Ok. Thanks.

mivanac:
With these last changes, test is not failing.

jhuynh1:
Cool, awesome job.  I'm going to run against some end to end tests and see what else crops up.  Please bear with me.

Also, is there a brief explanation of what was causing the hang and what change was required?  Just for those of us who are curious.

mivanac:
Problem was, that as we are stoping and starting gateway sender, while we are dispatching events, 
in some cases, stopping of gw sender can shutdown ackReaderThread, just after we dispatch batch, so we will never remove this events from queue.

mivanac:
Hi,
Are there any updates regarding this PR?

mivanac:
Hi, can we merge this PR, if there is no more comments?

PR #4390, opened 2019-12-02
GEODE-7454 Document Cluster Management Service (experimental)(

davebarnes97:
Add a description of the (still experimental) Cluster Management Service. Point to the Geode Wiki for details on the REST and Java APIs. Give instructions for querying the REST API directly for a self-description. ~~(Note: check-in is for develop branch, v1.12. Wiki links point to v1.11 because it's the most recent available at this time.)~~


geode-docs/configuring/running/cluster-management-service.html.md.erb
```diff
@@ -0,0 +1,141 @@
...
+
+This version of <%=vars.product_name%> uses the [1.11.0 Management REST API](https://cwiki.apache.org/confluence/display/GEODE/1.11.0+Management+REST+API+-+v1).
+As mentioned, the feature is experimental and still undergoing revision.
+
+For a definitive description of the cluster management service for your version of <%=vars.product_name%>, you can query the API directly to describe itself.
+The API returns a JSON body that contains a message describing the result along with standard HTTP status codes:
+
+1. Invoke `gfsh` on your current version of <%=vars.product_name%> and issue the `start locator` to start a locator using the default settings.
+2. Quit `gfsh` to return to the command prompt.
+3. Use `curl` to send a message asking the locator to describe its Cluster Management REST API. The API returns its description in JSON format. Pipe the output through the `jq` JSON formatting utility for readable output.
```
dschneider-pivotal: Instead of telling them to use curl and jq, the easiest way for them to learn the API is to go to: http://localhost:7070/management/swagger-ui.html with a web browser.

dschneider-pivotal: If they use a web browser then they have no need to quit the first gfsh. They can just stay in gfsh until done browsing and then to the shutdown command

geode-docs/configuring/running/cluster-management-service.html.md.erb
```diff
@@ -0,0 +1,75 @@
...
+
+The cluster management service is available as a [REST API](#cms-reference) and as a [Java API](#cms-java-api).
+
+## <a id='cms-reference'>Cluster Management Service REST API</a>
+
+The cluster management service REST API adheres to standard REST semantics, so users can use POST, PATCH, DELETE, and GET to create,
+update, delete, and read, respectively.
+
+This version of <%=vars.product_name%> uses the [1.11.0 Management REST API](https://cwiki.apache.org/confluence/display/GEODE/1.11.0+Management+REST+API+-+v1).
+As mentioned, the feature is experimental and still undergoing revision.
```
onichols-pivotal: This PR is against develop, which will become 1.12, so linking to 1.11 is mildly incorrect.  Even better would be to get the version from <%=vars so you don't have to update this by hand for every release.

geode-docs/configuring/running/cluster-management-service.html.md.erb
davebarnes97: Replaced curl with Swagger

geode-docs/configuring/running/cluster-management-service.html.md.erb
davebarnes97: Used the `product_version` template variable in the docs config to allow the text to track with the software version. Made possible by @onichols-pivotal PR #4394.

PR #4395, opened 2019-12-03
GEODE-7309: upgrade Lucene to 7.1.0

mkevo:

### For all changes:

### Note:


geode-lucene/src/upgradeTest/java/org/apache/geode/cache/lucene/RollingUpgradeQueryReturnsCorrectResultsAfterClientAndServersAreRolledOverAllBucketsCreated.java
```diff
@@ -31,6 +32,7 @@
 public class RollingUpgradeQueryReturnsCorrectResultsAfterClientAndServersAreRolledOverAllBucketsCreated
     extends LuceneSearchWithRollingUpgradeDUnit {
 
+  @Ignore("Not supported")
```
jhuynh1: I don't think we should add an ignore to this test... the test comments explicitly state it's to catch a problem from going from 6 -> 7 (which this PR does)

mkevo: Hi @jhuynh1 ,
As we agreed on dev list we need to **do query after all servers are upgraded**, but this is the case where one server is upgraded, lucene index created on it and then shutdown. Server with older version will get _IndexFormatTooNewException_ as lucene 7.x has different index format and it cannot read it. 
In that case this test is not supported.

jhuynh1: Was there a modified version of this test to show the end result works?  Just to prove that rolling still allows us to query older indexed data at the end of the rolling upgrade?  I thought there was some work/test added to that feature branch?

jhuynh1: Should there also be some doc change and a test added so that makes sure queries are "stuck" if versions differ?  My fear/concern is that a change then opens up this "problem" in the future.

mkevo: I think that we can add to _Apache Lucene Integration_ in _Requirements and Caveats_ that executing queries is possible only when rolling upgrade is finished.
Also we can try to write a test to upgrade one of two server, execute query and except that query will failed as you cannot execute it before all members are rolled.

geode-lucene/src/test/java/org/apache/geode/cache/lucene/internal/distributed/DistributedScoringJUnitTest.java
```diff
@@ -110,15 +109,8 @@ public void uniformDistributionProducesComparableScores() throws Exception {
 
     List<EntryScore<String>> distResult = manager.reduce(collectors).getEntries().getHits();
 
-    Assert.assertEquals(singleResult.size(), distResult.size());
     Assert.assertTrue(singleResult.size() > 0);
-
-    for (Iterator single = distResult.iterator(), dist = singleResult.iterator(); single.hasNext()
```
jhuynh1: The scoring is now not consistent for single node compared to distributed nodes... there is a way to configure the old scoring mechanism to keep them consistent... if we want to move to the new scoring system, we'll have to figure out a way to feed the correct metadata to the score algorithm...

geode-lucene/src/upgradeTest/java/org/apache/geode/cache/lucene/LuceneSearchWithRollingUpgradeTestBase.java
```diff
@@ -292,22 +293,27 @@ void verifyLuceneQueryResults(String regionName, int expectedRegionSize)
...
+  }
 
+  protected boolean isIndexingFinished(Object luceneQuery)
+      throws IllegalAccessException, NoSuchMethodException, InvocationTargetException {
+    boolean success;
+    try {
+      luceneQuery.getClass().getMethod("findKeys").invoke(luceneQuery);
+      success = true;
+    } catch (Exception ex) {
+      ex.printStackTrace();
```
nabarunnag: We can remove the ex.printStackTrace();

geode-lucene/src/main/java/org/apache/geode/cache/lucene/internal/IndexRepositoryFactory.java
```diff
@@ -102,6 +105,19 @@ private IndexRepository finishComputingRepository(Integer bucketId, LuceneSerial
     if (oldRepository != null) {
       oldRepository.cleanup();
     }
+
+    boolean hasOldMember = false;
+    if (userRegion.getCache() != null) {
```
nabarunnag: We can get this as a cache helper method to prevent repeated code. We can make it a part of the cache.
We can implement it how cache.hasPersistenceRegion() is implemented.
boolean cache.hasMemberOlderThan(Version version){<implementation>}

nabarunnag:
Is there a possibility in which we fail the lucene query when it is executed in mixed version servers, like by throwing a particular exception that does not trigger re-execution. I am also working on your branch with this commit and trying to figure out if we can stop the query execution with a failure.

geode-lucene/src/test/java/org/apache/geode/cache/lucene/internal/distributed/DistributedScoringJUnitTest.java
mkevo: Do we need to change `reduce` method to compare on key instead of score? Or it is enough to see that all entries exist in both singleResult and distResult?

jhuynh1: I guess it depends on what we consider acceptable.  I would think we'd want to modify the scoring to be consistent whether we have a 1 node cluster or a multinode cluster.  The classic scoring in lucene seemed to work for us in 6.0 but it looks like things have changed with their scoring.

This might be the relevant page. 
 It looks like we'd have to set the similarity in the index to use the older (I think tdidf one?) http://lucene.apache.org/core/8_0_0/core/org/apache/lucene/search/similarities/Similarity.html

mkevo: In case we want to use similarity(TFIDFSimilarity) we need to implement three methods: _tf, idf_ and _lengthNorm_.
There is a good article about tf-idf and BM25(default in lucene): https://opensourceconnections.com/blog/2015/10/16/bm25-the-next-generation-of-lucene-relevation/

I added a new commit with implementation of these methods.

mkevo: In case we want to use similarity(TFIDFSimilarity) we need to implement three methods: _tf, idf_ and _lengthNorm_.
There is a good article about tf-idf and BM25(default in lucene): https://opensourceconnections.com/blog/2015/10/16/bm25-the-next-generation-of-lucene-relevation/

I added a new commit with implementation of these methods.

jhuynh1: Awesome.  I think TFIDF was the classic scoring.  I think we should move to the newer scoring but I haven't done any research on how to get there.  If you feel ambitious, by all means, go for it :-)

Either way, nice job!

mkevo: Thanks @jhuynh1,
what is a way forward? Is the latest changes OK or we need to change something else?
Any thoughts/suggestions?

jhuynh1: Sorry, I got around to reading the implementation a bit more.  Is the sloppyFreq and scoringPayload scoring correctly?  Is it ok to send back 0 and 1 (I know it's acceptable but is that going to give a real score or are we just going to say everything is the same score now?)

jhuynh1: I think Nabarun had a concern about what type of exception to throw etc.  I'll resolve my conversation if you think the scoring works as intended.   Please check with nabarun about his concerns.  Thanks!

As an aside, i think you originally tried going to lucene 8 and we ran into some issues.  Perhaps those issues are actually resolved now (after doing all the intermediate work to get it this far).

mkevo: sloppyFreq and scoringPayload are deprecated in Lucene 7.x. So in case we change it, it doesn't have any effect on results. 
After this PR is merged I will go to upgrade to lucene 8.
Thanks!

mkevo:
@nabarunnag, I implemented your suggestions, is it ok with you?

nabarunnag:
We are still getting stack overflow errors in our rolling upgrade use cases, let me investigate a bit more.

onichols-pivotal:
> We are still getting stack overflow errors in my rolling upgrade use cases, let me investigate a bit more.

@nabarunnag do you have any update from your further investigation?

mkevo:
> We are still getting stack overflow errors in my rolling upgrade use cases, let me investigate a bit more.

Hi @nabarunnag, can you share your tests with us so we can help you? Also internal tests should not block merging to an open source project.

nabarunnag:
> > We are still getting stack overflow errors in my rolling upgrade use cases, let me investigate a bit more.
> 
> Hi @nabarunnag, can you share your tests with us so we can help you? Also internal tests should not block merging to an open source project.

We are in the process of open-sourcing the tests. More details incoming. 

Simple root cause is: When there is a newer version gemfire with older version, the aysnc listener is blocked waiting on the index repository to be created, but the queues continue to get filled up and ultimately result in overflow. Also, you can see that I have another pull request over your commit [https://github.com/apache/geode/pull/4826] which solves this problem.

nabarunnag:
@mkevo It is nearly complete, but the tests may be exposing another issue with the geode code base (un-related to lucene) which I am trying to fix, but I was unable to get to (other commitments) .. I will try to get it fixed shortly.

onichols-pivotal:
this PR appears to have been abandoned

mkevo:
Hi @nabarunnag,
Do you have any update on this?

mkevo:
Hi @nabarunnag ,
It passed a long period from your last comment. Is it still valid? 
Do you fixed your internal test cases? 
Can I go on with fixing conflicts and merging this PR?

mkevo:
Hi,
If there is no comments until the end of the week, I will go with resolving conflicts and merging this PR.
BR,
Mario

mkevo:
A new PR is opened rebased on the latest develeop branch. #6076

PR #4398, opened 2019-12-03
GEODE-7519 GMSHealthMonitorJUnitTest fails in testCheckIfAvailableNoH…

bschuchardt:
…eartBeatDontRemoveMember

Removing time-sensitive test.  There are other tests that already
exercise the checkIfAvailable health-monitor method.  Two of those tests make
sure that the initiateRemoval parameter (the last parameter) to
checkIfAvailable functions properly and that the value returned by the
method is correct.  Other tests already ensure that a member that
checks out okay is not removed from the cluster and that the value
returned by checkIfAvailable is correct.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/integrationTest/java/org/apache/geode/distributed/internal/membership/gms/fd/GMSHealthMonitorJUnitTest.java
+++ b/geode-core/src/integrationTest/java/org/apache/geode/distributed/internal/membership/gms/fd/GMSHealthMonitorJUnitTest.java
@@ -725,7 +725,48 @@ public class GMSHealthMonitorJUnitTest {
     verify(joinLeave).remove(isA(MemberIdentifier.class), isA(String.class));
   }
 
+  /*
+   * The next two tests are like the previous two except the next two do not set up failure
+   * detection ports.
+   */
 
+  @Test
+  public void testFailedCheckIfAvailableWithoutFailureDetectionPortDoesNotRemoveMember() {
+    useGMSHealthMonitorTestClass = true;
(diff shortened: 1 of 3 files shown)


geode-core/src/integrationTest/java/org/apache/geode/distributed/internal/membership/gms/fd/GMSHealthMonitorJUnitTest.java
```diff
@@ -450,25 +450,6 @@ public void testRemoveMemberCalledAfterDoingFinalCheckOnCoordinator() throws Exc
     Assert.assertTrue(gmsHealthMonitor.getStats().getSuspectsReceived() > 0);
   }
 
-  /***
-   * validates HealthMonitor.CheckIfAvailable api
-   */
-  @Test
-  public void testCheckIfAvailableNoHeartBeatDontRemoveMember() {
```
Bill: Is the point of this test to verify that calling `checkIfAvailable(x,x,true)` under the following conditions, returns `false`:

1. no heartbeats have been received from the member
2. we have not yet reached the first timeout time

Notwithstanding the reasoning in the description of this PR, I don't see this situation covered by the remaining tests.

Sure would be nice to be able to use virtual time in this test.

bschuchardt: The test testFailedCheckIfAvailableRemovesMember verifies that checkIfAvailable(x,x,true) returns false and sets the same pre-conditions that this test used.

PR #4406, opened 2019-12-03
Add LGTM Badges

moleske:
Makes it easier to get to alerts for people to then contribute to Geode


moleske:
Well we as a community we decided that LGTM was an integration worth having.  So if we don't like the alerts, I don't think we should be integrating with LGTM.

And yes I saw what you did there 😉

PR #4409, opened 2019-12-04
Geode 6374 - Fixing DistributedNoAckRegionCCEOffHeapDUnitTest.testNoLoaderWithInvalidEntry

mhansonp:


geode-dunit/src/main/java/org/apache/geode/test/dunit/cache/internal/JUnit4CacheTestCase.java
```diff
@@ -377,103 +377,99 @@ private static synchronized void remoteTearDown() {
...
+      throws CacheException {
 
-  public final <K, V> Region<K, V> createRootRegion(final String rootName,
-      final RegionFactory<K, V> regionFactory) throws CacheException {
     Region<K, V> root = getRootRegion(rootName);
     if (root == null) {
-      RegionFactory<K, V> regionFactoryRoot = getRootRegionRegionFactory(regionFactory);
-      regionFactory.setPartitionAttributes(null);
+      RegionFactory<K, V> regionFactoryRoot = getRootRegionFactory(regionFactory);
+      regionFactoryRoot.setPartitionAttributes(null);
```
mhansonp: The bug.

mhansonp:
The "org.apache.geode.internal.cache.PartitionedRegionSizeDUnitTest > testBug39868 FAILED"

is the basis for GEODE-7510 which has a PR on it already.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/partitioned/PersistentPartitionedRegionDistributedTest.java
```diff
@@ -1620,7 +1608,7 @@ public void recoveryFinished(Region region) {
 
     regionFactory.create(partitionedRegionName);
 
-    recoveryDone.await(2, MINUTES);
+    recoveryDone.await();
```
kirklund: CountDownLatch await() is infinite by default. You might want to change to use GeodeAwaitility timeout:
```
private static final long TIMEOUT_MILLIS = GeodeAwaitility.getTimeout().getValueInMS();

recoverDone.await(TIMEOUT_MILLIS, TimeUnit.MILLISECONDS);
```

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/partitioned/PersistentPartitionedRegionDistributedTest.java
```diff
@@ -1785,7 +1773,7 @@ public void beforeSendMessage(final ClusterDistributionManager dm,
         // make sure this is a bucket region doing a GII
         if (requestImageMessage.regionPath.contains("B_")) {
           try {
-            latch.await(2, MINUTES);
+            latch.await();
```
kirklund:
```
private static final long TIMEOUT_MILLIS = GeodeAwaitility.getTimeout().getValueInMS();

recoverDone.await(TIMEOUT_MILLIS, TimeUnit.MILLISECONDS);
```

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/partitioned/PersistentPartitionedRegionDistributedTest.java
```diff
@@ -1467,22 +1458,22 @@ public void recoveryFinished(final Region region) {
...
-    RegionFactory<?, ?> regionFactoryPR =
+    RegionFactory<Object, Object> regionFactoryPR =
         getCache().createRegionFactory(PARTITION_PERSISTENT);
     regionFactoryPR.setPartitionAttributes(partitionAttributesFactory.create());
 
     regionFactoryPR.createSubregion(parentRegion1, partitionedRegionName);
     regionFactoryPR.createSubregion(parentRegion2, partitionedRegionName);
 
-    recoveryDone.await(2, MINUTES);
+    recoveryDone.await();
```
kirklund: Infinite CountDownLatch await().

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/partitioned/PersistentPartitionedRegionDistributedTest.java
```diff
@@ -1066,9 +1061,9 @@ public void regionIsUsableBeforeRedundancyRecovery() throws Exception {
       AsyncInvocation<Void> createPartitionedRegionOnVM2 =
           vm2.invokeAsync(() -> createPartitionedRegionWithPersistence(redundancy));
 
-      createPartitionedRegionOnVM0.await(2, MINUTES);
-      createPartitionedRegionOnVM1.await(2, MINUTES);
-      createPartitionedRegionOnVM2.await(2, MINUTES);
+      createPartitionedRegionOnVM0.await();
```
kirklund: These are perfect. AsyncInvocation await() uses GeodeAwaitility timeout.

mhansonp:
This PR turned into a refactor of a bunch of files due to a number of periodic failures throughout.

This PR addresss GEODE-6374, GEODE-6326, GEODE-7472, GEODE-7512, and GEODE-7514.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/partitioned/PersistentPartitionedRegionDistributedTest.java
mhansonp: Done

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/partitioned/PersistentPartitionedRegionDistributedTest.java
mhansonp: Done

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/partitioned/PersistentPartitionedRegionDistributedTest.java
mhansonp: Done

geode-core/src/main/java/org/apache/geode/cache/Cache.java
```diff
@@ -144,6 +144,10 @@
    */
   <K, V> RegionFactory<K, V> createRegionFactory(RegionAttributes<K, V> regionAttributes);
 
+
+
```
mhansonp: Found a need to add a new API so that I could pass and modify a regionFactory.

geode-core/src/main/java/org/apache/geode/cache/RegionFactory.java
```diff
@@ -100,6 +100,16 @@ protected RegionFactory(InternalCache cache, RegionAttributes ra) {
     this.attrsFactory = new AttributesFactory<K, V>(ra);
   }
 
+  /**
+   * For internal use only.
+   *
+   * @since GemFire 9.10
+   */
```
mhansonp: Found a need to add a new API so that I could pass and modify a regionFactory.

geode-core/src/main/java/org/apache/geode/cache/RegionFactory.java
```diff
@@ -161,6 +171,10 @@ public RegionFactory(String regionAttributesId)
     this((InternalCache) new CacheFactory().create(), regionAttributesId);
   }
 
+  public RegionFactory(RegionFactory<K, V> regionFactory) {
```
mhansonp: Found a need to add a new API so that I could pass and modify a regionFactory.

geode-core/src/main/java/org/apache/geode/cache/RegionFactory.java
```diff
@@ -769,8 +784,7 @@ public void setPublisher(boolean v) {}
    * @throws CacheClosedException if the cache is closed
    * @since GemFire 7.0
    */
-  @SuppressWarnings("unchecked")
-  public Region<K, V> createSubregion(Region<?, ?> parent, String name)
+  public Region<K, V> createSubregion(Region<K, V> parent, String name)
```
mhansonp: Fixed typing warning.

geode-core/src/main/java/org/apache/geode/internal/cache/GemFireCacheImpl.java
```diff
@@ -4715,6 +4715,12 @@ public TXManagerImpl getTxManager() {
     return new RegionFactoryImpl<>(this, regionAttributes);
   }
 
+  @Override
+  public <K, V> RegionFactory<K, V> createRegionFactory(RegionFactory<K, V> regionFactory) {
+    throwIfClient();
```
mhansonp: Found a need to add a new API so that I could pass and modify a regionFactory.

geode-core/src/main/java/org/apache/geode/cache/RegionFactory.java
metatype: If it's on the public API, I don't think it should be marked for internal use.

geode-core/src/main/java/org/apache/geode/cache/RegionFactory.java
metatype: Is this a breaking API change?

mhansonp: Not that I can see. All relevant testing still passes. It was getting a warning that the type was not correct, hence the reason I changed it to clear the unchecked.

jhuynh1: I think this forces the parent and child to have the same K,V types?
 Cache cache = getCache();
      RegionFactory<String, String> regionFactory = cache.createRegionFactory(RegionShortcut.REPLICATE);
      Region<String, Integer> parentRegion = cache.getRegion(PARENT_REGION_NAME);
      regionFactory.createSubregion(parentRegion, CHILD_REGION_NAME);

I believe that would have worked before but now it will show as an issue since the parent and child regions have different types (which actually is allowed in geode)

jhuynh1: Does this force the parent and child to have the same K,V types?  If so, I don't think that is a valid requirement for Geode...

geode-core/src/main/java/org/apache/geode/cache/RegionFactory.java
```diff
@@ -100,6 +100,16 @@ protected RegionFactory(InternalCache cache, RegionAttributes ra) {
     this.attrsFactory = new AttributesFactory<K, V>(ra);
   }
 
+  /**
+   * For internal use only.
+   *
+   * @since GemFire 9.10
```
jhuynh1: The since should be a geode release not a gemfire release

mhansonp: I modified that in the latest change set.

mhansonp:
Sorry I rebased due to a conflict, so a force push was required.

geode-core/src/main/java/org/apache/geode/cache/RegionFactory.java
mhansonp: Good point...

geode-core/src/main/java/org/apache/geode/internal/cache/GemFireCacheImpl.java
```diff
@@ -4366,6 +4366,15 @@ public TXManagerImpl getTxManager() {
...
+    return new RegionFactoryImpl<>(regionFactory);
+  }
+
+  /**
+   * @since GemFire 6.5
```
jhuynh1: Was this intentional?  Is there a reason why we would put this on the class but not interface?

mhansonp: What do you mean? Which API? The above API  for createRegionFactory(RegionFactory<K,V> regionFactory) is in Cache.java

jhuynh1: I mean, In two parts:

1.) Do we want to annotate this in GemFireCacheImpl (when the other similar methods in GemFireCacheImpl aren't annotated?)  

2.) Shouldn't this be annotated in the Interface instead of the implementing class?

mhansonp: I think I have made this discussion moot by removing this functionality and creating a utility class which accesses the now protected RegionFactory method. This means I have removed this API that I added.

PR #4410, opened 2019-12-04
GEODE-7503: GemFireCacheImpl close will block until close completes

kirklund:
Subsequent calls to GemFireCacheImpl close will now block until the
first call to close completes.

Additional changes:
* Expand unit testing of GemFireCacheImpl
* Inject dependencies into GemFireCacheImpl constructor
* Cleanup existing tests of GemFireCacheImpl

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/GemFireCacheImpl.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/GemFireCacheImpl.java
@@ -623,6 +623,8 @@ public class GemFireCacheImpl implements InternalCache, InternalClientCache, Has
 
   private volatile boolean isClosing;
 
+  private final CountDownLatch isClosedLatch = new CountDownLatch(1);
+
   /**
    * Set of all gateway senders. It may be fetched safely (for enumeration), but updates must by
    * synchronized via {@link #allGatewaySendersLock}
@@ -1785,7 +1787,7 @@ public class GemFireCacheImpl implements InternalCache, InternalClientCache, Has
           }
         }
 
-        close("Shut down all members", null, false, true);
+        close("Shut down all members", null, false, true, false);
       } finally {
         shutDownAllFinished.countDown();
       }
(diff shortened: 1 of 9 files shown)


geode-core/src/main/java/org/apache/geode/internal/cache/GemFireCacheImpl.java
```diff
@@ -375,6 +379,8 @@
 
   volatile boolean isClosing = false; // used in Stopper inner class
 
+  private final CountDownLatch isClosedLatch = new CountDownLatch(1);
```
agingade: What happens when the cache is in reconnecting state. It did not looked like the latch is getting reset.

geode-core/src/main/java/org/apache/geode/internal/cache/GemFireCacheImpl.java
```diff
@@ -2070,11 +2163,12 @@ public void close(String reason, Throwable systemFailureCause, boolean keepAlive
     final boolean isDebugEnabled = logger.isDebugEnabled();
 
     synchronized (GemFireCacheImpl.class) {
-      // fix for bug 36512 "GemFireCache.close is not thread safe"
-      // ALL CODE FOR CLOSE SHOULD NOW BE UNDER STATIC SYNCHRONIZATION
-      // OF synchronized (GemFireCache.class) {
+      // ALL CODE FOR CLOSE SHOULD NOW BE UNDER STATIC SYNCHRONIZATION OF GemFireCacheImpl.class
       // static synchronization is necessary due to static resources
       if (isClosed()) {
+        if (!skipAwait) {
```
agingade: This changes the existing behavior...Where subsequent close calls are returned immediately...I guess its ok; as the first close can be any thread depending on the scheduling.

kirklund:
Closing PR.

kirklund:
I closed this PR after Anil and I discussed it further. The main concern is that the current behavior is expected by users and is de facto correct. Changing it would change the behavior and result in long-time users being surprised by the new behavior.

PR #4416, opened 2019-12-04
GEODE-7351: OQL Method Authorizer Constraints

jujoramos:
Added distributed tests to verify the required constraints are met
whenever the MethodInvocationAuthorizer is changed in runtime.

- Once the authorizer is changed, all queries executed afterwards
  should use the newly configured authorizer.
- Once a query execution starts, the authorizer used can not be
  changed for that particular query.
- Continuous queries already running should pick up the new authorizer
  the next time the query is internally executed to detect whether an
  event matches or not. If the CQ has methods not allowed by the newly
  configured authorizer, any matching events from that moment on should
  invoke 'onError' instead of 'onEvent' on the associated 'CqListener'.
- Indexes should pick up the newly configured authorizer the next time
  an entry is added or removed from the index, and the index itself
  should be marked as invalid if it uses method invocations not allowed
  by the newly configured authorizer.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/distributedTest/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceConstraintsDistributedTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceConstraintsDistributedTest.java
@@ -77,6 +77,9 @@ public class QueryConfigurationServiceConstraintsDistributedTest implements Seri
   private static final int UPDATE_KEY = ENTRIES - 3;
   private static final int REPLACE_KEY = ENTRIES - 4;
   private static final int INVALIDATE_KEY = ENTRIES - 5;
+  private static final String ID_INDEX_IDENTIFIER = "IdIndex";
+  private static final String NAME_INDEX_IDENTIFIER = "NameIndex";
+  private static final QueryObject TEST_VALUE = new QueryObject(999, "name_999");
   private File logFile;
   protected MemberVM server;
   protected ClientVM client;
@@ -130,11 +133,12 @@ public class QueryConfigurationServiceConstraintsDistributedTest implements Seri
         .withProperty("log-file", logFile.getAbsolutePath()));
 
     server.invoke(() -> {
-      assertThat(ClusterStartupRule.getCache()).isNotNull();
       InternalCache internalCache = ClusterStartupRule.getCache();
+      assertThat(internalCache).isNotNull();
       internalCache.getService(QueryConfigurationService.class).updateMethodAuthorizer(
           internalCache, false, TestMethodAuthorizer.class.getName(),
-          Stream.of("getId", "getName").collect(Collectors.toSet()));
+          Stream.of(QueryObject.GET_ID_METHOD, QueryObject.GET_NAME_METHOD)
+              .collect(Collectors.toSet()));
     });
 
     client = cluster.startClientVM(2, ccf -> ccf
@@ -150,8 +154,8 @@ public class QueryConfigurationServiceConstraintsDistributedTest implements Seri
 
   private void createAndPopulateRegion(String regionName, RegionShortcut shortcut) {
     server.invoke(() -> {
-      assertThat(ClusterStartupRule.getCache()).isNotNull();
       InternalCache internalCache = ClusterStartupRule.getCache();
+      assertThat(internalCache).isNotNull();
       Region<Integer, QueryObject> region =
           internalCache.<Integer, QueryObject>createRegionFactory(shortcut).create(regionName);
       IntStream.range(0, ENTRIES).forEach(id -> region.put(id, new QueryObject(id, "name_" + id)));
@@ -160,20 +164,20 @@ public class QueryConfigurationServiceConstraintsDistributedTest implements Seri
   }
 
   private ThrowableAssert.ThrowingCallable getRegionOperation(Operation operation,
-      Region<Integer, QueryObject> region, QueryObject testValue) {
+      Region<Integer, QueryObject> region) {
     switch (operation) {
       case PUT:
-        return () -> region.put(PUT_KEY, testValue);
+        return () -> region.put(PUT_KEY, TEST_VALUE);
       case CREATE:
-        return () -> region.create(CREATE_KEY, testValue);
+        return () -> region.create(CREATE_KEY, TEST_VALUE);
       case REMOVE:
         return () -> region.remove(REMOVE_KEY);
       case DESTROY:
         return () -> region.destroy(DESTROY_KEY);
       case UPDATE:
-        return () -> region.put(UPDATE_KEY, testValue);
+        return () -> region.put(UPDATE_KEY, TEST_VALUE);
       case REPLACE:
-        return () -> region.replace(REPLACE_KEY, testValue);
+        return () -> region.replace(REPLACE_KEY, TEST_VALUE);
       case INVALIDATE:
         return () -> region.invalidate(INVALIDATE_KEY);
       default:
@@ -182,15 +186,15 @@ public class QueryConfigurationServiceConstraintsDistributedTest implements Seri
     }
   }
 
-  private void assertRegionOperationResult(Operation operation, Region<Integer, QueryObject> region,
-      QueryObject testValue) {
+  private void assertRegionOperationResult(Operation operation,
+      Region<Integer, QueryObject> region) {
     // Assert operation result.
     switch (operation) {
       case PUT:
-        assertThat(region.get(PUT_KEY)).isEqualTo(testValue);
+        assertThat(region.get(PUT_KEY)).isEqualTo(TEST_VALUE);
         break;
       case CREATE:
-        assertThat(region.get(CREATE_KEY)).isEqualTo(testValue);
+        assertThat(region.get(CREATE_KEY)).isEqualTo(TEST_VALUE);
         break;
       case REMOVE:
         assertThat(region.get(REMOVE_KEY)).isNull();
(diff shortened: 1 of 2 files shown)


geode-core/src/distributedTest/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceConstraintsDistributedTest.java
```diff
@@ -0,0 +1,595 @@
...
+
+    server = cluster.startServerVM(1, cf -> cf
+        .withProperty(SECURITY_MANAGER, SimpleSecurityManager.class.getName())
+        .withProperty(SERIALIZABLE_OBJECT_FILTER, "org.apache.geode.cache.query.internal.*")
+        .withProperty("security-username", "cluster").withProperty("security-password", "cluster")
+        .withProperty("log-file", logFile.getAbsolutePath()));
+
+    server.invoke(() -> {
+      assertThat(ClusterStartupRule.getCache()).isNotNull();
+      InternalCache internalCache = ClusterStartupRule.getCache();
```
DonalEvans: Instead of two method calls here, it might be more efficient to assign the `internalCache` variable first and then assert that the variable is not null. This applies to almost every time `ClusterStartupRule` is used to retrieve a cache in this file and the other file in this PR.

geode-core/src/distributedTest/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceConstraintsDistributedTest.java
```diff
@@ -0,0 +1,595 @@
...
+          .isInstanceOf(ServerOperationException.class)
+          .hasCauseInstanceOf(NotAuthorizedException.class)
+          .hasStackTraceContaining(RestrictedMethodAuthorizer.UNAUTHORIZED_STRING + "getName");
+    });
+  }
+
+  /**
+   * The test registers a {@link QueryObserver} that changes the installed
+   * {@link MethodInvocationAuthorizer} to the default one (denies everything) as soon as the query
+   * starts, and makes sure the query succeeds (it is no affected by the authorizer change).
```
DonalEvans: Typo here, should be "it is not affected by".

geode-core/src/distributedTest/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceConstraintsDistributedTest.java
```diff
@@ -0,0 +1,595 @@
...
+    // Execute query again, it should fail as the restricted authorizer has been installed.
+    server.invoke(QueryObserverHolder::reset);
+    client.invoke(() -> {
+      assertThat(ClusterStartupRule.getClientCache()).isNotNull();
+      ClientCache clientCache = ClusterStartupRule.getClientCache();
+      Query newQuery = clientCache.getQueryService().newQuery(queryString);
+      assertThatThrownBy(newQuery::execute)
+          .isInstanceOf(ServerOperationException.class)
+          .hasCauseInstanceOf(NotAuthorizedException.class)
+          .hasStackTraceContaining(RestrictedMethodAuthorizer.UNAUTHORIZED_STRING + "getName");
```
DonalEvans: Here (and in other tests) could the "getName" (or "getId" if that's the method used) string be extracted to a variable so that we can avoid having the string hardcoded in two different places?

geode-core/src/distributedTest/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceConstraintsDistributedTest.java
```diff
@@ -0,0 +1,595 @@
...
+          .isInstanceOf(NotAuthorizedException.class)
+          .hasMessage(RestrictedMethodAuthorizer.UNAUTHORIZED_STRING + "getName");
+    });
+  }
+
+  /**
+   * The test creates an index with a method invocation as part of the expression, changes
+   * the {@link MethodInvocationAuthorizer} to a custom one that still allows the methods part of
+   * the index expression and executes a region operation from the client that would cause an
+   * index mapping removal/addition.
```
DonalEvans: Could this be reworded to be a little clearer? Maybe something like "changes the {@link MethodInvocationAuthorizer} to a custom one that allows the method that is invoked by the index and executes"

geode-core/src/distributedTest/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceConstraintsDistributedTest.java
```diff
@@ -0,0 +1,595 @@
...
+    QueryObject testValue = new QueryObject(999, "name_999");
+
+    server.invoke(() -> {
+      assertThat(ClusterStartupRule.getCache()).isNotNull();
+      InternalCache internalCache = ClusterStartupRule.getCache();
+
+      // Index is valid.
+      QueryService queryService = internalCache.getQueryService();
+      queryService.createIndex("NameIndex", "e.getName()", "/" + regionName + " e");
+      Index index = queryService.getIndex(internalCache.getRegion(regionName), "NameIndex");
```
DonalEvans: Here and in other tests, could we extract the index name "NameIndex" (or "IdIndex" in tests where that's used) to a variable, just to avoid having it hardcoded in two different places?

geode-core/src/distributedTest/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceConstraintsDistributedTest.java
```diff
@@ -0,0 +1,595 @@
...
+   * The operation should succeed, the index should still be valid and no errors should be logged.
+   */
+  @Test
+  @Parameters(method = "getRegionTypeAndOperation")
+  @TestCaseName("[{index}] {method}(RegionType:{0};Operation:{1})")
+  public void indexesShouldNotBeAffectedByMethodAuthorizerChangeAfterRegionOperationOnClientWhenIndexedExpressionContainsMethodsAllowedByTheNewAuthorizer(
+      RegionShortcut regionShortcut, Operation operation) {
+    String regionName = testName.getMethodName();
+    createAndPopulateRegion(regionName, regionShortcut);
+    QueryObject testValue = new QueryObject(999, "name_999");
```
DonalEvans: Would it be possible to have this `QueryObject` be a class field? The values passed to its constructor seem to be arbitrary so it should be fine for them to be the same in every test case, and it would avoid having to create a new one for every test case.

geode-core/src/distributedTest/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceConstraintsDistributedTest.java
```diff
@@ -0,0 +1,595 @@
...
+      ThrowableAssert.ThrowingCallable operationCallable =
+          getRegionOperation(operation, region, testValue);
+      assertThatCode(operationCallable).doesNotThrowAnyException();
+    });
+  }
+
+  /**
+   * The test registers a {@link QueryObserver} that changes the installed
+   * {@link MethodInvocationAuthorizer} to the default one (denies everything) as soon as the query
+   * starts, and makes sure the query succeeds (it is no affected by the authorizer change).
```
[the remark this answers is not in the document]
jujoramos: Done!.

geode-core/src/distributedTest/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceConstraintsDistributedTest.java
jujoramos: I didn't pay much attention when originally doing this as we're talking about a test class here, not sure how much more "efficient" avoiding a method call would be, and how important it actually is for the test, but I made the requested changed anyway 👍

geode-core/src/distributedTest/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceConstraintsDistributedTest.java
jujoramos: Done!

geode-core/src/distributedTest/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceConstraintsDistributedTest.java
jujoramos: Done!.

geode-core/src/distributedTest/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceConstraintsDistributedTest.java
jujoramos: Done!.

geode-core/src/distributedTest/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceConstraintsDistributedTest.java
jujoramos: Done!.

geode-core/src/distributedTest/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceConstraintsDistributedTest.java
jujoramos: Done!.

PR #4423, opened 2019-12-04
GEODE-7540: use max_in_flight to spread out peak load on CI

onichols-pivotal:
Instead of immediately launching all jobs at once, this PR evens out the load by setting lower max_in_flight settings for shorter jobs. I estimate this will cut peak load in half without builds taking any longer (for a scenario where 6 commits are in the pipeline at the same time).
Reducing load on concourse is desirable because we tend to have more unexplained rsync failures, etc, when concourse is under heavy load.

Also dial down unnecessarily-high retry counts from `10` to `2`, which rarely helped but greatly increased number of containers when things were already overloaded.

Note: This PR does NOT throttle PR pipeline.

### DIFF
--- a/ci/pipelines/examples/jinja.template.yml
+++ b/ci/pipelines/examples/jinja.template.yml
@@ -113,7 +113,7 @@ jobs:
           - name: attempts-log
             path: new
         timeout: 15m
-        attempts: 10
+        attempts: 2
       - task: rsync_code_up
         image: alpine-tools-image
         config:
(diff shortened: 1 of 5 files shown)


rhoughton-pivot:
My review was on an older change-set... I would have -1 this, for the `retry<=2` for create_instance jobs. The roulette nature of the work means that we can fail twice 1/4 times when a failure happens the first time.

onichols-pivotal:
I have no problem increasing it if we see a case where more retries could arguably make a difference.  Unfortunately with 10 we only saw the cases where retries didn't help.  Other pipelines I've looked at tend to stick with 2-3 retries.  Any issue that persists after 2-3 tries is probably worth bubbling up to our attention.

PR #4429, opened 2019-12-05
GEODE-7550: Param to only list gw receivers or senders

alb3rtobr:
This commit adds the options "--senders" & "--receivers" to the "list gateways" gfsh command, to show only senders or receivers respectively.
If both options are introduced, both senders and receivers are shown.
Tests were added verifiying the output of the command with each parameter, and with both parameters at the same time.


davebarnes97:
Added a 'docs' component to the JIRA ticket.

geode-core/src/main/java/org/apache/geode/management/internal/cli/commands/ListGatewayCommand.java
```diff
@@ -53,9 +53,21 @@ public ResultModel listGateway(
...
+      @CliOption(key = {CliStrings.LIST_GATEWAY__RECEIVERS_ONLY},
+          specifiedDefaultValue = "false", unspecifiedDefaultValue = "true",
+          help = CliStrings.LIST_GATEWAY__RECEIVERS_ONLY) boolean getSenders,
+      @CliOption(key = {CliStrings.LIST_GATEWAY__SENDERS_ONLY},
+          specifiedDefaultValue = "false", unspecifiedDefaultValue = "true",
+          help = CliStrings.LIST_GATEWAY__SENDERS_ONLY__HELP) boolean getReceivers)
+
       throws Exception {
 
+    if (!getSenders && !getReceivers) {
```
agingade: Looks like this is going to return both senders and receivers list even when user has explicitly set the values to false. Either we should throw error saying one needs to be true or return empty set.

Also, for the following case, due to specified default value "false"; it will list the receivers, wherein the expectation may be only senders.

> list gateways --senders

jujoramos: +1 to @agingade's comments above.

alb3rtobr: Notice that option `--senders` set the value for `getReceivers` variable, and `--receivers` set the value for `getSenders`. 

That is:
* When `list gateways` then `getReceivers=true` & `getSenders=true` 
* When `list gateways --senders` then `getReceivers=false` & `getSenders=true`
* When `list gateways --receivers` then `getReceivers=true` & `getSenders=false`
* When `list gateways --senders --receivers` then `getReceivers=false` & `getSenders=false`, so I set them to true, to get both senders & receivers, because for me its more intuitive. I would expect to get both senders and receivers if I introduce both parameters.

Thanks to your comment @agingade I realize there is a missing assert in the `testListGatewaySenderOnly` test, because Im not checking if the output has or not the gateway receiver section, as it is done in `testListGatewayReceiverOnly`.

agingade: I did not notice the inverted use of attributes...
For "sender" argument it is setting "getReceivers"
Can we change it to "noSender" that way we know its associated with the sender....

For "list gateways --senders --receivers"
Agree with you; it should list out both.

For "list gateways --senders=false --receivers=false"
Its not expected, but one could execute it; in that case, we just need to ignore the command...And list empty set. Or a message saying either one of them is required.

agingade: @alb3rtobr please let me know if you need any help.

alb3rtobr: @agingade Thanks, I have not been able to spend more time on this until today.

alb3rtobr:
Finally this is the only way I found to implement a consistent behavior:

The problem is that if the output of `list gateways --senders --receivers` is the same than `list gateways`, that means that `specifiedDefaultValue` and `unspecifiedDefaultValue` have to be `true` for both parameters. But in that case, when executing `list gateways --senders` or `list gateways --receivers` both parameters will be `true`.

So with the last commit:
* `list gateways --senders --receivers` and `list gateways --senders=true --receivers=true` will return error
* `list gateways --senders=false --receivers=false` and any combination using "one true and one false" will be like running `list gateways`.
* `list gateways --senders` and `list gateways --senders=true` will show only senders.
* `list gateways --receivers` and `list gateways --receivers=true` will show only receivers.

onichols-pivotal:
@alb3rtobr, to tickle the PR checks that never ran, please push an empty commit, thanks!

alb3rtobr:
> @alb3rtobr, to tickle the PR checks that never ran, please push an empty commit, thanks!

thanks for the heads up @onichols-pivotal , I had not realized the CI was stuck

jhuynh1:
The behavior as described:
`
list gateways --senders=false --receivers=false and any combination using "one true and one false" will be like running list gateways.`

sounds odd to me... if one is true and the other false, i would expect it to run similar to list gateways --sender=true or --receivers=true depending on which is true...

geode-core/src/main/java/org/apache/geode/management/internal/i18n/CliStrings.java
```diff
@@ -1653,6 +1653,14 @@
   public static final String LIST_FUNCTION__NO_FUNCTIONS_FOUND_ERROR_MESSAGE = "No Functions Found";
 
   public static final String LIST_GATEWAY = "list gateways";
+  public static final String LIST_GATEWAY__SHOW_SENDERS_ONLY = "senders";
```
jujoramos: It would be more intuitive to our users for this parameter to be called `senders-only` instead of just `sender`.

geode-core/src/main/java/org/apache/geode/management/internal/i18n/CliStrings.java
```diff
@@ -1653,6 +1653,14 @@
...
+  public static final String LIST_GATEWAY__SHOW_SENDERS_ONLY__HELP =
+      "Display gateway senders only.";
+  public static final String LIST_GATEWAY__SHOW_RECEIVERS_ONLY = "receivers";
```
jujoramos: It would be more intuitive to our users for this parameter to be called `receivers-only` instead of just `receivers`.

alb3rtobr:
> The behavior as described:
> ` list gateways --senders=false --receivers=false and any combination using "one true and one false" will be like running list gateways.`
> 
> sounds odd to me... if one is true and the other false, i would expect it to run similar to list gateways --sender=true or --receivers=true depending on which is true...

I think that with the change suggested by @jujoramos it will have more sense to you. The parameters `--senders` and `--receivers` are used to indicate "show only senders" and "show only receivers" respectively. So better if I rename them :)

geode-core/src/main/java/org/apache/geode/management/internal/i18n/CliStrings.java
```diff
@@ -1653,6 +1653,14 @@
...
 
   public static final String LIST_GATEWAY = "list gateways";
+  public static final String LIST_GATEWAY__SHOW_SENDERS_ONLY = "senders-only";
+  public static final String LIST_GATEWAY__SHOW_SENDERS_ONLY__HELP =
+      "Display gateway senders only.";
+  public static final String LIST_GATEWAY__SHOW_RECEIVERS_ONLY = "receivers-only";
+  public static final String LIST_GATEWAY__SHOW_RECEIVERS_ONLY__HELP =
+      "Display gateway receivers only.";
+  public static final String LIST_GATEWAY__ERROR_ON_SHOW_PARAMETERS =
+      "Options --senders and --receivers cannot be used together.";
```
jujoramos: In order to keep the messages consistent with the parameter names, this should read `Options --senders-only and --receivers-only cannot be used together.`

alb3rtobr:
I realized that the parameters can have three states ( false, true and empty ) and I was trying to use a boolean to store that state. I fixed that, and I have simplified the use cases:

* `list gateways --senders-only` and `list gateways --senders=true` will show only senders.
* `list gateways --receivers-only` and `list gateways --receivers-only=true` will show only receivers.
* Any attempt to use `--senders-only` and `--receivers-only` together will return error, no matter the value introduced.

I like this solution instead of the previous one, I think its easier to understand how the parameters can be used.

geode-core/src/main/java/org/apache/geode/management/internal/i18n/CliStrings.java
```diff
@@ -1653,6 +1653,23 @@
...
+      "Display gateway senders only.";
+  public static final String LIST_GATEWAY__SHOW_RECEIVERS_ONLY = "receivers-only";
+  public static final String LIST_GATEWAY__SHOW_RECEIVERS_ONLY_INPUT_ERROR =
+      "receivers-only must be 'true' or 'false'";
+  public static final String LIST_GATEWAY__SHOW_RECEIVERS_ONLY__HELP =
+      "Display gateway receivers only.";
+
+  public static final String LIST_GATEWAY__SHOW_PARAM_TRUE = "true";
+  public static final String LIST_GATEWAY__SHOW_PARAM_FALSE = "false";
+  public static final String LIST_GATEWAY__SHOW_PARAM_EMPTY = "empty";
```
jujoramos: I don't think these new constants are needed and, actually, I believe it makes things harder to understand than before... (see my other comments).

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/ListGatewayCommand.java
```diff
@@ -53,9 +53,51 @@ public ResultModel listGateway(
...
-          help = CliStrings.LIST_GATEWAY__GROUP__HELP) String[] onGroup)
+          help = CliStrings.LIST_GATEWAY__GROUP__HELP) String[] onGroup,
+      @CliOption(key = {CliStrings.LIST_GATEWAY__SHOW_RECEIVERS_ONLY},
+          specifiedDefaultValue = CliStrings.LIST_GATEWAY__SHOW_PARAM_TRUE,
+          unspecifiedDefaultValue = CliStrings.LIST_GATEWAY__SHOW_PARAM_EMPTY,
+          help = CliStrings.LIST_GATEWAY__SHOW_RECEIVERS_ONLY) String showReceiversOnly,
+      @CliOption(key = {CliStrings.LIST_GATEWAY__SHOW_SENDERS_ONLY},
+          specifiedDefaultValue = CliStrings.LIST_GATEWAY__SHOW_PARAM_TRUE,
+          unspecifiedDefaultValue = CliStrings.LIST_GATEWAY__SHOW_PARAM_EMPTY,
+          help = CliStrings.LIST_GATEWAY__SHOW_SENDERS_ONLY__HELP) String showSendersOnly)
```
jujoramos: Both parameters should be of type `boolean`, and you can achieve the same outcome directly through `spring-shell` features:

```
@CliOption(key = CliStrings.LIST_GATEWAY__SHOW_SENDERS_ONLY, specifiedDefaultValue = "true", unspecifiedDefaultValue = "false", help = CliStrings.LIST_GATEWAY__SHOW_SENDERS_ONLY) boolean showOnlySenders

@CliOption(key = CliStrings.LIST_GATEWAY__SHOW_RECEIVERS_ONLY, specifiedDefaultValue = "true", unspecifiedDefaultValue = "false", help = CliStrings.LIST_GATEWAY__SHOW_RECEIVERS_ONLY) boolean showOnlyReceivers
```

Spring shell will automatically populate the correct value for the `boolean` flags when parsing the user input, as below:

| Command  | showOnlyReceivers | showOnlySenders| 
| ------------- | ------------- | ------------- |
| list gateways  | false  | false |
| list gateways --senders-only  | false | true |
| list gateways --senders-only=true  | false  | true |
| list gateways --senders-only=false  | false  | false |
| list gateways --receivers-only  | true  | false |
| list gateways --receivers-only=true  | true  | false |
| list gateways --receivers-only=false  | false  | false |

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/ListGatewayCommand.java
```diff
@@ -53,9 +53,51 @@ public ResultModel listGateway(
...
+        && !showReceiversOnly.equals(CliStrings.LIST_GATEWAY__SHOW_PARAM_FALSE)) {
+      return ResultModel.createError(CliStrings.LIST_GATEWAY__SHOW_RECEIVERS_ONLY_INPUT_ERROR);
+    }
+
+    // Check value of --senders-only
+    if (!showSendersOnly.equals(CliStrings.LIST_GATEWAY__SHOW_PARAM_EMPTY)
+        && !showSendersOnly.equals(CliStrings.LIST_GATEWAY__SHOW_PARAM_TRUE)
+        && !showSendersOnly.equals(CliStrings.LIST_GATEWAY__SHOW_PARAM_FALSE)) {
+      return ResultModel.createError(CliStrings.LIST_GATEWAY__SHOW_SENDERS_ONLY_INPUT_ERROR);
+    }
```
jujoramos: You can delete all these validations once the flags have been changed to `boolean` instead of `String`.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/ListGatewayCommand.java
```diff
@@ -53,9 +53,51 @@ public ResultModel listGateway(
...
+
+    // Check --senders-only & --receivers-only have not been used together
+    if (!showReceiversOnly.equals(CliStrings.LIST_GATEWAY__SHOW_PARAM_EMPTY)
+        && !showSendersOnly.equals(CliStrings.LIST_GATEWAY__SHOW_PARAM_EMPTY)) {
+      return ResultModel.createError(CliStrings.LIST_GATEWAY__ERROR_ON_SHOW_PARAMETERS);
+    }
```
jujoramos: Can you move this logic to an interceptor -> `@CliMetaData(interceptor = "interceptor.class.name")`?, it's the standard way of doing this kind of input validation for GFSH commands. You can have a look at `AlterAsyncEventQueueCommand` and `CreateRegionCommand` for examples about its usage.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/ListGatewayCommand.java
alb3rtobr: > Both parameters should be of type `boolean`, and you can achieve the same outcome directly through `spring-shell` features:
> 
> ```
> @CliOption(key = CliStrings.LIST_GATEWAY__SHOW_SENDERS_ONLY, specifiedDefaultValue = "true", unspecifiedDefaultValue = "false", help = CliStrings.LIST_GATEWAY__SHOW_SENDERS_ONLY) boolean showOnlySenders
> 
> @CliOption(key = CliStrings.LIST_GATEWAY__SHOW_RECEIVERS_ONLY, specifiedDefaultValue = "true", unspecifiedDefaultValue = "false", help = CliStrings.LIST_GATEWAY__SHOW_RECEIVERS_ONLY) boolean showOnlyReceivers
> ```
> 
> Spring shell will automatically populate the correct value for the `boolean` flags when parsing the user input, as below:
> Command 	showOnlyReceivers 	showOnlySenders
> list gateways 	false 	false
> list gateways --senders-only 	false 	true
> list gateways --senders-only=true 	false 	true
> list gateways --senders-only=false 	false 	false
> list gateways --receivers-only 	true 	false
> list gateways --receivers-only=true 	true 	false
> list gateways --receivers-only=false 	false 	false

And how do you difference between `list gateways --receivers-only --senders-only=false` and `list gateways --senders-only=false`? In both cases, `showOnlyReceivers=true` and `showOnlySenders=false` but the first one should return error and the second one should work.

alb3rtobr: I have just read your other comment about interceptors. If using an interceptor Im able to return error when `--senders-only`& `--receivers-only` are used together, then I could go back to use a boolean.

geode-core/src/main/java/org/apache/geode/management/internal/i18n/CliStrings.java
jujoramos: Small typo, this should read `Options --senders-only and --receivers-only cannot be used together`

geode-docs/tools_modules/gfsh/command-pages/list.html.md.erb
```diff
@@ -336,12 +336,16 @@ list gateways [--members=value(,value)*] [--groups=value(,value)*]
 |------------------------------------------------|--------------------------------------------------------------------------------------------------------------------------------|
 | <span class="keyword parmname">&#8209;&#8209;members</span> | Member(s) whose gateways senders and receiver display.                                                                         |
 | <span class="keyword parmname">\\-\\-groups</span>  | Group(s) of members for which Gateway Senders and Receivers will be displayed. Use a comma separated list for multiple groups. |
+| \\-\\-senders-only                                  | List only gateway senders. This parameter cannot be used together with `--receivers-only`                                                                                                     |
+| \\-\\-receivers-only                                | List only gateway receivers This parameter cannot be used together with `--senders-only`.                                                                                                   |
```
jujoramos: Small typo, thee's a missing `.` between `receivers` and `This`.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/wancommand/ListGatewaysCommandDUnitTest.java
```diff
@@ -369,4 +369,144 @@ public void listGatewaysShouldCorrectlyUpdateSendersConnectedCountWhenReceiverSt
...
   }
+
+  @Test
+  public void testListGatewaySenderOnly() {
+    setupClusters();
+
+    String command = CliStrings.LIST_GATEWAY + " --" + CliStrings.LIST_GATEWAY__SHOW_SENDERS_ONLY;
+    CommandResultAssert commandAssert = gfsh.executeAndAssertThat(command).statusIsSuccess();
+    commandAssert
+        .hasTableSection("gatewaySenders").hasRowSize(4)
```
DonalEvans: Is it possible to extract this "4" to a constant/variable so that it's more obvious why we expect 4 rows here? This comment also applies to other uses of a hard-coded 4 in following test methods.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/wancommand/ListGatewaysCommandDUnitTest.java
```diff
@@ -369,4 +369,144 @@ public void listGatewaysShouldCorrectlyUpdateSendersConnectedCountWhenReceiverSt
         .hasRowSize(1)
         .hasColumn("Sender Count").containsExactly("0");
   }
+
+  @Test
+  public void testListGatewaySenderOnly() {
```
DonalEvans: Could the test names be changed slightly to include the expected behaviour as well as the conditions? For example, this test would change to "testListGatewaySenderOnlyListsOnlyGatewaySenders." It can feel redundant in cases where the expected behaviour should be obvious from the wording of the command, but in other cases, such as when both options are specified, it can help make the intent of the test clearer.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/wancommand/ListGatewaysCommandDUnitTest.java
```diff
@@ -369,4 +369,144 @@ public void listGatewaysShouldCorrectlyUpdateSendersConnectedCountWhenReceiverSt
...
+    CommandResultAssert commandAssert = gfsh.executeAndAssertThat(command).statusIsSuccess();
+    commandAssert
+        .hasTableSection("gatewaySenders").hasRowSize(4)
+        .hasColumns().contains("GatewaySender Id", "Member");
+    commandAssert.hasTableSection("gatewayReceivers")
+        .hasRowSize(1).hasColumns().contains("Port", "Member");
+  }
+
+  @Test
+  public void testListAllGatewaysBothFalse() {
```
DonalEvans: 3 of these 4 test cases are redundant and can be safely removed without losing coverage, since the behaviour being tested is just whether the options `--senders-only` and `--receivers-only` are specified at all, not what their values are. If one of these tests were to fail but not the other 3, that would indicate a deeper problem with the gfsh parser that is outside the scope of what this class is intended to be testing and should be already covered in an existing test class.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/ListGatewayCommand.java
```diff
@@ -53,9 +57,23 @@ public ResultModel listGateway(
           help = CliStrings.LIST_GATEWAY__MEMBER__HELP) String[] onMember,
       @CliOption(key = {CliStrings.GROUP, CliStrings.GROUPS},
           optionContext = ConverterHint.MEMBERGROUP,
-          help = CliStrings.LIST_GATEWAY__GROUP__HELP) String[] onGroup)
+          help = CliStrings.LIST_GATEWAY__GROUP__HELP) String[] onGroup,
+      @CliOption(key = {CliStrings.LIST_GATEWAY__SHOW_RECEIVERS_ONLY},
+          specifiedDefaultValue = "true",
+          help = CliStrings.LIST_GATEWAY__SHOW_RECEIVERS_ONLY__HELP) Boolean showReceiversOnly,
```
jinmeiliao: You can use boolean instead Boolean, if you put unspecifiedDefaultVaue="false"

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/ListGatewayCommand.java
```diff
@@ -53,9 +57,23 @@ public ResultModel listGateway(
...
+          help = CliStrings.LIST_GATEWAY__SHOW_SENDERS_ONLY__HELP) Boolean showSendersOnly)
+
       throws Exception {
 
+    if (showReceiversOnly == null) {
+      showReceiversOnly = false;
+    }
+    if (showSendersOnly == null) {
+      showSendersOnly = false;
+    }
```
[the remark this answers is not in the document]
alb3rtobr: If I set them to "false" I cannot differentiate when they are false because the command was executed with that option from when they have not been introduced. I check the null value in the interceptor.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/ListGatewayCommand.java
alb3rtobr: I need Boolean to have three possible values: true, false and null. I need the null value for the interceptor check.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/ListGatewayCommand.java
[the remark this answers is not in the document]
DonalEvans: You're right, having an `unspecifiedDefaultValue` set means you can't check if the user has specified both options in the interceptor.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/ListGatewayCommand.java
jinmeiliao: in the interceptors, you can error out if both parameters have "true" value. You don't need to check for null

alb3rtobr: if you use `unspecifiedDefaultVaue="false"` then how could you differentiate between `list gateways --receivers-only=false --senders-only=false` and `list gateways --senders-only=false`? In both cases, the interceptor would receive `showOnlyReceivers=false` and `showOnlySenders=false` but the first command should return error and the second one should work.

jinmeiliao: I might be confused, but to me 
`list gateways`
`list gateways --sends-only=false`
`list gateways --receivers-only=false
`list gateways --receivers-only=false --senders-only=false` 

should all behave the same, since "false" is the default value for these two options. In each case, it all means " I want to see all of them"

alb3rtobr: you forgot the extra restriction that both options cannot be used together, so the fourth command you wrote will return an error :)

jinmeiliao: that's only when both values are true at the same time. The only invalid command should be:
`list gateways --receivers-only=true--senders-only=true` or
`list gateways --receivers-only --senders-only`

when either or both values are false, there should not be an error.

alb3rtobr: I think that using both parameters at the same time has no sense at all, and if someone tries to use them, it will be due to a misunderstanding of the documentation, this is why I like the idea of returning an error if both parameters are introduced.

jinmeiliao: I think we are disagreeing on what should be the outcome of the following command
`list gateways --receivers-only=false --senders-only=false`
My take is we should go ahead and show all of them, and you would like to throw an error. Probably we should involve some doc's people to interpret this.

alb3rtobr: yes, I want to return an error due to I want to return an error whenever both options are used together. @davebarnes97 , could you help with this?

davebarnes97: I understand you're asking me as a writer to help with the description of the behavior, but I think there's an inherent difficulty in the design. So 2 answers, first the design level, then the language level.
There are three choices of what to display: all, receivers, senders. So this is not a binary/boolean choice, it's ternary. The proposed implementation is a pair of booleans giving 4 total choices, where one is forbidden. I would like to suggest, instead, a parameter that specifies what to show, and recognizes 3 choices. For example,
  list gateways --display=senders
  list gateways --display=receivers
  list gateways --display=all
A value other than (senders, receivers, all) is an error.
If --display is not specified, the default is 'all' (the current behavior).

davebarnes97: Leaving aside my design recommendation, and focusing only on language, I think the difficulty arises because the choices "--receivers-only" and "--senders-only" are actually negatives (meaning "not senders" and "not receivers, respectively). If you recast them as positives "--receivers" and "--senders", there is no need to assign them true/false values. Presence means true, absence means false. So:
  list gateways --senders      // show senders
  list gateways --receivers   // show receivers
  list gateways                       // show both (current behavior)
  list gateways --senders --receivers  // show both
In this scheme, since specifying both is the same as specifying neither, there is no 4th case to worry about.

onichols-pivotal: My 2 cents:
the user will probably ever only care about:
`list gateways`
`list gateways --receivers-only`
`list gateways --senders-only`

Anything else (explicitly appending `=false`, specifying both, etc) falls under "don't care" in my opinion.  You could throw an error in every permutation other than the 3 above, or do anything that feels consistent with principal-of-least-surprise.

@davebarnes97 has a great suggestion to think about maybe constraining the input space another way to avoid the need for so much validation, but it does lead to a slightly "less natural" command syntax imho.

jinmeiliao: @davebarnes97 assuming we want to leave the design as is, let's just focus on the semantic meaning of the following command:
`list gateways --receivers-only=false --senders-only=false`

Is the above command self-conflicting in anyway that we have to throw an error? I know `list gateways --receivers-only --senders-only` is self-conflicting, but with false value, I don't think it is, but it's your call.

Honestly,  it's very unlikely that users will issue such command, so exactly what it does might not be that important, it just introduces some coding complexity that I was just trying to get rid of in the first place.

onichols-pivotal: I agree that treating `--<flag>=false` any differently from omitting it entirely violates principal of least surprise.

alb3rtobr: Thanks you all for your contributions. This sentence from @onichols-pivotal changed my mind:

> I agree that treating `--<flag>=false` any differently from omitting it entirely violates principal of least surprise.

If I think about using `false` as if the parameter is omitted, then it has sense for me not returning an error when using `--receivers-only=false --senders-only=false`. Thanks again and sorry @jinmeiliao for the misunderstanding.

alb3rtobr:
If everything is ok, could any committer merge this PR? Thanks

jinmeiliao:
looks like the acceptance test failed. Can you rebase to develop and see if this test would pass? @alb3rtobr

alb3rtobr:
> looks like the acceptance test failed. Can you rebase to develop and see if this test would pass? @alb3rtobr

Done!

PR #4434, opened 2019-12-05
Geode-7498: show index type

jmelchio:
co-authored-by: Dale Emery <demery@pivotal.io>
co-authored-by: Jinmei Liao <jliao@pivotal.io>

### For all changes:

### Note:

### DIFF
--- a/geode-management/src/main/java/org/apache/geode/management/configuration/IndexType.java
+++ b/geode-management/src/main/java/org/apache/geode/management/configuration/IndexType.java
@@ -37,4 +37,19 @@ public enum IndexType {
   public String getName() {
     return this.name;
   }
+
+  public static IndexType valueOfSynonym(String name) {
+    name = name.toUpperCase();
+
+    switch (name) {
+      case "KEY":
+        return valueOf("PRIMARY_KEY");
+      case "RANGE":
+        return valueOf("FUNCTIONAL");
+      case "HASH":
+        return valueOf("HASH_DEPRECATED");
+    }
+
+    throw new IllegalArgumentException("Unknown IndexType: " + name);
+  }
 }
--- a/geode-management/src/main/java/org/apache/geode/management/configuration/IndexType.java
+++ b/geode-management/src/main/java/org/apache/geode/management/configuration/IndexType.java
@@ -23,33 +23,9 @@ package org.apache.geode.management.configuration;
  * @since Geode 1.12
  */
 public enum IndexType {
-  FUNCTIONAL("RANGE"),
+  KEY,
+  FUNCTIONAL,
+  // this is used to represent deprecated index types not supported by the management API.
   @Deprecated
-  HASH_DEPRECATED("HASH"),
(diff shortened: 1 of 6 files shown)


jmelchio:
@dschneider-pivotal @demery-pivotal @jinmeiliao

geode-management/src/main/java/org/apache/geode/management/configuration/IndexType.java
```diff
@@ -0,0 +1,55 @@
...
+  public String getName() {
+    return this.name;
+  }
+
+  public static IndexType valueOfSynonym(String name) {
+    name = name.toUpperCase();
+
+    switch (name) {
+      case "KEY":
+        return valueOf("PRIMARY_KEY");
```
demery-pivotal: Consider `return PRIMARY_KEY;` Similar for the other values.

geode-management/src/main/java/org/apache/geode/management/configuration/IndexType.java
```diff
@@ -0,0 +1,55 @@
...
+  HASH_DEPRECATED("HASH"),
+  PRIMARY_KEY("KEY");
+
+  private String name;
+
+  IndexType(String name) {
+    this.name = name;
+  }
+
+  public String getName() {
```
demery-pivotal: `name` is an unfortunate name for this property, given that every enum already has a `name()` method. It's and confusing awkward that `name()` and `getName()` return different values.

Below, we call this `synonym`. Consider changing the field to `synonym` and this method to `getSynonym()`.

jinmeiliao: if we don't define the synonyms inside IndexType class, then I agree with you. If the IndexType is aware of the synonyms, then it's the responsibility of the class itself to provide a translation.

jinmeiliao: agree!

jinmeiliao:
if we don't define the synonyms inside IndexType class, then I agree with you. If the IndexType is aware of the synonyms, then it's the responsibility of the class itself to provide a translation.

> The mapping between the new `IndexType` enum and the old names is needed only by the converter. I think it's a mistake to put that mapping into `IndexType`, and to make the mapping visible via a method (`valueOfSynonym()`). Putting that method into `IndexType` makes the mapping part of the public API.

geode-management/src/main/java/org/apache/geode/management/configuration/IndexType.java
jmelchio: Thanks @demery-pivotal I applied this change.

geode-management/src/main/java/org/apache/geode/management/configuration/IndexType.java
jmelchio: @demery-pivotal I've applied this change.

PR #4436, opened 2019-12-06
GEODE-7503: Cleanup GemFireCacheImpl

kirklund:
Improve readability of GemFireCacheImpl. I also touched up the javadocs on the various Cache interfaces we have including User-visible ones.

I'm leaving this in many commits to facilitate incremental reviews, but I'll squash it down to one commit during merge.

This is a follow-up to previous PR and merge that added more constructor injection and unit tests to GemFireCacheImpl.

Summary:
1) Move inner-classes to bottom of outer class
2) Remove bug system ticket numbers
3) Cleanup javadocs, toString, and comments
4) Pull up javadocs to interfaces where applicable
5) Move all fields and constants above constructor in this order: static final, static non-final, final, volatile, all other fields
6) Apply VisibleForTesting annotation where applicable
7) Adjust lots of white-space and line-feeds
8) Fix many minor IDE warnings
9) Adjusted internal interfaces (including removal method from interface because it was not needed there)
10) Remove extraneous uses of final from local vars and parameters
11) Use select static imports to improve readability
12) Change Consumer<Void> usages to use Runnable


geode-core/src/main/java/org/apache/geode/internal/cache/GemFireCacheImpl.java
```diff
@@ -260,100 +282,143 @@
 import org.apache.geode.pdx.internal.PdxInstanceImpl;
 import org.apache.geode.pdx.internal.TypeRegistry;
 
-// TODO: somebody Come up with more reasonable values for {@link #DEFAULT_LOCK_TIMEOUT}, etc.
 /**
  * GemFire's implementation of a distributed {@link Cache}.
  */
-@SuppressWarnings("deprecation")
+@SuppressWarnings({"TypeMayBeWeakened", "ThisEscapedInObjectConstruction", "OverlyComplexMethod",
```
kirklund: Oops, I didn't mean to leave these in. I'll delete these.

geode-core/src/main/java/org/apache/geode/internal/cache/GemFireCacheImpl.java
```diff
@@ -604,21 +584,162 @@
...
+
+  /**
+   * Synchronization mutex for {@link #ccpTimer}.
+   */
+  private final Object ccpTimerMutex = new Object();
+
+  private final ExpirationScheduler expirationScheduler;
+
+  /**
+   * TODO make this a simple int guarded by riWaiters and get rid of the double-check
```
kirklund: This is an old TODO.

geode-core/src/main/java/org/apache/geode/internal/cache/GemFireCacheImpl.java
```diff
@@ -2844,41 +2748,22 @@ public String getName() {
...
 
-  /**
-   * Returns the number of seconds that have elapsed since the Cache was created.
-   *
-   * @since GemFire 3.5
-   */
   @Override
-  public int getUpTime() {
-    return (int) (System.currentTimeMillis() - creationDate.getTime()) / 1000;
+  public long getUpTime() {
```
kirklund: This is internal-only. It's defined on the InternalCache interface. All callers except one use long already. I moved the (int) case to RemoteCacheInfo.

geode-core/src/main/java/org/apache/geode/internal/cache/GemFireCacheImpl.java
```diff
@@ -2924,13 +2809,13 @@ private void addLocalHostAsServer(PoolFactory poolFactory) {
...
   /**
-   * Used to set the default pool on a new GemFireCache.
+   * Set the default pool on a new GemFireCache.
    */
-  @Override
-  public synchronized void determineDefaultPool() {
+  private synchronized void determineDefaultPool() {
     if (!isClient()) {
-      throw new UnsupportedOperationException();
+      throw new UnsupportedOperationException("determineDefaultPool requires a client cache");
```
kirklund: TODO: Wording should be improved to not use the name of a private method.

geode-core/src/main/java/org/apache/geode/internal/cache/xmlcache/CacheCreation.java
```diff
@@ -1075,7 +1078,12 @@ public CancelCriterion getCancelCriterion() {
   }
 
   @Override
-  public InternalQueryService getQueryService() {
+  public QueryService getQueryService() {
+    return queryService;
+  }
+
+  @Override
+  public InternalQueryService getInternalQueryService() {
```
kirklund: Explanation: I changed impl methods that narrow the types (return or params)  to match the interface. I added new methods with narrowed types where applicable (such as here).

geode-core/src/main/java/org/apache/geode/internal/cache/GemFireCacheImpl.java
```diff
@@ -841,9 +937,9 @@ public static GemFireCacheImpl getForPdx(String reason) {
       Function<InternalCache, InternalResourceManager> internalResourceManagerFactory,
       Supplier<Integer> serialNumberSupplier,
       HeapEvictorFactory heapEvictorFactory,
-      Consumer<Void> typeRegistryInit,
-      Consumer<Void> typeRegistryOpen,
-      Consumer<Void> typeRegistryClose,
+      Runnable typeRegistryInit,
+      Runnable typeRegistryOpen,
+      Runnable typeRegistryClose,
```
demery-pivotal: Yeah, this is much better. Later we may be able to bundle these three runnables into a single interface. And similar for any other clumps of related parameters.

geode-core/src/main/java/org/apache/geode/internal/cache/GemFireCacheImpl.java
```diff
@@ -2795,21 +2708,18 @@ public LogWriterI18n getSecurityLoggerI18n() {
     return system.getSecurityInternalLogWriter();
   }
 
+  @Deprecated
```
demery-pivotal: Should these be deprecated on `InternalCache` instead?

geode-core/src/main/java/org/apache/geode/internal/cache/GemFireCacheImpl.java
demery-pivotal: Is it possible (and reasonable) to move this check up the call stack, so that the exception message can refer to more public methods?

geode-core/src/main/java/org/apache/geode/internal/cache/GemFireCacheImpl.java
kirklund: Yep, that's a great idea!

geode-core/src/main/java/org/apache/geode/internal/cache/GemFireCacheImpl.java
kirklund: These are all deprecated in the interface with both javadoc and annotation. I copied the Deprecated annotation down to the impl to help the reader realize they're looking at a deprecated method.

kirklund:
Looks like AnalyzeManagementSerializablesJUnitTest is failing on develop as well as this PR branch. I'll rebase on develop soon.

geode-core/src/main/java/org/apache/geode/internal/cache/GemFireCacheImpl.java
kirklund: I'll look into that.

geode-core/src/main/java/org/apache/geode/internal/cache/GemFireCacheImpl.java
kirklund: Done

PR #4441, opened 2019-12-09
GEODE-7561: GW sender allow single dispatcher thread

mkevo:

### For all changes:

### Note:

### DIFF
--- a/geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/functions/GatewaySenderCreateFunction.java
+++ b/geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/functions/GatewaySenderCreateFunction.java
@@ -122,10 +122,12 @@ public class GatewaySenderCreateFunction implements InternalFunction {
     }
 
     Integer dispatcherThreads = gatewaySenderCreateArgs.getDispatcherThreads();
-    if (dispatcherThreads != null && dispatcherThreads > 1) {
+    if (dispatcherThreads != null) {
       gateway.setDispatcherThreads(dispatcherThreads);
+    }
 
-      String orderPolicy = gatewaySenderCreateArgs.getOrderPolicy();
+    String orderPolicy = gatewaySenderCreateArgs.getOrderPolicy();
+    if (dispatcherThreads != null && dispatcherThreads > 1) {
       gateway.setOrderPolicy(OrderPolicy.valueOf(orderPolicy));
     }
 
--- a/geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/functions/GatewaySenderCreateFunction.java
+++ b/geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/functions/GatewaySenderCreateFunction.java
@@ -126,8 +126,8 @@ public class GatewaySenderCreateFunction implements InternalFunction {
       gateway.setDispatcherThreads(dispatcherThreads);
     }
 
-    String orderPolicy = gatewaySenderCreateArgs.getOrderPolicy();
     if (dispatcherThreads != null && dispatcherThreads > 1) {
+      String orderPolicy = gatewaySenderCreateArgs.getOrderPolicy();
       gateway.setOrderPolicy(OrderPolicy.valueOf(orderPolicy));
     }
 
--- a/geode-gfsh/src/test/java/org/apache/geode/management/internal/cli/commands/CreateGatewaySenderCommandTest.java
+++ b/geode-gfsh/src/test/java/org/apache/geode/management/internal/cli/commands/CreateGatewaySenderCommandTest.java
@@ -124,6 +124,22 @@ public class CreateGatewaySenderCommandTest {
   }
 
   @Test
+  public void testSingleDispatcherThread() {
+    doReturn(mock(Set.class)).when(command).getMembers(any(), any());
+    cliFunctionResult = new CliFunctionResult("member",
+        CliFunctionResult.StatusState.OK, "cliFunctionResult");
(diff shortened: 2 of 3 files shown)


geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/functions/GatewaySenderCreateFunction.java
```diff
@@ -122,10 +122,12 @@ private GatewaySender createGatewaySender(Cache cache,
...
 
     Integer dispatcherThreads = gatewaySenderCreateArgs.getDispatcherThreads();
-    if (dispatcherThreads != null && dispatcherThreads > 1) {
+    if (dispatcherThreads != null) {
       gateway.setDispatcherThreads(dispatcherThreads);
+    }
 
-      String orderPolicy = gatewaySenderCreateArgs.getOrderPolicy();
+    String orderPolicy = gatewaySenderCreateArgs.getOrderPolicy();
+    if (dispatcherThreads != null && dispatcherThreads > 1) {
```
jhuynh1: Is the idea here that if dispatcher threads is 1, the order policy doesn't really matter, otherwise we should remove this dispatcherThreads > 1 check too correct?

mkevo: Yes, If we have 1 dispatcher thread we don't need order policy. In case there is a more dispatcher threads than 1, order policy is needed.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/wancommand/CreateDestroyGatewaySenderCommandDUnitTest.java
```diff
@@ -182,6 +182,42 @@ public void testCreateDestroyGatewaySender() {
         server3);
   }
 
+  /**
+   * GatewaySender with single dispatcher thread
+   */
+  @Test
+  public void testCreateDestroyGatewaySenderWithSingleDispatcherThread() {
```
jhuynh1: Cool, I ran this test on develop without your changes and it fails (as you already knew, more for anyone else wondering), it passes with this diff and fails without it.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/functions/GatewaySenderCreateFunction.java
agingade: Thats not what it looks like from "GatewaySenderFactoryImpl". The order policy is required if its using more than 1 dispatcher thread and if its not provided, it gets set to default one. 
And one could set order policy without setting the dispatcher thread. This is from the GatewaySenderFactoryImpl; I haven't looked into how it is used during event dispatch.

Based on logic in GatewaySenderFactoryImpl,  this could be
if (dispatcherThreads != null) 
 gateway.setDispatcherThreads(dispatcherThreads);

if (orderPolicy != null) {
try {
OrderPolicy.valueOf(orderPolicy) // orderPolicy could be null or invalid string
gateway.setOrderPolicy(OrderPolicy.valueOf(orderPolicy));
catch (Exception) {
// Throw an error message to user?
}
}

geode-gfsh/src/test/java/org/apache/geode/management/internal/cli/commands/CreateGatewaySenderCommandTest.java
```diff
@@ -123,6 +123,22 @@ public void whenCommandOnMember() {
...
   }
 
+  @Test
+  public void testSingleDispatcherThread() {
+    doReturn(mock(Set.class)).when(command).getMembers(any(), any());
+    cliFunctionResult = new CliFunctionResult("member",
+        CliFunctionResult.StatusState.OK, "cliFunctionResult");
+    functionResults.add(cliFunctionResult);
+    gfsh.executeAndAssertThat(command,
+        "create gateway-sender --member=xyz --id=1 --remote-distributed-system-id=1 " +
```
agingade: If there is no test with null or invalid order policy; please add one.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/functions/GatewaySenderCreateFunction.java
mkevo: Hi @agingade, 
OrderPolicy cannot have an invalid string as it doesn't allow any string except this from the OrderPolicy.ENUM: KEY, THREAD and PARTITION. 

1. If dispatcher threads is more than 1 you need to set order-policy to one of this OrderPolicy.ENUM. In case you set any other string you will get:
_java.lang.IllegalArgumentException: Failed to convert 'XXX' to type OrderPolicy for option 'order-policy'
No enum constant org.apache.geode.cache.wan.GatewaySender.OrderPolicy.XXX_
2. If dispatcher threads is equal 1 it doesn't need order-policy as it is alone there(You can set order-policy to some of this values, but they don't have any impact as it will not be used). 
3. If it is lower than 1 you will get exception and message that dispatcher threads cannot be less than 1.

geode-gfsh/src/test/java/org/apache/geode/management/internal/cli/commands/CreateGatewaySenderCommandTest.java
mkevo: Test added.

mivanac:
Since all test are passing, I will merge this PR

PR #4189, opened 2019-10-21
Revert "GEODE-6807: cache adviseUpdate and adviseAllEventsOrCached"

echobravopapa:
Reverts apache/geode#4086 - CI failing

PR #4442, opened 2019-12-09
Reverting "GEODE-6807: cache adviseUpdate and adviseAllE…

jhuynh1:
…ventsOrCached" (#4189)" (#4212)"

This reverts commit 80efe5cae9848f4fab45bb5c199c1e1f1fdf0eae.

While working on GEODE-7510 and other distribution related issues, we noticed that this commit causes this specific test scenario to fail.  It fails consistently in that it will fail within < 100 tries (for me < 50 and usually under 15) and without this commit, will no longer fail (after 500+ runs).

It looks like caching of the distribution profiles is possibly causing us to not do correct distribution for some reason.  The end result for this test is lingering values in secondaries.  

We have seen other weird scenarios but have not been able to identify the correct tests to report just yet.

### For all changes:

### Note:


geode-core/src/main/java/org/apache/geode/distributed/internal/DistributionAdvisor.java
```diff
@@ -1319,7 +1313,8 @@ public Set adviseProfileRemove() {
   // must synchronize when modifying profile array
   private synchronized boolean basicAddProfile(Profile p) {
     // don't add more than once, but replace existing profile
-    profilesVersion++;
+    // try {
```
igodwin: Looks like leftovers from debugging.

geode-core/src/main/java/org/apache/geode/internal/cache/CacheDistributionAdvisor.java
```diff
@@ -153,35 +140,19 @@ public String toString() {
...
-            if (cp.getInRecovery()) {
-              return false;
-            }
-            return cp.cachedOrAllEventsWithListener();
-          }));
-          adviseAllEventsVersion = tempProfilesVersion;
-          adviseInRecoveryVersion = tempInRecoveryVersion;
-        }
+    return adviseFilter(profile -> {
+      assert profile instanceof CacheProfile;
```
igodwin: Maybe update to use AssertJ.

geode-core/src/main/java/org/apache/geode/internal/cache/CacheDistributionAdvisor.java
```diff
@@ -191,30 +162,18 @@ public String toString() {
...
-              return dp.withReplication()
-                  || (cp.allEvents() && (dp.withStorage() || cp.hasCacheListener));
-            }));
-            adviseUpdateVersion = tempProfilesVersion;
-          }
-        }
-      }
-      return adviseSetforUpdate;
+      return adviseFilter(profile -> {
+        assert profile instanceof CacheProfile;
```
igodwin: AssertJ.

jhuynh1: I'm not going to make modifications to this PR. It's going to be a pure revert so that things don't get munged...

geode-core/src/main/java/org/apache/geode/distributed/internal/DistributionAdvisor.java
jhuynh1: This is just a straight revert command.  I'm trying not to cause any confusion with git and/or review. If I make modifications, I think this will cause confusion for anyone tracing this in the future.

PR #4446, opened 2019-12-09
GEODE-7536: Add serialization analysis to geode-gfsh

bschuchardt:
The new geode-gfsh module left all of its serializables in the
sanctioned list from geode-core.  This commit adds serialization
analysis to the new module and removes geode-gfsh serializables from
sanctioned-geode-core-serializables.txt.

I had to modify the base test class to read the sanctioned list using a
resource stream instead of a file.  Geode-gfsh, for some reason, has the
module's JAR file on the path and the test is finding the resource there
instead of as a regular file.

### For all changes:

### Note:


geode-junit/src/main/java/org/apache/geode/codeAnalysis/AnalyzeSerializablesJUnitTestBase.java
```diff
@@ -88,7 +89,7 @@
   private File expectedDataSerializablesFile;
   private String expectedSerializablesFileName =
       "sanctioned-" + getModuleName() + "-serializables.txt";
-  private File expectedSerializablesFile;
+  private InputStream expectedSerializablesInputStream;
```
upthewaterspout: Why is this InputStream a field on the class? Doesn't seem like something we should keep around or reuse.

geode-junit/src/main/java/org/apache/geode/codeAnalysis/CompiledClassUtils.java
```diff
@@ -292,11 +294,10 @@ public static void storeClassesAndMethods(List<ClassAndMethods> cams, File file)
...
   }
 
-  public static List<ClassAndVariableDetails> loadClassesAndVariables(File file)
+  public static List<ClassAndVariableDetails> loadClassesAndVariables(InputStream stream)
       throws IOException {
-    List<ClassAndVariableDetails> result = new LinkedList<ClassAndVariableDetails>();
-    FileReader fr = new FileReader(file);
-    BufferedReader in = new BufferedReader(fr);
+    List<ClassAndVariableDetails> result = new LinkedList<>();
+    BufferedReader in = new BufferedReader(new InputStreamReader(stream));
```
upthewaterspout: Should this be in a try-with-resources? Doesn't look like we put the close in a finally block.

geode-junit/src/main/java/org/apache/geode/codeAnalysis/AnalyzeSerializablesJUnitTestBase.java
```diff
@@ -433,7 +433,7 @@ private void loadClasses() throws IOException {
     String ideaBuildDirName = Paths.get(getModuleName(), "out", "production", "classes").toString();
     System.out.println("ideaBuildDirName is " + ideaBuildDirName);
     String ideaFQCNBuildDirName = Paths.get("out", "production",
-        "org.apache.geode." + getModuleName() + ".main").toString();
+        "geode." + getModuleName() + ".main").toString();
```
upthewaterspout: I wish there was a way to do this without fishing for an intellij (or gradle) directory name.

bschuchardt: I've looked at how ArchUnit finds classes and it's also classpath & search based but is considerably more complex, hooking into analysis classes in gradle, idea and in one case a class available starting in jdk 9.

geode-junit/src/main/java/org/apache/geode/codeAnalysis/CompiledClassUtils.java
bschuchardt: I changed this to a try-with-resources and moved the closing back to AnalyzeSerializablesJUnitTestBase.

geode-junit/src/main/java/org/apache/geode/codeAnalysis/AnalyzeSerializablesJUnitTestBase.java
bschuchardt: I've deleted this instance variable

PR #4447, opened 2019-12-09
GEODE-7643: Gateway unprocessedTokensMap appears to grow without bounds with replicated regions and peer accessors

BenjaminPerryRoss:
Allow LocalRegion.virtualPut() to throw ConcurrentCacheModificationException
Added throwsConcurrentModifiaction boolean argument to LocalRegion.virtualPut() to determine if the exception should be thrown
Added invokeCallbacks boolean argument to LocalRegion.virtualPut() to determine if bridge clients and gateway senders should be notified of the event
Fixed issue with inheriting new virtualPut implementations
Added DUnit test to confirm fix

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/AbstractUpdateOperation.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/AbstractUpdateOperation.java
@@ -135,6 +135,8 @@ public abstract class AbstractUpdateOperation extends DistributedCacheOperation 
     try {
       boolean updated = false;
       boolean doUpdate = true; // start with assumption we have key and need value
+      boolean invokeCallbacks = true;
+
       if (shouldDoRemoteCreate(rgn, ev)) {
         if (logger.isDebugEnabled()) {
           logger.debug("doPutOrCreate: attempting to update or create entry");
@@ -149,14 +151,19 @@ public abstract class AbstractUpdateOperation extends DistributedCacheOperation 
           // if the oldValue is the DESTROYED token and overwrite is disallowed,
           // then basicPut will set the blockedDestroyed flag in the event
           boolean overwriteDestroyed = ev.getOperation().isCreate();
-          if (rgn.basicUpdate(ev, true /* ifNew */, false/* ifOld */, lastMod,
-              overwriteDestroyed)) {
-            rgn.getCachePerfStats().endPut(startPut, ev.isOriginRemote());
-            // we did a create, or replayed a create event
-            doUpdate = false;
-            updated = true;
-          } else { // already exists. If it was blocked by the DESTROYED token, then
-            // do no update.
+          try {
+            if (rgn.basicUpdate(ev, true, false, lastMod, overwriteDestroyed, true, true)) {
+              rgn.getCachePerfStats().endPut(startPut, ev.isOriginRemote());
+              // we did a create, or replayed a create event
+              doUpdate = false;
+              updated = true;
+            } else {
+              // already exists. If it was blocked by the DESTROYED token, then
+              // do no update.
+              doUpdate = checkIfToUpdateAfterCreateFailed(rgn, ev);
+            }
+          } catch (ConcurrentCacheModificationException ex) {
+            invokeCallbacks = false;
             doUpdate = checkIfToUpdateAfterCreateFailed(rgn, ev);
           }
         } finally {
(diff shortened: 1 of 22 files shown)


geode-core/src/main/java/org/apache/geode/internal/cache/AbstractUpdateOperation.java
```diff
@@ -179,8 +186,15 @@ public static boolean doPutOrCreate(LocalRegion rgn, EntryEventImpl ev, long las
             br.getPartitionedRegion().getPrStats().startApplyReplication();
           }
           try {
-            if (rgn.basicUpdate(ev, false/* ifNew */, true/* ifOld */, lastMod,
-                overwriteDestroyed)) {
+            boolean secondAttempt = false;
+            try {
+              secondAttempt = rgn.basicUpdate(ev, false/* ifNew */, true/* ifOld */, lastMod,
```
gesterzhou: I think this part's logic is incorrect. And the variable secondAttempt should be changed to a better name or omitted.

The rgn.basicUpdate can only have 3 possible results:
1) return true: which means the . update succeeded. Then we should return.
2) return false but not due to CME. It should be caused by the existing entry was removed by another thread. So update with ifOld==true failed. In that case, we should retry with basicUpdate(ifNew==false, ifOld==false, throwConcurrentModificaiton==false), i.e. do update and ignore the possible CME.
3) return false due to CME. It should retry with basicUpdate(ifNew==false, ifOld==false, shouldNotify==false, throwConcurrentModificaiton==false)

You can see in both case 2) and 3), we need to retry basicUpdate(ifNew==false, ifOld==false, , throwConcurrentModificaiton==false). The only difference is: if there's CME, the shouldNotify==false.

From above analysis, variable secondAttempt can be omitted.

geode-core/src/main/java/org/apache/geode/internal/cache/AbstractUpdateOperation.java
```diff
@@ -179,22 +186,32 @@ public static boolean doPutOrCreate(LocalRegion rgn, EntryEventImpl ev, long las
             br.getPartitionedRegion().getPrStats().startApplyReplication();
           }
           try {
-            if (rgn.basicUpdate(ev, false/* ifNew */, true/* ifOld */, lastMod,
-                overwriteDestroyed)) {
+            boolean secondUpdateSuccess;
```
gesterzhou: maybe "firstUpdateSuccess" is a better name, since there're 2 updates here.

DonalEvans: The first update attempt takes place on line 155. Would it be worth extracting that method call to a boolean similar to second- and thirdUpdateSuccess, for consistency?

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialGatewaySenderQueueDUnitTest.java
```diff
@@ -56,6 +58,85 @@
 @Category({WanTest.class})
 public class SerialGatewaySenderQueueDUnitTest extends WANTestBase {
 
+  @Test
+  public void unprocessEventsLinger() throws Exception {
```
jhuynh1: would we be able to have a better test name?

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/UpdateVersionDUnitTest.java
```diff
@@ -270,8 +271,11 @@ private VersionTag putEntryAndGetReplicatedRegionVersionTag(String key) {
 
     EntryEventImpl event =
         createNewEvent((DistributedRegion) region, versionTag, entry.getKey(), "value-3");
-
-    ((LocalRegion) region).basicUpdate(event, false, true, 0L, false);
+    try {
+      ((LocalRegion) region).basicUpdate(event, false, true, 0L, false);
+    } catch (ConcurrentCacheModificationException ex) {
```
jhuynh1: Is this catch needed?  Shouldn't the changes be a no change for all existing callers?

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialGatewaySenderQueueDUnitTest.java
DonalEvans: Done!

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/UpdateVersionDUnitTest.java
DonalEvans: You're right, it wasn't needed, it's been removed.

geode-core/src/main/java/org/apache/geode/internal/cache/AbstractUpdateOperation.java
DonalEvans: We have renamed the booleans to first-, second- and thirdBasicUpdateSuccess for each of the three calls to `LocalRegion.basicUpdate` in the method, for clarity.

DonalEvans:
> any ideas on how to make break the doPutOrCreate method down?

There are several problems with the doPutOrCreate method, both in terms of its complexity and size, and the fact that it always returns true, even if the update does not actually succeed. Allowing the method to return false in the even that all three calls to `LocalRegion.basicUpdate` fail results in several DUnit tests failing, so it might be necessary to create a separate Geode ticket for a refactor/fix of this method.

PR #4456, opened 2019-12-10
Mass test script change

mhansonp:
Updated the mass test script to use a local look at the concourse config.


mhansonp:
The failure in DistributedTestOpenJDK11 is not related to this change as the change is not in java code. it is a script for characterizing concourse run history.

ci/bin/concourse_job_performance.py
```diff
@@ -65,6 +69,16 @@ def main(url, team, pipeline, job, max_fetch_count, build_count, authorization_c
     print_results(len(build_to_examine), expected_failed_builds, long_list_of_failures, url, failure_url_base)
 
 
+def get_cookie(url):
+    data = yaml.load(open("/Users/mhanson/.flyrc"))
+    for target in data["targets"]:
```
onichols-pivotal: could we use `~` instead?

PR #4465, opened 2019-12-12
show "who" and "when" from cluster config for deployed jars

dschneider-pivotal:

### For all changes:

### Note:


geode-core/src/main/java/org/apache/geode/internal/security/IntegratedSecurityService.java
```diff
@@ -125,6 +126,20 @@ public Subject getSubject() {
     return currentUser;
   }
 
+  @Override
+  public Object getPrincipal() {
+    Subject subject;
+    try {
+      subject = getSubject();
+    } catch (AuthenticationRequiredException ignored) {
```
jinmeiliao: I think we should not catch this exception. When this happens, something is seriously wrong and nothing should behave correctly afterwards.

demery-pivotal: Fixed

jinmeiliao:
Would be great to have an upgrade test

PR #4468, opened 2019-12-12
GEODE-7563: eliminate home-grown ConcurrentHashSet

Bill:
[GEODE-7563](https://issues.apache.org/jira/browse/GEODE-7563)

### For all changes:

### Note:


geode-core/src/main/java/org/apache/geode/internal/cache/DiskInitFile.java
```diff
@@ -336,9 +335,7 @@
 
   private final LongOpenHashSet crfIds;
```
echobravopapa: I realize these variable names are pre-existing... any chance to make them human understandable...?

Bill: I'm calling that out of scope.

Bill:
Since I had to remove the `ConcurrentHashSet` entry from `geode-core/src/main/resources/org/apache/geode/internal/sanctioned-geode-core-serializables.txt` the next thing I have to do is analyze the places where `ConcurrentHashSet` was used (Java classes) to see which, if any, were actually serialized (and which, therefore, might break as a result of the changes in this PR)

geode-core/src/main/java/org/apache/geode/internal/cache/DiskInitFile.java
jhuynh1: Also, i think crf Ids is the readable form in this case :-)

PR #4484, opened 2019-12-16
GEODE-7424: create index on group

dschneider-pivotal:
Now makes sure that create index on a group works correctly.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/management/internal/configuration/validators/IndexValidator.java
+++ b/geode-core/src/main/java/org/apache/geode/management/internal/configuration/validators/IndexValidator.java
@@ -15,11 +15,22 @@
 
 package org.apache.geode.management.internal.configuration.validators;
 
+import org.apache.geode.cache.configuration.CacheConfig;
+import org.apache.geode.cache.configuration.RegionConfig;
+import org.apache.geode.distributed.ConfigurationPersistenceService;
+import org.apache.geode.management.api.ClusterManagementException;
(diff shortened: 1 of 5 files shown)


geode-core/src/main/java/org/apache/geode/management/internal/configuration/validators/IndexValidator.java
```diff
@@ -15,11 +15,22 @@
...
+import org.apache.geode.management.api.ClusterManagementException;
+import org.apache.geode.management.api.ClusterManagementResult;
 import org.apache.geode.management.configuration.Index;
 import org.apache.geode.management.configuration.IndexType;
 import org.apache.geode.management.internal.CacheElementOperation;
 
 public class IndexValidator implements ConfigurationValidator<Index> {
+  private final ConfigurationPersistenceService persistenceService;
+
+  public IndexValidator(ConfigurationPersistenceService persistenceService) {
```
jinmeiliao: probably injecting RegionConfigManager and use the `get` method on that would be easier to find the existence of the region.

dschneider-pivotal: I think it would actually be more work. RegionConfigManager.get(Region config, CacheConfig existing) still needs a CacheConfig so we still need to get that from the persistenceService and we would also need to create a "Region" instance and set its name. So I think the current code is easier.

PR #4488, opened 2019-12-17
GEODE-7582: Update initlocator list

mivanac:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/integrationTest/java/org/apache/geode/cache/client/internal/AutoConnectionSourceImplJUnitTest.java
+++ b/geode-core/src/integrationTest/java/org/apache/geode/cache/client/internal/AutoConnectionSourceImplJUnitTest.java
@@ -420,9 +420,12 @@ public class AutoConnectionSourceImplJUnitTest {
     assertEquals(updateLocatorInterval, source.getLocatorUpdateInterval());
   }
 
-
+  /**
+   * This tests validates update of IP of locator from initialLocators list.
+   * In this test we validate this using identityHashCode.
+   */
   @Test
-  public void testLocatorIpChange2() {
+  public void testinitialLocatorIpChange() {
     int port = 11011;
     List<InetSocketAddress> locators = new ArrayList<>();
     InetSocketAddress floc1 = InetSocketAddress.createUnresolved("localhost", port);
@@ -431,21 +434,21 @@ public class AutoConnectionSourceImplJUnitTest {
     locators.add(floc1);
     locators.add(floc2);
 
-    List<HostAddress> la = new ArrayList<>();
-    la.add(new HostAddress(floc1, floc1.getHostName()));
-    la.add(new HostAddress(floc2, floc2.getHostName()));
+    List<LocatorAddress> la = new ArrayList<>();
+    la.add(new LocatorAddress(floc1, floc1.getHostName()));
+    la.add(new LocatorAddress(floc2, floc2.getHostName()));
 
     AutoConnectionSourceImpl src = new AutoConnectionSourceImpl(la, "", 60 * 1000);
     src.setPool(pool);
 
     // This method will create a new InetSocketAddress of floc1
-    src.updateLocatorInLocatorList(new HostAddress(floc1, floc1.getHostName()));
+    src.updateLocatorInLocatorList(new LocatorAddress(floc1, floc1.getHostName()));
 
-    List<HostAddress> cLocList = src.getCurrentLocatorsAddresses();
+    List<LocatorAddress> cLocList = src.getCurrentLocatorsAddresses();
 
     Assert.assertEquals(2, cLocList.size());
 
-    for (HostAddress t : cLocList) {
+    for (LocatorAddress t : cLocList) {
       Assert.assertNotSame("Should have replaced floc1 instance", t.getSocketInetAddressNoLookup(),
           floc1);
     }
@@ -456,10 +459,10 @@ public class AutoConnectionSourceImplJUnitTest {
 
     src.updateLocatorList(response);
 
-    List<HostAddress> cLocList2 = src.getCurrentLocatorsAddresses();
+    List<LocatorAddress> cLocList2 = src.getCurrentLocatorsAddresses();
 
     Assert.assertEquals(2, cLocList2.size());
-    for (HostAddress t : cLocList2) {
+    for (LocatorAddress t : cLocList2) {
       Assert.assertNotSame("Should have replaced floc1 instance", t.getSocketInetAddressNoLookup(),
           floc1);
     }
--- a/geode-core/src/integrationTest/java/org/apache/geode/cache/client/internal/AutoConnectionSourceImplJUnitTest.java
+++ b/geode-core/src/integrationTest/java/org/apache/geode/cache/client/internal/AutoConnectionSourceImplJUnitTest.java
@@ -170,41 +170,6 @@ public class AutoConnectionSourceImplJUnitTest {
   }
 
   /**
-   * This test validates the AutoConnectionSourceImpl.updateLocatorInLocatorList method. That method
-   * takes InetSocketAddres of locator which unable to connect to locator. And update that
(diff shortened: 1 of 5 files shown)


geode-core/src/integrationTest/java/org/apache/geode/cache/client/internal/AutoConnectionSourceImplJUnitTest.java
```diff
@@ -411,6 +411,53 @@ public void testLocatorUpdateIntervalZero() {
     assertEquals(updateLocatorInterval, source.getLocatorUpdateInterval());
   }
 
+
+  @Test
+  public void testLocatorIpChange2() {
```
jhuynh1: Would you be able to modify the test name to describe what we are testing ?  I don't know much about the code and unable to quickly decipher what we are trying to test.

mivanac:
If IP of locator from initialLocators list is updated, we have problem eventualy, because updated LocatorAddress is overwritten with old LocatorAddress locator from initialLocators list (in procedure updateLocatorList). If we do not want additional filtering, possible solution is to update initialLocators list.

geode-core/src/main/java/org/apache/geode/cache/client/internal/AutoConnectionSourceImpl.java
```diff
@@ -267,10 +267,46 @@ protected void updateLocatorInLocatorList(HostAddress locator) {
...
+          locatorfound = true;
+          break;
+        }
+      }
+      if (!locatorfound) {
+        initLocators.add(initloc);
+      }
+    }
+    return initLocators;
+
```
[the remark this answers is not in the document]
mivanac: We just want to return Set of initial locators, with newest/updated HostAddress.

mivanac:
Solution updated, with update of initiallocators, if any of them change IP address.

mivanac:
Solution with HostAddress class is to use DNS query before contacting each locator, which is according to current implementation intentionally avoided. This the reason I would prefer last proposed solution.

mivanac:
If I understand your proposal correctly, you would like just to remove usage of getSocketInetAddressNoLookup() method, and always use getSocketInetAddress() which then perform DNS resolving (create the new instance of InetSocketAddress)

pivotal-jbarrett:
@mivanac That isn't exactly what I am asking but it would work too. In that way you would always return a new `InetSocketAddress` via `InetSocketAddress.createUnresolved()`.

My suggestion was to not even use an `InetSocketAddress` in the `LocatorAddress`, but I think your approach works too.

I think we should make sure that the `InetSocketAddress` inside the `LocatorAddress` is unresolved as well. Its not the end of the world if it is but there isn't much point to it ever resolving.

mivanac:
I just updated solution with, possibility to update InetSocketAddress in LocatorAddress. Or I can modify it with your proposal

pivotal-jbarrett:
@mivanac I think something like this attached change is really the only file that would need adjustment.

pivotal-jbarrett:
@mivanac Something like this attached change should really be the only thing necessary. 

[LocatorAddress.java.zip](https://github.com/apache/geode/files/4073085/LocatorAddress.java.zip)

pivotal-jbarrett:
Warning! My suggestion is untested. Just a POC.

mivanac:
Thanks for comments and proposal.

geode-core/src/integrationTest/java/org/apache/geode/cache/client/internal/AutoConnectionSourceImplJUnitTest.java
```diff
@@ -420,6 +420,56 @@ public void testLocatorUpdateIntervalZero() {
...
 
+  /**
+   * This tests validates update of IP of locator from initialLocators list.
+   * In this test we validate this using identityHashCode.
+   */
+  @Test
+  public void testinitialLocatorIpChange() {
+    int port = 11011;
+    List<InetSocketAddress> locators = new ArrayList<>();
+    InetSocketAddress floc1 = InetSocketAddress.createUnresolved("localhost", port);
```
moleske: `floc1` and `floc2` were confusing names to me (mostly because by the end of the test I forgot what they stood for and started thinking about a flock of birds, but I digress...)

I think some thing like `unresolvedLocalhost` and `fakeLocalhost` would be a bit more clearer

geode-core/src/integrationTest/java/org/apache/geode/cache/client/internal/AutoConnectionSourceImplJUnitTest.java
```diff
@@ -420,6 +420,56 @@ public void testLocatorUpdateIntervalZero() {
...
+    InetSocketAddress floc2 = new InetSocketAddress("fakeLocalHost2", port);
+
+    locators.add(floc1);
```
moleske: `locators` is actually never used, values are only added to this list and never accessed

geode-core/src/integrationTest/java/org/apache/geode/cache/client/internal/AutoConnectionSourceImplJUnitTest.java
aaronlindsey: I agree it would be good to have a more descriptive name here

mivanac:
With last update of  LocatorAddress, some methods become unnecessary, so I removed them. Also test with IP change are no more valid.

geode-tcp-server/src/main/java/org/apache/geode/distributed/internal/tcpserver/LocatorAddress.java
```diff
@@ -15,96 +15,77 @@
...
-  private final String hostname;
-  private final int port;
-  private final boolean isIpString;
 
   public LocatorAddress(InetSocketAddress loc, String locStr) {
-    this.socketInetAddress = loc;
-    this.hostname = locStr;
-    this.port = loc.getPort();
-    this.isIpString = InetAddressValidator.getInstance().isValid(locStr);
+    if (InetAddressValidator.getInstance().isValid(locStr)) {
```
moleske: I think we could add some unit tests for LocatorAddress to cover some of these logic changes.  I'm thinking things like the following pseudo code

```java
public class LocatorAddressTest {
    @Test
    public void getSocketInentAddress_whenResolvableAddress_returnsTheSocketAddress() {
        InetSocketAddress thing = new InetSocketAddress(8080);
        LocatorAddress subject = new LocatorAddress(thing, "localhost");

        InetSocketAddress actual = subject.getSocketInetAddress();

        assertThat(actual, isEqualTo(thing));
        assertThat(!actual.isUnresolved());
    }
}
```

Not 100% sure it will work out the way I think, but may be worth a shot

geode-tcp-server/src/test/java/org/apache/geode/distributed/internal/tcpserver/LocatorAddressTest.java
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.apache.geode.distributed.internal.tcpserver;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
```
aaronlindsey: It's generally been agreed upon by the community to use AssertJ instead of JUnit asserts (e.g. `assertThat(...)`). There was a dev list thread about this at some point. It's not strictly necessary, but good for consistency.

mivanac:
@bschuchardt are this changes OK with you.

mivanac:
@bschuchardt are this changes OK with you. If they are could you remove indication that changes are requested. Thanks.

mivanac:
Since we have 3 approvals for this PR, and requested changes are addressed, could we close this PR.

PR #4494, opened 2019-12-17
GEODE-7589: Added ability to configure wan to dispatch based on time

jhuynh1:
* GEODE-7589: Reduced allocation of arraylist to batch size instead of batch size * 2
  * GEODE-7589: Parameterize AEQ Test so same test can be used for Serial and Parallel for newly added test
  * If batchSize is -1 (making dispatch time based), then size for allocating batch collection will be default batch size

### For all changes:

### Note:


geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
```diff
@@ -1215,25 +1216,26 @@ public List peek(int batchSize, int timeToWait) throws InterruptedException, Cac
...
         Thread.sleep(50);
       } catch (InterruptedException e) {
         Thread.currentThread().interrupt();
       }
       blockProcessorThreadIfRequired();
-      return batch;
+      return Collections.EMPTY_LIST;
     }
+    List<GatewaySenderEventImpl> batch =
+        new ArrayList<>(batchSize == BATCH_BASED_ON_TIME_ONLY ? DEFAULT_BATCH_SIZE : batchSize);
```
agingade: This will create array of capacity 100 (DEFAULT_BATCH_SIZE). Earlier the array growth was taken care by java. If there are not many events in the queue, it may be overhead.

geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
```diff
@@ -1215,25 +1216,26 @@ public List peek(int batchSize, int timeToWait) throws InterruptedException, Cac
...
     long start = System.currentTimeMillis();
     long end = start + timeToWait;
 
     // Add peeked events
-    addPeekedEvents(batch, batchSize);
+    addPeekedEvents(batch, batchSize == BATCH_BASED_ON_TIME_ONLY ? DEFAULT_BATCH_SIZE : batchSize);
 
     int bId = -1;
-    while (batch.size() < batchSize) {
+    while (batchSize == BATCH_BASED_ON_TIME_ONLY || batch.size() < batchSize) {
```
agingade: Is there a need to control; how big a batch size can grow...If user has set a long time, the batch size could be huge.

geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
jhuynh1: This was an attempt to make serial and parallel implementations more consistent.  Javas default size is quite small and I would guess that most of the time we are having to grow to the "batchSize" that is configured anyways?

geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
jhuynh1: This is for a specific case where an application just wanted it time triggered.  The work around is to specify a large batch-size currently but that also has large overhead.  In this case, I believe the application developers can determine what an ideal time is.

PR #4497, opened 2019-12-18
add missing variable

onichols-pivotal:
CONCOURSE_HOST is referenced in deploy_metrics.sh but was not defined when deploying the meta pipeline.


onichols-pivotal:
@BalaKaza

PR #4503, opened 2019-12-18
GEODE-7425 delete index

dschneider-pivotal:

### For all changes:

### Note:

### DIFF
--- a/geode-assembly/src/distributedTest/java/org/apache/geode/management/internal/rest/ListIndexManagementDUnitTest.java
+++ b/geode-assembly/src/distributedTest/java/org/apache/geode/management/internal/rest/ListIndexManagementDUnitTest.java
@@ -281,7 +281,7 @@ public class ListIndexManagementDUnitTest {
     locator.waitUntilRegionIsReadyOnExactlyThisManyServers("/region2", 3);
 
     Index index = new Index();
-    index.setName("index");
+    index.setName("index.1");
     index.setExpression("key");
     index.setRegionPath("/region2");
     index.setIndexType(IndexType.KEY);
@@ -292,7 +292,7 @@ public class ListIndexManagementDUnitTest {
     List<IndexInfo> runtimeResult = indexResult.getRuntimeResult();
     assertSoftly(softly -> {
       softly.assertThat(indexConfig.getRegionName()).isEqualTo("region2");
-      softly.assertThat(indexConfig.getName()).isEqualTo("index");
+      softly.assertThat(indexConfig.getName()).isEqualTo("index.1");
       softly.assertThat(indexConfig.getRegionPath()).isEqualTo("/region2");
       softly.assertThat(indexConfig.getExpression()).isEqualTo("key");
       softly.assertThat(runtimeResult).extracting(IndexInfo::getMemberName)
@@ -304,7 +304,7 @@ public class ListIndexManagementDUnitTest {
     assertSoftly(softly -> {
       softly.assertThat(deleteIndexResult.isSuccessful()).isTrue();
       softly.assertThatThrownBy(() -> cms.get(index))
-          .hasMessageContaining("Index 'index' does not exist");
+          .hasMessageContaining("Index 'index.1' does not exist");
       softly.assertThat(deleteRegionResult.isSuccessful()).isTrue();
       softly.assertThatThrownBy(() -> cms.get(config))
           .hasMessageContaining("Region 'region2' does not exist");
--- a/geode-assembly/src/distributedTest/java/org/apache/geode/management/internal/rest/ListIndexManagementDUnitTest.java
+++ b/geode-assembly/src/distributedTest/java/org/apache/geode/management/internal/rest/ListIndexManagementDUnitTest.java
@@ -47,7 +47,7 @@ import org.apache.geode.test.junit.rules.MemberStarterRule;
 public class ListIndexManagementDUnitTest {
 
   private Region regionConfig;
-  private Index index;
+  private Index indexConfig;
 
   @ClassRule
   public static ClusterStartupRule lsRule = new ClusterStartupRule();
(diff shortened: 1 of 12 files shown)


geode-web-management/src/main/java/org/apache/geode/management/internal/rest/controllers/RegionManagementController.java
```diff
@@ -196,4 +196,20 @@ public ClusterManagementResult deleteRegion(
     return new ResponseEntity<>(result,
         HttpStatus.CREATED);
   }
+
+  @ApiOperation(value = "delete region index")
+  @PreAuthorize("@securityService.authorize('CLUSTER', 'MANAGE', 'QUERY')")
+  @DeleteMapping(REGION_CONFIG_ENDPOINT + "/{regionName}" + INDEXES + "/{indexName}")
```
jinmeiliao: Would the index name have "." in it? Should we add that :. thing like other end points?

geode-web-management/src/main/java/org/apache/geode/management/internal/rest/controllers/RegionManagementController.java
```diff
@@ -196,4 +196,20 @@ public ClusterManagementResult deleteRegion(
     return new ResponseEntity<>(result,
         HttpStatus.CREATED);
   }
+
+  @ApiOperation(value = "delete region index")
+  @PreAuthorize("@securityService.authorize('CLUSTER', 'MANAGE', 'QUERY')")
```
jinmeiliao: would be nice to add an entry in the "ClusterManagementSecurityRestIntegrationTest" to test this.

geode-web-management/src/main/java/org/apache/geode/management/internal/rest/controllers/RegionManagementController.java
dschneider-pivotal: Good catch. We should do that for all endpoints that end with a parameter

jmelchio: Updated the code to cover this scenario.

geode-web-management/src/main/java/org/apache/geode/management/internal/rest/controllers/RegionManagementController.java
jmelchio: Added context for testing permissions in `ClusterManagementSecurityRestIntegrationTest`.

jmelchio:
@demery-pivotal please review

geode-web-management/src/main/java/org/apache/geode/management/internal/rest/controllers/RegionManagementController.java
```diff
@@ -196,4 +196,20 @@ public ClusterManagementResult deleteRegion(
...
+  @DeleteMapping(REGION_CONFIG_ENDPOINT + "/{regionName}" + INDEXES + "/{indexName:.+}")
```
dschneider-pivotal: Did you consider adding a test to verify ":.+" is working to RegionManagementControllerSpringTest? Currently that is all that test does so you should be able to copy one of its current methods to cover this endpoint

geode-web-management/src/main/java/org/apache/geode/management/internal/rest/controllers/RegionManagementController.java
```diff
@@ -196,4 +196,20 @@ public ClusterManagementResult deleteRegion(
...
+  public ClusterManagementResult deleteRegion(
```
dschneider-pivotal: Shouldn't this method be named deleteIndex?

jmelchio: Yes. I'll make that change.

geode-web-management/src/main/java/org/apache/geode/management/internal/rest/controllers/RegionManagementController.java
jmelchio: We did make a change to the DUnit test to do the same and that will exercise the endpoint. I can add one to the `RegionManagementControllerSpringTest` as well.

jmelchio:
@demery-pivotal good feedback. I'll update the code to address the issues.

geode-assembly/src/distributedTest/java/org/apache/geode/management/internal/rest/ListIndexManagementDUnitTest.java
```diff
@@ -52,12 +53,14 @@
...
   private static ClusterManagementService cms;
+  private static MemberVM locator;
 
   @BeforeClass
   public static void beforeClass() {
-    MemberVM locator = lsRule.startLocatorVM(0, MemberStarterRule::withHttpService);
+    locator = lsRule.startLocatorVM(0, MemberStarterRule::withHttpService);
     MemberVM server1 = lsRule.startServerVM(1, locator.getPort());
     MemberVM server2 = lsRule.startServerVM(2, locator.getPort());
+    MemberVM server3 = lsRule.startServerVM(3, "group1", locator.getPort());
```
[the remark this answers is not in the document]
jmelchio: > Please name the members to indicate their purpose in the test. For example, t looks as if the key idea for `server3` is that it is in a group. Name both the variable and the server to indicate that intent. That way, when someone reads the test, they have a better chance of understanding why this variable or string is used as an input to something, or should appear in a result.

I'm not sure if we can name the server. I believe the server name is simply the word `server` with the number that is given in the `startServer` method of the rule. I can certainly change the name of the variable.

geode-assembly/src/distributedTest/java/org/apache/geode/management/internal/rest/ListIndexManagementDUnitTest.java
```diff
@@ -226,4 +229,85 @@ public void listNonExistingIndexesWithIdFilter() {
       softly.assertThat(list.isSuccessful()).isTrue();
     });
   }
+
+  @Test
+  public void createAndDeleteIndex_success_in_specific_group() {
+    Region config = new Region();
```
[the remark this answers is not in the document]
jmelchio: > Consider: If a variable represents a filter, name it to identify it as a filter. Like `regionFilter` and `indexFilter`. If a variable represents a config, name it to identify it as a config, like `regionConfig` and `indexConfig`.

A bit of a challenge here is that in some cases we use the variable to create something first, and as a filter to fetch to confirm the creation later.

geode-assembly/src/distributedTest/java/org/apache/geode/management/internal/rest/ListIndexManagementDUnitTest.java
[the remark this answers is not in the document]
demery-pivotal: Maybe in that case we can find a name that encompasses both intentions. Maybe something like `configSpecification`.

PR #4508, opened 2019-12-19
GEODE-7544 Break dependency on ClassPathLoader

bschuchardt:
Look for the jgroups configuration file using regular class loaders
instead of geode-core's ClassPathLoader.

### For all changes:

- [tests already exist] Have you written or updated unit tests to verify your changes?

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/distributed/internal/membership/gms/messenger/JGroupsMessenger.java
+++ b/geode-core/src/main/java/org/apache/geode/distributed/internal/membership/gms/messenger/JGroupsMessenger.java
@@ -91,7 +91,6 @@ import org.apache.geode.distributed.internal.membership.gms.locator.FindCoordina
 import org.apache.geode.distributed.internal.membership.gms.locator.FindCoordinatorResponse;
 import org.apache.geode.distributed.internal.membership.gms.messages.JoinRequestMessage;
 import org.apache.geode.distributed.internal.membership.gms.messages.JoinResponseMessage;
-import org.apache.geode.internal.ClassPathLoader;
 import org.apache.geode.internal.OSProcess;
 import org.apache.geode.internal.cache.DistributedCacheOperation;
 import org.apache.geode.internal.net.SocketCreator;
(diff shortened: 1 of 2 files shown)


geode-core/src/main/java/org/apache/geode/distributed/internal/membership/gms/messenger/JGroupsMessenger.java
```diff
@@ -201,15 +200,24 @@ public void init(Services<ID> s) throws MembershipConfigurationException {
...
     if (config.isMcastEnabled()) {
       r = JGROUPS_MCAST_CONFIG_FILE_NAME;
     } else {
       r = DEFAULT_JGROUPS_TCP_CONFIG;
     }
-    is = ClassPathLoader.getLatest().getResourceAsStream(getClass(), r);
+    ClassLoader contextClassLoader = Thread.currentThread().getContextClassLoader();
+    if (contextClassLoader != null) {
+      is = contextClassLoader.getResourceAsStream(r);
+    }
```
Bill: I see the new code searches first, using the current thread's context `ClassLoader`, and then tries the current class' one. Finally it searches for the resource as a "system resource" (in the classpath).

This seems fine, in the abstract. But I can't tell how it relates to the old code `ClassPathLoader.getLatest().getResourceAsStream()`.

bschuchardt: ClassPathLoader exists to find customer classes in multiple contexts including J2EE servers, Web servers, regular gfsh-launched servers, etc.  Here we only need to find the JGroups configuration file, which will normally be found with ClassLoader.getSystemResourceAsStream(r).  I added the others because you really should look at the thread-context class-loader first and then, why not look at the class-loader for JGroupsMessenger as well (that's where I thought it would be found but I was wrong for the common case).

Bill: thanks @bschuchardt

PR #4509, opened 2019-12-19
Geode 7398 show deployer and deploy time

jinmeiliao:

### For all changes:

### Note:


jinmeiliao:
@jmelchio

geode-core/src/main/java/org/apache/geode/management/internal/configuration/domain/Configuration.java
```diff
@@ -168,80 +164,75 @@ public void toData(DataOutput out) throws IOException {
...
+    cacheXmlContent = DataSerializer.readString(in);
+    propertiesFileName = DataSerializer.readString(in);
+    gemfireProperties = DataSerializer.readProperties(in);
+    HashSet<String> jarNames = DataSerializer.readHashSet(in);
+
+    Version version;
+    try {
+      // after 1.12, we would have version info in the input stream
+      version = Version.fromOrdinal(Version.readOrdinal(in));
+    } catch (Exception e) {
```
demery-pivotal: Is there a gentler way to determine whether there's an ordinal? Some kind of "peek" or "is EOF" test to see if there's more data?

I looked briefly and couldn't find a way, but perhaps I overlooked something.

dschneider-pivotal: DataInput does not have a way. So we would need to cast DataInput to some other class and it is hard to say if it would always be an instance of the other class.

dschneider-pivotal: I think I found a way. Pre 1.12 the jarNames HashSet was never null. So we can use it to determine if we are Pre 1.12 (jarNames !=null) or 1.12 and later (jarNames == null). I will code up this fix.

PR #4513, opened 2019-12-20
GEODE-7609: allow "Squash And Merge" only in GitHub

onichols-pivotal:
DO NOT MERGE yet.  This proposal is still being discussed on the dev list and has not been voted on yet.

This PR changes GitHub settings to make "Squash and merge" the only option for merging a PR.  The other options could be used to bypass our branch protection rules requiring every commit landing on develop to have passed required PR checks.

It also updates the language in the the PR template to relax the mandate to always _initially_ create PRs from a single, squashed commit; instead emphasizing to initially create PRs in whatever way is most amenable to review (such as separating out refactor/rename commits from the actual fix).

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -23,5 +23,5 @@ github:
     merge:   false
     # enable squash button:
     squash:  true
-    # enable rebase button:
-    rebase:  true
+    # disable rebase button:
+    rebase:  false
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -1,0 +1,27 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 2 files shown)


onichols-pivotal:
> Can we change this to default to squash-merge instead?

If GitHub provided such an option, I agree that would be nicer way to solve the problem.  However to the best of my knowledge, GitHub does not currently offer that level of control.  Please refer to https://help.github.com/en/github/administering-a-repository/configuring-commit-squashing-for-pull-requests for details on what is possible to configure.

onichols-pivotal:
This PR was premature given the lack of consensus. I am withdrawing it.

onichols-pivotal:
As this is still being discussed on the dev list, please leave your comments on the dev list, and do not merge this PR unless the is consensus and a successful vote on the dev list.

onichols-pivotal:
Based on dev list discussion, it sounds like the preference is to leave all 3 merge options enabled, therefore this PR is unneeded.

PR #4502, opened 2019-12-18
GEODE-7556: remove membership dependencies on geode-core exeptions

bschuchardt:
Removed use of geode-core exceptions from membership. DistributionImpl now converts membership exceptions into geode-core exceptions where necessary.

Except for MembershipClosedException the new membership exceptions are all checked exceptions. This let me isolate where the exceptions are used and ensure that they're changed into appropriate geode-core exceptions.

MemberShunnedException is now in the membership module instead of the TcpConduit module. It, too, is a checked exception.

CancelCriterion is also removed from use in the membership module. The Stopper in Services.java doesn't need to be a CancelCriterion to function properly.

Several tests had to be modified to handle our stress-test environement.

Thank you for submitting a contribution to Apache Geode.

In order to streamline the review of the contribution we ask you
to ensure the following steps have been taken:

### For all changes:
- [x] Is there a JIRA ticket associated with this PR? Is it referenced in the commit message?

- [x] Has your PR been rebased against the latest commit within the target branch (typically `develop`)?

- [x] Is your initial contribution a single, squashed commit?

- [x] Does `gradlew build` run cleanly?

- [n/a] Have you written or updated unit tests to verify your changes?

- [ ] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)?

### Note:
Please ensure that once the PR is submitted, check Concourse for build issues and
submit an update to your PR as soon as possible. If you need help, please send an
email to dev@geode.apache.org.

PR #4514, opened 2019-12-20
DRAFT Revert "GEODE-7556: remove membership dependencies on geode-core exeptions"

upthewaterspout:
Reverts apache/geode#4502

These changes seem to have broken the types of exceptions that are thrown when a split brain is detected.


upthewaterspout:
Not needed, @bschuchardt fixed the bug this change introduced.

PR #4538, opened 2019-12-27
GEODE-7626: Break dependency on LocalViewMessage in membership

bschuchardt:
LocalViewMessage was a DistributionMessage executed in an executor owned
by ClusterDistributionmanager.  This arrangement was very convoluted
because CDM only had upstream involvement in membership view
installation.

This PR moves view installation into GMSMembership using a
single-threaded executor similar to what CDM used but without
statistics.  Stats for the view installation thread have never been
useful so I have not retained that functionality.

There are already many tests for view installation, so while I've
modified a couple I haven't added any new tests.

### For all changes:

### Note:


geode-core/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSMembership.java
```diff
@@ -2107,4 +2111,31 @@ public boolean isReconnectingDS() {
 
   }
 
+
+  static class ViewExecutorBlockHandler implements RejectedExecutionHandler {
```
echobravopapa: Wondering if the access level ought to be specified...

geode-core/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSMembership.java
```diff
@@ -652,6 +660,10 @@ public GMSMembership(MembershipListener<ID> listener, MessageListener<ID> messag
     this.listener = listener;
     this.messageListener = messageListener;
     this.gmsManager = new ManagerImpl();
+    LinkedBlockingQueue<Runnable> feed = new LinkedBlockingQueue<>();
+    ThreadFactory threadFactory = new LoggingThreadFactory("Geode View Processor");
+    this.viewExecutor = new ThreadPoolExecutor(1, 1, 30,
+        TimeUnit.SECONDS, feed, threadFactory, new ViewExecutorBlockHandler(feed));
```
upthewaterspout: Could this have used LoggingExecutors?

upthewaterspout: Since this is an unbounded queue, do you really need the ViewExecutorBlockHandler? Could you just use the AbortPolicy in the JDK?

bschuchardt: I was trying to capture the config of the old executor but I think you're right, Dan.  I'll get rid of that class and use a simpler executor.

PR #4546, opened 2019-12-30
GEODE-7579: introduce RegionAware

jinmeiliao:
* Index configuration is ReqionAware, but not groupable
* index's group depends on what region it's defined with

### For all changes:

### Note:

### DIFF
--- a/geode-assembly/src/distributedTest/java/org/apache/geode/management/internal/rest/ListIndexOnPartitionRegionTest.java
+++ b/geode-assembly/src/distributedTest/java/org/apache/geode/management/internal/rest/ListIndexOnPartitionRegionTest.java
@@ -57,6 +57,7 @@ public class ListIndexOnPartitionRegionTest {
         .setHostAddress("localhost", locator.getHttpPort())
         .build();
 
+    // set up the same region in 3 different groups, one is a proxy region, but with the same name
     Region config = new Region();
     config.setName("testRegion");
     config.setType(RegionType.PARTITION);
@@ -121,9 +122,12 @@ public class ListIndexOnPartitionRegionTest {
       assertThat(group4.findRegionConfiguration("testRegion").getIndexes()).isEmpty();
     });
 
-    // try to create index again
+    // try to create index again will throw an error since index already exists on some groups
     assertThatThrownBy(() -> cms.create(index))
         .hasMessageContaining("Index 'index' already exists");
+
+    // demonstrate that the only way to put the index on the newly created region on group4 is
+    // to delete the index and then recreate the index again.
 
     // delete the index
     ClusterManagementRealizationResult deleteResult = cms.delete(index);
--- a/geode-assembly/src/distributedTest/java/org/apache/geode/management/internal/rest/ListIndexOnPartitionRegionTest.java
+++ b/geode-assembly/src/distributedTest/java/org/apache/geode/management/internal/rest/ListIndexOnPartitionRegionTest.java
@@ -1,0 +1,107 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package org.apache.geode.management.internal.rest;
+
+import static org.apache.geode.test.junit.assertions.ClusterManagementListResultAssert.assertManagementListResult;
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Test;
+
+import org.apache.geode.cache.configuration.CacheConfig;
+import org.apache.geode.cache.configuration.RegionConfig;
+import org.apache.geode.distributed.internal.InternalConfigurationPersistenceService;
+import org.apache.geode.management.api.ClusterManagementListResult;
+import org.apache.geode.management.api.ClusterManagementService;
+import org.apache.geode.management.client.ClusterManagementServiceBuilder;
+import org.apache.geode.management.configuration.Index;
+import org.apache.geode.management.configuration.IndexType;
+import org.apache.geode.management.configuration.Region;
+import org.apache.geode.management.configuration.RegionType;
+import org.apache.geode.management.runtime.IndexInfo;
+import org.apache.geode.test.dunit.rules.ClusterStartupRule;
+import org.apache.geode.test.dunit.rules.MemberVM;
+import org.apache.geode.test.junit.rules.MemberStarterRule;
+
+public class ListIndexOnPartitionRegionTest {
+  @ClassRule
+  public static ClusterStartupRule lsRule = new ClusterStartupRule();
+
+  private static ClusterManagementService cms;
+  private static MemberVM locator;
+
+  @BeforeClass
+  public static void beforeClass() {
+    locator = lsRule.startLocatorVM(0, MemberStarterRule::withHttpService);
+    lsRule.startServerVM(1, "group1", locator.getPort());
+    lsRule.startServerVM(2, "group2", locator.getPort());
+    lsRule.startServerVM(3, "group3", locator.getPort());
+
+    cms = ClusterManagementServiceBuilder.buildWithHostAddress()
+        .setHostAddress("localhost", locator.getHttpPort())
+        .build();
+
+    Region config = new Region();
+    config.setName("testRegion");
+    config.setType(RegionType.PARTITION);
+    config.setGroup("group1");
+    cms.create(config);
+
+    config.setType(RegionType.PARTITION);
+    config.setGroup("group2");
+    cms.create(config);
+
+    config.setType(RegionType.PARTITION_PROXY);
+    config.setGroup("group3");
+    cms.create(config);
+
+    locator.waitUntilRegionIsReadyOnExactlyThisManyServers("/testRegion", 3);
+  }
+
+  @Test
+  public void createAndListIndex() {
+    // create index on group2
+    Index index = new Index();
+    index.setName("index");
+    index.setExpression("id");
+    index.setRegionPath("/testRegion");
+    index.setIndexType(IndexType.KEY);
+    cms.create(index);
+
+    // verify that index configuration is saved on all groups
+    locator.invoke(() -> {
+      InternalConfigurationPersistenceService cps =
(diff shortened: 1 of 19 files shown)


jinmeiliao:
@jmelchio

geode-core/src/main/java/org/apache/geode/management/internal/api/LocatorClusterManagementService.java
```diff
@@ -185,10 +202,17 @@ public LocatorClusterManagementService(InternalConfigurationPersistenceService p
...
+      String groupName = groups.iterator().next();
+      configurationManager.add(config, groupName);
+      result.setStatus(StatusCode.OK,
+          "Successfully updated configuration for " + groupName + ".");
+    } else {
+      for (String groupName : groups) {
+        configurationManager.add(config, groupName);
+      }
+      result.setStatus(StatusCode.OK, "Successfully updated configuration");
+    }
```
jmelchio: It appears the else branch covers would work for a size of `1` as well.

jinmeiliao: yes, all elements nested under <region> should be RegionAware, this include index, lucene index and any other custom elements that will be nested under "region".

geode-core/src/main/java/org/apache/geode/management/internal/api/LocatorClusterManagementService.java
```diff
@@ -157,20 +156,38 @@ public LocatorClusterManagementService(InternalConfigurationPersistenceService p
...
     }
 
-    Set<DistributedMember> targetedMembers = memberValidator.findServers(group);
+    // find the targeted members
+    Set<String> groups = new HashSet<>();
+    Set<DistributedMember> targetedMembers;
+    if (config instanceof RegionAware) {
+      String regionName = ((RegionAware) config).getRegionName();
+      groups = memberValidator.findGroups(regionName);
+      if (groups.size() == 0) {
```
dschneider-pivotal: Better would be "if (groups.isEmpty())"

geode-core/src/main/java/org/apache/geode/management/internal/api/LocatorClusterManagementService.java
```diff
@@ -185,10 +202,17 @@ public LocatorClusterManagementService(InternalConfigurationPersistenceService p
...
+    if (groups.size() == 1) {
+      String groupName = groups.iterator().next();
+      configurationManager.add(config, groupName);
+      result.setStatus(StatusCode.OK,
+          "Successfully updated configuration for " + groupName + ".");
+    } else {
+      for (String groupName : groups) {
+        configurationManager.add(config, groupName);
+      }
+      result.setStatus(StatusCode.OK, "Successfully updated configuration");
```
dschneider-pivotal: Add a "." to the end of the message

dschneider-pivotal: I think this message should list all the groups it updated. Then you could get rid if the if size == 1 and just have all the code loop over the groups (even when it is just a single group).

geode-core/src/main/java/org/apache/geode/management/internal/configuration/validators/MemberValidator.java
```diff
@@ -52,16 +55,27 @@ public void validateCreate(AbstractConfiguration config, CacheConfigurationManag
...
           config.getClass().getSimpleName() + " '" + config.getId()
               + "' already exists in group " + configGroup);
     }
 
+    // if other group and this new group has common members, then throw exception
+    String[] groups = existingElementsAndTheirGroups.keySet().toArray(new String[0]);
     Set<DistributedMember> membersOfExistingGroups =
-        findServers(existingElementsAndTheirGroups.keySet().toArray(new String[0]));
-    Set<DistributedMember> membersOfNewGroup = findServers(config.getGroup());
+        findMembers(false, groups);
```
dschneider-pivotal: I liked the old findServers(groups) instead of findMembers(false, groups). Can we bring it back and have it call findMembers(false, groups)?

geode-core/src/main/java/org/apache/geode/management/internal/api/LocatorClusterManagementService.java
```diff
@@ -185,10 +202,11 @@ public LocatorClusterManagementService(InternalConfigurationPersistenceService p
...
 
     // persist configuration in cache config
-
-    configurationManager.add(config, groupName);
+    for (String groupName : groups) {
+      configurationManager.add(config, groupName);
+    }
     result.setStatus(StatusCode.OK,
-        "Successfully updated configuration for " + groupName + ".");
+        "Successfully updated configuration for " + StringUtils.join(groups, ", ") + ".");
```
demery-pivotal: Prefer `String.join()`, which is built into Java.

geode-management/src/main/java/org/apache/geode/management/api/ClusterManagementRealizationResult.java
```diff
@@ -55,4 +58,9 @@ public void addMemberStatus(RealizationResult result) {
   public List<RealizationResult> getMemberStatuses() {
     return memberStatuses;
   }
+
+  @Override
+  public String toString() {
+    return super.toString() + " " + StringUtils.join(memberStatuses, ";");
```
demery-pivotal: Prefer `String.join()`, which is built into Java.

geode-core/src/test/java/org/apache/geode/management/internal/api/LocatorClusterManagementServiceTest.java
```diff
@@ -132,7 +132,8 @@ public void create_persistenceIsNull() {
 
   @Test
   public void create_validatorIsCalledCorrectly() {
-    doReturn(Collections.emptySet()).when(memberValidator).findServers(anyString());
+    doReturn(Collections.emptySet()).when(memberValidator);
+    memberValidator.findMembers(eq(false), new String[] {anyString()});
```
demery-pivotal: This is very puzzling use of Mockito. Are these two lines trying to set up `findMembers()` to return an empty set? I haven't experimented with this, but I don't this does the right thing. Without a function call after the `when()`, I don't think the `doReturn()` does anything. I don't think the call to `findMembers()` also does anything. And I don't think the argument matchers (`eq()` and `anyString()`) are doing anything.

I think the net result of all of this is that you have not mocked `findMembers()` at all. So mockito, by default, returns an empty collection when the method is called.

And why this odd mocking syntax? Is it because of the varargs? Perhaps it would be better to create a non-varargs version or two—one that takes a single string, and another that takes an iterable or collection.

geode-core/src/main/java/org/apache/geode/management/internal/api/LocatorClusterManagementService.java
```diff
@@ -157,20 +156,38 @@ public LocatorClusterManagementService(InternalConfigurationPersistenceService p
...
+    // find the targeted members
+    Set<String> groups = new HashSet<>();
+    Set<DistributedMember> targetedMembers;
+    if (config instanceof RegionAware) {
+      String regionName = ((RegionAware) config).getRegionName();
+      groups = memberValidator.findGroups(regionName);
+      if (groups.isEmpty()) {
+        raise(StatusCode.ENTITY_NOT_FOUND, "Region provided does not exist: " + regionName);
+      }
+      targetedMembers = memberValidator.findServers(groups.toArray(new String[0]));
```
demery-pivotal: This varargs method forces too many other parts of the code to convert collections into arrays. Consider adding another version or two of `findServers()`—one that takes a single string, and another that takes an iterator or collection. Then our production code and test code won't have to jump through so many hoops to do something so simple.

geode-management/src/main/java/org/apache/geode/management/configuration/RegionAware.java
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package org.apache.geode.management.configuration;
+
+public interface RegionAware {
```
demery-pivotal: Given how this interface is used, it seems to me that implementers are not only _aware,_ of a region, but _scoped_ to a region. Is that the intention? If so, `RegionScoped` would be a better name.

geode-core/src/main/java/org/apache/geode/management/internal/api/LocatorClusterManagementService.java
```diff
@@ -305,7 +323,11 @@ public LocatorClusterManagementService(InternalConfigurationPersistenceService p
             }
           });
         }
-        resultList.addAll(list);
+        list.stream().forEach(t -> {
+          if (!resultList.contains(t)) {
+            resultList.add(t);
+          }
+        });
```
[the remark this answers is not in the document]
jinmeiliao: done

geode-assembly/src/distributedTest/java/org/apache/geode/management/internal/rest/ListIndexOnPartitionRegionTest.java
```diff
@@ -0,0 +1,154 @@
...
+    cms.create(config);
+
+    config.setType(RegionType.PARTITION_PROXY);
+    config.setGroup("group3");
+    cms.create(config);
+
+    locator.waitUntilRegionIsReadyOnExactlyThisManyServers("/testRegion", 3);
+  }
+
+  @Test
```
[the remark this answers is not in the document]
jinmeiliao: added more comments in the test to show the intention of the test

geode-core/src/main/java/org/apache/geode/management/internal/api/LocatorClusterManagementService.java
jinmeiliao: done

geode-management/src/main/java/org/apache/geode/management/api/ClusterManagementRealizationResult.java
jinmeiliao: In this case, StringUtils is better since it takes Iterable<?> rather than just Iterable<CharSequence>

geode-core/src/test/java/org/apache/geode/management/internal/api/LocatorClusterManagementServiceTest.java
jinmeiliao: those should be one line, I am not sure since when they become 2 lines of code.

geode-core/src/main/java/org/apache/geode/management/internal/api/LocatorClusterManagementService.java
jinmeiliao: this makes passing in one single string or nothing easier. I updated the tests to not make the unnecessary conversions if passing in a single string or nothing.

geode-management/src/main/java/org/apache/geode/management/configuration/RegionAware.java
jinmeiliao: You are right. I was struggling with a better name.

jinmeiliao:
the only part that needs to consider regionAware is in the LocatorClusterManagementService (the main pipeline). All other changes are test changes.

PR #4551, opened 2019-12-30
[DRAFT] Extract a superclass of InternalDistributedMember

upthewaterspout:
Extracting a superclass of InternalDistributedMember
    
We need a concrete MemberIdentifier implementation in geode-membership
to test membership.
    
Extracting a superclass of InternalDistributedMember,
MemberIdentifierImpl, which contains pretty much all of the
InternalDistributedMember logic. InternalDistributedMember now just
extends that class and implements DistributedMember.

Removing uses of SocketCreator from MemberIdentifierImpl.


upthewaterspout:
Going to redo this work a different way.

PR #4563, opened 2020-01-03
GEODE-7648: elide hostname resolution if no bindAddress given

Bill:
[GEODE-7648](https://issues.apache.org/jira/browse/GEODE-7648)

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSUtil.java
+++ b/geode-core/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSUtil.java
@@ -135,18 +135,20 @@ public class GMSUtil {
 
       final InetSocketAddress isa = new InetSocketAddress(host, port);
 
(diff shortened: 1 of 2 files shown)


metatype:
I don't think I follow the bind address logic.  Do you know why we throw and exception in one case of unresolvable host name but not the other?  IMHO it seems like we should expect DNS to be unreliable and not assume misconfiguration as the only cause of unresolv-ability.

pivotal-jbarrett:
I believe that local host name check is pretty old behavior. However, it could bite us in any orchestrated deployment where DNS setup is delayed until after a service is running. I’d suggest addressing that issue in another ticket and letting this one through as is to correct the bug.

Bill:
so @metatype you'd like to see the whole `if (checkLoopback)` block removed entirely right? Yeah, I don't have insight into why we want to do that (conditional) prechecking

PR #4570, opened 2020-01-07
GEODE-7655:Restore context with distinct state after a recursion step.

nabarunnag:
* After a recursion step within Compiled Select's evaluate statement, the distinct flag was lost for the outer query.
	* The solution was to reset the distinct state back using a temp variable to store the initial distinct flag value.
	* Also, in this diff a else part section was missing converting PDX strings to normal strings for comparison of distinct.
	* When we place this else section converting the PDX string to string for distinct comparison, we don't need to perform this check in the 'element' query function.

Removing the JPMC related configurations

### For all changes:

### Note:

### DIFF
--- a/geode-dunit/src/main/java/org/apache/geode/management/internal/cli/commands/NestedQueryClassCastExceptionFailureDUnitTest.java
+++ b/geode-dunit/src/main/java/org/apache/geode/management/internal/cli/commands/NestedQueryClassCastExceptionFailureDUnitTest.java
@@ -1,0 +1,79 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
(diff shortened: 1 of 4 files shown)


geode-dunit/src/main/java/org/apache/geode/management/internal/cli/commands/NestedQueryClassCastExceptionFailureDUnitTest.java
```diff
@@ -0,0 +1,79 @@
...
+          "org.apache.geode.management.internal.cli.commands.*");
+      Cache cache = new CacheFactory(properties)
+          .setPdxSerializer(new ReflectionBasedAutoSerializer(
+              "org.apache.geode.management.internal.cli.commands.*"))
+          .create();
+
+      CacheServer cacheServer = cache.addCacheServer();
+      cacheServer.setPort(0);
+      cacheServer.start();
+      Region region = cache.createRegionFactory(RegionShortcut.REPLICATE).create("product");
```
jujoramos: Nitpick: this chunk of code can be replaced with something "easier" to read:

```
  @Rule
  public ServerStarterRule server =
      new ServerStarterRule()
          .withRegion(RegionShortcut.REPLICATE, "product")
          .withProperty(SERIALIZABLE_OBJECT_FILTER, "org.apache.geode.management.internal.cli.commands.*")
          .withPdxSerializer(new ReflectionBasedAutoSerializer("org.apache.geode.management.internal.cli.commands.*"))
          .withAutoStart();

  @Test
  public void classCastExceptionShouldNotBeThrownWhileExecutionNestedQueries() throws Exception {
      Cache cache = server.getCache();
      Region<Integer, Product> region = cache.getRegion("product");
      (...)
}
```

nabarunnag: I agree, but I personally prefer using creating the setup as we say in our documentation. Previously I had issues with Rules. I will try to refactor the tests better from now on to make it more readable

PR #4580, opened 2020-01-09
GEODE-5407: Fix JMXMBeanReconnectDUnitTest

kirklund:
```
GEODE-5407: Fix JMXMBeanReconnectDUnitTest
```
```
* Rewrite test with standard Geode and JMX APIs
* Use CountDownLatch instead of Blackboard
* Use SharedErrorCollector for remote listener errors
* Always consistently use Awaitility before checking GemFire MXBeans
    
Co-authored-by: Ivan Godwin <igodwin@pivotal.io>
```

### DIFF
--- a/geode-core/src/distributedTest/java/org/apache/geode/management/JMXMBeanReconnectDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/management/JMXMBeanReconnectDUnitTest.java
@@ -12,22 +12,40 @@
  * or implied. See the License for the specific language governing permissions and limitations under
  * the License.
  */
-
 package org.apache.geode.management;
 
-import static java.util.stream.Collectors.toList;
-import static org.apache.geode.management.ManagementService.getExistingManagementService;
+import static java.lang.management.ManagementFactory.getPlatformMBeanServer;
+import static java.util.Arrays.asList;
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
+import static javax.management.ObjectName.getInstance;
+import static org.apache.geode.cache.Region.SEPARATOR;
+import static org.apache.geode.distributed.ConfigurationProperties.JMX_MANAGER;
+import static org.apache.geode.distributed.ConfigurationProperties.JMX_MANAGER_PORT;
+import static org.apache.geode.distributed.ConfigurationProperties.JMX_MANAGER_START;
+import static org.apache.geode.distributed.ConfigurationProperties.LOCATORS;
+import static org.apache.geode.distributed.ConfigurationProperties.MAX_WAIT_TIME_RECONNECT;
+import static org.apache.geode.distributed.ConfigurationProperties.MEMBER_TIMEOUT;
+import static org.apache.geode.distributed.internal.InternalDistributedSystem.addReconnectListener;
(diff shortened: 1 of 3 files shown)


geode-core/src/main/java/org/apache/geode/distributed/internal/InternalDistributedSystem.java
```diff
@@ -2212,15 +2212,20 @@ public Properties getSecurityProperties() {
      * @param oldSystem the old DS, which is in a partially disconnected state and cannot be used
      *        for messaging
      */
-    void reconnecting(InternalDistributedSystem oldSystem);
+    default void reconnecting(InternalDistributedSystem oldSystem) {
```
igodwin: Is this change to default still required? It looks like we are overriding for each implementation of ReconnectListener.

kirklund: It's not required but I think it's a good addition. This is an internal listener, so we're not impacting any users. Adding default just means that implementers (tests) can implement just the method(s) it needs to instead of including empty impl methods.

geode-core/src/distributedTest/java/org/apache/geode/management/JMXMBeanReconnectDUnitTest.java
```diff
@@ -156,47 +341,51 @@ public void testLocalBeans_MaintainLocatorAndCrashServer() {
...
-      intermediateL2Beans.addAll(jmxConToLocator2.getGemfireFederatedBeans());
-
-      assertThat(intermediateL2Beans)
-          .containsExactlyElementsOf(expectedIntermediateBeanList)
-          .hasSameSizeAs(expectedIntermediateBeanList);
+  public void locatorMXBeansOnOtherLocatorAreRestoredAfterCrashedLocatorReturns() {
+    locator1VM.invoke(() -> {
+      RECONNECT_LATCH.set(new CountDownLatch(1));
+
+      addReconnectListener(new ReconnectListener() {
```
kirklund: @igodwin Here's one of the implementations of ReconnectListener in the new test which doesn't implement both methods. If we remove `default`, then this impl would have to provide an empty impl of `onReconnect`.

geode-core/src/distributedTest/java/org/apache/geode/management/JMXMBeanReconnectDUnitTest.java
```diff
@@ -36,117 +54,284 @@
...
+        assertThat(getPlatformMBeanServer().queryNames(getInstance("GemFire:*"), null))
+            .containsAll(expectedMXBeansOnServer());
+      });
+      return getPlatformMBeanServer().queryNames(getInstance("GemFire:*"), null);
+    });
 
-    gfsh.connectAndVerify(locator1);
-    gfsh.executeAndAssertThat("create region --type=REPLICATE --name=" + REGION_PATH
-        + " --enable-statistics=true").statusIsSuccess();
+    mxbeansOnLocator1 = locator1VM.invoke(() -> {
```
mhansonp: There are a lot of collapsible lambdas. I presume you intentionally did not collapse them for readability.

PR #4587, opened 2020-01-12
GEODE-7690: make it less difficult to run benchmarks against a PR

onichols-pivotal:
To run Benchmarks against a personal fork, it's not immediately obvious where to override the repo default of "apache/geode"

This PR adds variables for this at the top of run_benchmarks.sh so they can be easily set in your environment without having to figure out how to update the very long run_against_baseline commandline.

### DIFF
--- a/ci/scripts/run_benchmarks.sh
+++ b/ci/scripts/run_benchmarks.sh
@@ -37,6 +37,9 @@ RESULTS_DIR=$(pwd)/results/benchmarks-${CLUSTER_TAG}
 CLUSTER_COUNT=4
 BENCHMARKS_BRANCH=${BENCHMARKS_BRANCH:-develop}
 
+GEODE_REPO=${GEODE_REPO:-apache/geode}
+BASELINE_REPO=${BASELINE_REPO:-${GEODE_REPO}}
+
 pushd geode
 GEODE_SHA=$(git rev-parse --verify HEAD)
 popd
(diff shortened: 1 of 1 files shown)


ci/scripts/run_benchmarks.sh
```diff
@@ -37,6 +37,9 @@ RESULTS_DIR=$(pwd)/results/benchmarks-${CLUSTER_TAG}
 CLUSTER_COUNT=4
 BENCHMARKS_BRANCH=${BENCHMARKS_BRANCH:-develop}
 
+GEODE_REPO=${GEODE_REPO:-apache/geode}
```
pivotal-jbarrett: Are changes needed in the rest of the CI to get the `GEODE_REPO` passed through?

onichols-pivotal: see my change on line 80

onichols-pivotal: for the public Geode develop and release pipelines, the default initialized on line 40 is sufficient.  As described in the PR, if you are trying to do any testing against a PR, fork, etc prior to merging to develop, the hardcoded defaults in run_against_baseline made it impossible without hacking some scripts.   This PR eliminates the need for hacking, by improving the default for GEODE_REPO, and further allowing it to be overridden externally if the default computed on line 40 doesn't work for some reason.

PR #4588, opened 2020-01-13
GEODE-7691: Instance of parameter checked in equals method.

nabarunnag:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/cache/query/internal/CqEntry.java
+++ b/geode-core/src/main/java/org/apache/geode/cache/query/internal/CqEntry.java
@@ -82,7 +82,7 @@ public class CqEntry implements DataSerializableFixedID {
 
   @Override
   public boolean equals(Object other) {
-    if (other == null) {
+    if (!(other instanceof CqEntry)) {
       return false;
     }
     return this.key.equals(((CqEntry) other).getKey());
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
@@ -9665,8 +9665,7 @@ public class PartitionedRegion extends LocalRegion
 
     @Override
     public boolean equals(Object other) {
-      // TODO: equals should always check the class of other
-      if (other == null) {
+      if (!(other instanceof IndexTask)) {
         return false;
       }
       IndexTask otherIndexTask = (IndexTask) other;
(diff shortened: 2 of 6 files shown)


geode-core/src/main/java/org/apache/geode/cache/query/internal/CqEntry.java
```diff
@@ -82,6 +82,9 @@ public int hashCode() {
 
   @Override
   public boolean equals(Object other) {
+    if (!(other instanceof CqEntry)) {
+      return false;
+    }
     if (other == null) {
```
jhuynh1: I think this if check is no longer needed with the addition of the instanceof check

geode-core/src/main/java/org/apache/geode/cache/query/internal/index/DummyQRegion.java
```diff
@@ -68,6 +68,9 @@ public DummyQRegion(Region region) {
 
   @Override
   public boolean equals(Object o) { // for findbugs
+    if (!(o instanceof DummyQRegion)) {
```
jhuynh1: Not sure if we want this check here... the super implementation just checks values.  So a DummyQRegion could in theory be checked against a base QRegion and still return true.. in this case it would return false.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
```diff
@@ -9665,7 +9665,9 @@ public String toString() {
 
     @Override
     public boolean equals(Object other) {
-      // TODO: equals should always check the class of other
+      if (!(other instanceof IndexTask)) {
+        return false;
+      }
       if (other == null) {
```
jhuynh1: same, this can probably be removed since the instanceof check will return early if other is null

geode-core/src/main/java/org/apache/geode/cache/query/internal/CqEntry.java
nabarunnag: my bad, I missed that!!

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
nabarunnag: my bad, I missed that!!

geode-core/src/main/java/org/apache/geode/cache/query/internal/index/DummyQRegion.java
nabarunnag: My bad again. Modified the superclass to do an instance check.

PR #4596, opened 2020-01-15
GEODE-7699: use ConfigurationInfo to wrap the get result

jinmeiliao:

### For all changes:

### Note:

### DIFF
--- a/geode-management/src/main/java/org/apache/geode/management/api/ConfigurationInfo.java
+++ b/geode-management/src/main/java/org/apache/geode/management/api/ConfigurationInfo.java
@@ -28,17 +28,24 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import org.apache.geode.management.configuration.AbstractConfiguration;
 import org.apache.geode.management.runtime.RuntimeInfo;
 
+/**
+ * This holds the configuration of a particular entity in the cluster. It includes the id of the
+ * entity and a list of configuration per group on the cluster
+ *
+ * @param <T> the config type
+ * @param <R> the runtimeInfo type
+ */
 public class ConfigurationInfo<T extends AbstractConfiguration<R>, R extends RuntimeInfo> {
   private String id;
   @JsonInclude
   @JsonProperty
-  private List<ConfigurationResult<T, R>> groupResults = new ArrayList<>();
+  private List<ConfigurationResult<T, R>> configurationByGroup = new ArrayList<>();
 
   public ConfigurationInfo() {}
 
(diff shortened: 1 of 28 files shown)


jinmeiliao:
@jmelchio

geode-management/src/main/java/org/apache/geode/management/api/ConfigurationInfo.java
```diff
@@ -0,0 +1,106 @@
...
+import org.apache.geode.management.runtime.RuntimeInfo;
+
+/**
+ * This holds the configuration of a particular entity in the cluster. It includes the id of the
+ * entity and a list of configuration per group on the cluster
+ *
+ * @param <T> the config type
+ * @param <R> the runtimeInfo type
+ */
+public class ConfigurationInfo<T extends AbstractConfiguration<R>, R extends RuntimeInfo> {
```
dschneider-pivotal: On the topic of naming: given how this class is currently used and the other classes names it seems like this class should include "Result" in it. It is only used as part of the result of a get. I think we should consider also using it in the  result of list but in either case this class is a "result".
The primary thing this class has (compared to other results) is the identity of the thing it describes. So would a good name be "IdentifiableResult"?

geode-management/src/main/java/org/apache/geode/management/api/ConfigurationInfo.java
```diff
@@ -0,0 +1,106 @@
...
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import org.apache.geode.management.configuration.AbstractConfiguration;
+import org.apache.geode.management.runtime.RuntimeInfo;
+
+/**
+ * This holds the configuration of a particular entity in the cluster. It includes the id of the
+ * entity and a list of configuration per group on the cluster
```
dschneider-pivotal: I think this comment should also describe how RuntimeInfo per member is part of this class

geode-management/src/main/java/org/apache/geode/management/api/ConfigurationResult.java
```diff
@@ -46,7 +51,12 @@ public ConfigurationResult() {}
    * for internal use only
    */
   public ConfigurationResult(T configuration) {
+    this(configuration, emptyList());
+  }
+
+  public ConfigurationResult(T configuration, List<R> runtimeInfo) {
```
dschneider-pivotal: It seems like a better name for this class would be "GroupResult". I like that you updated the class comment to say it is per group. It seems like this class should include a group name so that it would be self describing.

PR #4598, opened 2020-01-16
GEODE-7706: throw correct TransactionDataRebalancedException.

pivotal-eshu:


geode-core/src/test/java/org/apache/geode/internal/cache/PartitionedRegionTest.java
```diff
@@ -439,6 +443,66 @@ public void generatePRIdShouldNotThrowNumberFormatExceptionIfAnErrorOccursWhileR
...
+    doThrow(new ForceReattemptException("")).when(dataStore)
+        .getInitializedBucketWithKnownPrimaryForId(key, 1);
+    doReturn(mock(InternalDistributedMember.class)).when(spyPartitionedRegion).createBucket(1, 0,
+        null);
+
+    Throwable caughtException =
+        catchThrowable(() -> spyPartitionedRegion.getDataRegionForWrite(keyInfo));
+
+    assertThat(caughtException).isInstanceOf(TransactionDataRebalancedException.class)
+        .hasMessage("Transactional data moved, due to rebalancing.");
```
DonalEvans: Is it worth extracting this string to a constant in the `PartitionedRegion` class so that it can be consistently referred to?

geode-core/src/test/java/org/apache/geode/internal/cache/PartitionedRegionTest.java
```diff
@@ -439,6 +443,66 @@ public void generatePRIdShouldNotThrowNumberFormatExceptionIfAnErrorOccursWhileR
...
+
+    KeyInfo keyInfo = mock(KeyInfo.class);
+    when(keyInfo.getBucketId()).thenReturn(1);
+    doReturn(null).when(spyPartitionedRegion).getDataStore();
+
+    Throwable caughtException =
+        catchThrowable(() -> spyPartitionedRegion.getDataRegionForWrite(keyInfo));
+
+    assertThat(caughtException).isInstanceOf(TransactionException.class).hasMessage(
+        "PartitionedRegion Transactions cannot execute on nodes with local max memory zero");
```
jujoramos: These assertions within the tests can be further simplified by using something like the following:

```
    assertThatThrownBy(() -> spyPartitionedRegion.getDataRegionForWrite(keyInfo))
        .isInstanceOf(TransactionException.class)
        .hasMessage("PartitionedRegion Transactions cannot execute on nodes with local max memory zero");
```

geode-core/src/test/java/org/apache/geode/internal/cache/PartitionedRegionTest.java
pivotal-eshu: Good idea, will make the change next time.

PR #4601, opened 2020-01-17
GEODE-7715: Introduce Transport and Config interfaces to ClusterManagementServiceBuilder

yozaner1324:
This introduces several new interfaces to the `ClusterManagementService`. These are:

- `ClusterManagmentServiceTransport` which represents the message exchange between client and endpoint.
- `ClusterManagementServiceConnectionConfig` which is a data object and represents various properties required to establish a connection.

### For all changes:

### Note:


jdeppe-pivotal:
@jmelchio for your review.

geode-core/src/main/java/org/apache/geode/management/internal/api/GeodeClusterManagementServiceConnectionConfig.java
```diff
@@ -44,62 +48,105 @@
...
 
+/**
+ * Concrete implementation of {@link ClusterManagementServiceConnectionConfig} which can be used
+ * to derive most (if not all) of the connection properties from an existing {@link Cache} or
+ * {@link ClientCache}.
+ *
+ * @see ClusterManagementServiceBuilder
+ */
+@Experimental
+public class GeodeClusterManagementServiceConnectionConfig
```
jinmeiliao: this pretty much holds the same stuff as BasicClusterManagementServiceConnectionConfig. Is there anyway to make this delegate to that or inherit that?

geode-assembly/src/acceptanceTest/resources/ManagementClientCreateRegion.java
```diff
@@ -31,14 +31,15 @@ public static void main(String[] args) throws Exception {
     if (useSsl) {
       // The default SSLContext will pull in all necessary javax.net.ssl properties
       cms =
-          ClusterManagementServiceBuilder.buildWithHostAddress()
-              .setHostAddress("localhost", httpPort)
-              .setSslContext(SSLContext.getDefault()).build();
+          new ClusterManagementServiceBuilder().setConnectionConfig(
+              new BasicClusterManagementServiceConnectionConfig("localhost", port)
+                  .setSslContext(SSLContext.getDefault()))
+              .buildWithHostAddress();
```
jinmeiliao: shouldn't this be just "build()"?

geode-assembly/src/distributedTest/java/org/apache/geode/management/internal/rest/DeployToMultiGroupDUnitTest.java
```diff
@@ -64,10 +65,10 @@ public static void beforeClass() throws Exception {
     server1 = cluster.startServerVM(1, "group1", locator.getPort());
     server2 = cluster.startServerVM(2, "group2", locator.getPort());
 
-    client =
-        ClusterManagementServiceBuilder.buildWithHostAddress()
-            .setHostAddress("localhost", locator.getHttpPort())
-            .build();
+    client = new ClusterManagementServiceBuilder().setConnectionConfig(
+        new BasicClusterManagementServiceConnectionConfig("localhost", locator.getHttpPort()))
```
jinmeiliao: "BasicClusterManagementServiceConnectionConfig" seems like an awful long name? Any thing shorter and sweeter? Preferably not something "Basic" :-).

Maybe just ConnectionConfig and CacheConnectionConfig

geode-management/src/main/java/org/apache/geode/management/api/ClusterManagementServiceTransport.java
```diff
@@ -0,0 +1,104 @@
...
+  /**
+   * Submit a message for a specific command which returns a list result of operations in progress.
+   * This supports the {@link ClusterManagementService#list(ClusterManagementOperation)} command.
+   *
+   * @param responseType the expected response class
+   * @param <A> operation of type {@link ClusterManagementOperation}
+   * @return {@link ClusterManagementListResult}
+   */
+  <A extends ClusterManagementOperation<V>, V extends OperationResult> ClusterManagementListOperationsResult<V> submitMessageForListOperation(
+      A opType, Class<? extends ClusterManagementListOperationsResult> responseType);
```
demery-pivotal: Is there a strong reason not to use `Class<? extends ClusterManagementListOperationsResult<V>` for `responseType`? If not, please use that.

geode-management/src/test/java/org/apache/geode/management/client/ClusterManagementServiceBuilderTest.java
```diff
@@ -0,0 +1,85 @@
...
+  @BeforeClass
+  public static void setup() {
+    connectionConfig = new BasicClusterManagementServiceConnectionConfig("localhost", 7777);
+  }
+
+  @Test
+  public void buildWithTransportOnlyHavingConnectionConfig() {
+    new ClusterManagementServiceBuilder().setTransport(
+        new RestTemplateClusterManagementServiceTransport(connectionConfig)).build();
+  }
```
demery-pivotal: These tests seem to verify only that the builder doesn't throw. Surely there's something more test here.

geode-core/src/main/java/org/apache/geode/management/internal/api/GeodeClusterManagementServiceConnectionConfig.java
yozaner1324: It now delegates.

geode-assembly/src/acceptanceTest/resources/ManagementClientCreateRegion.java
yozaner1324: You're right, that file had several errors that I didn't catch because it wasn't being compiled. I fixed this and the other problems.

geode-assembly/src/distributedTest/java/org/apache/geode/management/internal/rest/DeployToMultiGroupDUnitTest.java
yozaner1324: It does seem pretty cumbersome. I changed the names to be shorter.

jinmeiliao: Maybe instead of ConnectionConfigImpl which implies it's an internal implementation, change it to "BaseConnectionConfig"?

geode-management/src/main/java/org/apache/geode/management/api/ClusterManagementServiceTransport.java
yozaner1324: We can't do that because of type erasure. To call that method we would have to do something like "ClusterManagementListOperationsResult<V>.class" which isn't allowed.

geode-management/src/test/java/org/apache/geode/management/client/ClusterManagementServiceBuilderTest.java
yozaner1324: You're totally right, I'm updating it to also test that the CMS client that gets built is able to connect properly.

geode-assembly/src/distributedTest/java/org/apache/geode/management/internal/rest/DeployToMultiGroupDUnitTest.java
yozaner1324: Done

PR #4603, opened 2020-01-17
GEODE-7714: add disabling stats-archive-file

mkevo:

### For all changes:

### Note:

### DIFF
--- a/geode-docs/managing/statistics/setting_up_statistics.html.md.erb
+++ b/geode-docs/managing/statistics/setting_up_statistics.html.md.erb
@@ -32,7 +32,7 @@ gfsh>start locator --name=l1 --enable-cluster-configuration=true
 gfsh>alter runtime --enable-statistics=true -–statistic-archive-file=myStatisticsArchiveFile.gfs
 ```
 
-Note that an empty value of `statistic-archive-file` still calculates statistics, but they are not archived to a file.
+Note that a value of `statistic-archive-file` without .gfs suffix still calculates statistics, but they are not archived to a file.
 
 You can also configure sample rate and the filename of your statistic archive files. See [alter runtime](../../tools_modules/gfsh/command-pages/alter.html#topic_7E6B7E1B972D4F418CB45354D1089C2B) for more command options.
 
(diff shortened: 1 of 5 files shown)


davebarnes97:
@mkevo Do I understand correctly that setting the filename to the empty string `""` will now disable writing stats to a file? Formatting of the JIRA description may have deleted quote marks, so just double-checking.

mkevo:
> @mkevo Do I understand correctly that setting the filename to the empty string `""` will now disable writing stats to a file? Formatting of the JIRA description may have deleted quote marks, so just double-checking.

To disable writing stats to file you can use this two options:

alter runtime --statistic-archive-file=
alter runtime --statistic-archive-file=""

mkevo:
As _alter runtime_ command go through all parameters and check it, in case we don't specified any parameters some commands like `alter runtime --member=server` will be successful. So I made a change to accept only file names that has _.gfs_ suffix. To disable it just provide some random string without _.gfs_.

davebarnes97:
@mkevo If the change you describe is a response to my comment on the JIRA ticket, thanks. To keep all discussion in a single location, I'll summarize my comment here:
  I am uncomfortable with the syntax `alter runtime --statistic-archive-file=` where no value follows 
  the equals sign. I would like to see comments from the Apache Geode developers' mailing list on this 
  proposal.
However, I am also uncomfortable with this latest change. If I understand it correctly, to enable this property, the value following the equals sign must be a filename with one of two suffixes: '.gfs' or '.gz'. Any other string disables the archive-to-a-file function.
I find this counterintuitive. In my opinion:
The simplest way to add a turn-off function to this existing property would be to specify an empty string, using a syntax you suggested earlier: `alter runtime --statistic-archive-file=""`.
If that isn't workable, then (again, my opinion) the next best approach would be to add a boolean option that sets statistics archiving to 'true' or 'false'. 
Again, I would like to see this put to the dev list for additional thoughts.

mkevo:
Hi @davebarnes97,
is it OK to close this PR and open a new one with doc change on this part:
`The file to which the running system member writes statistic samples. For example: “StatisticsArchiveFile.gfs”. An empty string disables archiving. Adding .gz suffix to the file name causes it to be compressed.`

to

`The file to which the running system member writes statistic samples. For example: “StatisticsArchiveFile.gfs”. To disable archiving set it to /dev/null. Adding .gz suffix to the file name causes it to be compressed.`

davebarnes97:
+1 OK with me

PR #4604, opened 2020-01-17
GEODE-7716: geode-benchmarks src should be included in release artifacts

onichols-pivotal:
mailing list thread: https://lists.apache.org/thread.html/r67fb7bf3a9d249a0f9566a46bbd5667105e0ba564db3674a878d66d9%40%3Cdev.geode.apache.org%3E

DO NOT MERGE unless consensus has been reached and/or a vote has passed on the dev list.  Until then, this PR is provided only as an exhibit of the changes being proposed.

### DIFF
--- a/dev-tools/release/prepare_rc.sh
+++ b/dev-tools/release/prepare_rc.sh
@@ -99,6 +99,7 @@ GEODE=$WORKSPACE/geode
 GEODE_DEVELOP=$WORKSPACE/geode-develop
 GEODE_EXAMPLES=$WORKSPACE/geode-examples
 GEODE_NATIVE=$WORKSPACE/geode-native
+GEODE_BENCHMARKS=$WORKSPACE/geode-benchmarks
 BREW_DIR=$WORKSPACE/homebrew-core
 SVN_DIR=$WORKSPACE/dist/dev/geode
 set +x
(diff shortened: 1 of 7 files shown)


dev-tools/release/prepare_rc.sh
```diff
@@ -228,10 +250,12 @@ mkdir ${FULL_VERSION}
 cp ${GEODE}/geode-assembly/build/distributions/* ${FULL_VERSION}
 cp ${GEODE_EXAMPLES}/build/distributions/* ${FULL_VERSION}
 cp ${GEODE_NATIVE}/build/apache-geode-native-${VERSION}* ${FULL_VERSION}
+cp ${GEODE_BENCHMARKS}/apache-geode-benchmarks-${VERSION}* ${FULL_VERSION}
+rm ${FULL_VERSION}/apache-geode-examples-*.zip*
```
upthewaterspout: Why this change to remove the examples zip?

onichols-pivotal: thanks, split this out into #4606

onichols-pivotal:
@davebarnes97 if approved, perhaps the Apache Geode website should also have a link to this new artifact starting with 1.12.0

PR #4606, opened 2020-01-17
GEODE-7719: no need to release examples as both .tgz AND .zip

onichols-pivotal:
as requested here: https://lists.apache.org/thread.html/c7bd84b6e6f5464ed674ed447fe8922097237932967b4ed1966e79d3%40%3Cdev.geode.apache.org%3E

Since Geode 1.8.0, we stopped distributing both .tgz AND .zip for Geode, instead favoring .tgz only. However, geode-examples is still shipping both.

For the same reasons cited in GEODE-5694, we only need to ship examples as a .tar.gz, no need for multiple archive formats.

DO NOT MERGE unless consensus has been reached and/or a vote has passed on the dev list. Until then, this PR is provided only as an exhibit of the changes being proposed.


onichols-pivotal:
> Maybe it would be better to just not build the zip file the first place? You can do something like this in gradle/release.gradle in the `geode-examples` project.
> 
> ```
> tasks.distZip.enabled = false
> ```

fixed in https://github.com/apache/geode-examples/pull/91

PR #4610, opened 2020-01-21
GEODE-7740: Native Client enhancements to instance creation.

steve-sienk:
WIP to consolidate instance creation scripts for common CI between Geode main and native client.

### DIFF
--- a/ci/scripts/create_instance.sh
+++ b/ci/scripts/create_instance.sh
@@ -176,6 +176,11 @@ else
       sleep 5
     done
     echo ""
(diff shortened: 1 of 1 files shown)


steve-sienk:
Added as GEODE-7740.

steve-sienk:
[GEODE-7740] Native Client enchancements to instance creation.

    * Add use-local-ssd switch during GCE instance creation for local
    scratch drives
    * Include instance-id in instance-data for use in NC's destroy instance.
    * Initialize scratch drive on Windows instance as drive Z:

PR #4618, opened 2020-01-22
GEODE-7663: Fix delta update inconsistency in client cache.

pivotal-eshu:
* When performing delta update on client, there is a possibility that
   some delta update is not yet applied on client cache.
 * If apply delta update, some queued delta update would be lost -- as
   entry versioning will discard these delta updates when received later.
 * Now it will check if there is a versioning mismatch, if it happened,
   the full value from server will be applied instead of just the delta.

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/cache/client/internal/PutOp.java
+++ b/geode-core/src/main/java/org/apache/geode/cache/client/internal/PutOp.java
@@ -315,7 +315,7 @@ public class PutOp {
         return;
       }
       VersionStamp versionStamp = regionEntry.getVersionStamp();
-      if (deltaSent && versionTag.getEntryVersion() != versionStamp.getEntryVersion() + 1) {
+      if (deltaSent && versionTag.getEntryVersion() > versionStamp.getEntryVersion() + 1) {
         // Delta can't be applied, need to get full value.
         if (logger.isDebugEnabled()) {
           logger.debug("Version is out of order. Need to get from server to perform delta update.");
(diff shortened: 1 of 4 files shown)


geode-core/src/test/java/org/apache/geode/cache/client/internal/PutOpJUnitTest.java
```diff
@@ -16,19 +16,38 @@
...
+    when(entry.getVersionStamp()).thenReturn(versionStamp);
+  }
 
   private EntryEventImpl getEntryEvent() {
-    EntryEventImpl entryEvent = Mockito.mock(EntryEventImpl.class);
-    Mockito.when(entryEvent.getEventId()).thenReturn(new EventID());
+    EntryEventImpl entryEvent = mock(EntryEventImpl.class);
+    when(entryEvent.getEventId()).thenReturn(new EventID());
     return entryEvent;
   }
```
DonalEvans: I think this method can be entirely removed now that there is an `event` field in the test, and calls to it can be replaced with usages of the field.

geode-core/src/main/java/org/apache/geode/cache/client/internal/PutOp.java
```diff
@@ -299,13 +300,39 @@ protected Object processResponse(Message msg, Connection con) throws Exception {
...
 
+    void checkForDeltaConflictAndSetVersionTag(VersionTag versionTag, Connection connection)
+        throws Exception {
+      RegionEntry regionEntry = ((EntryEventImpl) event).getRegionEntry();
+      if (regionEntry == null) {
+        event.setVersionTag(versionTag);
+        return;
+      }
+      VersionStamp versionStamp = regionEntry.getVersionStamp();
+      if (deltaSent && versionTag.getEntryVersion() != versionStamp.getEntryVersion() + 1) {
```
gesterzhou: It should be ">" not "!=", in the case of "<", you should not call the get.

pivotal-eshu: I am not sure if client delta put will generate region version less than current local cache version. However, use ">" might be more clear in this case.

geode-core/src/test/java/org/apache/geode/cache/client/internal/PutOpJUnitTest.java
pivotal-eshu: Done

PR #4627, opened 2020-01-23
GEODE-7565: Allow gw receivers with same host and port

alb3rtobr:
There is a problem with Geode WAN replication when GW receivers are configured with the same hostname-for-senders and port on all servers. The reason for such a setup is deploying Geode cluster on a Kubernetes cluster where all GW receivers are reachable from the outside world on the same VIP and port.
 
The problem experienced is that shutting down one server is stopping replication to this cluster until the server is up again. This is because Geode incorrectly assumes there are no more alive servers when just one of them is down, because since they share hostname-for-senders and port, they are treated as one same server.

With these changes locator is able to distinguish the different receivers using the same hostname and port so replication is not impacted when one server is stopped.

( Link to thread in dev mailing list: https://markmail.org/thread/6qakx67rxiokdsec )


geode-core/src/main/java/org/apache/geode/distributed/internal/LocatorLoadSnapshot.java
```diff
@@ -50,7 +50,8 @@
 
   private final Map<ServerLocation, String[]> serverGroupMap = new HashMap<>();
 
-  private final Map<String, Map<ServerLocation, LoadHolder>> connectionLoadMap = new HashMap<>();
+  private final Map<String, Map<ServerLocationAndMemberId, LoadHolder>> connectionLoadMap =
+      new HashMap<>();
 
   private final Map<String, Map<ServerLocation, LoadHolder>> queueLoadMap = new HashMap<>();
```
pivotal-jbarrett: This map needs to change for the same reasons the `connectionLoadMap` needs changing.

geode-core/src/main/java/org/apache/geode/distributed/internal/LocatorLoadSnapshot.java
```diff
@@ -50,7 +50,8 @@
 
   private final Map<ServerLocation, String[]> serverGroupMap = new HashMap<>();
```
pivotal-jbarrett: It is possible now, with duplicate `ServerLocation` that `serverGroupMap` could have the wrong groups. It should also use the member ID as its key.

alb3rtobr:
> Why use the strongly typed InternalDistributedMember rather than it's string representation?

I don't see which part of the code is your comment about, but I suppose is referred to `ServerLocationAndMemberId` class, am I right? I used the string representation because its the minimum information I think I need. Thanks for the fast review!

pivotal-jbarrett:
> > Why use the strongly typed InternalDistributedMember rather than it's string representation?
> 
> I don't see which part of the code is your comment about, but I suppose is referred to `ServerLocationAndMemberId` class, am I right? I used the string representation because its the minimum information I think I need. Thanks for the fast review!

Sorry, I should have attached it to a line of code. Yes, the issues is with using the `String` since it we have a more strongly typed value the string is derived from why not use it. 

If you consider my other comment about the ID really being the only unique part of the key anyway it makes a stronger argument for the strong type over the string as the key for the maps.

geode-core/src/main/java/org/apache/geode/distributed/internal/LocatorLoadSnapshot.java
alb3rtobr: done

geode-core/src/main/java/org/apache/geode/distributed/internal/LocatorLoadSnapshot.java
alb3rtobr: done

alb3rtobr:
Hi @pivotal-jbarrett , the changes you requested are done and all checks pass, could you please take a look? Thanks!

alb3rtobr:
This PR has become a bit messy, so Im closing it.
I have created a new one: https://github.com/apache/geode/pull/4824
And also, a wiki page describing the problem more in detail: https://cwiki.apache.org/confluence/display/GEODE/Allow+same+host+and+port+for+all+gateway+receivers

PR #4629, opened 2020-01-24
GEODE-7727: modify sender thread to detect release of connection

mivanac:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/tcp/Connection.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/tcp/Connection.java
@@ -1486,6 +1486,8 @@ public class Connection implements Runnable {
         }
         asyncClose(false);
         owner.removeAndCloseThreadOwnedSockets();
+      } else {
+        asyncClose(false);
       }
       releaseInputBuffer();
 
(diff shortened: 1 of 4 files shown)


geode-core/src/main/java/org/apache/geode/internal/tcp/Connection.java
```diff
@@ -1692,10 +1700,8 @@ private void readMessages() {
         }
       }
     } finally {
-      if (!isHandShakeReader) {
-        synchronized (stateLock) {
-          connectionState = STATE_IDLE;
-        }
+      synchronized (stateLock) {
```
bschuchardt: The (!isHandShakeReader) test was added here to fix a race condition between the handshake thread and an application thread using the thread-owned Connection.  I think this test needs to remain in place.

geode-core/src/main/java/org/apache/geode/internal/tcp/Connection.java
```diff
@@ -1637,8 +1640,13 @@ private void readMessages() {
               }
             }
             isHandShakeReader = true;
-            // Once we have read the handshake the reader can go away
-            break;
+            notifyHandshakeWaiter(false);
```
bschuchardt: handshakeRead and handshakeCancelled are only set to true in notifyHandshakeWaiter().  Why do we need to call that method again in this spot?

geode-core/src/main/java/org/apache/geode/internal/tcp/Connection.java
```diff
@@ -1637,8 +1640,13 @@ private void readMessages() {
...
+            // Once we have read the handshake for unshared connections, the reader can skip
+            // processing messages
+            if (!sharedResource) {
```
bschuchardt: good change here.  It shouldn't be needed since a sharedResource will only invoke this method if isReaderThread==true but having an explicit test is a good idea.

geode-core/src/main/java/org/apache/geode/internal/tcp/Connection.java
mivanac: thanks for comments. I will return check on !isHandShakeReader.

geode-core/src/main/java/org/apache/geode/internal/tcp/Connection.java
mivanac: updated

PR #4633, opened 2020-01-27
GEODE-7109: add test case that multiple sessions can be maintained.

pivotal-eshu:


geode-assembly/src/distributedTest/java/org/apache/geode/session/tests/CargoTestBase.java
```diff
@@ -139,22 +150,25 @@ public void getKeyValueDataOnAllClients(String key, String expectedValue, String
         String value = resp.getResponse();
         if (!expectedValue.equals(value)) {
           LogService.getLogger().info("verifying container {} for expected value of {}"
-              + " for key {}, but gets response value of {}. Waiting for update from server.", i,
-              expectedValue, key, value);
+              + " for key {}, but gets response value of {}. Waiting for update from server. for client {}",
```
DonalEvans: This would read better as "but gets response value of {} from client {}. Waiting for update from server."

pivotal-eshu: Will try to change this in next commit for the test.

PR #4638, opened 2020-01-28
GEODE-7683: introduce BR.cmnClearRegion

gesterzhou:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/BucketRegion.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/BucketRegion.java
@@ -557,6 +557,36 @@ public class BucketRegion extends DistributedRegion implements Bucket {
     }
   }
 
+  @Override
+  void cmnClearRegion(RegionEventImpl regionEvent, boolean cacheWrite, boolean useRVV) {
+    if (!getBucketAdvisor().isPrimary()) {
+      if (logger.isDebugEnabled()) {
+        logger.debug("Not primary bucket when doing clear, do nothing");
+      }
+      return;
+    }
+
+    boolean enableRVV = useRVV && getConcurrencyChecksEnabled();
+    RegionVersionVector rvv = null;
+    if (enableRVV) {
+      rvv = getVersionVector().getCloneForTransmission();
+    }
+
+    // get rvvLock
+    Set<InternalDistributedMember> participants =
+        getCacheDistributionAdvisor().adviseInvalidateRegion();
+    try {
+      obtainWriteLocksForClear(regionEvent, participants);
+      // no need to dominate my own rvv.
+      // Clear is on going here, there won't be GII for this member
+      clearRegionLocally(regionEvent, cacheWrite, null);
+      distributeClearOperation(regionEvent, rvv, participants);
+
+      // TODO: call reindexUserDataRegion if there're lucene indexes
+    } finally {
+      releaseWriteLocksForClear(regionEvent, participants);
+    }
+  }
 
   long generateTailKey() {
     long key = eventSeqNum.addAndGet(partitionedRegion.getTotalNumberOfBuckets());
@@ -2094,11 +2124,9 @@ public class BucketRegion extends DistributedRegion implements Bucket {
       // if GII has failed, because there is not primary. So it's safe to set these
       // counters to 0.
       oldMemValue = bytesInMemory.getAndSet(0);
-    }
-
-    else {
-      throw new InternalGemFireError(
-          "Trying to clear a bucket region that was not destroyed or in initialization.");
+    } else {
+      // BucketRegion's clear is supported now
+      oldMemValue = bytesInMemory.getAndSet(0);
     }
     if (oldMemValue != BUCKET_DESTROYED) {
       partitionedRegion.getPrStats().incDataStoreEntryCount(-sizeBeforeClear);
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/DistributedRegion.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/DistributedRegion.java
@@ -2003,6 +2003,10 @@ public class DistributedRegion extends LocalRegion implements InternalDistribute
     super.basicClear(regionEvent, cacheWrite);
   }
 
+  void distributeClearOperation(RegionEventImpl regionEvent, RegionVersionVector rvv,
+      Set<InternalDistributedMember> participants) {
+    DistributedClearOperation.clear(regionEvent, rvv, participants);
+  }
 
   @Override
   void cmnClearRegion(RegionEventImpl regionEvent, boolean cacheWrite, boolean useRVV) {
@@ -2025,7 +2029,7 @@ public class DistributedRegion extends LocalRegion implements InternalDistribute
             obtainWriteLocksForClear(regionEvent, participants);
             clearRegionLocally(regionEvent, cacheWrite, null);
             if (!regionEvent.isOriginRemote() && regionEvent.getOperation().isDistributed()) {
-              DistributedClearOperation.clear(regionEvent, null, participants);
+              distributeClearOperation(regionEvent, null, participants);
             }
           } finally {
             releaseWriteLocksForClear(regionEvent, participants);
@@ -2081,10 +2085,12 @@ public class DistributedRegion extends LocalRegion implements InternalDistribute
   /**
    * obtain locks preventing generation of new versions in other members
    */
-  private void obtainWriteLocksForClear(RegionEventImpl regionEvent,
+  protected void obtainWriteLocksForClear(RegionEventImpl regionEvent,
       Set<InternalDistributedMember> participants) {
     lockLocallyForClear(getDistributionManager(), getMyId(), regionEvent);
-    DistributedClearOperation.lockAndFlushToOthers(regionEvent, participants);
+    if (!isUsedForPartitionedRegionBucket()) {
+      DistributedClearOperation.lockAndFlushToOthers(regionEvent, participants);
+    }
   }
 
   /**
@@ -2121,7 +2127,7 @@ public class DistributedRegion extends LocalRegion implements InternalDistribute
   /**
    * releases the locks obtained in obtainWriteLocksForClear
    */
-  private void releaseWriteLocksForClear(RegionEventImpl regionEvent,
+  protected void releaseWriteLocksForClear(RegionEventImpl regionEvent,
       Set<InternalDistributedMember> participants) {
 
     ARMLockTestHook armLockTestHook = getRegionMap().getARMLockTestHook();
@@ -2129,8 +2135,13 @@ public class DistributedRegion extends LocalRegion implements InternalDistribute
       armLockTestHook.beforeRelease(this, regionEvent);
     }
 
-    getVersionVector().unlockForClear(getMyId());
-    DistributedClearOperation.releaseLocks(regionEvent, participants);
+    RegionVersionVector rvv = getVersionVector();
+    if (rvv != null) {
+      rvv.unlockForClear(getMyId());
+    }
+    if (!isUsedForPartitionedRegionBucket()) {
+      DistributedClearOperation.releaseLocks(regionEvent, participants);
+    }
 
     if (armLockTestHook != null) {
       armLockTestHook.afterRelease(this, regionEvent);
--- a/geode-core/src/test/java/org/apache/geode/internal/cache/BucketRegionJUnitTest.java
+++ b/geode-core/src/test/java/org/apache/geode/internal/cache/BucketRegionJUnitTest.java
@@ -14,7 +14,9 @@
  */
 package org.apache.geode.internal.cache;
 
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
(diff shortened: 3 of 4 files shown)


geode-core/src/main/java/org/apache/geode/internal/cache/BucketRegion.java
```diff
@@ -557,6 +558,42 @@ public boolean virtualPut(EntryEventImpl event, boolean ifNew, boolean ifOld,
...
+    // lock the primary from moving
+    DistributedLockService lockService = DistributedLockService
+        .getServiceNamed(PartitionedRegionHelper.PARTITION_LOCK_SERVICE_NAME);
+    String lockName = this.getFullPath();
+    while (!lockService.lock(lockName, 100, -1)) {
+      if (!getBucketAdvisor().isPrimary()) {
+        PartitionedRegionException pre =
+            new PartitionedRegionException(
+                "The bucket " + this.getId() + " is no longer a primary. Retry the clear");
+        throw pre;
```
BenjaminPerryRoss: Since we locked the primary bucket just now, should we first unlock it before we throw due to the primary no longer being where we expect?

geode-core/src/main/java/org/apache/geode/internal/cache/BucketRegion.java
```diff
@@ -557,6 +557,34 @@ public boolean virtualPut(EntryEventImpl event, boolean ifNew, boolean ifOld,
     }
   }
 
+  @Override
+  void cmnClearRegion(RegionEventImpl regionEvent, boolean cacheWrite, boolean useRVV) {
+    if (getBucketAdvisor().isPrimary()) {
```
jhuynh1: This is opposite of what the comment says

geode-core/src/main/java/org/apache/geode/internal/cache/BucketRegion.java
```diff
@@ -557,6 +557,34 @@ public boolean virtualPut(EntryEventImpl event, boolean ifNew, boolean ifOld,
...
+      logger.info("Not primary bucket when doing clear, do nothing");
+      return;
+    }
+
+    boolean enableRVV = useRVV && getConcurrencyChecksEnabled();
+    RegionVersionVector rvv = null;
+    if (enableRVV) {
+      rvv = getVersionVector();
```
jhuynh1: Does this call just return null if it's not enabled?  Do we need the if check?

geode-core/src/main/java/org/apache/geode/internal/cache/BucketRegion.java
```diff
@@ -557,6 +557,34 @@ public boolean virtualPut(EntryEventImpl event, boolean ifNew, boolean ifOld,
...
+    }
+
+    // get rvvLock
+    Set<InternalDistributedMember> participants =
+        getCacheDistributionAdvisor().adviseInvalidateRegion();
+    try {
+      obtainWriteLocksForClear(regionEvent, participants);
+      clearRegionLocally(regionEvent, cacheWrite, null);
+      if (!regionEvent.isOriginRemote() && regionEvent.getOperation().isDistributed()) {
```
jhuynh1: Shouldn't it only be the origin that distributes?

geode-core/src/main/java/org/apache/geode/internal/cache/DistributedRegion.java
```diff
@@ -2081,10 +2081,12 @@ private void distributedUnlockForClear() {
   /**
    * obtain locks preventing generation of new versions in other members
    */
-  private void obtainWriteLocksForClear(RegionEventImpl regionEvent,
+  protected void obtainWriteLocksForClear(RegionEventImpl regionEvent,
       Set<InternalDistributedMember> participants) {
     lockLocallyForClear(getDistributionManager(), getMyId(), regionEvent);
-    DistributedClearOperation.lockAndFlushToOthers(regionEvent, participants);
+    if (!isUsedForPartitionedRegionBucket()) {
+      DistributedClearOperation.lockAndFlushToOthers(regionEvent, participants);
```
jhuynh1: Not part of this diff, but I assume we are taking a less granular lock earlier in the process for the entire partitioned region?

geode-core/src/main/java/org/apache/geode/internal/cache/BucketRegion.java
gesterzhou: Yes, after I carefully reviewed the code, we don't need this check.

geode-core/src/main/java/org/apache/geode/internal/cache/DistributedRegion.java
gesterzhou: BR does not need to send msg to secondary to do rvvLock.

geode-core/src/main/java/org/apache/geode/internal/cache/BucketRegion.java
gesterzhou: Yes, then just use null for rvv.

geode-core/src/main/java/org/apache/geode/internal/cache/BucketRegion.java
gesterzhou: Yes, I have fixed.

geode-core/src/main/java/org/apache/geode/internal/cache/BucketRegion.java
jhuynh1: If this returns null, then the if check isn't needed.  It'll be null either way.

geode-core/src/main/java/org/apache/geode/internal/cache/BucketRegion.java
```diff
@@ -557,6 +557,32 @@ public boolean virtualPut(EntryEventImpl event, boolean ifNew, boolean ifOld,
...
+
+    // get rvvLock
+    Set<InternalDistributedMember> participants =
+        getCacheDistributionAdvisor().adviseInvalidateRegion();
+    try {
+      obtainWriteLocksForClear(regionEvent, participants);
+      clearRegionLocally(regionEvent, cacheWrite, rvv);
+      distributeClearOperation(regionEvent, rvv, participants);
+
+      // TODO: call reindexUserDataRegion if there're lucene indexes
```
boglesby: I see you made reindexUserDataRegion public. Is this something you need to call here?

gesterzhou: Yes, that will be called in future as a todo work.

geode-lucene/src/main/java/org/apache/geode/cache/lucene/internal/IndexRepositoryFactory.java
```diff
@@ -152,7 +152,7 @@ protected IndexWriter buildIndexWriter(int bucketId, BucketRegion fileAndChunkBu
     return new IndexWriter(dir, config);
   }
 
-  private boolean reindexUserDataRegion(Integer bucketId, PartitionedRegion userRegion,
+  public boolean reindexUserDataRegion(Integer bucketId, PartitionedRegion userRegion,
```
jchen21: There is no need to change the scope to `public`, at least not in this pull request.

geode-core/src/main/java/org/apache/geode/internal/cache/BucketRegion.java
gesterzhou: There's basicLocalClear whose useRVV is false. I need to consider that.

geode-core/src/main/java/org/apache/geode/internal/cache/BucketRegion.java
gesterzhou: But I changed it back for the time being

gesterzhou:
> I'm not 100% sure I understand what this is implementing. Clearing a BucketRegion still fails the same way it has in the past.
> 
> I wrote a function that attempts to clear a PartitionedRegion by clearing all of its primary BucketRegions. To do that, it tried these API:
> 
> ```
> br.basicClear(regionEvent);
> br.basicClear(regionEvent, true);
> br.cmnClearRegion(regionEvent, true, true);
> ```
> 
> No matter which of those it called, this exception occurs:
> 
> ```
> Caused by: org.apache.geode.InternalGemFireError: Trying to clear a bucket region that was not destroyed or in initialization.
> 	at org.apache.geode.internal.cache.BucketRegion.updateSizeOnClearRegion(BucketRegion.java:2128)
> 	at org.apache.geode.internal.cache.AbstractRegionMap.clear(AbstractRegionMap.java:460)
> 	at org.apache.geode.internal.cache.BucketRegion.clearEntries(BucketRegion.java:2479)
> 	at org.apache.geode.internal.cache.LocalRegion.clearRegionLocally(LocalRegion.java:8544)
> 	at org.apache.geode.internal.cache.BucketRegion.cmnClearRegion(BucketRegion.java:578)
> 	at org.apache.geode.internal.cache.BasicClearBucketRegionFunction.basicClear(BasicClearBucketRegionFunction.java:32)
> 	at org.apache.geode.internal.cache.BasicClearBucketRegionFunction.execute(BasicClearBucketRegionFunction.java:21)
> 	at org.apache.geode.internal.cache.execute.AbstractExecution.executeFunctionLocally(AbstractExecution.java:328)
> 	at org.apache.geode.internal.cache.execute.AbstractExecution.lambda$executeFunctionOnLocalPRNode$0(AbstractExecution.java:273)
> 	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
> 	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
> 	at org.apache.geode.distributed.internal.ClusterOperationExecutors.runUntilShutdown(ClusterOperationExecutors.java:449)
> 	at org.apache.geode.distributed.internal.ClusterOperationExecutors.doFunctionExecutionThread(ClusterOperationExecutors.java:379)
> 	at org.apache.geode.logging.internal.executors.LoggingThreadFactory.lambda$newThread$0(LoggingThreadFactory.java:119)
> 	at java.lang.Thread.run(Thread.java:745)
> ```
> 
> AFAICT, a BucketRegion still can't be cleared.
> 
> When I hack-implemented this feature, I changed BucketRegion.updateSizeOnClearRegion to do this instead of throwing the InternalGemFireError. I'm not sure if this is valid, though.
> 
> ```
> oldMemValue = this.bytesInMemory.getAndSet(0);
> ```

This looks like a loose-end I did not found earlier. I will do that fix. And I will add this dunit test case in PR.clear.

geode-core/src/main/java/org/apache/geode/internal/cache/BucketRegion.java
```diff
@@ -2087,19 +2113,10 @@ void updateSizeOnClearRegion(int sizeBeforeClear) {
     if (isDestroyed || isDestroyingDiskRegion) {
       // If this region is destroyed, mark the stat as destroyed.
       oldMemValue = bytesInMemory.getAndSet(BUCKET_DESTROYED);
-
-    } else if (!isInitialized()) {
-      // This case is rather special. We clear the region if the GII failed.
-      // In the case of bucket regions, we know that there will be no concurrent operations
-      // if GII has failed, because there is not primary. So it's safe to set these
-      // counters to 0.
+    } else {
```
jhuynh1: This is not quite the same.  Is there a reason why we removed the if check?  We will now always set oldMemValue even if it's been initialized?

why wasn't this just a removal of the else and leave the else if? Just want to understand the logic here a bit...

geode-core/src/main/java/org/apache/geode/internal/cache/BucketRegion.java
jhuynh1: Lets think of the case where useRVV is false.
if (false) {} 
rvv is already null, so rvv ends up null;
OR
//remove the if check
rvv = getVersionVector() <-returns null anyways

The end result is the same.  why have the if check?

gesterzhou: In BR.cmnClearRegion(), since we have locked the rvvLock, no more operations will be intiated here. So no need to explicitly use RVV. I will fix this part.

geode-core/src/main/java/org/apache/geode/internal/cache/BucketRegion.java
gesterzhou: OK, I will keep the "else if" with comments and fix into the "else" only.

geode-core/src/main/java/org/apache/geode/internal/cache/BucketRegion.java
```diff
@@ -557,6 +557,32 @@ public boolean virtualPut(EntryEventImpl event, boolean ifNew, boolean ifOld,
     }
   }
 
+  @Override
+  void cmnClearRegion(RegionEventImpl regionEvent, boolean cacheWrite, boolean useRVV) {
+    if (!getBucketAdvisor().isPrimary()) {
+      logger.info("Not primary bucket when doing clear, do nothing");
```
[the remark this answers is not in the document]
gesterzhou: done

gesterzhou:
there's some dunit tests fail from time to time on different test. I think it's unstable in base branch i.e feature/GEODE-7665

PR #4639, opened 2020-01-28
GEODE-7728: Fix Assertion Failures in OQL

jujoramos:
- Added unit and integration tests.
- Back off from the single index optimisation during join queries if
  we can't find the (single) best filter, allowing the regular
  execution to continue and use multiple indexes.
- If there are multiple indexes, add the filter evaluable operand at
  the beginning of the internal list to evaluate it first in the chain.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/cache/query/internal/CompiledComparison.java
+++ b/geode-core/src/main/java/org/apache/geode/cache/query/internal/CompiledComparison.java
@@ -296,15 +296,23 @@ public class CompiledComparison extends AbstractCompiledValue
   public int getSizeEstimate(ExecutionContext context) throws FunctionDomainException,
       TypeMismatchException, NameResolutionException, QueryInvocationTargetException {
     IndexInfo[] idxInfo = getIndexInfo(context);
-    if ((idxInfo == null) || (idxInfo.length > 1)) {
-      // Asif: This implies it is an independent condition. So evaluate it first
-      // in filter operand
+
+    // Both operands are indexed, evaluate it first in the filter operand.
+    if (idxInfo != null && idxInfo.length > 1) {
       return 0;
     }
+
+    // Asif: This implies it is an independent condition. So evaluate it second in filter operand.
+    if (idxInfo == null) {
+      return 1;
+    }
+
     assert idxInfo.length == 1;
     Object key = idxInfo[0].evaluateIndexKey(context);
+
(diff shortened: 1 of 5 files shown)


geode-core/src/main/java/org/apache/geode/cache/query/internal/CompiledComparison.java
```diff
@@ -296,7 +296,7 @@ public boolean isRangeEvaluatable() {
   public int getSizeEstimate(ExecutionContext context) throws FunctionDomainException,
       TypeMismatchException, NameResolutionException, QueryInvocationTargetException {
     IndexInfo[] idxInfo = getIndexInfo(context);
-    if (idxInfo == null) {
+    if ((idxInfo == null) || (idxInfo.length > 1)) {
```
jhuynh1: Just a question, should this condition be on it's own and does it make sense to give it a score of 0?  Should these types of conditions be waited more heavily or considered "better?"  I'm ok with 0 just curious if you tried different cases out :-)

jujoramos: Good point, and I think you're right... having more than one index available for a `CompiledComparison` implies that both the left and right operands are indexed, so that case should have more "weight" when compared to an independent operator that would imply scanning the full region.
Will make the changes and add some more tests, thanks for finding this!.

jujoramos: Done!

geode-core/src/main/java/org/apache/geode/cache/query/internal/CompiledComparison.java
```diff
@@ -296,15 +296,23 @@ public boolean isRangeEvaluatable() {
...
+      return 1;
+    }
+
     assert idxInfo.length == 1;
     Object key = idxInfo[0].evaluateIndexKey(context);
+
+    // Evaluate it second as we do with the independent condition if key was not found.
     if (key != null && key.equals(QueryService.UNDEFINED)) {
-      return 0;
+      return 1;
```
jhuynh1: I think if the key is UNDEFINED and we have indexes present, the index actually has a mapping for UNDEFINED so the lookup should itself be really fast so 0 might be ok

jujoramos: Changed! 👍

PR #4641, opened 2020-01-28
GEODE-7604: Fix QueryConfigurationServiceConstraintsDistributedTest

kirklund:
GEODE-7604: Fix QueryConfigurationServiceConstraintsDistributedTest

* Rewrite test with standard Geode and JMX APIs
* Rewrite CountingCqListener with AtomicIntegers
* Replace if-else with enum

### DIFF
--- a/geode-cq/src/distributedTest/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceConstraintsDistributedTest.java
+++ b/geode-cq/src/distributedTest/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceConstraintsDistributedTest.java
@@ -27,7 +27,6 @@ import static org.apache.geode.test.dunit.VM.getVM;
 import static org.assertj.core.api.Assertions.assertThat;
 
 import java.io.File;
-import java.io.IOException;
 import java.io.Serializable;
 import java.lang.reflect.Method;
 import java.util.Arrays;
@@ -39,6 +38,7 @@ import java.util.stream.IntStream;
 import java.util.stream.Stream;
 
 import junitparams.JUnitParamsRunner;
+import junitparams.NamedParameters;
 import junitparams.Parameters;
 import junitparams.naming.TestCaseName;
 import org.junit.Before;
@@ -73,8 +73,8 @@ import org.apache.geode.test.junit.categories.SecurityTest;
 import org.apache.geode.test.junit.rules.serializable.SerializableTemporaryFolder;
 import org.apache.geode.test.junit.rules.serializable.SerializableTestName;
 
-@RunWith(JUnitParamsRunner.class)
 @Category({OQLQueryTest.class, SecurityTest.class})
+@RunWith(JUnitParamsRunner.class)
 @SuppressWarnings("serial")
 public class QueryConfigurationServiceConstraintsDistributedTest implements Serializable {
 
@@ -123,7 +123,17 @@ public class QueryConfigurationServiceConstraintsDistributedTest implements Seri
     int serverPort = getRandomAvailableTCPPort();
 
     serverVM.invoke(() -> {
-      createServer(serverPort);
+      serverLauncher = new ServerLauncher.Builder()
+          .setMemberName("server")
+          .setServerPort(serverPort)
+          .setWorkingDirectory(temporaryFolder.newFolder("server").getAbsolutePath())
+          .set(SERIALIZABLE_OBJECT_FILTER, "org.apache.geode.cache.query.internal.*")
+          .set(SECURITY_MANAGER, SimpleSecurityManager.class.getName())
+          .set(USER_NAME, "cluster")
+          .set(PASSWORD, "cluster")
+          .build();
+
+      serverLauncher.start();
 
       InternalCache internalCache = (InternalCache) serverLauncher.getCache();
       assertThat(internalCache).isNotNull();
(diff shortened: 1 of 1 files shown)


geode-cq/src/distributedTest/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceConstraintsDistributedTest.java
```diff
@@ -72,158 +90,74 @@
...
+
+  private VM serverVM;
+  private VM clientVM;
+
+  private String regionName;
+  private String queryString;
 
   @Rule
-  public ClusterStartupRule cluster = new ClusterStartupRule();
+  public DistributedRule distributedRule = new DistributedRule();
```
jujoramos: I thought we were encouraging people to use the `ClusterStartuoRule` instead of `DistributedRule`... did that change?.

geode-cq/src/distributedTest/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceConstraintsDistributedTest.java
```diff
@@ -378,4 +418,28 @@ public boolean authorize(Method method, Object target) {
       return authorizedMethods.contains(method.getName());
     }
   }
+
+  private static class CountingCqListener implements CqListener {
+
+    private final AtomicInteger eventCount = new AtomicInteger();
```
kirklund: The fix: Use new inner class with AtomicIntegers instead of TestCqListener which is not thread safe.

geode-cq/src/distributedTest/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceConstraintsDistributedTest.java
```diff
@@ -233,41 +167,57 @@ private void createClientCq(String queryString, boolean executeWithInitialResult
...
+      "REPLICATE, UPDATE, true", "REPLICATE, UPDATE, false",
+      "REPLICATE, REPLACE, true", "REPLICATE, REPLACE, false",
+      "REPLICATE, INVALIDATE, true", "REPLICATE, INVALIDATE, false",
+      "PARTITION, PUT, true", "REPLICATE, PUT, false",
+      "PARTITION, CREATE, true", "REPLICATE, CREATE, false",
+      "PARTITION, REMOVE, true", "REPLICATE, REMOVE, false",
+      "PARTITION, DESTROY, true", "REPLICATE, DESTROY, false",
+      "PARTITION, UPDATE, true", "REPLICATE, UPDATE, false",
+      "PARTITION, REPLACE, true", "REPLICATE, REPLACE, false",
+      "PARTITION, INVALIDATE, true", "REPLICATE, INVALIDATE, false"})
```
jujoramos: The set of parameters is duplicated across the test, I believe the previous approach of having a specific method to return the full list of parameters is easier to read. Am I missing something here?.

geode-cq/src/distributedTest/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceConstraintsDistributedTest.java
kirklund: It's possible that changing to Geode and JMX APIs may have contributed to fixing the flakiness.

geode-cq/src/distributedTest/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceConstraintsDistributedTest.java
kirklund: Actually, we're encouraging people t use DistributedRule with Geode APIs instead of ClusterStartupRule. ClusterStartupRule provides an entire new API layer for Geode and I'm 100% against that approach for writing tests.

geode-cq/src/distributedTest/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceConstraintsDistributedTest.java
kirklund: I can restore this. The reason I did this was because the JUnitParams parameter methods are always detected as unused by the IDE, so then you have to mark it with `SuppressWarnings("unused")`.

kirklund: I also prefer to see the parameters with the test method instead of looking for another method. I could try to find a 3rd way to provide the parameters to the tests that isn't redundant.

mhansonp: I am don't like the duplication, but I think this way is easier to read. I think it doesn't matter a lot in this case. if the test were getting duplicated a lot or if this was in core (not test) code, I might object more.

kirklund:
I pushed some more changes including changing from inlined parameters to using a method annotated with `NamedParameters`.

geode-cq/src/distributedTest/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceConstraintsDistributedTest.java
jujoramos: Okay, good to know this... so, as the rule of thumb, we should use `DisributedRule` and `CacheRule` for distributed tests, and `ServerStarterRule` / `LocatorStarterRule` for integration tests, is that right?.

geode-cq/src/distributedTest/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceConstraintsDistributedTest.java
jujoramos: 👍

kirklund: I ended up replacing it with `NamedParameters` method.

geode-cq/src/distributedTest/java/org/apache/geode/cache/query/internal/QueryConfigurationServiceConstraintsDistributedTest.java
kirklund: I would just `DistributedRule` for distributed tests. Don't even use `CacheRule`. I set ServerLauncher and LocatorLauncher statics or fields in the test for starting Server and Locator. These are the same APIs that a User would use (or GFSH).

PR #4642, opened 2020-01-28
GEODE-7600-4: Further cleanup

mhansonp:
Cleaning the teardown approach
change to wait for event history to be correct.
getting rid of a timing issue

### DIFF
--- a/geode-core/src/distributedTest/java/org/apache/geode/cache/ConnectionPoolDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/cache/ConnectionPoolDUnitTest.java
@@ -112,7 +112,7 @@ public class ConnectionPoolDUnitTest extends JUnit4CacheTestCase {
   @Before
   public void setup() {
     vm0 = VM.getVM(0);
-    vm1 = VM.getVM(-1);
+    vm1 = VM.getVM(1);
     vm2 = VM.getVM(2);
     vm3 = VM.getVM(3);
   }
--- a/geode-core/src/distributedTest/java/org/apache/geode/cache/ConnectionPoolDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/cache/ConnectionPoolDUnitTest.java
@@ -34,7 +34,6 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.LinkedList;
 import java.util.List;
-import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Stream;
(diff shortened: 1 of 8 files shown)


geode-core/src/distributedTest/java/org/apache/geode/cache/ConnectionPoolDUnitTest.java
```diff
@@ -120,24 +119,14 @@ public void setup() {
...
-      }
-    });
-  }
-
-  @Override
-  public final void postTearDownCacheTestCase() {
     Invoke.invokeInEveryVM(() -> {
-      Map<String, Pool> pools = PoolManager.getAll();
-      assertThat(pools).describedAs("found pools remaining after teardown: " + pools).isEmpty();
+      if (basicGetCache() != null) {
```
mhansonp: Part of this fix was to cleanout the teardown code so it would run correctly.

geode-core/src/distributedTest/java/org/apache/geode/cache/ConnectionPoolDUnitTest.java
```diff
@@ -1298,75 +1311,70 @@ public void run2() throws CacheException {
...
-      CertifiableTestCacheListener<Object, Object> ctl =
-          (CertifiableTestCacheListener<Object, Object>) region.getAttributes()
-              .getCacheListeners()[0];
-      List<CacheEvent<Object, Object>> list = ctl.getEventHistory();
-      logger
-          .info("history (should be empty): " + list);
-      assertThat(list).isEmpty();
-      // now see if we can get it from the server
+
+      await().untilAsserted(() -> {
```
mhansonp: This is important to wait so that the server has all the keys so that we don't get invalidates in addition to the Local_Load_Creates

geode-core/src/distributedTest/java/org/apache/geode/cache/ConnectionPoolDUnitTest.java
```diff
@@ -333,9 +322,6 @@ public void beforeCreate2(EntryEvent event) throws CacheWriterException {
       Region<Object, Object> region = getRootRegion().getSubregion(name);
       region.localDestroyRegion();
     });
-
```
mhansonp: These are no longer necessary.

geode-core/src/distributedTest/java/org/apache/geode/cache/ConnectionPoolDUnitTest.java
kirklund: I learned today that you can just do:
```
PoolManager.close()
```
And that'll close all pools.

mhansonp: Sweet, next time I will use that...

PR #4653, opened 2020-01-31
GEODE-5595: Fix DeltaPropagationDUnitTest flakiness

kirklund:
Improve testability of CacheClientProxy
* Extract inner classes
* Introduce CacheClientProxyFactory with support for property injection
* Pull getPrimaryPort up to InternalPool interface
* Define InternalPoolFactory and pull up init and getPoolAttributes

Replace flaky sleep-based test-hook in product code with custom
MessageDispatcher injected by property in CacheClientProxyFactory
which uses CountDownLatch.


kirklund:
DistributedTestOpen failure was Tomcat7CachingClientServerTest which is unrelated to the changes for DeltaPropagationDUnitTest.

PR #4668, opened 2020-02-03
GEODE-7760: NPE in Locator during auto-reconnect

bschuchardt:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/distributed/internal/InternalLocator.java
+++ b/geode-core/src/main/java/org/apache/geode/distributed/internal/InternalLocator.java
@@ -773,15 +773,19 @@ public class InternalLocator extends Locator implements ConnectListener, LogConf
     startClusterManagementService();
   }
 
-  private void startClusterManagementService() throws IOException {
+  @VisibleForTesting
+  void startClusterManagementService() throws IOException {
     startConfigurationPersistenceService();
 
-    if (internalCache == null) {
+    InternalCache myCache = this.internalCache;
+    InternalLocator currentLocator = InternalLocator.locator;
+
+    if (myCache == null || currentLocator == null) {
       return;
     }
 
-    clusterManagementService = new LocatorClusterManagementService(locator.internalCache,
-        locator.configurationPersistenceService);
+    clusterManagementService = new LocatorClusterManagementService(currentLocator.internalCache,
+        currentLocator.configurationPersistenceService);
 
     // start management rest service
     AgentUtil agentUtil = new AgentUtil(GemFireVersion.getGemFireVersion());
@@ -796,7 +800,7 @@ public class InternalLocator extends Locator implements ConnectListener, LogConf
 
     Map<String, Object> serviceAttributes = new HashMap<>();
     serviceAttributes.put(HttpService.SECURITY_SERVICE_SERVLET_CONTEXT_PARAM,
-        internalCache.getSecurityService());
+        myCache.getSecurityService());
     serviceAttributes.put(HttpService.CLUSTER_MANAGEMENT_SERVICE_CONTEXT_PARAM,
         clusterManagementService);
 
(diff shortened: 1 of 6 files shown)


geode-core/src/main/java/org/apache/geode/distributed/internal/InternalLocator.java
```diff
@@ -773,15 +773,19 @@ private void startCache(DistributedSystem system) throws IOException {
     startClusterManagementService();
   }
 
-  private void startClusterManagementService() throws IOException {
+  @VisibleForTesting
+  void startClusterManagementService() throws IOException {
     startConfigurationPersistenceService();
 
-    if (internalCache == null) {
+    InternalCache myCache = this.internalCache;
```
echobravopapa: don't want to mind-read here, what was the rationale for a local copy?

geode-core/src/main/java/org/apache/geode/distributed/internal/InternalLocator.java
```diff
@@ -773,15 +773,19 @@ private void startCache(DistributedSystem system) throws IOException {
...
+    InternalLocator currentLocator = InternalLocator.locator;
+
+    if (myCache == null || currentLocator == null) {
       return;
     }
 
-    clusterManagementService = new LocatorClusterManagementService(locator.internalCache,
-        locator.configurationPersistenceService);
+    clusterManagementService = new LocatorClusterManagementService(currentLocator.internalCache,
```
dschneider-pivotal: should "currentLocator.internalCache" be changed to "myCache"?

geode-core/src/main/java/org/apache/geode/distributed/internal/InternalLocator.java
```diff
@@ -773,15 +773,19 @@ private void startCache(DistributedSystem system) throws IOException {
...
-  private void startClusterManagementService() throws IOException {
+  @VisibleForTesting
+  void startClusterManagementService() throws IOException {
     startConfigurationPersistenceService();
 
-    if (internalCache == null) {
+    InternalCache myCache = this.internalCache;
+    InternalLocator currentLocator = InternalLocator.locator;
+
+    if (myCache == null || currentLocator == null) {
```
dschneider-pivotal: If this expression is true it seems like we just silently returned without starting all of the locator's services. I have seen runs (recently) in which this happens because the InternalDistributedSystem does a concurrent InternalLocator.stop while another thread is executing the restart of this locator. So before returning here should this locator be marked in some way as not having been started? I think the thread that waits for this thread (the one executing startClusterManagementService) just joins and then says we are started.

geode-core/src/main/java/org/apache/geode/distributed/internal/InternalLocator.java
dschneider-pivotal: Since this is an instance method on InternalLocator, why don't we just use this.internalCache and this.configurationPersistenceService instead of trying to get the locator from a static. This method should be initializing itself.

geode-core/src/main/java/org/apache/geode/distributed/internal/InternalLocator.java
bschuchardt: this caches the value to ensure that it doesn't go null after this check

geode-core/src/main/java/org/apache/geode/distributed/internal/InternalLocator.java
bschuchardt: yes.  It looks like Aditya Anchuri added that code early in 2019.  I'll change it.

bschuchardt: yes, I agree.  I'm going to make a pass through InternalLocator and remove the use of the "locator" static wherever possible.  These uses all seem to come from cluster-configuration work.

PR #4674, opened 2020-02-05
GEODE-7356: Fix FixedPartitioningWithColocationAndPersistenceDUnitTest

kirklund:
Main fix for flakiness involved using ResourceObservers. Search for these lines:
```
await().until(() -> observer.isRecoveryFinished());
```
The secondary fix involved using custom enum, serializable classes, etc with proper Atomic fields. The shared classes that this test was using from geode-dunit are not thread safe but were used from multiple threads.

### DIFF
--- a/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/partitioned/fixed/FixedPartitioningWithColocationAndPersistenceDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/partitioned/fixed/FixedPartitioningWithColocationAndPersistenceDUnitTest.java
@@ -14,565 +14,683 @@
  */
 package org.apache.geode.internal.cache.partitioned.fixed;
 
-import static org.junit.Assert.fail;
+import static java.util.Arrays.asList;
+import static java.util.Collections.emptySet;
+import static java.util.Locale.US;
+import static java.util.Objects.requireNonNull;
+import static org.apache.geode.cache.FixedPartitionAttributes.createFixedPartition;
+import static org.apache.geode.distributed.ConfigurationProperties.SERIALIZABLE_OBJECT_FILTER;
+import static org.apache.geode.internal.cache.partitioned.fixed.FixedPartitioningWithColocationAndPersistenceDUnitTest.Month.APR;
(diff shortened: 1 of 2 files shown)


kirklund:
Cross-section testing eventually showed some failures similar to:
```
org.apache.geode.internal.cache.partitioned.fixed.FixedPartitioningWithColocationAndPersistenceDUnitTest > testColocation_FPR_Persistence_Colocation_OneMemberAlive FAILED
    org.apache.geode.test.dunit.RMIException: While invoking org.apache.geode.test.dunit.rules.DistributedRestoreSystemProperties$$Lambda$184/0x0000000840b72040.run in VM 0 running on Host 08134b6f2ec4 with 4 VMs

        Caused by:
        java.rmi.ConnectException: Connection refused to host: 172.17.0.3; nested exception is: 
                java.net.ConnectException: Connection refused (Connection refused)

            Caused by:
            java.net.ConnectException: Connection refused (Connection refused)
```
I don't see a cause for this other than general RMI failure caused by networking problem on the worker.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/partitioned/fixed/FixedPartitioningWithColocationAndPersistenceDUnitTest.java
```diff
@@ -14,565 +14,683 @@
...
+        .addFixedPartitionAttributes(createFixedPartition("30", true, 5))
+        .addFixedPartitionAttributes(createFixedPartition("40", true, 5))
+        .addFixedPartitionAttributes(createFixedPartition("10", false, 5))
+        .addFixedPartitionAttributes(createFixedPartition("20", false, 5))
+        .redundantCopies(2)
+        .localMaxMemory(50)
+        .totalNumBuckets(20)
+        .partitionResolver(new CustomerFixedPartitionResolver<>())
+        .create(CUSTOMERS_REGION));
+
```
mhansonp: minor.... why not just build the list once rather than 3 times...

kirklund:
@mhansonp I already refactored the createRegion method into a builder style inner-class to help reduce redundancy (ie passing lots of false and null parameters). The definition of regions is so similar but actually unique in every test.

PR #4677, opened 2020-02-06
GEODE-7756: Do not use authorization cache for CQs

jujoramos:
CQs are executed on individual entries whenever an event is triggered
for them and the execution context is always cleared before the actual
evaluation, thus using an internal cache to keep already authorized
methods is useless (will always be a cache miss and the computation
required just adds overhead).

- Fixed the CQ creation to always set the 'cqQueryContext' flag.
- Modified the query engine to avoid using the internal cache for
  already authorized methods when the context belongs to a CQ.
- Avoid the creation and concatenation of unnecessary Strings, using
  the Method class directly instead.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/cache/query/internal/AttributeDescriptor.java
+++ b/geode-core/src/main/java/org/apache/geode/cache/query/internal/AttributeDescriptor.java
@@ -34,6 +34,7 @@ import org.apache.geode.cache.EntryDestroyedException;
 import org.apache.geode.cache.query.NameNotFoundException;
 import org.apache.geode.cache.query.QueryInvocationTargetException;
 import org.apache.geode.cache.query.QueryService;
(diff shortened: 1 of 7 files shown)


jujoramos:
Hey @nabarunnag ,

Thanks for reviewing!.
I don't know the answer to your question, though, apparently the flag was not always set on the `QueryExecutionContext` for CQs, that's why I had to add the constructor parameter. On the other hand, the `reset()` method was added for CQs only a long time ago, so we apparently want to reset the context before the actual execution... we can create another ticket and investigate this further, I'm hesitant to change this behaviour right now without knowing all the implications.

```
  /**
   * Added to reset the state from the last execution. This is added for CQs only.
   */
  @Override
  public void reset() {
    super.reset();
    this.execCacheStack.clear();
  }
```

PR #4679, opened 2020-02-06
GEODE-7776 - Proposed changes to remove build cycle

rhoughton-pivot:
There is a build / runtime cycle between geode-core and geode-gfsh, defined as a runtime-only dependency in Gradle. Can we break this optional, runtime-only declaration?

### DIFF
--- a/geode-core/build.gradle
+++ b/geode-core/build.gradle
@@ -263,10 +263,11 @@ dependencies {
   //Log4j is used everywhere
   implementation('org.apache.logging.log4j:log4j-api')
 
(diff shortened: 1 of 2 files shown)


rhoughton-pivot:
> shouldn't there be more to this...?

Like what? They should actually be line deletions.

rhoughton-pivot:
I would like to get @jdeppe-pivotal opinion, as the committer that made the original breaking change.

rhoughton-pivot:
I have a wip on that additional work. Let's collaborate on Monday.

PR #4689, opened 2020-02-11
GEODE-7684: Create messaging for PR Clear

BenjaminPerryRoss:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/partitioned/ClearPRMessage.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/partitioned/ClearPRMessage.java
@@ -64,6 +64,8 @@ public class ClearPRMessage extends PartitionMessageWithDirectReply {
       "The bucket region on target member is no longer primary";
   public static final String BUCKET_REGION_LOCK_UNAVAILABLE_MESSAGE =
       "A lock for the bucket region could not be obtained.";
+  public static final String EXCEPTION_THROWN_DURING_CLEAR_OPERATION =
+      "An exception was thrown during the local clear operation: ";
 
   /**
    * state from operateOnRegion that must be preserved for transmission from the waiting pool
@@ -83,7 +85,7 @@ public class ClearPRMessage extends PartitionMessageWithDirectReply {
     this.posDup = false;
   }
 
-  public void setEventId(RegionEventImpl event) {
+  public void setRegionEvent(RegionEventImpl event) {
     regionEvent = event;
   }
 
@@ -201,8 +203,14 @@ public class ClearPRMessage extends PartitionMessageWithDirectReply {
         throw new ForceReattemptException(BUCKET_NON_PRIMARY_MESSAGE);
       }
 
-      // call new cmnClearRegion on the target bucket region
-      bucketRegion.cmnClearRegion(regionEvent, true, true);
+      try {
+        // call new cmnClearRegion on the target bucket region
+        bucketRegion.cmnClearRegion(regionEvent, true, true);
+      } catch (Exception ex) {
+        throw new ForceReattemptException(
+            EXCEPTION_THROWN_DURING_CLEAR_OPERATION + ex.getClass().getName(), ex);
+      }
+
     } finally {
       lockService.unlock(lockName);
     }
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/partitioned/ClearPRMessage.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/partitioned/ClearPRMessage.java
@@ -1,0 +1,400 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package org.apache.geode.internal.cache.partitioned;
+
+import java.io.DataInput;
+import java.io.DataOutput;
+import java.io.IOException;
+import java.util.Collections;
+import java.util.Set;
+
+import org.apache.logging.log4j.Logger;
+
+import org.apache.geode.DataSerializer;
+import org.apache.geode.cache.CacheException;
+import org.apache.geode.cache.EntryExistsException;
+import org.apache.geode.distributed.DistributedLockService;
+import org.apache.geode.distributed.DistributedMember;
+import org.apache.geode.distributed.internal.ClusterDistributionManager;
+import org.apache.geode.distributed.internal.DirectReplyProcessor;
+import org.apache.geode.distributed.internal.DistributionManager;
+import org.apache.geode.distributed.internal.InternalDistributedSystem;
+import org.apache.geode.distributed.internal.ReplyException;
+import org.apache.geode.distributed.internal.ReplyMessage;
+import org.apache.geode.distributed.internal.ReplyProcessor21;
+import org.apache.geode.distributed.internal.ReplySender;
+import org.apache.geode.distributed.internal.membership.InternalDistributedMember;
+import org.apache.geode.internal.Assert;
+import org.apache.geode.internal.InternalDataSerializer;
+import org.apache.geode.internal.NanoTimer;
+import org.apache.geode.internal.cache.BucketRegion;
+import org.apache.geode.internal.cache.DataLocationException;
+import org.apache.geode.internal.cache.EventID;
+import org.apache.geode.internal.cache.ForceReattemptException;
+import org.apache.geode.internal.cache.PartitionedRegion;
+import org.apache.geode.internal.cache.PartitionedRegionException;
+import org.apache.geode.internal.cache.PartitionedRegionHelper;
+import org.apache.geode.internal.cache.RegionEventImpl;
+import org.apache.geode.internal.logging.log4j.LogMarker;
+import org.apache.geode.internal.serialization.DeserializationContext;
+import org.apache.geode.internal.serialization.SerializationContext;
+import org.apache.geode.logging.internal.log4j.api.LogService;
+
+public class ClearPRMessage extends PartitionMessageWithDirectReply {
+  private static final Logger logger = LogService.getLogger();
+
+  private RegionEventImpl regionEvent;
+
+  private Integer bucketId;
+
+  private Object callbackArg;
+
+  /** The time in ms to wait for a lock to be obtained during doLocalClear() */
+  public static final int LOCK_WAIT_TIMEOUT_MS = 100;
+
+  protected static final short HAS_BRIDGE_CONTEXT = UNRESERVED_FLAGS_START;
+
+  /**
+   * state from operateOnRegion that must be preserved for transmission from the waiting pool
+   */
+  transient boolean result = false;
+
+  /**
+   * Empty constructor to satisfy {@link DataSerializer}requirements
+   */
+  public ClearPRMessage() {}
+
+  public ClearPRMessage(int bucketId, boolean notificationOnly, boolean posDup,
+      Object callbackArg) {
+    this.bucketId = bucketId;
+    this.notificationOnly = notificationOnly;
+    this.posDup = posDup;
+    this.callbackArg = callbackArg;
+    initTxMemberId();
+  }
+
+  public void setEventId(RegionEventImpl event) {
+    regionEvent = event;
+  }
+
+  public void initMessage(PartitionedRegion region, Set<InternalDistributedMember> recipients,
+      boolean notifyOnly, DirectReplyProcessor replyProcessor) {
+    this.resetRecipients();
+    if (recipients != null) {
+      setRecipients(recipients);
+    }
+    this.regionId = region.getPRId();
+    this.processor = replyProcessor;
+    this.processorId = replyProcessor == null ? 0 : replyProcessor.getProcessorId();
+    if (replyProcessor != null && this.isSevereAlertCompatible()) {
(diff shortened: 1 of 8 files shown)


BenjaminPerryRoss:
After talking with Gester - we should remove notify only and the transactional related stuff, make sure that doLocalClear() throws ForceReattemptException instead of PartitionedRegionException, and use a while loop to get the lock/check primary (see lucine IndexRepositoryFactory.java)

geode-core/src/main/java/org/apache/geode/internal/cache/partitioned/ClearPRMessage.java
```diff
@@ -0,0 +1,405 @@
...
+    return regionEvent.getEventId();
+  }
+
+  /**
+   * This method is called upon receipt and make the desired changes to the PartitionedRegion Note:
+   * It is very important that this message does NOT cause any deadlocks as the sender will wait
+   * indefinitely for the acknowledgement
+   */
+  @Override
+  protected boolean operateOnPartitionedRegion(ClusterDistributionManager distributionManager,
```
DonalEvans: This method always returns false. Is that intentional? Also, I'm not sure if this method can actually throw any exceptions, since we catch the `ForceReattemptException` from `doLocalClear()` and the other two don't seem to be thrown by anything we're calling here.

geode-core/src/main/java/org/apache/geode/internal/cache/partitioned/ClearPRMessage.java
```diff
@@ -0,0 +1,405 @@
...
+  public void initMessage(PartitionedRegion region, Set<InternalDistributedMember> recipients,
+      boolean notifyOnly, DirectReplyProcessor replyProcessor) {
+    this.resetRecipients();
+    if (recipients != null) {
+      setRecipients(recipients);
+    }
+    this.regionId = region.getPRId();
+    this.processor = replyProcessor;
+    this.processorId = replyProcessor == null ? 0 : replyProcessor.getProcessorId();
+    if (replyProcessor != null && this.isSevereAlertCompatible()) {
```
DonalEvans: `this.isSevereAlertCompatible()` is always true for this class, so it can probably be removed from this if statement.

geode-core/src/test/java/org/apache/geode/internal/cache/partitioned/ClearPRMessageTest.java
```diff
@@ -0,0 +1,116 @@
...
+import org.junit.Test;
+
+import org.apache.geode.distributed.DistributedLockService;
+import org.apache.geode.internal.cache.BucketRegion;
+import org.apache.geode.internal.cache.ForceReattemptException;
+import org.apache.geode.internal.cache.PartitionedRegion;
+import org.apache.geode.internal.cache.PartitionedRegionDataStore;
+import org.apache.geode.internal.cache.PartitionedRegionException;
+
+public class ClearPRMessageTest {
```
DonalEvans: It might be worth including tests to confirm that `initMessage()` correctly sets the `processorId` and doesn't call `enableSevereAlertProcessing()` on the `ReplyProcessor` if the processor is null, a test to confirm that `send()` throws an exception if `putOutgoing(this)` returns a non-null set, a test to confirm that `operateOnPartitionedRegion()` calls `sendReply()` with the correct arguments if a `ForceReattemptException` is thrown during `doLocalClear()`, a test to confirm that we call `endPartitionMessagesProcessing()` in `sendReply()` when appropriate and a test to confirm that we only call `replyProcessor.process()` in `ClearReplyMessage.process()` when the replyProcessor is not null.

geode-core/src/main/java/org/apache/geode/internal/cache/partitioned/ClearPRMessage.java
BenjaminPerryRoss: OperateOnPartitionedRegion() is invoked by the process() method on the message class which is defined in the parent, and the return value is used to indicate to the Process() method whether or not it needs to construct and send a reply. Since we construct a reply ourselves in both return cases in this method, we don't need a case which returns true. 

It also doesn't need to throw any ForceReattemptExceptions since it's bundling those in the reply sent from the case in which those are thrown by the doLocalClear() method.

geode-core/src/main/java/org/apache/geode/internal/cache/partitioned/ClearPRMessage.java
```diff
@@ -0,0 +1,405 @@
...
+  private Object callbackArg;
+
+  /** The time in ms to wait for a lock to be obtained during doLocalClear() */
+  public static final int LOCK_WAIT_TIMEOUT_MS = 100;
+  public static final String BUCKET_NON_PRIMARY_MESSAGE =
+      "The bucket region on target member is no longer primary";
+  public static final String BUCKET_REGION_LOCK_UNAVAILABLE_MESSAGE =
+      "A lock for the bucket region could not be obtained.";
+
+  protected static final short HAS_BRIDGE_CONTEXT = UNRESERVED_FLAGS_START;
```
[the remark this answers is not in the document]
BenjaminPerryRoss: Done.

geode-core/src/main/java/org/apache/geode/internal/cache/partitioned/ClearPRMessage.java
```diff
@@ -0,0 +1,405 @@
...
+import org.apache.geode.logging.internal.log4j.api.LogService;
+
+public class ClearPRMessage extends PartitionMessageWithDirectReply {
+  private static final Logger logger = LogService.getLogger();
+
+  private RegionEventImpl regionEvent;
+
+  private Integer bucketId;
+
+  private Object callbackArg;
```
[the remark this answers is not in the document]
BenjaminPerryRoss: Done.

geode-core/src/main/java/org/apache/geode/internal/cache/partitioned/ClearPRMessage.java
```diff
@@ -0,0 +1,405 @@
...
+  /**
+   * Empty constructor to satisfy {@link DataSerializer}requirements
+   */
+  public ClearPRMessage() {}
+
+  public ClearPRMessage(int bucketId, boolean notificationOnly, boolean posDup,
+      Object callbackArg) {
+    this.bucketId = bucketId;
+    this.notificationOnly = notificationOnly;
+    this.posDup = posDup;
```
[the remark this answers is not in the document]
BenjaminPerryRoss: I don't think we should remove this entirely as they are fields of the parent class and are called by methods in the parent class that will end up running when this message is sent - however, as we never need this to be anything but false, I have changed it to automatically set this value to false in the constructer, rather than allowing a custom value for each instance of the message via the constructor.

geode-core/src/main/java/org/apache/geode/internal/cache/partitioned/ClearPRMessage.java
```diff
@@ -0,0 +1,405 @@
...
+
+  /**
+   * Empty constructor to satisfy {@link DataSerializer}requirements
+   */
+  public ClearPRMessage() {}
+
+  public ClearPRMessage(int bucketId, boolean notificationOnly, boolean posDup,
+      Object callbackArg) {
+    this.bucketId = bucketId;
+    this.notificationOnly = notificationOnly;
```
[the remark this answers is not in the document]
BenjaminPerryRoss: This is the same as the posDup above, always false, but needed by the parent. I've set it to false in the constructor as well.

geode-core/src/main/java/org/apache/geode/internal/cache/partitioned/ClearPRMessage.java
```diff
@@ -0,0 +1,405 @@
...
+  @Override
+  public void toData(DataOutput out, SerializationContext context) throws IOException {
+    super.toData(out, context);
+    if (bucketId == null) {
+      InternalDataSerializer.writeSignedVL(-1, out);
+    } else {
+      InternalDataSerializer.writeSignedVL(bucketId, out);
+    }
+    DataSerializer.writeObject(this.callbackArg, out);
+    // BR: This may not work properly
```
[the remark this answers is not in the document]
BenjaminPerryRoss: Done.

geode-core/src/main/java/org/apache/geode/internal/cache/partitioned/ClearPRMessage.java
BenjaminPerryRoss: Done.

geode-core/src/main/java/org/apache/geode/internal/cache/partitioned/ClearPRMessage.java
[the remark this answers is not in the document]
DonalEvans: Since they're `boolean` you don't need to explicitly set them to false; they default to that value even if they're not initialized, so we can avoid cluttering the constructor with assignments that are never used by the class.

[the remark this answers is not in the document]
BenjaminPerryRoss: This is true, but I feel expressly setting them is more clear. If they were expressly set to false in the parent class as a default that would be different, but just allowing the compiler to default their values to false kind of hides the fact that this is an intentional value. Our message class will use these values (albeit in methods defined by its parent) and it's important that they're set to false, otherwise we run the risk of improperly processing the message.

geode-core/src/main/java/org/apache/geode/internal/cache/partitioned/ClearPRMessage.java
```diff
@@ -0,0 +1,405 @@
...
+    this.notificationOnly = notifyOnly;
+  }
+
+  @Override
+  public boolean isSevereAlertCompatible() {
+    // allow forced-disconnect processing for all cache op messages
+    return true;
+  }
+
+  public void setPossibleDuplicate(boolean posDup) {
```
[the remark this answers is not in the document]
BenjaminPerryRoss: Done.

geode-core/src/test/java/org/apache/geode/internal/cache/partitioned/ClearPRMessageTest.java
BenjaminPerryRoss: I added all these tests - a few of them were just making sure that methods weren't called on fields that were null, and these tests felt a little unnecessary (they pretty much just involved calling the method and relying on an NPE to fail) but I've added them anyway just to be safe. Let me know if you think these are unnecessary.

geode-core/src/test/java/org/apache/geode/internal/cache/partitioned/ClearPRMessageTest.java
```diff
@@ -0,0 +1,294 @@
...
+    assertThatThrownBy(() -> message.doLocalClear(region))
+        .isInstanceOf(ForceReattemptException.class)
+        .hasMessageContaining(ClearPRMessage.EXCEPTION_THROWN_DURING_CLEAR_OPERATION);
+
+    // Confirm that cmnClearRegion was called
+    verify(bucketRegion, times(1)).cmnClearRegion(any(), anyBoolean(), anyBoolean());
+
+    // Confirm that we actually obtained and released the lock
+    verify(mockLockService, times(1)).lock(any(), anyLong(), anyLong());
+    verify(mockLockService, times(1)).unlock(any());
```
DonalEvans: This test isn't trying to verify anything about whether we do or don't obtain locks, just the behaviour when we get an exception from cmnClearRegion(), so these last couple of verifies aren't needed.

geode-core/src/test/java/org/apache/geode/internal/cache/partitioned/ClearPRMessageTest.java
```diff
@@ -0,0 +1,294 @@
...
+  }
+
+  @Test
+  public void doLocalClearInvokesCmnClearRegionWhenBucketIsPrimaryAndLockIsObtained()
+      throws ForceReattemptException {
+    DistributedLockService mockLockService = mock(DistributedLockService.class);
+    doReturn(mockLockService).when(message).getPartitionRegionLockService();
+
```
DonalEvans: There's an extra line here that could be tidied up.

geode-core/src/main/java/org/apache/geode/internal/cache/partitioned/ClearPRMessage.java
```diff
@@ -0,0 +1,389 @@
...
+      }
+
+      // Double check if we are still primary, as this could have changed between our first check
+      // and obtaining the lock
+      if (!bucketRegion.isPrimary()) {
+        throw new ForceReattemptException(BUCKET_NON_PRIMARY_MESSAGE);
+      }
+
+      try {
+        // call new cmnClearRegion on the target bucket region
```
DonalEvans: I think this comment is probably no longer needed, as it was originally just a placeholder and the thing it's describing is very self-explanatory.

geode-core/src/test/java/org/apache/geode/internal/cache/partitioned/ClearPRMessageTest.java
```diff
@@ -0,0 +1,294 @@
...
+        .isInstanceOf(ForceReattemptException.class)
+        .hasMessageContaining(ClearPRMessage.BUCKET_NON_PRIMARY_MESSAGE);
+    // Confirm that we actually obtained and released the lock
+    verify(mockLockService, times(1)).lock(any(), anyLong(), anyLong());
+    verify(mockLockService, times(1)).unlock(any());
+  }
+
+  @Test
+  public void doLocalClearThrowsForceReattemptExceptionWhenAnExceptionIsThrownDuringClearOperation()
+      throws ForceReattemptException {
```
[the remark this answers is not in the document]
BenjaminPerryRoss: Done.

geode-core/src/test/java/org/apache/geode/internal/cache/partitioned/ClearPRMessageTest.java
BenjaminPerryRoss: Done.

geode-core/src/test/java/org/apache/geode/internal/cache/partitioned/ClearPRMessageTest.java
BenjaminPerryRoss: Done.

geode-core/src/main/java/org/apache/geode/internal/cache/partitioned/ClearPRMessage.java
BenjaminPerryRoss: Done.

PR #4690, opened 2020-02-11
GEODE-7746: onServers function throws a NPE if the distributed system is shutdown

jchen21:

### For all changes:

### Note:


jchen21:
Will squash the commits when merging to develop

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/execute/FunctionExecutionDUnit.java
```diff
@@ -0,0 +1,96 @@
...
+    public String getId() {
+      return getClass().getSimpleName();
+    }
+  }
+
+  @Rule
+  public ClusterStartupRule cluster = new ClusterStartupRule();
+
+  @Test
+  public void test() throws Exception {
```
jhuynh1: perhaps a more detailed test name describing what the test is trying to test for?

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/execute/FunctionExecutionDUnit.java
```diff
@@ -0,0 +1,96 @@
...
+    AsyncInvocation async = client.invokeAsync(() -> {
+      ClientCache clientCache = ClusterStartupRule.getClientCache();
+      while (true) {
+        FunctionService.onServers(clientCache).execute("TestFunction").getResult();
+      }
+    });
+    server.invoke(() -> {
+      Region region = null;
+      while (region == null) {
+        region = ClusterStartupRule.getCache().getRegion("testRegion");
```
jhuynh1: Perhaps use awaitility instead of spinning?

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/execute/FunctionExecutionDUnit.java
```diff
@@ -0,0 +1,96 @@
...
+    cluster.stop(1);
+    cluster.stop(0);
+    waitForException(async);
+  }
+
+  private void waitForException(AsyncInvocation async) {
+    try {
+      async.await();
+    } catch (Throwable e) {
+      assertThat(hasCauseType(e, NullPointerException.class)).isFalse();
```
jhuynh1: I think there is a chance these asserts don't get called.  I think you'll have to add a throw on line 90-91 so that if the await() doesn't fail, then the test should fail.

jhuynh1: Also, is there a way we can convert to awaitility so there is no chance a test can hang forever?

jchen21:
Removed the FunctionExecutionDUnit in the previous commits. Since the `NullPointerException` can be reproduced by `ClientServerSessionCacheDUnitTest.startingAClientWithoutServersFails()`.

extensions/geode-modules/src/distributedTest/java/org/apache/geode/modules/util/ClientServerSessionCacheDUnitTest.java
```diff
@@ -105,7 +105,7 @@ public void startingAClientWithoutServersFails() {
     final VM client = VM.getVM(2);
 
     assertThatThrownBy(() -> client.invoke(this::startClientSessionCache))
-        .hasCauseInstanceOf(FunctionException.class);
+        .hasCauseInstanceOf(NoAvailableServersException.class);
```
jhuynh1: Now that this changes the exception thrown... will this affect any other parts of function service?  such as retries?

jchen21: This is a test code change. I don't think it will affect the product behavior of function service. 

When there is no server in the cluster, I don't think there is retry for function service. It should throw `NoAvailableServersException`.

The NullPointerException(NPE) bug (GEODE-7746) can be reproduced by this test from develop branch without any code change. The test originally expects a `FunctionException` which wraps a NPE as its cause. The NPE is exactly GEODE-7746, same stack trace. The purpose of the test to test the case of function execution when there is no server. It should throw `NoAvailableServersException`(`ServerConnectivityException`), instead of `FunctionException` which wraps a NPE. The related catch and throw code is https://github.com/apache/geode/blob/develop/geode-core/src/main/java/org/apache/geode/internal/cache/execute/ServerFunctionExecutor.java#L182-L190
and https://github.com/apache/geode/blob/develop/geode-core/src/main/java/org/apache/geode/internal/cache/execute/ServerFunctionExecutor.java#L227-L235

PR #4692, opened 2020-02-12
GEODE-7799: Distribute rebalance status to other locators

dschneider-pivotal:

### For all changes:

### Note:


dschneider-pivotal:
@yozaner1324

geode-management/src/main/java/org/apache/geode/management/api/ClusterManagementOperation.java
```diff
@@ -29,12 +31,15 @@
  * @param <V> the result type of the operation
  */
 @SuppressWarnings("unused")
-public interface ClusterManagementOperation<V extends OperationResult> {
+public interface ClusterManagementOperation<V extends OperationResult>
```
jinmeiliao: I know this isn't introduced by this PR, but now I am wondering: V really isn't used by this interface, why do we need to declare it here?

geode-assembly/src/distributedTest/java/org/apache/geode/management/internal/rest/RebalanceManagementDunitTest.java
```diff
@@ -36,92 +37,110 @@
...
     assertThat(end).isBetween(now - 60000, now)
-        .isGreaterThanOrEqualTo(cmr.getOperationStart().getTime());
+        .isGreaterThanOrEqualTo(endResult.getOperationStart().getTime());
+    RebalanceResult result = endResult.getOperationResult();
     assertThat(result.getRebalanceRegionResults().size()).isEqualTo(2);
     RebalanceRegionResult firstRegionSummary = result.getRebalanceRegionResults().get(0);
     assertThat(firstRegionSummary.getRegionName()).isIn("customers1", "customers2");
   }
 
+  private ClusterManagementOperationResult<RebalanceOperation, RebalanceResult> waitForStartToEnd(
```
jinmeiliao: this is something that I had hoped that could be built in the result that's coming back from the "start" operation.

geode-core/src/main/java/org/apache/geode/internal/cache/RegionFactoryImpl.java
```diff
@@ -43,4 +53,22 @@ public RegionFactoryImpl(InternalCache cache, String regionAttributesId) {
...
     super(regionFactory);
   }
+
+  public void setInternalRegionArguments(
+      InternalRegionArguments internalRegionArguments) {
+    this.internalRegionArguments = internalRegionArguments;
+  }
+
+  @Override
+  public Region<K, V> create(String name)
```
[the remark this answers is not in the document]
dschneider-pivotal: That is a good idea but I think we should do it in another pull request. I will file a refactor ticket about not using the deprecated AttributesFactory to create internal regions. So this PR introduces the new RegionFactoryImpl but will only use it in the new code in this PR. The other ticket can look for other places in the code (like the one you mention) that use deprecated apis to create internal regions.

geode-management/src/main/java/org/apache/geode/management/internal/ClusterManagementOperationStatusResult.java
```diff
@@ -23,6 +23,7 @@
 import org.apache.geode.management.runtime.OperationResult;
 
 @Experimental
+// TODO can this class be removed?
```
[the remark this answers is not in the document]
dschneider-pivotal: Thanks for catching this! We intended to remove it but forgot to do so

geode-assembly/src/distributedTest/java/org/apache/geode/management/internal/rest/ClientClusterManagementServiceDunitTest.java
```diff
@@ -115,7 +119,8 @@ public void createRegionWithGroup() {
 
     locator.invoke(() -> {
       InternalConfigurationPersistenceService persistenceService =
-          ClusterStartupRule.getLocator().getConfigurationPersistenceService();
+          Objects.requireNonNull(ClusterStartupRule.getLocator())
```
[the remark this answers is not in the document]
jmelchio: > why do we need this requireNonNull here?

It's just to reduce the warnings you get in the code.

geode-management/src/test/java/org/apache/geode/management/api/ClusterManagementOperationResultTest.java
```diff
@@ -37,20 +39,32 @@ public void setUp() {
...
-  static class TestOperationResult implements OperationResult {
+  @Test
+  public void serializeRebal() throws Exception {
+    ClusterManagementResult result1 = new ClusterManagementResult();
+    result1.setStatus(StatusCode.OK, "Success!!");
+    RebalanceOperation rebalanceOperation = new RebalanceOperation();
+    rebalanceOperation.setOperator("operator");
+    ClusterManagementOperationResult<RebalanceOperation, RebalanceResult> result =
+        new ClusterManagementOperationResult(result1, new Date(), new Date(),
+            rebalanceOperation, "id", new RebalanceResultImpl(), null);
```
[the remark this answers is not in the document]
jmelchio: Yes, let us look at that. Agree that given that we're extending an interface already in `RebalanceResult` it makes sense to make `RebalanceResult` the implemenation.

geode-management/src/main/java/org/apache/geode/management/api/ClusterManagementOperation.java
jmelchio: > I know this isn't introduced by this PR, but now I am wondering: V really isn't used by this interface, why do we need to declare it here?

We need this to maintain the coupling between the operation and the result in the implementations of `ClusterManagementOperation`.

geode-management/src/test/java/org/apache/geode/management/api/ClusterManagementOperationResultTest.java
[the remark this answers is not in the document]
dschneider-pivotal: This PR did not introduce RebalanceResult nor RebalanceResultImpl.
I prefer keeping the current interface and having an internal implementation class. This prevents our external apis from having internal things exposed. For example, in this case, the external api does not need a constructor nor settors.
I think it would be good to have more of our external api contract defined by interfaces and try to minimize the number of classes in it.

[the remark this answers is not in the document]
dschneider-pivotal: ps: I'm fine with some other name for the class that implements the RebalanceResult.
I also find it odd that RebalanceResult has "getSuccess()" and "getStatusMessage()" but its super interface "OperationResult" is empty. It seems like both of these would be reasonable things to expect from every OperationResult.

geode-core/src/main/java/org/apache/geode/internal/cache/RegionFactoryImpl.java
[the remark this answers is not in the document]
dschneider-pivotal: I filed: https://issues.apache.org/jira/browse/GEODE-7803 for this

geode-core/src/main/java/org/apache/geode/management/internal/operation/OperationHistoryManager.java
```diff
@@ -38,146 +33,88 @@
...
+  private final OperationStateDistributionService operationStateDistributionService;
 
   /**
    * set a default retention policy to keep results for 2 hours after completion
    */
-  public OperationHistoryManager() {
-    this(2, TimeUnit.HOURS);
+  public OperationHistoryManager(
+      OperationStateDistributionService operationStateDistributionService) {
+    this(2, TimeUnit.HOURS, operationStateDistributionService);
```
jinmeiliao: Gang: please change it to one week.

geode-core/src/main/java/org/apache/geode/management/internal/operation/RegionOperationStateDistributionService.java
```diff
@@ -0,0 +1,111 @@
...
+import org.apache.geode.internal.cache.RegionFactoryImpl;
+import org.apache.geode.management.api.ClusterManagementOperation;
+import org.apache.geode.management.runtime.OperationResult;
+
+public class RegionOperationStateDistributionService
+    implements OperationStateDistributionService {
+  private final Supplier<String> uniqueIdSupplier;
+  private final Region<String, OperationState<ClusterManagementOperation<OperationResult>, OperationResult>> region;
+
+  public static final String OPERATION_STATE_REGION_NAME = "OperationStateRegion";
```
jinmeiliao: internal region needs to be started with "_" or "__".

geode-core/src/main/java/org/apache/geode/management/internal/operation/OperationHistoryManager.java
```diff
@@ -38,146 +33,88 @@
...
   }
 
   /**
    * set a custom retention policy to keep results for X amount of time after completion
    */
-  public OperationHistoryManager(long keepCompleted, TimeUnit timeUnit) {
```
demery-pivotal: Consider using Java's very nice `Duration` and `Instant` classes to carry information about durations and instants.

geode-management/src/test/java/org/apache/geode/management/api/ClusterManagementOperationResultTest.java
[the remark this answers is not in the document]
dschneider-pivotal: I filed a geode ticket about moving those methods up to OperationResult.
The team talked and we decided it would be best to keep the impl class

geode-assembly/src/distributedTest/java/org/apache/geode/management/internal/rest/RebalanceManagementDunitTest.java
dschneider-pivotal: We decided as a team to look into adding a way for the java api user to block until the started op has completed.

geode-core/src/main/java/org/apache/geode/management/internal/operation/RegionOperationStateDistributionService.java
dschneider-pivotal: This has been changed to start with "__". I found a validator that does not allow names to start with "__".

geode-assembly/src/distributedTest/java/org/apache/geode/management/internal/rest/RebalanceManagementDunitTest.java
dschneider-pivotal: We added "getFuture" to the ClusterManagementService. We thought this was cleaner than adding "getFuture" to ClusterManagementOperationResult because if we did that it would need to have a reference to the ClusterManagementService to repeatedly call "get".

geode-core/src/main/java/org/apache/geode/management/internal/operation/OperationHistoryManager.java
dschneider-pivotal: done

geode-core/src/main/java/org/apache/geode/management/internal/operation/OperationHistoryManager.java
dschneider-pivotal: constructor now takes a Duration instead of a long and a TimeUnit

PR #4709, opened 2020-02-14
GEODE-7109: add test case that multiple sessions can be maintained.

pivotal-eshu:
* work around an known issue by verifying after HARegionQueue is drained.

### DIFF
--- a/extensions/session-testing-war/src/main/java/org/apache/geode/modules/session/CommandServlet.java
+++ b/extensions/session-testing-war/src/main/java/org/apache/geode/modules/session/CommandServlet.java
@@ -131,19 +131,19 @@ public class CommandServlet extends HttpServlet {
     if (logQueues) {
       for (Object object : list) {
         for (Object queue : ((Map) object).keySet()) {
-          manager.getLogger().info("client cache has queue: " + queue);
+          System.out.println("client cache has queue: " + queue);
         }
       }
     }
     for (Object object : list) {
       for (Object size : ((Map) object).values()) {
         if ((Integer) size != 0) {
-          manager.getLogger().info("checkQueueDrained not drained with size " + size);
+          System.out.println("checkQueueDrained not drained with size " + size);
           return false;
         }
       }
     }
-    manager.getLogger().info("checkQueueDrained drained");
+    System.out.println("checkQueueDrained drained");
     return true;
   }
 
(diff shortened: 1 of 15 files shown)


geode-assembly/src/distributedTest/java/org/apache/geode/session/tests/TomcatTest.java
```diff
@@ -0,0 +1,267 @@
...
+  }
+
+  private void verifySessionGets(Client client, HashMap<String, String> attributes,
+      Client.Response resp, int operatingContainer) throws IOException, URISyntaxException {
+    int count = 0;
+    int whichOne = random.nextInt(attributes.size());
+    Iterator iterator = attributes.entrySet().iterator();
+    while (iterator.hasNext()) {
+      Map.Entry entry = (Map.Entry) iterator.next();
+      if (count == whichOne) {
```
jchen21: Why not verify all the entries in `attributes`?

extensions/session-testing-war/src/main/java/org/apache/geode/modules/session/CommandServlet.java
```diff
@@ -92,10 +109,47 @@ protected void doGet(HttpServletRequest request, HttpServletResponse response)
...
+    Execution execution = FunctionService.onServers(cache);
+
+    ResultCollector collector = execution.execute(GetQueueSize.ID);
+    List list = (List) collector.getResult();
+    for (Object object : list) {
+      for (Object queue : ((Map) object).keySet()) {
+        manager.getLogger().info("client cache has queue: " + queue);
+      }
+    }
+    for (Object object : list) {
```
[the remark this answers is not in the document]
pivotal-eshu: You are right for this. However, the original intention was to only log the queue names once. Code has been changed to achieve this.

geode-assembly/src/distributedTest/java/org/apache/geode/session/tests/TomcatTest.java
pivotal-eshu: All attributes are verified at the end of test. No need to verify the same attributes after each operation (basically verify same attributes again.) Test will just randomly check one of the attributes and verify all at the end of the test.

onichols-pivotal:

onichols-pivotal:
closed due to lack of activity

PR #4720, opened 2020-02-19
GEODE-6532: Use implementation, not compile, as build configuration

rhoughton-pivot:

### For all changes:

### Note:


rhoughton-pivot:
@PurelyApplied I have walked through the expected-pom.xml changes and looked at what all has moved from `compile` to `runtime` and nothing jumped out as a "But Robert, we use that symbol in a public API" kind of problem. The JIRA does call this out as a potential problem for downstream customers.

rhoughton-pivot:
Just tested against `geode-examples` and saw that `commons-lang3` is no longer being emitted by us, and was expected by the example. Going back to inspect the packages to see if `geode` emits any symbols owned by `commons`

Bill:
@rhoughton-pivot I pulled this branch and built it in IntelliJ and ran some JUnit tests in IntelliJ and didn't see any hiccups ✓

making some comments inline…

geode-gfsh/build.gradle
```diff
@@ -21,26 +21,33 @@ apply from: "${project.projectDir}/../gradle/publish-java.gradle"
...
     api(platform(project(':boms:geode-all-bom')))
-    compile(project(':geode-core'))
-    compile(project(':geode-logging'))
-    compile(project(':geode-membership'))
-    compile(project(':geode-serialization'))
-    compile(project(':geode-unsafe'))
 
-    compile('com.healthmarketscience.rmiio:rmiio')
+    api(project(':geode-core'))
+    api(project(':geode-common'))
```
Bill: why `api` and not merely `implementation` here?

rhoughton-pivot: In geode-gfsh, the non-internal, public abstract class `org.apache.geode.management.cli.GfshCommand` has public methods such as `authorize` that takes a `ResourePermission.Resource`, defined in the  `geode-core` project. My understanding is, this makes `geode-core` an API-level dependency of `geode-gfsh`.

Similar reasoning for `geode-common`, in that it uses `org.apache.geode.annotations.Immutable` in a non-internal abstract class.

Bill: Golly, I didn't know that `GfshCommand` was meant to be extended by users. I see it's in a non-internal package. Does that mean we really have to actually support users extending that thing? @jinmeiliao do you have an opinion? Seems to me like it was a mistake to make `GfshCommand` non-internal.

onichols-pivotal: for a long time there was a vision to make Geode more modular, with each module providing its own gfsh commands (the ability for users to also write their own gfsh commands is sort of a side-effect of this...I too am curious if any customers actually have done so).  If we wanted to change course and say that gfsh is a fixed, non-extensible set of commands, we could also get rid of fast-classpath-scanner and a lot of other baggage...

jinmeiliao: Yes, gfsh commands are meant to be extensible, you can write your own gfsh commands and have them loaded in the command executor. `GfshCommand` is public for that reason.

Bill: ok thanks. @rhoughton-pivot this code is fine.

PR #4726, opened 2020-02-22
Revert "GEODE-7727: modify sender thread to detect release of connect…

DonalEvans:
…ion (#4629)"

In internal tests, this commit was identified as causing deadlocks. Until the root cause can be discovered and addressed, we propose reverting the change.

This reverts commit 1a0d9769e482f49e0c725c0d6adc75d324f88958.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/distributedTest/java/org/apache/geode/internal/tcp/TCPConduitDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/internal/tcp/TCPConduitDUnitTest.java
@@ -16,7 +16,6 @@ package org.apache.geode.internal.tcp;
 
 import static org.apache.geode.test.awaitility.GeodeAwaitility.await;
 import static org.apache.geode.test.util.ResourceUtils.createTempFileFromResource;
-import static org.assertj.core.api.Assertions.assertThat;
 
 import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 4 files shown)


mivanac:
Hi,

can you share some info regarding this internal tests, so we can speed up detection of problem in this PR?

Thanks

DonalEvans:
> Hi,
> 
> can you share some info regarding this internal tests, so we can speed up detection of problem in this PR?
> 
> Thanks

Hi Mario,

The purpose of the internal test is to force async message queueing and confirm that no deadlocks occur. The test in question starts 6 servers with 5 regions in the cluster, each using a custom CacheListener that sleeps the thread for 2 ms when onCreate() is called with a particular event key in order to encourage queueing. Random events are delivered to the regions, including ones using the key that will trigger a 2ms sleep.

Each server is configured to run with 5 threads, conserve sockets is set to true, async distribution timeout is 5ms, async queue timeout is 60000ms and async max queue size is 10.

We consistently see the test hang when two members log that they are sleeping due to the onCreate() listener call. If only one member has logged that call then the test continues regardless of how many times its called, but as soon as a second member logs it, the test encounters problems and a hang is declared 5 minutes later.

We will continue to investigate this issue after the weekend, but I hope that the information here is helpful to you in the meantime.

mivanac:
Hi Donal,

thanx for info.
But if you have any failing tests, that are reason for revert, could you push them to community.
With current approach, it is hard to troubleshoot, and this way all will be transparent.

Regards.

DonalEvans:
> Hi Donal,
> 
> thanx for info.
> But if you have any failing tests, that are reason for revert, could you push them to community.
> With current approach, it is hard to troubleshoot, and this way all will be transparent.
> 
> Regards.

Hi Mario,

Regrettably, I'm not able to share the actual test that caught this problem due to licensing issues; it's a proprietary test used as part of ensuring the product I work on is enterprise-ready. We've been trying to reproduce the issue in a DUnit test that can be shared with the community but haven't been able to do so yet, as the exact circumstances needed to cause it to show up are proving elusive.

I will be certain to let you know as soon as I have a way of reproducing the issue locally and will push any tests to a branch on my fork to allow debugging, as well as creating a Jira ticket with all the details.

DonalEvans:
> Hi Donal,
> 
> thanx for info.
> But if you have any failing tests, that are reason for revert, could you push them to community.
> With current approach, it is hard to troubleshoot, and this way all will be transparent.
> 
> Regards.

A JIRA ticket has now been created for this issue, containing a test that reliably reproduces the issue: https://issues.apache.org/jira/browse/GEODE-7817
It turns out that the problem only presents itself when running the test with JDK 11, which is why it took so long to reproduce locally. Now that a reliable reproduction has been provided, I will be reverting the original change.

PR #4727, opened 2020-02-24
GEODE-7421: deploy using management rest api

jinmeiliao:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/management/internal/api/LocatorClusterManagementService.java
+++ b/geode-core/src/main/java/org/apache/geode/management/internal/api/LocatorClusterManagementService.java
@@ -45,6 +45,7 @@ import org.apache.geode.annotations.VisibleForTesting;
 import org.apache.geode.cache.execute.Execution;
 import org.apache.geode.cache.execute.Function;
 import org.apache.geode.cache.execute.FunctionService;
+import org.apache.geode.cache.execute.ResultCollector;
 import org.apache.geode.distributed.DistributedMember;
 import org.apache.geode.distributed.internal.InternalConfigurationPersistenceService;
 import org.apache.geode.internal.cache.InternalCache;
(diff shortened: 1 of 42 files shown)


geode-core/src/main/java/org/apache/geode/management/internal/api/LocatorClusterManagementService.java
```diff
@@ -202,11 +214,19 @@ public LocatorClusterManagementService(InternalConfigurationPersistenceService p
...
 
     // persist configuration in cache config
+    List<String> updatedGroups = new ArrayList<>();
+    List<String> failedGroups = new ArrayList<>();
     for (String groupName : groups) {
-      configurationManager.add(config, groupName);
+      try {
+        configurationManager.add(config, groupName);
+        updatedGroups.add(groupName);
+      } catch (Exception e) {
```
dschneider-pivotal: Could we make some of the new "catch (Exception ...)" more specific? Do we not know what exceptions we expect configurationManager.add to throw?

geode-core/src/main/java/org/apache/geode/management/internal/api/LocatorClusterManagementService.java
```diff
@@ -499,17 +542,50 @@ public void close() {
...
+      file = ((HasFile) configuration).getFile();
+    }
+
+    for (DistributedMember member : targetMembers) {
+      RemoteInputStream remoteInputStream = null;
+      if (file != null && exporter != null) {
+        try {
+          remoteInputStream = exporter
+              .export(new SimpleRemoteInputStream(new FileInputStream(file.getAbsolutePath())));
+        } catch (Exception e) {
```
dschneider-pivotal: another catch Exception to take a look at

geode-core/src/main/java/org/apache/geode/management/internal/configuration/mutators/ConfigurationManager.java
```diff
@@ -27,9 +27,9 @@
 @Experimental
 public interface ConfigurationManager<T extends AbstractConfiguration> {
 
-  void add(T config, String groupName);
+  void add(T config, String groupName) throws Exception;
```
dschneider-pivotal: Could we make this more specific instead of "throws Exception"?

geode-core/src/test/java/org/apache/geode/management/internal/configuration/validators/DeploymentValidatorTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+  @Test
+  public void validateCreateWithIncorrectJarFile() {
+    deployment.setFile(noJar);
+
+    Throwable throwable = catchThrowable(() -> deploymentValidator.validate(CREATE, deployment));
+
+    assertThat(throwable).isNotNull();
+    assertThat(throwable).isInstanceOf(IllegalArgumentException.class);
+    assertThat(throwable).hasMessageContaining("File does not contain valid JAR content:");
```
demery-pivotal: You can chain assertions if you like:
```
assertThat(throwable)
    .isInstanceOf(IllegalArgumentException.class);
    .hasMessageContaining("File does not contain valid JAR content:");
```
Also note that AssertJ's assertions do a null check, so you don't need `isNotNull()` unless it helps to clarify the test.

geode-core/src/main/java/org/apache/geode/management/internal/api/LocatorClusterManagementService.java
jinmeiliao: it depends on each configuration manager's implementation. I don't think we can dictate on what kind of exception it can throw. I added this signature here so that the handling of the exception falls on the caller, not the implementer.

geode-core/src/main/java/org/apache/geode/management/internal/configuration/mutators/ConfigurationManager.java
jinmeiliao: as commented before.

geode-core/src/main/java/org/apache/geode/management/internal/api/LocatorClusterManagementService.java
dschneider-pivotal: If this is an SPI then this interface should dictate what the implementor needs to do. It already says you have to implement an "add" method etc. It s reasonable if we want to support exceptions from some of these methods to say so on the interface and then the implementer has to make sure and throw that type of exception. That would let the implementor explicitly say "I could not do an add for the following reason" and we could treat that as a handled exception. Other exceptions (like say an NPE caused by their coding error) could then be highlighted as unexpected. But with this code we have to treat all of them as unexpected and they end up with a big call stack in the log which is not needed if it is just something like "I did not have file system permissions".

jinmeiliao:
closing this for PR checks.

PR #4728, opened 2020-02-24
GEODE-7810: Change alert listener connection warning from fatal

kirklund:
Change from FATAL log level to WARNING log level for log statement with stack trace when unable to create a connection to an alert listener.

Note to reviewers: let me know if you think this should be ERROR log level instead of warning. I picked WARNING because I think that FATAL is too high. This doesn't prevent any operations or represent data corruption or the inability to continue to run as a data server. It just means that we couldn't create a new connection for a remote alert listener (probably because the member left or is about to depart the view).


kirklund:
AnalyzeSerializablesJUnitTest needs to be updated for AlertingIOException.

kirklund:
DistributedTest had two unrelated failures:
* DistributedMemberDUnitTest
* DestroyRegionDuringGIIDistributedTest

PR #4736, opened 2020-02-26
GEODE-7839: add gfsh oauth token support

dschneider-pivotal:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/security/shiro/JMXShiroAuthenticator.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/security/shiro/JMXShiroAuthenticator.java
@@ -48,13 +48,16 @@ public class JMXShiroAuthenticator implements JMXAuthenticator, NotificationList
 
   @Override
   public Subject authenticate(Object credentials) {
-    String username = null;
+    String username;
     Properties credProps = new Properties();
     if (credentials instanceof String[]) {
       final String[] aCredentials = (String[]) credentials;
       username = aCredentials[0];
       credProps.setProperty(ResourceConstants.USER_NAME, aCredentials[0]);
       credProps.setProperty(ResourceConstants.PASSWORD, aCredentials[1]);
+    } else if (credentials instanceof String) {
+      username = "";
+      credProps.setProperty(ResourceConstants.TOKEN, (String) credentials);
     } else {
       throw new AuthenticationFailedException(MISSING_CREDENTIALS_MESSAGE);
     }
--- a/geode-core/src/main/java/org/apache/geode/internal/security/shiro/JMXShiroAuthenticator.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/security/shiro/JMXShiroAuthenticator.java
@@ -56,7 +56,7 @@ public class JMXShiroAuthenticator implements JMXAuthenticator, NotificationList
       credProps.setProperty(ResourceConstants.USER_NAME, aCredentials[0]);
       credProps.setProperty(ResourceConstants.PASSWORD, aCredentials[1]);
     } else if (credentials instanceof String) {
-      username = "";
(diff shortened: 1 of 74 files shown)


dschneider-pivotal:
@jmelchio

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/ConnectCommand.java
```diff
@@ -124,6 +126,14 @@ public ResultModel connect(
       useSsl = true;
     }
 
+    if ("".equals(token)) {
+      return ResultModel.createError("--token requires a value, for example --token=foo");
+    }
+
```
jinmeiliao: Like we discussed in the meeting, maybe checks for empty strings for username and password here as well, so that you don't need to deal with them anymore.

geode-core/src/main/java/org/apache/geode/internal/security/shiro/JMXShiroAuthenticator.java
```diff
@@ -48,13 +48,16 @@ public JMXShiroAuthenticator(SecurityService securityService) {
...
-    String username = null;
+    String username;
     Properties credProps = new Properties();
     if (credentials instanceof String[]) {
       final String[] aCredentials = (String[]) credentials;
       username = aCredentials[0];
       credProps.setProperty(ResourceConstants.USER_NAME, aCredentials[0]);
       credProps.setProperty(ResourceConstants.PASSWORD, aCredentials[1]);
+    } else if (credentials instanceof String) {
+      username = "";
```
jinmeiliao: this probably is not good. maybe use a unique string based on the token would be nice.

geode-core/src/main/java/org/apache/geode/internal/security/shiro/JMXShiroAuthenticator.java
```diff
@@ -48,13 +48,16 @@ public JMXShiroAuthenticator(SecurityService securityService) {
...
   public Subject authenticate(Object credentials) {
-    String username = null;
+    String username;
     Properties credProps = new Properties();
     if (credentials instanceof String[]) {
       final String[] aCredentials = (String[]) credentials;
       username = aCredentials[0];
       credProps.setProperty(ResourceConstants.USER_NAME, aCredentials[0]);
       credProps.setProperty(ResourceConstants.PASSWORD, aCredentials[1]);
+    } else if (credentials instanceof String) {
```
jinmeiliao: did we decide to do something about this now or later?

dschneider-pivotal: I think we should do this later when we need it. It is not clear that we will ever need a more general container (for example a Map) of these properties.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/ConnectCommand.java
dschneider-pivotal: I think the current code on this branch is correct. We were very careful to not change the existing behavior of --user and --password in order to be backwards compatible. But we also wanted to be able to tell if both --token and --user or --password were used together. That is why we added a specified default of "" for --user and --password. But the old behavior if you specified --user with no value was to just ignore it (it was not flagged as an error). If we do want to say this old behavior is a bug that needs to be fixed lets do that with its own ticket instead of trying to add it into this PR that is about adding --token support.

geode-core/src/main/java/org/apache/geode/internal/security/shiro/JMXShiroAuthenticator.java
dschneider-pivotal: I addressed this by setting username to "token-" + credential.hashCode()

PR #4738, opened 2020-02-26
Deploy3

jinmeiliao:

### For all changes:

### Note:


jinmeiliao:
> dschneider-pivotal 20 hours ago Member
If this is an SPI then this interface should dictate what the implementor needs to do. It already says you have to implement an "add" method etc. It s reasonable if we want to support exceptions from some of these methods to say so on the interface and then the implementer has to make sure and throw that type of exception. That would let the implementor explicitly say "I could not do an add for the following reason" and we could treat that as a handled exception. Other exceptions (like say an NPE caused by their coding error) could then be highlighted as unexpected. But with this code we have to treat all of them as unexpected and they end up with a big call stack in the log which is not needed if it is just something like "I did not have file system permissions".

I think I am trying to protect the system from any unhandled exceptions, even if it's NPE, since this is other people's code. this way, at least on our side, we logged it and know exactly what went wrong.

PR #4751, opened 2020-03-01
GEODE-7727: modify sender thread to detect relese of connection

mivanac:

### For all changes:

### Note:


geode-core/src/distributedTest/java/org/apache/geode/management/DistributedSystemMXBeanWithAlertsDistributedTest.java
```diff
@@ -137,6 +137,7 @@ public void setUp() throws Exception {
     memberVM3 = getVM(3);
 
     managerMember = managerVM.invoke(() -> createManager());
+    IgnoredException.addIgnoredException("Cannot form connection to alert listener");
```
DonalEvans: What is the reason for adding this as an ignored exception?  Why do we now expect to see this exception and feel okay ignoring it?

upthewaterspout:
I have couple questions about this PR. The original JIRA says:

> The connection is left hanging (CLOSE-WAIT state). The peers are then unable to establish proper P2P communication later. 

Can you clarify what "unable to establish proper P2P communication later" means? If this is the actual bug, it would be nice to write a test that demonstrates how peers later cannot establish communications.

The test included with PR doesn't seem to test this case. It starts a locator and 3 servers. It then uses a test utility to "crash" the 3 servers. It then checks a connection count statistic, but it stops there. 

In addition, by crashing the 3 servers, the test is actually causing the locator to detect a network partition and shut itself down. So the test is actually testing a connection count on an object that is no longer really valid or in use.

From my understanding, the connection *should* already get cleaned up when the peer crashes or closes the connection by a couple of mechanisms:
* The *writer* thread may detect that the connection is closed and clean things up
* The membership system will notice a departed member and trigger cleanup of all of the connections to that member.

If there is a bug where we have problems connecting later, maybe it might involve fixing a problem with one of the above mechanisms? By introducing a third way to clean up the connection it seems like we might be introducing more possibilities for race conditions in maintaining the connection state.

geode-core/src/distributedTest/java/org/apache/geode/internal/tcp/TCPConduitDUnitTest.java
```diff
@@ -113,6 +116,88 @@ public void basicAcceptConnection() throws Exception {
...
+          }
+        });
+      }
+      assertThat(ConnectionTable.getNumSenderSharedConnections()).isEqualTo(0);
+      DistributedTestUtils.crashDistributedSystem(system);
+    }
+  }
+
+  @Test
+  public void testAsyncHanging() throws Exception {
```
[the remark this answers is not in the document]
mivanac: await() functionality has default timeout of 5 minutes, which as I was told should not be changed.

geode-core/src/distributedTest/java/org/apache/geode/internal/tcp/TCPConduitDUnitTest.java
```diff
@@ -113,6 +116,88 @@ public void basicAcceptConnection() throws Exception {
     }
   }
 
+
+  @Test
+  public void testClosingOfSenderConnections() throws Exception {
```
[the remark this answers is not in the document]
mivanac: This added assert is for testing of modification added in this PR.

geode-core/src/distributedTest/java/org/apache/geode/management/DistributedSystemMXBeanWithAlertsDistributedTest.java
mivanac: Due to previous solution, sender thread did not detect release of connection. So this is the reason why this exception did not appear in test logs.

mivanac:
Just to describe faulty scenario. In our network configuration, we have proxy between sites. So if some proxy resets connections, sender side will not detect that, because sender thread is terminated after handshake is finished. And there will be no reestablishing of connection, since sender do not know that connection is released.

upthewaterspout:
> In our network configuration, we have proxy between sites.

This PR is changing peer-to-peer connections, not client server or WAN connections. Do you have a proxy between peers? Asking because I see other PRs related to WAN proxies.

> And there will be no reestablishing of connection, since sender do not know that connection is released.

This is what I'm not quite sure I understand here. I believe when a thread tries to send on this closed P2P socket, it should get an exception and retry, recreating the connection. This seems like what we need to be testing as part of this change - is there actually a problem sending another message if a proxy between the peers closes the socket?

mivanac:
Yes you are right. Proxy is between peers, and after proxy resets connections,  sender connection hangs in CLOSE-WAIT state. After that when sending any message over that connection, there is no exception. So our fix was to detect closing of sender connection.

upthewaterspout:
I thought I would try to write the kind of test I was talking about so I could see this bug reproduce. It did reproduce the issue you described - https://github.com/Nordix/geode/pull/5.

I'm still looking into whether this is going to cause any race conditions, I'm a little worried that the Connection is left in a state of READING during this time, when the writer threads are also modifying the state.

mivanac:
Thanks

geode-core/src/distributedTest/java/org/apache/geode/internal/tcp/TCPConduitDUnitTest.java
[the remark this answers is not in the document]
mivanac: This test is removed

geode-core/src/distributedTest/java/org/apache/geode/internal/tcp/ConnectionChangeHangTest.java
```diff
@@ -0,0 +1,78 @@
...
+
+    try {
+      locator.invoke(() -> await("for message to be sent").until(() -> {
+        final SerialAckedMessage serialAckedMessage = new SerialAckedMessage();
+        serialAckedMessage.send(system.getAllOtherMembers(), false);
+        return true;
+      }));
+    } finally {
+
+    }
```
DonalEvans: There is an empty `finally` block here that should either be removed along with the `try`, or have something relevant done in it.

geode-core/src/distributedTest/java/org/apache/geode/internal/tcp/ConnectionChangeHangTest.java
```diff
@@ -0,0 +1,78 @@
...
+
+  private MemberVM startServer(final int vmIndex) {
+    return cluster.startServerVM(
+        vmIndex, s -> s.withConnectionToLocator(locator.getPort())
+            .withProperty("async-distribution-timeout", "5"));
+  }
+
+  @Test
+
+  public void testNoHanging() {
```
DonalEvans: This test name would be more descriptive if it was something like "serverStartupDoesNotHangWhenAsyncDistributionTimeoutIsSet"

geode-core/src/distributedTest/java/org/apache/geode/internal/tcp/ConnectionChangeHangTest.java
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  private MemberVM startServer(final int vmIndex) {
+    return cluster.startServerVM(
+        vmIndex, s -> s.withConnectionToLocator(locator.getPort())
+            .withProperty("async-distribution-timeout", "5"));
+  }
+
+  @Test
+
```
DonalEvans: There is an extra line break here

geode-core/src/distributedTest/java/org/apache/geode/internal/tcp/ConnectionChangeHangTest.java
```diff
@@ -0,0 +1,78 @@
...
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+
+import org.apache.geode.distributed.internal.InternalDistributedSystem;
+import org.apache.geode.distributed.internal.SerialAckedMessage;
+import org.apache.geode.test.dunit.rules.ClusterStartupRule;
+import org.apache.geode.test.dunit.rules.MemberVM;
+
+public class ConnectionChangeHangTest implements Serializable {
```
DonalEvans: This class should be renamed. The original name was a placeholder for use while debugging.

PR #4753, opened 2020-03-02
Feature/geode 6536 1: modify borrowConnection singleHop solution

mivanac:

### For all changes:

### Note:


geode-core/src/main/java/org/apache/geode/cache/client/internal/pooling/ConnectionManagerImpl.java
```diff
@@ -301,31 +301,68 @@ public Connection borrowConnection(long acquireTimeout)
...
-   */
   @Override
-  public PooledConnection borrowConnection(ServerLocation server,
-      boolean onlyUseExistingCnx) throws AllConnectionsInUseException, NoAvailableServersException {
-    PooledConnection connection =
-        availableConnectionManager.useFirst((c) -> c.getServer().equals(server));
-    if (null != connection) {
-      return connection;
-    }
+  public PooledConnection borrowConnection(ServerLocation server, long acquireTimeout,
```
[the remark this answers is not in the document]
mivanac: As I see this, previous behavior was only to search existing connection in pool, toward designated server. And if we did not find it, then in case onlyUseExistingCnx is set to false we will create new connection, regardless of number of connections.
 My proposal is, simular to borrowConnections(timeout), if we can not find connection in pool, and number of connections did not reach maximum, we create new connection toward designated server. And implement same retry logic from borrowConnections(timeout).

geode-core/src/integrationTest/java/org/apache/geode/cache/client/internal/pooling/ConnectionManagerJUnitTest.java
```diff
@@ -361,6 +363,82 @@ public void testBug41516()
         elapsedMillis >= BORROW_TIMEOUT_MILLIS - ALLOWABLE_ERROR_IN_MILLIS);
   }
 
+
+  @Test
+  public void testSingleHopRetryConnection()
```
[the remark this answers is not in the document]
mivanac: I will update test names.
These tests are aligned with previous tests. And connections are assigned just to fill up connection pool till maximum.

geode-core/src/integrationTest/java/org/apache/geode/cache/client/internal/pooling/ConnectionManagerImplTest.java
```diff
@@ -92,9 +92,18 @@ public void startShouldEatRejectedExecutionException() {
   @Test
   public void borrowConnectionThrowsWhenUsingExistingConnectionsAndNoConnectionsExist() {
```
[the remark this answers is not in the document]
mivanac: We are still testing same thing since there is no free connections in pool, since you can not define pool with 0 connections. If you want I can change name ...AndNoFreeConnectionsExist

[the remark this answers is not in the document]
DonalEvans: Looking at the old code vs the new, it seems like originally we just tried to call `borrowConnection()` and asserted that we throw an exception, but in the new test we first successfully call the single-argument `borrowConnection()` and then call the three-arg version and assert that that throws. Given that there are two calls to `borrowConnection()` in the test and only one of them throws, a test name that explains this difference in behaviour between the calls would be best.

geode-core/src/main/java/org/apache/geode/cache/client/internal/pooling/ConnectionManagerImpl.java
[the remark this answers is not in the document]
agingade: @mivanac You have introduced the wait which was not there before, this changes the behavior from caller perspective; It may have impact on the performance of some of the operation.

geode-core/src/main/java/org/apache/geode/cache/client/internal/pooling/ConnectionManagerImpl.java
```diff
@@ -301,31 +301,68 @@ public Connection borrowConnection(long acquireTimeout)
...
+              }
+            }
+          }
+        }
+
+        if (checkShutdownInterruptedOrTimeout(timeout)) {
+          break;
+        }
+
+        if (!onlyUseExistingCnx) {
```
agingade: If the request is not for existing connection; it doesn't has to wait, right? can this be moved before wait.

geode-core/src/main/java/org/apache/geode/cache/client/internal/pooling/ConnectionManagerImpl.java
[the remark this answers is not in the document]
mivanac: Added new parameter, which as default behavior disables waiting till timeout.

DonalEvans:
The most recent change resolves the issue we were seeing in internal tests that would cause a hang. Thank you for your patience with this; it's been extremely difficult to try and reproduce the failure in a DUnit environment to allow for debugging, and I wish I had been able to provide you with a test that could help get to the root cause of the issue.

geode-core/src/main/java/org/apache/geode/cache/client/internal/pooling/ConnectionManagerImpl.java
```diff
@@ -301,31 +301,68 @@ public Connection borrowConnection(long acquireTimeout)
...
-    if (onlyUseExistingCnx) {
-      throw new AllConnectionsInUseException();
-    }
+    PooledConnection connection;
+    logger.trace("Connection borrowConnection single hop connection");
+
+    long waitStart = NOT_WAITING;
+    try {
+      long timeout = System.nanoTime() + MILLISECONDS.toNanos(acquireTimeout);
+      while (true) {
```
agingade: While trying to find a connection, for given time (acquireTimeout); this may get into a hot-loop; consuming the CPU; can you see if you can wait for a notification which indicates a server connection is available.

geode-core/src/main/java/org/apache/geode/cache/client/PoolFactory.java
```diff
@@ -73,6 +73,14 @@
    */
   int DEFAULT_FREE_CONNECTION_TIMEOUT = 10000;
 
+  /**
+   * The default amount of time, in milliseconds, which we will wait for a server connection if max
+   * connections is set and there is no free connections towards designated server.
+   * <p>
+   * Current value: <code>-1</code>.
+   */
+  int DEFAULT_SERVER_CONNECTION_TIMEOUT = -1;
```
DonalEvans: A timeout of -1 is usually used to imply that the timeout is infinite, i.e. that we will wait forever to obtain a server connection if max connections is set and there is no free connections towards designated server. Would it be possible to change the default value to 0 but still ensure that the original behaviour is preserved?

geode-core/src/test/java/org/apache/geode/cache/client/internal/OpExecutorImplJUnitTest.java
```diff
@@ -99,7 +99,7 @@ public RuntimeException generateCancelledException(Throwable e) {
   @Test
   public void testExecute() throws Exception {
     OpExecutorImpl exec = new OpExecutorImpl(manager, queueManager, endpointManager, riTracker, 3,
-        10, cancelCriterion, null);
+        10, -1, cancelCriterion, null);
```
DonalEvans: When the default value for the timeout is being used, instead of hard-coding the value like here, it would be better to use a reference to the `DEFAULT_SERVER_CONNECTION_TIMEOUT` defined in PoolFactory.java.

geode-core/src/main/java/org/apache/geode/cache/client/PoolFactory.java
mivanac: For timeout, if you look at the code (checkShutdownInterruptedOrTimeout) -1 means no timeout, because you compare current time against, current + timeout. If we set 0, you will get the code to retry several times until 1ms passes.

DonalEvans: If the comparison in `checkShutdownInterruptedOrTimeout` was `timeout <= System.nanoTime()` then this problem would be solved, I think.

mivanac: OK. Updated.

PR #4758, opened 2020-03-03
GEODE-7837: add gemfire.security.sni-proxy system property

Bill:
[GEODE-7837](https://issues.apache.org/jira/browse/GEODE-7837) is an MVP to add SNI support in the Geode Java client. This PR lets you set an SNI proxy (host:port) in a system property. A follow-on PR [GEODE-7852](https://issues.apache.org/jira/browse/GEODE-7852) will replace the system property with a first-class connection pool configuration property.

This PR adds some unit tests of the system property parsing but doesn't have an integration test. Integration testing was manual for the MVP. We'll have a follow-on story TBD to add an integration test.

### For all changes:

### Note:


geode-core/src/main/java/org/apache/geode/internal/net/SCAdvancedSocketCreator.java
```diff
@@ -83,8 +86,33 @@ public Socket connect(HostAndPort addr, int timeout,
...
       }
-      socket.connect(sockaddr, Math.max(timeout, 0));
-      coreSocketCreator.configureClientSSLSocket(socket, timeout);
+
+      final InetSocketAddress sniProxyAddress =
+          coreSocketCreator.getSslConfig().getSniProxyAddress();
+      final boolean provideSNI = sniProxyAddress != null;
+
+      if (provideSNI) {
+        logger.error("DEBUG CREATING SSL PROXY SOCKET FOR {}", addr.getHostName());
```
pivotal-jbarrett: This line looks like a mistake.

geode-core/src/main/java/org/apache/geode/internal/net/SCAdvancedSocketCreator.java
```diff
@@ -83,8 +86,33 @@ public Socket connect(HostAndPort addr, int timeout,
...
+        final InetSocketAddress sockaddr;
+        InetSocketAddress sockaddrTemp = addr.getSocketInetAddress();
+        if (sockaddrTemp.getAddress() == null) {
+          final InetAddress address = InetAddress.getByName(sockaddrTemp.getHostName());
+          sockaddr = new InetSocketAddress(address, sockaddrTemp.getPort());
+        } else {
+          sockaddr = sockaddrTemp;
+        }
+        socket.connect(sockaddr, Math.max(timeout, 0));
+        coreSocketCreator.configureClientSSLSocket(socket, timeout, _ignored -> {
```
pivotal-jbarrett: Why not provide the SNI regardless of proxy?

geode-core/src/test/java/org/apache/geode/internal/net/SocketCreatorWithSNIProxyTest.java
```diff
@@ -0,0 +1,86 @@
...
+
+public class SocketCreatorWithSNIProxyTest {
+
+  /*
+   * This SNI proxy is not suitable for CI tests. It's for developer testing only.
+   */
+  @Before
+  public void before() {
+    System.setProperty(GEMFIRE_PREFIX + "security.sni-proxy",
+        "sni-proxy.moscow.cf-app.com:15443");
```
pivotal-jbarrett: This test address seem pretty specific, is that intentional?

geode-core/src/main/java/org/apache/geode/internal/admin/SSLConfig.java
```diff
@@ -237,10 +249,26 @@ public void toDSProperties(Properties props) {
...
 
     public SSLConfig build() {
+
+      final String sniProxyProperty = System.getProperty(GEMFIRE_PREFIX + "security.sni-proxy");
+      final InetSocketAddress sniProxyAddress;
+      if (sniProxyProperty == null) {
+        sniProxyAddress = null;
+      } else {
+        try {
+          final URL parser = new URL("http://" + sniProxyProperty);
```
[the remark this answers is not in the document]
Bill: Good point. There are many host+port parsers in the codebase. The one I was thinking of was the one that parses locator host+port: `GMSUtil.parseLocators()`. I opted not to use that one because it dictates a nonstandard syntax e.g. `hostname[123]`. That format is used in a few places in Geode but it's unusual to folks who are used to web browsers.

Do you know, offhand, of another parser we've got that parses the more orthodox colon-separated port (and supports fqdn's, IPv4 addys and IPv6 ones too?)

geode-core/src/main/java/org/apache/geode/internal/admin/SSLConfig.java
```diff
@@ -237,10 +249,26 @@ public void toDSProperties(Properties props) {
     public Builder() {}
 
     public SSLConfig build() {
+
+      final String sniProxyProperty = System.getProperty(GEMFIRE_PREFIX + "security.sni-proxy");
```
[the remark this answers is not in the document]
Bill: Yeah, it's currently a "security" property merely because SNI is part of TLS and TLS is configured via "security" properties.

The URL idea is interesting! I'm letting it soak in now…

geode-core/src/main/java/org/apache/geode/internal/net/SCAdvancedSocketCreator.java
Bill: fixed

geode-core/src/main/java/org/apache/geode/internal/net/SCAdvancedSocketCreator.java
Bill: If I understand your question, I chose to pass a lambda to `configureClientSSLSocket()` rather than passing in a (possibly-`null` SNI hostname).

geode-core/src/test/java/org/apache/geode/internal/net/SocketCreatorWithSNIProxyTest.java
Bill: Yes indeed. (recall this is a draft PR.) From the comments:

```
This SNI proxy is not suitable for CI tests. It's for developer testing only.
```

The actual test is `@Ignore`d because:

```
This test is ignored because it's connecting to a private toolsmith's BOSH environment.
```

Is it reasonable to have such an (`@Ignore`d) test make it into Geode as a temporary development utility? Or should I be sure to delete this whole test before publishing the (non-draft) PR?

As it stands, any developer with access to an SNI proxy, could edit these settings and test against their own environment. I realize it isn't ideal, but it's something.

Bill:
that IntegrationTestOpenJDK11 is a well-known flaky test outside this PR: https://issues.apache.org/jira/browse/GEODE-7843?jql=text%20~%20%22testConcurrentHSetNX%22

geode-core/src/main/java/org/apache/geode/internal/net/SCAdvancedSocketCreator.java
pivotal-jbarrett: My question is, There is always a host name so why not always set the SNI header?

geode-core/src/test/java/org/apache/geode/internal/net/SocketCreatorWithSNIProxyTest.java
pivotal-jbarrett: Gotcha...

You might look at https://github.com/dlundquist/sniproxy for a simple integration test.

geode-core/src/main/java/org/apache/geode/internal/net/SCAdvancedSocketCreator.java
Bill: Oh I see what you mean. I'd rather not add that header, which is superfluous in the common case i.e. no proxy. But yeah like you, I think functionally that could work.

geode-core/src/test/java/org/apache/geode/internal/net/SocketCreatorWithSNIProxyTest.java
Bill: What I'd like to do, if y'all are ok with it, is to promote this PR out of draft and to merge it with just unit tests and my manual integration-level testing. This will put the SNI client _mechanism_ into the `develop` branch, MVP-style. This will set the stage for a couple follow-on stories:

1. [GEODE-7852](https://issues.apache.org/jira/browse/GEODE-7852) to replace the system property with a first-class configuration property on connection pools
2. another follow-on story TBD to add an honest-to-goodness integration test with an SNI proxy

On (2) I'm experimenting right now to see if I can set up HAproxy in front of a Geode cluster all Dockerized. That would run nice in development and in CI if I can get it going.

geode-core/src/main/java/org/apache/geode/internal/admin/SSLConfig.java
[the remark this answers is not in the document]
Bill: I found many host+port parsers but none jumped out as the one to reuse. I'd like to go ahead with the use of the `URL` class for this PR and come in with a more fulsome approach in the follow-on story: [GEODE-7852](https://issues.apache.org/jira/browse/GEODE-7852). That story will replace the system property with a connection pool configuration property.

geode-core/src/main/java/org/apache/geode/internal/admin/SSLConfig.java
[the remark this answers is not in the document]
Bill: this will be addressed in a follow-on story [GEODE-7852](https://issues.apache.org/jira/browse/GEODE-7852)

geode-core/src/main/java/org/apache/geode/internal/admin/SSLConfig.java
```diff
@@ -237,10 +249,26 @@ public void toDSProperties(Properties props) {
     public Builder() {}
 
     public SSLConfig build() {
+
```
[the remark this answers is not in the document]
Bill: see above

geode-core/src/test/java/org/apache/geode/internal/net/SocketCreatorWithSNIProxyTest.java
Bill: deleting this test from this PR…

geode-core/src/main/java/org/apache/geode/internal/net/SCAdvancedSocketCreator.java
pivotal-jbarrett: I am thinking from the standpoint that everything is the same except for where it can't. So in this case the initialization of the SSL socket is identical regardless of the proxy. Less moving parts.

geode-core/src/test/java/org/apache/geode/internal/net/SocketCreatorWithSNIProxyTest.java
pivotal-jbarrett: I think until that point we should make it clear this is experimental.

Bill:
This PR needs an automated test agin a real SNI proxy. I'm closing the PR until that test is in place.

PR #4763, opened 2020-03-03
GEODE-7841 Add benchmarking scripts for Geode-Redis module

ringles:
We have created some scripts to measure the performance of Geode-Redis versus native Redis. They use the standard 'redis-benchmark' tool to run various Redis commands against either Geode or native Redis.

The benchmark tools consist of three scripts:

- aggregator.sh runs the actual tests
- benchmark.sh handles validating the Redis server (e.g. starting and stopping Geode Redis if needed) then runs aggregator.sh
- shacompare.sh takes two git SHAs as arguments and runs benchmark.sh on each of them, generating separate output files for each.

### DIFF
--- a/geode-redis/README.md
+++ b/geode-redis/README.md
@@ -1,0 +1,71 @@
+# Geode Redis Module
+
+## Contents
+1. [Overview](#overview)
+2. [Performance Test](#performance-test)
+
+## <a name="overview"></a>Overview
+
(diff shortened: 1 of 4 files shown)


moleske:
There's not much in ways of documentation.  I don't think most folks would want to go find the original JIRA or pull request to see the notes describing the files added.  Do you think a readme or other instruction file makes sense?  I'm honestly not sure where I'd want to put those docs

PR #4765, opened 2020-03-03
GEODE-7808: standardize on use of HostAndPort for creating connections

bschuchardt:
The base commits for this PR have already been approved and passed PR checks.  This revision reverts the removal of the SocketCreator InetAddress->hostname cache as it severely impacted Benchmark performance in CI runs.  Removal of that cache wasn't part of the charter for this PR, and reinstating it has no impact on the standardization of use of HostAndPort for creating connections.

I have not reverted a number of places that used SocketCreator to get the hostname for localhost.  Those all go to LocalHostUtil and access the same InetAddress, which will have a cached hostname and won't affect performance.

I've run Benchmark tests against the impacted tests on this branch and results show that performance is no longer affected.

### For all changes:

### Note:


geode-core/src/main/java/org/apache/geode/internal/net/SCClientSocketCreator.java
```diff
@@ -12,16 +12,24 @@
...
 
-import java.net.InetSocketAddress;
+import java.io.IOException;
+import java.net.Socket;
 
-import org.apache.geode.distributed.internal.tcpserver.LocatorAddress;
+import org.apache.geode.distributed.internal.tcpserver.ClientSocketCreatorImpl;
 
-public class HostAddress extends LocatorAddress {
+class SCClientSocketCreator extends ClientSocketCreatorImpl {
```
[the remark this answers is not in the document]
bschuchardt: SocketCreator.  I wanted to rename all of this since the purpose of SocketCreator is to add TLS support but it would kill backward-compatibility tests to change SocketCreator's name.  Maybe we can do that in Geode 2.0.

geode-core/src/main/java/org/apache/geode/distributed/internal/membership/InternalDistributedMember.java
```diff
@@ -100,7 +100,7 @@ public InternalDistributedMember(InetAddress i, int membershipPort, boolean spli
   }
 
   private static String getHostName(InetAddress i) {
-    return SocketCreator.resolve_dns ? i.getHostName() : i.getHostAddress();
+    return SocketCreator.resolve_dns ? SocketCreator.getHostName(i) : i.getHostAddress();
```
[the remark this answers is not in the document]
bschuchardt: yes

geode-core/src/main/java/org/apache/geode/internal/net/SCAdvancedSocketCreator.java
```diff
@@ -0,0 +1,119 @@
...
+    if (!useSSL) {
+      return super.connect(addr, timeout, optionalWatcher, allowClientSocketFactory,
+          socketBufferSize,
+          useSSL);
+    }
+
+    // create an SSL connection
+
+    Socket socket;
+    InetSocketAddress sockaddr = addr.getSocketInetAddress();
```
Bill: `getSocketInetAddress()` will try to resolve the hostname stored in `addr`…

geode-core/src/main/java/org/apache/geode/internal/net/SCAdvancedSocketCreator.java
```diff
@@ -0,0 +1,119 @@
...
+    if (sockaddr.getAddress() == null) {
+      InetAddress address = InetAddress.getByName(sockaddr.getHostString());
```
Bill: …then, if address resolution failed, we try `InetAddress.getByName()` again. With the exact same hostname.

Two things:

1. I don't see how this second address resolution will ever work if the first one failed (barring some rare race with DNS data)
2. In the case where we enter the conditional at line 61, we end up with `sockaddr` referring to an `InetSocketAddress` that has a `null` address and _no_ hostname at all. Wheres if we do not enter the conditional `sockaddr` refers to an `InetSocketAddress` that has both hostname and port.

Should we remove this conditional entirely?

Bill: Ah now I see that line 61 throws an `UnknownHostException` if address resolution fails. This, while quirky, is reasonable. Inasmuch as this logic has been in the product a long time, I'm ok with leaving it as-is.

bschuchardt: I had to do some recollection on this one Bill.
HostAndPort.getSocketInetAddress() is required to resolve hostnames if possible.  It's supposed to hold onto hostname:port pairs and leave the name unresolved until this method is called.
Yes, the second resolution will fail 100% of the time, but it will give us the expected UnknownHostException with the correct text.  Without that lookup there was at least one test failure that asserted that it should get this exception when attempting to use an invalid host name.

geode-core/src/main/java/org/apache/geode/internal/net/SCAdvancedSocketCreator.java
bschuchardt: Yes, that's what the LocatorAddress method was supposed to do.  I renamed LocatorAddress to HostAndPort since it can hold arbitrary addresses, not just locator addresses.  For your PR you can bypass getSocketInetAddress and just get the hostname since that's all you need for the SNIServerName parameter.

PR #4768, opened 2020-03-04
Some code simplification in ConnectionProxyJUnitTest

alb3rtobr:
I was taking a look to `ConnectionProxyJUnitTest` and I realized that `verifyAckSend`and `verifyExpiry`methods were not using their `timeToWait` parameter, so in fact they were using the default `GeodeAwaitability` timeout value (5 minutes).

I changed that and then I realized different calls to both methods were using different timeouts. And the one used for `verifyExpiry` in `testPeriodicAckSendByClient` was causing the test to fail, so I decided to homogenize calls to the same method, using a common timeout (the two new Duration objects in the class).

And finally I removed the usage of deprecated `WaitCriterion` class.


alb3rtobr:
I did not create a jira ticket for this change as I think its a small change, but let me know if you consider it is necessary

PR #4769, opened 2020-03-04
GEODE-7830: fix race in rebalance start

dschneider-pivotal:
The result from doing a rebalance start will now be consistent.
Previously the OperationState used to compute the result
could be changing while making the result causing the result
to be inconsistent. Now a copy of the OperationState is made
in a thread safe manner and then that immutable copy is used
to compute the result.

### For all changes:

### Note:


dschneider-pivotal:
@jmelchio

aaronlindsey:
I think you may have found a different scenario where the result of a rebalance could be inconsistent. I think the issue I saw might have been caused by `RebalanceOperationPerformer.executeRebalanceOnDS`. On line 330 it returns rebalanceResult with success=false. Then on line 347 it returns rebalanceResult with success=true. Both times they have the same status message.

dschneider-pivotal:
The two dunit failures had existing open jira tickets: GEODE-7710 and GEODE-7812

PR #4773, opened 2020-03-05
GEODE-7838: add row to rebalance command result

mkevo:

### For all changes:

### Note:

### DIFF
--- a/geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/RebalanceCommand.java
+++ b/geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/RebalanceCommand.java
@@ -170,6 +170,13 @@ public class RebalanceCommand extends GfshCommand {
     resultStr.append(CliStrings.REBALANCE__MSG__TOTALTIME).append(" = ").append(rstlist.get(8))
         .append(newLine);
 
+    table1.accumulate("Rebalanced Stats", CliStrings.REBALANCE__MSG__MEMBER_COUNT);
+    table1.accumulate("Value", String.valueOf(cache.getMembers().size()));
+    resultStr.append(CliStrings.REBALANCE__MSG__MEMBER_COUNT).append(" = ")
+        .append(cache.getMembers().size())
+        .append(newLine);
+
+
     String headerText;
     if (simulate) {
       headerText = "Simulated partition regions";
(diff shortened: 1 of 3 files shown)


mkevo:
> I'm not sure this `PR` fully addresses the original JIRA ticket: if a member goes down or comes up **after** the rebalance operation finishes but **before** the output is shown to the user, the member count will still be incorrect, right?.

Hi @jujoramos , the output will represent on how much members rebalance command is executed. If some member go out or come back after command is finished and before output is shown it will print just num of members on which rebalance command is executed. So user can see that he has different number of members in the system and in the output of command and have a conclusion that rebalance command is not executed on all currently members in the system.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/RebalanceCommand.java
```diff
@@ -170,6 +170,13 @@ private void toCompositeResultData(ResultModel result,
     resultStr.append(CliStrings.REBALANCE__MSG__TOTALTIME).append(" = ").append(rstlist.get(8))
         .append(newLine);
 
+    table1.accumulate("Rebalanced Stats", CliStrings.REBALANCE__MSG__MEMBER_COUNT);
+    table1.accumulate("Value", String.valueOf(cache.getMembers().size()));
```
jujoramos: @mkevo: this is what I was talking about, the output will contain the number of members _currently_ present in the distributed system, **_not_** the amount of members on which the rebalance was executed. Easier to show with an example:
**T0:** There are 3 members in the system.
**T1:** Users executes the rebalance command. 
**T2:** Rebalance finishes and control is returned to the `gfsh` command while a member joins the distributed system.
**T3:** The command use `cache.getMembers().size()` in the output , which at this time is 4, not 3; but the rebalance was executed only on the original 3 members and not the one that just joined.

Makes sense?.

mkevo: Yes, you are right. I add some sleep in between rebalance finished and output is not yet shown and it has a wrong result. 
I think that change is needed in RebalanceRegionResult, add new functions to get members.
I will close this PR and go on with different implementation.
Thanks @jujoramos !

PR #4781, opened 2020-03-09
GEODE-7828: Convert backing store for Redis Hashes and Sets to single regions

ringles:
- GEODE-7843 fixes sporadic test failures for HSETNX


jdeppe-pivotal:
Failed test in distributedTests is unrelated to this PR

PR #4790, opened 2020-03-10
Geode prefers not to use merge-commits, so do not test against them.

rhoughton-pivot:
The PR resource supports rebase instead of merging against the latest
base-branch. Do that instead of merging.

Co-authored-by: Robert Houghton <rhoughton@pivotal.io>
Co-authored-by: Owen Nichols <onichols@pivotal.io>


upthewaterspout:
I guess I'm a bit confused here - won't an automatic rebase produce the same results as an automatic merge, if it is successful?

Is the issue really that rebase works and merge doesn't with this PR resource?

PR #4792, opened 2020-03-11
GEODE-7869: Cleanup and enforce compiler warnings.

pivotal-jbarrett:

### For all changes:

### Note:

### DIFF
--- a/geode-dunit/src/main/java/org/apache/geode/management/internal/cli/HeadlessGfsh.java
+++ b/geode-dunit/src/main/java/org/apache/geode/management/internal/cli/HeadlessGfsh.java
@@ -20,6 +20,7 @@ import java.io.FileDescriptor;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.PrintStream;
+import java.nio.file.Files;
 import java.util.Properties;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.LinkedBlockingQueue;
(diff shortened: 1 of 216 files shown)


geode-dunit/src/main/java/org/apache/geode/management/internal/cli/HeadlessGfsh.java
```diff
@@ -260,16 +259,12 @@ public boolean hasError() {
...
-            endOfShell.await();
-          } catch (InterruptedException e) {
-            // e.printStackTrace();
-          }
-          this.exitShellRequest = ExitShellRequest.NORMAL_EXIT;
-          setShellStatus(Status.SHUTTING_DOWN);
-          break;
+        try {
+          endOfShell.await();
+        } catch (InterruptedException ignore) {
```
rhoughton-pivot: Isn't there a new warning for an empty catch block?

geode-gfsh/src/distributedTest/java/org/apache/geode/management/internal/cli/NetstatDUnitTest.java
```diff
@@ -165,6 +166,7 @@ public void testOutputWithLsofToFile() throws Exception {
 
     File outputFile = new File(temp.newFolder(), "command.log.txt");
 
+    @SuppressWarnings("deprecation")
```
rhoughton-pivot: Do we have a JIRA to replace this test with one using the non-deprecated code path? I understand that work is beyond the scope of 7869

geode-dunit/src/main/java/org/apache/geode/management/internal/cli/HeadlessGfsh.java
pivotal-jbarrett: Not in the compiler but in IJ and other tools. IJ is ok if you name it ignore.

geode-gfsh/src/distributedTest/java/org/apache/geode/management/internal/cli/NetstatDUnitTest.java
pivotal-jbarrett: No. I think I am going to propose that any code changes after this point that produce a deprecation failure should be addressed correctly. Also many of these deprecated items will be stripped from the code in the next major, forcing this issue. The goal of this ticket is to not make this problem worse over time. A committer deprecating code will be forced to address it before it will build. If they use suppression it will be obvious in the PR and should be challenged.

PR #4807, opened 2020-03-13
GEODE-7710: Fix race condition in federation of MXBeans

kirklund:
JMX Manager may fail to federate one or more MXBeans during startup
because of a race condition involving FederatingManager
ManagementCacheListener.

Change ManagementCacheListener to be ready upon construction.

Since ManagementCacheListener is a test without unit test, I have created
ManagementCacheListenerTest which follows these unit testing standards:
* it uses Mockito for all dependencies
* follows Michael Feather's definition of a unit test
* uses terminology and patterns from http://xunitpatterns.com/
* uses TextDox inspired test method naming

PR #4808, opened 2020-03-13
GEODE-7710: Fix race condition in sending JMX notifications

kirklund:
JMX Manager may fail to broadcast one or more JMX notifications
during startup because of a race condition involving FederatingManager
and NotificationCacheListener.

Change NotificationCacheListener to be ready upon construction.

Since NotificationCacheListener is a test without unit test, I have created
NotificationCacheListenerTest which follows these unit testing standards:
* it uses Mockito for all dependencies
* follows Michael Feather's definition of a unit test
* uses terminology and patterns from http://xunitpatterns.com/
* uses TextDox inspired test method naming

### DIFF
--- a/geode-core/src/test/java/org/apache/geode/management/internal/NotificationCacheListenerTest.java
+++ b/geode-core/src/test/java/org/apache/geode/management/internal/NotificationCacheListenerTest.java
@@ -1,0 +1,70 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.apache.geode.management.internal;
+
+import static org.apache.geode.internal.cache.util.UncheckedUtils.cast;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+
+import javax.management.Notification;
+
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.mockito.junit.MockitoJUnit;
+import org.mockito.junit.MockitoRule;
+import org.mockito.quality.Strictness;
+
+import org.apache.geode.cache.EntryEvent;
+
+/**
+ * Unit tests for {@link NotificationCacheListener} (ie the SUT). These are characterization tests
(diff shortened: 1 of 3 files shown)


geode-core/src/distributedTest/java/org/apache/geode/management/MultipleManagersRegressionTest.java
```diff
@@ -0,0 +1,191 @@
...
+import org.junit.experimental.categories.Category;
+
+import org.apache.geode.distributed.LocatorLauncher;
+import org.apache.geode.distributed.internal.InternalLocator;
+import org.apache.geode.test.dunit.VM;
+import org.apache.geode.test.dunit.rules.DistributedRule;
+import org.apache.geode.test.dunit.rules.SharedErrorCollector;
+import org.apache.geode.test.junit.categories.JMXTest;
+import org.apache.geode.test.junit.rules.serializable.SerializableTemporaryFolder;
+
```
mhansonp: I would like to see a description of the purpose and the approach of these tests.

geode-core/src/test/java/org/apache/geode/management/internal/ManagementCacheListenerTest.java
```diff
@@ -0,0 +1,154 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+import org.mockito.junit.MockitoJUnit;
+import org.mockito.junit.MockitoRule;
+import org.mockito.quality.Strictness;
+
+import org.apache.geode.cache.EntryEvent;
+import org.apache.geode.cache.Region;
+import org.apache.geode.distributed.DistributedMember;
+
```
mhansonp: I would like to see a description of the purpose and the approach of these tests.

geode-core/src/test/java/org/apache/geode/management/internal/NotificationCacheListenerTest.java
```diff
@@ -0,0 +1,65 @@
...
+
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.mockito.junit.MockitoJUnit;
+import org.mockito.junit.MockitoRule;
+import org.mockito.quality.Strictness;
+
+import org.apache.geode.cache.EntryEvent;
+
```
mhansonp: I would like to see a description of the purpose and the approach of these tests.

geode-core/src/distributedTest/java/org/apache/geode/management/MultipleManagersRegressionTest.java
kirklund: I've deleted MultipleManagersRegressionTest because it's redundant with coverage in JMXMBeanReconnectDUnitTest

geode-core/src/test/java/org/apache/geode/management/internal/ManagementCacheListenerTest.java
kirklund: ManagementCacheListenerTest is part of #4807. It's duplicated in this PR only because #4808 builds upon #4807 and because Geode dev community no longer wants to allow a PR to have multiple commits.

geode-core/src/test/java/org/apache/geode/management/internal/NotificationCacheListenerTest.java
kirklund: Java description has been added to each unit test class.

PR #4809, opened 2020-03-13
GEODE-7710: Add getRedundancyZone to DistributionManager

kirklund:
Avoids having to check instanceof and cast to
ClusterDistributionManager in MemberMBeanBridge.


geode-core/src/main/java/org/apache/geode/management/internal/beans/MemberMBeanBridge.java
```diff
@@ -193,12 +192,8 @@
 
     dm = system.getDistributionManager();
```
kirklund: TODO: get rid of `dm` field and just use the `system` field directly instead of fetching from dm within `shutDownMember()`

kirklund: Done

PR #4815, opened 2020-03-16
GEODE-7838: getNumOfServers while rebalance is executed

mkevo:

### For all changes:

### Note:


aaronlindsey:
I don't see what this PR has to do with GEODE-7383. Did you mean to specify a different JIRA ticket?

mkevo:
> I don't see what this PR has to do with GEODE-7383. Did you mean to specify a different JIRA ticket?

Sorry, wrong JIRA ticket number.

geode-core/src/main/java/org/apache/geode/cache/partition/PartitionRebalanceInfo.java
```diff
@@ -137,4 +137,11 @@
    * @return the time, in milliseconds, spent transferring primaries for this region
    */
   long getPrimaryTransferTime();
+
+  /**
+   * Returns the total number of members on which command is executed.
+   *
+   * @return the total number of members on which command is executed
+   */
+  int getPrimaryNumOfMembers();
```
DonalEvans, line 146: This name is confusing, since the method has nothing to do with primary buckets. A better name might be "getNumberOfMembersExecutedOn".

geode-core/src/main/java/org/apache/geode/cache/control/RebalanceResults.java
```diff
@@ -99,4 +99,11 @@
    * @return the total time, in milliseconds, spent transferring primaries
    */
   long getTotalPrimaryTransferTime();
+
+  /**
+   * Returns the total number of members on which command is executed.
+   *
+   * @return the total number of members on which command is executed
+   */
+  int getTotalNumOfMembers();
```
DonalEvans, line 108: This method might be better named "getTotalMembersExecutedOn", or something else that makes it clear that it's specifically returning the number of members the command was executed on.

geode-core/src/main/java/org/apache/geode/management/internal/i18n/CliStrings.java
```diff
@@ -1873,6 +1873,8 @@
       "Total primaries transferred during this rebalance";
   public static final String REBALANCE__MSG__TOTALTIME =
       "Total time (in milliseconds) for this rebalance";
+  public static final String REBALANCE__MSG__MEMBER_COUNT =
+      "Total number of members in system during rebalance";
```
DonalEvans, line 1877: This description is potentially misleading, since the count is actually the number of members on which the rebalance was executed, which might be different from the total number of members in the system.

geode-management/src/main/java/org/apache/geode/management/runtime/RebalanceRegionResult.java
```diff
@@ -74,4 +74,9 @@
    * Returns the time, in milliseconds, that the rebalance operation took for this region.
    */
   long getTimeInMilliseconds();
+
+  /**
+   * Returns the number of members on which rebalance operation is executed.
+   */
+  int getNum0fMembers();
```
DonalEvans, line 81: Typo here, should be `Of`, not `0f`

mkevo:
> Some typos and naming suggestions. Also, there is no test confirming the new intended behaviour as described in the JIRA ticket. A test is needed that starts a rebalance, stops and/or starts a member, then confirms that the output of the rebalance command is correct with respect to how many members were actually rebalanced.

As this change doesn't have impacts on functionality of rebalance command just on it's output, I added a test to reproduce a scenario when user list members to see how many server it has, after that do rebalance command which is executed on lower number of servers as one is stopped. After rebalance is completed server is started again and user can see in list member that all members are there and think that rebalance is executed on all of them, but it doesn't. Regarding that new line in rebalance command output is added to be aware how many servers is rebalanced.

geode-dunit/src/main/java/org/apache/geode/management/internal/cli/commands/RebalanceCommandDistributedTestBase.java
```diff
@@ -94,6 +98,40 @@ public void testRebalanceResultOutput() {
...
+    Map<String, List<String>> rebalanceResult = gfsh.executeAndAssertThat("rebalance")
+        .statusIsSuccess().hasTableSection().getActual().getContent();
+    
+    server3.waitTilFullyReconnected();
+
+    listMembersResult = gfsh.executeAndAssertThat("list members")
+        .hasTableSection().getActual().getContent();
+    assertThat(listMembersResult.get("Name").size()).isEqualTo(4);
+    assertThat(rebalanceResult.get("Rebalanced Stats").get(9))
+        .isEqualTo("Total number of members in system on which rebalance is executed");
```
DonalEvans, line 131: This string should ideally be replaced with a reference to the `REBALANCE__MSG__MEMBER_COUNT` constant.

mkevo: Done! :)

DonalEvans:
It only just occurred to me that there should also be some docs changes to go with this, and since the change will be very small, it probably doesn't warrant a separate JIRA ticket. The content in `geode-docs/tools_modules/gfsh/command-pages/rebalance.html.md.erb` should be updated to reflect the new output of the command.

mkevo:
> It only just occurred to me that there should also be some docs changes to go with this, and since the change will be very small, it probably doesn't warrant a separate JIRA ticket. The content in `geode-docs/tools_modules/gfsh/command-pages/rebalance.html.md.erb` should be updated to reflect the new output of the command.

It seems that this part is not updated for a long time. I will add new output to the document.

DonalEvans:
I just spotted a bug in this commit and filed a JIRA ticket for it: https://issues.apache.org/jira/browse/GEODE-7980. When colocated regions are present, the number of members reported in the output of the rebalance command is incorrect.

mkevo:
Nice catch! Thanks!

PR #4818, opened 2020-03-16
GEODE-7667: Add a 'clear' gfsh command for PR and RR clear

BenjaminPerryRoss:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/distributedTest/java/org/apache/geode/cache30/RegionReliabilityTestCase.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/cache30/RegionReliabilityTestCase.java
@@ -1341,7 +1341,7 @@ public abstract class RegionReliabilityTestCase extends ReliabilityTestCase {
       // pass
     }
 
-    // CLEAR
+    // CLEAR_REGION
     reset.run();
     Host.getHost(0).getVM(1).invoke(createRegion);
     try {
--- a/geode-core/src/distributedTest/java/org/apache/geode/cache30/RegionReliabilityTestCase.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/cache30/RegionReliabilityTestCase.java
@@ -1341,7 +1341,7 @@ public abstract class RegionReliabilityTestCase extends ReliabilityTestCase {
       // pass
     }
 
-    // CLEAR_REGION
+    // CLEAR
     reset.run();
     Host.getHost(0).getVM(1).invoke(createRegion);
     try {
--- a/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClearRvvLockingDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClearRvvLockingDUnitTest.java
@@ -51,7 +51,7 @@ import org.apache.geode.test.dunit.cache.internal.JUnit4CacheTestCase;
 
 /**
  * Test class to verify proper RVV locking interaction between entry operations such as PUT/REMOVE
- * and the CLEAR_REGION region operation
+ * and the CLEAR region operation
  *
  * GEODE-599: After an operation completed, it would unlock the RVV. This was occurring before the
  * operation was distributed to other members which created a window in which another operation
--- a/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClearRvvLockingDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClearRvvLockingDUnitTest.java
@@ -51,7 +51,7 @@ import org.apache.geode.test.dunit.cache.internal.JUnit4CacheTestCase;
 
 /**
  * Test class to verify proper RVV locking interaction between entry operations such as PUT/REMOVE
- * and the CLEAR region operation
+ * and the CLEAR_REGION region operation
  *
  * GEODE-599: After an operation completed, it would unlock the RVV. This was occurring before the
  * operation was distributed to other members which created a window in which another operation
--- a/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClearTXLockingDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClearTXLockingDUnitTest.java
@@ -49,14 +49,13 @@ import org.apache.geode.test.dunit.rules.DistributedRule;
  * <p>
  * GEODE-1740: It was observed that operations performed within a transaction were not holding
  * region modification locks for the duration of commit processing. This lock is used to ensure
- * region consistency during CLEAR_REGION processing. By not holding the lock for the duration of
- * commit
+ * region consistency during CLEAR processing. By not holding the lock for the duration of commit
  * processing, a window was opened that allowed region operations such as clear to occur in
  * mid-commit.
  *
  * <p>
  * The fix for GEODE-1740 was to acquire and hold read locks for any region involved in the commit.
- * This forces CLEAR_REGION to wait until commit processing is complete.
+ * This forces CLEAR to wait until commit processing is complete.
  *
  * <p>
  * This test performs operations within a transaction and during commit processing schedules a
@@ -274,7 +273,7 @@ public class ClearTXLockingDUnitTest implements Serializable {
   }
 
   /**
-   * Wait to be notified it is time to perform region operation (i.e. CLEAR_REGION)
+   * Wait to be notified it is time to perform region operation (i.e. CLEAR)
    */
   private static void regionOperationWait(CountDownLatch latch) throws InterruptedException {
     latch.await();
--- a/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClearTXLockingDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClearTXLockingDUnitTest.java
@@ -43,19 +43,19 @@ import org.apache.geode.test.dunit.VM;
 import org.apache.geode.test.dunit.rules.DistributedRule;
 
 /**
- * Test class to verify proper locking interaction between transactions and the CLEAR region
+ * Test class to verify proper locking interaction between transactions and the CLEAR_REGION region
  * operation.
  *
  * <p>
  * GEODE-1740: It was observed that operations performed within a transaction were not holding
  * region modification locks for the duration of commit processing. This lock is used to ensure
- * region consistency during CLEAR processing. By not holding the lock for the duration of commit
+ * region consistency during CLEAR_REGION processing. By not holding the lock for the duration of commit
  * processing, a window was opened that allowed region operations such as clear to occur in
  * mid-commit.
  *
  * <p>
  * The fix for GEODE-1740 was to acquire and hold read locks for any region involved in the commit.
- * This forces CLEAR to wait until commit processing is complete.
+ * This forces CLEAR_REGION to wait until commit processing is complete.
  *
  * <p>
  * This test performs operations within a transaction and during commit processing schedules a
@@ -273,7 +273,7 @@ public class ClearTXLockingDUnitTest implements Serializable {
   }
 
   /**
-   * Wait to be notified it is time to perform region operation (i.e. CLEAR)
+   * Wait to be notified it is time to perform region operation (i.e. CLEAR_REGION)
    */
   private static void regionOperationWait(CountDownLatch latch) throws InterruptedException {
     latch.await();
--- a/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClearTXLockingDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClearTXLockingDUnitTest.java
@@ -43,7 +43,7 @@ import org.apache.geode.test.dunit.VM;
 import org.apache.geode.test.dunit.rules.DistributedRule;
 
 /**
- * Test class to verify proper locking interaction between transactions and the CLEAR_REGION region
+ * Test class to verify proper locking interaction between transactions and the CLEAR region
  * operation.
  *
  * <p>
--- a/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClearTXLockingDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClearTXLockingDUnitTest.java
@@ -49,7 +49,8 @@ import org.apache.geode.test.dunit.rules.DistributedRule;
  * <p>
  * GEODE-1740: It was observed that operations performed within a transaction were not holding
  * region modification locks for the duration of commit processing. This lock is used to ensure
- * region consistency during CLEAR_REGION processing. By not holding the lock for the duration of commit
+ * region consistency during CLEAR_REGION processing. By not holding the lock for the duration of
+ * commit
  * processing, a window was opened that allowed region operations such as clear to occur in
  * mid-commit.
  *
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/DiskStoreImpl.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/DiskStoreImpl.java
@@ -3509,7 +3509,7 @@ public class DiskStoreImpl implements DiskStore {
       if (de != null) {
         sb.append(" key=").append(de.getKey());
       } else {
-        sb.append(" <END CLEAR_REGION>");
+        sb.append(" <END CLEAR>");
       }
       return sb.toString();
     }
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/DiskStoreImpl.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/DiskStoreImpl.java
@@ -3509,7 +3509,7 @@ public class DiskStoreImpl implements DiskStore {
       if (de != null) {
         sb.append(" key=").append(de.getKey());
       } else {
-        sb.append(" <END CLEAR>");
+        sb.append(" <END CLEAR_REGION>");
       }
       return sb.toString();
     }
--- a/geode-core/src/main/java/org/apache/geode/management/internal/i18n/CliStrings.java
+++ b/geode-core/src/main/java/org/apache/geode/management/internal/i18n/CliStrings.java
@@ -818,7 +818,8 @@ public class CliStrings {
 
   /* clear region */
   public static final String CLEAR_REGION = "clear region";
-  public static final String CLEAR_REGION_HELP = "Clears/Removes all keys from the specified region.";
+  public static final String CLEAR_REGION_HELP =
+      "Clears/Removes all keys from the specified region.";
   public static final String CLEAR_REGION_REGION_NAME = "region";
   public static final String CLEAR_REGION_REGION_NAME_HELP = "Region to clear keys from.";
 
@@ -1921,8 +1922,9 @@ public class CliStrings {
   public static final String REMOVE__MSG__REGION_NOT_FOUND = "Region <{0}> Not Found";
   public static final String REMOVE__MSG__KEY_NOT_FOUND_REGION = "Key is not present in the region";
   public static final String REMOVE__MSG__CLEARED_ALL_CLEARS = "Cleared all keys in the region";
-  public static final String REMOVE__MSG__CLEARALL_DEPRECATION_WARNING = "Warning: The --all option for the 'remove' command is deprecated. Please"
-      + " use the 'clear region' command instead.";
+  public static final String REMOVE__MSG__CLEARALL_DEPRECATION_WARNING =
+      "Warning: The --all option for the 'remove' command is deprecated. Please"
+          + " use the 'clear region' command instead.";
   public static final String REMOVE__MSG__CLEARALL_NOT_SUPPORTED_FOR_PARTITIONREGION =
       "Option --" + REMOVE__ALL + " is not supported on partitioned region";
 
--- a/geode-core/src/main/java/org/apache/geode/management/internal/i18n/CliStrings.java
+++ b/geode-core/src/main/java/org/apache/geode/management/internal/i18n/CliStrings.java
@@ -822,6 +822,7 @@ public class CliStrings {
       "Clears/Removes all keys from the specified region.";
   public static final String CLEAR_REGION_REGION_NAME = "name";
   public static final String CLEAR_REGION_REGION_NAME_HELP = "Region to clear keys from.";
+  public static final String CLEAR_REGION_CLEARED_ALL_KEYS = "Cleared all keys in the region";
 
   /* create region */
   public static final String CREATE_REGION = "create region";
@@ -1921,7 +1922,6 @@ public class CliStrings {
   public static final String REMOVE__MSG__KEY_EMPTY = "Key is Null";
   public static final String REMOVE__MSG__REGION_NOT_FOUND = "Region <{0}> Not Found";
   public static final String REMOVE__MSG__KEY_NOT_FOUND_REGION = "Key is not present in the region";
-  public static final String REMOVE__MSG__CLEARED_ALL_KEYS = "Cleared all keys in the region";
   public static final String REMOVE__MSG__CLEARALL_DEPRECATION_WARNING =
       "Warning: The --all option for the 'remove' command is deprecated. Please"
           + " use the 'clear' command instead.";
--- a/geode-core/src/main/java/org/apache/geode/management/internal/i18n/CliStrings.java
+++ b/geode-core/src/main/java/org/apache/geode/management/internal/i18n/CliStrings.java
@@ -816,6 +816,12 @@ public class CliStrings {
   public static final String CLEAR_DEFINED_INDEX__SUCCESS__MSG =
       "Index definitions successfully cleared";
 
+  /* clear region */
+  public static final String CLEAR_REGION = "clear region";
+  public static final String CLEAR_REGION_HELP = "Clears/Removes all keys from the specified region.";
+  public static final String CLEAR_REGION_REGION_NAME = "region";
+  public static final String CLEAR_REGION_REGION_NAME_HELP = "Region to clear keys from.";
+
   /* create region */
   public static final String CREATE_REGION = "create region";
   public static final String CREATE_REGION__HELP =
@@ -1915,6 +1921,8 @@ public class CliStrings {
   public static final String REMOVE__MSG__REGION_NOT_FOUND = "Region <{0}> Not Found";
   public static final String REMOVE__MSG__KEY_NOT_FOUND_REGION = "Key is not present in the region";
   public static final String REMOVE__MSG__CLEARED_ALL_CLEARS = "Cleared all keys in the region";
+  public static final String REMOVE__MSG__CLEARALL_DEPRECATION_WARNING = "Warning: The --all option for the 'remove' command is deprecated. Please"
+      + " use the 'clear region' command instead.";
   public static final String REMOVE__MSG__CLEARALL_NOT_SUPPORTED_FOR_PARTITIONREGION =
       "Option --" + REMOVE__ALL + " is not supported on partitioned region";
 
--- a/geode-core/src/main/java/org/apache/geode/management/internal/i18n/CliStrings.java
+++ b/geode-core/src/main/java/org/apache/geode/management/internal/i18n/CliStrings.java
@@ -1921,7 +1921,7 @@ public class CliStrings {
   public static final String REMOVE__MSG__KEY_EMPTY = "Key is Null";
   public static final String REMOVE__MSG__REGION_NOT_FOUND = "Region <{0}> Not Found";
   public static final String REMOVE__MSG__KEY_NOT_FOUND_REGION = "Key is not present in the region";
-  public static final String REMOVE__MSG__CLEARED_ALL_CLEARS = "Cleared all keys in the region";
+  public static final String REMOVE__MSG__CLEARED_ALL_KEYS = "Cleared all keys in the region";
   public static final String REMOVE__MSG__CLEARALL_DEPRECATION_WARNING =
       "Warning: The --all option for the 'remove' command is deprecated. Please"
           + " use the 'clear' command instead.";
--- a/geode-core/src/main/java/org/apache/geode/management/internal/i18n/CliStrings.java
+++ b/geode-core/src/main/java/org/apache/geode/management/internal/i18n/CliStrings.java
@@ -817,11 +817,11 @@ public class CliStrings {
       "Index definitions successfully cleared";
 
   /* clear region */
-  public static final String CLEAR_REGION = "clear region";
-  public static final String CLEAR_REGION_HELP =
+  public static final String CLEAR = "clear";
+  public static final String CLEAR_HELP =
       "Clears/Removes all keys from the specified region.";
-  public static final String CLEAR_REGION_REGION_NAME = "region";
-  public static final String CLEAR_REGION_REGION_NAME_HELP = "Region to clear keys from.";
+  public static final String CLEAR_REGION_NAME = "region";
+  public static final String CLEAR_REGION_NAME_HELP = "Region to clear keys from.";
 
   /* create region */
   public static final String CREATE_REGION = "create region";
@@ -1924,9 +1924,7 @@ public class CliStrings {
   public static final String REMOVE__MSG__CLEARED_ALL_CLEARS = "Cleared all keys in the region";
   public static final String REMOVE__MSG__CLEARALL_DEPRECATION_WARNING =
       "Warning: The --all option for the 'remove' command is deprecated. Please"
-          + " use the 'clear region' command instead.";
-  public static final String REMOVE__MSG__CLEARALL_NOT_SUPPORTED_FOR_PARTITIONREGION =
-      "Option --" + REMOVE__ALL + " is not supported on partitioned region";
+          + " use the 'clear' command instead.";
 
   /* resume gateway-sender */
   public static final String RESUME_GATEWAYSENDER = "resume gateway-sender";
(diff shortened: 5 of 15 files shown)


geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/ClearCommand.java
```diff
@@ -0,0 +1,81 @@
...
+      request.setRemoveAllKeys("ALL");
+      request.setRegionName(regionPath);
+      dataResult = callFunctionForRegion(request, removefn, memberList);
+    } else {
+      dataResult = removefn.remove(null, null, regionPath, "ALL",
+          (InternalCache) cache);
+    }
+
+    // BR: is this needed?
+    dataResult.setKeyClass(null);
```
gesterzhou: can you try to remove this line?

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/ClearCommand.java
```diff
@@ -0,0 +1,81 @@
...
+import org.apache.geode.management.internal.i18n.CliStrings;
+import org.apache.geode.security.ResourcePermission.Operation;
+import org.apache.geode.security.ResourcePermission.Resource;
+
+public class ClearCommand extends GfshCommand {
+  public static final String REGION_NOT_FOUND = "Region <%s> not found in any of the members";
+
+  @CliMetaData(relatedTopic = {CliStrings.TOPIC_GEODE_DATA, CliStrings.TOPIC_GEODE_REGION})
+  @CliCommand(value = {CliStrings.CLEAR_REGION}, help = CliStrings.CLEAR_REGION_HELP)
+  public ResultModel remove(
```
gesterzhou: should we use clear() instead of remove() for method name?

geode-core/src/main/java/org/apache/geode/management/internal/i18n/CliStrings.java
```diff
@@ -1915,6 +1922,9 @@
   public static final String REMOVE__MSG__REGION_NOT_FOUND = "Region <{0}> Not Found";
   public static final String REMOVE__MSG__KEY_NOT_FOUND_REGION = "Key is not present in the region";
   public static final String REMOVE__MSG__CLEARED_ALL_CLEARS = "Cleared all keys in the region";
+  public static final String REMOVE__MSG__CLEARALL_DEPRECATION_WARNING =
+      "Warning: The --all option for the 'remove' command is deprecated. Please"
+          + " use the 'clear region' command instead.";
   public static final String REMOVE__MSG__CLEARALL_NOT_SUPPORTED_FOR_PARTITIONREGION =
```
agingade: REMOVE__MSG__CLEARALL_NOT_SUPPORTED_FOR_PARTITIONREGION  is not true anymore, right?

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/ClearCommand.java
```diff
@@ -0,0 +1,81 @@
...
+    DataCommandResult dataResult;
+    if (region == null) {
+      Set<DistributedMember> memberList = findAnyMembersForRegion(regionPath);
+
+      if (CollectionUtils.isEmpty(memberList)) {
+        return new ResultModel().createError(String.format(REGION_NOT_FOUND, regionPath));
+      }
+
+      DataCommandRequest request = new DataCommandRequest();
+      request.setCommand(CliStrings.REMOVE);
```
agingade: As Gester suggested; how about using "clear" instead of "remove/all".

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/functions/DataCommandFunction.java
```diff
@@ -314,17 +314,13 @@ public DataCommandResult remove(String key, String keyClass, String regionName,
...
-          region.clear();
-          if (logger.isDebugEnabled()) {
-            logger.debug("Cleared all keys in the region - {}", regionName);
-          }
-          return DataCommandResult.createRemoveInfoResult(key, null, null,
-              CliStrings.format(CliStrings.REMOVE__MSG__CLEARED_ALL_CLEARS, regionName), true);
-        } else {
-          return DataCommandResult.createRemoveInfoResult(key, null, null,
-              CliStrings.REMOVE__MSG__CLEARALL_NOT_SUPPORTED_FOR_PARTITIONREGION, false);
+        region.clear();
```
agingade: Can we add clear() instead of using remove()? Till the deprecated remove-all is removed, the remove-all can clear().

geode-gfsh/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/ClearCommandDUnitTest.java
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  @Test
+  public void clearReplicateRegion() {
+    String command = commandString + " --region=" + REPLICATE_REGION_NAME;
+
+    gfsh.executeAndAssertThat("list regions").statusIsSuccess();
+    gfsh.executeAndAssertThat(command).statusIsSuccess();
+
+    assertThat(gfsh.getGfshOutput()).contains("Cleared all keys in the region");
```
DonalEvans: The "Cleared all keys in the region" string should ideally be extracted to a constant where it's defined so other places can refer to that rather than hard-coding.

geode-core/src/main/java/org/apache/geode/management/internal/i18n/CliStrings.java
```diff
@@ -1915,6 +1922,9 @@
   public static final String REMOVE__MSG__REGION_NOT_FOUND = "Region <{0}> Not Found";
   public static final String REMOVE__MSG__KEY_NOT_FOUND_REGION = "Key is not present in the region";
   public static final String REMOVE__MSG__CLEARED_ALL_CLEARS = "Cleared all keys in the region";
+  public static final String REMOVE__MSG__CLEARALL_DEPRECATION_WARNING =
+      "Warning: The --all option for the 'remove' command is deprecated. Please"
```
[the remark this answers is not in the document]
BenjaminPerryRoss: We can double check this with Charlie, but the new command 'clear' has no way to remove specific entries, while the remove command does. We don't want to deprecate the whole remove command or we would be eventually losing that functionality.

geode-gfsh/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/ClearCommandDUnitTest.java
```diff
@@ -0,0 +1,123 @@
...
+
+    gfsh.executeAndAssertThat(command).statusIsError()
+        .containsOutput(String.format(REGION_NOT_FOUND, "/NotAValidRegion"));
+  }
+
+  @Test
+  public void clearReplicateRegion() {
+    String command = commandString + " --region=" + REPLICATE_REGION_NAME;
+
+    gfsh.executeAndAssertThat("list regions").statusIsSuccess();
```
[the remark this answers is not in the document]
BenjaminPerryRoss: This makes the test more readable in the event of a failure (although the assertion of success is probably unnecessary).

geode-gfsh/src/integrationTest/java/org/apache/geode/management/internal/cli/GfshParserAutoCompletionIntegrationTest.java
```diff
@@ -381,7 +381,7 @@ public void testObtainHintForData() {
     String hintArgument = "data";
     String hintsProvided = gfshParserRule.getCommandManager().obtainHint(hintArgument);
     String[] hintsProvidedArray = hintsProvided.split(lineSeparator());
-    assertThat(hintsProvidedArray.length).isEqualTo(15);
+    assertThat(hintsProvidedArray.length).isEqualTo(16);
```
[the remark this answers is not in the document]
BenjaminPerryRoss: This is an existing test measuring our parsing of gfsh hints. Because a new hint was added for clear, it's now finding one more hint than it was before, but this is an expected difference so I've just updated the tests assertion to reflect that.

geode-gfsh/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/ClearCommandDUnitTest.java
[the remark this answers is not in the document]
DonalEvans: I guess I'm not understanding how the list regions command makes things more readable here. We only have two regions, and we already assert that they both exist in the setup phase, so having a list of them here in the even of a failure seems redundant. If they don't both exist by the time we get to calling `clear region` then the test should have already failed, and if the test fails after or during calling clear region, we already know that it's not because the regions don't exist.

geode-core/src/main/java/org/apache/geode/management/internal/i18n/CliStrings.java
```diff
@@ -1924,9 +1924,7 @@
   public static final String REMOVE__MSG__CLEARED_ALL_CLEARS = "Cleared all keys in the region";
```
DonalEvans, line 1924: I know this wasn't something introduced in this commit, but I feel that this naming was a typo and should be `REMOVE__MSG__CLEARED_ALL_KEYS `

geode-gfsh/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/ClearCommandDUnitTest.java
```diff
@@ -70,43 +72,49 @@ private static void populateTestRegions() {
...
 
   @Test
-  public void clearPartitionedRegion() {
-    String command = commandString + " --region=" + PARTITIONED_REGION_NAME;
+  public void clearSucceedsWithValidPartitionedRegion() {
+    String command = new CommandStringBuilder(CliStrings.CLEAR)
+        .addOption(CliStrings.CLEAR_REGION_NAME, PARTITIONED_REGION_NAME).getCommandString();
 
-    gfsh.executeAndAssertThat(command).statusIsSuccess();
+    gfsh.executeAndAssertThat(command);
```
DonalEvans, line 115: If we're executing and asserting, we should assert that the command is successful. If we don't care if the command succeeds or not, we should just use `execute()`. I think that in this case, we probably do care about the status of the command.

davebarnes97:
I see some names and details are still being finalized. Friendly reminder: don't forget the docs!

geode-core/src/main/java/org/apache/geode/management/internal/i18n/CliStrings.java
```diff
@@ -1914,9 +1921,10 @@
   public static final String REMOVE__MSG__KEY_EMPTY = "Key is Null";
   public static final String REMOVE__MSG__REGION_NOT_FOUND = "Region <{0}> Not Found";
   public static final String REMOVE__MSG__KEY_NOT_FOUND_REGION = "Key is not present in the region";
-  public static final String REMOVE__MSG__CLEARED_ALL_CLEARS = "Cleared all keys in the region";
-  public static final String REMOVE__MSG__CLEARALL_NOT_SUPPORTED_FOR_PARTITIONREGION =
-      "Option --" + REMOVE__ALL + " is not supported on partitioned region";
+  public static final String REMOVE__MSG__CLEARED_ALL_KEYS = "Cleared all keys in the region";
```
DonalEvans, line 1924: It might be best to rename and move this constant to reflect that going forward, only the clear command should be using the message, since remove --all is now deprecated.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/ClearCommand.java
```diff
@@ -0,0 +1,80 @@
...
+
+
+    Region region = cache.getRegion(regionPath);
+    DataCommandFunction clearfn = new DataCommandFunction();
+    DataCommandResult dataResult;
+    if (region == null) {
+      Set<DistributedMember> memberList = findAnyMembersForRegion(regionPath);
+
+      if (CollectionUtils.isEmpty(memberList)) {
+        return new ResultModel().createError(String.format(REGION_NOT_FOUND, regionPath));
```
DonalEvans, line 63: `new` is not needed here, since `createError` is a static call.

geode-gfsh/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/ClearCommandDUnitTest.java
```diff
@@ -0,0 +1,130 @@
...
+    String command = new CommandStringBuilder(CliStrings.CLEAR)
+        .addOption(CliStrings.CLEAR_REGION_NAME, PARTITIONED_REGION_NAME).getCommandString();
+
+    gfsh.executeAndAssertThat(command).statusIsSuccess();
+
+    assertThat(gfsh.getGfshOutput()).contains(REMOVE__MSG__CLEARED_ALL_KEYS);
+
+    server1.invoke(() -> verifyAllKeysAreRemoved(PARTITIONED_REGION_NAME));
+    server2.invoke(() -> verifyAllKeysAreRemoved(PARTITIONED_REGION_NAME));
+  }
```
DonalEvans, line 120: These tests could be replaced with a single parameterized version, which would prevent code duplication, and also allow easy addition of other region types, such as those using persistence, should the test need to be expanded to cover them.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/RemoveCommand.java
```diff
@@ -90,7 +90,14 @@ public ResultModel remove(
     }
 
     dataResult.setKeyClass(keyClass);
+    ResultModel result = null;
```
DonalEvans, line 93: Setting the value of `result` here is redundant, since it will always be assigned in the if/else block below.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/domain/DataCommandResult.java
```diff
@@ -240,6 +242,16 @@ public static DataCommandResult createRemoveResult(Object inputKey, Object value
     return result;
   }
 
+  public static DataCommandResult createClearResult(Throwable error, String errorString,
```
DonalEvans, line 245: This method is currently not being used anywhere.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/domain/DataCommandResult.java
```diff
@@ -381,6 +397,12 @@ public ResultModel toResultModel() {
     }
 
     ResultModel result = new ResultModel();
+
+    if (warningMessage != null && !warningMessage.isEmpty()) {
+      InfoResultModel info = result.addInfo(HEADER_INFO_SECTION);
```
DonalEvans, line 402: It might be more consistent to either call `setHeader` here instead of `addInfo`, or to rename `HEADER_INFO_SECTION` to `WARNING_INFO_SECTION` or something similar.

geode-gfsh/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/ClearCommandDUnitTest.java
```diff
@@ -0,0 +1,130 @@
...
+    locator.waitUntilRegionIsReadyOnExactlyThisManyServers("/" + PARTITIONED_REGION_NAME, 2);
+
+    VMProvider.invokeInEveryMember(ClearCommandDUnitTest::populateTestRegions, server1, server2);
+  }
+
+  private static void populateTestRegions() {
+    Cache cache = CacheFactory.getAnyInstance();
+
+    Region<String, String> replicateRegion = cache.getRegion(REPLICATE_REGION_NAME);
+    replicateRegion.put(EMPTY_STRING, "valueForEmptyKey");
```
DonalEvans, line 74: Is there a particular reason we put a value with an empty string as the key here? Just curious.

geode-core/src/main/java/org/apache/geode/management/internal/i18n/CliStrings.java
BenjaminPerryRoss: I've removed this message.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/functions/DataCommandFunction.java
BenjaminPerryRoss: I agree this could be confusing. I've added clear() as a helper method invoked by remove() when called with the 'ALL' flag to help make the relationship more visible.

geode-core/src/main/java/org/apache/geode/management/internal/i18n/CliStrings.java
```diff
@@ -816,6 +816,13 @@
   public static final String CLEAR_DEFINED_INDEX__SUCCESS__MSG =
       "Index definitions successfully cleared";
 
+  /* clear region */
+  public static final String CLEAR_REGION = "clear region";
```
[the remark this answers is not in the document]
BenjaminPerryRoss: I initially tried changing this to 'clear' as described in the ticket but this caused a conflict with the 'clear defined index' command. I've discussed it with Charlie and we decided that 'clear region' is appropriate (and it works around that issue) so I've updated the JIRA ticket/tracker story to reflect this.

geode-gfsh/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/ClearCommandDUnitTest.java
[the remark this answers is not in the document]
BenjaminPerryRoss: I've gone ahead and removed this command/assertion since it's not really necessary.

geode-core/src/main/java/org/apache/geode/management/internal/i18n/CliStrings.java
BenjaminPerryRoss: I agree, this could easily be corrected here.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/ClearCommand.java
```diff
@@ -0,0 +1,80 @@
...
+      }
+
+      DataCommandRequest request = new DataCommandRequest();
+      request.setCommand(CliStrings.REMOVE);
+      request.setRemoveAllKeys("ALL");
+      request.setRegionName(regionPath);
+      dataResult = callFunctionForRegion(request, clearfn, memberList);
+    } else {
+      dataResult = clearfn.remove(null, null, regionPath, "ALL",
```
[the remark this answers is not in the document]
BenjaminPerryRoss: The remove function already has logic for 'clearing' a region (since you can remove --all) so a clear is just a remove with the all flag set. Anil and I discussed this as well and I refactored the remove all code into a 'clear()' method on the function class so it was more obvious where the code for clear was being performed but there were challenges in implementing a new clear function and it would also have involved duplicate code.

geode-gfsh/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/ClearCommandDUnitTest.java
BenjaminPerryRoss: I was following the pattern set up by the Remove command (since they're very similar) for a lot of these choices and this is the pattern that was used in the DUnit tests for that command. With the value specifically I'm not sure it matters whether it's empty or not but it's easy to change if we think there's value.

geode-gfsh/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/ClearCommandDUnitTest.java
BenjaminPerryRoss: This was removed by mistake, it should be back in there now.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/functions/DataCommandFunction.java
```diff
@@ -276,7 +276,7 @@ public DataCommandResult remove(String key, String keyClass, String regionName,
 
     if (StringUtils.isEmpty(removeAllKeys) && (key == null)) {
       return DataCommandResult.createRemoveResult(null, null, null,
-          CliStrings.REMOVE__MSG__KEY_EMPTY, false);
+          "BR" + CliStrings.REMOVE__MSG__KEY_EMPTY, false);
```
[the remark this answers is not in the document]
BenjaminPerryRoss: Done

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/functions/DataCommandFunction.java
```diff
@@ -313,22 +313,21 @@ public DataCommandResult remove(String key, String keyClass, String regionName,
...
-          return DataCommandResult.createRemoveInfoResult(key, null, null,
-              CliStrings.REMOVE__MSG__CLEARALL_NOT_SUPPORTED_FOR_PARTITIONREGION, false);
-        }
+        return clear(region, regionName);
       }
     }
   }
 
+  public DataCommandResult clear(Region region, String regionName) {
+    DataPolicy policy = region.getAttributes().getDataPolicy();
```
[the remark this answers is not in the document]
BenjaminPerryRoss: I've removed it.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/ClearCommand.java
BenjaminPerryRoss: I've removed this.

geode-gfsh/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/ClearCommandDUnitTest.java
BenjaminPerryRoss: This is a good idea, I'll make this parameterized.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/domain/DataCommandResult.java
BenjaminPerryRoss: I'm reluctant to change the addInfo to setHeader because I want the structure of the output to match remove as closely as possible so I opted to change the name to WARNING_INFO_SECTION as this section isn't used in any other context.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/RemoveCommand.java
BenjaminPerryRoss: I've removed the redundant initialization.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/domain/DataCommandResult.java
BenjaminPerryRoss: It was left in by mistake after making and then reverting some changes based on PR feedback. I'll remove this.

geode-core/src/distributedTest/java/org/apache/geode/cache30/RegionReliabilityTestCase.java
```diff
@@ -1341,7 +1341,7 @@ public void run() {
       // pass
     }
 
-    // CLEAR
+    // CLEAR_REGION
```
DonalEvans, line 1344: I think that several of these renames were accidentally included when changing the Cli string. They should be put back how they were.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClearRvvLockingDUnitTest.java
```diff
@@ -51,7 +51,7 @@
 
 /**
  * Test class to verify proper RVV locking interaction between entry operations such as PUT/REMOVE
- * and the CLEAR region operation
+ * and the CLEAR_REGION region operation
```
DonalEvans, line 54: Another accidental rename.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClearTXLockingDUnitTest.java
```diff
@@ -43,19 +43,20 @@
 import org.apache.geode.test.dunit.rules.DistributedRule;
 
 /**
- * Test class to verify proper locking interaction between transactions and the CLEAR region
+ * Test class to verify proper locking interaction between transactions and the CLEAR_REGION region
```
DonalEvans, line 46: More accidental renames in this file.

geode-core/src/main/java/org/apache/geode/internal/cache/DiskStoreImpl.java
```diff
@@ -3509,7 +3509,7 @@ public String toString() {
       if (de != null) {
         sb.append(" key=").append(de.getKey());
       } else {
-        sb.append(" <END CLEAR>");
+        sb.append(" <END CLEAR_REGION>");
```
DonalEvans, line 3512: Accidental rename.

geode-gfsh/src/test/java/org/apache/geode/management/internal/cli/commands/ClearCommandTest.java
```diff
@@ -0,0 +1,93 @@
...
+import org.apache.geode.cache.Region;
+import org.apache.geode.distributed.DistributedMember;
+import org.apache.geode.internal.cache.InternalCache;
+import org.apache.geode.management.internal.cli.domain.DataCommandRequest;
+import org.apache.geode.management.internal.cli.domain.DataCommandResult;
+import org.apache.geode.management.internal.cli.functions.DataCommandFunction;
+import org.apache.geode.management.internal.cli.result.model.ResultModel;
+import org.apache.geode.test.junit.rules.GfshParserRule;
+
+public class ClearCommandTest {
```
DonalEvans, line 32: This feels like more of an integration test as it's currently written, since it's not directly testing the ClearCommand class but rather how that class is used by gfsh. Instead of using `gfsh.executeAndAssertThat`, it might better to call `ClearCommand.clear()` directly and assert on the returned `ResultModel`.

It would also be good to verify that `callFunctionForRegion()` and `clearfn.remove()` are called with the expected/correct arguments, rater than just asserting the success/failure of the command, since otherwise the last two test cases end up being indistinguishable in terms of what they're asserting and the actual difference in behaviour between the two code paths isn't being verified.

geode-gfsh/src/test/java/org/apache/geode/management/internal/cli/commands/ClearCommandTest.java
```diff
@@ -0,0 +1,93 @@
...
+  DataCommandResult dataResult;
+
+  @Before
+  public void setup() {
+    cache = mock(InternalCache.class);
+    command = spy(new ClearCommand());
+    region = mock(Region.class);
+    dataResult = mock(DataCommandResult.class);
+
+    membersList = new HashSet<DistributedMember>();
```
DonalEvans, line 54: It's not necessary to specify the `DistributedMember` here, since it's implicit in the declaration of the `membersList` set.

geode-gfsh/src/test/java/org/apache/geode/management/internal/cli/commands/ClearCommandTest.java
```diff
@@ -0,0 +1,93 @@
...
+
+  @ClassRule
+  public static GfshParserRule gfsh = new GfshParserRule();
+
+  static final String regionName = "regionName";
+  static final String success = "SUCCESS";
+
+  InternalCache cache;
+  ClearCommand command;
+  Region region;
```
DonalEvans, line 42: To prevent compiler warnings, this should probably be `Region<Object, Object>` and a `@SuppressWarnings("unchecked")` tag should be added to the `setup()` method.

geode-gfsh/src/test/java/org/apache/geode/management/internal/cli/commands/ClearCommandTest.java
BenjaminPerryRoss: Originally I was calling clear directly but that introduce complications with the test needing to hardcode the parsing of ResultModel since we weren't using the gfsh tools that come with the gfsh rule. Ultimately I decided that using the gfsh rule and relying on the gfsh command line parsing to call the the actual method we're testing was the lesser of two evils in that regard. 

I have added the verifications, I agree they improve the quality of the test.

geode-gfsh/src/test/java/org/apache/geode/management/internal/cli/commands/ClearCommandTest.java
BenjaminPerryRoss: Done.

geode-gfsh/src/test/java/org/apache/geode/management/internal/cli/commands/ClearCommandTest.java
BenjaminPerryRoss: Done.

geode-core/src/distributedTest/java/org/apache/geode/cache30/RegionReliabilityTestCase.java
BenjaminPerryRoss: Yep, this was a careless rename on my part. I changed these back.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClearTXLockingDUnitTest.java
```diff
@@ -49,13 +49,14 @@
  * <p>
  * GEODE-1740: It was observed that operations performed within a transaction were not holding
  * region modification locks for the duration of commit processing. This lock is used to ensure
- * region consistency during CLEAR processing. By not holding the lock for the duration of commit
+ * region consistency during CLEAR_REGION processing. By not holding the lock for the duration of
```
DonalEvans, line 52: Here. Also consider fixing the formatting on the comment, since spotless has kind of messed it up following the accidental rename.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClearTXLockingDUnitTest.java
```diff
@@ -49,13 +49,14 @@
...
+ * commit
  * processing, a window was opened that allowed region operations such as clear to occur in
  * mid-commit.
  *
  * <p>
  * The fix for GEODE-1740 was to acquire and hold read locks for any region involved in the commit.
- * This forces CLEAR to wait until commit processing is complete.
+ * This forces CLEAR_REGION to wait until commit processing is complete.
```
DonalEvans, line 59: Here.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClearTXLockingDUnitTest.java
```diff
@@ -273,7 +274,7 @@ private void resetArmHook(String rname) {
   }
 
   /**
-   * Wait to be notified it is time to perform region operation (i.e. CLEAR)
+   * Wait to be notified it is time to perform region operation (i.e. CLEAR_REGION)
```
DonalEvans, line 277: Here.

geode-core/src/main/java/org/apache/geode/management/internal/i18n/CliStrings.java
DonalEvans: I still think that renaming this to something like "CLEAR_REGION_CLEARED_ALL_KEYS" and moving it to the "clear region" section of constants would be worthwhile. It seems like a good idea to keep all clear-related stuff in one place.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClearTXLockingDUnitTest.java
BenjaminPerryRoss: Done.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClearTXLockingDUnitTest.java
BenjaminPerryRoss: Done.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClearTXLockingDUnitTest.java
BenjaminPerryRoss: Done.

geode-core/src/main/java/org/apache/geode/management/internal/i18n/CliStrings.java
BenjaminPerryRoss: Done.

BenjaminPerryRoss:
@davebarnes97 So this ticket is being merged against a feature branch containing a bunch of changes for the PR clear feature, and I believe that there is a separate story (potentially still to be created) that covers documenting this alongside the other changes being made for PR clear. If you feel that we should do that work on this story instead please let me know and I'll add it.

davebarnes97:
> @davebarnes97 So this ticket is being merged against a feature branch containing a bunch of changes for the PR clear feature, and I believe that there is a separate story (potentially still to be created) that covers documenting this alongside the other changes being made for PR clear. If you feel that we should do that work on this story instead please let me know and I'll add it.

Just want to make sure the docs get done. A "separate story potentially still to be created" is kinda lame. If that doc ticket exists, please cite it here; if not, please create it -- that will satisfy my request. Thank you.

BenjaminPerryRoss:
@davebarnes97 that seems like a good idea. I've created the Jira for the documentation work and added it as a child of the base feature here https://issues.apache.org/jira/browse/GEODE-8074.

PR #4819, opened 2020-03-16
GEODE-7885: Improve RedisLockService - fix synchronization and memory leak

jdeppe-pivotal:
- RedisLockService was not able to automatically free keys/locks after
  unlocking since keys were also beging referenced in regions.
- Fix synchronization while iterating over keySet to avoid
  ConcurrentModifiactionExceptions.

Co-Authored-By: Sarah Abbey <sabbey@pivotal.io>
Co-Authored-By: Ray Ingles <ringles@pivotal.io>
Co-Authored-By: John Hutchison <jhutchison@pivotal.io>

### DIFF
--- a/geode-redis/src/main/java/org/apache/geode/redis/internal/RedisLockService.java
+++ b/geode-redis/src/main/java/org/apache/geode/redis/internal/RedisLockService.java
@@ -72,37 +72,40 @@ public class RedisLockService implements RedisLockServiceMBean {
     }
 
     KeyHashIdentifier lockKey = new KeyHashIdentifier(key.toBytes());
-    Lock lock = new ReentrantLock();
-    Lock oldLock = weakReferencesTolocks.putIfAbsent(lockKey, lock);
+    KeyHashIdentifier referencedKey = lockKey;
 
-    if (oldLock != null) {
-      lock = oldLock;
+    Lock lock;
+    do {
+      lock = new ReentrantLock();
+      Lock oldLock = weakReferencesTolocks.putIfAbsent(lockKey, lock);
 
-      // we need to get a reference to the actual key object
-      // so that the backing WeakHashMap does not clean it up
-      // when garbage collection happens.
-      lockKey = getReferenceToLockKey(lockKey);
-    }
+      if (oldLock != null) {
+        lock = oldLock;
+
+        // we need to get a reference to the actual key object
+        // so that the backing WeakHashMap does not clean it up
+        // when garbage collection happens.
+        referencedKey = getReferenceToLockKey(lockKey);
+      }
+    } while (referencedKey == null);
 
     if (!lock.tryLock(timeoutMS, TimeUnit.MILLISECONDS)) {
       throw new TimeoutException("Couldn't get lock for " + lockKey.toString());
     }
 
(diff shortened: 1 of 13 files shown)


jdeppe-pivotal:
@prettyClouds @xtreme-prasath - for your review

geode-redis/src/main/java/org/apache/geode/redis/internal/RedisLockService.java
```diff
@@ -70,29 +71,38 @@ public AutoCloseableLock lock(ByteArrayWrapper key) throws InterruptedException
...
-        if (keyInSet.equals(key)) {
-          key = keyInSet;
-          break;
-        }
-      }
+
+      // we need to get a reference to the actual key object
+      // so that the backing WeakHashMap does not clean it up
+      // when garbage collection happens.
+      lockKey = getReferenceToLockKey(lockKey);
```
kohlmu-pivotal: I don't understand this logic. If `weakReferencesTolocks.putIfAbsent(lockKey, lock)` returns a value, then you already have a reference to the lockKey. OR am I mistaken about this?

geode-redis/src/main/java/org/apache/geode/redis/internal/RedisLockService.java
```diff
@@ -70,29 +71,38 @@ public AutoCloseableLock lock(ByteArrayWrapper key) throws InterruptedException
...
 
-  int getMapSize() {
-    return map.size();
+  private KeyHashIdentifier getReferenceToLockKey(KeyHashIdentifier lockKey) {
+    KeyHashIdentifier referenceToLockKey = null;
+
+    synchronized (weakReferencesTolocks) {
+      for (KeyHashIdentifier keyInSet : weakReferencesTolocks.keySet()) {
+        if (keyInSet.equals(lockKey)) {
+          referenceToLockKey = keyInSet;
```
kohlmu-pivotal: you can just return `keyInSet` without having to set it onto `referenceToLockKey` and return `null` on line 105

geode-redis/src/main/java/org/apache/geode/redis/internal/RedisLockService.java
jdeppe-pivotal: The essence here is that you need to get a reference to the actual key in the map (if you don't it might get GC'd). Consider the first pass through all of this and I'm trying to lock key "ABC" and succeed. I have a reference to the key and all is good. If another thread comes to look for key "ABC" it will have a different 'key' object. So if we get the lock out of the map but don't also get the actual key then that lock might get GC'd if we don't retain a hard reference to its key.

kohlmu-pivotal: What I was trying to say was... in the java doc for `putIfAbsent`... 

> If the specified key is not already associated with a value (or is mapped
     * to {@code null}) associates it with the given value and returns
     * {@code null}, else returns the current value.

Which makes the whole loop of iterating over the keySet just extra work.

kohlmu-pivotal: Correction... it return the VALUE not the KEY.... ok.... LGTM

PR #4820, opened 2020-03-16
GEODE-7826: start jmx-manager if the Geode Management Rest Service is enabled

dschneider-pivotal:
Before the geode management rest service web app is added,
the code now starts the jmx-manager. If it can not be started then
an error will be reported.

### For all changes:

### Note:


dschneider-pivotal:
@jmelchio

geode-core/src/test/java/org/apache/geode/distributed/internal/InternalLocatorTest.java
```diff
@@ -0,0 +1,157 @@
...
+import org.apache.geode.internal.cache.InternalRegionFactory;
+import org.apache.geode.internal.security.SecurableCommunicationChannel;
+import org.apache.geode.logging.internal.LoggingSession;
+import org.apache.geode.management.internal.BaseManagementService;
+
+public class InternalLocatorTest {
+  @Test
+  public void startClusterManagementServiceWithRestServiceEnabledInvokesStartManager()
+      throws IOException {
+    LoggingSession loggingSession = mock(LoggingSession.class);
```
demery-pivotal: I think most of the common setup stuff could be moved into a setup method. Then only the stuff that differs from one test to another would remain. That would make the test-specific details stand out more, and would more clearly express the intent of each test.

If only a few setup details remain in each test, and the intent is obvious, that's enough. If there are more than a few details, perhaps they can be hidden in helper methods with that express the intent.

geode-core/src/test/java/org/apache/geode/distributed/internal/InternalLocatorTest.java
```diff
@@ -0,0 +1,157 @@
...
+        null, null, distributionConfig, null);
+    InternalCacheForClientAccess cache = mock(InternalCacheForClientAccess.class);
+    InternalRegionFactory regionFactory = mock(InternalRegionFactory.class);
+    when(cache.createInternalRegionFactory(RegionShortcut.REPLICATE)).thenReturn(regionFactory);
+    when(cache.getOptionalService(HttpService.class))
+        .thenReturn(Optional.of(mock(HttpService.class)));
+    when(cache.getCacheForProcessingClientRequests()).thenReturn(cache);
+    BaseManagementService managementService = mock(BaseManagementService.class);
+    BaseManagementService.setManagementService(cache, managementService);
+    internalLocator.setInternalCache(cache);
```
demery-pivotal: Instead of adding a new test-only method, try this:
- Add a new `Supplier<InternalCache>` parameter to the `InternalLocator` constructor, and let the constructor store the supplier in a field for later use.
- Change `InternalLocator.getCache()` to call the supplier instead of calling `GemFireCacheImpl.getInstance()` directly.
- Change `InternalLocator.createLocator(…)` and `InternalLocatorIntegrationTest` to pass `GemFireCacheImpl::getInstance` to the `InternalLocator` constructor.
- In these new tests, pass the mock cache to the `InternalLocator` constructor.

geode-core/src/main/java/org/apache/geode/distributed/internal/InternalLocator.java
```diff
@@ -815,6 +822,10 @@ void startClusterManagementService() throws IOException {
 
     if (distributionConfig.getEnableManagementRestService()) {
       myCache.getOptionalService(HttpService.class).ifPresent(x -> {
+        ManagementService managementService = ManagementService.getManagementService(myCache);
```
jinmeiliao: Maybe put a comment here stating why jmxManager is needed when starting management rest service

dschneider-pivotal: added a comment

geode-core/src/test/java/org/apache/geode/distributed/internal/InternalLocatorTest.java
dschneider-pivotal: I refactored common code into a couple of helper methods. I hope that at some point this unit test will have more coverage for this class so I did not put this code into a common setup method. I think most of this code is specific to testing the "startClusterManagementService" method.

geode-core/src/test/java/org/apache/geode/distributed/internal/InternalLocatorTest.java
dschneider-pivotal: I was able to get rid of the new test-only method. But I did it in a different way that did not require the constructor to change. The method being tested now takes as a parameter the cache it should use. Let me know what you think.

dschneider-pivotal:
the failure in IntegrationTestOpenJDK11 is the known issue GEODE-7901

PR #4821, opened 2020-03-17
GEODE-7886 Update strings tests to cover sad as well as happy paths

ringles:
This patch updates the tests for the Redis String commands to cover error conditions and edge cases, as well as concurrent executions. It updates the command executor unit tests to cover low-level API errors (incorrect arguments, etc.).


ringles:
@prettyClouds @sabbeyPivotal @xtreme-prasath  @jhutchison - For review and delectation

geode-redis/src/integrationTest/java/org/apache/geode/redis/StringsIntegrationTest.java
```diff
@@ -219,30 +351,313 @@ public void testMSetAndGet() {
...
+
+    future1.get();
+    future2.get();
 
-    assertTrue(response == 1);
+    assertThat(jedis.get("contestedKey")).isEqualTo(Integer.toString(2 * ITERATION_COUNT));
+  }
+
+  @Test
+  // @todo: not looked over for current release
```
jdeppe-pivotal: There are a couple of these comments. Since all of these good changes are already being applied to this class, is there any reason not to validate these remaining few tests? Or is this more work than what it appears on the surface?

geode-redis/src/main/java/org/apache/geode/redis/internal/RedisLockKey.java
```diff
@@ -0,0 +1,48 @@
...
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ *
+ */
+
+package org.apache.geode.redis.internal;
+
+import org.apache.commons.codec.digest.MurmurHash3;
+
+public class RedisLockKey {
```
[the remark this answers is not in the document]
sabbeyPivotal: Deleted! Thank you.

geode-redis/src/integrationTest/java/org/apache/geode/redis/StringsIntegrationTest.java
sabbeyPivotal: I think this story had gone on a bit longer than was originally planned and the scope kept expanding, so creating stories to go back and review the implementation/tests for the commands that weren't part of this release was a way of reining in the scope.  @ringles may be able to provide more context though.

ringles: The issue is that there are higher-priority commands that need love and attention before we circle back to these. There are stories in the backlog for them, but they aren't our current top priority.

geode-redis/src/main/java/org/apache/geode/redis/internal/RedisLockKey.java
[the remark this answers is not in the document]
ringles: On it!

PR #4822, opened 2020-03-17
GEODE-7884: server hangs due to IllegalStateException

bschuchardt:
Added cancellation check before scheduling an idle-timeout or
ack-wait-threshold timer task.  I had to add a new method to
SystemTimerTask and then noticed there were no tests for SystemTimer, so
I cleaned up that class and added tests.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/SystemTimer.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/SystemTimer.java
@@ -62,41 +62,49 @@ public class SystemTimer {
 
   @Override
   public String toString() {
-    StringBuffer sb = new StringBuffer();
-    sb.append("SystemTimer[");
-    sb.append("swarm = " + swarm);
-    sb.append("]");
-    return sb.toString();
+    return "SystemTimer["
+        + "swarm = " + swarm
+        + "]";
   }
 
   /**
    * List of all of the swarms in the system
    */
   @MakeNotStatic
-  private static final HashMap<DistributedSystem, List> allSwarms = new HashMap();
+  private static final HashMap<DistributedSystem, List<WeakReference<SystemTimer>>> allSwarms =
+      new HashMap<>();
 
   /**
    * Add the given timer is in the given swarm. Used only by constructors.
    *
    * @param swarm swarm to add the timer to
-   * @param t timer to add
+   * @param systemTimer timer to add
    */
-  private static void addToSwarm(DistributedSystem swarm, SystemTimer t) {
+  private static void addToSwarm(DistributedSystem swarm, SystemTimer systemTimer) {
     // Get or add list of timers for this swarm...
-    ArrayList /* ArrayList<WeakReference<SystemTimer>> */ swarmSet;
+    List<WeakReference<SystemTimer>> swarmSet;
     synchronized (allSwarms) {
-      swarmSet = (ArrayList) allSwarms.get(swarm);
+      swarmSet = allSwarms.get(swarm);
       if (swarmSet == null) {
-        swarmSet = new ArrayList();
+        swarmSet = new ArrayList<>();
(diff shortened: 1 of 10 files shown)


geode-core/src/main/java/org/apache/geode/internal/tcp/ConnectionTable.java
```diff
@@ -519,8 +519,12 @@ private void scheduleIdleTimeout(Connection conn) {
           if (!closed) {
             IdleConnTT task = new IdleConnTT(conn);
             conn.setIdleTimeoutTask(task);
-            getIdleConnTimer().scheduleAtFixedRate(task, owner.idleConnectionTimeout,
-                owner.idleConnectionTimeout);
+            synchronized (task) {
```
echobravopapa: does synchronizing here relate to removing the `continue`s in the timer code?

geode-core/src/main/java/org/apache/geode/internal/SystemTimer.java
```diff
@@ -181,88 +167,67 @@ private static void sweepAllSwarms() {
...
-            continue;
-          }
-          if (t2 == t) {
-            it.remove();
+        Iterator<WeakReference<SystemTimer>> swarmIterator = swarmSet.iterator();
+        while (swarmIterator.hasNext()) {
+          WeakReference<SystemTimer> ref = swarmIterator.next();
+          SystemTimer timer = ref.get();
+          if (timer == null || timer == timerToRemove) {
+            swarmIterator.remove();
```
echobravopapa: no more `continue`...

geode-core/src/main/java/org/apache/geode/internal/tcp/ConnectionTable.java
bschuchardt: The synchronizations and cancel checks fix GEODE-7884.  The SystemTimer changes, except for adding isCancelled(), are unrelated.  I made those changes so that I could test isCancelled() and add test coverage for SystemTimer.

geode-core/src/main/java/org/apache/geode/internal/SystemTimer.java
bschuchardt: _continue_ isn't needed - this is the end of the loop

geode-core/src/main/java/org/apache/geode/internal/tcp/ConnectionTable.java
echobravopapa: roger that

geode-core/src/main/java/org/apache/geode/internal/tcp/Connection.java
```diff
@@ -1411,11 +1411,15 @@ private void close(String reason, boolean cleanupEndpoint, boolean p_removeEndpo
     // This cancels the idle timer task, but it also removes the tasks reference to this connection,
     // freeing up the connection (and it's buffers for GC sooner.
     if (idleTask != null) {
-      idleTask.cancel();
+      synchronized (idleTask) {
```
Bill: I see `idleTask` is never changed from non-null back to `null` ✓ OTOH since the variable is neither `volatile` nor guarded when we do the null-check, it is possible that if `setIdleTimeoutTask()` was called from a different thread than `close()` we could leak tasks.

I feel like maybe `idleTask` and `ackTimeoutTask` need to be volatile for this reason. Or maybe they need to be `final` and set from the constructor.

bschuchardt: It's likely that these variables are set in one thread and then accessed in another, so I've followed your advice and made them volatile.  I've also added checks in the two timer tasks to cancel themselves if they see that the Connection they're monitoring has closed.  That should stop any remaining leaks.

PR #4823, opened 2020-03-18
GEODE-7851: Pulse Oauth Configuration

jinmeiliao:
* create an OauthSecurityConfig to configure spring using oauth
* add PULSE as one of the oauth-enabled-component, and if pulse is set to use oauth, set the OauthSecurityConfig as the active security profile
* use pulse.properties in the locator's working dir to externalize pulse authentication provider configuration

### For all changes:

### Note:


jinmeiliao:
@jmelchio

jinmeiliao:
This PR only takes care of the how to configure Pulse to use oAuth. There will be a 2nd PR to funnel the access token to the security manager in the jmx connection.

geode-core/src/main/java/org/apache/geode/management/internal/ManagementAgent.java
```diff
@@ -192,9 +194,20 @@ private void loadWebApplications() {
...
       }
     } else {
+      String[] authTokenEnabledComponents = config.getSecurityAuthTokenEnabledComponents();
+      boolean pulseOauth = Arrays.stream(authTokenEnabledComponents)
+          .anyMatch(AuthTokenEnabledComponents::hasPulse);
       String pwFile = this.config.getJmxManagerPasswordFile();
-      if (securityService.isIntegratedSecurity() || StringUtils.isNotBlank(pwFile)) {
+      if (securityService.isIntegratedSecurity()) {
+        if (pulseOauth) {
+          System.setProperty("spring.profiles.active", "pulse.authentication.oauth");
```
dschneider-pivotal: this code repeats "spring.profiles.active" four times. Consider changing it to just call System.setProperty once (at the end). You could introduce a localvar "String activeProfileValue" and have each of these lines set that local var.

geode-core/src/main/java/org/apache/geode/management/internal/ManagementAgent.java
```diff
@@ -192,9 +194,20 @@ private void loadWebApplications() {
         logger.debug(message);
       }
     } else {
+      String[] authTokenEnabledComponents = config.getSecurityAuthTokenEnabledComponents();
+      boolean pulseOauth = Arrays.stream(authTokenEnabledComponents)
```
dschneider-pivotal: Since you only use "pulseOauth" inside the if(isIntegratedSecurity) would it be better to initialize the variable inside that if?

geode-pulse/src/main/java/org/apache/geode/tools/pulse/internal/security/OauthSecurityConfig.java
```diff
@@ -0,0 +1,100 @@
...
+  @Override
+  protected void configure(HttpSecurity http) throws Exception {
+    http.authorizeRequests(authorize -> authorize
+        .anyRequest().authenticated())
+        .oauth2Login(withDefaults());
+  }
+
+  @Bean
+  public ClientRegistrationRepository clientRegistrationRepository() {
+    return new InMemoryClientRegistrationRepository(this.uaaClientRegistration());
```
dschneider-pivotal: It looks like we only have uaa support here. So if a customer is using oauth2 will it only work with uaa or can the configure it for some other oauth2 authentication server?

jinmeiliao: actually that name can be changed

geode-pulse/src/main/java/org/apache/geode/tools/pulse/internal/security/OauthSecurityConfig.java
```diff
@@ -0,0 +1,100 @@
...
+  private String userInfoUri;
+  @Value("${pulse.oauth.jwkSetUri}")
+  private String jwkSetUri;
+  @Value("${pulse.oauth.userNameAttributeName}")
+  private String userNameAttributeName;
+
+  @Override
+  protected void configure(HttpSecurity http) throws Exception {
+    http.authorizeRequests(authorize -> authorize
+        .anyRequest().authenticated())
```
jmelchio: Do we need to add something here to take of where the app goes after logging out? For OAuth this should be different from Default or Gemfire flows I believe.

jinmeiliao: In your test, what does it do?

jinmeiliao: > Does this PR need to update some of the doc files to describe how to configure pulse to use oauth2?

Yes, I will add a task to the tracker story to update the doc. PR for that is to be followed.

jmelchio: > In your test, what does it do?

Don't have that covered in test yet but it follows the `logoutHandler` set up in the `DefaultSecurityConfig`. In the OAuth case I think it should invalidate the token someway. I assume there is a built-in way that Spring handles this so we should try to make use of that.

jinmeiliao: In this changeset, I have OauthSecurityConfig directly extends `WebSecurityConfigurerAdapter`. It does not use anything defined in DefaultSecurityConfig. So I hope spring would pick up the flow.

geode-assembly/src/integrationTest/java/org/apache/geode/tools/pulse/PulseSecurityConfigOauthProfileTest.java
```diff
@@ -0,0 +1,83 @@
...
+    // copy the pulse.properties to the locator's working dir. Pulse will use the locator's working
+    // dir as classpath to search for this property file
+    pulsePropertyFile = new File(locator.getWorkingDir(), "pulse.properties");
+    Properties properties = new Properties();
+    properties.setProperty("pulse.oauth.provider", "uaa");
+    properties.setProperty("pulse.oauth.clientId", "pulse");
+    properties.setProperty("pulse.oauth.clientSecret", "secret");
+    // have the authorization uri point to a known uri that locator itself can serve
+    properties.setProperty("pulse.oauth.authorizationUri",
+        "http://localhost:" + locator.getHttpPort() + "/management");
```
demery-pivotal: It looks as if this uses the /management API itself to somehow mimic UAA (or some other OAuth server). How does that work?

geode-assembly/src/integrationTest/java/org/apache/geode/tools/pulse/PulseSecurityConfigOauthProfileTest.java
```diff
@@ -0,0 +1,83 @@
...
+  @Rule
+  public GeodeHttpClientRule client = new GeodeHttpClientRule(locator::getHttpPort);
+
+  @Test
+  public void testLogin() throws Exception {
+    HttpResponse response = client.get("/pulse/login.html");
+    // the request is redirect to the authorization uri configured before
+    assertResponse(response).hasStatusCode(200).hasResponseBody()
+        .containsOnlyOnce("latest")
+        .containsOnlyOnce("supported");
```
demery-pivotal: Why only once? This looks fragile: If someone changes the page to show these words more than once, the test will fail, even if the request redirects properly.

geode-assembly/src/integrationTest/java/org/apache/geode/tools/pulse/PulseSecurityConfigOauthProfileTest.java
```diff
@@ -0,0 +1,83 @@
...
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+
+import org.apache.geode.examples.SimpleSecurityManager;
+import org.apache.geode.test.junit.categories.PulseTest;
+import org.apache.geode.test.junit.rules.GeodeHttpClientRule;
+import org.apache.geode.test.junit.rules.LocatorStarterRule;
+
+@Category({PulseTest.class})
+public class PulseSecurityConfigOauthProfileTest {
```
demery-pivotal: Capitalize the A in OAuth (here and elsewhere)

geode-pulse/src/integrationTest/java/org/apache/geode/tools/pulse/security/OauthSecurityConfigTest.java
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  @Test
+  public void redirectToOauth() throws Exception {
+    mvc.perform(get("/login.html")).andExpect(status().is3xxRedirection())
+        .andExpect(redirectedUrl(("http://localhost/oauth2/authorization/uaa")));
+
+    mvc.perform(get("http://localhost/oauth2/authorization/uaa"))
+        .andExpect(status().is3xxRedirection())
+        .andExpect(redirectedUrlPattern("http://xyz.com/uaa/oauth/**"));
```
demery-pivotal: xyz.com is an actual domain name. Please use a test-specific domain name such as:

- `example.com` (`example.com` is intended mostly for documentation, but people use it for testing)
- `<any-subdomain>.test` (`.test` is a special top-level domain name designed for testing)
- `<any-subdomain>.example` (`.example` is another special top-level domain name)

geode-pulse/src/integrationTest/resources/pulse.properties
```diff
@@ -0,0 +1,23 @@
...
+# the License.
+#
+
+pulse.oauth.provider=uaa
+pulse.oauth.clientId=pulse
+pulse.oauth.clientSecret=secret
+pulse.oauth.authorizationUri=http://xyz.com/uaa/oauth/authorize
+pulse.oauth.tokenUri=http://xyz.com/uaa/oauth/token
+pulse.oauth.userInfoUri=http://xyz.com/uaa/userinfo
+pulse.oauth.jwkSetUri=http://xyz.com/uaa/token_keys
```
demery-pivotal: Someone else owns xyz.com, so we should not use it.

geode-assembly/src/integrationTest/java/org/apache/geode/tools/pulse/PulseSecurityConfigOauthProfileTest.java
jinmeiliao: this changeset only makes sure pulse can be configured to talk to an authorizationUri. Whatever happens after that, is going to be in Joris's PR.

jinmeiliao: As the test comment states "this test just makes sure the property file in the locator's working dir
gets properly read and used in the oauth security configuration". so it only makes sure the configuration in the property file gets used by Pulse, exactly where that URI points to, is irrelavant.

geode-assembly/src/integrationTest/java/org/apache/geode/tools/pulse/PulseSecurityConfigOauthProfileTest.java
jinmeiliao: This is a random URL I picked up only for testing purpose. I happen to pick the /management uri, which returns json content {"latest":"...", "supported":[...]}. It is just a way to test the redirect works. It's not likely that url response would change and if it does, yes, developer will need to change the test.

geode-pulse/src/integrationTest/java/org/apache/geode/tools/pulse/security/OauthSecurityConfigTest.java
jinmeiliao: will do

geode-pulse/src/integrationTest/resources/pulse.properties
jinmeiliao: will do.

geode-assembly/src/integrationTest/java/org/apache/geode/tools/pulse/PulseSecurityConfigOauthProfileTest.java
jmelchio: > Why only once? This looks fragile: If someone changes the page to show these words more than once, the test will fail, even if the request redirects properly.

From what I know that is highly unlikely. I think it is safe in this case.

demery-pivotal: Each assertion should express a requirement that the test cares about. "Only once" is not a requirement here. Fix it.

jinmeiliao: done.

PR #4824, opened 2020-03-18
GEODE-7565: Allow gateway receivers with same host and port

alb3rtobr:
**Long version:**
https://cwiki.apache.org/confluence/display/GEODE/Allow+same+host+and+port+for+all+gateway+receivers

**Summary:**
There is a problem with Geode WAN replication when GW receivers are configured with the same hostname-for-senders and port on all servers. The reason for such a setup is deploying Geode cluster on a Kubernetes cluster where all GW receivers are reachable from the outside world on the same VIP and port.

The problem experienced is that shutting down one server is stopping replication to this cluster until the server is up again. This is because Geode incorrectly assumes there are no more alive servers when just one of them is down, because since they share hostname-for-senders and port, they are treated as one same server.

With these changes locator is able to distinguish the different receivers using the same hostname and port so replication is not impacted when one server is stopped.

( Link to thread in dev mailing list: https://markmail.org/thread/6qakx67rxiokdsec )

### DIFF
--- a/geode-core/src/integrationTest/java/org/apache/geode/cache/client/internal/ConnectionPoolImplJUnitTest.java
+++ b/geode-core/src/integrationTest/java/org/apache/geode/cache/client/internal/ConnectionPoolImplJUnitTest.java
@@ -168,7 +168,7 @@ public class ConnectionPoolImplJUnitTest {
     assertTrue(pool2.isDestroyed());
   }
 
-  @Test(timeout = 5000)
+  @Test
   public void testExecuteOp() throws Exception {
     CacheServer server1 = cache.addCacheServer();
     CacheServer server2 = cache.addCacheServer();
--- a/geode-core/src/integrationTest/java/org/apache/geode/cache/client/internal/ConnectionPoolImplJUnitTest.java
+++ b/geode-core/src/integrationTest/java/org/apache/geode/cache/client/internal/ConnectionPoolImplJUnitTest.java
@@ -168,7 +168,7 @@ public class ConnectionPoolImplJUnitTest {
     assertTrue(pool2.isDestroyed());
   }
 
-  @Test
+  @Test(timeout = 5000)
   public void testExecuteOp() throws Exception {
     CacheServer server1 = cache.addCacheServer();
     CacheServer server2 = cache.addCacheServer();
--- a/geode-core/src/main/java/org/apache/geode/cache/client/internal/EndpointManagerImpl.java
+++ b/geode-core/src/main/java/org/apache/geode/cache/client/internal/EndpointManagerImpl.java
@@ -37,7 +37,7 @@ import org.apache.geode.logging.internal.log4j.api.LogService;
 public class EndpointManagerImpl implements EndpointManager {
   private static final Logger logger = LogService.getLogger();
 
-  private volatile Map<ServerLocation, Endpoint> endpointMap = Collections.emptyMap();
+  private volatile Map<DistributedMember, Endpoint> endpointMap = Collections.emptyMap();
   private final Map<ServerLocation, ConnectionStats> statMap = new HashMap<>();
   private final DistributedSystem ds;
   private final String poolName;
@@ -56,17 +56,17 @@ public class EndpointManagerImpl implements EndpointManager {
 
   @Override
   public Endpoint referenceEndpoint(ServerLocation server, DistributedMember memberId) {
-    Endpoint endpoint = endpointMap.get(server);
+    Endpoint endpoint = endpointMap.get(memberId);
     boolean addedEndpoint = false;
     if (endpoint == null || endpoint.isClosed()) {
       synchronized (this) {
-        endpoint = endpointMap.get(server);
+        endpoint = endpointMap.get(memberId);
         if (endpoint == null || endpoint.isClosed()) {
           ConnectionStats stats = getStats(server);
-          Map<ServerLocation, Endpoint> endpointMapTemp = new HashMap<>(endpointMap);
+          Map<DistributedMember, Endpoint> endpointMapTemp = new HashMap<>(endpointMap);
           endpoint = new Endpoint(this, ds, server, stats, memberId);
           listener.clearPdxRegistry(endpoint);
-          endpointMapTemp.put(server, endpoint);
+          endpointMapTemp.put(memberId, endpoint);
           endpointMap = Collections.unmodifiableMap(endpointMapTemp);
           addedEndpoint = true;
           poolStats.setServerCount(endpointMap.size());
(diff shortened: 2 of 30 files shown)


geode-core/src/integrationTest/java/org/apache/geode/cache/client/internal/ConnectionPoolImplJUnitTest.java
```diff
@@ -168,7 +168,7 @@ public void testCacheClose() throws Exception {
     assertTrue(pool2.isDestroyed());
   }
 
-  @Test
+  @Test(timeout = 5000)
```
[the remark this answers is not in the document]
alb3rtobr: This test case enters in a infinite loop, I still have to fix it. This timeout is a temporary workaround to make it fail, so the CI can at least finish.

geode-core/src/main/java/org/apache/geode/cache/client/internal/EndpointManagerImpl.java
```diff
@@ -56,17 +56,17 @@ public EndpointManagerImpl(String poolName, DistributedSystem ds, CancelCriterio
 
   @Override
   public Endpoint referenceEndpoint(ServerLocation server, DistributedMember memberId) {
-    Endpoint endpoint = endpointMap.get(server);
+    Endpoint endpoint = endpointMap.get(memberId);
```
bschuchardt: This is what is causing the ConnectionPoolImplJUnitTest failure.  That test creates two servers and a pool in the same JVM, so both servers have the same membership ID.  If you delete that test and replace it with a distributedTest that creates the servers in different JVMs it ought to pass.

bschuchardt: Actually I think this is a problem that needs to be brought to the dev email list.  Either the PR needs to change or we need to disallow creating multiple CacheServers in the same cache.

alb3rtobr:
In the first PR I sent for review some time ago, it was using a new class combining ServerLocation and member id, instead of just member id. After your comment, I think I have to go back to that version then.

alb3rtobr:
> In the first PR I sent for review some time ago, it was using a new class combining ServerLocation and member id, instead of just member id. After your comment, I think I have to go back to that version then.

I forgot to add the link to that PR: https://github.com/apache/geode/pull/4627

geode-core/src/main/java/org/apache/geode/distributed/internal/LocatorLoadSnapshot.java
```diff
@@ -85,49 +86,52 @@ public LocatorLoadSnapshot() {
   }
 
   public void addServer(ServerLocation location, String[] groups, ServerLoad initialLoad) {
-    addServer(location, groups, initialLoad, 30000);
+    addServer(location, "", groups, initialLoad, 30000);
```
upthewaterspout, line 89: It looks like this is only used by tests, is that right? It might be better to remove this method, which uses and incorrect memberId of ""

geode-core/src/main/java/org/apache/geode/distributed/internal/LocatorLoadSnapshot.java
```diff
@@ -448,6 +489,65 @@ private void updateMap(Map map, ServerLocation location, float load, float loadP
...
+  /**
+   *
+   * @param groupServers the servers to consider
+   * @param excludedServers servers to exclude
+   * @param count how many you want. a negative number means all of them in order of best to worst
+   * @return a list of best...worst server LoadHolders
+   */
+  private List<LoadHolder> findBestServersUsingMemberId(
+      Map<ServerLocationAndMemberId, LoadHolder> groupServers,
+      Set<ServerLocation> excludedServers, int count) {
```
upthewaterspout, line 514: I'm wondering about how this method is handling excludedServers. If the hostname-for-senders of all of the servers is the same, doesn't that mean that as soon as one server is in this excluded server list, no results will be returned?

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/command/Ping.java
```diff
@@ -60,6 +78,29 @@ public void cmdExecute(final Message clientMessage, final ServerConnection serve
...
+    }
+    if (!serverConnection.getCache().getDistributionManager().isCurrentMember(targetServer)) {
+      logger.warn("Unable to ping non-member {} for client {}", targetServer,
+          serverConnection.getProxyID());
+      writeErrorResponse(clientMessage, MessageType.PING, serverConnection);
+      serverConnection.setAsTrue(RESPONDED);
+    } else {
+      // send a ping message to the server. This is a one-way message that doesn't send a reply
+      final DistributedPingMessage distributedPingMessage =
+          new DistributedPingMessage(targetServer, serverConnection.getProxyID());
```
upthewaterspout, line 98: Does this need to take into account the version of the member that is being pinged? Could we be pinging a member with an older version and it will fail to deserialize this message? Or are we assuming all of the receivers will be upgraded before we start getting pings that need to be forwarded to the correct server.

geode-core/src/main/java/org/apache/geode/distributed/internal/LocatorLoadSnapshot.java
alb3rtobr: You are right, the excluded servers should have the member id into account

alb3rtobr: Ok, the method is not involved in the receiver failover, I was wrong. So, you are right about the behavior of the method. And in the case of configuring same host and port for all the receivers it doesn't seem to be an issue.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/command/Ping.java
alb3rtobr: In our case its ok if the ping is not working while the upgrade of all members is not finished

geode-core/src/main/java/org/apache/geode/distributed/internal/LocatorLoadSnapshot.java
alb3rtobr: I have just added a commit to fix this.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/command/Ping.java
bschuchardt: I don't think Ping needs to check the version of the recipient of the DistributedPingMessage.  Only  new clients will include the new targetServer message part, and all servers should be upgraded before new clients come on-line.

geode-core/src/main/java/org/apache/geode/distributed/internal/LocatorLoadSnapshot.java
upthewaterspout: I'm digging through the code a bit more. I think this method does get called, but maybe we are retrying at a higher level so it doesn't matter for the gateway code.

upthewaterspout: I think it's possible to call this method, but it does look like the WAN will retry and come back with an empty excluded servers set. So I think you're right, it will be ok for the WAN case.

PR #4830, opened 2020-03-19
GEODE-7852: Adding a SocketFactory configuration for client Pools

upthewaterspout:
Adding the ability to provide a SocketFactory to a client pool for use in
creating sockets. Adding an implementation of this SocketFactory that
configures the pool to use an SNI proxy.

Adding an automated test of the SniSocketFactory that uses docker compose to launch the locator and server in docker containers, and uses a real SNI proxy to ensure that clients connect to those locators and servers through the proxy.


geode-assembly/src/acceptanceTest/resources/org/apache/geode/client/docker-compose.yml
```diff
@@ -0,0 +1,48 @@
...
+#
+version: '3'
+services:
+  geode:
+    container_name: 'geode'
+    image: 'geode:develop'
+    expose:
+      - '10334'
+      - '40404'
+#   entrypoint: gfsh
```
pivotal-jbarrett: Delete commented code.

geode-core/src/distributedTest/java/org/apache/geode/cache/client/SocketFactoryDUnitTest.java
```diff
@@ -0,0 +1,128 @@
...
+  private void verifyConnection(ClientCache client) {
+    // Verify connectivity with servers
+    Object functionResult =
+        FunctionService.onServers(client).execute(new TestFunction()).getResult();
+
+    assertThat(functionResult).isEqualTo(Arrays.asList("test"));
+  }
+
+
+  public static class TestFunction implements Function, DataSerializable {
```
pivotal-jbarrett: Please don't leave raw type warnings. Use proper generic types or suppress warnings if generics can't work here. 

```java
public static class TestFunction implements Function<Void> ... {
...
public void execute<FunctionContext<Void> context);
}
```
```

File: geode-core/src/distributedTest/java/org/apache/geode/cache/client/SocketFactoryDUnitTest.java [hunk 3]
```diff
@@ -0,0 +1,128 @@
...
+
+
+    // Verify the socket factory switched the hostname so we can connect
+    verifyConnection(client);
+  }
+
+  private void verifyConnection(ClientCache client) {
+    // Verify connectivity with servers
+    Object functionResult =
+        FunctionService.onServers(client).execute(new TestFunction()).getResult();
```
pivotal-jbarrett: Cleanup rawtype warnings.

upthewaterspout: Done.

PR #4833, opened 2020-03-20
Feature/geode 6536 2: Added retry in borrowConnection/single hop

mivanac:

### For all changes:

### Note:


geode-core/src/main/java/org/apache/geode/cache/client/internal/pooling/ConnectionManagerImpl.java
```diff
@@ -338,6 +335,10 @@ public PooledConnection borrowConnection(ServerLocation server, long acquireTime
               }
             }
           }
+          if (null != connection) {
+            return connection;
+          }
+          throw new ServerConnectivityException(BORROW_CONN_ERROR_MSG + server);
         }
 
         if (!onlyUseExistingCnx) {
```
agingade: Sorry, if i missed this in earlier comments. The "onlyUseExistingCnx" indicates no new connection to be created, by the time this gets checked, a create connection attempt will be made if "tryCreate" is true (above logic). 
It seems right to keep the condition check as If(onlyUseExistingCnx) { throw new AllConnectionsInUseException();} if there are no connection to this server available, the exception may be handled differently based on the caller.

geode-core/src/main/java/org/apache/geode/cache/client/internal/pooling/ConnectionManagerImpl.java
```diff
@@ -338,6 +335,10 @@ public PooledConnection borrowConnection(ServerLocation server, long acquireTime
               }
             }
           }
+          if (null != connection) {
+            return connection;
+          }
+          throw new ServerConnectivityException(BORROW_CONN_ERROR_MSG + server);
```
agingade: If the tryCreate() is true, and for any reason, unable to create connection, it should be in the loop till the specified timeout (server-timeout), right? throwing an exception here will return before the specified wait time.

geode-core/src/main/java/org/apache/geode/cache/client/internal/pooling/ConnectionManagerImpl.java
mivanac: I introduced logic simular to borrowConnection, because if connectins in pool are not up to maximum, why we could not create new connection. And if "onlyUseExistingCnx" is set to false, we will create new connection, even if we exceed maximum.
But, I can return to previous implementation.

geode-core/src/main/java/org/apache/geode/cache/client/internal/pooling/ConnectionManagerImpl.java
mivanac: It is simular to borrowConnection() implementation, so we are reatempting, until we can try to create connection (until pool size drops bellow maximum). If server is unreachable, we will throw exception.

geode-core/src/main/java/org/apache/geode/cache/client/internal/pooling/ConnectionManagerImpl.java
agingade: Agree with your thinking in relation to max-connection; but this arguments seems to enforce that only existing connection needs to be used.

geode-core/src/main/java/org/apache/geode/cache/client/internal/pooling/ConnectionManagerImpl.java
agingade: OK, i get your point. I was thinking of the case, where server is busy when this connection is attempted. Also, do we need to handle GemFireSecurityException?

mivanac: As I said, logic is the same as for borrowConnection(), so I would not change this logic.

PR #4836, opened 2020-03-20
GEODE-7812: Cleanup for PutAllOp

kirklund:
Cleanup some collection types and interfaces for PutAllOp and
PutAllClientServerDistributedTest.

This does NOT fix GEODE-7812. I'm submitting this as its own PR because I want it to be a separate commit.

### DIFF
--- a/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/SingleHopStatsDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/SingleHopStatsDUnitTest.java
@@ -22,13 +22,18 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
 
+import java.io.DataInput;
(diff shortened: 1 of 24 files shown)


geode-core/src/distributedTest/java/org/apache/geode/internal/cache/SingleHopStatsDUnitTest.java
```diff
@@ -331,12 +337,12 @@ private void createdColocatedPRData(GemFireCacheImpl cache) {
 
       customerRegion.put(custid, customer);
       for (int j = 1; j <= 10; j++) {
-        int oid = (i * 10) + j;
+        int oid = i * 10 + j;
```
igodwin: You're going to make me go look up the Java operator precedence to make sure this is doing what I really think it is?

kirklund: I just cheat and let the IDE tell me the precedence! Multiplication happens before addition.

PR #4837, opened 2020-03-21
Ignore ClientSNIAcceptanceTest on windows

upthewaterspout:
Our windows CI machines don't have docker-compose, so don't run this test.


upthewaterspout:
Note - it appears on Windows we only run geode-assembly:acceptanceTest, rather than all acceptance tests. @jdeppe-pivotal Is this intentional? 

We reused this docker compose rule because the connector tests use it, but we added to add our test to geode-assembly which is why it actually got run on windows, unlike the rest of the acceptanceTests.

PR #4841, opened 2020-03-23
GEODE-7901: Redis testGetSet_shouldBeAtomic and testGetRange_rangePastEndOfValue_returnsEmptyString tests failing

ringles:
One test for atomicity had too small a timeout to run reliably in CI. This has been increased to a safe value. (The timeout sometimes caused the subsequent test to fail as well.)

### DIFF
--- a/geode-redis/src/integrationTest/java/org/apache/geode/redis/StringsIntegrationTest.java
+++ b/geode-redis/src/integrationTest/java/org/apache/geode/redis/StringsIntegrationTest.java
@@ -33,7 +33,6 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
-import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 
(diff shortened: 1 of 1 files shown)


geode-redis/src/integrationTest/java/org/apache/geode/redis/StringsIntegrationTest.java
```diff
@@ -281,7 +281,7 @@ public void testGetSet_shouldBeAtomic()
 
     latch.countDown();
 
-    Integer getSetSum = future2.get(5, TimeUnit.SECONDS);
+    Integer getSetSum = future2.get(60, TimeUnit.SECONDS);
```
Bill: Rather than hard-coding a wait limit, most Geode tests use `GeodeAwaitility`. The advantage of the latter is that we can (and have) set a global default maximum timeout (currently 5 minutes).

ringles: Aha! The part of the code we're working on is old enough that I don't think we've seen an example of that. Thanks for the tip - we'll adopt that going forward.

PR #4843, opened 2020-03-23
GEODE-7900: Only set the SNI hostname if it is not present

upthewaterspout:
It turns out the JDK sometimes automatically sets the SNI hostname and
sometimes does not.  With openjdk 8, it only sets SNI hostname if there is a
dot in the hostname. See sun.security.ssl.Utilities#addToSNIServerNameList

For this reason, we need to only set the SNI hostname if the JDK did not.
Otherwise it will complain that there is a `Duplicated server name of type 0`

Note, this fixes the Windows distributed tests failures we are seeing - for example this run https://concourse.apachegeode-ci.info/teams/main/pipelines/apache-develop-main/jobs/WindowsGfshDistributedTestOpenJDK8/builds/1391. Those failures aren't really because of windows, its because our windows CI instances have fully qualified hostnames as opposed to the linux docker containers.

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/net/SocketCreator.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/net/SocketCreator.java
@@ -56,6 +56,7 @@ import javax.net.ssl.SSLParameters;
 import javax.net.ssl.SSLPeerUnverifiedException;
 import javax.net.ssl.SSLProtocolException;
 import javax.net.ssl.SSLSocket;
+import javax.net.ssl.StandardConstants;
 import javax.net.ssl.TrustManager;
 import javax.net.ssl.TrustManagerFactory;
 import javax.net.ssl.X509ExtendedKeyManager;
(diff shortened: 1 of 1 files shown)


geode-core/src/main/java/org/apache/geode/internal/net/SocketCreator.java
```diff
@@ -774,6 +775,14 @@ private void setServerNames(SSLParameters modifiedParams, HostAndPort addr) {
...
     oldNames = oldNames == null ? Collections.emptyList() : oldNames;
     final List<SNIServerName> serverNames = new ArrayList<>(oldNames);
+
+    if (serverNames.stream()
+        .mapToInt(SNIServerName::getType)
+        .anyMatch(type -> type == StandardConstants.SNI_HOST_NAME)) {
+      // we already have a SNI hostname set. Do nothing.
+      return;
+    }
+
```
Bill: Two things:

1. a nit: the allocation can be moved after the conditional :)
2. is the type check sufficient or is a content check necessary?

on (2) I wonder if there is a scenario where SNI might be set already, but maybe it _disagrees_ with `addr.getHostName()`

upthewaterspout: For (2) - it certainly could disagree since the JDK is setting it to whatever it wants ... But the JDK also only allows one SNI name of a given type. So I guess we just need to decide what the behavior is that we want. I thought it would best to leave the JDK setting in place.

Bill: Maybe we could unconditionally delete whatever is there and replace it with our own content?

Bill: Having looked at the path into this method, it's clear that this framework owns this socket and the only way it can set set will cause it to have the value we expect so I'm ok w/ this code as-is. Thanks for playing!

PR #4845, opened 2020-03-24
GEODE-7902: fix flaky test

jinmeiliao:

### For all changes:

### Note:

### DIFF
--- a/geode-assembly/src/distributedTest/java/org/apache/geode/management/internal/rest/DeploymentManagementRedployDUnitTest.java
+++ b/geode-assembly/src/distributedTest/java/org/apache/geode/management/internal/rest/DeploymentManagementRedployDUnitTest.java
@@ -20,13 +20,16 @@ import static org.apache.geode.test.junit.assertions.ClusterManagementRealizatio
 import static org.assertj.core.api.Assertions.assertThat;
 
 import java.io.File;
+import java.io.IOException;
 import java.io.Serializable;
 import java.net.URL;
+import java.nio.file.Path;
 import java.util.List;
 import java.util.Properties;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 
@@ -124,16 +127,16 @@ public class DeploymentManagementRedployDUnitTest {
   }
 
   @Test
-  public void redeployJarsWithNewVersionsOfFunctionsAndMultipleLocators()
-      throws InterruptedException {
+  public void redeployJarsWithNewVersionsOfFunctionsAndMultipleLocators() throws IOException {
     Properties props = new Properties();
     props.setProperty("locators", "localhost[" + locator.getPort() + "]");
-    lsRule.startLocatorVM(2, props);
+    MemberVM locator2 = lsRule.startLocatorVM(2, props);
 
     deployment.setFile(jarAVersion1);
     assertManagementResult(client.create(deployment)).isSuccessful();
     server.invoke(() -> assertThatCanLoad(JAR_NAME_A, FUNCTION_A));
     server.invoke(() -> assertThatFunctionHasVersion(FUNCTION_A, VERSION1));
+
 
     deployment.setFile(jarAVersion2);
     assertManagementResult(client.create(deployment)).isSuccessful();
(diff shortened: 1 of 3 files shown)


jinmeiliao:
> I don't understand this fix.
> 
> 1. Why do we have to sleep between stopping and restarting the server?
> 2. What does it mean for the new jars to "settle" on the file server?
> 3. If the version was correct before the server stopped, how did it revert to an earlier version after it was restarted?
> 
> Also, I'm hoping for a fix that does not involve a fixed wait. If we didn't do that fixed wait at this spot, would the server eventually pick up the correct version of the jar? If not, why not?

To be honest, I don't quite understand either, but after adding the wait, the test ran more than 100 times without failure, before that, it failed at the 2nd time.

geode-assembly/src/distributedTest/java/org/apache/geode/management/internal/rest/DeploymentManagementRedployDUnitTest.java
```diff
@@ -134,14 +135,16 @@ public void redeployJarsWithNewVersionsOfFunctionsAndMultipleLocators() {
...
 
-
     deployment.setFile(jarAVersion2);
     assertManagementResult(client.create(deployment)).isSuccessful();
     server.invoke(() -> assertThatCanLoad(JAR_NAME_A, FUNCTION_A));
     server.invoke(() -> assertThatFunctionHasVersion(FUNCTION_A, VERSION2));
 
     server.stop(false);
 
+    // GEODE-7902: wait for a while for new jars to settle in the file system
```
dschneider-pivotal: What is the sleep waiting for? What does "settle" mean? Seems like whatever it means couldn't we then do a file system await to figure out if the file has settled?

demery-pivotal: Would an `awaitUntilAsserted()` around the final assertion (or the final two) suffice, or is it necessary to wait (for whatever the condition is) before doing the restart?

If `awaitUntilAsserted()` works, that's a much more robust fix than a fixed-duration sleep.

jinmeiliao: I believe the wait has to be before the server restarts to get the new version of jars from one of the locators.

geode-core/src/main/java/org/apache/geode/management/internal/configuration/functions/DownloadJarFunction.java
```diff
@@ -61,6 +61,13 @@ public void execute(FunctionContext<Object[]> context) {
         RemoteInputStreamServer istream = null;
         try {
           File jarFile = sharedConfig.getPathToJarOnThisLocator(group, jarName).toFile();
+          // if file is modified within the last second, wait for a second before downloading it
+          // It looks like if the file was modified within the same second, the file downloaded
+          // will be the old content. On mac/linux, the lastModified time of the file is already
+          // reported to the precision of seconds.
+          if (jarFile.lastModified() > System.currentTimeMillis() - 1000) {
```
pivotal-jbarrett: This should be replaced with an await so we don't get stuck here forever.

jinmeiliao: I tried that but await isn't available in core.

kirklund:
Outside of test src sets, you can code your own await. Take a look at geode-core/src/main/java/org/apache/geode/distributed/internal/locks/DLockService.java's sleep method.

Maybe change it to just throw InterruptedException and add a condition (for example, if want to wait until the pidFile disappears from the JVM's working directory):
```
  private void awaitPidFileDeletion(long millis) throws InterruptedException {
    StopWatch timer = new StopWatch(true);
    while (pidFile.exists()) {
      long timeLeft = millis - timer.elapsedTimeMillis();
      if (timeLeft <= 0) {
        break;
      }
      Thread.sleep(timeLeft);
      break;
    }
  }
```

jinmeiliao:
> Outside of test src sets, you can code your own await. Take a look at geode-core/src/main/java/org/apache/geode/distributed/internal/locks/DLockService.java's sleep method.
> 
> Maybe change it to just throw InterruptedException and add a condition (for example, if want to wait until the pidFile disappears from the JVM's working directory):
> 
> ```
>   private void awaitPidFileDeletion(long millis) throws InterruptedException {
>     StopWatch timer = new StopWatch(true);
>     while (pidFile.exists()) {
>       long timeLeft = millis - timer.elapsedTimeMillis();
>       if (timeLeft <= 0) {
>         break;
>       }
>       Thread.sleep(timeLeft);
>       break;
>     }
>   }
> ```

Thanks! Decided to put the wait in the test code. But this is helpful.

PR #4846, opened 2020-03-24
GEODE-7908: Throws IllegalArgumentException if client config uses SNI…

echobravopapa:
… without SSL enabled.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/cache/client/proxy/ProxySocketFactories.java
+++ b/geode-core/src/main/java/org/apache/geode/cache/client/proxy/ProxySocketFactories.java
@@ -14,7 +14,13 @@
  */
 package org.apache.geode.cache.client.proxy;
 
+import static org.apache.geode.distributed.ConfigurationProperties.SSL_ENABLED_COMPONENTS;
+
+import java.util.Properties;
+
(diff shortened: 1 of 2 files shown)


geode-core/src/main/java/org/apache/geode/cache/client/proxy/ProxySocketFactories.java
```diff
@@ -14,19 +14,29 @@
...
   /**
    * Create a {@link SocketFactory} that will connect a geode client through
    * the configured SNI proxy.
    *
-   *
    * @param hostname the hostname of the sni proxy
    * @param port the port of the sni proxy
    */
-  public static SocketFactory sni(String hostname, int port) {
+  public static SocketFactory sni(String hostname, int port, Properties geodeProperties) {
```
pivotal-jbarrett: I am really not keen on using `Properties` configure proxies. How does the user know which arbitrary `Properties` object to pass? A strongly typed value expresses the correct value without support documentation. Also are we saying that the user is responsible for passing the configuration rather than the system passing it on the `createSocket` method or some lifecycle method?

(Javadocs are missing)

PR #4848, opened 2020-03-25
GEODE-7670: Add Tests for PR clear

jujoramos:
Added distributed tests to verify that the clear operation on
Partitioned Regions works as expected when there are other
concurrent operations happening on the cache (puts, gets, members
added and members removed).

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithConcurrentOperationsDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithConcurrentOperationsDUnitTest.java
@@ -28,6 +28,11 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.stream.IntStream;
 
 import junitparams.JUnitParamsRunner;
@@ -56,6 +61,7 @@ import org.apache.geode.distributed.internal.membership.api.MembershipManagerHel
 import org.apache.geode.internal.cache.versions.RegionVersionHolder;
 import org.apache.geode.internal.cache.versions.RegionVersionVector;
 import org.apache.geode.internal.cache.versions.VersionSource;
+import org.apache.geode.test.awaitility.GeodeAwaitility;
 import org.apache.geode.test.dunit.AsyncInvocation;
 import org.apache.geode.test.dunit.VM;
 import org.apache.geode.test.dunit.rules.CacheRule;
@@ -79,7 +85,9 @@ public class PartitionedRegionClearWithConcurrentOperationsDUnitTest implements 
   @Rule
   public CacheRule cacheRule = CacheRule.builder().createCacheInAll().build();
 
-  private VM accessor, server1, server2;
+  private VM server1;
+  private VM server2;
+  private VM accessor;
 
   private enum TestVM {
     ACCESSOR(0), SERVER1(1), SERVER2(2);
@@ -91,7 +99,6 @@ public class PartitionedRegionClearWithConcurrentOperationsDUnitTest implements 
     }
   }
 
-  @SuppressWarnings("unused")
   static RegionShortcut[] regionTypes() {
     return new RegionShortcut[] {
         RegionShortcut.PARTITION, RegionShortcut.PARTITION_REDUNDANT
@@ -126,26 +133,24 @@ public class PartitionedRegionClearWithConcurrentOperationsDUnitTest implements 
   }
 
   private void initAccessor(RegionShortcut regionShortcut) {
-    @SuppressWarnings("rawtypes")
-    PartitionAttributes attributes = new PartitionAttributesFactory<String, String>()
+    PartitionAttributes<String, String> attrs = new PartitionAttributesFactory<String, String>()
         .setTotalNumBuckets(BUCKETS)
         .setLocalMaxMemory(0)
         .create();
 
     cacheRule.getCache().createRegionFactory(regionShortcut)
-        .setPartitionAttributes(attributes)
+        .setPartitionAttributes(attrs)
         .create(REGION_NAME);
 
   }
 
   private void initDataStore(RegionShortcut regionShortcut) {
-    @SuppressWarnings("rawtypes")
-    PartitionAttributes attributes = new PartitionAttributesFactory<String, String>()
+    PartitionAttributes<String, String> attrs = new PartitionAttributesFactory<String, String>()
         .setTotalNumBuckets(BUCKETS)
         .create();
 
     cacheRule.getCache().createRegionFactory(regionShortcut)
-        .setPartitionAttributes(attributes)
+        .setPartitionAttributes(attrs)
         .create(REGION_NAME);
   }
 
@@ -195,11 +200,10 @@ public class PartitionedRegionClearWithConcurrentOperationsDUnitTest implements 
    * @param bucketDump1 First bucketDump.
    * @param bucketDump2 Second bucketDump.
    */
-  @SuppressWarnings({"unchecked", "rawtypes"})
   private void assertRegionVersionVectorsConsistency(int bucketId, BucketDump bucketDump1,
       BucketDump bucketDump2) {
-    RegionVersionVector rvv1 = bucketDump1.getRvv();
-    RegionVersionVector rvv2 = bucketDump2.getRvv();
+    RegionVersionVector<?> rvv1 = bucketDump1.getRvv();
+    RegionVersionVector<?> rvv2 = bucketDump2.getRvv();
 
     if (rvv1 == null) {
       assertThat(rvv2)
@@ -217,14 +221,14 @@ public class PartitionedRegionClearWithConcurrentOperationsDUnitTest implements 
 
     assertThat(rvv1).isNotNull();
     assertThat(rvv2).isNotNull();
-    Map<VersionSource, RegionVersionHolder> rvv2Members =
-        new HashMap<VersionSource, RegionVersionHolder>(rvv1.getMemberToVersion());
-    Map<VersionSource, RegionVersionHolder> rvv1Members =
-        new HashMap<VersionSource, RegionVersionHolder>(rvv1.getMemberToVersion());
-    for (Map.Entry<VersionSource, RegionVersionHolder> entry : rvv1Members.entrySet()) {
-      VersionSource memberId = entry.getKey();
-      RegionVersionHolder versionHolder1 = entry.getValue();
-      RegionVersionHolder versionHolder2 = rvv2Members.remove(memberId);
+    Map<VersionSource<?>, RegionVersionHolder<?>> rvv2Members =
+        new HashMap<>(rvv1.getMemberToVersion());
+    Map<VersionSource<?>, RegionVersionHolder<?>> rvv1Members =
+        new HashMap<>(rvv1.getMemberToVersion());
+    for (Map.Entry<VersionSource<?>, RegionVersionHolder<?>> entry : rvv1Members.entrySet()) {
+      VersionSource<?> memberId = entry.getKey();
+      RegionVersionHolder<?> versionHolder1 = entry.getValue();
+      RegionVersionHolder<?> versionHolder2 = rvv2Members.remove(memberId);
       assertThat(versionHolder1)
           .as("RegionVersionVector for bucket " + bucketId + " on member " + bucketDump1.getMember()
               + " is not consistent with member " + bucketDump2.getMember())
@@ -272,12 +276,12 @@ public class PartitionedRegionClearWithConcurrentOperationsDUnitTest implements 
   }
 
   /**
-   * Continuously execute get operations on the PartitionedRegion for the given durationInSeconds.
+   * Continuously execute get operations on the PartitionedRegion for the given durationInMillis.
    */
-  private void executeGets(final int numEntries, final long durationInSeconds) {
+  private void executeGets(final int numEntries, final long durationInMillis) {
     Cache cache = cacheRule.getCache();
     Region<String, String> region = cache.getRegion(REGION_NAME);
-    Instant finishTime = Instant.now().plusSeconds(durationInSeconds);
+    Instant finishTime = Instant.now().plusMillis(durationInMillis);
 
     while (Instant.now().isBefore(finishTime)) {
       // Region might have been cleared in between, that's why we check for null.
@@ -289,12 +293,12 @@ public class PartitionedRegionClearWithConcurrentOperationsDUnitTest implements 
   }
 
   /**
-   * Continuously execute put operations on the PartitionedRegion for the given durationInSeconds.
+   * Continuously execute put operations on the PartitionedRegion for the given durationInMillis.
    */
-  private void executePuts(final int numEntries, final long durationInSeconds) {
+  private void executePuts(final int numEntries, final long durationInMillis) {
     Cache cache = cacheRule.getCache();
     Region<String, String> region = cache.getRegion(REGION_NAME);
-    Instant finishTime = Instant.now().plusSeconds(durationInSeconds);
+    Instant finishTime = Instant.now().plusMillis(durationInMillis);
 
     while (Instant.now().isBefore(finishTime)) {
       IntStream.range(0, numEntries).forEach(i -> region.put(String.valueOf(i), "Value_" + i));
@@ -303,15 +307,15 @@ public class PartitionedRegionClearWithConcurrentOperationsDUnitTest implements 
 
   /**
    * Continuously execute putAll operations on the PartitionedRegion for the given
-   * durationInSeconds.
+   * durationInMillis.
    */
-  private void executePutAlls(final int start, final int finish, final long durationInSeconds) {
+  private void executePutAlls(final int startKey, final int finalKey, final long durationInMillis) {
     Cache cache = cacheRule.getCache();
     Map<String, String> valuesToInsert = new HashMap<>();
     Region<String, String> region = cache.getRegion(REGION_NAME);
-    IntStream.range(start, finish)
+    IntStream.range(startKey, finalKey)
         .forEach(i -> valuesToInsert.put(String.valueOf(i), "Value_" + i));
-    Instant finishTime = Instant.now().plusSeconds(durationInSeconds);
+    Instant finishTime = Instant.now().plusMillis(durationInMillis);
 
     while (Instant.now().isBefore(finishTime)) {
       region.putAll(valuesToInsert);
@@ -320,12 +324,12 @@ public class PartitionedRegionClearWithConcurrentOperationsDUnitTest implements 
 
   /**
    * Continuously execute remove operations on the PartitionedRegion for the given
-   * durationInSeconds.
+   * durationInMillis.
    */
-  private void executeRemoves(final int numEntries, final long durationInSeconds) {
+  private void executeRemoves(final int numEntries, final long durationInMillis) {
     Cache cache = cacheRule.getCache();
     Region<String, String> region = cache.getRegion(REGION_NAME);
-    Instant finishTime = Instant.now().plusSeconds(durationInSeconds);
+    Instant finishTime = Instant.now().plusMillis(durationInMillis);
 
     while (Instant.now().isBefore(finishTime)) {
       // Region might have been cleared in between, that's why we check for null.
@@ -338,14 +342,15 @@ public class PartitionedRegionClearWithConcurrentOperationsDUnitTest implements 
 
   /**
    * Continuously execute removeAll operations on the PartitionedRegion for the given
-   * durationInSeconds.
+   * durationInMillis.
    */
-  private void executeRemoveAlls(final int start, final int finish, final long durationInSeconds) {
+  private void executeRemoveAlls(final int startKey, final int finalKey,
+      final long durationInMillis) {
     Cache cache = cacheRule.getCache();
     List<String> keysToRemove = new ArrayList<>();
     Region<String, String> region = cache.getRegion(REGION_NAME);
-    IntStream.range(start, finish).forEach(i -> keysToRemove.add(String.valueOf(i)));
-    Instant finishTime = Instant.now().plusSeconds(durationInSeconds);
+    IntStream.range(startKey, finalKey).forEach(i -> keysToRemove.add(String.valueOf(i)));
+    Instant finishTime = Instant.now().plusMillis(durationInMillis);
 
     while (Instant.now().isBefore(finishTime)) {
       region.removeAll(keysToRemove);
@@ -373,18 +378,27 @@ public class PartitionedRegionClearWithConcurrentOperationsDUnitTest implements 
   }
 
   /**
-   * Continuously execute clear operations on the PartitionedRegion for the given durationInSeconds.
+   * Continuously execute clear operations on the PartitionedRegion every periodInMillis for the
+   * given
+   * durationInMillis.
    */
-  private void executeClears(final long durationInSeconds, final long waitTimeInMilliseconds)
+  private void executeClears(final long durationInMillis, final long periodInMillis)
       throws InterruptedException {
     Cache cache = cacheRule.getCache();
+    AtomicLong invocationCount = new AtomicLong(0);
     Region<String, String> region = cache.getRegion(REGION_NAME);
-    Instant finishTime = Instant.now().plusSeconds(durationInSeconds);
-
-    while (Instant.now().isBefore(finishTime)) {
+    Long minimumInvocationCount = (durationInMillis / periodInMillis);
+    ScheduledExecutorService executor = Executors.newScheduledThreadPool(1);
+    ScheduledFuture<?> scheduledFuture = executor.scheduleWithFixedDelay(() -> {
       region.clear();
-      Thread.sleep(waitTimeInMilliseconds);
-    }
+      invocationCount.incrementAndGet();
+    }, 0, periodInMillis, TimeUnit.MILLISECONDS);
+
+    await().untilAsserted(
+        () -> assertThat(invocationCount.get()).isGreaterThanOrEqualTo(minimumInvocationCount));
+    scheduledFuture.cancel(false);
+    executor.shutdown();
+    executor.awaitTermination(GeodeAwaitility.getTimeout().getSeconds(), TimeUnit.SECONDS);
   }
 
   /**
@@ -400,17 +414,17 @@ public class PartitionedRegionClearWithConcurrentOperationsDUnitTest implements 
   public void clearWithConcurrentPutGetRemoveShouldWorkCorrectly(TestVM coordinatorVM,
       RegionShortcut regionShortcut) throws InterruptedException {
     final int entries = 15000;
-    final int workSeconds = 60;
+    final int workMillis = 60000;
     parametrizedSetup(regionShortcut);
 
     // Let all VMs continuously execute puts and gets for 60 seconds.
     List<AsyncInvocation<Void>> asyncInvocationList = Arrays.asList(
-        server1.invokeAsync(() -> executePuts(entries, workSeconds)),
-        server2.invokeAsync(() -> executeGets(entries, workSeconds)),
-        accessor.invokeAsync(() -> executeRemoves(entries, workSeconds)));
+        server1.invokeAsync(() -> executePuts(entries, workMillis)),
+        server2.invokeAsync(() -> executeGets(entries, workMillis)),
+        accessor.invokeAsync(() -> executeRemoves(entries, workMillis)));
 
     // Clear the region every second for 60 seconds.
-    getVM(coordinatorVM.vmNumber).invoke(() -> executeClears(workSeconds, 1000));
+    getVM(coordinatorVM.vmNumber).invoke(() -> executeClears(workMillis, 1000));
 
     // Let asyncInvocations finish.
     for (AsyncInvocation<Void> asyncInvocation : asyncInvocationList) {
@@ -434,20 +448,20 @@ public class PartitionedRegionClearWithConcurrentOperationsDUnitTest implements 
   @Parameters(method = "coordinatorsAndRegionTypes")
   public void clearWithConcurrentPutAllRemoveAllShouldWorkCorrectly(TestVM coordinatorVM,
       RegionShortcut regionShortcut) throws InterruptedException {
-    final int workSeconds = 15;
+    final int workMillis = 15000;
     parametrizedSetup(regionShortcut);
 
     // Let all VMs continuously execute putAll and removeAll for 15 seconds.
     List<AsyncInvocation<Void>> asyncInvocationList = Arrays.asList(
-        server1.invokeAsync(() -> executePutAlls(0, 2000, workSeconds)),
-        server1.invokeAsync(() -> executeRemoveAlls(0, 2000, workSeconds)),
-        server2.invokeAsync(() -> executePutAlls(2000, 4000, workSeconds)),
-        server2.invokeAsync(() -> executeRemoveAlls(2000, 4000, workSeconds)),
-        accessor.invokeAsync(() -> executePutAlls(4000, 6000, workSeconds)),
-        accessor.invokeAsync(() -> executeRemoveAlls(4000, 6000, workSeconds)));
+        server1.invokeAsync(() -> executePutAlls(0, 2000, workMillis)),
+        server1.invokeAsync(() -> executeRemoveAlls(0, 2000, workMillis)),
+        server2.invokeAsync(() -> executePutAlls(2000, 4000, workMillis)),
+        server2.invokeAsync(() -> executeRemoveAlls(2000, 4000, workMillis)),
+        accessor.invokeAsync(() -> executePutAlls(4000, 6000, workMillis)),
+        accessor.invokeAsync(() -> executeRemoveAlls(4000, 6000, workMillis)));
 
     // Clear the region every half second for 15 seconds.
-    getVM(coordinatorVM.vmNumber).invoke(() -> executeClears(workSeconds, 500));
+    getVM(coordinatorVM.vmNumber).invoke(() -> executeClears(workMillis, 500));
 
     // Let asyncInvocations finish.
     for (AsyncInvocation<Void> asyncInvocation : asyncInvocationList) {
@@ -516,17 +530,17 @@ public class PartitionedRegionClearWithConcurrentOperationsDUnitTest implements 
   public void clearOnRedundantPartitionRegionWithConcurrentPutGetRemoveShouldWorkCorrectlyWhenNonCoordinatorMembersAreBounced(
       TestVM coordinatorVM) throws InterruptedException {
     final int entries = 7500;
-    final int workSeconds = 30;
+    final int workMillis = 30000;
     parametrizedSetup(RegionShortcut.PARTITION_REDUNDANT);
     populateRegion(accessor, entries, asList(accessor, server1, server2));
     server2.invoke(() -> DistributionMessageObserver.setInstance(new MemberKiller(false)));
 
     // Let all VMs (except the one to kill) continuously execute gets, put and removes for 30"
     List<AsyncInvocation<Void>> asyncInvocationList = Arrays.asList(
-        server1.invokeAsync(() -> executeGets(entries, workSeconds)),
-        server1.invokeAsync(() -> executePuts(entries, workSeconds)),
-        accessor.invokeAsync(() -> executeGets(entries, workSeconds)),
-        accessor.invokeAsync(() -> executeRemoves(entries, workSeconds)));
+        server1.invokeAsync(() -> executeGets(entries, workMillis)),
+        server1.invokeAsync(() -> executePuts(entries, workMillis)),
+        accessor.invokeAsync(() -> executeGets(entries, workMillis)),
+        accessor.invokeAsync(() -> executeRemoves(entries, workMillis)));
 
     // Retry the clear operation on the region until success (server2 will go down, but other
     // members will eventually become primary for those buckets previously hosted by server2).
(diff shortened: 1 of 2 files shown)


jujoramos:
The `:geode-redis:distributedTest` task has failed twice while running CI for this PR:

```
> Task :geode-redis:distributedTest

org.apache.geode.redis.RedisDistDUnitTest > classMethod FAILED
    java.lang.AssertionError: Suspicious strings were written to the log during this run.
    Fix the strings or use IgnoredException.addIgnoredException to ignore.
    -----------------------------------------------------------------------
    Found suspect string in log4j at line 5300

    [error 2020/03/26 13:30:55.981 GMT <GeodeRedisServer-WorkerThread-3> tid=98] org.apache.geode.ToDataException: toData failed on DataSerializer with id=0 for class class java.util.HashMap
```

The above can be safely ignored while reviewing the changes as this `PR` only adds new `DistributedTests` for the new `clear()` feature on Partition Regions.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithConcurrentOperationsDUnitTest.java
```diff
@@ -0,0 +1,525 @@
...
+    final int vmNumber;
+
+    TestVM(int vmNumber) {
+      this.vmNumber = vmNumber;
+    }
+  }
+
+  static RegionShortcut[] regionTypes() {
+    return new RegionShortcut[] {
+        RegionShortcut.PARTITION,
```
gesterzhou: can you add more shortcuts?

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithConcurrentOperationsDUnitTest.java
```diff
@@ -0,0 +1,525 @@
...
+  @Parameters(method = "vmAndRegionTypes")
+  public void clearWithConcurrentGetsShouldWorkCorrectly(TestVM coordinatorVM,
+      RegionShortcut regionShortcut) throws InterruptedException {
+    final int entries = 10000;
+    parametrizedSetup(regionShortcut);
+    populateRegion(accessor, entries, asList(accessor, server1, server2));
+
+    // Let all VMs continuously execute gets for 10 seconds.
+    List<AsyncInvocation<Void>> asyncInvocationList = Arrays.asList(
+        server1.invokeAsync(() -> executeGets(entries, 10)),
```
gesterzhou: after the first clear, all other gets() will not really do anything. You should re-populateRegion again before next clear.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithConcurrentOperationsDUnitTest.java
```diff
@@ -0,0 +1,525 @@
...
+   * Continuously execute clear operations on the PartitionedRegion for the given durationInSeconds.
+   */
+  private void executeClears(final long durationInSeconds, final long waitTimeInSeconds)
+      throws InterruptedException {
+    Cache cache = cacheRule.getCache();
+    Region<String, String> region = cache.getRegion(REGION_NAME);
+    Instant finishTime = Instant.now().plusSeconds(durationInSeconds);
+
+    while (Instant.now().isBefore(finishTime)) {
+      region.clear();
```
gesterzhou: you need to measure how long a clear() took. If it's more than 5 seconds, something must be wrong.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithConcurrentOperationsDUnitTest.java
```diff
@@ -0,0 +1,525 @@
...
+  /**
+   * Continuously execute put operations on the PartitionedRegion for the given durationInSeconds.
+   */
+  private void executePuts(final int numEntries, final long durationInSeconds) {
+    Cache cache = cacheRule.getCache();
+    Region<String, String> region = cache.getRegion(REGION_NAME);
+    Instant finishTime = Instant.now().plusSeconds(durationInSeconds);
+
+    while (Instant.now().isBefore(finishTime)) {
+      IntStream.range(0, numEntries).forEach(i -> region.put(String.valueOf(i), "Value_" + i));
```
gesterzhou: it's better to change the continuous put into continuous putAll.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithConcurrentOperationsDUnitTest.java
```diff
@@ -0,0 +1,525 @@
...
+    populateRegion(accessor, entries, asList(accessor, server1, server2));
+
+    // Let all VMs continuously execute gets for 10 seconds.
+    List<AsyncInvocation<Void>> asyncInvocationList = Arrays.asList(
+        server1.invokeAsync(() -> executeGets(entries, 10)),
+        server2.invokeAsync(() -> executeGets(entries, 10)),
+        accessor.invokeAsync(() -> executeGets(entries, 10)));
+
+    // Clear the region every 2 seconds for 10 seconds.
+    getVM(coordinatorVM.vmNumber).invoke(() -> executeClears(10, 2));
```
agingade: 2 secs may be long pauses...Since there is continuous cache operation, how about sleeping for few mili secs (say 50)

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithConcurrentOperationsDUnitTest.java
```diff
@@ -0,0 +1,525 @@
...
+    parametrizedSetup(regionShortcut);
+
+    // Let all VMs continuously execute puts for 15 seconds.
+    List<AsyncInvocation<Void>> asyncInvocationList = Arrays.asList(
+        server1.invokeAsync(() -> executePuts(entries, 15)),
+        server2.invokeAsync(() -> executePuts(entries, 15)),
+        accessor.invokeAsync(() -> executePuts(entries, 15)));
+
+    // Clear the region every 3 seconds for 15 seconds.
+    getVM(coordinatorVM.vmNumber).invoke(() -> executeClears(15, 3));
```
agingade: 3 secs long sleep.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithConcurrentOperationsDUnitTest.java
```diff
@@ -0,0 +1,525 @@
...
+  }
+
+  /**
+   * Waits until there are not PR messages in flight for the given region.
+   */
+  private void waitForSilence() {
+    PartitionedRegion region = (PartitionedRegion) cacheRule.getCache().getRegion(REGION_NAME);
+
+    await().untilAsserted(() -> {
+      PartitionedRegionStats prStats = region.getPrStats();
```
agingade: This waits for message pertaining to region creation or status change. There is DistributionMessageObserver that can be used to see if any messages are still in progress.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithConcurrentOperationsDUnitTest.java
```diff
@@ -0,0 +1,525 @@
...
+    Region<String, String> region = cacheRule.getCache().getRegion(REGION_NAME);
+    region.forEach(regionSnapshot::put);
+
+    return regionSnapshot;
+  }
+
+  /**
+   * Asserts that the region data is consistent on requested VMs.
+   */
+  private void assertRegionDataConsistency(List<VM> vms) {
```
agingade: This is going over the PRs entry/key set getting all the values. This will be consistent where-ever this is called. 
Are we trying to see if the data in one vm is same as in another vm. Or are we trying to see data in primary buckets are same as in secondary buckets.
Fetching the local PR data-set and iterating over it will give the data stored in that vm. 
Or just to see an op could be executed successfully after clear?

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithConcurrentOperationsDUnitTest.java
```diff
@@ -0,0 +1,525 @@
...
+    }
+
+    // Assert Region Data is consistent across members.
+    assertRegionDataConsistency(asList(accessor, server1, server2));
+  }
+
+  /**
+   * The test does the following (the clear coordinator is chosen through parameters):
+   * - Launches threads to continuously execute puts and gets for 60 seconds.
+   * - Clears the Partition Region continuously (once every 10 seconds for 60 seconds).
```
agingade: No need to specify time; if someone changes the time and doesn't update it, this will become incorrect.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithConcurrentOperationsDUnitTest.java
```diff
@@ -0,0 +1,525 @@
...
+  }
+
+  /**
+   * Shutdowns a member while the clear operation is in progress.
+   * The writer is only installed on the member the test wants to shutdown, doesn't matter whether
+   * it's the clear coordinator or another member holding primary buckets.
+   *
+   * TODO: Review once GEODE-7678 / GEODE-7912 are merged (CacheWriter lifecycle might change).
+   */
+  public static class MemberKiller extends CacheWriterAdapter<String, String> {
```
agingade: Instead of CacheWriter, you can use DistributionMessageObserver. And can use VM.bounce* to kill vms. No need to wait for CacheWriter messaging implementation.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithConcurrentOperationsDUnitTest.java
```diff
@@ -0,0 +1,525 @@
...
+import org.apache.geode.test.dunit.rules.CacheRule;
+import org.apache.geode.test.dunit.rules.DistributedRule;
+
+/**
+ * Tests to verify that {@link PartitionedRegion#clear()} can be executed multiple times on
+ * the same region while other cache operations are being executed concurrently and members are
+ * added or removed.
+ */
+@RunWith(JUnitParamsRunner.class)
+public class PartitionedRegionClearWithConcurrentOperationsDUnitTest implements Serializable {
```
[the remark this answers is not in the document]
agingade: This is a good start; other cache ops that can be added in mix are Region ops (destroyRegion, invalidateRegion), querying; Region with index, rebalance, etc.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithConcurrentOperationsDUnitTest.java
jujoramos: There are other tickets aimed at testing concurrent operations with overflow, persistence, etc. Adding more region types should be done as part of the other tickets (that's why I set the `RegionShortcut` to be configurable through parameters), not this one.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithConcurrentOperationsDUnitTest.java
jujoramos: The test is still valid, though, we aim to verify that there are no issues with concurrent `gets` and concurrent `clears`, doesn't matter how many entries (if any) the region has. There is another method within the class that tests concurrent `gets` + `puts` + `clears`.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithConcurrentOperationsDUnitTest.java
jujoramos: I'm not following you on this one... why only 5 seconds?, and why should the test verify how long the operation actually took?. The method basically executes the `clear` operation for a specific period of time to be able to interleave this operation with other concurrent ones in the cache. Measuring how long the operation took should be part of another ticket (a performance one, probably).

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithConcurrentOperationsDUnitTest.java
jujoramos: Any reason for that?. I can certainly do it, just wondering what's the actual reasoning behind the request.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithConcurrentOperationsDUnitTest.java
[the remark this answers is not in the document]
jujoramos: I'm in favour of having clear names in tests and classes, that's why I prefer to use long ones :-). Regarding the second comment, strongly agreed!.
This PR is to specifically address [GEODE-7670](https://issues.apache.org/jira/browse/GEODE-7670), that's why it only addresses `gets` and `puts`, we can certainly improve the test and add extra operations (or region types) as part of other tickets.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithConcurrentOperationsDUnitTest.java
jujoramos: Changed.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithConcurrentOperationsDUnitTest.java
jujoramos: Changed.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithConcurrentOperationsDUnitTest.java
jujoramos: Agreed, the `DistributionMessageObserver` can be used but it's a test hook and I've heard we want to get rid of that eventually, that's why I chose to use another approach (which seems to work fine, so I'm not sure I should to apply the requested change here).

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithConcurrentOperationsDUnitTest.java
jujoramos: The method asserts that the data is consistent across VMs by tacking a snapshot of the region per VM and comparing it against the other running VMs, it doesn't check primaries vs secondaries.

```
    vms.forEach(vm -> vm.invoke(() -> {
      final Map<String, String> thisSnapshot = waitForSilenceAndGetRegionSnapshot();
      assertThat(thisSnapshot).isEqualTo(vm0Snapshot);
      thisSnapshot.forEach((key, value) -> assertThat(value).isEqualTo("Value_" + key));
    }));

```

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithConcurrentOperationsDUnitTest.java
jujoramos: Done!.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithConcurrentOperationsDUnitTest.java
jujoramos: Same as above, I'm still inclined to use the `CacheWriter` for now instead of the `DistributionMessageObserver` hook. If the lifecycle of the listener changes, I'll be more than happy to use another approach :-). If you think this should block the PR from being merged, though, I'll explore using the `DistributionMessageObserver`.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithConcurrentOperationsDUnitTest.java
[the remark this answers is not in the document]
agingade: @jujoramos name change and adding new ops I will leave it to you. The acceptance criteria and ticket heading doesn't match; anyways we could add additional tickets to try out other ops.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithConcurrentOperationsDUnitTest.java
agingade: I won't foresee removal of observers; again there are multiple ways to do it; it is your choice :)

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithConcurrentOperationsDUnitTest.java
agingade: What I am trying to says is, since you are iterating over the PR region; it will get the data from all the nodes. Thats remains same across all the nodes.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithConcurrentOperationsDUnitTest.java
agingade: I don't think lifecycle of listeners changing, your choice.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithConcurrentOperationsDUnitTest.java
gesterzhou: 5 seconds is rvv domination timeout. If clear operation took more than 5 seconds, something must be wrong. No need to measure other operations' time.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithConcurrentOperationsDUnitTest.java
gesterzhou: putAll will have longer window to reproduce synchronization issues

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithConcurrentOperationsDUnitTest.java
jujoramos: @gesterzhou : I still don't understand what these tests have to do with the fact that a `clear` operation should take less than 5 seconds. The ticket is to purely verify that `clear` can be executed while other cache operations are being executed at the same time on the same region, shouldn't we create an extra ticket to test the scenario you're talking about?.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithConcurrentOperationsDUnitTest.java
jujoramos: Cool, will add another test to execute multiple `putAll` instead of regular `put` opertations.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithConcurrentOperationsDUnitTest.java
jujoramos: @agingade : got your point, thanks for catching this. Will change the method to verify that the data both on the primary and secondary buckets are in sync.

jujoramos:
@gesterzhou 
I've rebased the `PR` against `feature/GEODE-7665` to include the latest changes from the `clear` feature. There are still some failures while running the tests, concurrent `putAll` with `clear` (test method `clearWithConcurrentPutAllShouldWorkCorrectly`) always reproduces a discrepancy between primary buckets a secondary buckets for `PARTITION_REDUNDANT` regions.
Other tests also sporadically fail when I uncomment the following lines in the `assertRegionBucketsConsistency` method too:
```
assertThat(otherDump.getRvv())
    .as("RegionVersionVector for bucket " + bucketId + " on member " 
        + otherDump.getMember() + " is not consistent with member " + firstDump.getMember())
    .isEqualTo(firstDump.getRvv());
```

jujoramos:
@agingade 
According to what we discussed offline about the `clear` operation:
> - The clear may be successful or partial.
> - In case of partial, it will throw `PartitionedRegionPartialClearException`.
> - In both cases, the secondary bucket data should be same as primary (even with member departing).
> - The `RegionVersionVector` between primary and secondary should be same, no matter what.

Having that in mind, I've re-based the `PR` against the latest changes from the `feature/GEODE-7665` branch. There are still some test failures, below are the details:

- Test **clearWithConcurrentPutsShouldWorkCorrectly** always fails for `PARTITION_REDUNDANT` regions, doesn't matter whether the clear coordinator is the actual data host or the accessor node. After the clear finishes, the `RegionVersionVector` is not the same on the primary and secondary copies for a particular bucket.
- Test **clearShouldFailWhenCoordinatorMemberIsBounced** always fails for `PARTITION_REDUNDANT` regions. After the clear finishes, the `RegionVersionVector` is not the same on the primary and secondary copies for a particular bucket.
- Test **clearWithConcurrentPutAllShouldWorkCorrectly** always fails for `PARTITION_REDUNDANT` regions, doesn't matter whether the clear coordinator is the actual data host or the accessor node. After the clear finishes, the `RegionVersionVector` is not the same on the primary and secondary copies for a particular bucket.
- Test **clearWithConcurrentPutsAndGetsShouldWorkCorrectly** always fails for `PARTITION_REDUNDANT` regions, doesn't matter whether the clear coordinator is the actual data host or the accessor node. After the clear finishes, the `RegionVersionVector` is not the same on the primary and secondary copies for a particular bucket.
- Test **clearWithConcurrentPutsShouldWorkCorrectlyWhenNonCoordinatorMembersAreBounced** randomly fails (for `PARTITION_REDUNDANT` regions) with `PartitionedRegionPartialClearException` or inconsistent `RegionVersionVector` across buckets copies, I believe I'm hitting the same scenario as the one @DonalEvans described [here](https://github.com/apache/geode/pull/5095/files#r432000642).

Thoughts?.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithConcurrentOperationsDUnitTest.java
```diff
@@ -0,0 +1,680 @@
...
+          .isEqualTo(expectedCopies);
+
+      // Check that all copies of the bucket have the same data.
+      if (bucketDumps.size() > 1) {
+        BucketDump firstDump = bucketDumps.get(0);
+
+        for (int j = 1; j < bucketDumps.size(); j++) {
+          BucketDump otherDump = bucketDumps.get(j);
+
+          assertThat(otherDump.getRvv())
```
agingade, line 226: This validation needs to be changed....The RVV's are maintained as local-rvv (local to the node/member) and array of remote-rvvs (all other members). The RVVs has to be taken for corresponding member from both dumps and compared.

jujoramos:
@agingade 

I've updated the `PR` with the changes you suggested (add `bucketRegion.lockRVVForBulkOp()` and `bucketRegion.unlockRVVForBulkOp()` to `PutAllPRMessage.doLocalPutAll`). I've also modified the test to correctly compare the RVVs from the buckets.
I'm still seeing some random failures when comparing (primary vs secondary) the `values` and `versions` stored within the `Bucket`, specially whenever the `clear` is executed concurrently with `putAll`.
That said, I'd suggest to move your suggestions (`bucketRegion.lockRVVForBulkOp()` and `bucketRegion.unlockRVVForBulkOp()`) to a new ticket and only use this `PR` for the actual distributed test.
Please let me know what you think.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithConcurrentOperationsDUnitTest.java
```diff
@@ -0,0 +1,715 @@
...
+    // seconds.
+    List<AsyncInvocation<Void>> asyncInvocationList = Arrays.asList(
+        server1.invokeAsync(() -> executeGets(entries, workSeconds)),
+        server1.invokeAsync(() -> executePuts(entries, workSeconds)),
+        accessor.invokeAsync(() -> executeGets(entries, workSeconds)),
+        accessor.invokeAsync(() -> executeRemoves(entries, workSeconds)));
+
+    // Clear the region.
+    getVM(coordinatorVM.vmNumber).invoke(() -> {
+      assertThatThrownBy(() -> cacheRule.getCache().getRegion(REGION_NAME).clear())
```
gesterzhou, line 600: The PartialClearException could happen here but not 100% will happen. Your assertion could fail here.

jujoramos: @gesterzhou 

There are three members in this test: `ACCESSOR`, `SERVER1` and `SERVER2`, both servers host primary buckets locally for the region. The `clear` operation is executed either from the `ACCESSOR` or `SERVER1`, and `SERVER2` is always killed through the `MemberKiller` class (`CacheWriter.beforeRegionClear()`) before the `clear` is locally executed.
According to what I understand, the `PartitionedRegionPartialClearException` should always be thrown as `SERVER2` doesn't get started (after the initial shutdown) until the `clear` "finishes", meaning that the clear coordinator **_will never be able to get the lock for all the primary buckets_** (the region is configured as `PARTITION`, not as `PARTITION_REDUNDANT`, so there are no secondary buckets that can become primary while `SERVER2` is down).
Am I missing something here?, shouldn't the exception be thrown no matter what, given the above conditions?.

---

As a side note, the test `clearOnNonRedundantPartitionRegionWithConcurrentPutGetRemoveShouldFailWhenNonCoordinatorMembersAreBounced` failed 4 times out of 100 executions, the actual failure depends on whether the `clear` is executed on the `ACCESSOR` or `SERVER1` vm.
The two failures seen when the `clear` is executed on the the `ACCESSOR` node are basically assertion errors, caused by the fact the the `PartitionedRegionPartialClearException` is not thrown (discussed above).
The two failures seen when the `clear` is executed on `SERVER1`, on the other hand, are caused by the fact that the exception message is different from the one expected by the test (I'll change this and update the `PR`).

jujoramos: @gesterzhou 
I've modified the test assertion to always expect the `PartitionedRegionPartialClearException` but don't check the actual message, all the related failures are gone now.
The question still remains, though, the `PartitionedRegionPartialClearException` is always thrown under the conditions set by the test (that's what I would expect), why do you think the mentioned exception shouldn't be thrown 100% of time under these conditions?.

gesterzhou: Only when the primary holder cannot be found any where, the PartitionedRegionPartialClearException will be throw. 

In your case, only killing server2. There're chances and races that the primary buckets are recreated quick enough on server1. So PartitionedRegionPartialClearException is not 100%.

jujoramos: Maybe I'm missing something here, but there's no activity on the cluster at the time `server2` is killed, so the primaries previously hosted by that server shouldn't be recreated anywhere (and there are no secondaries to promote as the region is just `PARTITION`, not `REDUNDANT_PARTITION`).

jujoramos:
@agingade @gesterzhou 

After changing the approach used to shutdown the members during the tests (`DistributionMessageObserver` instead of `CacheWriter`) there are no more failures related to the `Timeout during netsearch/netload/netwrite` I was seeing before, CI is fully green so this `PR` is ready to merge once you approve it.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithConcurrentOperationsDUnitTest.java
```diff
@@ -0,0 +1,733 @@
...
+  private static final String TEST_CASE_NAME =
+      "[{index}] {method}(Coordinator:{0}, RegionType:{1})";
+
+  @Rule
+  public DistributedRule distributedRule = new DistributedRule(3);
+
+  @Rule
+  public CacheRule cacheRule = CacheRule.builder().createCacheInAll().build();
+
+  private VM accessor, server1, server2;
```
DonalEvans, line 82: It's recommended not to declare multiple variables inline, so this should ideally be split into three separate lines.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithConcurrentOperationsDUnitTest.java
```diff
@@ -0,0 +1,733 @@
...
+    ACCESSOR(0), SERVER1(1), SERVER2(2);
+
+    final int vmNumber;
+
+    TestVM(int vmNumber) {
+      this.vmNumber = vmNumber;
+    }
+  }
+
+  @SuppressWarnings("unused")
```
DonalEvans, line 94: This warning suppression is not necessary, as the method is used in `coordinatorsAndRegionTypes()` below.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithConcurrentOperationsDUnitTest.java
```diff
@@ -0,0 +1,733 @@
...
+
+  @Before
+  public void setUp() throws Exception {
+    server1 = getVM(TestVM.SERVER1.vmNumber);
+    server2 = getVM(TestVM.SERVER2.vmNumber);
+    accessor = getVM(TestVM.ACCESSOR.vmNumber);
+  }
+
+  private void initAccessor(RegionShortcut regionShortcut) {
+    @SuppressWarnings("rawtypes")
```
DonalEvans, line 129: This warning suppression can be removed if the below line is changed to `PartitionAttributes<String, String> attributes = new PartitionAttributesFactory<String, String>()`

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithConcurrentOperationsDUnitTest.java
```diff
@@ -0,0 +1,733 @@
...
+        .create();
+
+    cacheRule.getCache().createRegionFactory(regionShortcut)
+        .setPartitionAttributes(attributes)
+        .create(REGION_NAME);
+
+  }
+
+  private void initDataStore(RegionShortcut regionShortcut) {
+    @SuppressWarnings("rawtypes")
```
DonalEvans, line 142: This warning suppression can be removed if the below line is changed to `PartitionAttributes<String, String> attributes = new PartitionAttributesFactory<String, String>()`

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithConcurrentOperationsDUnitTest.java
```diff
@@ -0,0 +1,733 @@
...
+  }
+
+  /**
+   * Asserts that the RegionVersionVectors for both buckets are consistent.
+   *
+   * @param bucketId Id of the bucket to compare.
+   * @param bucketDump1 First bucketDump.
+   * @param bucketDump2 Second bucketDump.
+   */
+  @SuppressWarnings({"unchecked", "rawtypes"})
```
DonalEvans, line 198: These warning suppressions can be safely removed if some changes are made in this method, as described below.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithConcurrentOperationsDUnitTest.java
```diff
@@ -0,0 +1,733 @@
...
+  private void assertRegionVersionVectorsConsistency(int bucketId, BucketDump bucketDump1,
+      BucketDump bucketDump2) {
+    RegionVersionVector rvv1 = bucketDump1.getRvv();
+    RegionVersionVector rvv2 = bucketDump2.getRvv();
```
DonalEvans, line 202: These can be changed to `RegionVersionVector<?>` to prevent "rawtypes" warnings.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithConcurrentOperationsDUnitTest.java
```diff
@@ -0,0 +1,733 @@
...
+    assertThat(rvv1).isNotNull();
+    assertThat(rvv2).isNotNull();
+    Map<VersionSource, RegionVersionHolder> rvv2Members =
+        new HashMap<VersionSource, RegionVersionHolder>(rvv1.getMemberToVersion());
+    Map<VersionSource, RegionVersionHolder> rvv1Members =
+        new HashMap<VersionSource, RegionVersionHolder>(rvv1.getMemberToVersion());
+    for (Map.Entry<VersionSource, RegionVersionHolder> entry : rvv1Members.entrySet()) {
+      VersionSource memberId = entry.getKey();
+      RegionVersionHolder versionHolder1 = entry.getValue();
+      RegionVersionHolder versionHolder2 = rvv2Members.remove(memberId);
```
DonalEvans, line 227: The following change set will remove the need for suppressed "unchecked" and "rawtypes" warnings in this method:
```suggestion
    Map<VersionSource<?>, RegionVersionHolder<?>> rvv2Members =
        new HashMap<>(rvv1.getMemberToVersion());
    Map<VersionSource<?>, RegionVersionHolder<?>> rvv1Members =
        new HashMap<>(rvv1.getMemberToVersion());
    for (Map.Entry<VersionSource<?>, RegionVersionHolder<?>> entry : rvv1Members.entrySet()) {
      VersionSource<?> memberId = entry.getKey();
      RegionVersionHolder<?> versionHolder1 = entry.getValue();
      RegionVersionHolder<?> versionHolder2 = rvv2Members.remove(memberId);
```

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithConcurrentOperationsDUnitTest.java
```diff
@@ -0,0 +1,733 @@
...
+    while (Instant.now().isBefore(finishTime)) {
+      IntStream.range(0, numEntries).forEach(i -> region.put(String.valueOf(i), "Value_" + i));
+    }
+  }
+
+  /**
+   * Continuously execute putAll operations on the PartitionedRegion for the given
+   * durationInSeconds.
+   */
+  private void executePutAlls(final int start, final int finish, final long durationInSeconds) {
```
DonalEvans, line 308: The arguments "start" and "finish" here might be better named as "startKey" and "endKey" or something similar, to make it clear that they are not related to time.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithConcurrentOperationsDUnitTest.java
```diff
@@ -0,0 +1,733 @@
...
+        nullableValue.ifPresent(value -> assertThat(value).isEqualTo("Value_" + i));
+      });
+    }
+  }
+
+  /**
+   * Continuously execute removeAll operations on the PartitionedRegion for the given
+   * durationInSeconds.
+   */
+  private void executeRemoveAlls(final int start, final int finish, final long durationInSeconds) {
```
DonalEvans, line 343: The arguments "start" and "finish" here might be better named as "startKey" and "endKey" or something similar, to make it clear that they are not related to time.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithConcurrentOperationsDUnitTest.java
```diff
@@ -0,0 +1,733 @@
...
+   */
+  private void executeClears(final long durationInSeconds, final long waitTimeInMilliseconds)
+      throws InterruptedException {
+    Cache cache = cacheRule.getCache();
+    Region<String, String> region = cache.getRegion(REGION_NAME);
+    Instant finishTime = Instant.now().plusSeconds(durationInSeconds);
+
+    while (Instant.now().isBefore(finishTime)) {
+      region.clear();
+      Thread.sleep(waitTimeInMilliseconds);
```
DonalEvans, line 386: Instead of a `while` loop with a `Thread.sleep()`, it might be better to use the [Timer](https://docs.oracle.com/javase/7/docs/api/java/util/Timer.html) and [TimerTask](https://docs.oracle.com/javase/7/docs/api/java/util/TimerTask.html) classes to execute clears (specifically the `scheduleAtFixedRate()` method.) Entirely personal preference, but I do know that in general we try not to use `Thread.sleep()` directly in tests.

PR #4850, opened 2020-03-26
GEODE-7911: fix release script issues encountered in 1.12.0 release

onichols-pivotal:


geode-assembly/build.gradle
```diff
@@ -113,6 +113,14 @@ publishing {
...
 }
 
+gradle.taskGraph.whenReady( { graph ->
+  tasks.withType(Zip).each { zip ->
+    if ( "task ':geode-assembly:dockerfileZip'" == "${zip}") {
+      zip.enabled = false
+    }
+  }
+})
+
```
rhoughton-pivot: Instead, we can add the task to the list of other zip tasks we are disabling:
```
[tasks.named('distZip'),
 tasks.named('srcDistZip'),
 tasks.named('dockerfileZip')
 ]*.configure {
  enabled = false
}
```

onichols-pivotal: nice, that is much cleaner!

echobravopapa:
@rhoughton-pivot look good?

PR #4852, opened 2020-03-26
GEODE-7914: create missing unit test for Redis Module Expire Command

jhutchison:

### For all changes:

### Note:

### DIFF
--- a/geode-redis/src/test/java/org/apache/geode/redis/internal/executor/general/ExpireExecutorJUnitTest.java
+++ b/geode-redis/src/test/java/org/apache/geode/redis/internal/executor/general/ExpireExecutorJUnitTest.java
@@ -1,0 +1,116 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package org.apache.geode.redis.internal.executor.general;
+
+import static java.nio.charset.Charset.defaultCharset;
+import static org.assertj.core.api.AssertionsForInterfaceTypes.assertThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+import java.nio.charset.Charset;
+import java.util.ArrayList;
+import java.util.List;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.UnpooledByteBufAllocator;
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.ArgumentCaptor;
+
+import org.apache.geode.redis.internal.Command;
+import org.apache.geode.redis.internal.ExecutionHandlerContext;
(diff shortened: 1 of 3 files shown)


geode-redis/src/test/java/org/apache/geode/redis/internal/executor/general/ExpireExecutorJUnitTest.java
```diff
@@ -0,0 +1,116 @@
...
+package org.apache.geode.redis.internal.executor.general;
+
+import static java.nio.charset.Charset.defaultCharset;
+import static org.assertj.core.api.AssertionsForInterfaceTypes.assertThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+import java.nio.charset.Charset;
```
moleske: This and `import org.apache.geode.redis.internal.executor.string.GetRangeExecutor` are unused according to spotless

geode-redis/src/test/java/org/apache/geode/redis/internal/executor/general/ExpireExecutorJUnitTest.java
```diff
@@ -0,0 +1,116 @@
...
+    commandsAsBytesWithTooFewArguments.add("EXPIRE".getBytes());
+    commandsAsBytesWithTooFewArguments.add("key".getBytes());
+
+    ArgumentCaptor<ByteBuf> argsErrorCaptor = ArgumentCaptor.forClass(ByteBuf.class);
+
+    when(context.getByteBufAllocator()).thenReturn(byteBuf);
+    when(command.getProcessedCommand()).thenReturn(commandsAsBytesWithTooFewArguments);
+
+    executor.executeCommand(command, context);
+    verify(command, times(1)).setResponse(argsErrorCaptor.capture());
```
moleske: The default for `verify` is `times(1)`, so you can remove the explicit `times(1)` (same for the tests below)

geode-redis/src/test/java/org/apache/geode/redis/internal/executor/general/ExpireExecutorJUnitTest.java
jhutchison: yeah, I confess I was soling on this at the end of the day and couldn't remember how to run spotless.  We'll catch it next commit!

geode-redis/src/test/java/org/apache/geode/redis/internal/executor/general/ExpireExecutorJUnitTest.java
jhutchison: In the past I've adopted a sort of informal convention of leaving  the (unneeded) times(1) in there to  make it really obvious what's going on (in case a reader is not familiar with the default value), but not sure if that actually improves readability or not?  thoughts....?

geode-redis/src/test/java/org/apache/geode/redis/internal/executor/general/ExpireExecutorJUnitTest.java
moleske: `./gradlew spotlessApply` will auto apply the fixes it can (like removing imports)

geode-redis/src/test/java/org/apache/geode/redis/internal/executor/general/ExpireExecutorJUnitTest.java
moleske: Oh that's interesting.  I haven't run into that, mostly because I rarely want a test to verify something happens multiple times.  I will say when I looked through the test, I was like "Well that's interesting, there's a times, wonder why" and then saw "oh its one, that's boring."  So I guess seeing verify with a times is a trigger for me to look closer since it is not doing the normal just one time.  I'm not that strong against it which is why I marked the PR as approved.  Don't want to block good tests on something minor.  I think in the Geode code base we've only put `times(someNumber)` when that number is not one if you were to go looking in Geode for prior art.

jhutchison: yeah, makes sense.  Will take into consideration in future verifies :)   thanks for the feedback

PR #4857, opened 2020-03-27
Use Gradle lint rule to remove all deprecated dependency configurations

rhoughton-pivot:

### For all changes:

### Note:

### DIFF
--- a/buildSrc/build.gradle
+++ b/buildSrc/build.gradle
@@ -28,7 +28,7 @@ repositories {
 }
 
 dependencies {
-  testRuntime('org.apache.geode:geode-junit:1.3.0') {
+  testRuntimeOnly('org.apache.geode:geode-junit:1.3.0') {
     exclude group: 'org.apache.logging.log4j'
   }
   implementation(gradleApi())
--- a/geode-assembly/geode-assembly-test/build.gradle
+++ b/geode-assembly/geode-assembly-test/build.gradle
@@ -21,7 +21,7 @@ apply from: "${rootDir}/${scriptDir}/standard-subproject-configuration.gradle"
 
 
 dependencies {
-  implementation(platform(project(':boms:geode-all-bom')))
+  api(platform(project(':boms:geode-all-bom')))
   compileOnly(project(':extensions:geode-modules-test'))
   compileOnly(project(':geode-core'))
   compileOnly(project(':geode-pulse'))
--- a/geode-assembly/geode-assembly-test/build.gradle
+++ b/geode-assembly/geode-assembly-test/build.gradle
@@ -21,7 +21,7 @@ apply from: "${rootDir}/${scriptDir}/standard-subproject-configuration.gradle"
 
 
 dependencies {
-  compile(platform(project(':boms:geode-all-bom')))
+  implementation(platform(project(':boms:geode-all-bom')))
   compileOnly(project(':extensions:geode-modules-test'))
   compileOnly(project(':geode-core'))
   compileOnly(project(':geode-pulse'))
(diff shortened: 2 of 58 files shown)


extensions/geode-modules-test/build.gradle
```diff
@@ -22,15 +22,15 @@ apply from: "${rootDir}/${scriptDir}/standard-subproject-configuration.gradle"
 
 
 dependencies {
-  compile(platform(project(':boms:geode-all-bom')))
-  compile(project(':extensions:geode-modules')) {
+  api(platform(project(':boms:geode-all-bom')))
```
pivotal-jbarrett: I ran into issues where using the wrong scope here meant that the bom was not applied across the other scopes. It seems like `api` scope should be more correct but did you verify that the bom is locking down the versions correctly in compile and runtimeOnly scopes?

geode-assembly/geode-assembly-test/build.gradle
```diff
@@ -21,10 +21,11 @@ apply from: "${rootDir}/${scriptDir}/standard-subproject-configuration.gradle"
 
 
 dependencies {
-  compile(platform(project(':boms:geode-all-bom')))
+  api(platform(project(':boms:geode-all-bom')))
   compileOnly(project(':extensions:geode-modules-test'))
   compileOnly(project(':geode-core'))
   compileOnly(project(':geode-pulse'))
+  implementation(project(':geode-junit'))
```
pivotal-jbarrett: Why is this being added here? I wouldn't expect these changes to be adding any new dependencies.

rhoughton-pivot: `geode-dunit` passes `geode-junit` along as a `compile` dependency prior to this change. Modules were using `geode-junit` without explicitly asking for it. 
Alternatively, I could make `geode-dunit` declare `geode-junit` as an `api` dependency, but I think that would be lazy and technically incorrect.

extensions/geode-modules-test/build.gradle
rhoughton-pivot: `api` does not lock down the `compile` or `runtimeOnly` scopes. The bom declaration would need to be repeated for `runtimeOnly` dependencies.

We aren't supposed to use `compile` anywhere, so that's ok.

geode-assembly/geode-assembly-test/build.gradle
pivotal-jbarrett: Given that some of the classes in geode-dunit extend/implement classes in good-junit it more appropriate to make it `api`.

extensions/geode-modules-test/build.gradle
pivotal-jbarrett: Ok, good, so it does lock down `implementation`?

rhoughton-pivot: Ah, ok. According to [Gradle](https://docs.gradle.org/current/userguide/java_library_plugin.html) the `runtimeOnly` configuration does not inherit constraints from `api`, but *both* are inherited by `runtimeClasspath` which is what hits the POM. At that time, the constraint applies 👍

buildSrc/build.gradle
```diff
@@ -25,7 +25,7 @@ repositories {
 }
 
 dependencies {
-  testRuntime('org.apache.geode:geode-junit:1.3.0') {
+  testRuntimeOnly('org.apache.geode:geode-junit:1.3.0') {
     exclude group: 'org.apache.logging.log4j'
```
pivotal-jbarrett: Do we end up excluding log4j all over? Maybe we should make it `compileOnly` in `geode-junit`?

rhoughton-pivot:
I agree. Maybe beyond this scope of work.

onichols-pivotal:

PR #4861, opened 2020-03-27
GEODE-7905: RedisDistDUnitTest failing intermittently in CI

sabbeyPivotal:
Fix RedisDistDUnitTest, which fails intermittently:
https://concourse.apachegeode-ci.info/builds/141130
 
```
org.apache.geode.redis.RedisDistDUnitTest > classMethod FAILED
 java.lang.AssertionError: Suspicious strings were written to the log during this run.
 Fix the strings or use IgnoredException.addIgnoredException to ignore.
 -----------------------------------------------------------------------
 Found suspect string in log4j at line 5030
[error 2020/03/23 19:26:51.559 GMT <GeodeRedisServer-WorkerThread-3> tid=99] org.apache.geode.ToDataException: toData failed on DataSerializer with id=0 for class class java.util.HashSet
6 tests completed, 1 failed
```

Updates SADD executor to use optimistic instead of pessimistic locking.

### DIFF
--- a/geode-redis/src/main/java/org/apache/geode/redis/internal/executor/set/SAddExecutor.java
+++ b/geode-redis/src/main/java/org/apache/geode/redis/internal/executor/set/SAddExecutor.java
@@ -17,10 +17,9 @@ package org.apache.geode.redis.internal.executor.set;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicLong;
 
 import org.apache.geode.cache.Region;
-import org.apache.geode.cache.TimeoutException;
-import org.apache.geode.redis.internal.AutoCloseableLock;
 import org.apache.geode.redis.internal.ByteArrayWrapper;
 import org.apache.geode.redis.internal.Coder;
 import org.apache.geode.redis.internal.Command;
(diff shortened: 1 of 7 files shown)


geode-redis/src/main/java/org/apache/geode/redis/internal/executor/set/SAddExecutor.java
```diff
@@ -30,49 +28,58 @@
...
+//    Long entriesAdded = 0L;
+    Region<ByteArrayWrapper, Set<ByteArrayWrapper>> region = getRegion(context);
+
+    ByteArrayWrapper keyPlaceHolder = null;
 
-      Set<ByteArrayWrapper> entries = region.get(key);
-      if (entries == null) {
-        entries = new HashSet<>();
-      }
+    region.compute(
```
dschneider-pivotal, line 47: FYI, compute on Region does not have a region specific implementation. It is implemented with the default impl from ConcurrentMap. So it basically does a region.get, executes your lambda, and does a region.put. It will retry but I think this will have the same concurrent modification issue you saw before because you modifying directly the "oldRegionData"

ringles: It's been tweaked to modify a copy.

jdeppe-pivotal:
Since we're actively working on various approaches to this, and it's probably still going to take a while, I'm going to close this PR for now.

PR #4862, opened 2020-03-27
GEODE-7919: Moving membership integration test to geode-membership

upthewaterspout:
Moving MembershipOnlyTest to MembershipIntegrationTest in the geode-membership
module and adding some more tests.


geode-membership/src/integrationTest/java/org/apache/geode/distributed/internal/membership/gms/MembershipIntegrationTest.java
```diff
@@ -0,0 +1,206 @@
...
+
+    final Membership<MemberIdentifier> membership = createMembership(locator,
+        locator.getPort());
+    start(membership);
+
+    assertThat(membership.getView().getMembers()).hasSize(1);
+  }
+
+  @Test
+  public void twoMembersCanStartWithOneLocator()
```
Bill: do we want to say `twoMembershhipsCanStart…` to be consistent w/ other test names?

geode-membership/src/integrationTest/java/org/apache/geode/distributed/internal/membership/gms/MembershipIntegrationTest.java
```diff
@@ -0,0 +1,206 @@
...
+      public String getLocators() {
+        return getLocatorString(locatorPorts);
+      }
+
+      // TODO - the Membership system starting in the locator *MUST* be told that is
+      // is a locator through this flag. Ideally it should be able to infer this from
+      // being associated with a locator
+      @Override
+      public int getVmKind() {
+        return isALocator ? MemberIdentifier.LOCATOR_DM_TYPE : MemberIdentifier.NORMAL_DM_TYPE;
```
Bill: would it be better to send in the VM kind instead of the boolean `isALocator`? doing that would eliminate the conditional here.

geode-membership/src/integrationTest/java/org/apache/geode/distributed/internal/membership/gms/MembershipIntegrationTest.java
```diff
@@ -0,0 +1,206 @@
...
+  private InetAddress localHost;
+  private DSFIDSerializer dsfidSerializer;
+  private TcpSocketCreator socketCreator;
+
+  @Rule
+  public TemporaryFolder temporaryFolder = new TemporaryFolder();
+
+  @Before
+  public void before() throws IOException, MembershipConfigurationException {
+    localHost = InetAddress.getLocalHost();
```
bschuchardt: This should use LocalHostUtil.getLocalHost()

geode-membership/src/integrationTest/java/org/apache/geode/distributed/internal/membership/gms/MembershipIntegrationTest.java
upthewaterspout: Seems like that would just shift the conditional? I think I'll leave it the way it is for now since we want to get rid of this flag entirely when we fix the membership code.

geode-membership/src/integrationTest/java/org/apache/geode/distributed/internal/membership/gms/MembershipIntegrationTest.java
upthewaterspout: Done

PR #4866, opened 2020-03-28
GEODE-7864: Overriding hashCode when equals is overridden.

nabarunnag:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/cache/client/internal/pooling/AvailableConnectionManager.java
+++ b/geode-core/src/main/java/org/apache/geode/cache/client/internal/pooling/AvailableConnectionManager.java
@@ -15,6 +15,7 @@
 package org.apache.geode.cache.client.internal.pooling;
 
 import java.util.Deque;
+import java.util.Objects;
 import java.util.concurrent.ConcurrentLinkedDeque;
 import java.util.function.Predicate;
 
@@ -148,6 +149,11 @@ public class AvailableConnectionManager {
         return true;
       }
       return false;
+    }
+
+    @Override
+    public int hashCode() {
+      return Objects.hash(predicate, connectionThatMatched);
     }
 
     public Connection getConnectionThatMatched() {
--- a/geode-dunit/src/main/java/org/apache/geode/internal/cache/functions/TestFunction.java
+++ b/geode-dunit/src/main/java/org/apache/geode/internal/cache/functions/TestFunction.java
@@ -23,6 +23,7 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Properties;
 import java.util.Set;
 import java.util.concurrent.CancellationException;
@@ -1028,6 +1029,11 @@ public class TestFunction<T> implements Function<T>, Declarable2, DataSerializab
     }
     TestFunction function = (TestFunction) obj;
     return this.props.equals(function.getConfig());
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(props);
   }
 
   @Override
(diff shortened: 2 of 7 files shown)


geode-core/src/main/java/org/apache/geode/cache/client/internal/pooling/AvailableConnectionManager.java
```diff
@@ -146,6 +146,11 @@ public boolean equals(Object o) {
       return false;
     }
```
pivotal-jbarrett: I don’t think there is even a reason this class implements equals. I’d say just let them be default.

geode-dunit/src/main/java/org/apache/geode/internal/cache/functions/TestFunction.java
```diff
@@ -1032,6 +1032,11 @@ public boolean equals(Object obj) {
     return this.props.equals(function.getConfig());
   }
 
+  @Override
+  public int hashCode() {
```
pivotal-jbarrett: This implementation seems to violate the general rules of hashcode. It would use the same members as equals to generate a hashcode.

geode-core/src/main/java/org/apache/geode/cache/client/internal/pooling/AvailableConnectionManager.java
nabarunnag: - This was implemented as a refactoring of the ConnectionManagerImpl as a part of improving performance.
- I believe that we cannot use the default implementation because it needs to store the first connection which was removed from the deque while executing  removeFirstOccurrence method in Deque.

geode-dunit/src/main/java/org/apache/geode/internal/cache/functions/TestFunction.java
nabarunnag: Fixed. We now depend on the hashcode of the properties map,  if they have the same properties then they are equal and will have the same hashcode.

geode-core/src/main/java/org/apache/geode/cache/client/internal/pooling/AvailableConnectionManager.java
pivotal-jbarrett: Oh yes, that is correct. I didn't notice this was that special inner class. Thanks!

geode-dunit/src/main/java/org/apache/geode/internal/cache/functions/TestFunction.java
pivotal-jbarrett: I think the same issue exists on the other hashcodes added in this commit.

geode-junit/src/main/java/org/apache/geode/pdx/TestObjectForJSONFormatter.java
```diff
@@ -992,6 +993,34 @@ boolean compareMaps(Map m1, Map m2) {
...
   }
 
+  @Override
+  public int hashCode() {
+    int result =
+        Objects
+            .hash(p_bool, p_byte, p_short, p_int, p_long, p_float, p_double, w_bool, w_byte,
+                w_short,
+                w_int, w_long, w_bigInt, w_float, w_bigDec, w_double, w_string, c_list, c_set,
+                c_queue, c_deque, m_empByCity, day);
```
DonalEvans, line 1003: Could this be formatted slightly neater? The following seems to make spotless happy on my machine:
`int result = Objects.hash(p_bool, p_byte, p_short, p_int, p_long, p_float, p_double, w_bool,
        w_byte, w_short, w_int, w_long, w_bigInt, w_float, w_bigDec, w_double, w_string, c_list,
        c_set, c_queue, c_deque, m_empByCity, day);`

geode-junit/src/main/java/org/apache/geode/pdx/TestObjectForJSONFormatter.java
```diff
@@ -992,6 +993,31 @@ boolean compareMaps(Map m1, Map m2) {
     return true;
   }
 
+  @Override
+  public int hashCode() {
+    int result = Objects.hash(p_bool, p_byte, p_short, p_int, p_long, p_float, p_double, w_bool,
```
pivotal-jbarrett, line 998: What about:
```java
return Arrays.deepHashCode(new Object[] {p_pool, p_poolArray, ...});
```

PR #4871, opened 2020-03-29
GEODE-7864: Override the default implementation of write(byte[],int,int)

nabarunnag:
* Prevent overheads of writing one byte at a time.

### For all changes:

### Note:

### DIFF
--- a/geode-common/src/main/java/org/apache/geode/util/internal/TeeOutputStream.java
+++ b/geode-common/src/main/java/org/apache/geode/util/internal/TeeOutputStream.java
@@ -62,6 +62,15 @@ public class TeeOutputStream extends FilterOutputStream {
   }
 
   @Override
+  public void write(byte[] b, int off, int len) throws IOException {
+    super.write(b, off, len);
+    OutputStream os = this.branch;
+    if (os != null) {
+      os.write(b, off, len);
+    }
+  }
+
+  @Override
   public void flush() throws IOException {
     super.flush();
     OutputStream os = this.branch;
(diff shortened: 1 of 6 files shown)


geode-common/src/main/java/org/apache/geode/util/internal/TeeOutputStream.java
```diff
@@ -61,6 +61,15 @@ public void write(int b) throws IOException {
     }
   }
 
+  @Override
+  public void write(byte[] b, int off, int len) throws IOException {
+    super.write(b, off, len);
+    OutputStream os = this.branch;
```
agingade: Looks liker we are calling write twice now. Earlier behavior was calling write on parent class FilterOutputStream.

pivotal-jbarrett: I don't see any unit tests for this class. I think you are right that this will write the bytes twice the branch stream.

nabarunnag: This is the existing write command:
`@Override
  public void write(int b) throws IOException {
    super.write(b);
    OutputStream os = this.branch;
    if (os != null) {
      os.write(b);
    }
  }`

There is no `private volatile OutputStream branch;`  in the parent class. I did not understand the comment. @agingade

nabarunnag: The Constructor can provide two output streams. the super call will only write into the one of the stream not on the other and hence we have to write into the other stream in the overridden method

PR #4872, opened 2020-03-29
GEODE-7864: Prevented non synchronized override of synchronized methods.

nabarunnag:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/DistTXPrecommitMessage.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/DistTXPrecommitMessage.java
@@ -315,7 +315,7 @@ public class DistTXPrecommitMessage extends TXMessage {
     }
 
     @Override
-    protected void processException(DistributionMessage msg, ReplyException ex) {
+    protected synchronized void processException(DistributionMessage msg, ReplyException ex) {
       if (msg instanceof ReplyMessage) {
         synchronized (this) {
           if (this.exception == null) {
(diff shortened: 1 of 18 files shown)


geode-core/src/main/java/org/apache/geode/internal/cache/DistTXPrecommitMessage.java
```diff
@@ -315,7 +315,7 @@ public void waitForPrecommitCompletion() {
     }
 
     @Override
-    protected void processException(DistributionMessage msg, ReplyException ex) {
+    protected synchronized void processException(DistributionMessage msg, ReplyException ex) {
```
agingade: We may not need synchronized(this) in this method.

geode-core/src/main/java/org/apache/geode/internal/cache/DistTXRollbackMessage.java
```diff
@@ -311,7 +311,7 @@ public void waitForPrecommitCompletion() {
     }
 
     @Override
-    protected void processException(DistributionMessage msg, ReplyException ex) {
+    protected synchronized void processException(DistributionMessage msg, ReplyException ex) {
```
agingade: We may not need synchronized(this) in this method.

geode-core/src/main/java/org/apache/geode/internal/cache/DistTXPrecommitMessage.java
nabarunnag: The parents are synchronized and the override is not. This is not considered a good practice. I will look into this a bit more.

geode-core/src/main/java/org/apache/geode/internal/cache/DistTXRollbackMessage.java
nabarunnag: The parents are synchronized and the override is not. This is not considered a good practice. I will look into this a bit more.

PR #4885, opened 2020-03-30
GEODE-7938: change dev rest api to support slashes in key parameters

dschneider-pivotal:
Tests have been added that confirm that the key can now contain a '/'.
This was done by changing "/{keys}" to "/**" and adding code to parse the keys.

### For all changes:

### Note:


geode-web-api/src/main/java/org/apache/geode/rest/internal/web/controllers/PdxBasedCrudController.java
```diff
@@ -295,12 +296,13 @@ protected String getRestApiVersion() {
       @ApiResponse(code = 500, message = "GemFire throws an error or exception.")})
   @PreAuthorize("@securityService.authorize('WRITE', #region, #keys)")
   public ResponseEntity<?> update(@PathVariable("region") String region,
-      @PathVariable("keys") final String[] keys,
+      @PathVariable("keys") String[] keys,
```
pivotal-jbarrett: Have you considered just using the Spring annotations and patterns to match the remainder of the URL path?

`@RequestMapping(value = "/{region}/**", ...)`
and
`String[] keys = request.getAttribute(HandlerMapping.PATH_WITHIN_HANDLER_MAPPING_ATTRIBUTE).split(",")`

dschneider-pivotal: It is not clear to me why using /** would be an better than the current solution. The current code is allowing spring to parse the pathvar into an array of String. Why is your way of doing it explicitly better? And how does this relate to supporting a slash in a key? Would your solution not require the keys to be encoded by the user? This /** solution is discouraged here: https://www.baeldung.com/spring-slash-character-in-url but so is encoding/decoding. The current code base already did decoding of each {region} pathvar so that is why I went with also decoding the {keys} pathvar. Also if we don't decode then the swagger page will not work with keys because swagger automatically encodes path and query vars when you execute the apis from the web page.

pivotal-jbarrett: It relates to supporting slashes in the path without needing to encode them. It makes sense to do it they way you proposed if we are already expecting the region to be encoded.

dschneider-pivotal:
@pivotal-jbarrett the main, unaddressed, issue with the current solution is that calling decode on each key may  break existing users that currently are using keys that contain characters that decode would change to something else. The breaking of backwards compatibility needs to be addressed. Jake's solution appears to address that since it allows us to have keys that contain slashes without needing to decode them. So I'm going to try that out and see how it works.

geode-web-api/src/main/java/org/apache/geode/rest/internal/web/controllers/PdxBasedCrudController.java
```diff
@@ -196,10 +198,9 @@ protected String getRestApiVersion() {
    * Reading data for set of keys
    *
    * @param region gemfire region name
-   * @param keys string containing comma separated keys
    * @return JSON document
    */
-  @RequestMapping(method = RequestMethod.GET, value = "/{region}/{keys}",
+  @RequestMapping(method = RequestMethod.GET, value = "/{region}/**",
```
pivotal-jbarrett, line 203: So does doing this now effect any swagger documentation or code generation?

geode-web-api/src/main/java/org/apache/geode/rest/internal/web/controllers/AbstractBaseController.java
```diff
@@ -983,4 +974,16 @@ protected QueryService getQueryService() {
     targetedMembers.add(cache.getDistributedSystem().getDistributedMember());
     return targetedMembers;
   }
+
+  protected String[] parseKeys(HttpServletRequest request, String region) {
+    String uri = request.getRequestURI();
```
pivotal-jbarrett, line 979: Did `request.getAttribute(HandlerMapping.PATH_WITHIN_HANDLER_MAPPING_ATTRIBUTE)` not work?

dschneider-pivotal: No it did not. It gave the entire uri. This was in the context of the RestAccessControllerTest

geode-web-api/src/integrationTest/java/org/apache/geode/rest/internal/web/controllers/RestAccessControllerTest.java
```diff
@@ -182,8 +183,7 @@ public void postEntry() throws Exception {
   @Test
   @WithMockUser
   public void postEntryWithSlashKey() throws Exception {
-    String decodedKey = "1/2";
-    String key = URLEncoder.encode(decodedKey, "UTF-8");
+    String key = "1" + SLASH + "2";
```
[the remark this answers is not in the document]
dschneider-pivotal: I just wanted on easy way to be able to change the value of SLASH and have all the tests use it. I was curious about other special characters

geode-web-api/src/main/java/org/apache/geode/rest/internal/web/controllers/AbstractBaseController.java
dschneider-pivotal: Another thing that did not do what was expected was "request.getContextPath()". It returned an empty string. I found an example that expected it to be the first part of the URL up to the start of {region}.

geode-web-api/src/main/java/org/apache/geode/rest/internal/web/controllers/PdxBasedCrudController.java
dschneider-pivotal: swagger shows the endpoint as /v1/{region}/** which seems okay
But if you click try it out it gives you no way (that I could find) to specify a value for the **

pivotal-jbarrett: Feels like there is no real good solution to this problem.

PR #4898, opened 2020-04-01
GEODE-7921: NullPointerExceptions logged during auto-reconnect

bschuchardt:
Don't deliver cache-level messages that were queued while disconnected
for auto-reconnect.  During auto-reconnect there is a QuorumChecker that
receives messages on the jgroups channel and tries to establish
communications with a quorum of the old membership view.  It may also
get JoinRequest messages and other membership-level messages but I
observed one case where it also queued cache-level messages when the
property disable-tcp was set to true (funnelling all comms through
jgroups).

I also added some null checks to the LatestLastAccessTimeMessage and a
small test for that.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/LatestLastAccessTimeMessage.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/LatestLastAccessTimeMessage.java
@@ -62,17 +62,23 @@ public class LatestLastAccessTimeMessage<K> extends PooledDistributionMessage
   @Override
   protected void process(ClusterDistributionManager dm) {
     long latestLastAccessTime = 0L;
+    InternalCache cache = dm.getCache();
+    if (cache == null) {
+      return;
+    }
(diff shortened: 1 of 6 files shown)


geode-core/src/main/java/org/apache/geode/internal/cache/LatestLastAccessTimeMessage.java
```diff
@@ -62,17 +62,23 @@ public int getDSFID() {
...
-    if (region != null) {
-      RegionEntry entry = region.getRegionEntry(this.key);
-      if (entry != null) {
-        try {
-          latestLastAccessTime = entry.getLastAccessed();
-        } catch (InternalStatisticsDisabledException ignored) {
-          // last access time is not available
-        }
-      }
+        (InternalDistributedRegion) cache.getRegion(this.regionName);
```
echobravopapa, line 70: so the NPE could happen right here, got it

PR #4899, opened 2020-04-02
GEODE-7945: Replace Map with ConcurrentMap for GemFireCacheImpl.rootRegions

jchen21:
Authored-by: Jianxia Chen <jchen21@apache.org>

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/GemFireCacheImpl.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/GemFireCacheImpl.java
@@ -408,7 +408,7 @@ public class GemFireCacheImpl implements InternalCache, InternalClientCache, Has
 
   private final DistributionManager dm;
 
-  private final Map<String, InternalRegion> rootRegions;
+  private final ConcurrentMap<String, InternalRegion> rootRegions;
 
   /**
    * True if this cache is being created by a ClientCacheFactory.
@@ -998,7 +998,7 @@ public class GemFireCacheImpl implements InternalCache, InternalClientCache, Has
         throw new IllegalStateException("Cannot create a Cache in an admin-only VM.");
       }
 
-      rootRegions = new HashMap<>();
+      rootRegions = new ConcurrentHashMap<>();
 
       cqService = cqServiceFactory.apply(this);
 
@@ -2176,9 +2176,8 @@ public class GemFireCacheImpl implements InternalCache, InternalClientCache, Has
           prepareDiskStoresForClose();
 
           List<InternalRegion> rootRegionValues;
-          synchronized (rootRegions) {
-            rootRegionValues = new ArrayList<>(rootRegions.values());
-          }
+
+          rootRegionValues = new ArrayList<>(rootRegions.values());
 
           Operation op;
           if (forcedDisconnect) {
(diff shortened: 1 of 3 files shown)


geode-core/src/main/java/org/apache/geode/internal/cache/GemFireCacheImpl.java
```diff
@@ -3438,16 +3436,16 @@ public boolean getCopyOnRead() {
...
-      if (found == rootRgn) {
-        InternalRegion previous = rootRegions.remove(regionName);
-        Assert.assertTrue(previous == rootRgn);
-        return true;
-      }
-      return false;
+
+    String regionName = rootRgn.getName();
+    InternalRegion found = rootRegions.get(regionName);
+    if (found == rootRgn) {
```
dschneider-pivotal, line 3442: This code no longer looks correct. The synchronization made this thread safe.
Without the sync you need to just call a method on rootRegions once.
I think you just want to do this: return rootRegions.remove(regionName, rootRgn);

geode-core/src/main/java/org/apache/geode/internal/cache/GemFireCacheImpl.java
```diff
@@ -2997,12 +2995,12 @@ public void validatePoolFactory(PoolFactory poolFactory) {
           } finally {
             // clean up if initialize fails for any reason
             setRegionByPath(region.getFullPath(), null);
-            synchronized (rootRegions) {
-              Region rootRegion = rootRegions.get(name);
-              if (rootRegion == region) {
-                rootRegions.remove(name);
-              }
+
+            Region rootRegion = rootRegions.get(name);
```
dschneider-pivotal, line 2999: This does not look thread safe. Instead of calling get and then remove you need a single call.
Here I think you want: rootRegions.remove(name, region);

geode-core/src/main/java/org/apache/geode/internal/cache/GemFireCacheImpl.java
```diff
@@ -2905,45 +2904,44 @@ public void validatePoolFactory(PoolFactory poolFactory) {
...
+                  statisticsClock);
             }
+          }
 
-            rootRegions.put(name, region);
-            if (isReInitCreate) {
-              regionReinitialized(region);
-            }
-            break;
+          rootRegions.put(name, region);
```
dschneider-pivotal, line 2940: This does not look thread safe. I don't think we want to do a get followed by a put here. This allows two threads to both see null on the get and then they both create it and the last one wins. Maybe a single call of computeIfAbsent instead of the get+put is what you want here

geode-core/src/main/java/org/apache/geode/internal/cache/GemFireCacheImpl.java
```diff
@@ -2176,9 +2175,8 @@ public void close(String reason, Throwable systemFailureCause, boolean keepAlive
           prepareDiskStoresForClose();
 
           List<InternalRegion> rootRegionValues;
-          synchronized (rootRegions) {
-            rootRegionValues = new ArrayList<>(rootRegions.values());
-          }
+
+          rootRegionValues = new ArrayList<>(rootRegions.values());
```
dschneider-pivotal, line 2179: I see no reason for continuing to have this method make a copy of "rootRegions.values()". All it does with "rootRegionValues" is an iteration. So now that you have changed to a ConcurrentMap, I think you can safely get rid of "rootRegionValues" and down on line 2194 do this: for (InternalRegion lr: rootRegions.values())

jchen21: If we do `for (InternalRegion lr : rootRegions.vales())`, other threads may change `rootRegions` while the `for` loop is still in progress. Since each `for` iteration evaluates `rootRegions.values()`, the resulting behavior could be different than the original behavior, which iterates over a snapshot of `rootRegions`.

dschneider-pivotal: I you think about what you are doing now when you make the copy, it is actually iterating over the values to copy them into your ArrayList; without any type of lock. Basically ConcurrentMap gives you a Set that you can safely iterate without locking. It may change but you would only notice this if you did something like iterate it twice and compared the first iteration to the second. Since all this code does is a simple iteration you don't need to make a copy to be safe. It is not going to break the code; but you are wasting some memory and cpu and possibly making future coders think you need to make a copy to be safe :-)

dschneider-pivotal: By the way, each iteration does not evaluate rootRegion.values(). Values is called once and the result is iterated. Each iteration gets the next value from the Set that values() returned.

PR #4900, opened 2020-04-02
GEODE-6819: Fix PartitionedRegionSingleHopDUnitTest BindExceptions

kirklund:
Intermittent failures were caused by trying to stop and then restart
server(s) with the previously used ports which are no longer free on
the host machine.

This PR overhauls all of PartitionedRegionSingleHopDUnitTest while 
also fixing flaky failures caused by underlying BindExceptions in both:
* testMetadataIsSameOnAllServersAndClients
* testSingleHopWithHAWithLocator

I've added comments below that point to the actual BindException
fixes to aid in reviewing.

### DIFF
--- a/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionSingleHopDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionSingleHopDUnitTest.java
@@ -16,35 +16,40 @@ package org.apache.geode.internal.cache;
 
 import static java.util.Arrays.asList;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
+import static java.util.concurrent.TimeUnit.SECONDS;
+import static org.apache.geode.cache.RegionShortcut.LOCAL;
+import static org.apache.geode.cache.RegionShortcut.PARTITION;
+import static org.apache.geode.cache.RegionShortcut.PARTITION_PERSISTENT;
 import static org.apache.geode.distributed.ConfigurationProperties.ENABLE_CLUSTER_CONFIGURATION;
 import static org.apache.geode.distributed.ConfigurationProperties.LOCATORS;
+import static org.apache.geode.internal.cache.util.UncheckedUtils.cast;
 import static org.apache.geode.internal.lang.SystemPropertyHelper.GEMFIRE_PREFIX;
 import static org.apache.geode.management.ManagementService.getExistingManagementService;
 import static org.apache.geode.test.awaitility.GeodeAwaitility.await;
 import static org.apache.geode.test.awaitility.GeodeAwaitility.getTimeout;
-import static org.apache.geode.test.dunit.Disconnect.disconnectAllFromDS;
 import static org.apache.geode.test.dunit.IgnoredException.addIgnoredException;
 import static org.apache.geode.test.dunit.VM.getController;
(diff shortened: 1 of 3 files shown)


geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionSingleHopDUnitTest.java
```diff
@@ -620,61 +666,63 @@ public void testMetadataFetchOnlyThroughputAll() {
...
+          .hasSize(1)
+          .containsKey(partitionedRegion.getFullPath());
+      assertThat(prMetadata.getBucketServerLocationsMap_TEST_ONLY())
+          .hasSize(totalNumberOfBuckets);
     });
   }
 
   @Test
   public void testMetadataIsSameOnAllServersAndClients() {
+    int locatorPort = DUnitEnv.get().getLocatorPort();
```
kirklund, line 695: dunitrunner revealed that testMetadataIsSameOnAllServersAndClients fails intermittently with underlying BindExceptions thrown when restarting server in vm0 or vm1 with the same port those two vms used to have a CacheServer running on.

The only way to fix this is to use a Locator in the pool instead of hardcoded server ports, and allow the CacheServers to come back up on any port.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionSingleHopDUnitTest.java
```diff
@@ -386,75 +428,70 @@ public void testSingleHopWithHA() {
...
-      testRegion.put(i, i + 10);
+      partitionedRegion.put(i, i + 10);
     }
   }
 
   @Test
   public void testSingleHopWithHAWithLocator() {
-    int port3 = AvailablePort.getRandomAvailablePort(AvailablePort.SOCKET);
-    String locator = "localhost[" + port3 + "]";
+    int locatorPort = vm3.invoke(() -> startLocator());
```
kirklund, line 437: dunitrunner revealed intermittent failures in testSingleHopWithHAWithLocator with an underlying BindException as the cause.

The easiest fix is to change startLocator to use port zero and then return whatever port the Locator ended up binding to.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionSingleHopDUnitTest.java
```diff
@@ -868,585 +916,338 @@ public void testMetadataIsSameOnAllServersAndClientsHA() {
   public void testClientMetadataForPersistentPrs() throws Exception {
     LATCH.set(new CountDownLatch(4));
 
-    int port0 = vm0.invoke(() -> createPersistentPrsAndServer());
-    int port1 = vm1.invoke(() -> createPersistentPrsAndServer());
-    int port2 = vm2.invoke(() -> createPersistentPrsAndServer());
-    int port3 = vm3.invoke(() -> createPersistentPrsAndServer());
+    int locatorPort = DUnitEnv.get().getLocatorPort();
```
kirklund, line 919: dunitrunner revealed flaky failures in testClientMetadataForPersistentPrs caused by underlying BindExceptions.

The fix involves pointing the client at the servers via the locator and then let the servers use whatever ports they can find instead of restarting the servers with a previously used port.

PR #4909, opened 2020-04-06
GEODE-7953: Restore Redundancy Internal API

DonalEvans:
- Add RestoreRedundancyBuilder interface
- Add accessor methods for RestoreRedundancyBuilder to ResourceManager interface
- Add RestoreRedundancyResults interface
- Add RestoreRedundancyDirector for use in PartitionedRegionRebalanceOp
- Add stats for restore redundancy operations
- Add RestoreRedundancyRegionResult to describe the redundancy status of
a partitioned region
- Add unit and DUnit tests for all the above
- Add rolling upgrade test

### For all changes:

### Note:


geode-core/src/main/java/org/apache/geode/internal/cache/control/ResourceManagerStats.java
```diff
@@ -92,6 +95,16 @@
...
+            f.createLongCounter("restoreRedundanciesInProgress",
+                "Current number of cache restore redundancy operations being directed by this process.",
+                "operations"),
+            f.createLongCounter("restoreRedundanciesCompleted",
+                "Total number of cache restore redundancy operations directed by this process.",
+                "operations"),
+            f.createLongCounter("restoreRedundancyTime",
+                "Total time spent directing cache restore redundancy operations.", "nanoseconds",
+                false),
+
```
jujoramos, line 107: I think we should add these new statistics to `geode-docs/reference/statistics_list.html.md.erb`.

geode-core/src/main/java/org/apache/geode/internal/cache/control/RestoreRedundancyRegionResult.java
```diff
@@ -0,0 +1,139 @@
...
+    desiredRedundancy = region.getRedundantCopies();
+    actualRedundancy = calculateLowestRedundancy(region);
+    status = determineStatus(desiredRedundancy, actualRedundancy);
+  }
+
+  public String getRegionName() {
+    return regionName;
+  }
+
+  public int getDesiredRedundancy() {
```
jujoramos, line 85: According to the IDE this method is never invoked.
Unless we foresee somebody using it within the near future, can we just delete it?.

geode-core/src/upgradeTest/java/org/apache/geode/internal/cache/rollingupgrade/RollingUpgradeRestoreRedundancy.java
```diff
@@ -0,0 +1,111 @@
...
+  @Parameterized.Parameters(name = "from_v{0}")
+  public static Collection<String> data() {
+    List<String> result = VersionManager.getInstance().getVersionsWithoutCurrent();
+    if (result.size() < 1) {
+      throw new RuntimeException("No older versions of Geode were found to test against");
+    } else {
+      // The restore redundancy feature was added in 1.13.0, so versions newer than that do not need
+      // to be tested in a rolling upgrade environment
+      result.removeIf(version -> TestVersion.compare(version, "1.13.0") > 0);
+      System.out.println("running against these versions: " + result);
```
jujoramos, line 52: Do we need this debug message in the test?.

geode-core/src/main/java/org/apache/geode/cache/control/RestoreRedundancyResults.java
```diff
@@ -0,0 +1,151 @@
...
+import org.apache.geode.internal.cache.partitioned.PartitionedRegionRebalanceOp;
+
+/**
+ * A class to collect the results of restore redundancy operations for one or more regions and
+ * determine the success of failure of the operation.
+ */
+public interface RestoreRedundancyResults {
+
+  /**
+   * {@link #SUCCESS} is defined as every included region having at least one redundant copy.
```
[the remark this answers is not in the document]
DonalEvans: The idea is that as long as redundancy has been established, at any level, then the operation has succeeded, since in 99% of cases, a user cares far more about whether they have redundancy or not vs how much redundancy they have. A user can get information about how many regions (if any) do not have fully satisfied redundancy using the `RestoreRedundancyResults.getUnderRedundancyRegionResults()` method if they need that information.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/control/RestoreRedundancyOperationDUnitTest.java
```diff
@@ -0,0 +1,348 @@
...
+      assertRedundancyStatus(childRegionName, false);
+      assertRedundancyStatus(lowRedundancyRegionName, false);
+      assertPrimariesBalanced(parentRegionName, serversToStart, false);
+      assertPrimariesBalanced(childRegionName, serversToStart, false);
+      assertPrimariesBalanced(lowRedundancyRegionName, serversToStart, false);
+    }));
+  }
+
+  @Test
+  public void statsAreUpdatedWhenRestoreRedundnacyIsCalled() {
```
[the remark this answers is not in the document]
DonalEvans: I swear I must have made this exact same typo a hundred times when writing this code. My brain just had a problem typing the word redundancy I guess. Thanks for catching it.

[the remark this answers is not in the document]
jujoramos: I didn't catch it, `IntelliJ` did, I won't take the credit  :-P

geode-core/src/main/java/org/apache/geode/cache/control/RestoreRedundancyResults.java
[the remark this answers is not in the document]
jujoramos: Sounds good, thanks for the clarification!.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/control/RestoreRedundancyOperationDUnitTest.java
```diff
@@ -0,0 +1,348 @@
...
+public class RestoreRedundancyOperationDUnitTest {
+  private List<MemberVM> servers;
+  private static final int serversToStart = 3;
+  private static final String parentRegionName = "parentColocatedRegion";
+  private static final String childRegionName = "childColocatedRegion";
+  private static final int desiredRedundancyCopies = 2;
+  private static final String lowRedundancyRegionName = "lowRedundancyRegion";
+  private static final int lowRedundancyCopies = 1;
+  private static final int numBuckets = GLOBAL_MAX_BUCKETS_DEFAULT;
+  private static final int entries = 5 * numBuckets;
```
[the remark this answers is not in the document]
DonalEvans: Good call.

geode-core/src/main/java/org/apache/geode/internal/cache/control/ResourceManagerStats.java
DonalEvans: Thanks for pointing this out, I was unaware of this doc. I'll add the changes to the final PR for this set of changes, which will include all docs updates.

geode-core/src/main/java/org/apache/geode/internal/cache/control/RestoreRedundancyRegionResult.java
DonalEvans: Sure

geode-core/src/upgradeTest/java/org/apache/geode/internal/cache/rollingupgrade/RollingUpgradeRestoreRedundancy.java
DonalEvans: Nope, I was just sloppy when reusing an existing test as a framework and didn't notice it. I've removed it now.

geode-core/src/main/java/org/apache/geode/cache/control/RestoreRedundancyBuilder.java
```diff
@@ -0,0 +1,62 @@
...
+
+  /**
+   * Set whether the restore redundancy operation should reassign primary buckets. The default,
+   * <code>false</code>, will result in primary buckets being reassigned for better load balancing
+   * across members.
+   *
+   * @param shouldNotReassign A boolean indicating whether or not the operation created by this
+   *        class should reassign primary bucket hosts.
+   */
+  RestoreRedundancyBuilder doNotReassignPrimaries(boolean shouldNotReassign);
```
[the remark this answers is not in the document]
DonalEvans: Done

geode-core/src/main/java/org/apache/geode/cache/control/RestoreRedundancyResults.java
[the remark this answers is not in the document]
DonalEvans: After a bit more discussion, it was decided to go with the suggestion here. `SUCCESS` will now only be returned if all regions have their redundancy fully satisfied.

geode-core/src/main/java/org/apache/geode/cache/control/RestoreRedundancyResults.java
```diff
@@ -0,0 +1,148 @@
...
+   *        one.
+   */
+  void addRegionResults(RestoreRedundancyResults results);
+
+  /**
+   * Adds information regarding the number of primaries reassigned and the time taken to reassign
+   * them during a restore redundancy operation.
+   *
+   * @param details a {@link PartitionRebalanceInfo} generated by a
+   *        {@link PartitionedRegionRebalanceOp} operation.
```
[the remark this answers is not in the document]
pivotal-jbarrett: If we had a proper API module these would be a compile time failure for any tests. 
@upthewaterspout Seems like there should be a test in the static analyzer or other that should reject APIs like this.

geode-core/src/main/java/org/apache/geode/cache/control/RestoreRedundancyResults.java
```diff
@@ -0,0 +1,148 @@
...
+   */
+  Map<String, RegionRedundancyStatus> getRegionResults();
+
+  /**
+   * Returns the total number of primaries that were transferred as part of the restore redundancy
+   * operations.
+   *
+   * @return the total number of primaries that were transferred
+   */
+  int getTotalPrimaryTransfersCompleted();
```
pivotal-jbarrett, line 139: Is it reasonable to assume we will never be able to transfer 2^31 or more primaries?

geode-core/src/main/java/org/apache/geode/cache/control/RestoreRedundancyResults.java
```diff
@@ -0,0 +1,148 @@
...
+
+  /**
+   * Returns the total time, in milliseconds, spent transferring primaries as part of the restore
+   * redundancy operations.
+   *
+   * @return the total time, in milliseconds, spent transferring primaries
+   */
+  long getTotalPrimaryTransferTime();
```
pivotal-jbarrett, line 147: We had discussion long ago about making sure time based APIs used `TimeUnit`. This method should be `long getTotalPrimaryTransferTime(TimeUnit unit)`. Alternatively we could consider `java.time.Duration`. Although that does incur an object allocation, given the infrequency that it is called the overhead is probably worth the correctness.

geode-core/src/main/java/org/apache/geode/cache/control/RestoreRedundancyBuilder.java
```diff
@@ -0,0 +1,62 @@
...
+
+  /**
+   * Set whether the restore redundancy operation should reassign primary buckets. The default,
+   * <code>true</code>, will result in primary buckets being reassigned for better load balancing
+   * across members.
+   *
+   * @param shouldReassign A boolean indicating whether or not the operation created by this
+   *        class should reassign primary bucket hosts.
+   */
+  RestoreRedundancyBuilder setReassignPrimaries(boolean shouldReassign);
```
pivotal-jbarrett, line 51: I think `builder.shouldReassignPrimaries(true)` is more consistent with the other builder methods here. Otherwise I would change all to use the `set` prefix. Naming convention consistency is preferable.

geode-core/src/main/java/org/apache/geode/cache/control/RestoreRedundancyBuilder.java
```diff
@@ -0,0 +1,62 @@
...
+
+  /**
+   * Asynchronously starts a new restore redundancy operation. Only the cache resources used by this
+   * member will have redundancy restored. The operation may queue as needed for resources in
+   * contention by other active restore redundancy operations.
+   *
+   * @return a {@link CompletableFuture} which will return the results of the restore redundancy
+   *         operation started by this method.
+   */
+  CompletableFuture<RestoreRedundancyResults> start();
```
pivotal-jbarrett, line 61: A builder with a "start" method? I get that "build" doesn't feel right either. Maybe there is a better name for this class? `¯\_(ツ)_/¯`

geode-core/src/main/java/org/apache/geode/internal/cache/control/InternalResourceManager.java
```diff
@@ -288,20 +294,20 @@ public RebalanceFactory createRebalanceFactory() {
 
   @Override
   public Set<RebalanceOperation> getRebalanceOperations() {
-    synchronized (this.inProgressOperationsLock) {
-      return new HashSet<RebalanceOperation>(this.inProgressOperations);
+    synchronized (this.inProgressRebalanceOperationsLock) {
+      return new HashSet<>(this.inProgressRebalanceOperations);
```
pivotal-jbarrett, line 298: Maybe an unmodifiable set wrapped around a concurrent map?

Also the use of `this.` qualifier is not necessary in this context.

geode-core/src/main/java/org/apache/geode/internal/cache/control/RegionRedundancyStatus.java
```diff
@@ -0,0 +1,123 @@
...
+package org.apache.geode.internal.cache.control;
+
+import java.io.Serializable;
+
+import org.apache.geode.internal.cache.PartitionedRegion;
+
+/**
+ * Used to calculate and store the redundancy status for a {@link PartitionedRegion}.
+ */
+public class RegionRedundancyStatus implements Serializable {
```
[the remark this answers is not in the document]
pivotal-jbarrett: The only time I would say we should have Java Serializable classes is for testing. All other uses for client/server or p2p messaging should be `DataSerializableFixedId`. I wonder if our static analyzer could be updated to catch things like this?

geode-core/src/main/java/org/apache/geode/internal/cache/control/RegionRedundancyStatus.java
```diff
@@ -0,0 +1,123 @@
...
+    desiredRedundancy = region.getRedundantCopies();
+    actualRedundancy = calculateLowestRedundancy(region);
+    status = determineStatus(desiredRedundancy, actualRedundancy);
+  }
+
+  public String getRegionName() {
+    return regionName;
+  }
+
+  public int getActualRedundancy() {
```
pivotal-jbarrett, line 75: Javadoc, what does "Actual Redundancy" mean?

geode-core/src/main/java/org/apache/geode/internal/cache/control/RestoreRedundancyResultsImpl.java
```diff
@@ -0,0 +1,169 @@
...
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+import org.apache.geode.cache.control.RestoreRedundancyResults;
+import org.apache.geode.cache.partition.PartitionRebalanceInfo;
+
+public class RestoreRedundancyResultsImpl implements RestoreRedundancyResults, Serializable {
```
pivotal-jbarrett, line 27: Again with the serialization here.

geode-core/src/main/java/org/apache/geode/cache/control/RestoreRedundancyResults.java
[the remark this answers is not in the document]
pivotal-jbarrett: Perhaps this is the opportunity to start a `geode-control-api.jar` that contains all these interfaces. We have to start somewhere and this seems like a good small start.

geode-core/src/main/java/org/apache/geode/cache/control/RestoreRedundancyResults.java
```diff
@@ -0,0 +1,148 @@
...
+   *        {@link PartitionedRegionRebalanceOp} operation.
+   */
+  void addPrimaryReassignmentDetails(PartitionRebalanceInfo details);
+
+  /**
+   * Adds one {@link RegionRedundancyStatus} to the result set.
+   *
+   * @param regionResult The {@link RegionRedundancyStatus} to be added to the result set.
+   */
+  void addRegionResult(RegionRedundancyStatus regionResult);
```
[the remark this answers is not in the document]
DonalEvans: Good call. I've removed the mutators from the interface and moved `RegionRedundancyStatus` to no longer be internal.

geode-core/src/main/java/org/apache/geode/internal/cache/control/RegionRedundancyStatus.java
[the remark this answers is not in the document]
DonalEvans: Done.

geode-core/src/main/java/org/apache/geode/internal/cache/control/RestoreRedundancyBuilderImpl.java
```diff
@@ -0,0 +1,177 @@
...
+  }
+
+  @Override
+  public RestoreRedundancyBuilder setReassignPrimaries(boolean shouldReassign) {
+    this.shouldReassign = shouldReassign;
+    return this;
+  }
+
+  @Override
+  public CompletableFuture<RestoreRedundancyResults> start() {
```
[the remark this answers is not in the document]
DonalEvans: If child colocated regions are automatically rejected, then if we explicitly include a child region in the restore redundancy operation but not its parent, no work will be done. This could make life difficult for users who might not know which region in a colocation group is the parent, so the behaviour implemented here is to restore redundancy to all regions in the colocation group if any one of them is included, regardless of whether it's the parent or not.

The missing logic is also not strictly needed here, since if we call `execute()` on a `PartitionedRegionRebalanceOp` for a colocated region, we will either rebalance the entire colocation group (if it's the first region in the group to be executed on) or we'll exit early because `isRebalanceNecessary()` will return false (if any of the other regions in the group have already been executed on), so no additional work is being done at the region level.

The design for this class was originally much closer to what is seen in `RebalanceFactoryImpl`, but following feedback on the RFC in which several people requested that the `start()` method return a `CompletableFuture` rather than a separate Operation class, this approach was taken. I agree that it would be better to have both operations use the same logic, but `RebalanceOperation` and `RebalanceFactory` are both public APIs and so can't be easily changed, and there was a strong agreement on the RFC to use a `CompletableFuture` rather than the existing approach, so I'm not sure it's possible to make everyone happy here.

DonalEvans:
> it seems like there are several classes that are a tweaked copy of existing classes - RestoreRedundancyBuilderImpl, RestoreRedundancyDirector, RegionRedundancyStatus. Is there a way we could extend/generalize the existing classes rather than copy/paste/tweak?

I've modified `CompositeDirector` slightly to avoid the need for the `RestoreRedundancyDirector` class, but as described in my comment above, doing the same for `RestoreRedundancyBuilderImpl` would be very difficult at this point. As for `RegionRedundancyStatus`; it's fundamentally very different from `RebalanceResults` in terms of the information it contains, since it represents a snapshot of the state of a region rather than the work done to get it to that state (unless I've missed something here and there is another class that already exists and is similar to `RegionRedundancyStatus`).

geode-core/src/main/java/org/apache/geode/cache/control/RegionRedundancyStatus.java
```diff
@@ -0,0 +1,158 @@
...
+import org.apache.geode.internal.cache.PartitionedRegion;
+import org.apache.geode.internal.serialization.DataSerializableFixedID;
+import org.apache.geode.internal.serialization.DeserializationContext;
+import org.apache.geode.internal.serialization.SerializationContext;
+import org.apache.geode.internal.serialization.Version;
+
+/**
+ * Used to calculate and store the redundancy status for a {@link PartitionedRegion}.
+ */
+public class RegionRedundancyStatus implements DataSerializableFixedID {
```
upthewaterspout, line 31: I think this is still leaking internal classes (in this case DataSerializableFixedID). This probably needs another (read only) interface as part of the public API and an internal implementation class.

geode-core/src/main/java/org/apache/geode/cache/control/RestoreRedundancyResults.java
jhuynh1: I would hope we don't have to transfer that many primaries... ever...Can we even handle that many buckets (not theoretically but do we know of any cases where we have over 100 million buckets❓)

DonalEvans: The most primaries that could possibly be transferred for a single region during a restore redundancy operation is equal to the number of buckets in the region (this scenario would also require there to be at least as many servers as there are buckets). Assuming the default number of buckets (113) is used, there would need to be a little over 19 million regions for an integer overflow to occur, which seems like an extremely unlikely scenario.

geode-core/src/main/java/org/apache/geode/internal/cache/control/InternalResourceManager.java
DonalEvans: Done

geode-core/src/main/java/org/apache/geode/internal/cache/control/RegionRedundancyStatus.java
DonalEvans: Updated with a description. The actual redundancy is the number of redundant copies that existed when the `RegionRedundancyStatus` was created, which may be different from the configured desired redundancy.

geode-core/src/main/java/org/apache/geode/internal/cache/control/RestoreRedundancyResultsImpl.java
DonalEvans: Fixed

geode-core/src/main/java/org/apache/geode/cache/control/RestoreRedundancyBuilder.java
DonalEvans: Done

geode-core/src/main/java/org/apache/geode/cache/control/RestoreRedundancyResults.java
DonalEvans: Done

geode-core/src/main/java/org/apache/geode/cache/control/RegionRedundancyStatus.java
DonalEvans: Thanks for catching this

geode-core/src/main/java/org/apache/geode/cache/control/ResourceManager.java
```diff
@@ -77,6 +78,21 @@
    */
   Set<RebalanceOperation> getRebalanceOperations();
 
+  /**
+   * Creates a class for defining and starting restore redundancy operations.
```
upthewaterspout, line 82: I think we should improve these javadocs and maybe clarify the difference between "restore redundancy" and rebalance. Restore redundancy is really just a subset of the rebalance operation - there is no need to invoke both, for example.

geode-serialization/src/main/java/org/apache/geode/internal/serialization/DataSerializableFixedID.java
```diff
@@ -225,7 +225,10 @@
 
   byte REMOVE_CACHESERVER_PROFILE_UPDATE = -19;
 
-  // IDs -18 through -10 unused
+  byte REGION_REDUNDANCY_STATUS = -18;
+  byte RESTORE_REDUNDANCY_RESULTS = -17;
```
upthewaterspout, line 229: Nitpick - but maybe we should save these small message ids for more frequent messages?

DonalEvans: Sounds sensible

geode-core/src/main/java/org/apache/geode/cache/control/ResourceManager.java
DonalEvans: How would this be:

PR #4912, opened 2020-04-07
Decouple geode in sadd,srem,smembers

prettyClouds:
This commit builds a GeodeRedisSetSynchronized class that brings
together "set" methods.  These three are necessary for Spring-Session so
they are the only ones included.

Hey team,

The commit brings together two ideas in this PR:

1.  It synchronizes all calls to the set (at least sadd,srem,smembers) via compute.
2. It decouples all geode related code from the Executors.  We have discussed this a couple times, but I wanted to show it, and see if you all are on board.

### DIFF
--- a/geode-redis/src/main/java/org/apache/geode/redis/GeodeRedisServer.java
+++ b/geode-redis/src/main/java/org/apache/geode/redis/GeodeRedisServer.java
@@ -306,7 +306,7 @@ public class GeodeRedisServer {
    * @return {@link RegionShortcut}
    */
   private static RegionShortcut setRegionType() {
-    String regionType = System.getProperty(DEFAULT_REGION_SYS_PROP_NAME, "PARTITION");
+    String regionType = System.getProperty(DEFAULT_REGION_SYS_PROP_NAME, "REPLICATE");
     RegionShortcut type;
     try {
       type = RegionShortcut.valueOf(regionType);
--- a/geode-redis/src/main/java/org/apache/geode/redis/GeodeRedisServer.java
+++ b/geode-redis/src/main/java/org/apache/geode/redis/GeodeRedisServer.java
@@ -306,7 +306,7 @@ public class GeodeRedisServer {
    * @return {@link RegionShortcut}
    */
   private static RegionShortcut setRegionType() {
-    String regionType = System.getProperty(DEFAULT_REGION_SYS_PROP_NAME, "REPLICATE");
+    String regionType = System.getProperty(DEFAULT_REGION_SYS_PROP_NAME, "PARTITION");
     RegionShortcut type;
     try {
       type = RegionShortcut.valueOf(regionType);
--- a/geode-redis/src/main/java/org/apache/geode/redis/internal/executor/hash/GeodeRedisHashSynchronized.java
+++ b/geode-redis/src/main/java/org/apache/geode/redis/internal/executor/hash/GeodeRedisHashSynchronized.java
@@ -39,7 +39,7 @@ class GeodeRedisHashSynchronized implements RedisHash {
 
   @Override
   public int hset(List<ByteArrayWrapper> fieldsToSet,
-      boolean NX) {
+                  boolean NX) {
     AtomicInteger fieldsAdded = new AtomicInteger();
 
     Map<ByteArrayWrapper, ByteArrayWrapper> computedHash =
@@ -57,18 +57,17 @@ class GeodeRedisHashSynchronized implements RedisHash {
             ByteArrayWrapper field = fieldsToSet.get(i);
             ByteArrayWrapper value = fieldsToSet.get(i + 1);
 
-            Object abc;
             if (NX) {
-              abc = newHash.putIfAbsent(field, value);
+              newHash.putIfAbsent(field, value);
             } else {
-              abc = newHash.put(field, value);
-            }
-
-            if (abc == null) {
-              fieldsAdded.getAndIncrement();
+              newHash.put(field, value);
             }
           }
-
+          if (oldHash == null) {
+            fieldsAdded.set(newHash.size());
+          } else {
+            fieldsAdded.set(newHash.size() - oldHash.size());
+          }
           return newHash;
         });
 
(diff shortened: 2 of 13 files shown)


geode-redis/src/main/java/org/apache/geode/redis/GeodeRedisServer.java
```diff
@@ -306,7 +306,7 @@
    * @return {@link RegionShortcut}
    */
   private static RegionShortcut setRegionType() {
-    String regionType = System.getProperty(DEFAULT_REGION_SYS_PROP_NAME, "PARTITION");
+    String regionType = System.getProperty(DEFAULT_REGION_SYS_PROP_NAME, "REPLICATE");
```
upthewaterspout, line 309: Why the change to REPLICATE region type as the default? Should it be PARTITION_REDUNDANT, perhaps?

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/hash/GeodeRedisHashSynchronized.java
```diff
@@ -0,0 +1,113 @@
...
+          numDeleted.incrementAndGet();
+        }
+      }
+      return newHash;
+    });
+    return numDeleted.intValue();
+  }
+
+  @Override
+  public Collection<Map.Entry<ByteArrayWrapper, ByteArrayWrapper>> hgetall() {
```
upthewaterspout, line 99: Because you set methods are never modifying the old value in place, I think would be safe just to return the result of region.get() here, without the compute or the copying.

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/hash/GeodeRedisHashSynchronized.java
```diff
@@ -0,0 +1,113 @@
...
+
+  @Override
+  public int hset(List<ByteArrayWrapper> fieldsToSet,
+      boolean NX) {
+    AtomicInteger fieldsAdded = new AtomicInteger();
+
+    Map<ByteArrayWrapper, ByteArrayWrapper> computedHash =
+        region().compute(key, (_unused_, oldHash) -> {
+
+          fieldsAdded.set(0);
```
upthewaterspout, line 48: Might want to comment about why you called fieldsAdded.set(0) (because compute can be called multiple times).

The logic you used in the set methods to just call fieldsAdded.set(newHash.size() - oldHash.size()) might be less confusing to someone reading the code.

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/hash/GeodeRedisHashSynchronized.java
```diff
@@ -0,0 +1,113 @@
...
+    if (computedHash != null) {
+      context.getKeyRegistrar().register(this.key, RedisDataType.REDIS_HASH);
+    }
+
+    return fieldsAdded.get();
+  }
+
+  @Override
+  public int hdel(List<ByteArrayWrapper> subList) {
+    AtomicLong numDeleted = new AtomicLong();
```
upthewaterspout, line 84: Need to set numDeleted to 0 at the beginning of the compute block.

geode-redis/src/main/java/org/apache/geode/redis/GeodeRedisServer.java
prettyClouds: i was just messing around...i'll rebase this change out of the PR

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/hash/GeodeRedisHashSynchronized.java
prettyClouds: i was concerned about a CME if somebody iterates over the returned value...i think the copy is necessary but let me know.

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/hash/GeodeRedisHashSynchronized.java
prettyClouds: yea...i only caught that because there was luckily a test that was counting the return values. i can add something.

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/hash/GeodeRedisHashSynchronized.java
prettyClouds: good catch...no test for this, so i missed it.  you are a human junit...DanUnit

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/hash/GeodeRedisHashSynchronized.java
prettyClouds: i changed the logic...i was hesitant to run .size(), but it's not a concurrenthashmap, so it should be O(1) and consistent

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/hash/GeodeRedisHashSynchronized.java
prettyClouds: fixed...changed the logic to call AtomicLong.set(<value>) like you recommended

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/hash/GeodeRedisHashSynchronized.java
upthewaterspout: Well, the function you pass into computeIfPresent is also technically iterating over the oldHash. And since the function is not necessarily executed under lock if anything changes oldHash the function could also through a CME. The compute* methods don't necessarily lock while executing your function, they could be doing a replace followed by a retry of your function for a concurrent modification exception.

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/set/GeodeRedisSetSynchronized.java
```diff
@@ -0,0 +1,86 @@
...
+import org.apache.geode.cache.Region;
+import org.apache.geode.redis.internal.ByteArrayWrapper;
+import org.apache.geode.redis.internal.ExecutionHandlerContext;
+
+class GeodeRedisSetSynchronized implements RedisSet {
+
+  private ByteArrayWrapper key;
+  private ExecutionHandlerContext context;
+
+  public GeodeRedisSetSynchronized(ByteArrayWrapper key, ExecutionHandlerContext context) {
```
jhutchison, line 32: I would suggest a refactor to the name here-  this class seems like it's more about operating on sets than a container for data?

PR #4913, opened 2020-04-07
GEODE-7943 add synchronization to Subscriptions class

jdeppe-pivotal:

### For all changes:

### Note:

### DIFF
--- a/geode-redis/src/main/java/org/apache/geode/redis/internal/Subscriptions.java
+++ b/geode-redis/src/main/java/org/apache/geode/redis/internal/Subscriptions.java
@@ -32,7 +32,7 @@ public class Subscriptions {
    * @param channelOrPattern channel or pattern
    * @param client a client connection
    */
-  public boolean exists(Object channelOrPattern, Client client) {
+  public synchronized boolean exists(Object channelOrPattern, Client client) {
     return subscriptions.stream()
         .anyMatch(subscription -> subscription.isEqualTo(channelOrPattern, client));
   }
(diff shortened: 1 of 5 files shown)


jdeppe-pivotal:
@ringles @sabbeyPivotal @xtreme-prasath @prettyClouds @jhutchison - for your review

geode-redis/src/main/java/org/apache/geode/redis/internal/Subscriptions.java
```diff
@@ -55,30 +55,37 @@ public boolean exists(Object channelOrPattern, Client client) {
...
-  public void remove(Object channel, Client client) {
-    this.subscriptions.removeIf(subscription -> subscription.isEqualTo(channel, client));
+  public synchronized void remove(Object channel, Client client) {
+    subscriptions.removeIf(subscription -> subscription.isEqualTo(channel, client));
+  }
+
+  /**
+   * @return the total number of all local subscriptions
+   */
+  public int size() {
```
upthewaterspout, line 88: Technically I think size() needs to be synchronized as well.

jdeppe-pivotal: I've switched to using a `CopyOnWriteArrayList`.

jdeppe-pivotal:
@prettyClouds Good idea! I've switched to using that.

jdeppe-pivotal:
@dschneider-pivotal - Is there any underlying technical difference? I had played around with using a `Vector` which, I think, is practically the same thing. The 'lookup' methods still need to have explicit synchronization although the `add` and `remove` don't. Since there is only a single instance of this class in play, having the synchronization at the method level vs at the level of the backing structure also does not make any difference. Although I could buy the argument that that is just how it is currently being used. I think I just convinced myself... :)

dschneider-pivotal:
why is subscriptionIntegrationTest running out of memory?

PR #4924, opened 2020-04-07
GEODE-7964: Upgrade Mockito to 3.3.3

kirklund:
GEODE-7964: Upgrade Mockito to 3.3.3

Fixup all tests touched for Mockito upgrade.
Update usage of Mockito deprecated APIs.

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/cache/client/internal/ConnectionImpl.java
+++ b/geode-core/src/main/java/org/apache/geode/cache/client/internal/ConnectionImpl.java
@@ -224,6 +224,16 @@ public class ConnectionImpl implements Connection {
   }
 
   @Override
+  public long getBirthDate() {
+    return 0; // TODO:KIRK
+  }
+
+  @Override
+  public void setBirthDate(long ts) {
+    // TODO:KIRK
+  }
+
+  @Override
   public OutputStream getOutputStream() {
     return out;
   }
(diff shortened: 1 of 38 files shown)


kirklund:
Note: 2.23.4 and 2.24.0 passed IntegrationTest. 2.24.5 started to see OOME in integrationTest. First OOME was HARegionQueueIntegrationTest and then the *ConcurrentTests using Mockito.

HARegionQueueIntegrationTest was fixed by adding these lines to tearDown:
```
Mockito.framework().clearInlineMocks();
Mockito.reset();
```

And using stubOnly when creating mocks:
```
Mockito.mock(classToMock, withSettings().stubOnly());
```

kirklund:
DistributedTest failure is JMXMBeanReconnectDUnitTest

kirklund:
The Concurrent Tests that don't use Mockito still seem to hang in StressNewTest (I removed `@Ignore` so they now need to pass StressNewTest).

kirklund:
Another unrelated JMXMBeanReconnectDUnitTest failure.

geode-core/src/main/java/org/apache/geode/cache/client/internal/ConnectionImpl.java
```diff
@@ -223,6 +223,16 @@ public Socket getSocket() {
     return theSocket;
   }
 
+  @Override
+  public long getBirthDate() {
+    return 0; // TODO:KIRK
```
upthewaterspout, line 228: Clean up these TODO:KIRK parts.

kirklund: Thanks for catching those!

kirklund: Removed

PR #4934, opened 2020-04-09
GEODE-7940: Fix Tracking of ShadowBuckets Destroyed

jujoramos:
The BucketAdvisor can now keep track of more than just one shadow
bucket to avoid incorrectly marking all of them as destroyed.

- Added unit and distributed tests.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
@@ -701,17 +701,17 @@ public class ParallelGatewaySenderQueue implements RegionQueue {
     AbstractBucketRegionQueue brq =
         (AbstractBucketRegionQueue) prQ.getDataStore().getLocalBucketById(bucketId);
 
+    // Full path of the bucket:
+    final String bucketFullPath =
+        Region.SEPARATOR + PartitionedRegionHelper.PR_ROOT_REGION_NAME + Region.SEPARATOR
+            + prQ.getBucketName(bucketId);
+
     try {
       if (brq == null) {
         // Set the threadInitLevel to BEFORE_INITIAL_IMAGE.
         final InitializationLevel oldLevel =
             LocalRegion.setThreadInitLevelRequirement(BEFORE_INITIAL_IMAGE);
         try {
-          // Full path of the bucket:
-
-          final String bucketFullPath =
-              Region.SEPARATOR + PartitionedRegionHelper.PR_ROOT_REGION_NAME + Region.SEPARATOR
-                  + prQ.getBucketName(bucketId);
 
           brq = (AbstractBucketRegionQueue) prQ.getCache().getInternalRegionByPath(bucketFullPath);
           if (isDebugEnabled) {
@@ -739,8 +739,8 @@ public class ParallelGatewaySenderQueue implements RegionQueue {
             // chain is getting destroyed one by one starting from child region
             // i.e this bucket due to moveBucket operation
             // In that case we don't want to store this event.
-            if (((PartitionedRegion) prQ.getColocatedWithRegion()).getRegionAdvisor()
-                .getBucketAdvisor(bucketId).getShadowBucketDestroyed()) {
+            if (prQ.getColocatedWithRegion().getRegionAdvisor().getBucketAdvisor(bucketId)
+                .isShadowBucketDestroyed(bucketFullPath)) {
               if (isDebugEnabled) {
                 logger.debug(
                     "ParallelGatewaySenderOrderedQueue not putting key {} : Value : {} as shadowPR bucket is destroyed.",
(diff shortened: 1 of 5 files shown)


jujoramos:
@DonalEvans : good observation, I don't think we ever create that many shadow buckets as to introduce a memory leak with this map; _**but**_ is a valid concern nevertheless... I will go over the code again and check whether we can change the way we track shadow buckets destroyed. Thanks for catching this.

jujoramos:
@DonalEvans

I've been digging through the code and we don't track created shadow buckets at the bucket level, we only track the entire shadow `PartitionRegion` instead... adding a fine-grained tracking per bucket doesn't seem to be a quick, nor easy, task. It would also imply getting the full list of buckets per region and, for each bucket, find every related shadow bucket and register it in the relevant `BucketAdvisor` upon creation, implying more messaging and delays.
Instead, I've modified the `BucketAdvisor` to only keep track of known destroyed buckets, returning `false` by default (as before) if we don't have a record for that particular shadow bucket. The `markAllShadowBucketsAsNonDestroyed` now clears the `destroyedShadowBuckets` map instead of changing the boolean flag to `false` as well. Another thing to notice, according to what I see, is that there's only one `BucketAdvisor` per bucket and, in the general use case, there will be only one shadow bucket per bucket -> meaning that the `Map` itself will contain only one entry (if any) at any given time. For those cases where a single `Region` has multiple `gateway-senders`, the `Map` will have at most N entries (if any), being N the amount of `gateway-senders` attached to the region (the mapping `1 shadow bucket per bucket` becomes `N shadow buckets per bucket`).
I might be missing something, though, and I know you've been working on the redundancy stuff (highly related to PR regions and bucket movement) _**so we can certainly discuss if you think otherwise**_.
Please let me know your thoughts!.

DonalEvans:
@jujoramos The proposed change you have of `markAllShadowBucketsAsNonDestroyed` clearing the map sounds good to me.

geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
```diff
@@ -790,17 +790,17 @@ public boolean put(Object object) throws InterruptedException, CacheException {
           LocalRegion.setThreadInitLevelRequirement(oldLevel);
         }
       } else {
-        boolean thisbucketDestroyed = false;
+        boolean thisBucketDestroyed;
```
jhuynh1, line 793: just a thought, would this simplify to :
boolean thisBucketDestroyed = brq.isDestroyed();
if (!isDREvent) {
  thisBucketDestroyed |= prQ.getColocatedWithRegion()...;
}

geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
```diff
@@ -701,17 +701,17 @@ public boolean put(Object object) throws InterruptedException, CacheException {
     AbstractBucketRegionQueue brq =
         (AbstractBucketRegionQueue) prQ.getDataStore().getLocalBucketById(bucketId);
 
+    // Full path of the bucket:
+    final String bucketFullPath =
```
jhuynh1, line 705: Pulling this out will always construct this string now.  Any ideas on perf impact?

geode-core/src/test/java/org/apache/geode/internal/cache/BucketAdvisorTest.java
```diff
@@ -150,4 +155,42 @@ public void volunteerForPrimaryIgnoresMissingPrimaryElector() {
...
+    // Return false by default.
+    assertThat(bucketAdvisor.isShadowBucketDestroyed("/bucket")).isFalse();
+
+    // Return correct value when found.
+    bucketAdvisor.markShadowBucketAsDestroyed(shadowBuckets.get(1));
+    assertThat(bucketAdvisor.isShadowBucketDestroyed(shadowBuckets.get(1))).isTrue();
+
+    // Mark all shadow buckets values as destroyed
+    bucketAdvisor.markAllShadowBucketsAsDestroyed();
+    shadowBuckets
```
jhuynh1, line 188: No change needed, but there might be a way to do this with a stream filter instead .

geode-core/src/test/java/org/apache/geode/internal/cache/BucketAdvisorTest.java
```diff
@@ -150,4 +155,42 @@ public void volunteerForPrimaryIgnoresMissingPrimaryElector() {
...
+        .forEach(b -> assertThat(assertThat(bucketAdvisor.isShadowBucketDestroyed(b)).isTrue()));
+
+    // Mark all shadow buckets values as non destroyed
+    bucketAdvisor.markAllShadowBucketsAsNonDestroyed();
```
jhuynh1, line 192: maybe break this into two tests, one just for marking?  it might seem redundant so it's up to you

geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
jujoramos: Changed!.

geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
jujoramos: Good point, I haven't ran any performance tests related to the change.
I've refactored the first part of the `String` as a constant and moved the initialisation _**only**_ to places where is needed, hopefully this will reduce the impact. Let me know if you think that's enough.

geode-core/src/test/java/org/apache/geode/internal/cache/BucketAdvisorTest.java
jujoramos: 👍

geode-core/src/test/java/org/apache/geode/internal/cache/BucketAdvisorTest.java
jujoramos: Changed!

PR #4936, opened 2020-04-09
GEODE-7851: Pulse Oauth Support

demery-pivotal:
- create an OauthSecurityConfig to configure spring using oauth
- add PULSE as an oauth-enabled-component, and if pulse is set to use
  oauth, set the OauthSecurityConfig as the active security profile
- use pulse.properties in the locator's working dir to externalize pulse
  authentication provider configuration

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/management/internal/ManagementAgent.java
+++ b/geode-core/src/main/java/org/apache/geode/management/internal/ManagementAgent.java
@@ -196,7 +196,6 @@ public class ManagementAgent {
       }
     } else {
       String pwFile = this.config.getJmxManagerPasswordFile();
-      String springActiveProfile = "spring.profiles.active";
       if (securityService.isIntegratedSecurity()) {
         String[] authTokenEnabledComponents = config.getSecurityAuthTokenEnabledComponents();
         boolean pulseOauth = Arrays.stream(authTokenEnabledComponents)
--- a/geode-core/src/main/java/org/apache/geode/management/internal/ManagementAgent.java
+++ b/geode-core/src/main/java/org/apache/geode/management/internal/ManagementAgent.java
@@ -196,6 +196,7 @@ public class ManagementAgent {
       }
     } else {
       String pwFile = this.config.getJmxManagerPasswordFile();
+      String springActiveProfile = "spring.profiles.active";
       if (securityService.isIntegratedSecurity()) {
         String[] authTokenEnabledComponents = config.getSecurityAuthTokenEnabledComponents();
         boolean pulseOauth = Arrays.stream(authTokenEnabledComponents)
--- a/geode-pulse/src/main/java/org/apache/geode/tools/pulse/internal/security/LogoutHandler.java
+++ b/geode-pulse/src/main/java/org/apache/geode/tools/pulse/internal/security/LogoutHandler.java
@@ -26,7 +26,6 @@ import org.springframework.beans.BeansException;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationContextAware;
 import org.springframework.security.core.Authentication;
-import org.springframework.security.web.authentication.logout.LogoutSuccessHandler;
 import org.springframework.security.web.authentication.logout.SimpleUrlLogoutSuccessHandler;
 
 import org.apache.geode.tools.pulse.internal.data.Repository;
@@ -34,7 +33,7 @@ import org.apache.geode.tools.pulse.internal.data.Repository;
 /**
  * Handler is used to close jmx connection maintained at user-level
  */
-public class LogoutHandler extends SimpleUrlLogoutSuccessHandler implements LogoutSuccessHandler,
+public class LogoutHandler extends SimpleUrlLogoutSuccessHandler implements
     ApplicationContextAware {
   private static final Logger logger = LogManager.getLogger();
   private ApplicationContext applicationContext;
--- a/geode-pulse/src/main/java/org/apache/geode/tools/pulse/internal/security/LogoutHandler.java
+++ b/geode-pulse/src/main/java/org/apache/geode/tools/pulse/internal/security/LogoutHandler.java
@@ -22,6 +22,9 @@ import javax.servlet.http.HttpServletResponse;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.springframework.beans.BeansException;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.ApplicationContextAware;
 import org.springframework.security.core.Authentication;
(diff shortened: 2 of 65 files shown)


demery-pivotal:
@pnikonowicz please review

geode-pulse/src/main/java/org/apache/geode/tools/pulse/internal/security/LogoutHandler.java
```diff
@@ -30,24 +33,31 @@
 
 /**
  * Handler is used to close jmx connection maintained at user-level
- *
  */
-public class LogoutHandler extends SimpleUrlLogoutSuccessHandler implements LogoutSuccessHandler {
+public class LogoutHandler extends SimpleUrlLogoutSuccessHandler implements LogoutSuccessHandler,
```
jinmeiliao, line 37: I believe you don't need to implement `LogoutSuccessHandler` since `SimpleUrlLogoutSuccessHandler` already implements that

you don't need to implement `ContextAware` either, Spring should take care of auto-wiring for you if you declare
<code>
@Autowired
  private Repository repository;
<code>

geode-pulse/src/main/java/org/apache/geode/tools/pulse/internal/security/LogoutHandler.java
```diff
@@ -30,24 +33,31 @@
...
+    ApplicationContextAware {
   private static final Logger logger = LogManager.getLogger();
+  private ApplicationContext applicationContext;
```
jinmeiliao, line 40:
```suggestion
  @Autowired
  private Repository repository;
```

geode-pulse/src/main/java/org/apache/geode/tools/pulse/internal/service/ClusterDetailsService.java
```diff
@@ -44,14 +45,20 @@
 public class ClusterDetailsService implements PulseService {
 
   private final ObjectMapper mapper = new ObjectMapper();
+  private final Repository repository;
```
jinmeiliao, line 48:
```suggestion
  @Autowired
  private Repository repository;
```

geode-pulse/src/main/java/org/apache/geode/tools/pulse/internal/service/ClusterDetailsService.java
```diff
@@ -44,14 +45,20 @@
...
+
+  @Autowired
```
jinmeiliao, line 50: no need for this constructor if Repository variable is declared as `autowired` already.

Same should be the same for all the service classes.

geode-core/src/main/java/org/apache/geode/management/internal/ManagementAgent.java
```diff
@@ -196,6 +196,7 @@ private void loadWebApplications() {
       }
     } else {
       String pwFile = this.config.getJmxManagerPasswordFile();
+      String springActiveProfile = "spring.profiles.active";
```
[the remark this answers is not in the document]
demery-pivotal: I've removed it.

geode-pulse/src/main/java/org/apache/geode/tools/pulse/internal/security/LogoutHandler.java
demery-pivotal: I've removed the unnecessary `implements LogoutSuccessHandler`.

Removing `ContextAware` is trickier. Unfortunately, simply autowiring the repository this way causes a circular dependency that Spring can't resolve.

In my logout spike, I have fixed this by untangling the circular dependency. So here are the options I can think of:

1. Move the dependency-untangling code from my logout spike into this PR. This involves changes to 8 or 9 files.
2. Leave `LogoutHandler` implementing `ContextAware` for now, and fix it in a later PR.

Which would you prefer?

geode-pulse/src/main/java/org/apache/geode/tools/pulse/internal/service/ClusterDetailsService.java
demery-pivotal: How strongly do you feel about this? The general (strong) consensus is to avoid field injection. And we've found that constructor injection makes it way easier to write unit tests (if we ever get around to writing unit tests).

geode-pulse/src/main/java/org/apache/geode/tools/pulse/internal/security/LogoutHandler.java
jinmeiliao: we can leave it for now

geode-pulse/src/main/java/org/apache/geode/tools/pulse/internal/service/ClusterDetailsService.java
jinmeiliao: I only prefer field injection because it's less code. If this makes testing easier, I am OK with it.

geode-pulse/src/main/java/org/apache/geode/tools/pulse/internal/security/OAuthSecurityConfig.java
```diff
@@ -84,16 +107,15 @@ public OAuth2AuthorizedClientRepository authorizedClientRepository(
...
         .redirectUriTemplate("{baseUrl}/login/oauth2/code/{registrationId}")
-        .scope("openid", "CLUSTER:READ", "CLUSTER:WRITE", "DATA:READ", "DATA:WRITE")
+        .clientId(clientId)
+        .clientSecret(clientSecret)
         .authorizationUri(authorizationUri)
         .tokenUri(tokenUri)
         .userInfoUri(userInfoUri)
         .jwkSetUri(jwkSetUri)
-        .clientName("Pulse")
+        .clientName(providerName)
```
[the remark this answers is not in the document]
demery-pivotal: Despite the name of the method `clientName()`, the value really, really, really does name the provider. If you look at the `ClientRegistration` code (line 561), you'll see that the default value for "client name" is `registrationId` (aka provider id).

Given that, we have some options:

1. Name the property pulse.oauth.clientName. I think this name is very confusing for whoever writes the properties file.
2. Name the property pulse.oauth.providerName. This name is clearer for whoever writes the properties file, and is a bit of a puzzle for a Geode developer reading the `OAuthSecurityConfig` code. We could add a comment explaining the puzzle.
3. Name the property pulse.oauth.providerName, but name the field `clientName`. Then this method call becomes clearer. But then a Geode developer may wonder why the field name differs from the property name. Again, we could explain with a comment.

Which do you prefer?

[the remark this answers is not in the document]
jinmeiliao: 2 with a comment probably would be better.

PR #4941, opened 2020-04-10
 GEODE-7979: Implement a test class for the PERSIST Redis API command

sabbeyPivotal:
The Redis PERSIST command removes the existing timeout on a key (https://redis.io/commands/persist).  The PERSIST command was previously implemented, but not tested. This PR adds tests for this command.

### DIFF
--- a/geode-redis/src/acceptanceTest/java/org/apache/geode/redis/PersistDockerAcceptanceTest.java
+++ b/geode-redis/src/acceptanceTest/java/org/apache/geode/redis/PersistDockerAcceptanceTest.java
@@ -23,12 +23,12 @@ import org.junit.rules.TestRule;
 import org.testcontainers.containers.GenericContainer;
 import redis.clients.jedis.Jedis;
 
-import org.apache.geode.redis.general.ExpireIntegrationTest;
+import org.apache.geode.redis.general.PersistIntegrationTest;
 import org.apache.geode.test.junit.categories.RedisTest;
 import org.apache.geode.test.junit.rules.IgnoreOnWindowsRule;
 
 @Category({RedisTest.class})
-public class ExpireDockerAcceptanceTest extends ExpireIntegrationTest {
+public class PersistDockerAcceptanceTest extends PersistIntegrationTest {
 
   private static GenericContainer redisContainer;
 
@@ -42,10 +42,11 @@ public class ExpireDockerAcceptanceTest extends ExpireIntegrationTest {
     redisContainer = new GenericContainer<>("redis:5.0.6").withExposedPorts(6379);
     redisContainer.start();
     jedis = new Jedis("localhost", redisContainer.getFirstMappedPort(), REDIS_CLIENT_TIMEOUT);
+    jedis2 = new Jedis("localhost", redisContainer.getFirstMappedPort(), REDIS_CLIENT_TIMEOUT);
   }
 
   @AfterClass
-  public static void classLevelTearDown() {
+  public static void tearDown() {
     jedis.close();
   }
 
--- a/geode-redis/src/integrationTest/java/org/apache/geode/redis/general/PersistIntegrationTest.java
+++ b/geode-redis/src/integrationTest/java/org/apache/geode/redis/general/PersistIntegrationTest.java
@@ -1,0 +1,216 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package org.apache.geode.redis.general;
+
+import static org.apache.geode.distributed.ConfigurationProperties.LOCATORS;
+import static org.apache.geode.distributed.ConfigurationProperties.LOG_LEVEL;
+import static org.apache.geode.distributed.ConfigurationProperties.MCAST_PORT;
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.concurrent.atomic.AtomicLong;
+
+import org.junit.After;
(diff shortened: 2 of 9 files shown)


geode-redis/src/integrationTest/java/org/apache/geode/redis/general/PersistIntegrationTest.java
```diff
@@ -0,0 +1,214 @@
...
+import org.apache.geode.cache.CacheFactory;
+import org.apache.geode.cache.GemFireCache;
+import org.apache.geode.internal.AvailablePortHelper;
+import org.apache.geode.redis.GeodeRedisServer;
+
+public class PersistIntegrationTest {
+
+  public static Jedis jedis;
+  public static Jedis jedis2;
+  public static int REDIS_CLIENT_TIMEOUT = 10000000;
```
prettyClouds, line 41: why is this so high?  do we even need to set this below, or can we rely on the default?

geode-redis/src/integrationTest/java/org/apache/geode/redis/general/PersistIntegrationTest.java
```diff
@@ -0,0 +1,214 @@
...
+  private static GeodeRedisServer server;
+  private static GemFireCache cache;
+  private static int ITERATION_COUNT = 5000;
```
prettyClouds, line 44: i think this should be a local variable to the concurrent test, not a global to this whole class.

geode-redis/src/integrationTest/java/org/apache/geode/redis/general/PersistIntegrationTest.java
```diff
@@ -0,0 +1,214 @@
...
+
+  @Test
+  public void shouldReturnZero_givenKeyDoesNotExist() {
+    assertThat(jedis.persist("key")).isEqualTo(0L);
+  }
+
+  @Test
+  public void shouldPersistKey_givenKeyWithSetValue() {
+    String key = "key";
+    String value = "value";
```
prettyClouds, line 92: can you rename this to element or set_element...it is really hard to distinguish this test from the one above without looking VERY carefully

geode-redis/src/integrationTest/java/org/apache/geode/redis/general/PersistIntegrationTest.java
```diff
@@ -0,0 +1,214 @@
...
+
+    thread1.start();
+    thread2.start();
+    thread1.join();
+    thread2.join();
+
+    assertThat(persistedFromThread1.get() + persistedFromThread2.get()).isEqualTo(ITERATION_COUNT);
+  }
+
+  private void doABunchOfSetEXsWithBlockingQueue(Jedis jedis) {
```
prettyClouds, line 199: where's the blocking queue here?  this name may be a relic, and needs to be updated?

"do a bunch of" is odd.  maybe the method is just expireKeys, and  below one is persistKeys

geode-redis/src/test/java/org/apache/geode/redis/internal/executor/general/PersistExecutorJUnitTest.java
```diff
@@ -0,0 +1,91 @@
...
+public class PersistExecutorJUnitTest {
+
+  private ExecutionHandlerContext context;
+  private Command command;
+  private UnpooledByteBufAllocator byteBuf;
+
+  @Before
+  public void setUp() {
+    context = mock(ExecutionHandlerContext.class);
+    command = mock(Command.class);
```
prettyClouds, line 48: I think you should use the actual Command class and not a mock.  I tried it out and the tests are simpler. Give it a try and see what you think.

geode-redis/src/acceptanceTest/java/org/apache/geode/redis/PersistDockerAcceptanceTest.java
```diff
@@ -0,0 +1,45 @@
...
+import org.junit.BeforeClass;
+import org.junit.experimental.categories.Category;
+import org.testcontainers.containers.GenericContainer;
+import redis.clients.jedis.Jedis;
+
+import org.apache.geode.redis.general.PersistIntegrationTest;
+import org.apache.geode.test.junit.categories.RedisTest;
+
+@Category({RedisTest.class})
+public class PersistDockerAcceptanceTest extends PersistIntegrationTest {
```
dschneider-pivotal, line 28: is this test still a work in progress? I don't see that it tests anything

geode-redis/src/integrationTest/java/org/apache/geode/redis/general/PersistIntegrationTest.java
sabbeyPivotal: I set it to the default from awaitility.  I'll double check with Jens though.  For some reason I remember him having to set it really high, but it may not be necessary for Persist.

geode-redis/src/integrationTest/java/org/apache/geode/redis/general/PersistIntegrationTest.java
sabbeyPivotal: Sounds good, just made it local.

geode-redis/src/integrationTest/java/org/apache/geode/redis/general/PersistIntegrationTest.java
sabbeyPivotal: Just put an underscore before the data types and updated the key and value names.  Does that help in distinguishing?

geode-redis/src/integrationTest/java/org/apache/geode/redis/general/PersistIntegrationTest.java
sabbeyPivotal: It was a relic! We forgot to rename the method names.  Thank you!  As far as "do a bunch of", we were following a naming convention from a few other test suites.  I changed it though.  Maybe we should discuss as a team or rename the other methods as well?

geode-redis/src/test/java/org/apache/geode/redis/internal/executor/general/PersistExecutorJUnitTest.java
sabbeyPivotal: We wanted the unit test to remain pure and only test the relevant executor.  We also mock the Command class in other JUnit tests.  We could bring this up at techtro though if we want to change the way we're doing the executor unit tests?

geode-redis/src/acceptanceTest/java/org/apache/geode/redis/PersistDockerAcceptanceTest.java
sabbeyPivotal: It actually runs all the PersistIntegrationTests against native Redis running in a Docker container.

PR #4946, opened 2020-04-11
GEODE-7870: Fix PubSubIntegrationTest flakiness

jdeppe-pivotal:
The complete fix also builds on the following prior fixes

- GEODE-7943 add synchronization to Subscriptions class
- GEODE-7946: Fix redis publish/subscribe leaking netty buffers

### For all changes:

### Note:

### DIFF
--- a/geode-redis/src/integrationTest/java/org/apache/geode/redis/PubSubIntegrationTest.java
+++ b/geode-redis/src/integrationTest/java/org/apache/geode/redis/PubSubIntegrationTest.java
@@ -257,7 +257,7 @@ public class PubSubIntegrationTest {
     assertThat(result).isEqualTo(1);
 
     assertThat(mockSubscriber.getReceivedMessages()).isEmpty();
-    GeodeAwaitility.await().until(() -> mockSubscriber.getReceivedPMessages().size() != 0);
+    GeodeAwaitility.await().until(() -> mockSubscriber.getReceivedPMessages().isEmpty());
     assertThat(mockSubscriber.getReceivedPMessages()).containsExactly(message);
 
     mockSubscriber.punsubscribe("sal*s");
--- a/geode-redis/src/integrationTest/java/org/apache/geode/redis/PubSubIntegrationTest.java
+++ b/geode-redis/src/integrationTest/java/org/apache/geode/redis/PubSubIntegrationTest.java
@@ -41,7 +41,6 @@ import org.apache.geode.test.awaitility.GeodeAwaitility;
 import org.apache.geode.test.junit.categories.RedisTest;
 
 @Category({RedisTest.class})
-@Ignore("GEODE-7870")
 public class PubSubIntegrationTest {
   static Jedis publisher;
   static Jedis subscriber;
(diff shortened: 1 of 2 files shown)


geode-redis/src/integrationTest/java/org/apache/geode/redis/PubSubIntegrationTest.java
```diff
@@ -253,11 +251,14 @@ public void testPatternSubscribe() {
...
 
-    Long result = publisher.publish("salutations", "hello");
+    String message = "hello-" + System.currentTimeMillis();
+
+    Long result = publisher.publish("salutations", message);
     assertThat(result).isEqualTo(1);
 
     assertThat(mockSubscriber.getReceivedMessages()).isEmpty();
-    assertThat(mockSubscriber.getReceivedPMessages()).containsExactly("hello");
+    GeodeAwaitility.await().until(() -> mockSubscriber.getReceivedPMessages().size() != 0);
```
dschneider-pivotal, line 260: instead of size() != 0 !isEmpty() is a bit better

geode-redis/src/integrationTest/java/org/apache/geode/redis/PubSubIntegrationTest.java
```diff
@@ -253,11 +251,14 @@ public void testPatternSubscribe() {
 
     waitFor(() -> mockSubscriber.getSubscribedChannels() == 1);
 
-    Long result = publisher.publish("salutations", "hello");
+    String message = "hello-" + System.currentTimeMillis();
+
+    Long result = publisher.publish("salutations", message);
     assertThat(result).isEqualTo(1);
 
     assertThat(mockSubscriber.getReceivedMessages()).isEmpty();
```
dschneider-pivotal, line 259: It seems like this old assert that we have no receivedMessages should be AFTER the new await that we did receive a PMessage. I could be wrong about this but it seems like you should have the test wait for the event and then do the asserts

jdeppe-pivotal: There are actually 2 categories of received messages - one is as a result of a 'regular' subscription and one is as a result of a pattern subscription. They are both sent and processed separately. So it doesn't matter in which order these asserts are done in. In this case I just want to make sure I received 1 'pmessage' and did not receive any other messages.

dschneider-pivotal: But if you made sure AFTER you had received the message you are expecting, you would give the test more of a chance to catch that it also received the unexpected message. I don't think you should wait for something you never expect to happen; that would cause all tests to take too long to run. I'm okay if you leave the test as is

geode-redis/src/integrationTest/java/org/apache/geode/redis/PubSubIntegrationTest.java
```diff
@@ -253,11 +251,14 @@ public void testPatternSubscribe() {
...
-    assertThat(mockSubscriber.getReceivedPMessages()).containsExactly("hello");
+    GeodeAwaitility.await().until(() -> mockSubscriber.getReceivedPMessages().isEmpty());
```
dschneider-pivotal, line 260: Doesn't this need to be !isEmpty()?

jdeppe-pivotal: Doh! Thanks.

PR #4949, opened 2020-04-13
GEODE-7982: Close the client first in rolling upgrade test

upthewaterspout:
To prevent suspect strings from a background thread in the client, close the
client before the locator in this test.


upthewaterspout:
JMXMBeanReconnectDUnitTest failed, but that is not related to this test.

PR #4955, opened 2020-04-14
GEODE-7980: update rebalance output on colocated region

mkevo:

### For all changes:

### Note:

### DIFF
--- a/geode-gfsh/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/RebalanceMembersColocationTest.java
+++ b/geode-gfsh/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/RebalanceMembersColocationTest.java
@@ -89,8 +89,9 @@ public class RebalanceMembersColocationTest {
 
     gfsh.connectAndVerify(locator);
 
-    Map<String, List<String>> rebalanceResult = gfsh.executeAndAssertThat("rebalance")
-        .statusIsSuccess().hasTableSection().getActual().getContent();
+    Map<String, List<String>> rebalanceResult =
+        gfsh.executeAndAssertThat("rebalance --include-region=/" + PARENT_REGION_NAME)
+            .statusIsSuccess().hasTableSection().getActual().getContent();
 
     assertThat(rebalanceResult.get("Value").get(9)).isEqualTo("2");
   }
--- a/geode-gfsh/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/RebalanceMembersColocationTest.java
+++ b/geode-gfsh/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/RebalanceMembersColocationTest.java
@@ -1,0 +1,97 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package org.apache.geode.management.internal.cli.commands;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.stream.IntStream;
+
+import junitparams.JUnitParamsRunner;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.runner.RunWith;
+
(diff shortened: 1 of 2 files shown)


geode-gfsh/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/RebalanceMembersColocationTest.java
```diff
@@ -0,0 +1,97 @@
...
+
+    locator.waitUntilRegionIsReadyOnExactlyThisManyServers("/" + PARENT_REGION_NAME, 2);
+    locator.waitUntilRegionIsReadyOnExactlyThisManyServers("/" + CHILD_REGION_NAME, 2);
+
+    gfsh.connectAndVerify(locator);
+
+    Map<String, List<String>> rebalanceResult = gfsh.executeAndAssertThat("rebalance")
+        .statusIsSuccess().hasTableSection().getActual().getContent();
+
+    assertThat(rebalanceResult.get("Value").get(9)).isEqualTo("2");
```
DonalEvans, line 95: It would be better to assert that the output for both regions is correct rather than just the first one that appears in the gfsh output. I think that the child region still reports 0 for the total members executed on, which is not correct.

mkevo: I'm not sure but I think that it is normal to get zeros for colocated region as the whole rebalance is executed on parent region. 
As I see when execute rebalance it add all results to parent region. It goes in this part of code
```
for (PartitionRebalanceInfo details : detailSet) {
   results.addDetails(details);
}

```
If it is not normal behaviour than this whole logic is wrong and should be changed.

DonalEvans: I think that it's incorrect for "number of members rebalance was executed on" to be zero for colocated regions, since the rebalance *is* being executed on members for those regions. The other details like buckets moved and primaries transferred aren't zero for colocated regions, so members executed on shouldn't be zero either.

mkevo: Hi @DonalEvans,
I tested this without changes due to adding "number of members rebalance was executed on" row and got the following output:
```

Command result for <rebalance>: 
Rebalanced partition regions /parentRegion

                                       Rebalanced Stats                                         | Value
----------------------------------------------------------------------------------------------- | -----
Total bytes in all redundant bucket copies created during this rebalance                        | 0
Total time (in milliseconds) spent creating redundant bucket copies during this rebalance       | 0
Total number of redundant copies created during this rebalance                                  | 0
Total bytes in buckets moved during this rebalance                                              | 18426
Total time (in milliseconds) spent moving buckets during this rebalance                         | 4119
Total number of buckets moved during this rebalance                                             | 132
Total time (in milliseconds) spent switching the primary state of buckets during this rebalance | 0
Total primaries transferred during this rebalance                                               | 0
Total time (in milliseconds) for this rebalance                                                 | 8372

Rebalanced partition regions /childRegion

                                       Rebalanced Stats                                         | Value
----------------------------------------------------------------------------------------------- | -----
Total bytes in all redundant bucket copies created during this rebalance                        | 0
Total time (in milliseconds) spent creating redundant bucket copies during this rebalance       | 0
Total number of redundant copies created during this rebalance                                  | 0
Total bytes in buckets moved during this rebalance                                              | 0
Total time (in milliseconds) spent moving buckets during this rebalance                         | 0
Total number of buckets moved during this rebalance                                             | 0
Total time (in milliseconds) spent switching the primary state of buckets during this rebalance | 0
Total primaries transferred during this rebalance                                               | 0
Total time (in milliseconds) for this rebalance                                                 | 0
```

From this it seems that all details are zero. Also what I saw is that stats for parentRegion is twice updated with stats from both regions, parentRegion and childRegion.  
Before breaking Gfsh out into separate module it print result for both regions together, as it is below.

```
Command result for <rebalance>: 
Rebalanced partition regions /parentRegion, /childRegion

                                       Rebalanced Stats                                         | Value
----------------------------------------------------------------------------------------------- | -----
Total bytes in all redundant bucket copies created during this rebalance                        | 0
Total time (in milliseconds) spent creating redundant bucket copies during this rebalance       | 0
Total number of redundant copies created during this rebalance                                  | 0
Total bytes in buckets moved during this rebalance                                              | 18426
Total time (in milliseconds) spent moving buckets during this rebalance                         | 4119
Total number of buckets moved during this rebalance                                             | 132
Total time (in milliseconds) spent switching the primary state of buckets during this rebalance | 0
Total primaries transferred during this rebalance                                               | 0
Total time (in milliseconds) for this rebalance                                                 | 8372

```
From that I can see that it is normal that output for childRegion is all zeros or the whole logic is wrong and should be changed. 
In case it is normal that result is all zeros than this change has a sense to get correct number of servers on which rebalance is executed.

DonalEvans: Ah, I see, thanks for the clarification. In that case, yes, the output provided for "members executed on" is consistent with the output for the other stats, and I withdraw my request for additional changes.

That being said, it might be worth discussing on the dev list whether showing 0 for colocated regions is actually correct, and filing a new bug ticket if people think that we should output differently. Having a section with all the values 0 seems potentially misleading and confusion to me.

PR #4958, opened 2020-04-14
GEODE-7852: test ClientHealthMonitor functionality behind a SNI gateway

bschuchardt:
This ensures that a server sitting behind an SNI gateway detects the
loss of a client and cleans up after it.  In this case the test detects
that the server has closed CQs created by the non-durable client.

Since test code is not accessible in the Docker container that's running
the server I've enhanced the StatArchiveReader to be able to report the
values of a statistic and have enabled statistics recording in the
server.  Much of this code came from the deprecated SystemAdmin class.

(cherry picked from commit 376df4cc7abc6db38b4b298f8af6f3c53baedf15)

### For all changes:

### Note:

### DIFF
--- a/geode-assembly/src/acceptanceTest/java/org/apache/geode/client/sni/ClientSNICQAcceptanceTest.java
+++ b/geode-assembly/src/acceptanceTest/java/org/apache/geode/client/sni/ClientSNICQAcceptanceTest.java
@@ -184,7 +184,6 @@ public class ClientSNICQAcceptanceTest {
           arguments("java", "-cp", "/geode/lib/geode-dependencies.jar",
               "org.apache.geode.internal.statistics.StatArchiveReader",
               "stat", "server-dolores/statArchive.gfs", "CqServiceStats.numCqsClosed"));
-      System.out.println("stats from server are :" + stats);
       // the stat should transition from zero to one at some point
       assertThat(stats).contains("0.0 1.0");
     });
--- a/geode-assembly/src/acceptanceTest/java/org/apache/geode/client/sni/ClientSNICQAcceptanceTest.java
+++ b/geode-assembly/src/acceptanceTest/java/org/apache/geode/client/sni/ClientSNICQAcceptanceTest.java
@@ -33,9 +33,10 @@ import java.util.concurrent.atomic.AtomicInteger;
 
 import com.palantir.docker.compose.DockerComposeRule;
 import org.junit.After;
+import org.junit.AfterClass;
 import org.junit.Before;
+import org.junit.BeforeClass;
 import org.junit.ClassRule;
-import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.TestRule;
 
@@ -67,15 +68,17 @@ public class ClientSNICQAcceptanceTest {
   @ClassRule
   public static TestRule ignoreOnWindowsRule = new IgnoreOnWindowsRule();
 
-  @Rule
-  public DockerComposeRule docker = DockerComposeRule.builder()
-      .file(DOCKER_COMPOSE_PATH.getPath())
-      .build();
+  @ClassRule
+  public static NotOnWindowsDockerRule docker =
+      new NotOnWindowsDockerRule(() -> DockerComposeRule.builder()
+          .file(DOCKER_COMPOSE_PATH.getPath()).build());
+
   private CqQuery cqTracker;
 
   AtomicInteger eventCreateCounter = new AtomicInteger(0);
   AtomicInteger eventUpdateCounter = new AtomicInteger(0);
   private ClientCache cache;
+  private Region<String, Integer> region;
 
   class SNICQListener implements CqListener {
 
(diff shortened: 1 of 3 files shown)


geode-assembly/src/acceptanceTest/java/org/apache/geode/client/sni/ClientSNICQAcceptanceTest.java
```diff
@@ -107,9 +111,18 @@ public void before() throws IOException, InterruptedException {
 
   }
 
+  @After
+  public void after() throws Exception {
+    String output =
+        docker.exec(options("-T"), "geode", arguments("cat", "server-dolores/server-dolores.log"));
+    System.out.println("Server log file--------------------------------\n" + output);
```
Bill, line 118: do we need this log in the test output?

geode-assembly/src/acceptanceTest/java/org/apache/geode/client/sni/ClientSNICQAcceptanceTest.java
```diff
@@ -22,6 +22,7 @@
 import static org.apache.geode.distributed.ConfigurationProperties.SSL_REQUIRE_AUTHENTICATION;
 import static org.apache.geode.distributed.ConfigurationProperties.SSL_TRUSTSTORE;
 import static org.apache.geode.distributed.ConfigurationProperties.SSL_TRUSTSTORE_PASSWORD;
+import static org.apache.geode.test.awaitility.GeodeAwaitility.await;
```
Bill, line 25: Now that we have `NotOnWindowsDockerRule`, we have the ability for a test to start containers once and to break actual testing into multiple test methods. We do it that way (in `SingleServerSNIAcceptanceTest` and `DualServerSNIAcceptanceTest`) for performance reasons—we don't want to take too many minutes out of our CI pipeline, for SNI testing.

So for these reasons I think we should do one of two things:

1. use `NotOnWindowsDockerRule` here, and break these tests into multiple test methods
2. go further, and transfer these CQ tests over to the `SingleServerSNIAcceptanceTest` class (eliminating this test class entirely)

I can see why we might want to do (1) even though it will add more Docker startup/teardown time (versus (2)), since there is some CQ-specific stuff here: 3 fields and an inner (`CqListener`) class.

geode-assembly/src/acceptanceTest/java/org/apache/geode/client/sni/ClientSNICQAcceptanceTest.java
```diff
@@ -152,6 +165,21 @@ public void performSimpleCQOverSNIProxy()
...
+    ((PoolImpl) cache.getDefaultPool()).killPrimaryEndpoint();
+    // since we can't run code in the server let's grab the CQ statistics and verify that
+    // the CQ has been closed. StatArchiveReader has a main() that we can use to get a printout
+    // of stat values
+    await().untilAsserted(() -> {
+      String stats = docker.exec(options("-T"), "geode",
+          arguments("java", "-cp", "/geode/lib/geode-dependencies.jar",
+              "org.apache.geode.internal.statistics.StatArchiveReader",
+              "stat", "server-dolores/statArchive.gfs", "CqServiceStats.numCqsClosed"));
+      System.out.println("stats from server are :" + stats);
```
Bill, line 178: do we need/want these stats in the test output?

geode-assembly/src/acceptanceTest/resources/org/apache/geode/client/sni/geode-config/gemfire.properties
```diff
@@ -15,4 +15,5 @@
 # limitations under the License.
 #
 
-#empty
+statistic-sampling-enabled=true
```
Bill, line 18: this turns on stats for all our SNI tests—not just this one. I assume that is ok but just wanted to remind everybody

geode-assembly/src/acceptanceTest/java/org/apache/geode/client/sni/ClientSNICQAcceptanceTest.java
bschuchardt: Dan thought it was a good idea for the SingleServer test & I thought so, too.  If something goes wrong you probably want this output in the logs.

geode-assembly/src/acceptanceTest/java/org/apache/geode/client/sni/ClientSNICQAcceptanceTest.java
bschuchardt: There's only one test in this class.  No need for the ClassRule, though we could standardize on using it if we want.  The SingleServer test expects the same content to be in the server cache for each test, so I don't think it would be appropriate to add this test to it without addressing that.  I would rather keep this class and have it focus on subscriptions.

geode-assembly/src/acceptanceTest/java/org/apache/geode/client/sni/ClientSNICQAcceptanceTest.java
bschuchardt: Yes, if something goes wrong we can see why things didn't match in the awaitility failure message.  Helped me a lot.

geode-assembly/src/acceptanceTest/resources/org/apache/geode/client/sni/geode-config/gemfire.properties
bschuchardt: yes, it does.  I think that's okay.

geode-assembly/src/acceptanceTest/java/org/apache/geode/client/sni/ClientSNICQAcceptanceTest.java
bschuchardt: I've changed this to a ClassRule

PR #4963, opened 2020-04-15
GEODE-7503: Block Cache.close() until everything is disconnected

kirklund:
Subsequent calls to GemFireCacheImpl close will now block until the
first call to close completes instead of returning from early-outs.

Use boolean argument to skip await if cache close is invoked from
the DisconnectListener in InternalLocator.

This change has ONLY been tested against the precheckin test jobs
invoked by this PR.

I ended up using a ThreadLocal to make Cache.close() reentrant for 
the same thread because Geode has so many calls to Cache.close().


kirklund:
IntegrationTest exceeded timeout, and DistributedTest had a common JMXMBeanReconnectDUnitTest failure.

kirklund:
Memcached IntegrationJUnitTest hangs IntegrationTest because Cache.close() calls GeodeMemcachedService.close() which again calls Cache.close(). Looks like the code base has lots of Cache.close() calls -- all of them could theoretically cause issues.

upthewaterspout:
The combination of a `ThreadLocal<Thread>`, a `CountDownLatch`, and `synchronized()` seems a bit complicated.

The whole close block is already inside of  `synchronized(GemFireCacheImpl.class) {}`. It seems like the race condition is just with the early out at the beginning of close:

```
if (isClosed()) {
      return;
}
```

Couldn't that just be changed to 

```
if (!skipWait && isClosed()) {
      return;
}
```

To handle thread reentering cache.close, just don't add a skipWait check inside the synchronized block here:

```
synchronized (GemFireCacheImpl.class) {
      // ALL CODE FOR CLOSE SHOULD NOW BE UNDER STATIC SYNCHRONIZATION OF GemFireCacheImpl.class
      // static synchronization is necessary due to static resources
      if (isClosed()) {

====> Remove the below check
        if (!skipAwait...)
```

Any code that gets into the synchronized block is guaranteed that either (a) the cache has not been closed by another thread (b) the cache is completely closed already or (c) the thread is reentering cache close, in which case it should early out.

There is probably some wrinkle that I'm missing here...

kirklund:
@upthewaterspout, I have some questions about your review of Cache.close() changes.
{noformat}
  @Override
  public void close(String reason, Throwable systemFailureCause, boolean keepAlive,
      boolean keepDS, boolean skipAwait) {
    securityService.close();
    if (isClosed()) {
      if (!skipAwait && !Thread.currentThread().equals(CLOSING_THREAD.get())) {
        waitUntilClosed();
      }
      return;
    }
    if (!keepDS && systemFailureCause == null
        && (isReconnecting() || system.getReconnectedSystem() != null)) {
      logger.debug(
          "Cache is shutting down distributed system connection. isReconnecting={} reconnectedSystem={} keepAlive={} keepDS={}",
          isReconnecting(), system.getReconnectedSystem(), keepAlive, keepDS);
      system.stopReconnectingNoDisconnect();
      if (system.getReconnectedSystem() != null) {
        system.getReconnectedSystem().disconnect();
      }
      return;
    }
    synchronized (GemFireCacheImpl.class) {
      // ALL CODE FOR CLOSE SHOULD NOW BE UNDER STATIC SYNCHRONIZATION OF GemFireCacheImpl.class
      // static synchronization is necessary due to static resources
      if (isClosed()) {
        if (!skipAwait && !Thread.currentThread().equals(CLOSING_THREAD.get())) {
          waitUntilClosed();
        }
        return;
      }
      CLOSING_THREAD.set(Thread.currentThread());
{noformat}
If I remove that 1st early-out, then all subsequent calls will repeatedly execute the block about the reconnecting system. I’ve read the code that it invokes — such as 
{noformat}
system.stopReconnectingNoDisconnect();
{noformat}
and 
{noformat}
system.getReconnectedSystem().disconnect()
{noformat}
… so far, I’m not convinced that it’s safe to remove the 1st early out. Are you sure it’s safe to invoke these calls more than once? The current behavior never invokes them more than once.

upthewaterspout:
> … so far, I’m not convinced that it’s safe to remove the 1st early out. Are you sure it’s safe to invoke these calls more than once? The current behavior never invokes them more than once.

It looks like it might be ok to have a second thread go in there? If DistributedSystem.disconnect() is safe to call twice, it should be ok. It looks like that method also synchronizes on GemFireCacheImpl and also has some interesting wait logic to wait until other threads have finished disconnecting. Hmm.

kirklund:
I'm currently running a lot of manual tests against this change (ie hydra tests).

PR #4966, opened 2020-04-16
GEODE-7851: Pulse logout requests end of OAuth session

demery-pivotal:
When Pulse is configured to use OAuth, and a user logs out of Pulse,
Pulse redirects the browser to a page where the user can take action to
end their session. The available actions depend on the OAuth provider,
but may include revoking the token or logging out of the OAuth provider
entirely.

Main changes:

- Changed OAuthSecurityConfig to install two logout handlers: A
  RepositoryLogoutHandler (renamed from LogoutHandler) and an
  OidcClientInitiatedLogoutSuccessHandler.

- Added a pulse.security.oauth.endSessionEndpoint property to specify
  the URL to which the OidcClientInitiatedLogoutSuccessHandler should
  redirect the browser on logout.

- Configured the OAuthSecurityConfig to add the "end session endpoint"
  property value to the client configuration metadata.  On logout, the
  OidcClientInitiatedLogoutSuccessHandler redirects the browser to this
  endpoint, where the user can take action to end the session.

- In the OAuthClientConfig class (extracted from OAuthSecurityConfig),
  restored the code to explicitly list the scopes that Pulse is
  requesting, in particular to list "openid" in the scopes. Though
  authentication works just fine without that explicit list, the
  OidcClientInitiatedLogoutSuccessHandler does not. The
  OidcClientInitiatedLogoutSuccessHandler handles logout only if the the
  principal is an OidcUser. If "openid" is not explicitly listed in the
  client's scopes. Spring creates OAuth2User principals instead of
  OidcUser principals, and OidcClientInitiatedLogoutSuccessHandler
  return without redirecting the browser.

Also refactored to support the above changes:

- Moved the oauth client service configuration from OAuthSecurityConfig
  to a new OAuthClientConfig class. This breaks Respository's dependence
  on OAuthSecurityConfig, which in turn (through the LogoutHandler)
  depended on Repository. Repository now gets its
  OAuth2AuthorizedClientService from the OAuthClientConfig class, which
  does not in turn depend on Repository.

- Marked two Repository constructors as non-required. Spring will pick
  whichever one has the most dependencies it can satisfy. So if the
  profile specifies an OAuth2AuthorizedClientService, Spring will call
  the constructor that takes one of those. Otherwise Spring will call
  the no-args constructor.

- Renamed LogoutHandler to RepositoryLogoutHandler to better reflect its
  specific responsibilities.

- Changed RepositoryLogoutHandler to implement LogoutHandler instead of
  LogoutSuccessHandler. Now it does its work *during* logout instead of
  *after.*

- Changed DefaultSecurityConfig to specify the logout success URL
  directly instead of via a logout success handler. (OAuthSecurityConfig
  no longer needs a logout success URL, because the OIDC logout handler
  redirects to the OAuth provider instead.)

### For all changes:

### Note:

### DIFF
--- a/geode-pulse/src/main/java/org/apache/geode/tools/pulse/internal/security/OAuthClientConfig.java
+++ b/geode-pulse/src/main/java/org/apache/geode/tools/pulse/internal/security/OAuthClientConfig.java
@@ -1,0 +1,109 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package org.apache.geode.tools.pulse.internal.security;
+
+import static java.util.Collections.singletonMap;
+
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Profile;
+import org.springframework.context.annotation.PropertySource;
+import org.springframework.security.oauth2.client.InMemoryOAuth2AuthorizedClientService;
+import org.springframework.security.oauth2.client.OAuth2AuthorizedClientService;
+import org.springframework.security.oauth2.client.oidc.web.logout.OidcClientInitiatedLogoutSuccessHandler;
+import org.springframework.security.oauth2.client.registration.ClientRegistration;
+import org.springframework.security.oauth2.client.registration.ClientRegistrationRepository;
+import org.springframework.security.oauth2.client.registration.InMemoryClientRegistrationRepository;
+import org.springframework.security.oauth2.client.web.AuthenticatedPrincipalOAuth2AuthorizedClientRepository;
+import org.springframework.security.oauth2.client.web.OAuth2AuthorizedClientRepository;
+import org.springframework.security.oauth2.core.AuthorizationGrantType;
+
+/**
+ * Configures Pulse to use the OAuth 2 provider defined by properties in {@code pulse.properties}.
+ */
+@Configuration
+@Profile("pulse.authentication.oauth")
+@PropertySource("classpath:pulse.properties")
+public class OAuthClientConfig {
+  @Value("${pulse.oauth.providerId}")
+  private String providerId;
+  @Value("${pulse.oauth.providerName}")
+  private String providerName;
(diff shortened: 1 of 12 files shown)


davebarnes97:
Thanks for documenting, @demery-pivotal. One small editorial request: "End-User" should simply be "end user".

geode-pulse/src/main/java/org/apache/geode/tools/pulse/internal/security/OAuthClientConfig.java
```diff
@@ -0,0 +1,109 @@
...
+import org.springframework.security.oauth2.client.web.OAuth2AuthorizedClientRepository;
+import org.springframework.security.oauth2.core.AuthorizationGrantType;
+
+/**
+ * Configures Pulse to use the OAuth 2 provider defined by properties in {@code pulse.properties}.
+ */
+@Configuration
+@Profile("pulse.authentication.oauth")
+@PropertySource("classpath:pulse.properties")
+public class OAuthClientConfig {
```
jinmeiliao, line 41: Any particular reason we move this out of the OauthSecurityConfig?

geode-pulse/src/main/java/org/apache/geode/tools/pulse/internal/security/OAuthClientConfig.java
```diff
@@ -0,0 +1,109 @@
...
+        .authorizationGrantType(AuthorizationGrantType.AUTHORIZATION_CODE)
+        .redirectUriTemplate("{baseUrl}/login/oauth2/code/{registrationId}")
+        .clientId(clientId)
+        .clientSecret(clientSecret)
+        .scope("openid", "CLUSTER:READ", "CLUSTER:WRITE", "DATA:READ", "DATA:WRITE")
+        .authorizationUri(authorizationUri)
+        .tokenUri(tokenUri)
+        .userInfoUri(userInfoUri)
+        .jwkSetUri(jwkSetUri)
+        .providerConfigurationMetadata(
```
jinmeiliao, line 75: Is this "end_session_endpoint" provider specific? i.e. it's a strictly UAA thing or others have that as well? Can it be null?

geode-pulse/src/main/java/org/apache/geode/tools/pulse/internal/security/OAuthClientConfig.java
demery-pivotal: Yes: To break a dependency cycle. For details, see the commit message.

geode-pulse/src/main/java/org/apache/geode/tools/pulse/internal/security/OAuthClientConfig.java
demery-pivotal: That's defined by a draft standard: https://openid.net/specs/openid-connect-session-1_0.html

As far as I can tell, that standard has never been approved, though some OpenID Connect providers do implement it. See for example this Okta documentation: https://developer.okta.com/docs/reference/api/oidc/#well-known-openid-configuration

The sad thing is that UAA does _not_ implement it. But redirecting to UAA's "profile" page has a similar effect. Not exactly what that draft standard defines, but similar.

With the current commit, if you leave the property undefined, Spring assigns the property value expression ("${pulse.oauth.userNameAttributeName}") to the field, and the logout success handler fails on logout.

We did not try supplying null. I'll run a few experiments now. I think the logout success handler will simply skip the "end session" endpoint, and go straight to the "logout success" URL (which we will need to define if we allow the end session endpoint to be null).

PR #4971, opened 2020-04-17
GEODE-7996 Create DUNIT tests for SADD command

ringles:
Adds distributed unit tests for the Redis SADD command.

### DIFF
--- a/geode-redis/src/distributedTest/java/org/apache/geode/redis/SaddDistDunitTest.java
+++ b/geode-redis/src/distributedTest/java/org/apache/geode/redis/SaddDistDunitTest.java
@@ -1,0 +1,239 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package org.apache.geode.redis;
+
+import static org.apache.geode.distributed.ConfigurationProperties.MAX_WAIT_TIME_RECONNECT;
+import static org.apache.geode.distributed.ConfigurationProperties.REDIS_BIND_ADDRESS;
+import static org.apache.geode.distributed.ConfigurationProperties.REDIS_PORT;
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Properties;
+import java.util.Set;
+
+import org.junit.AfterClass;
+import org.junit.Before;
+import org.junit.BeforeClass;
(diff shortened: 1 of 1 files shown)


geode-redis/src/distributedTest/java/org/apache/geode/redis/SaddDistDunitTest.java
```diff
@@ -0,0 +1,239 @@
...
+
+    serverProperties3.setProperty(REDIS_PORT, "6373");
+    serverProperties3.setProperty(REDIS_BIND_ADDRESS, LOCAL_HOST);
+
+    locator = clusterStartUp.startLocatorVM(0, locatorProperties);
+    server1 = clusterStartUp.startServerVM(1, serverProperties1, locator.getPort());
+    server2 = clusterStartUp.startServerVM(2, serverProperties2, locator.getPort());
+    server3 = clusterStartUp.startServerVM(3, serverProperties3, locator.getPort());
+
+    jedis1 = new Jedis(LOCAL_HOST, 6371);
```
dschneider-pivotal, line 83: I wonder if these hard coded port numbers (6371, 6372, 6373) are what is causing the stressNewTest to fail.
I think you could instead have the test call AvailablePortHelper.getRandomAvailableTCPPort(3) which would give you back an array of 3 ints

geode-redis/src/distributedTest/java/org/apache/geode/redis/SaddDistDunitTest.java
```diff
@@ -0,0 +1,239 @@
...
+    List<String> members2 = makeMemberList(SET_SIZE, "member2-");
+
+    Runnable addSetsWithClient1 = makeSADDRunnable(key, members1, jedis1);
+    Runnable addSetsWithClient2 = makeSADDRunnable(key, members2, jedis2);
+
+    runConcurrentThreads(addSetsWithClient1, addSetsWithClient2);
+
+    Set<String> results = jedis3.smembers(key);
+
+    members1.addAll(members2);
```
dschneider-pivotal, line 135: I think this method would be clearer if at the top, when you initialize members1 and members2 you also created a list of "allMembers" by adding members1 and members2 to it. I think this is better than mutating members1 by adding members2 to it.

geode-redis/src/distributedTest/java/org/apache/geode/redis/SaddDistDunitTest.java
```diff
@@ -0,0 +1,239 @@
...
+    Set<String> results1 = jedis3.smembers(key1);
+    Set<String> results2 = jedis3.smembers(key2);
+
+    assertThat(results1.toArray()).containsExactlyInAnyOrder(members1.toArray());
+    assertThat(results2.toArray()).containsExactlyInAnyOrder(members2.toArray());
+
+  }
+
+  @Test
+  public void should_distributeDataAmongMultipleServers_givenTwoSetsOfClients_OperatingOnTheSamSetConcurrently()
```
dschneider-pivotal, line 183: Add a flavor of this method that has each client adding different members to the set?

ringles: Done!

geode-redis/src/distributedTest/java/org/apache/geode/redis/SaddDistDunitTest.java
ringles: I agree, that's clearer. Implemented!

geode-redis/src/distributedTest/java/org/apache/geode/redis/SaddDistDunitTest.java
ringles: Good idea, I've implemented that and we'll see what the stress test thinks of it.

PR #4975, opened 2020-04-20
Logging documentation fixes

alb3rtobr:
For reference, link to the conversation on dev list: https://markmail.org/thread/a6gobaphywmkm7gq


davebarnes97:
Thanks for this contribution, @alb3rtobr. 
For the gfsh command reference writeup on change loglevel, I'd like to suggest a minor re-wording, which replaces future tense with present tense.

Now:
Changes the logging level on specified members. This command will take effect if the default <%=vars.product_name%> logging configuration is used.

In case of using custom `Log4J` configuration, this command will not work unless the member whose logging level you want to change was started using the `--J=-Dgeode.LOG_LEVEL_UPDATE_OCCURS=ALWAYS` system property.

Suggested revision:
Changes the logging level on specified members. This command takes effect if the default <%=vars.product_name%> logging configuration is used.

When using a custom `Log4J` configuration, this command takes effect only if the member whose logging level you want to change was started using the `--J=-Dgeode.LOG_LEVEL_UPDATE_OCCURS=ALWAYS` system property.

mkevo:
Since it was changed according to the comment, I will merge this.

PR #4976, opened 2020-04-20
GEODE-7997: Document needed location of parallel gateway sender disk …

karensmolermiller:
…stores

Reviewers: I put the same information in 3 distinct locations. It is my hope that Geode-users who need this info will encounter it in one of those places when they look at the docs.


geode-docs/managing/disk_storage/using_disk_stores.html.md.erb
```diff
@@ -170,7 +170,10 @@ Example of using a named disk store for PDX serialization metadata (cache.xml):
 
 Gateway sender queues are always overflowed and may be persisted. Assign them to overflow disk stores if you do not persist, and to persistence disk stores if you do.
 
-Example of using a named disk store for gateway sender queue persistence:
+Persisted data from a parallel gateway sender must go to the same disk
+store as used by the region.
```
nabarunnag, line 174: This is not required, but it would be nice to tell the user why this needs to be done. i.e. parallel gateway sender queues are  collocated with the parent region

PR #4987, opened 2020-04-22
GEODE-7678: Support for cache-listener and client-notification for Partitioned Region Clear operation.

agingade:
The changes are made to PR clear messaging and locking mechanism to preserve
cache-listener and client-events ordering during concurrent cache operation
while clear in progress.

### For all changes:

- [Y] Is your initial contribution a single, squashed commit?

### Note:


geode-core/src/main/java/org/apache/geode/internal/cache/ClearPartitionedRegion.java
```diff
@@ -0,0 +1,342 @@
...
+    } catch (IllegalStateException e) {
+      partitionedRegion.lockCheckReadiness();
+    } catch (Exception ex) {
+      logger.warn("Caught exception while unlocking clear distributed lock", ex.getMessage());
+    }
+  }
+
+  void obtainWriteLockForClear(RegionEventImpl event) {
+    sendLocalClearRegionMessage(event,
+        ClearPartitionedRegionMessage.OperationType.OP_LOCK_FOR_CLEAR);
```
gesterzhou, line 73: It's better to use OP_LOCK_FOR_PR_CLEAR, OP_UNLOCK_FOR_PR_CLEAR. 

The OP_LOCK_FOR_CLEAR is used for DistributedRegion's CLEAR. Don't mix the operation type name.

geode-core/src/main/java/org/apache/geode/internal/cache/ClearPartitionedRegion.java
```diff
@@ -0,0 +1,342 @@
...
+    obtainClearLockLocal(partitionedRegion.getDistributionManager().getId());
```
gesterzhou, line 74: Should we do local lock first, then distribute the message to other members? The replicated region's clear is using this sequence. I think it's better to follow the same sequence, unless we have special reason. 

Same thing for releaseWriteLockForClear.

geode-core/src/main/java/org/apache/geode/internal/cache/ClearPartitionedRegion.java
```diff
@@ -0,0 +1,342 @@
...
+      RegionVersionVector vector) {
+    clearRegionLocal(regionEvent, cacheWrite, null);
+    sendLocalClearRegionMessage(regionEvent,
+        ClearPartitionedRegionMessage.OperationType.OP_CLEAR);
+  }
+
+  public void clearRegionLocal(RegionEventImpl regionEvent, boolean cacheWrite,
+      RegionVersionVector vector) {
+    // Synchronized to handle the requester departure.
+    synchronized (lockForListenerAndClientNotification) {
```
gesterzhou, line 94: Do you need to sync here again while you have synchronized (clearLock) in PR?

geode-core/src/main/java/org/apache/geode/internal/cache/ClearPartitionedRegion.java
```diff
@@ -0,0 +1,347 @@
...
+
+      if (localCqInterestFrInfo != null) {
+        event.setLocalFilterInfo(localCqInterestFrInfo.getLocalFilterInfo());
+      }
+    }
+    partitionedRegion.notifyBridgeClients(event);
+  }
+
+  protected void obtainClearLockLocal(InternalDistributedMember requester) {
+    synchronized (lockForListenerAndClientNotification) {
```
gesterzhou, line 151: do you need to sync here? There's only one thread in the whole cluster can do the clear at any time.

geode-core/src/main/java/org/apache/geode/internal/cache/ClearPartitionedRegion.java
```diff
@@ -0,0 +1,347 @@
...
+      if (!partitionedRegion.getDistributionManager().isCurrentMember(requester)) {
```
gesterzhou, line 152: is this possible?

geode-core/src/main/java/org/apache/geode/internal/cache/ClearPartitionedRegion.java
```diff
@@ -0,0 +1,347 @@
...
+        return;
+      }
+
+      lockForListenerAndClientNotification.setLocked(requester);
```
gesterzhou, line 156: BTW, where is the code to pre-create all the buckets? The code will prevent a new primary bucket created somewhere else. I remember previous version has that code.

geode-core/src/main/java/org/apache/geode/internal/cache/ClearPartitionedRegion.java
```diff
@@ -0,0 +1,347 @@
...
+      partitionedRegion.getPartitionedRegionLockService().unlock(clearLock);
+    } catch (IllegalStateException e) {
+      partitionedRegion.lockCheckReadiness();
+    } catch (Exception ex) {
+      logger.warn("Caught exception while unlocking clear distributed lock", ex.getMessage());
+    }
+  }
+
+  void obtainWriteLockForClear(RegionEventImpl event) {
+    sendLocalClearRegionMessage(event,
```
gesterzhou, line 72: Since the message is to let other member lock rvv and prepare for clear. The "sendLocalClearRegionMessage" is better to be rename to sendLockForClearPRMessage

gesterzhou, line 72: sendLocalClearRegionMessage() created a message whose operation is ClearPartitionedRegionMessage.OperationType.OP_LOCK_FOR_CLEAR, but the event's operation 
 is Operation.REGION_LOCAL_CLEAR. It's confusing here. Actually this message is not "local clear", it's only "lock for pr clear (locally)".

geode-core/src/main/java/org/apache/geode/internal/cache/ClearPartitionedRegion.java
```diff
@@ -0,0 +1,347 @@
...
+  void releaseWriteLockForClear(RegionEventImpl event) {
+    sendLocalClearRegionMessage(event,
+        ClearPartitionedRegionMessage.OperationType.OP_UNLOCK_FOR_CLEAR);
+    releaseClearLockLocal();
+  }
+
+  void clearRegion(RegionEventImpl regionEvent, boolean cacheWrite,
+      RegionVersionVector vector) {
+    clearRegionLocal(regionEvent, cacheWrite, null);
+    sendLocalClearRegionMessage(regionEvent,
```
gesterzhou, line 86: sendLocalClearRegionMessage() method is shared by all the 3 mesgs: obtainWriteLockForClear, clearRegion, releaseWriteLockForClear. So it's better to rename it to sendPRClearOperationMessage, or other generic name. It should not be "LocalClear".

geode-core/src/main/java/org/apache/geode/internal/cache/ClearPartitionedRegion.java
gesterzhou: I still think to do local lock, local clear, local release first, then send the message, especially current local clear is before sending mesg.

geode-core/src/main/java/org/apache/geode/internal/cache/ClearPartitionedRegion.java
```diff
@@ -0,0 +1,379 @@
...
+  }
+
+  void obtainLockForClear(RegionEventImpl event) {
+    sendClearRegionMessage(event,
+        ClearPartitionedRegionMessage.OperationType.OP_LOCK_FOR_PR_CLEAR);
+    obtainClearLockLocal(partitionedRegion.getDistributionManager().getId());
+  }
+
+  void releaseLockForClear(RegionEventImpl event) {
+    sendClearRegionMessage(event,
```
gesterzhou, line 80: sendClearRegionMessage is better to rename to sendClearPartitionedRegionMessage, because later you are sending ClearPartitionedRegionMessage

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
```diff
@@ -10373,4 +10377,27 @@ void updatePartitionRegionConfig(
...
+
+  protected ClearPartitionedRegion getClearPartitionedRegion() {
+    return clearPartitionedRegion;
+  }
+
+  @Override
+  void cmnClearRegion(RegionEventImpl regionEvent, boolean cacheWrite, boolean useRVV) {
+    // Synchronized to avoid other threads invoking clear on this vm/node.
+    synchronized (clearLock) {
+      clearPartitionedRegion.doClear(regionEvent, cacheWrite, this);
```
gesterzhou, line 10389: I wonder before calling doClear(), we need to call: final TXStateProxy tx = cache.getTXMgr().pauseTransaction();

Since we will get rvvLock for all buckets before clear, there's no guaranteed sequence for getting rvvlock for buckets, if some on-going tx requested some rvvLock on some buckets (such as bucket1) and is requesting rvvLock on bucket2. If clear thread did not pause the transaction, there could be deadlock.

geode-core/src/main/java/org/apache/geode/internal/cache/ClearPartitionedRegion.java
```diff
@@ -0,0 +1,379 @@
...
+import org.apache.geode.internal.cache.versions.RegionVersionVector;
+import org.apache.geode.logging.internal.log4j.api.LogService;
+
+public class ClearPartitionedRegion {
+
+  private static final Logger logger = LogService.getLogger();
+
+  private final PartitionedRegion partitionedRegion;
+
+  private LockForListenerAndClientNotification lockForListenerAndClientNotification =
```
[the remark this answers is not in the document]
jujoramos: +1

geode-core/src/main/java/org/apache/geode/internal/cache/InternalRegion.java
```diff
@@ -459,4 +459,7 @@ VersionedObjectList basicRemoveAll(Collection<Object> keys,
    * @return true if synchronization should be attempted
    */
   boolean shouldSyncForCrashedMember(InternalDistributedMember id);
+
+  void clearRegionLocal(RegionEventImpl regionEvent, boolean cacheWrite,
```
jujoramos, line 463: Maybe it's just me, but the old method name , `clearRegionLocally`, seems to be easier to understand for the reader?.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionDataStore.java
```diff
@@ -980,6 +980,14 @@ protected void lockBucketCreationAndVisit(BucketVisitor visitor) {
     }
   }
 
+  protected void lockBucketCreationForRegionClear() {
+    bucketCreationLock.writeLock().lock();
+  }
+
+  protected void unLockBucketCreationForRegionClear() {
```
jujoramos, line 987: Should probably be `unlockBucketCreationForRegionClear`.

geode-core/src/main/java/org/apache/geode/internal/cache/partitioned/RegionAdvisor.java
```diff
@@ -834,7 +834,7 @@ public InternalDistributedMember adviseFixedPrimaryPartitionDataStore(final int
     });
   }
 
-  Set adviseAllServersWithInterest() {
+  public Set adviseAllServersWithInterest() {
```
jujoramos, line 837: No need to make it `public`, it's only invoked from `PRTombstoneMessage`, which belongs to the same package.

geode-core/src/distributedTest/java/org/apache/geode/cache/ReplicateCacheListenerDistributedTest.java
```diff
@@ -302,4 +349,64 @@ public void afterDestroy(final EntryEvent<String, Integer> event) {
       errorCollector.checkThat(event.getNewValue(), nullValue());
     }
   }
+
+  protected class ClearCountingCacheListener extends BaseCacheListener {
```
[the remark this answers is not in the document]
agingade: It looks like from original implementation, along with after* event, the tests also verifies certain states with different operation. You are right there is opportunity to clean this test and refactor. I will try to file a separate ticket for this.

geode-core/src/distributedTest/java/org/apache/geode/cache/ReplicateCacheListenerDistributedTest.java
```diff
@@ -302,4 +349,64 @@ public void afterDestroy(final EntryEvent<String, Integer> event) {
...
+
+      if (!event.getRegion().getAttributes().getDataPolicy().withPartitioning()) {
+        if (event.isOriginRemote()) {
+          errorCollector.checkThat(event.getDistributedMember(),
+              not(cacheRule.getSystem().getDistributedMember()));
+        } else {
+          errorCollector.checkThat(event.getDistributedMember(),
+              equalTo(cacheRule.getSystem().getDistributedMember()));
+        }
+      }
```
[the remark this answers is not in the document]
agingade: In the tests, i have seen both types of implementation, where keeping duplicate code in the test, so that its easier to understand the test and modify that specific test. In this case, i will keep it separate and as i commented for Donal's earlier review we can come back to this test and refactor it.

geode-core/src/main/java/org/apache/geode/internal/cache/ClearPartitionedRegion.java
```diff
@@ -0,0 +1,379 @@
...
+  void clearRegion(RegionEventImpl regionEvent, boolean cacheWrite,
+      RegionVersionVector vector) {
+    clearRegionLocal(regionEvent, cacheWrite, null);
+    sendClearRegionMessage(regionEvent,
+        ClearPartitionedRegionMessage.OperationType.OP_PR_CLEAR);
+  }
+
+  private void waitForPrimary() {
+    boolean retry;
+    int retryTime = 2 * 60 * 1000 /* partitionedRegion.getRetryTimeout() */;
```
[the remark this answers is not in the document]
agingade: Are you suggesting we should wait for some notification of status change. My initial thinking was that, but could not find a way to do it. Also, this is similar to other wait approach used in the Partitioned region.

geode-core/src/main/java/org/apache/geode/internal/cache/ClearPartitionedRegion.java
```diff
@@ -0,0 +1,379 @@
...
+            for (BucketRegion localPrimaryBucketRegion : partitionedRegion.getDataStore()
+                .getAllLocalPrimaryBucketRegions()) {
+              if (localPrimaryBucketRegion.size() > 0) {
+                localPrimaryBucketRegion.clear();
+              }
+            }
+            if (membershipchange) {
+              membershipchange = false;
+              retry = true;
+              waitForPrimary();
```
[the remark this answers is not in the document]
agingade: The retry should happen with the membership change. We know or could estimate approximate time it takes to elect a secondary to primary when there is a member departure; the reason it is in the waitForPrimary. If we encounter timeout with clear; we need to adjust the timeout; similar approach is used other time consuming operation.

geode-core/src/main/java/org/apache/geode/internal/cache/ClearPartitionedRegion.java
agingade: Yes...This is to guard against member departure thread trying to unlock.

geode-core/src/main/java/org/apache/geode/internal/cache/ClearPartitionedRegion.java
agingade: This is added check to see if the member is still in the membership view.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionClear.java
```diff
@@ -30,18 +31,22 @@
...
 import org.apache.geode.logging.internal.log4j.api.LogService;
 
-public class ClearPartitionedRegion {
+public class PartitionedRegionClear {
 
   private static final Logger logger = LogService.getLogger();
 
+  private static final String CLEAR_OPERATION = "_clearOperation";
+
+  private final int retryTime = 2 * 60 * 1000 /* partitionedRegion.getRetryTimeout() */;
```
DonalEvans, line 40: This timeout still seems arbitrary. Is there a system property or default timeout that we could reference here instead of hard-coding the value? Why can't we use the value of partitionedRegion.getRetryTimeout() and set it in the constructor? Having a custom timeout just for the clear operation that is different from other region operations seems inconsistent.

geode-core/src/main/java/org/apache/geode/internal/cache/InternalRegion.java
```diff
@@ -459,4 +459,7 @@ VersionedObjectList basicRemoveAll(Collection<Object> keys,
    * @return true if synchronization should be attempted
    */
   boolean shouldSyncForCrashedMember(InternalDistributedMember id);
+
+  void clearRegionLocally(RegionEventImpl regionEvent, boolean cacheWrite,
```
gesterzhou, line 463: I wonder if we should change the API here? Will that impact our backward compatibility?

Do we have workaround not to change InternalRegion.java?

gesterzhou: I remove the code changes in InternalRegion, change the clearRegionLocally into public, then the code can compile.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionClear.java
gesterzhou: Our current clear timeout is actually determined by rvv domination timeout, i.e. 5 seconds. 

If a clear took more than 5 seconds, that means a rvv domination timeout happened. Something must have been wrong.

geode-core/src/main/java/org/apache/geode/internal/cache/InternalRegion.java
agingade: Not sure, what you meant here...InternalRegion is an interface and I added method in that to have access of it in the implemented classes.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionClear.java
agingade: The "partitionedRegion.getRetryTimeout()" is an hour long. In clear operation case, it doesn't make sense to wait for so long. Instead we throw partial clear exception and letting user to handle that.

agingade: Gester, This is the timeout used to wait for secondary buckets to become primary...Looking at the code and the test failures i have seen, its indeterministic and there is no easy way to wait and notify on this action.

agingade: @DonalEvans Also, adding system property, it doesn't look like it will be of any help to application and I am not convinced that this needs a system property. In this case if we are unable to find Primary bucket getting elected in a given time we are throwing partial clear exception; based on which user can take further action, like retry.

geode-core/src/main/java/org/apache/geode/internal/cache/ClearPartitionedRegion.java
```diff
@@ -0,0 +1,379 @@
...
+          doAfterClear(regionEvent);
+        } finally {
+          partitionedRegion.getDataStore().unLockBucketCreationForRegionClear();
+        }
+      } else {
+        // Non data-store with client queue and listener
+        doAfterClear(regionEvent);
+      }
+    }
+  }
```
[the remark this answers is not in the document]
agingade: In the if case, it is done under lock. If its moved out of there, it will be executed outside the lock.

geode-core/src/main/java/org/apache/geode/cache/PartitionedRegionPartialClearException.java
```diff
@@ -0,0 +1,40 @@
...
+ * attempts.
+ *
+ * @since GemFire 5.1
+ */
+public class PartitionedRegionPartialClearException extends CacheRuntimeException {
+
+  public PartitionedRegionPartialClearException() {
+    super();
+  }
+
```
gesterzhou, line 28: This exception is a placeholder. I have not seen a place to throw it. Let another story to throw and handle the exception.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionClear.java
DonalEvans: @agingade I see. So this timeout is kind of unique to this step in the clear operation and so isn't needed/tracked anywhere else.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithExpirationDUnitTest.java
```diff
@@ -281,6 +283,19 @@ private void registerVMKillerAsCacheWriter(List<VM> vmsToBounce) {
     }));
   }
```
gesterzhou, line 285: I don't know what the code change here caused the test PartitionedRegionClearWithExpirationDUnitTest to take 72 minutes to finish, while it took 17 minutes(still quite long) in GEODE-7665 branch

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
pivotal-eshu: clear as a region operations can not be executed under transaction. It will throw UnsupportedOperation.

geode-core/src/distributedTest/java/org/apache/geode/cache/PRCacheListenerDistributedTest.java
```diff
@@ -79,4 +118,206 @@ protected int expectedInvalidates() {
...
   }
+
+  @Test
+  public void afterRegionDestroyIsInvokedInEveryMember() {
+    CacheListener<String, Integer> listener = new RegionDestroyCountingCacheListener();
+    Region<String, Integer> region = createRegion(regionName, listener);
+
+    for (int i = 0; i < getVMCount(); i++) {
+      getVM(i).invoke(() -> {
+        withData(createRegion(regionName, listener));
```
gesterzhou, line 129: Why you did not use withData(region) here?

geode-core/src/distributedTest/java/org/apache/geode/cache/PRCacheListenerDistributedTest.java
```diff
@@ -79,4 +118,206 @@ protected int expectedInvalidates() {
...
+  }
+
+  @Test
+  public void afterRegionDestroyIsInvokedOnNodeWithListener() {
+    CacheListener<String, Integer> listener = new RegionDestroyCountingCacheListener();
+    Region<String, Integer> region = createRegion(regionName, listener);
+
+    for (int i = 0; i < getVMCount(); i++) {
+      getVM(i).invoke(() -> {
+        withData(createRegion(regionName, null));
```
gesterzhou, line 145: same as previous comment

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionAfterClearNotificationDUnitTest.java
```diff
@@ -0,0 +1,379 @@
...
+
+  @Test
+  public void invokeClearOnDataStoreAndVerifyListenerCount() {
+    accessor.invoke(() -> feed(false));
+    verifyServerRegionSize(NUM_ENTRIES);
+
+    dataStore1.invoke(() -> getRegion(false).clear());
+
+    verifyServerRegionSize(0);
+    verifyCacheListenerTriggerCount(dataStore1);
```
gesterzhou, line 195: it's better to add following line here:
    verifyCacheListenerTriggerCount(accessor);

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionAfterClearNotificationDUnitTest.java
```diff
@@ -0,0 +1,379 @@
...
+  }
+
+  @Test
+  public void invokeClearOnAccessorAndVerifyListenerCount() {
+    accessor.invoke(() -> feed(false));
+    verifyServerRegionSize(NUM_ENTRIES);
+    accessor.invoke(() -> getRegion(false).clear());
+    verifyServerRegionSize(0);
+    verifyCacheListenerTriggerCount(accessor);
```
gesterzhou, line 204: It's better to add following line here:
verifyCacheListenerTriggerCount(dataStore1);

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionAfterClearNotificationDUnitTest.java
```diff
@@ -0,0 +1,379 @@
...
+    client1.invoke(() -> getRegion(true).clear());
+
+    verifyServerRegionSize(0);
+    verifyClientRegionSize(0);
+    verifyCacheListenerTriggerCount(null);
+  }
+
+  @Test
+  public void invokeClearFromClientWithAccessorAsServer() {
+    dataStore1.invoke(this::stopServers);
```
gesterzhou, line 225: These 3 lines can be removed.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
agingade: We have ticket/story for transaction and PR.clear. Lets revisit this as part of that story.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionClear.java
agingade: Thats right.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithExpirationDUnitTest.java
agingade: I have run this on machine many a times, and it finishes in 11 or 12 min. I am guessing it could be an environment issue.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionAfterClearNotificationDUnitTest.java
agingade: This is to make sure clients only connect to accessor and not any other server. And to force client HAQueue to be created on accessor.

geode-core/src/distributedTest/java/org/apache/geode/cache/PRCacheListenerDistributedTest.java
```diff
@@ -79,4 +118,206 @@ protected int expectedInvalidates() {
...
+      });
+    }
+
+    region.destroyRegion();
+
+    assertThat(sharedCountersRule.getTotal(REGION_DESTROY)).isEqualTo(1);
+  }
+
+  @Test
+  public void afterRegionDestroyIsInvokedNonOnNonAccessor() {
```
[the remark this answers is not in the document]
agingade: Good catch. Will be addressed in the next checkin.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionPersistentClearDUnitTest.java
```diff
@@ -0,0 +1,26 @@
...
+package org.apache.geode.internal.cache;
+
+
+
+import org.apache.geode.cache.RegionShortcut;
+
+public class PartitionedRegionPersistentClearDUnitTest extends PartitionedRegionClearDUnitTest {
+
+  protected RegionShortcut getRegionShortCut() {
+    return RegionShortcut.PARTITION_REDUNDANT_PERSISTENT;
```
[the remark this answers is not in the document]
agingade: If you are referring to PartitionedRegionPersistentClearDUnitTest, this PR did not add that test.

geode-core/src/distributedTest/java/org/apache/geode/cache/ReplicateCacheListenerDistributedTest.java
```diff
@@ -283,7 +325,12 @@ public void afterInvalidate(final EntryEvent<String, Integer> event) {
 
     @Override
     public void afterCreate(final EntryEvent<String, Integer> event) {
-      // ignore
+      sharedCountersRule.increment(CREATES);
```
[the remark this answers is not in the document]
agingade: This is to override the unsupported exception in base cache listener.

geode-core/src/distributedTest/java/org/apache/geode/cache/ReplicateCacheListenerDistributedTest.java
```diff
@@ -283,7 +325,12 @@ public void afterInvalidate(final EntryEvent<String, Integer> event) {
...
+    }
+
+    @Override
+    public void afterUpdate(final EntryEvent<String, Integer> event) {
+      sharedCountersRule.increment(UPDATES);
```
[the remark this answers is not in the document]
agingade: This is to override the unsupported exception in base cache listener.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionAfterClearNotificationDUnitTest.java
```diff
@@ -0,0 +1,379 @@
...
+
+  protected int locatorPort;
+  protected MemberVM locator;
+  protected MemberVM dataStore1, dataStore2, dataStore3, accessor;
+  protected ClientVM client1, client2;
+
+  private static volatile DUnitBlackboard blackboard;
+
+  @Rule
+  public ClusterStartupRule cluster = new ClusterStartupRule(7);
```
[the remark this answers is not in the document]
agingade: Copied this test from an existing test.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionAfterClearNotificationDUnitTest.java
```diff
@@ -0,0 +1,379 @@
...
+      if (shortcut == RegionShortcut.PARTITION_PERSISTENT) {
+        shortcut = RegionShortcut.PARTITION;
+      } else if (shortcut == RegionShortcut.PARTITION_PERSISTENT_OVERFLOW) {
+        shortcut = RegionShortcut.PARTITION_OVERFLOW;
+      } else if (shortcut == RegionShortcut.PARTITION_REDUNDANT_PERSISTENT) {
+        shortcut = RegionShortcut.PARTITION_REDUNDANT;
+      } else if (shortcut == RegionShortcut.PARTITION_REDUNDANT_PERSISTENT_OVERFLOW) {
+        shortcut = RegionShortcut.PARTITION_REDUNDANT_OVERFLOW;
+      }
+    }
```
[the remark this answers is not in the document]
agingade: Good catch :)

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearDUnitTest.java
```diff
@@ -0,0 +1,400 @@
...
+
+public class PartitionedRegionClearDUnitTest implements Serializable {
+  protected static final String REGION_NAME = "testPR";
+  protected static final int TOTAL_BUCKET_NUM = 10;
+  protected static final int NUM_ENTRIES = 1000;
+
+  protected int locatorPort;
+  protected MemberVM locator;
+  protected MemberVM dataStore1, dataStore2, dataStore3, accessor;
+  protected ClientVM client1, client2;
```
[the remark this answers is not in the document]
agingade: Got the same comment from Jinxia...Fixed in the new checkin.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionAfterClearNotificationDUnitTest.java
```diff
@@ -0,0 +1,379 @@
...
+    dataStore2.invoke(this::initDataStore);
+    dataStore3.invoke(this::initDataStore);
+    accessor.invoke(this::initAccessor);
+
+    getBlackboard().initBlackboard();
+  }
+
+  protected RegionShortcut getRegionShortCut() {
+    return RegionShortcut.PARTITION_REDUNDANT;
+  }
```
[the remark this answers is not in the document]
agingade: Yes...I will it as is now.

geode-core/src/main/java/org/apache/geode/cache/Region.java
```diff
@@ -1303,6 +1303,10 @@ Object selectValue(String queryPredicate) throws FunctionDomainException, TypeMi
    * @see CacheListener#afterRegionClear
    * @see CacheWriter#beforeRegionClear
    * @throws UnsupportedOperationException If the region is a partitioned region
+   * @throws PartitionedRegionPartialClearException when data is partially cleared on partitioned
+   *         region. Its caller responsibility to handle the partial data clear either by retrying
+   *         the clear
+   *         operation or continue working with the partially cleared partitioned region.
```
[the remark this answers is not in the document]
agingade: Fixed the first one.
I wanted highlight the consequence partial clear and how it can be addressed by the caller.

PR #4988, opened 2020-04-23
GEODE-7851: Add slf4j implementation to Pulse

demery-pivotal:
Pulse was not logging because its war file had no slf4j implementation.
This commit adds the log4j2 implementation to the war file.

Authored-by: Dale Emery <demery@pivotal.io>

### For all changes:

### Note:

### DIFF
--- a/geode-pulse/build.gradle
+++ b/geode-pulse/build.gradle
@@ -92,11 +92,7 @@ dependencies {
 
   runtimeOnly('org.springframework:spring-expression')
 
-  // added only to ensure common version with other geode modules
-  runtimeOnly('org.slf4j:slf4j-api')
-  runtimeOnly('org.apache.logging.log4j:log4j-slf4j-impl') {
-    exclude module: 'slf4j-api'
-  }
+  runtimeOnly('org.apache.logging.log4j:log4j-slf4j-impl')
 
   providedCompile('commons-logging:commons-logging')
 
--- a/geode-pulse/build.gradle
+++ b/geode-pulse/build.gradle
@@ -50,7 +50,6 @@ dependencies {
   // Needed to fully use log4j instead of commons-logging.
   implementation('org.apache.logging.log4j:log4j-jcl')
   implementation('org.apache.logging.log4j:log4j-api')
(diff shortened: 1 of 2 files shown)


jmelchio:
> we probably need to add a test to assert the existence of the pulse.log in the geode-assembly module. Will help you add one tomorrow.

It appears that the `pulse.log` file is automatically created in the root directory of the locator.

geode-pulse/build.gradle
```diff
@@ -95,6 +94,9 @@ dependencies {
 
   // added only to ensure common version with other geode modules
   runtimeOnly('org.slf4j:slf4j-api')
+  runtimeOnly('org.apache.logging.log4j:log4j-slf4j-impl') {
+    exclude module: 'slf4j-api'
```
pivotal-jbarrett, line 98: Why are we including the slf4j above and excluding it below. If impl is runtime only then all its deps will also be runtime only. We should just collapse this into a single include.

jinmeiliao: I believe by declaring `runtimeOnly('org.slf4j:slf4j-api')` explicitly will use the version we chose specifically, if we only do `runtimeOnly('org.apache.logging.log4j:log4j-slf4j-impl')`, then the version will depend on whatever version log4j-slf4j-impl uses.

doing this: 
runtimeOnly('org.slf4j:slf4j-api')
runtimeOnly('org.apache.logging.log4j:log4j-slf4j-impl')

will end up using the versions we declare for both jars.

If I understand how gradle dependency works.

demery-pivotal: I'll declare both dependencies explicitly and remove the exclusion.

pivotal-jbarrett: Our BOM binds the version regardless of how it is imported. You can verify versions with ‘dependencies’ target. https://docs.gradle.org/current/userguide/viewing_debugging_dependencies.html

demery-pivotal: With or without the explicit dependency on slf4j-api, the scan shows that we get version 2.13.1. So I'll leave it out.

PR #4989, opened 2020-04-23
GEODE 8014: Sets and Hashes should be deleted once they are empty

sabbeyPivotal:
Redis completely removes keys for sets and hashes once they are empty.  Currently the implementation leaves a key for an empty set/hash once the last element is removed.

### DIFF
--- a/geode-redis/src/main/java/org/apache/geode/redis/internal/executor/hash/GeodeRedisHashSynchronized.java
+++ b/geode-redis/src/main/java/org/apache/geode/redis/internal/executor/hash/GeodeRedisHashSynchronized.java
@@ -90,7 +90,7 @@ class GeodeRedisHashSynchronized implements RedisHash {
       return newHash;
     });
 
-    if (hgetall().size() == 0) {
+    if (hgetall().isEmpty()) {
       RedisDataType type = context.getKeyRegistrar().getType(key);
       context.getRegionProvider().removeKey(key, type);
     }
--- a/geode-redis/src/main/java/org/apache/geode/redis/internal/executor/hash/GeodeRedisHashSynchronized.java
+++ b/geode-redis/src/main/java/org/apache/geode/redis/internal/executor/hash/GeodeRedisHashSynchronized.java
@@ -92,7 +92,9 @@ class GeodeRedisHashSynchronized implements RedisHash {
 
     if (hgetall().isEmpty()) {
       RedisDataType type = context.getKeyRegistrar().getType(key);
-      context.getRegionProvider().removeKey(key, type);
+      if (type == RedisDataType.REDIS_HASH) {
+        context.getRegionProvider().removeKey(key, type);
+      }
     }
     return numDeleted.intValue();
   }
--- a/geode-redis/src/main/java/org/apache/geode/redis/internal/executor/hash/GeodeRedisHashSynchronized.java
+++ b/geode-redis/src/main/java/org/apache/geode/redis/internal/executor/hash/GeodeRedisHashSynchronized.java
@@ -89,6 +89,11 @@ class GeodeRedisHashSynchronized implements RedisHash {
       numDeleted.set(oldHash.size() - newHash.size());
       return newHash;
     });
+
+    if (hgetall().size() == 0) {
+      RedisDataType type = context.getKeyRegistrar().getType(key);
+      context.getRegionProvider().removeKey(key, type);
+    }
     return numDeleted.intValue();
   }
 
--- a/geode-redis/src/main/java/org/apache/geode/redis/internal/executor/set/GeodeRedisSetSynchronized.java
+++ b/geode-redis/src/main/java/org/apache/geode/redis/internal/executor/set/GeodeRedisSetSynchronized.java
@@ -61,10 +61,14 @@ class GeodeRedisSetSynchronized implements RedisSet {
       newValue.removeAll(membersToRemove);
       removedCount.set(oldValue.size() - newValue.size());
 
+      // if(newValue.isEmpty()) {
+      // context.getRegionProvider().removeKey(key, context.getKeyRegistrar().getType(key));
+      // }
+
       return newValue;
     });
 
-    if (members().size() == 0) {
+    if (members().isEmpty()) {
       RedisDataType type = context.getKeyRegistrar().getType(key);
       context.getRegionProvider().removeKey(key, type);
     }
--- a/geode-redis/src/main/java/org/apache/geode/redis/internal/executor/set/GeodeRedisSetSynchronized.java
+++ b/geode-redis/src/main/java/org/apache/geode/redis/internal/executor/set/GeodeRedisSetSynchronized.java
@@ -66,7 +66,9 @@ class GeodeRedisSetSynchronized implements RedisSet {
 
     if (members().isEmpty()) {
       RedisDataType type = context.getKeyRegistrar().getType(key);
-      context.getRegionProvider().removeKey(key, type);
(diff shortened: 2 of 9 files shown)


geode-redis/src/main/java/org/apache/geode/redis/internal/executor/hash/GeodeRedisHashSynchronized.java
```diff
@@ -89,6 +89,11 @@ public int hdel(List<ByteArrayWrapper> subList) {
       numDeleted.set(oldHash.size() - newHash.size());
       return newHash;
     });
+
+    if (hgetall().size() == 0) {
```
dschneider-pivotal, line 93: Instead of ".size() == 0" I think ".isEmpty()" is better.

dschneider-pivotal: Calling "hgetall" and then checking the size does not seem thread safe. Couldn't another  thread add to this set and find the existing one we emptied out and add to it right after this thread checks the size and then we will unregister a non-empty hash?
If we do have a concurrency issue I'm still okay with this change since it fixes the non-concurrent case. Maybe we could just add a TODO comment that concurrency needs to be addressed?

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/hash/GeodeRedisHashSynchronized.java
```diff
@@ -89,6 +89,11 @@ public int hdel(List<ByteArrayWrapper> subList) {
...
+      RedisDataType type = context.getKeyRegistrar().getType(key);
```
dschneider-pivotal, line 94: It seems like this code should only unregister if the current type is REDIS_HASH. I thought you were unregistering here but does getRegionProvider().removeKey change the keyRegistar? I will talk with Sarah about this. I might just not understand this code

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/set/GeodeRedisSetSynchronized.java
```diff
@@ -59,8 +60,15 @@ public long srem(Collection<ByteArrayWrapper> membersToRemove) {
       Set<ByteArrayWrapper> newValue = createSet(oldValue);
       newValue.removeAll(membersToRemove);
       removedCount.set(oldValue.size() - newValue.size());
+
       return newValue;
     });
+
+    if (members().size() == 0) {
```
dschneider-pivotal, line 67: Same comments on this block as the above hdel

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/hash/GeodeRedisHashSynchronized.java
sabbeyPivotal: True, I will update to .isEmpty()

sabbeyPivotal: Yeah, I think we just put in the non-concurrent case for now since we are still deciding on an approach (compound key vs. function/delta vs. something else).  I will add a TODO, though.

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/set/GeodeRedisSetSynchronized.java
sabbeyPivotal: I will update and add a TODO.

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/hash/GeodeRedisHashSynchronized.java
sabbeyPivotal: Yes, the removeKey method unregisters the key from the keyRegistrar in addition to deleting it.

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/set/GeodeRedisSetSynchronized.java
```diff
@@ -59,8 +60,19 @@ public long srem(Collection<ByteArrayWrapper> membersToRemove) {
...
+
+      // if(newValue.isEmpty()) {
+      // context.getRegionProvider().removeKey(key, context.getKeyRegistrar().getType(key));
+      // }
+
       return newValue;
     });
+
+    if (members().isEmpty()) {
+      RedisDataType type = context.getKeyRegistrar().getType(key);
```
dschneider-pivotal, line 72: It seems wrong here to refetch the RedisDataType and then blow it away without checking if the type is what we expect. For example here should we check to see if it is == REDIS_SET before calling removeKey?

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/hash/GeodeRedisHashSynchronized.java
```diff
@@ -89,6 +89,11 @@ public int hdel(List<ByteArrayWrapper> subList) {
       numDeleted.set(oldHash.size() - newHash.size());
       return newHash;
     });
+
+    if (hgetall().isEmpty()) {
+      RedisDataType type = context.getKeyRegistrar().getType(key);
```
dschneider-pivotal, line 94: It seems wrong here to refetch the RedisDataType and then blow it away without checking if the type is what we expect. For example here should we check to see if it is == REDIS_HASH before calling removeKey?

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/set/GeodeRedisSetSynchronized.java
```diff
@@ -59,8 +60,19 @@ public long srem(Collection<ByteArrayWrapper> membersToRemove) {
       Set<ByteArrayWrapper> newValue = createSet(oldValue);
       newValue.removeAll(membersToRemove);
       removedCount.set(oldValue.size() - newValue.size());
+
+      // if(newValue.isEmpty()) {
```
dschneider-pivotal, line 64: should this commented out code be removed?

sabbeyPivotal: Yes.  We were experimenting with adding it into compute. Thank you for finding that!

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/set/GeodeRedisSetSynchronized.java
sabbeyPivotal: We actually check the data type in the SREM executor before we create the synchronized set and call SREM.  Do you think we need to check it again?

sabbeyPivotal: We added a test to make sure the error is thrown and check the message.

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/hash/GeodeRedisHashSynchronized.java
sabbeyPivotal: Turns out the type was not getting checked for HDel, so we added a test and updated the HDelExecutor.  Thanks for pointing this out!

PR #4990, opened 2020-04-24
[DO NOT REVIEW]

jujoramos:


geode-core/src/distributedTest/java/org/apache/geode/internal/cache/tier/sockets/PingOpDistributedTest.java
```diff
@@ -0,0 +1,219 @@
...
+
+    asList(server1, server2).forEach(vm -> vm.invoke(() -> {
+      ClientHealthMonitor chm = ClientHealthMonitor.getInstance();
+      assertThat(chm.getClientHeartbeats()).isEmpty();
+    }));
+
+    System.out.println("########## FIRST PING START");
+    client.invoke(() -> executePing(poolName, server1Port, distributedMember2));
+    Long firstHeartbeatServer1 = server1.invoke(this::getSingleHeartBeat);
+    Long firstHeartbeatServer2 = server2.invoke(this::getSingleHeartBeat);
```
alb3rtobr, line 166: @jujoramos There is something I dont get here: if there is a ClientHealthMonitor per server, how it is possible that a ping send for distributedMember2 creates an entry in the ClientHealthMonitor of server1? I have been debugging this and I only see one call in ClientHealthMonitor::receivedPing in server 2 (as expected) after executing the ping (line 164). Could it be a problem of the testing framework? Or am I missing something?

jujoramos: Hey @alb3rtobr 
Sorry for the confusion here, the test `memberShouldCorrectlyRedirectPingMessage` is not finished at all, I was just using it to debug the issue.

alb3rtobr: Ok! Anyway I found a possible cause for the unexpected ping. I think the assert in line 160 is not correct. Sometimes the test fails there, although no ping has been sent yet. The `ServerConnection::processHandshake` method calls `ClientHealthMonitor::registerClient`, which will cause that `ClientHealthMonitor.getClientHeartbeats()` will not be empty. I think sometimes the test is fast enough to pass the assert, and the heartbeat got in line 165 is the one created by `ServerConnection::processHandshake`

jujoramos: Yeah, do not worry about this particular test method, the one showing the issue is `memberShouldNotRedirectPingMessageWhenClientCachedViewIdIsWrong`. You can use the other two as starting points to develop your own tests for the `GEODE-7565`.

PR #4994, opened 2020-04-24
GEODE-8020: buffer corruption in SSL communications

bschuchardt:
revert change in GEODE-6661 that made NioSslEngine use a direct buffer.
This class is not designed to share its buffer with a pool in
BufferPool.  Connection is also modified to use a heap buffer for
reading encrypted SSL packets for consistency.  New tests ensure that
these buffers are the correct type when using SSL or plain sockets.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/tcp/Connection.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/tcp/Connection.java
@@ -1662,8 +1662,8 @@ public class Connection implements Runnable {
         } catch (IOException e) {
           // "Socket closed" check needed for Solaris jdk 1.4.2_08
           if (!isSocketClosed() && !"Socket closed".equalsIgnoreCase(e.getMessage())) {
-            if (logger.isDebugEnabled() && !isIgnorableIOException(e)) {
-              logger.debug("{} io exception for {}", p2pReaderName(), this, e);
+            if (logger.isInfoEnabled() && !isIgnorableIOException(e)) {
+              logger.info("{} io exception for {}", p2pReaderName(), this, e);
             }
             if (e.getMessage().contains("interrupted by a call to WSACancelBlockingCall")) {
               if (logger.isDebugEnabled()) {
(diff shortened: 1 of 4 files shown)


geode-core/src/main/java/org/apache/geode/internal/tcp/Connection.java
```diff
@@ -1763,9 +1763,13 @@ private static boolean isIgnorableIOException(Exception e) {
     }
 
     msg = msg.toLowerCase();
-    return msg.contains("forcibly closed")
-        || msg.contains("reset by peer")
-        || msg.contains("connection reset");
+
+    if (e instanceof SSLException && msg.contains("status = closed")) {
```
pivotal-jbarrett, line 1767: The commit message references reverting the direct buffer changes. Is this change intentional because it doesn't look like part of the reverted commit?

geode-core/src/main/java/org/apache/geode/internal/tcp/Connection.java
```diff
@@ -1662,8 +1662,8 @@ private void readMessages() {
         } catch (IOException e) {
           // "Socket closed" check needed for Solaris jdk 1.4.2_08
           if (!isSocketClosed() && !"Socket closed".equalsIgnoreCase(e.getMessage())) {
-            if (logger.isDebugEnabled() && !isIgnorableIOException(e)) {
-              logger.debug("{} io exception for {}", p2pReaderName(), this, e);
+            if (logger.isInfoEnabled() && !isIgnorableIOException(e)) {
```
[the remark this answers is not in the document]
bschuchardt: yes!  I only saw the SSLException that explained the message-loss when debug-level logging was enabled.  With this log level change I had to augment isIgnorableException to filter out some exceptions being caught here when the socket was closed.

geode-core/src/main/java/org/apache/geode/internal/tcp/Connection.java
bschuchardt: I'm confused by this comment because I very clearly included changing back to non-direct buffers in Connection and NioSslEngine in this commit.

PR #5009, opened 2020-04-27
GEODE-8048: change redis sets to use functions and deltas

prettyClouds:

### For all changes:

### Note:


geode-redis/src/main/java/org/apache/geode/redis/internal/executor/set/DeltaSet.java
```diff
@@ -0,0 +1,261 @@
...
+      }
+    }
+    return membersRemoved;
+  }
+
+  /**
+   * This exception is thrown if a modification fails because some other
+   * thread changed what is stored in the region.
+   */
+  static class RetryDueToConcurrentModification extends RuntimeException {
```
prettyClouds, line 237: This exception name tells the User what to do, i.e. "retry", instead of just stating what happened and letting the user decide how to handle it.

Maybe just throwing a ConcurrentModificationException, or something like RegionUnexpectedModificationException

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/set/DeltaSet.java
```diff
@@ -0,0 +1,261 @@
...
+ * Currently the only "correct" methods on this class are:
+ * members, delete, customAddAll, customRemoveAll, and the
+ * serialization methods.
+ */
+public class DeltaSet implements Delta, DataSerializable {
+
+  public static long sadd(Region<ByteArrayWrapper, DeltaSet> region,
+      ByteArrayWrapper key,
+      ArrayList<ByteArrayWrapper> membersToAdd) {
+    while (true) {
```
prettyClouds, line 48: If we synchronize "lower" in the stack, can we avoid the complexity of looping and throwing exceptions. Can we somehow guarantee that on a given key, only one of these operations is happening simultaneously? Maybe an exclusive lock on the key?

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/set/SaddFunction.java
```diff
@@ -0,0 +1,54 @@
...
+
+import java.util.ArrayList;
+
+import org.apache.geode.cache.Region;
+import org.apache.geode.cache.execute.Function;
+import org.apache.geode.cache.execute.FunctionContext;
+import org.apache.geode.internal.cache.execute.RegionFunctionContextImpl;
+import org.apache.geode.redis.internal.ByteArrayWrapper;
+
+class SaddFunction implements Function {
```
prettyClouds, line 26: When we first implemented this a few weeks back, Jens suggested having just one remote function that encapsulated all the Set commands, and the execute could just switch on some parameter.  I don't have a strong feeling either way, but I can't remember Jens' rationale...just wanted to capture the thought here, before it gets lost.

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/set/SaddFunction.java
```diff
@@ -0,0 +1,54 @@
...
+
+  public static final String ID = "SADD_FUNCTION";
+
+  @SuppressWarnings("unchecked")
```
prettyClouds, line 30: How do we get rid of these? Adding a type to the class declaration, like `... implements Function<Object>`?

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/set/DeltaSet.java
prettyClouds: We could even use some kind of custom read-write lock...sadds should be able to happen in parallel, and srems should be able to happen in parallel exclusive of sadds, and dels just have to be exclusive of everything.

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/set/SaddFunction.java
dschneider-pivotal: I like that idea, just to reduce the number of functions

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/set/DeltaSet.java
dschneider-pivotal: Keep in mind this exception is for our internal implementation; not the user. It is possible we could refactor this code so that we don't need to throw on exception to force a retry.

PR #5014, opened 2020-04-28
GEODE-8035: Parallel Disk Store Recovery when Cluster Restarts

jchen21:

### For all changes:

### Note:


geode-core/src/main/java/org/apache/geode/internal/cache/xmlcache/CacheCreation.java
```diff
@@ -521,12 +521,12 @@ void create(InternalCache cache)
 
     cache.initializePdxRegistry();
 
-    for (DiskStore diskStore : diskStores.values()) {
+    diskStores.values().parallelStream().forEach(diskStore -> {
```
DonalEvans, line 524: Would it be possible to put a bound on the number of threads used here? In cases with large numbers of disk stores this might cause resource issues if unbounded, but maybe we never have so many disk stores that it will matter.

geode-core/src/main/java/org/apache/geode/internal/cache/DiskStoreFactoryImpl.java
```diff
@@ -144,7 +144,10 @@ public DiskStore create(String name) {
     // As a simple fix for 41290, only allow one DiskStore to be created
     // at a time per cache by syncing on the cache.
```
DonalEvans, line 145: This comment is no longer entirely accurate.

geode-core/src/main/java/org/apache/geode/internal/cache/GemFireCacheImpl.java
```diff
@@ -1086,6 +1088,32 @@ public static GemFireCacheImpl getForPdx(String reason) {
...
+      } catch (InterruptedException e) {
+        throw new InternalGemFireError(e);
+      }
+    }
+  }
+
+  public void unlockDiskStore(String diskStoreName) {
+    if (diskStoreLatches.get(diskStoreName) != null) {
+      diskStoreLatches.get(diskStoreName).countDown();
+    }
```
DonalEvans, line 1114: I think there is a small possibility of a race condition causing an NPE here if the latch associated with `diskStoreName` is removed after the first `get()` call. Would it be possible to change this method to use only one `get()` call?

geode-core/src/main/java/org/apache/geode/internal/cache/xmlcache/CacheCreation.java
gesterzhou: In Geode use case, it won't be large number of diskstore. The worst case is less than region number.

geode-core/src/main/java/org/apache/geode/internal/cache/DiskStoreFactoryImpl.java
```diff
@@ -165,6 +168,10 @@ public DiskStore create(String name) {
           creation.addDiskStore(result);
         }
       }
+    } finally {
+      if (this.cache instanceof GemFireCacheImpl) {
+        ((GemFireCacheImpl) this.cache).unlockDiskStore(name);
+      }
```
jujoramos, line 174: Can you the methods `lockDiskStore` and `unlockDiskStore` to the `InternalCache` interface instead?. It's cleaner as we won't need to cast to `GemFireCacheImpl`, besides that, we could use these methods in the future from other parts of the code.

geode-core/src/main/java/org/apache/geode/internal/cache/GemFireCacheImpl.java
```diff
@@ -1086,6 +1088,32 @@ public static GemFireCacheImpl getForPdx(String reason) {
...
+        }
+      }
+    } else {
+      try {
+        countDownLatch.await();
+      } catch (InterruptedException e) {
+        throw new InternalGemFireError(e);
+      }
+    }
+  }
```
jujoramos, line 1109: The above can be simplified with something like:

```
public void lockDiskStore(String diskStoreName) {
	CountDownLatch latch = JavaWorkarounds.computeIfAbsent(diskStoreLatches, diskStoreName, k -> new CountDownLatch(1));

	try {
		countDownLatch.await();
		throw new InternalGemFireError(e);
	}
}
```

geode-core/src/main/java/org/apache/geode/internal/cache/GemFireCacheImpl.java
jujoramos: I second @DonalEvans's comments here, could we use `computeIfPresent` or something similar?.
As a side note, if somebody invokes `unlockDiskStore` before calling `lockDiskStore` for exactly the same `diskName` first (aka the entry is `null`), shouldn't we throw an exception directly as the caller is using the API in a wrong way?.

geode-core/src/main/java/org/apache/geode/internal/cache/GemFireCacheImpl.java
jchen21: In your simplified code, all thread will `await()`, which is not correct.

geode-core/src/main/java/org/apache/geode/internal/cache/GemFireCacheImpl.java
jchen21: Agree. There could be an NPE.

geode-core/src/main/java/org/apache/geode/internal/cache/xmlcache/CacheCreation.java
jchen21: Parallel stream by default uses [`ForkJoinPool`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ForkJoinPool.html) with parallelism equals the number of processors.

geode-core/src/main/java/org/apache/geode/internal/cache/DiskStoreFactoryImpl.java
jchen21: If `lockDiskStore`and `unlockDiskStore` are moved to `InternalCache`, all the classes that implement the interface has to implement `lockDiskStore`and `unlockDiskStore`. However, only `GemFireCacheImpl` actually needs `lockDiskStore`and `unlockDiskStore`. I don't think `CacheCreation` and `InternalCacheForClientAccess` need them.

geode-core/src/main/java/org/apache/geode/internal/cache/DiskStoreFactoryImpl.java
jchen21: Agree.

geode-core/src/main/java/org/apache/geode/internal/cache/DiskStoreFactoryImpl.java
jchen21: OK. Let's use the `default` implementation in `InternalCache`.

geode-core/src/main/java/org/apache/geode/internal/cache/xmlcache/CacheCreation.java
```diff
@@ -521,12 +521,12 @@ void create(InternalCache cache)
...
       DiskStoreAttributesCreation creation = (DiskStoreAttributesCreation) diskStore;
       if (creation != pdxRegDSC) {
         createDiskStore(creation, cache);
       }
-    }
+    });
```
agingade, line 529: How is the exception handled?

geode-core/src/main/java/org/apache/geode/internal/cache/GemFireCacheImpl.java
```diff
@@ -1086,6 +1088,35 @@ public static GemFireCacheImpl getForPdx(String reason) {
     clientMetadataService = clientMetadataServiceFactory.apply(this);
   }
 
+  @Override
+  public void lockDiskStore(String diskStoreName) {
+    CountDownLatch countDownLatch = diskStoreLatches.get(diskStoreName);
```
agingade, line 1093: Is get() required, when using putIfAbsent()? putIfAbsent will return either null or old value.

geode-core/src/main/java/org/apache/geode/internal/cache/DiskStoreFactoryImpl.java
```diff
@@ -141,10 +141,9 @@ DiskStoreImpl createOwnedByRegion(String name, boolean isOwnedByPR,
   @Override
   public DiskStore create(String name) {
     this.attrs.name = name;
-    // As a simple fix for 41290, only allow one DiskStore to be created
-    // at a time per cache by syncing on the cache.
     DiskStore result;
-    synchronized (this.cache) {
+    try {
```
agingade, line 145: Based on the previous comment for 41290; it seems there was a need to create disk-store one at a time. 
From the comment its not clear, if its when the disk-stores are created through java API or cache xml (ClusterConfig).
Why do we need to change the synchronization...Here the disk-stores are just added into the map and reading from the disk happens in the CacheCreation.java.

geode-core/src/main/java/org/apache/geode/internal/cache/xmlcache/CacheCreation.java
agingade: Need to add unit/integration tests for both cases, with multiple and single thread trying to create disk-stores.

geode-core/src/main/java/org/apache/geode/internal/cache/DiskStoreFactoryImpl.java
jchen21: Before I created this pull request, I actually looked into 41290, which was created more than 10 years ago. It was a totally different implementation. 41290 is no longer a concern.

We do need to remove the synchronization, in order to make the disk store recovery parallel. This synchronized block not only adds disk stores into the map, but also reads the disks, please see `initializeDiskStore(dsi);` in line 157. `initializeDiskStore()` eventually invokes `Oplog.recoverCrf()` and `Oplog.readKrf()`. Without removing the synchronization, the disk store can only be recovered one at a time.

geode-core/src/main/java/org/apache/geode/internal/cache/GemFireCacheImpl.java
jchen21: Good point! I will make the change per your suggestion.

geode-core/src/main/java/org/apache/geode/internal/cache/xmlcache/CacheCreation.java
jchen21: I am not sure which exception you are talking about. I don't see `parallelStream()` and `forEach()` introduce new exception, based on the Javadoc.

For test coverage, there are quite a few existing tests that already cover the single thread or multiple disk stores cases. We don't have to add new tests for that.

For multi-thread test, it is a very tricky one. The purpose of this pull request is to introduce parallel disk recovery and improve the performance. We can assert that the stream should be parallel using `isParallel()`. Or we can scan the logs and make sure there are multiple thread IDs that recover the disk stores. But it doesn't serve the purpose of testing performance improvement. It doesn't guarantee that disk recovery is parallel, even if we can assert that the stream is parallel or there are multiple threads that recover the disk store. For example, if we keep the synchronized block in `DiskStoreFactoryImple` line 147, the disk stores recovery will be sequential, even though there are multiple threads recovering the disk stores.

geode-core/src/main/java/org/apache/geode/internal/cache/GemFireCacheImpl.java
jchen21: @jujoramos if the entry is `null`, we can't tell whether it is because the disk store is removed, or somebody invokes `unlockDiskStore` before `lockDiskStore`. We don't want to throw an exception, when the entry is `null` because the disk store is removed.

geode-core/src/main/java/org/apache/geode/internal/cache/xmlcache/CacheCreation.java
agingade: Based on the DiskStoreFactory.create() it can throw IllegalArgumentException; probably IOException based on the disk read failures. In case there are multiple disk-store failures, we will leave the parallel stream to throw the exception or we will capture all the exception and throw a single exception with all the disk=store names in it...

I was thinking from unit test level; if there are already tests covering the new code. No need of new tests.

geode-core/src/main/java/org/apache/geode/internal/cache/GemFireCacheImpl.java
```diff
@@ -1086,6 +1088,27 @@ public static GemFireCacheImpl getForPdx(String reason) {
     clientMetadataService = clientMetadataServiceFactory.apply(this);
   }
 
+  @Override
+  public void lockDiskStore(String diskStoreName) {
+    CountDownLatch countDownLatch =
+        diskStoreLatches.putIfAbsent(diskStoreName, new CountDownLatch(1));
+    if (countDownLatch != null) {
+      try {
+        countDownLatch.await();
```
kirklund, line 1097: `putIfAbsent` returns the old value (if any) or null. Are you trying to await() on the old Latch or the new one you just put in the ConcurrentMap?

jchen21: It will `await()` on the old latch. More specifically, the first thread will put a new `CountDownLatch` in the map. The other threads will `await()` on the latch that the first thread put in the map.

geode-core/src/test/java/org/apache/geode/internal/cache/GemFireCacheImplTest.java
```diff
@@ -620,6 +625,40 @@ public void getCacheServers_isCanonical() {
...
+          if (unlockResult) {
+            nTrue.incrementAndGet();
+          } else {
+            nFalse.incrementAndGet();
+          }
+        }
+      });
+    });
+    executorService.shutdown();
+    executorService.awaitTermination(Long.MAX_VALUE, TimeUnit.NANOSECONDS);
```
kirklund, line 655: If you use the Rule, you shouldn't need this or the shutdown. In general, you should however use `GeodeAwaitility.getTimeout()` instead of `Long.MAX_VALUE`.

jchen21: Will change it.

geode-core/src/test/java/org/apache/geode/internal/cache/GemFireCacheImplTest.java
```diff
@@ -620,6 +625,40 @@ public void getCacheServers_isCanonical() {
         .isSameAs(gemFireCacheImpl.getCacheServers());
   }
 
+  @Test
+  public void testLockDiskStore() throws InterruptedException {
+    int nThread = 10;
+    String diskStoreName = "MyDiskStore";
+    AtomicInteger nTrue = new AtomicInteger();
+    AtomicInteger nFalse = new AtomicInteger();
+    ExecutorService executorService = Executors.newFixedThreadPool(nThread);
```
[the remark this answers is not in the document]
jchen21: I had thought about using `ExecutorServiceRule`. However, the return value of `doLockDiskStore()` is non-deterministic. I can't assert the return value of individual thread. I can only count and assert the number of returning values when all threads are done. I am not sure how `ExecutorServiceRule` will serve this test case.

jchen21:
> Changes look good to me, but I'd really prefer to see one or more new tests or an old one modified if possible.
> 
> I think it's good to add the new methods to InternalCache in order to avoid casting to concrete impl (GemFireCacheImpl), but it would be best to avoid making them default empty methods unless they're part of some sort of deprecated-replacement if that makes sense. I would just remove `default` from them and modify all the code (mostly tests) that implements InternalCache to provide empty implementations of the new methods.

@kirklund I am not sure about the disadvantage of default empty methods in `InternalCache`. I can remove the `default` keyword from `InternalCache` and let `CacheCreation` and `InternalCacheForClientAccess` implement the empty method. No test change needed. I don't see significant difference between these two options.

geode-core/src/main/java/org/apache/geode/internal/cache/xmlcache/CacheCreation.java
jchen21: The system property is added. Tests are added as well.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ParallelDiskStoreRecoveryDUnitTest.java
```diff
@@ -0,0 +1,171 @@
...
+import org.apache.geode.test.junit.rules.ExecutorServiceRule;
+import org.apache.geode.test.junit.rules.GfshCommandRule;
+
+public class ParallelDiskStoreRecoveryDUnitTest {
+
+  @Rule
+  public GfshCommandRule gfsh = new GfshCommandRule();
+
+  @Rule
+  public ClusterStartupRule cluster = new ClusterStartupRule();
```
kirklund, line 40: The Green Team provided a presentation and recommendations for DUnit tests that result in tests that are easy for other developers to maintain. One important recommendation is to use the Geode User APIs directly instead of using any sort of API replacement/wrapper including ClusterStartupRule. See [The Green Report](https://docs.google.com/presentation/d/13FrlS6AJQayi5Iyr_jG3c0kqwxPz-aSkEqBOHRSCj_w/edit?usp=sharing) (if you don't have access please let me know -- I'm happy to share it with anyone).

Basically, if a Rule or Wrapper hides anything about configuring, starting, stopping, or doing anything with Geode then we should not use it because it obscures Geode.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ParallelDiskStoreRecoveryDUnitTest.java
```diff
@@ -0,0 +1,171 @@
...
+
+  @Rule
+  public ExecutorServiceRule executorServiceRule = new ExecutorServiceRule();
+
+  private MemberVM locator, server1, server2;
```
kirklund, line 45: Our coding standard has always disallowed declaration of multiple fields on one line. This may have slipped with Patrick setup spotless for Geode, but there was never a group discussion on the dev-list to get rid of this style requirement.

geode-core/src/test/java/org/apache/geode/internal/cache/GemFireCacheImplTest.java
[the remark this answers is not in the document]
kirklund: I recommend using ExecutorServiceRule for several reasons. Most important is that it always performs tearDown (ie shutdownNow) even if the test blows up before invoking `executorService.shutdown();`

geode-core/src/main/java/org/apache/geode/internal/cache/DiskStoreFactoryImpl.java
```diff
@@ -141,10 +141,9 @@ DiskStoreImpl createOwnedByRegion(String name, boolean isOwnedByPR,
...
+      this.cache.lockDiskStore(name);
```
agingade, line 146: Does this needs to be before "try"? if lock throws exception without locking we may be trying to unlock it...

geode-core/src/main/java/org/apache/geode/distributed/internal/DistributionConfigImpl.java
```diff
@@ -668,6 +668,8 @@
    */
   private int threadMonitorTimeLimit = DEFAULT_THREAD_MONITOR_TIME_LIMIT;
 
+  private boolean parallelDiskStoreRecovery = DEFAULT_PARALLEL_DISK_STORE_RECOVERY;
```
agingade, line 671: Adding it here, makes it a distribution config setup...For system property you need to add in "SystemPropertyHelper". Look into "EVICTION_SCAN_MAX_THREADS" as an example.
Also, tests to verify the system property setting are working as expected....

geode-core/src/test/java/org/apache/geode/internal/cache/GemFireCacheImplTest.java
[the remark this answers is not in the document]
kirklund: Another alternative is to extract your `executorService` as a field so that shutdown or shutdownNow is always invoked even if the test has a failure:
```
private ExecutorService executorService;

@Before
public void setUp() {
  executorService = Executors.newCachedThreadPool();
}

@After
public void tearDown() {
  executorService.shutdownNow();
}
```

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ParallelDiskStoreRecoveryDUnitTest.java
jchen21: Will change it.

geode-core/src/main/java/org/apache/geode/internal/cache/DiskStoreFactoryImpl.java
jchen21: The `unlock` in the finally block does nothing, if the `lock` fails. The `unlock` is essentially a `CountDownLatch.countDown()`. If the current count is greater than zero then it is decremented. If the current count equals zero then nothing happens.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ParallelDiskStoreRecoveryDUnitTest.java
kirklund: Here's The Green Report on the Apache Geode Wiki: https://cwiki.apache.org/confluence/display/GEODE/The+Green+Report

geode-core/src/main/java/org/apache/geode/internal/cache/xmlcache/CacheCreation.java
agingade: Can we move this code block to helper method...The create() is becoming too long...

agingade: How about test where one/two of the recovery thread throws IllegalArgumentException...

jchen21: Refactored the code block to helper method. Done.

jchen21: I am not sure which method throws `IllegalArgumentException`. To the best of my knowledge, the new code I introduced does not throw `IllegalArgumentException`. If existing code throws `IllegalArgumentException`, it is beyond the scope of this JIRA.

geode-core/src/test/java/org/apache/geode/internal/cache/GemFireCacheImplTest.java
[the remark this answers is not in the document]
jchen21: I have changed new test in `GemFireCacheImplTest` to use `ExecutorServiceRule`.

geode-core/src/test/java/org/apache/geode/internal/cache/GemFireCacheImplTest.java
jchen21: Done.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ParallelDiskStoreRecoveryDUnitTest.java
jchen21: Done.

geode-core/src/main/java/org/apache/geode/distributed/internal/DistributionConfigImpl.java
jchen21: This is done, including the new tests.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ParallelDiskStoreRecoveryDUnitTest.java
jchen21: Thanks for the references! I have address all three issues:

1. Removed the use `ClusterStartupRule`.
2. Uses `ExecutorServiceRule` instead of `ExecutorService`.
3. Removed default empty interface methods.

PR #5018, opened 2020-04-29
GEODE-8039: update incorrect versions in LICENSE

onichols-pivotal:


onichols-pivotal:
It seems that anything we distribute that is licensed under Apache 2.0 License does not need to be called out.  However, I found 3 jars that will be in the 1.13 release that are not currently mentioned anywhere in LICENSE, should they be added?

* HikariCP v3.4.2 - Eclipse Public License 1.0
* findbugs-annotations v1.3.9-1 - BSD License (?)
* swagger-annotations v1.5.23 - MIT License

metatype:
> It seems that anything we distribute that is licensed under Apache 2.0 License does not need to be called out. However, I found 3 jars that will be in the 1.13 release that are licensed under some other license but not currently mentioned in LICENSE, should they be added?
> 
> * HikariCP v3.4.2 - Eclipse Public License 1.0
> * findbugs-annotations v1.3.9-1 - BSD License (?)
> * swagger-annotations v1.5.23 - MIT License

HikariCP is Apache though, right?  See https://github.com/brettwooldridge/HikariCP/blob/dev/LICENSE.

Same for findbugs-annotations (note the group):  https://mvnrepository.com/artifact/com.github.stephenc.findbugs/findbugs-annotations

Also for swagger-annotations:  https://mvnrepository.com/artifact/io.swagger/swagger-annotations/1.5.23

onichols-pivotal:
Thanks for catching those mistakes, Anthony.  You are correct, I had googled something else with similar name by mistake.  I also forgot to expand war files to look at their embedded dependencies; I've made several additional corrections and also included the script I created for this review process and an rc pipeline job to run it again before shipping.

onichols-pivotal:
LICENSE currently gets used 5 different places:
* the source distribution
* geode-pulse.war
* geode-web.war
* geode-web-api.war
* geode-web-management.war
It sounds like we should break this up and tailor each one.

rhoughton-pivot:
I'm not a huge fan that we have yet another list of dependencies to manage :(

gradle/java.gradle
```diff
@@ -73,7 +73,7 @@ gradle.taskGraph.whenReady({ graph ->
       }
     }
     jar.metaInf {
-      from("$rootDir/LICENSE")
+      from("$rootDir/geode-assembly/src/main/dist/LICENSE")
```
rhoughton-pivot, line 76: Two things:
1) Why do we have two copies of the `LICENSE` file?
2) Why are they different!?

onichols-pivotal: > 1. Why do we have two copies of the `LICENSE` file?
> 2. Why are they different!?

Thanks for asking, it's not totally intuitive at first.

`LICENSE` is intended to cover Geode's license plus any 3rd party products we redistribute in source form (such as javascript libraries).  This LICENSE file is added to the apache-geode-x.y.z-src.tgz release artifact.

`geode-assembly/src/main/dist/LICENSE` includes everything in LICENSE, and also covers any 3rd party libraries redistributed in apache-geode-x.y.z.tgz (binary) releases.  This license file is added to apache-geode-x.y.z.tgz and also placed in every Geode .jar and .war file.

It turns out, for several releases now we had been accidentally putting the source license instead of the binary license into .jar and .war files, which added further confusion to which license was for which purpose and which one things should be added to.  This PR straightens out all of those issues.

The reason we need separate source and binary licenses might have to do with Apache rules.  Only the source release is an official act of the Apache Foundation.  Geode also provides a compiled binary release as a courtesy only.  Please ask a PMC member for more details on Apache release governance.

LICENSE
```diff
@@ -280,8 +280,6 @@ Apache Geode bundles the following files under the MIT license:
     Foundation and other contributors, http://jquery.org
   - jScrollPane (http://jscrollpane.kelvinluck.com/), Copyright (c) 2010
     Kelvin Luck
-  - matchMedia() polyfill (https://github.com/paulirish/matchMedia.js),
```
metatype, line 283: Why is this removed?  I believe one of the js libraries that we include has bundled this in their source code.  See https://lists.apache.org/thread.html/ef3c70fcae89a05adbc46598439208a2e920d71b4f7062662a814fff%401460088338%40%3Cdev.geode.apache.org%3E.

buildSrc/src/main/groovy/org/apache/geode/gradle/plugins/DependencyConstraints.groovy
```diff
@@ -101,6 +101,7 @@ class DependencyConstraints implements Plugin<Project> {
         api(group: 'com.nimbusds', name:'nimbus-jose-jwt', version:'8.11')
         api(group: 'com.sun.activation', name: 'javax.activation', version: '1.2.0')
         api(group: 'com.sun.istack', name: 'istack-commons-runtime', version: '3.0.11')
+        api(group: 'com.sun.mail', name: 'javax.mail', version: '1.6.2')
```
metatype, line 104: Why is this removed?

geode-assembly/src/main/dist/LICENSE
```diff
@@ -1053,14 +1095,12 @@ Apache Geode bundles the following files under the MIT license:
     Foundation and other contributors, http://jquery.org
   - jScrollPane (http://jscrollpane.kelvinluck.com/), Copyright (c) 2010
     Kelvin Luck
-  - matchMedia() polyfill (https://github.com/paulirish/matchMedia.js),
```
metatype, line 1056: This should stay, see above.

geode-assembly/src/main/dist/LICENSE
```diff
@@ -261,17 +263,18 @@ POSSIBILITY OF SUCH DAMAGE.
 The CDDL Version 1.1 (https://javaee.github.io/glassfish/LICENSE)
 ---------------------------------------------------------------------------
 
-Apache Geode bundles the following file under the Common Development and
+Apache Geode bundles the following files under the Common Development and
 Distribution License:
 
-  - jaxb-istack-commons v2.2 (https://javaee.github.io/jaxb-istack-commons/)
+  - activation v1.1.0
```
metatype, line 269: Something seems wrong that we are bundling both activation and Java.activation.  Can you file a JIRA for followup?

gradle/java.gradle
```diff
@@ -73,9 +73,9 @@ gradle.taskGraph.whenReady({ graph ->
...
       if (jar.source.filter({ it.name.contains('NOTICE') }).empty) {
-        from("$rootDir/NOTICE")
+        from("$rootDir/geode-assembly/src/main/dist/NOTICE")
```
metatype, line 78: I'm not 100% sure on this one.  If there are no bundled libs (e.g. for geode jars) using the source LICENSE/NOTICE files seems better.  For war files using the full contents of LICENSE/NOTICE from geode-assembly seems better.  But I could be convinced otherwise.

LICENSE
onichols-pivotal: if it's in our source I should be able to find it in our source...I could not find any trace...

buildSrc/src/main/groovy/org/apache/geode/gradle/plugins/DependencyConstraints.groovy
onichols-pivotal: this addition is because we get javax.mail from two different places, due to a change in a Spring 5.3 dependency.  Without this, it brings in 1.6.1 and it looked funny to have both 1.6.1 and 1.6.2 in our license.

geode-assembly/src/main/dist/LICENSE
onichols-pivotal: filed GEODE-8066 for this

gradle/java.gradle
onichols-pivotal: I think gradle treats .war and .jar files as equivalent for applying this clause.  Maybe @rhoughton-pivot can offer some more insight.  I confess my understand of what NOTICE goes where is not as clear as my understanding of LICENSE...

onichols-pivotal: I have reverted this change.  Sounds like NOTICE cleanup might be best for a separate PR.

LICENSE
onichols-pivotal: The only reference to polyfill, which I tracked down from the email thread you shared, is in https://github.com/apache/geode-site/blob/master/website/content/js/respond.min.js

Since we don't distribute geode-site in source form as part of Geode releases, do we still need it in the Geode LICENSE?

buildSrc/src/main/groovy/org/apache/geode/gradle/plugins/DependencyConstraints.groovy
onichols-pivotal: this is intertwined with the activation issue, so GEODE-8066 will cover both

LICENSE
onichols-pivotal: It looks like maybe geode-site used to be a module of geode back when polyfill was originally added to LICENSE.  Now geode-site is a separate repo and isn't shipped as part of Geode releases.  Coincidentally all the other javascript libraries other than polyfill are also used in geode-pulse, which we do ship.

metatype: Nice work!  Thanks for filling in the gaps in my memory :-)

gradle/java.gradle
metatype: Makes sense

PR #5027, opened 2020-04-30
Revert "GEODE-8020: buffer corruption in SSL communications (#4994)"

bschuchardt:
This reverts commit ec8db54ad7f342542762beb8f3e912dff44e3a53.

The fix for GEODE-8020 reduced performance in SSL benchmark tests.
Unfortunately this revert leaves SSL communications broken, but there
are no CI tests that show that this is the case.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/distributedTest/java/org/apache/geode/ClusterCommunicationsDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/ClusterCommunicationsDUnitTest.java
@@ -43,8 +43,6 @@ import java.io.DataOutput;
 import java.io.File;
 import java.io.IOException;
 import java.io.Serializable;
-import java.lang.reflect.Field;
-import java.nio.Buffer;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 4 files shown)


pivotal-jbarrett:
By reverting this we are re-introducing the issue that GEODE-8020 was trying to solve, correct?

bschuchardt:
@pivotal-jbarrett yes, that's true.  However, there is a system property that folks can use to disable use of pooled direct-buffers in comms: p2p.nodirectBuffers=true

pivotal-jbarrett:
> @pivotal-jbarrett yes, that's true. However, there is a system property that folks can use to disable use of pooled direct-buffers in comms: p2p.nodirectBuffers=true, and that will avoid the problem.

Does disabling with the `p2p.nodirectBuffers=true` property result in the same performance issue that this revert corrects?

bschuchardt:
@pivotal-jbarrett yes, the performance issue would be back, **but** the cluster would not hang.  That is currently the case with 1.10, 1.11 and 1.12.  They will hang.  It seems to only occur during HA events.  The performance drop is relative to builds containing the change that caused this hang.  There are no CI tests demonstrating the hang, so we released several times with this bug.  I can't leave in the functional fix w/o failing benchmark tests.  Taking out the functional fix lets the pipeline pass while I work on a new fix that doesn't degrade performance.

pivotal-jbarrett:
> @pivotal-jbarrett yes, the performance issue would be back, **but** the cluster would not hang. That is currently the case with 1.10, 1.11 and 1.12. They will hang. It seems to only occur during HA events. The performance drop is relative to builds containing the change that caused this hang. There are no CI tests demonstrating the hang, so we released several times with this bug. I can't leave in the functional fix w/o failing benchmark tests. Taking out the functional fix lets the pipeline pass while I work on a new fix that doesn't degrade performance.

@bschuchardt Sounds reasonable to me. We should have a discussion on the dev list to decide what we want to do with the drop in performance. What percentage of drop are we experiencing over SSL?

PR #5033, opened 2020-04-30
DRAFT: GEODE-8040: Make BootstrappingFunction unit testable

kirklund:


kirklund:
@jdeppe-pivotal do you have any ideas why any changes to BootstrappingFunction causes the Tomcat session tests in UpgradeTest to fail?

PR #5035, opened 2020-05-01
GEODE-8059: replace multiple redis set functions with one CommandFunction

dschneider-pivotal:

### For all changes:

### Note:


geode-redis/src/main/java/org/apache/geode/redis/internal/executor/set/GeodeRedisSetWithFunctions.java
```diff
@@ -40,30 +47,18 @@ public GeodeRedisSetWithFunctions(ByteArrayWrapper key,
...
   @Override
   public long sadd(ArrayList<ByteArrayWrapper> membersToAdd) {
-    ResultCollector<ArrayList<ByteArrayWrapper>, List<Long>> results = FunctionService
-        .onRegion(region)
-        .withFilter(Collections.singleton(key))
-        .setArguments(membersToAdd)
-        .execute(SaddFunction.ID);
-
+    ResultCollector<Object[], List<Long>> results = executeFunction(SADD, membersToAdd);
     return results.getResult().get(0);
```
prettyClouds, line 56: this line is duplicated in each function. I don't think the `resultcollector` abstraction really makes sense in our case, because we are really trying to just call this on one server and get back one result.  Can we short circuit this by just calling this  `results.getResult().get(0)` in `executeFunction()`

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/set/GeodeRedisSetWithFunctions.java
```diff
@@ -40,30 +47,18 @@ public GeodeRedisSetWithFunctions(ByteArrayWrapper key,
...
   @Override
   public long srem(ArrayList<ByteArrayWrapper> membersToRemove, AtomicBoolean setWasDeleted) {
-    ResultCollector<ArrayList<ByteArrayWrapper>, List<Long>> results = FunctionService
-        .onRegion(region)
-        .withFilter(Collections.singleton(key))
-        .setArguments(membersToRemove)
-        .execute(SremFunction.ID);
+    ResultCollector<Object[], List<Long>> results = executeFunction(SREM, membersToRemove);
     List<Long> resultList = results.getResult();
     long membersRemoved = resultList.get(0);
```
prettyClouds, line 63: Could we just send back an object like SremResult?  this code is dependent on ordering of calls to "sendResult" when it doesn't need to be.

dschneider-pivotal: I think having srem return two things will not be needed for long. Once we have a single region (instead of the metaType region and the dataStore region) then we no longer need to tell the caller that the srem did a delete. So instead of making a new serialize class (like SremResult) we could just have a single result that is an long[] in this case. Then we only need to do resultSender.get(0) like the other operations.

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/set/GeodeRedisSetWithFunctions.java
dschneider-pivotal: srem is different but I think we could make it the same. And then you have the generic challenge of getting executeFunction to return a generic value that the caller can specify. This seems doable and I like having as much of the common code in executeFunction as possible.

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/set/DeltaSet.java
```diff
@@ -30,10 +30,37 @@
...
 import org.apache.geode.cache.Region;
+import org.apache.geode.cache.execute.ResultSender;
 import org.apache.geode.redis.internal.ByteArrayWrapper;
 
 public class DeltaSet implements Delta, DataSerializable {
 
+  public static void sadd(ResultSender<Long> resultSender,
+      Region<ByteArrayWrapper, DeltaSet> localRegion, ByteArrayWrapper key,
+      ArrayList<ByteArrayWrapper> membersToAdd) {
+    resultSender.lastResult(sadd(localRegion, key, membersToAdd));
```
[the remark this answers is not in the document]
dschneider-pivotal: ResultSender is just an interface so someone else could implement it. But I think at this point we don't want anyone except the Function interacting with DeltaSet. We have some work to do (the other set ops that directly read the region) but once they all use the function I'm okay with the only public methods on DeltaSet being ones that encourage the methods to only be used by Functions. But I would also be okay with DeltaSet not even having these ResultSender methods and instead the DeltaSet methods return a value and then we have some code in the Function that adds it to the ResultSender. At some point I'd like the static methods on DeltaSet to go away.

PR #5039, opened 2020-05-01
GEODE-8060: Fix flakiness in GemFireCacheImplCloseTest

kirklund:
I hate creating a special package-private method just for the test BUT I can't think of an effective test without getting some sort of state from GemFireCacheImpl that identifies which call or which thread actually performed the close.

Without identifying which call or thread performed the close, the test devolves to just asserting that two threads can call close but it can't really assert which call did the actual work of closing GemFireCacheImpl. From a really high level, we don't really care. But at a low level, we don't want the 2nd callers to return before the 1st caller has finished the close (doesn't matter if the 1st caller returns from close() after 2nd callers though which is what causes the flakiness).

### DIFF
--- a/geode-core/src/test/java/org/apache/geode/internal/cache/GemFireCacheImplCloseTest.java
+++ b/geode-core/src/test/java/org/apache/geode/internal/cache/GemFireCacheImplCloseTest.java
@@ -28,7 +28,6 @@ import static org.mockito.Mockito.when;
 import java.util.Properties;
 import java.util.concurrent.CyclicBarrier;
 import java.util.concurrent.Future;
-import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Supplier;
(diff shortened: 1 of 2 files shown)


geode-core/src/test/java/org/apache/geode/internal/cache/GemFireCacheImplCloseTest.java
```diff
@@ -190,41 +188,33 @@ public void close_doesNothingIfAlreadyClosed() {
...
 
-    long threadId1 = close1.get();
-    long threadId2 = close2.get();
+    boolean closedCache1 = close1.get();
+    boolean closedCache2 = close2.get();
 
-    assertThat(winner.get())
-        .as("ThreadId1=" + threadId1 + " and threadId2=" + threadId2)
-        .isEqualTo(threadId1);
+    assertThat(closedCache1)
```
jchen21, line 215: Just for safe, can we also assert `closedCache2` is `false`?

kirklund: Good idea!

kirklund:
IntegrationTest failure is caused by this PR. Do NOT merge this in until I fix it.

PR #5040, opened 2020-05-02
GEODE-7414_2: modify init() method argument

mivanac:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/net/SSLParameterExtension.java
+++ b/geode-core/src/main/java/org/apache/geode/net/SSLParameterExtension.java
@@ -15,9 +15,9 @@
 
 package org.apache.geode.net;
 
-import javax.net.ssl.SSLParameters;
+import java.util.Properties;
 
-import org.apache.geode.distributed.internal.DistributionConfig;
+import javax.net.ssl.SSLParameters;
 
 /**
  * User implementation of a SSLParameter extension logic.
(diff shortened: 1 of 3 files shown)


geode-core/src/main/java/org/apache/geode/net/SSLParameterExtension.java
```diff
@@ -29,9 +29,9 @@
   /**
    * Initialize the SSLParameterExtension.
    *
-   * @param config the DistributionConfig
+   * @param properties the Properties
    */
-  default void init(DistributionConfig config) {}
+  default void init(Properties properties) {}
```
onichols-pivotal, line 34: this is a breaking change on a public class that is not marked @Experimental.

why not keep `init(DistributionConfig)` and just make it call `init(distributionConfig.toProperties())` ?

pivotal-jbarrett: The API break is necessary because the original was not a valid API since it leaked internal types. The case of two wrongs make a right.

rhoughton-pivot:
This was reverted by GEODE-8087 due to breaking API compatibility between v1.12 and forward. I'm happy to pair on a new fix, and on writing the exemption to the API check job.

mivanac:
Hi, could you help me how to bypass this API check, since origin class is not marked @Experimental.

PR #5041, opened 2020-05-04
GEODE-8060: Fix flakiness in GemFireCacheImplCloseTest

kirklund:
I found and fixed the issues causing memcached IntegrationJUnitTest to fail. Please re-review and hopefully it should pass all precheckin jobs this time.

I hate creating a special package-private method just for the test BUT I can't think of an effective test without getting some sort of state from GemFireCacheImpl that identifies which call or which thread actually performed the close.

Without identifying which call or thread performed the close, the test devolves to just asserting that two threads can call close but it can't really assert which call did the actual work of closing GemFireCacheImpl. From a really high level, we don't really care. But at a low level, we don't want the 2nd callers to return before the 1st caller has finished the close (doesn't matter if the 1st caller returns from close() after 2nd callers though which is what causes the flakiness).

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/GemFireCacheImpl.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/GemFireCacheImpl.java
@@ -2391,10 +2391,15 @@ public class GemFireCacheImpl implements InternalCache, InternalClientCache, Has
   private boolean waitIfClosing(boolean skipAwait) {
     if (isClosing) {
       if (!skipAwait && !Thread.currentThread().equals(CLOSING_THREAD.get())) {
+        boolean interrupted = false;
         try {
           isClosedLatch.await();
-        } catch (InterruptedException ignore) {
-          // ignored
+        } catch (InterruptedException e) {
+          interrupted = true;
+        } finally {
+          if (interrupted) {
+            Thread.currentThread().interrupt();
+          }
         }
       }
       return true;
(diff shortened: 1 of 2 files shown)


kirklund:
@DonalEvans The problem in the previous PR were caused by me combining `if (isClosing` and `if (!skipAwait etc` into one if-block.

Broken:
```
  private boolean waitIfClosing(boolean skipAwait) {
    if (isClosing && !skipAwait && !Thread.currentThread().equals(CLOSING_THREAD.get())) {
      try {
        isClosedLatch.await();
      } catch (InterruptedException ignore) {
        // ignored
      }
    return true;
    }
    return false;
  }
```
Fixed:
```
  private boolean waitIfClosing(boolean skipAwait) {
    if (isClosing) {
      if (!skipAwait && !Thread.currentThread().equals(CLOSING_THREAD.get())) {
        try {
          isClosedLatch.await();
        } catch (InterruptedException ignore) {
          // ignored
        }
      }
      return true;
    }
    return false;
  }
```

geode-core/src/main/java/org/apache/geode/internal/cache/GemFireCacheImpl.java
```diff
@@ -2377,15 +2381,25 @@ public void close(String reason, Throwable systemFailureCause, boolean keepAlive
...
+  /**
+   * Returns true if caller waited on the {@code isClosedLatch}.
+   */
+  private boolean waitIfClosing(boolean skipAwait) {
+    if (isClosing) {
+      if (!skipAwait && !Thread.currentThread().equals(CLOSING_THREAD.get())) {
+        try {
+          isClosedLatch.await();
+        } catch (InterruptedException ignore) {
+          // ignored
```
kirklund, line 2397: The interrupt flag should be reset after catching `InterruptedException`

PR #5044, opened 2020-05-04
roll develop to 1.14 now that support/1.13 has been created

davebarnes97:


ci/pipelines/shared/jinja.variables.yml
```diff
@@ -66,7 +66,7 @@ java_test_versions:
   version: 11
 
 metadata:
-  initial_version: 1.13.0-((semver-prerelease-token))
+  initial_version: 1.14.0-SNAPSHOT
```
onichols-pivotal, line 69: I think this should have kept the `-((semver-prerelease-token))` suffix

PR #5048, opened 2020-05-04
GEODE-8020: buffer management problems

bschuchardt:
This fixes some buffer handling in MsgStreamerList and alters
MstStreamer to avoid creating MsgStreamerList and VersionedMsgStreamers
during normal, non-upgrade, operations.

It also changes NioSslEngine to use synchronization in more places,
notably the close() method, which was possibly allowing multiple threads to
change the state of the engine.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/net/BufferPool.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/net/BufferPool.java
@@ -19,13 +19,18 @@ import java.nio.ByteBuffer;
 import java.util.IdentityHashMap;
 import java.util.concurrent.ConcurrentLinkedQueue;
 
+import org.apache.logging.log4j.Logger;
+
 import org.apache.geode.distributed.internal.DMStats;
 import org.apache.geode.distributed.internal.DistributionConfig;
 import org.apache.geode.internal.Assert;
 import org.apache.geode.internal.tcp.Connection;
+import org.apache.geode.logging.internal.log4j.api.LogService;
+import org.apache.geode.util.internal.GeodeGlossary;
 
 public class BufferPool {
   private final DMStats stats;
+  private static final Logger logger = LogService.getLogger();
 
   /**
    * Buffers may be acquired from the Buffers pool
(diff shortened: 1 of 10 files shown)


geode-core/src/main/java/org/apache/geode/internal/net/NioSslEngine.java
```diff
@@ -398,6 +398,7 @@ public void close(SocketChannel socketChannel) {
     } finally {
       bufferPool.releaseBuffer(TRACKED_SENDER, myNetData);
       bufferPool.releaseBuffer(TRACKED_RECEIVER, peerAppData);
+      myNetData = null;
```
echobravopapa, line 401: that's cute, personalized data ;P

geode-core/src/main/java/org/apache/geode/internal/tcp/MsgStreamerList.java
```diff
@@ -63,25 +63,16 @@ public int writeMessage() throws IOException {
...
-      }
-      try {
-        result += streamer.writeMessage();
-        // if there is an exception we need to finish the
-        // loop and release the other streamer's buffers
-      } catch (RuntimeException e) {
-        ex = e;
-      } catch (IOException e) {
-        ioex = e;
+      } else {
```
echobravopapa, line 66: thx for removing all those crufty comments

geode-core/src/main/java/org/apache/geode/internal/net/BufferPool.java
```diff
@@ -69,7 +74,8 @@ public BufferPool(DMStats stats) {
   /**
    * use direct ByteBuffers instead of heap ByteBuffers for NIO operations
    */
-  public static final boolean useDirectBuffers = !Boolean.getBoolean("p2p.nodirectBuffers");
+  public static final boolean useDirectBuffers = !(Boolean.getBoolean("p2p.nodirectBuffers")
+      || Boolean.getBoolean(GeodeGlossary.GEMFIRE_PREFIX + "noDirectBuffers"));
```
echobravopapa, line 78: is the GF prefixed version of the property need to be exposed/documented?

bschuchardt: > is the GF prefixed version of the property need to be exposed/documented?

yes, but I'm not sure what the best way is to do that.  I updated properties.html in this PR with the new property.

pivotal-jbarrett:
Can we make the new geode property positive by dropping the no and flipping the default.

bschuchardt:
> Can we make the new geode property positive by dropping the no and flipping the default.

Good point, and the new property hints that nothing will use direct-buffers.  How about this?
GeodeGlossary.GEMFIRE_PREFIX + "BufferPool.useHeapBuffers"

PR #5054, opened 2020-05-05
GEODE-8071: Use daemon threads in RebalanceCommand

jujoramos:
Changed the ExecutorService within RebalanceCommand to use daemon 
threads, otherwise the locator refuses to gracefully shutdown.

- Fixed minor warnings.
- Added distributed tests.

### For all changes:

### Note:


jujoramos:
The [first commit ](https://github.com/apache/geode/pull/5054/commits/feb052e71cee12a5e7a2ca72809f9ad07a908e44) is just a refactor of the test class, the actual changes to fix the problem are in the [second commit](https://github.com/apache/geode/pull/5054/commits/bd481554d8add576c885c675c2350474eb398939). I've split them to make the review easier, will `squash` them before merging into `develop`.

geode-dunit/src/main/java/org/apache/geode/management/internal/cli/commands/RebalanceCommandDistributedTest.java
```diff
@@ -35,40 +37,91 @@
...
+      RegionFactory<String, String> dataRegionFactory =
+          cache.createRegionFactory(RegionShortcut.PARTITION);
+      Region<String, String> region = dataRegionFactory.create(REGION_ONE_NAME);
+      for (int i = 0; i < 10; i++) {
+        region.put("key" + (i + 200), "value" + (i + 200));
+      }
+      region = dataRegionFactory.create(REGION_TWO_NAME);
+      for (int i = 0; i < 100; i++) {
+        region.put("key" + (i + 200), "value" + (i + 200));
+      }
```
DonalEvans, line 78: Could the ints added to the keys and values be cleared up a bit? There doesn't seem to be any reason to add 200 to them here, and if the number of entries is extracted to a constant, then it can be used in the invocation on server2 to prevent overwriting the existing data.

jujoramos: I actually didn't add these tests and the entries looked a bit weird to me as well, I didn't want to change this but it makes sense, will do it, thanks.

jujoramos:
@DonalEvans 
> One thing I'm wondering though... do we know of any circumstances in which a locator might launch a non-deamon thread other than this? I'm concerned that the test might fail for unrelated reasons if something other than the rebalance command happens to start a thread.

Under the current test conditions no other `non-daemon` threads should be launched by the locator; it's a valid concern, though, so thanks for bringing it up.

PR #5055, opened 2020-05-05
GEODE-8073: Fix NPE after FetchKeysMessage failed.

pivotal-eshu:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
@@ -4593,39 +4593,58 @@ public class PartitionedRegion extends LocalRegion
         buckets = bucketKeys.keySet();
       }
 
-      for (Integer bucket : buckets) {
-        Set keys = null;
-        if (bucketKeys == null) {
-          try {
-            FetchKeysResponse fkr = FetchKeysMessage.send(member, this, bucket, true);
-            keys = fkr.waitForKeys();
-          } catch (ForceReattemptException ignore) {
-            failures.add(bucket);
-          }
-        } else {
-          keys = bucketKeys.get(bucket);
-        }
-
-        // TODO (ashetkar) Use single Get70 instance for all?
-        for (Object key : keys) {
-          Get70 command = (Get70) Get70.getCommand();
-          Get70.Entry ge = command.getValueAndIsObject(this, key, null, servConn);
-
-          if (ge.keyNotPresent) {
-            values.addObjectPartForAbsentKey(key, ge.value, ge.versionTag);
-          } else {
-            values.addObjectPart(key, ge.value, ge.isObject, ge.versionTag);
-          }
-
-          if (values.size() == BaseCommand.MAXIMUM_CHUNK_SIZE) {
-            BaseCommand.sendNewRegisterInterestResponseChunk(this, "keyList", values, false,
-                servConn);
-            values.clear();
-          }
-        }
-      }
+      fetchKeysAndValues(values, servConn, failures, member, bucketKeys, buckets);
     }
     return failures;
+  }
+
+  void fetchKeysAndValues(VersionedObjectList values, ServerConnection servConn,
+      Set<Integer> failures, InternalDistributedMember member,
+      HashMap<Integer, HashSet> bucketKeys, Set<Integer> buckets)
+      throws IOException {
+    for (Integer bucket : buckets) {
+      Set keys = null;
+      if (bucketKeys == null) {
+        try {
+          FetchKeysResponse fetchKeysResponse = getFetchKeysResponse(member, bucket);
+          keys = fetchKeysResponse.waitForKeys();
+        } catch (ForceReattemptException ignore) {
+          failures.add(bucket);
+        }
+      } else {
+        keys = bucketKeys.get(bucket);
+      }
+      if (keys != null) {
+        getValuesForKeys(values, servConn, keys);
+      }
+    }
+  }
+
+  FetchKeysResponse getFetchKeysResponse(InternalDistributedMember member,
+      Integer bucket)
+      throws ForceReattemptException {
+    return FetchKeysMessage.send(member, this, bucket, true);
+  }
+
+  void getValuesForKeys(VersionedObjectList values, ServerConnection servConn, Set keys)
+      throws IOException {
+    // TODO (ashetkar) Use single Get70 instance for all?
+    for (Object key : keys) {
+      Get70 command = (Get70) Get70.getCommand();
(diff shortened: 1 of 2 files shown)


geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
```diff
@@ -4593,39 +4593,58 @@ void updateNodeToBucketMap(
...
+      Set<Integer> failures, InternalDistributedMember member,
+      HashMap<Integer, HashSet> bucketKeys, Set<Integer> buckets)
+      throws IOException {
+    for (Integer bucket : buckets) {
+      Set keys = null;
+      if (bucketKeys == null) {
+        try {
+          FetchKeysResponse fetchKeysResponse = getFetchKeysResponse(member, bucket);
+          keys = fetchKeysResponse.waitForKeys();
+        } catch (ForceReattemptException ignore) {
```
DonalEvans, line 4611: The exception isn't ignored, so it shouldn't be called "ignore."

geode-core/src/test/java/org/apache/geode/internal/cache/PartitionedRegionTest.java
```diff
@@ -573,6 +578,50 @@ public void transactionThrowsTransactionDataRebalancedExceptionIfIsAForceReattem
...
+    VersionedObjectList values = mock(VersionedObjectList.class);
+    ServerConnection serverConnection = mock(ServerConnection.class);
+    Set<Integer> failures = mock(Set.class);
+    InternalDistributedMember member = mock(InternalDistributedMember.class);
+    Set<Integer> buckets = new HashSet<>();
+    buckets.add(1);
+    FetchKeysMessage.FetchKeysResponse fetchKeysResponse =
+        mock(FetchKeysMessage.FetchKeysResponse.class);
+    doReturn(fetchKeysResponse).when(spyPartitionedRegion).getFetchKeysResponse(member, 1);
+    Set keys = mock(Set.class);
```
DonalEvans, line 614: Here you should use a real Set instead of a mock.

geode-core/src/test/java/org/apache/geode/internal/cache/PartitionedRegionTest.java
```diff
@@ -573,6 +578,50 @@ public void transactionThrowsTransactionDataRebalancedExceptionIfIsAForceReattem
         .hasMessage(PartitionedRegion.DATA_MOVED_BY_REBALANCE).hasCause(exception);
   }
 
+  @Test
+  public void failuresSavedIfFetchKeysThrows() throws Exception {
+    PartitionedRegion spyPartitionedRegion = spy(partitionedRegion);
+
+    VersionedObjectList values = mock(VersionedObjectList.class);
+    ServerConnection serverConnection = mock(ServerConnection.class);
+    Set<Integer> failures = mock(Set.class);
```
DonalEvans, line 587: It's best not to mock Collections. It's fine to just use an actual Set here.

geode-core/src/test/java/org/apache/geode/internal/cache/PartitionedRegionTest.java
```diff
@@ -573,6 +578,50 @@ public void transactionThrowsTransactionDataRebalancedExceptionIfIsAForceReattem
...
+    verify(spyPartitionedRegion, never()).getValuesForKeys(values, serverConnection, null);
+  }
+
+  @Test
+  public void fetchKeysAndValuesInvokesGetValuesForKeys() throws Exception {
+    PartitionedRegion spyPartitionedRegion = spy(partitionedRegion);
+
+    VersionedObjectList values = mock(VersionedObjectList.class);
+    ServerConnection serverConnection = mock(ServerConnection.class);
+    Set<Integer> failures = mock(Set.class);
```
DonalEvans, line 607: Here you should use a real Set instead of a mock.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
pivotal-eshu: The exception was caught here, but the exception is itself being ignored. Seems that this was particularly changed to ignore in a previous checkin. Does this needs to change back definitely?

geode-core/src/test/java/org/apache/geode/internal/cache/PartitionedRegionTest.java
pivotal-eshu: The mocked failures as a set is used later to verify whether a method is invoked on the set. Otherwise it would be hard to verify.

geode-core/src/test/java/org/apache/geode/internal/cache/PartitionedRegionTest.java
pivotal-eshu: Is that a particular reason set should not be mocked? It is not being iterated through, just used as a placeholder. I would like to know the reason for the requested change so that I may gain some new knowledge.

geode-core/src/test/java/org/apache/geode/internal/cache/PartitionedRegionTest.java
pivotal-eshu: I'd like to know the reason why this has to be a real Set.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
DonalEvans: Ah, I see. We may catch the exception, but we don't actually do anything with it, even though we do _something_ if we catch it. This name is fine as it is.

geode-core/src/test/java/org/apache/geode/internal/cache/PartitionedRegionTest.java
DonalEvans: Instead of verifying the method call on the mocked set, you can verify the contents of the real set, which should contain one item, which is the `Integer` 1.

geode-core/src/test/java/org/apache/geode/internal/cache/PartitionedRegionTest.java
DonalEvans: It's part of the "don't mock what you don't own" principle in TDD, explained [here](https://github.com/testdouble/contributing-tests/wiki/Don't-mock-what-you-don't-own) and [here](https://blog.codecentric.de/en/2018/03/mock-what-when-how/). Also, as discussed [here](https://github.com/mockito/mockito/wiki/How-to-write-good-tests), if you can avoid mocking a class, as in this case, then that's probably a good thing.

pivotal-eshu: Thanks for the explanation. Change has been made.

PR #5064, opened 2020-05-06
Geode-8082: Refactor (reformat and renames) in GeodeRedisServer file in Geode-Redis-Module

jhutchison:

### For all changes:

### Note:

### DIFF
--- a/geode-redis/src/main/java/org/apache/geode/redis/GeodeRedisServer.java
+++ b/geode-redis/src/main/java/org/apache/geode/redis/GeodeRedisServer.java
@@ -596,44 +596,63 @@ public class GeodeRedisServer {
 
     bossGroup = null;
     workerGroup = null;
+
     Class<? extends ServerChannel> socketClass;
     if (singleThreadPerConnection) {
       socketClass =
-          startRedisServiceSingleThreadPerConnection(selectorThreadFactory, workerThreadFactory);
+          startRedisServiceSingleThreadPerConnection(
+              selectorThreadFactory,
+              workerThreadFactory);
     } else {
-      bossGroup = new NioEventLoopGroup(numSelectorThreads, selectorThreadFactory);
-      workerGroup = new NioEventLoopGroup(numWorkerThreads, workerThreadFactory);
+      bossGroup =
+          new NioEventLoopGroup(numSelectorThreads, selectorThreadFactory);
+      workerGroup =
+          new NioEventLoopGroup(numWorkerThreads, workerThreadFactory);
+
       socketClass = NioServerSocketChannel.class;
     }
-    InternalDistributedSystem system = (InternalDistributedSystem) cache.getDistributedSystem();
-    String pwd = system.getConfig().getRedisPassword();
-    final byte[] pwdB = Coder.stringToBytes(pwd);
-    ServerBootstrap b = new ServerBootstrap();
 
-    b.group(bossGroup, workerGroup).channel(socketClass)
-        .childHandler(new ChannelInitializer<SocketChannel>() {
-          @Override
-          public void initChannel(SocketChannel ch) {
-            if (logger.fineEnabled()) {
-              logger.fine("GeodeRedisServer-Connection established with " + ch.remoteAddress());
-            }
-            ChannelPipeline p = ch.pipeline();
-            addSSLIfEnabled(ch, p);
-            p.addLast(ByteToCommandDecoder.class.getSimpleName(), new ByteToCommandDecoder());
-            p.addLast(ExecutionHandlerContext.class.getSimpleName(),
-                new ExecutionHandlerContext(ch, cache, regionCache, GeodeRedisServer.this, pwdB,
-                    keyRegistrar, pubSub, hashLockService));
-          }
-        }).option(ChannelOption.SO_REUSEADDR, true).option(ChannelOption.SO_RCVBUF, getBufferSize())
+    InternalDistributedSystem system =
+        (InternalDistributedSystem) cache.getDistributedSystem();
+
+    String pwd =
+        system.getConfig().getRedisPassword();
+
+    final byte[] passwordAsByteArray =
+        Coder.stringToBytes(pwd);
+
+    ServerBootstrap serverBootstrap = new ServerBootstrap();
+
+    serverBootstrap
+        .group(bossGroup, workerGroup)
+        .channel(socketClass)
+        .childHandler(getChildHandler(passwordAsByteArray))
+        .option(ChannelOption.SO_REUSEADDR, true)
+        .option(ChannelOption.SO_RCVBUF, getBufferSize())
         .childOption(ChannelOption.SO_KEEPALIVE, true)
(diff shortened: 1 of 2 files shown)


geode-redis/src/main/java/org/apache/geode/redis/internal/ExecutionHandlerContext.java
```diff
@@ -56,7 +56,7 @@
   private static final Logger logger = LogService.getLogger();
   private static final int WAIT_REGION_DSTRYD_MILLIS = 100;
   private static final int MAXIMUM_NUM_RETRIES = (1000 * 60) / WAIT_REGION_DSTRYD_MILLIS; // 60
-                                                                                          // seconds
+  // seconds
```
dschneider-pivotal, line 59: This one still does not look correct. The comment is not even correct (this is a max retry count not a time. The 1000*60 represents 60 seconds (in millis). You could extract that into another final int and name it 60_SECONDS_IN_MILLIS. The JDK also has some classes that can help with these time unit conversions

jhutchison: I'll have to look at why that changed- I definitely didn't change any comments intentionally

jhutchison:
> What are the renames? I could not spot any of them with all the reformatting.
> 
> I would not reject new code formatted like this but I'd prefer that we don't go around to all our existing code and reformat it like this. I understand this is a stylistic issue.
> With so much whitespace and short lines it leads to not being able to have very many statements on a screen for one viewing. I don't understand the rational. Like it looks like variables can no longer be assigned on the same line. For example "bossGroup =" is not longer initialized on the same line with the constructor call (new NioEvent...). But it is okay to do this "bossGroup = null" on one line. If this kind of reformatting is valuable then it seems like we should have it in our coding standard. Without it being enforced I'm not sure what will cause it to persist.
> 
> One of the problems with reformatting existing code is can can cause spurious conflicts with others working on the code. I saw this just yesterday on the synchronized-deltas branch.


For sure stylistic-  I think I'm in the minority here, so I'm ok rejecting it if folks don't like it. For me it's much easier to look quickly and understand that an assignment is happening  when it's broken up like that.  In my mind, when an assignment is super short like x = 1, then it's really easy to take in the totality of what is happening on the line. when it's x = p.fill("foo").get(3).whatever(list<<object>>.class)- or something along those lines- it takes more time to process and even see quickly where the assignment is happening.  In a similar way I like to break chained calls up on lines because It find it much easier to quickly process the individual actions in a chain of events that way. But again, this is me.  I'm fine not merging the commit if folks are not of the same mind as me.   
The name changes are almost entirely thing like changing b into bootStrapServer and ch into Channel-  that kind of thing.

geode-redis/src/main/java/org/apache/geode/redis/GeodeRedisServer.java
```diff
@@ -596,61 +596,121 @@ public Thread newThread(Runnable r) {
...
+              selectorThreadFactory,
+              workerThreadFactory);
     } else {
-      bossGroup = new NioEventLoopGroup(numSelectorThreads, selectorThreadFactory);
-      workerGroup = new NioEventLoopGroup(numWorkerThreads, workerThreadFactory);
+      bossGroup =
+          new NioEventLoopGroup(numSelectorThreads, selectorThreadFactory);
+      workerGroup =
+          new NioEventLoopGroup(numWorkerThreads, workerThreadFactory);
+
```
jdeppe-pivotal, line 611: I realize this is perhaps a very personal preference, but I really dislike breaking up assignments (or method calls) like this when the line itself isn't really very long and doesn't already hit the max line length for code style. I find it makes methods longer such that I end up having to scroll back and forth to read the method.

geode-redis/src/main/java/org/apache/geode/redis/GeodeRedisServer.java
```diff
@@ -596,61 +596,121 @@ public Thread newThread(Runnable r) {
...
-          }
-        }).option(ChannelOption.SO_REUSEADDR, true).option(ChannelOption.SO_RCVBUF, getBufferSize())
+    InternalDistributedSystem system =
+        (InternalDistributedSystem) cache.getDistributedSystem();
+
+    String pwd =
+        system.getConfig().getRedisPassword();
+
+    final byte[] passwordAsByteArray =
+        Coder.stringToBytes(pwd);
```
jdeppe-pivotal, line 622: Same comment as above

geode-redis/src/main/java/org/apache/geode/redis/GeodeRedisServer.java
```diff
@@ -596,61 +596,121 @@ public Thread newThread(Runnable r) {
...
+        .option(ChannelOption.SO_RCVBUF, getBufferSize())
         .childOption(ChannelOption.SO_KEEPALIVE, true)
-        .childOption(ChannelOption.CONNECT_TIMEOUT_MILLIS, GeodeRedisServer.connectTimeoutMillis)
-        .childOption(ChannelOption.ALLOCATOR, PooledByteBufAllocator.DEFAULT);
+        .childOption(
+            ChannelOption.CONNECT_TIMEOUT_MILLIS,
+            GeodeRedisServer.connectTimeoutMillis)
+        .childOption(
+            ChannelOption.ALLOCATOR,
+            PooledByteBufAllocator.DEFAULT);
```
jdeppe-pivotal, line 638: I like this change! Definitely worth line-separating chained method calls like this.

geode-redis/src/main/java/org/apache/geode/redis/GeodeRedisServer.java
```diff
@@ -596,61 +596,121 @@ public Thread newThread(Runnable r) {
...
     if (logger.infoEnabled()) {
-      String logMessage = "GeodeRedisServer started {" + getBindAddress() + ":" + serverPort
-          + "}, Selector threads: " + numSelectorThreads;
+      String logMessage =
+          "GeodeRedisServer started {"
+              + getBindAddress()
+              + ":" + serverPort
+              + "}, Selector threads: "
+              + numSelectorThreads;
+
```
jdeppe-pivotal, line 655: 4 lines have expanded to 15. I find this harder to now quickly parse and see where subsequent new logic starts.

geode-redis/src/main/java/org/apache/geode/redis/GeodeRedisServer.java
```diff
@@ -596,61 +596,121 @@ public Thread newThread(Runnable r) {
...
     bossGroup =
-        new io.netty.channel.oio.OioEventLoopGroup(Integer.MAX_VALUE, selectorThreadFactory);
+        new io.netty.channel.oio.OioEventLoopGroup(
+            Integer.MAX_VALUE, selectorThreadFactory);
+
     workerGroup =
-        new io.netty.channel.oio.OioEventLoopGroup(Integer.MAX_VALUE, workerThreadFactory);
+        new io.netty.channel.oio.OioEventLoopGroup(
+            Integer.MAX_VALUE, workerThreadFactory);
+
```
jdeppe-pivotal, line 713: Can these two fully qualified class names be imported?

PR #5071, opened 2020-05-07
GEODE-8091: fix AuthenticationRequiredException when starting a locat…

jinmeiliao:
…or with security manager and preload cluster configuration.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/integrationTest/java/org/apache/geode/distributed/LoadClusterConfigFromDirIntegrationTest.java
+++ b/geode-core/src/integrationTest/java/org/apache/geode/distributed/LoadClusterConfigFromDirIntegrationTest.java
@@ -1,0 +1,73 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package org.apache.geode.distributed;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.Collection;
+
+import org.apache.commons.io.FileUtils;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+
+import org.apache.geode.distributed.internal.InternalConfigurationPersistenceService;
+import org.apache.geode.examples.SimpleSecurityManager;
+import org.apache.geode.management.configuration.Deployment;
+import org.apache.geode.management.internal.configuration.domain.Configuration;
+import org.apache.geode.test.compiler.JarBuilder;
+import org.apache.geode.test.junit.rules.LocatorStarterRule;
+
+public class LoadClusterConfigFromDirIntegrationTest {
+  @Rule
+  public LocatorStarterRule locator = new LocatorStarterRule();
+
+  private File clusterConfigDir;
+
+  @Before
+  public void before() throws IOException {
+    clusterConfigDir = new File(locator.getWorkingDir(), "cluster_config");
+    File groupDir = new File(clusterConfigDir, "cluster");
+    groupDir.mkdirs();
+    File jarFile = new File(groupDir, "test.jar");
+    JarBuilder jarBuilder = new JarBuilder();
+    jarBuilder.buildJarFromClassNames(jarFile, "TestFunction");
+  }
+
+  @Test
+  public void canStartWithDeployedJarInClusterConfig() {
+    locator.withSecurityManager(SimpleSecurityManager.class)
+        .withProperty("load-cluster-configuration-from-dir", "true")
+        .startLocator();
+
+    InternalConfigurationPersistenceService ccService =
+        locator.getLocator().getConfigurationPersistenceService();
+    Configuration config = ccService.getConfiguration("cluster");
+    Collection<Deployment> deployments = config.getDeployments();
+    assertThat(deployments).hasSize(1);
+    Deployment deployment = deployments.iterator().next();
+    assertThat(deployment.getFileName()).isEqualTo("test.jar");
+    assertThat(deployment.getDeployedBy()).isNull();
+  }
+
+  @After
+  public void after() {
+    FileUtils.deleteQuietly(clusterConfigDir);
+  }
+}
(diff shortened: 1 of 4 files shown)


geode-core/src/main/java/org/apache/geode/distributed/ConfigurationProperties.java
```diff
@@ -1144,7 +1144,10 @@
    * <U>Allowed values</U>: true or false
    * </p>
    * <U>Since</U>: GemFire 8.0
+   *
+   * @deprecated Since Geode 1.6 use "import cluster-configuration" command instead
    */
+  @Deprecated
```
jujoramos, line 1150: Regular users will likely never notice this property is deprecated as they rarely check the source code. 
The documentation is [already updated ](https://github.com/apache/geode/blob/develop/geode-docs/tools_modules/gfsh/command-pages/start.html.md.erb#L393), though, but maybe it's a good idea to log a warning as part of the `start locator` command when the property is actually used?.

jinmeiliao: the gfsh help message  contains the "deprecated" explanation since 1.6. Bug logging a warning seems a good idea, but not the scope of this PR.

geode-core/src/main/java/org/apache/geode/distributed/internal/InternalConfigurationPersistenceService.java
```diff
@@ -841,8 +842,15 @@ private Configuration readConfiguration(File groupConfigDir) throws IOException
...
   }
 
-  private String getDeployedBy() {
-    Subject subject = cache.getSecurityService().getSubject();
+  String getDeployedBy() {
+    Subject subject = null;
+    try {
+      subject = cache.getSecurityService().getSubject();
+    } catch (AuthenticationRequiredException exception) {
+      // ignored. No user logged in for the deployment
```
kirklund, line 850: Might want to log this at debug or trace level.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/StartLocatorCommand.java
```diff
@@ -139,6 +139,7 @@ public ResultModel startLocator(
 
   }
 
+  @SuppressWarnings("deprecation")
```
kirklund, line 142: In general, I would encourage you to not use `@SuppressWarnings("deprecation")`. All that does it hide how much we're using deprecated APIs, whereas I think we need to be aware of how much deprecated code we're calling and eventually change it. If it's hidden we'll forget and not revisit it.

jinmeiliao: our build would fail if we don't suppress it. I believe Jake just put in a check to prevent the compiler to build successfully if we introduced warning. but in this case, we had to use the deprecated variable since "start locator" still needs to support it.

geode-core/src/main/java/org/apache/geode/distributed/internal/InternalConfigurationPersistenceService.java
jinmeiliao: done.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/StartLocatorCommand.java
mhansonp: Hmm. That is actually really bad if true, because many of the API are not properly type and are raw. So in writing new tests, there would be a real problem.

pivotal-jbarrett: Fix the root problem. We need to stop propagating bad habits and warnings. If the method is deprecated, don't use it.

jinmeiliao: > Fix the root problem. We need to stop propagating bad habits and warnings. If the method is deprecated, don't use it.

@pivotal-jbarrett the code needs to refer to the `ConfigurationProperties.LOAD_CLUSTER_CONFIGURATION_FROM_DIR` directly, hence the warning, are you saying we need to change it to a string literal here?

pivotal-jbarrett: If the use of this property is truly deprecated then it should not be deprecated. If there is some legitimate reason for us of a deprecated property or method then isolate it and suppress that single statement. Suppressing this entire method could mask other issues. Suppression should be at the smallest scope possible.

pivotal-jbarrett: Replace the offending line 197 with a method like this:
```java
  @SuppressWarnings("deprecation")
  private void setLoadClusterConfigurationFromDir(Properties gemfireProperties,
                                                  boolean loadSharedConfigurationFromDirectory) {
        ConfigurationProperties.LOAD_CLUSTER_CONFIGURATION_FROM_DIR,
        loadSharedConfigurationFromDirectory);
  }
```

jinmeiliao:
@jujoramos is right, I will create a JIRA to do deprecation properly. Should not have muddled with this in this PR. This PR should focus on fixing the AuthenticationRequiredException.

geode-core/src/main/java/org/apache/geode/distributed/internal/InternalConfigurationPersistenceService.java
```diff
@@ -841,8 +842,16 @@ private Configuration readConfiguration(File groupConfigDir) throws IOException
...
-    Subject subject = cache.getSecurityService().getSubject();
+  String getDeployedBy() {
+    Subject subject = null;
+    try {
+      subject = cache.getSecurityService().getSubject();
+    } catch (AuthenticationRequiredException e) {
+      // ignored. No user logged in for the deployment
+      // this would happen for offline commands like "start locator" and loading the cluster config
+      // from a directory
+      logger.debug("getDeployedBy: no user information is found.", e.getMessage());
```
mhansonp, line 853: Are you missing a {} in the string.

geode-core/src/integrationTest/java/org/apache/geode/distributed/LoadClusterConfigFromDirIntegrationTest.java
```diff
@@ -0,0 +1,73 @@
...
+
+  private File clusterConfigDir;
+
+  @Before
+  public void before() throws IOException {
+    clusterConfigDir = new File(locator.getWorkingDir(), "cluster_config");
+    File groupDir = new File(clusterConfigDir, "cluster");
+    groupDir.mkdirs();
+    File jarFile = new File(groupDir, "test.jar");
+    JarBuilder jarBuilder = new JarBuilder();
```
mhansonp, line 49: Might be good to check for directory existence before we assume the creates were successful, or just issue the delete on the directory pre-emptively.

geode-core/src/main/java/org/apache/geode/distributed/internal/InternalConfigurationPersistenceService.java
```diff
@@ -841,8 +842,16 @@ private Configuration readConfiguration(File groupConfigDir) throws IOException
     return configuration;
   }
 
-  private String getDeployedBy() {
-    Subject subject = cache.getSecurityService().getSubject();
+  String getDeployedBy() {
+    Subject subject = null;
+    try {
+      subject = cache.getSecurityService().getSubject();
+    } catch (AuthenticationRequiredException e) {
```
agingade, line 849: Can the "getDeployedBy()" be called other than "offline command like start locator"; can this be valid exception in those cases that needs to be handled. Sorry, looking at the code I could not make it if its called in offline mode.

jinmeiliao: only for offline commands we can get this kind of exception. For online commands, there is always a subject present.

PR #5076, opened 2020-05-08
Geode 7812: Changes to PutAllClientServerDistributionTest

mhansonp:
Fixed issues causing testEventIdOutOfOrderInPartitionRegionSingleHop to fail

### DIFF
--- a/geode-cq/src/distributedTest/java/org/apache/geode/internal/cache/PutAllClientServerDistributedTest.java
+++ b/geode-cq/src/distributedTest/java/org/apache/geode/internal/cache/PutAllClientServerDistributedTest.java
@@ -59,7 +59,6 @@ import java.util.function.Consumer;
 import junitparams.JUnitParamsRunner;
 import junitparams.Parameters;
 import junitparams.naming.TestCaseName;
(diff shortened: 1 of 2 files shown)


geode-cq/src/distributedTest/java/org/apache/geode/internal/cache/PutAllClientServerDistributedTest.java
```diff
@@ -2305,24 +2300,21 @@ public void testEventIdMisorderInPRSingleHop() {
...
 
     // server2 add slow listener
     server2.invoke(() -> {
       Region<String, TickerData> region = getCache().getRegion(regionName);
       region.getAttributesMutator()
           .addCacheListener(new SlowCountingCacheListener<>(new Action<>(Operation.CREATE,
-              creates -> executorServiceRule.submit(() -> {
-                closeCacheConditionally(creates, 10);
-              }))));
+              creates -> closeCacheConditionally(creates, 30))));
```
agingade: It could so happen the number of buckets on this server could be less than 30; making putAll to success, changing it to less number will help.

PR #5085, opened 2020-05-08
GEODE-8090: implement function/delta for redis Hash ops

dschneider-pivotal:
Only a few of the hash ops use function/delta. But all the others also work with the new data model.
The hash ops that use function delta now are:
HDEL
DEL
HSET
HMSET
HSETNX
HGETALL

### For all changes:

### Note:


geode-redis/src/main/java/org/apache/geode/redis/internal/executor/hash/HashExecutor.java
```diff
@@ -42,14 +41,14 @@
    * @param key the region hash key region:<key>
    * @return the map data
    */
-  protected Map<ByteArrayWrapper, ByteArrayWrapper> getMap(ExecutionHandlerContext context,
+  protected RedisHash getMap(ExecutionHandlerContext context,
```
prettyClouds, line 44: can we deprecate this method?

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/hash/RedisHashCommands.java
```diff
@@ -0,0 +1,32 @@
...
+
+public interface RedisHashCommands {
+  int hset(ByteArrayWrapper key, List<ByteArrayWrapper> fieldsToSet, boolean NX);
+
+  int hdel(ByteArrayWrapper key, List<ByteArrayWrapper> fieldsToRemove);
+
+  // TODO: refactor this onto something like RedisKeyCommands
+  boolean del(ByteArrayWrapper key);
+
+  Collection<ByteArrayWrapper> hgetall(ByteArrayWrapper key);
```
prettyClouds, line 31: i think it would be nice if hgetall returned a collection of enttries.  That has more semantic meaning, but I understand that the caller would end up just flattening it to send it over the wire.

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/CommandFunction.java
```diff
@@ -16,26 +16,53 @@
...
 import org.apache.geode.internal.cache.execute.RegionFunctionContextImpl;
 import org.apache.geode.redis.internal.ByteArrayWrapper;
 import org.apache.geode.redis.internal.RedisCommandType;
+import org.apache.geode.redis.internal.RedisDataType;
+import org.apache.geode.redis.internal.executor.hash.RedisHash;
 import org.apache.geode.redis.internal.executor.set.RedisSet;
 import org.apache.geode.redis.internal.executor.set.StripedExecutor;
+import org.apache.geode.redis.internal.executor.set.SynchronizedStripedExecutor;
 
 public class CommandFunction implements Function<Object[]> {
```
prettyClouds, line 42: this class is going to grow with each new command added.  Is there any way we can break this up?

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/hash/RedisHashCommands.java
dschneider-pivotal: And it would not be serializable

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/CommandFunction.java
dschneider-pivotal: Probably. I don't have any ideas. But it does seem like we have a bunch of things that are related to each other and maybe they can be refactored in some way. I do like having a single function. For example we could have a function for each command but that would cause a bunch of busy work each time we add function support for another redis command. I would suggest we wait on this refactoring until we get all the commands we want to support for our first release done. Then we can look at those and figure out the best way to organize the code for those commands.

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/hash/RedisHashCommands.java
dschneider-pivotal: I do think it makes more sense for hgetall to return a Collection<ByteArrayWrapper> instead of Collection<Entry<ByteArrayWrapper, ByteArrayWrapper>> because the output of the redis HGETALL command is a flat list. The output is "key1 value1 key2 value2". It does not have structure (for example "{key1 value1} {key2 value2}". So by having hgetall return a simple Collection it actually corresponds to the public REDIS HGETALL output.

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/hash/HashExecutor.java
dschneider-pivotal: We could but we still will use it internally until the other HASH commands are converted to use functions. I don't think it is worth marking the internal apis we need to support this older implementation as deprecated. We should just spend a day converting the other HASH commands to use a function and then we can get rid of this, and other, methods.

PR #5095, opened 2020-05-12
GEODE-7680: PR.clear must be successful when interacting with rebalance

DonalEvans:
- Added DUnit tests to confirm that clear does not interfere with
rebalance or vice versa
- Fixed typo in PartitionedRegionClearWithExpirationDUnitTest
- Fixed typo in PartitionedRegion

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithRebalanceDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithRebalanceDUnitTest.java
@@ -22,7 +22,6 @@ import static org.apache.geode.internal.util.ArrayUtils.asList;
 import static org.apache.geode.test.awaitility.GeodeAwaitility.await;
 import static org.apache.geode.test.dunit.VM.getVM;
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 import java.io.Serializable;
 import java.util.Collection;
@@ -42,7 +41,6 @@ import org.junit.runner.RunWith;
 
 import org.apache.geode.cache.CacheWriterException;
 import org.apache.geode.cache.PartitionAttributesFactory;
-import org.apache.geode.cache.PartitionedRegionPartialClearException;
 import org.apache.geode.cache.Region;
 import org.apache.geode.cache.RegionEvent;
 import org.apache.geode.cache.RegionFactory;
@@ -88,6 +86,7 @@ public class PartitionedRegionClearWithRebalanceDUnitTest implements Serializabl
   private VM accessor;
   private VM server1;
   private VM server2;
+  private VM server3;
 
   private enum TestVM {
     ACCESSOR(0), SERVER1(1), SERVER2(2), SERVER3(3);
@@ -126,6 +125,7 @@ public class PartitionedRegionClearWithRebalanceDUnitTest implements Serializabl
     getBlackboard().initBlackboard();
     server1 = getVM(TestVM.SERVER1.vmNumber);
     server2 = getVM(TestVM.SERVER2.vmNumber);
+    server3 = getVM(TestVM.SERVER3.vmNumber);
     accessor = getVM(TestVM.ACCESSOR.vmNumber);
   }
 
@@ -192,6 +192,7 @@ public class PartitionedRegionClearWithRebalanceDUnitTest implements Serializabl
           .setPartitionAttributes(attributesFactory.create())
           .setCacheWriter(new BlackboardSignaller());
 
+      // Set up the disk store if the region is persistent
       if (regionShortcut.isPersistent()) {
         factory.setDiskStoreName(cacheRule.getCache()
             .createDiskStoreFactory()
@@ -203,7 +204,8 @@ public class PartitionedRegionClearWithRebalanceDUnitTest implements Serializabl
     });
   }
 
-  private void parametrizedSetup(RegionShortcut regionShortcut, Collection<String> regionNames) {
+  private void parametrizedSetup(RegionShortcut regionShortcut, Collection<String> regionNames,
+      boolean useAccessor) {
     // Create and populate the region on server1 first, to create an unbalanced distribution of data
     server1.invoke(() -> {
       initDataStore(regionShortcut, regionNames);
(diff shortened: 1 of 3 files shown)


geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithRebalanceDUnitTest.java
```diff
@@ -0,0 +1,376 @@
...
+  @Parameters(method = "vmsAndRegionTypes")
+  @TestCaseName("[{index}] {method}(ClearCoordinator:{0}, RebalanceCoordinator:{1}, RegionType:{2})")
+  public void clearRegionDuringRebalancePrimaryReassignmentClearsRegion(TestVM clearCoordinatorVM,
+      TestVM rebalanceVM, RegionShortcut regionType) throws InterruptedException {
+    AsyncInvocation<?> clearInvocation = setupAndPrepareClear(clearCoordinatorVM, regionType);
+
+    getVM(rebalanceVM.vmNumber).invoke(() -> {
+      // Start a rebalance and wait until primary reassignment has started before signalling the
+      // blackboard
+      RebalanceOperation rebalanceOp =
```
agingade, line 297: Its same as "startRebalanceAndGetResults()"; can it be called here as done in previous test.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithRebalanceDUnitTest.java
```diff
@@ -0,0 +1,376 @@
...
+  public static class MemberKiller extends CacheWriterAdapter<String, String> {
+
+    @Override
+    public synchronized void beforeRegionClear(RegionEvent<String, String> event)
+        throws CacheWriterException {
+      InternalDistributedSystem.getConnectedInstance().stopReconnectingNoDisconnect();
+      MembershipManagerHelper.crashDistributedSystem(
+          InternalDistributedSystem.getConnectedInstance());
+      await().untilAsserted(
+          () -> Assertions.assertThat(InternalDistributedSystem.getConnectedInstance()).isNull());
```
agingade, line 372: Do we need additional check to see cache is closed...When this vm is restarted, it creates the cache again.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithRebalanceDUnitTest.java
DonalEvans: This case is actually slightly different; it waits for `getRebalance**PrimaryTransfers**Completed()` to be greater than zero instead of `getRebalance**BucketCreates**Completed()`. This case is intended to test the behaviour of clear when rebalance is doing the reassigning primaries step specifically.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithRebalanceDUnitTest.java
```diff
@@ -0,0 +1,376 @@
...
+    });
+  }
+
+  private RebalanceResults startRebalanceAndGetResults() throws InterruptedException {
+    // Start a rebalance and wait until bucket creation for redundancy recovery (the first stage of
+    // a rebalance operation) has started before signalling the blackboard
+    RebalanceOperation rebalanceOp =
+        cacheRule.getCache().getResourceManager().createRebalanceFactory().start();
+    await().untilAsserted(() -> assertThat(cacheRule.getCache().getInternalResourceManager()
+        .getStats().getRebalanceBucketCreatesCompleted(), greaterThan(0)));
```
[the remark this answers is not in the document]
DonalEvans: There are 10,000 entries in the region, so the rebalance process takes some time. The clear should always start before the rebalance finishes, unless the rebalance manages to finish in only a few milliseconds. Is there a solution to this timing issue other than just making sure the rebalance takes longer? The `clear()` call will always have some possible delay in starting due to the `waitForGate()` call.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithRebalanceDUnitTest.java
DonalEvans: I can add a check.

mhansonp:
Sorry all, I don't know how I closed this! I didn't intend to.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithRebalanceDUnitTest.java
```diff
@@ -0,0 +1,386 @@
...
+import org.apache.geode.test.dunit.AsyncInvocation;
+import org.apache.geode.test.dunit.DUnitBlackboard;
+import org.apache.geode.test.dunit.SerializableRunnableIF;
+import org.apache.geode.test.dunit.VM;
+import org.apache.geode.test.dunit.rules.CacheRule;
+import org.apache.geode.test.dunit.rules.DistributedDiskDirRule;
+import org.apache.geode.test.dunit.rules.DistributedRule;
+
+@RunWith(JUnitParamsRunner.class)
+public class PartitionedRegionClearWithRebalanceDUnitTest implements Serializable {
```
agingade, line 63: Looks good. If it is fine with you, can you add one more scenario, where there are 3 data nodes/servers and one of the server is brought down (case 1) and brought up (case 2); during clear, rebalance in progress....

DonalEvans: In the first scenario, would we expect clear to succeed? If a member departs during clear, we get `PartitionedRegionPartialClearException`, right? I can definitely add the second scenario though.

agingade: Based on the redundant copies available; the clear should finish successfully, even with server down. The clear operation waits certain time for secondary to become primary, if it is unable to find the primary in given time, then it will throw Partial Clear exception.

DonalEvans: Using three servers, with a PartitionedRegion with one redundant copy, and killing the non-coordinator server when the clear operation starts, I am still seeing a PartialClearException some of the time. It seems to be flaky whether the clear operation has enough time to find the new primary in the case that a server goes down. This happens even if there is no rebalance happening at the time; as long as some data is hosted on the member that is killed, there is a chance of a PartialClearException, it seems. I don't know if it will be possible to write a useful test in this scenario, since we have to either assert that the clear succeeds, or that it fails, but it doesn't do either reliably.

PR #5097, opened 2020-05-12
GEODE-8111: Fix create gw receiver --manual-start doc

alb3rtobr:
Documentation of `--manual-start` option of `create gateway-receiver` gfsh command wrongly states that the default option is `true`.
But taking a look at the code it can be verified that the default option is false.


alb3rtobr:
> +1 Much clearer now. Thanks!

@davebarnes97 I dont have permissions for merging the PR, could you do it for me? DistributedTest check failed but not due to this documentation change :)

PR #5099, opened 2020-05-12
GEODE-8029: Delete orphaned drf files (2nd try)

jujoramos:
This reverts the revert of the original commit and also fixes the
originally introduced issues.

----

The OpLog initialization now delete unused drf files to prevent the
proliferation of unused records and files within the system, which
can could cause members to fail during startup while recovering
disk-stores (especially when they are isolated for gateway-senders).

- Added distributed tests.
- Delete orphaned drf files when deleting the corresponding
  crf during recovery.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/Oplog.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/Oplog.java
@@ -937,14 +937,21 @@ public class Oplog implements CompactableOplog, Flushable {
         // this.crf.raf.seek(this.crf.currSize);
       } else if (!offline) {
         // drf exists but crf has been deleted (because it was empty).
-        // I don't think the drf needs to be opened. It is only used during
-        // recovery.
-        // At some point the compacter my identify that it can be deleted.
         this.crf.RAFClosed = true;
         deleteCRF();
+
+        // See GEODE-8029.
+        // The drf file needs to be deleted, especially when the disk-store is *only* used by
+        // gateway-senders, otherwise there will be orphaned drfs that are never deleted by
+        // compaction (unless there are pending events in the queue upon restart - crf exists - or
+        // a manual compaction is executed).
+        deleteDRF();
+        setHasDeletes(false);
+
         this.closed = true;
         this.deleted.set(true);
       }
+
       this.drf.RAFClosed = true; // since we never open it on a recovered oplog
     } catch (IOException ex) {
       getParent().getCancelCriterion().checkCancelInProgress(ex);
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/Oplog.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/Oplog.java
@@ -940,13 +940,13 @@ public class Oplog implements CompactableOplog, Flushable {
         this.crf.RAFClosed = true;
         deleteCRF();
 
-        // See GEODE-8029.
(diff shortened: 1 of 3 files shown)


geode-core/src/main/java/org/apache/geode/internal/cache/Oplog.java
```diff
@@ -937,14 +937,21 @@ void initAfterRecovery(boolean offline) {
         // this.crf.raf.seek(this.crf.currSize);
       } else if (!offline) {
         // drf exists but crf has been deleted (because it was empty).
-        // I don't think the drf needs to be opened. It is only used during
-        // recovery.
-        // At some point the compacter my identify that it can be deleted.
         this.crf.RAFClosed = true;
         deleteCRF();
+
+        // See GEODE-8029.
```
agingade, line 943: This change comes into picture after the recovery is completed - "initAfterRecovery".
The real problem seems to be during recovery. If there is a periodic recovery this could be fine, but if there are large number of deleted records during the first recovery, the issue may still arise...We may need to have some checks during recovery (before reading the drfs) to avoid reading large deleted records. Please correct me if my understanding is wrong. 

Also, looking at the other part of the code where "setHasDeletes" is called, it looks like it is called before the deleteDRF() - there could be a reason for doing this.
And there is also call to "getOplogSet().removeDrf(this);" This may be needed here...

jujoramos: > This change comes into picture after the recovery is completed - "initAfterRecovery".
The real problem seems to be during recovery. If there is a periodic recovery this could be fine, but if there are large number of deleted records during the first recovery, the issue may still arise...We may need to have some checks during recovery (before reading the drfs) to avoid reading large deleted records. Please correct me if my understanding is wrong.

Your understanding is correct, however, the change are to prevent the issue from happening in the first place. By deleting the unused `drf` files here and avoid storing them on disk when not needed, users won't hit the `IllegalStateException` at all.
As a side note, in order to hit the problem in the first place, the user would need to have more than `805306401` delete operations within the `opLog` files in one single run, and compaction should have not run at all, which is highly unlikely.
If you think we should add the count while reading the files instead and, somehow, expand the load factor when needed or something similar, let me know and I'll give it a try. I'm worried, though, about the performance impact this might have while recovering files from disk... the recovery time will suffer during every single startup and, considering that there's only a handful of scenarios on which the actual issue can happen, I believe making sure we delete unused files is a better approach.

> Also, looking at the other part of the code where "setHasDeletes" is called, it looks like it is called before the deleteDRF() - there could be a reason for doing this.

Will change this, thanks for catching it.

> And there is also call to "getOplogSet().removeDrf(this);" This may be needed here...

This is already done within the `deleteDRF()` method.

---

Please let me know what you think about point 1 (add an extra check while recovering to make sure we are below the load-factor threshold), so I can go ahead and make all the changes at once.

agingade: I am fine with the approach you have taken. Wanted to bring it up, there could be distinct possibility.
Looks good to me.

jujoramos: @agingade 
Thanks, I've made the requested changes already and verified that no new regressions have been introduced due to the modifications, please go ahead and approve the `PR` (if there's nothing else you want me to change) so I can merge it 👍.

PR #5100, opened 2020-05-12
GEODE-8114: Refactor set

prettyClouds:
I broke up the `RedisSet` and `RedisHash` classes by moving the static methods into separate non-static classes.  I also made canonical empty classes, so we don't have to do as much null-checking.

### For all changes:

### Note:

### DIFF
--- a/geode-redis/src/main/java/org/apache/geode/redis/internal/executor/set/RedisSet.java
+++ b/geode-redis/src/main/java/org/apache/geode/redis/internal/executor/set/RedisSet.java
@@ -46,6 +46,7 @@ import org.apache.geode.redis.internal.Coder;
  */
 public class RedisSet implements Delta, DataSerializable {
 
+  public static transient RedisSet EMPTY = new EmptyRedisSet();
   private HashSet<ByteArrayWrapper> members;
   private transient ArrayList<ByteArrayWrapper> deltas;
   // true if deltas contains adds; false if removes
--- a/geode-redis/src/main/java/org/apache/geode/redis/internal/executor/set/RedisSet.java
+++ b/geode-redis/src/main/java/org/apache/geode/redis/internal/executor/set/RedisSet.java
@@ -16,7 +16,6 @@
 package org.apache.geode.redis.internal.executor.set;
 
 import static java.util.Collections.emptyList;
-import static java.util.Collections.emptySet;
 
 import java.io.DataInput;
 import java.io.DataOutput;
@@ -64,101 +63,7 @@ public class RedisSet implements Delta, DataSerializable {
   // for serialization
   public RedisSet() {}
 
-  public static long sadd(Region<ByteArrayWrapper, RedisSet> region,
-      ByteArrayWrapper key,
-      ArrayList<ByteArrayWrapper> membersToAdd) {
-
-    RedisSet redisSet = region.get(key);
-
-    if (redisSet != null) {
-      // update existing value
-      return redisSet.doSadd(membersToAdd, region, key);
-    } else {
-      region.create(key, new RedisSet(membersToAdd));
-      return membersToAdd.size();
-    }
-  }
-
-  public static long srem(Region<ByteArrayWrapper, RedisSet> region,
-      ByteArrayWrapper key,
-      ArrayList<ByteArrayWrapper> membersToRemove,
-      AtomicBoolean setWasDeleted) {
-    RedisSet redisSet = region.get(key);
-    long numRemoved;
-    if (redisSet == null) {
-      numRemoved = 0;
-    } else {
-      numRemoved = redisSet.doSrem(membersToRemove, region, key, setWasDeleted);
-    }
-    return numRemoved;
-  }
-
-  public static boolean del(Region<ByteArrayWrapper, RedisSet> region,
-      ByteArrayWrapper key) {
-    return region.remove(key) != null;
-  }
-
-  public static Set<ByteArrayWrapper> smembers(Region<ByteArrayWrapper, RedisSet> region,
-      ByteArrayWrapper key) {
-    RedisSet redisSet = region.get(key);
-    if (redisSet != null) {
-      return redisSet.members();
-    } else {
-      return emptySet();
-    }
-  }
-
-  public static int scard(Region<ByteArrayWrapper, RedisSet> region, ByteArrayWrapper key) {
-    RedisSet redisSet = region.get(key);
-    if (redisSet != null) {
-      return redisSet.size();
-    } else {
-      return 0;
-    }
-  }
-
-  public static boolean sismember(Region<ByteArrayWrapper, RedisSet> region,
-      ByteArrayWrapper key, ByteArrayWrapper member) {
-    RedisSet redisSet = region.get(key);
-    if (redisSet != null) {
-      return redisSet.contains(member);
-    } else {
-      return false;
-    }
-  }
-
-  public static Collection<ByteArrayWrapper> srandmember(Region<ByteArrayWrapper, RedisSet> region,
-      ByteArrayWrapper key, int count) {
-    RedisSet redisSet = region.get(key);
-    if (redisSet != null) {
-      return redisSet.srandmember(count);
-    } else {
-      return emptyList();
-    }
-  }
-
-  public static Collection<ByteArrayWrapper> spop(Region<ByteArrayWrapper, RedisSet> region,
-      ByteArrayWrapper key, int popCount) {
-    RedisSet redisSet = region.get(key);
-    if (redisSet != null) {
-      return redisSet.doSpop(region, key, popCount);
-    } else {
-      return emptyList();
-    }
-  }
-
-  public static List<Object> sscan(Region<ByteArrayWrapper, RedisSet> region,
-      ByteArrayWrapper key, Pattern matchPattern, int count, int cursor) {
-    RedisSet RedisSet = region.get(key);
-    if (RedisSet != null) {
-      return RedisSet.doSscan(matchPattern, count, cursor);
-    } else {
-      return emptyList();
-    }
-  }
-
-
-  private synchronized List<Object> doSscan(Pattern matchPattern, int count, int cursor) {
+  synchronized List<Object> doSscan(Pattern matchPattern, int count, int cursor) {
     List<Object> returnList = new ArrayList<>();
     int size = members.size();
     int beforeCursor = 0;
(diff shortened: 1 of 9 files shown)


geode-redis/src/main/java/org/apache/geode/redis/internal/executor/set/RedisSetInRegion.java
```diff
@@ -0,0 +1,115 @@
...
+    return l;
+  }
+
+  @Override
+  public long srem(
+      ByteArrayWrapper key,
+      ArrayList<ByteArrayWrapper> membersToRemove, AtomicBoolean setWasDeleted) {
+    RedisSet redisSet = get(key);
+    long l = redisSet.srem(key, membersToRemove, setWasDeleted);
+    if (get(key).scard(key) == 0) {
```
dschneider-pivotal, line 57: change "get(key)" to "redisSet"

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/set/RedisSetInRegion.java
```diff
@@ -0,0 +1,115 @@
...
+      region.remove(key);
+    } else {
+      region.put(key, get(key));
```
dschneider-pivotal, line 60: change "get(key)" to "redisSet"

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/set/RedisSetInRegion.java
```diff
@@ -0,0 +1,115 @@
...
+      ByteArrayWrapper key, int count) {
+    return get(key).srandmember(key, count);
+  }
+
+  @Override
+  public Collection<ByteArrayWrapper> spop(
+      ByteArrayWrapper key, int popCount) {
+    RedisSet redisSet = get(key);
+    Collection<ByteArrayWrapper> byteArrayWrappers = redisSet.spop(key, popCount);
+    if (get(key).scard(key) == 0) {
```
dschneider-pivotal, line 97: change "get(key)" to "redisSet"

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/set/RedisSetInRegion.java
```diff
@@ -0,0 +1,115 @@
...
+      region.remove(key);
+    } else {
+      region.put(key, get(key));
```
dschneider-pivotal, line 100: change "get(key)" to "redisSet"

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/set/RedisSetInRegion.java
```diff
@@ -0,0 +1,115 @@
...
+
+  @Override
+  public List<Object> sscan(
+      ByteArrayWrapper key, Pattern matchPattern, int count, int cursor) {
+    return get(key).sscan(key, matchPattern, count, cursor);
+  }
+
+  private RedisSet get(ByteArrayWrapper key) {
+    RedisSet redisSet = region.get(key);
+    return redisSet == null ? new RedisSet(new ArrayList<>()) : redisSet;
```
dschneider-pivotal, line 113: I prefer getting null back from this method to prevent garbage creation. I know you prefer keeping the callers simpler by always returning a non-null. Can you return a canonical empty instance instead of always creating a new  RedisSet  and a new ArrayList?

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/set/RedisSet.java
```diff
@@ -304,70 +211,60 @@ public void fromData(DataInput in) throws IOException, ClassNotFoundException {
...
    * @return the number of members actually removed; -1 if concurrent modification
    */
-  private synchronized long doSrem(ArrayList<ByteArrayWrapper> membersToRemove,
-      Region<ByteArrayWrapper, RedisSet> region,
-      ByteArrayWrapper key, AtomicBoolean setWasDeleted) {
-
+  @Override
+  public synchronized long srem(ByteArrayWrapper key, ArrayList<ByteArrayWrapper> membersToRemove,
+      AtomicBoolean setWasDeleted) {
+    deltas = null;
```
dschneider-pivotal, line 240: I don't think they way you have refactored the delta code is thread safe. As long as we have set ops that do not use the function+delta (that is why this method is synchronized still) then we need to send the delta before this synchronization is released.
I also don't like that we leave deltas set until the next operation is done. This will just extend its life when it should have been garbage before this method released in sync. I think this issue with the delta management is throughout this class.

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/set/RedisSet.java
```diff
@@ -34,18 +33,15 @@
...
- * underlying HashSet even though all writers do so under
- * the {@link SynchronizedStripedExecutor}. The synchronization on this
- * class can be removed once readers are changed to
- * also use the {@link SynchronizedStripedExecutor}.
+ * This class still uses "synchronized" to protect the underlying HashSet even though all writers do
+ * so under the {@link SynchronizedStripedExecutor}. The synchronization on this class can be
+ * removed once readers are changed to also use the {@link SynchronizedStripedExecutor}.
  */
-public class RedisSet implements Delta, DataSerializable {
+public class RedisSet implements Delta, DataSerializable, RedisSetCommands {
```
dschneider-pivotal, line 44: I don't think you should force RedisSetCommands on this class. None of the methods on this class currently need the "key" parameter because they do not need to find themselves in the region. But RedisSetCommands methods do  need a "key". I think that abstraction really does not fit this class.

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/set/RedisSet.java
dschneider-pivotal: Maybe we could pass the RedisSetInRegion into the methods that need to update the region. If we added the "key" to RedisSetInRegion then it would know what key it is supposed to remove or put when we callback to it.
This would also allow the RedisSetInRegion and RedisSet to both implement a common interface (neither of their methods would have "key" parameters).

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/set/RedisSetInRegion.java
```diff
@@ -0,0 +1,115 @@
...
+import java.util.regex.Pattern;
+
+import org.apache.geode.cache.Region;
+import org.apache.geode.redis.internal.ByteArrayWrapper;
+
+/**
+ * This class checks whether a RedisSet exists before forwarding the message.
+ */
+public class RedisSetInRegion implements RedisSetCommands {
+  private Region<ByteArrayWrapper, RedisSet> region;
```
dschneider-pivotal, line 32: make this field final

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/set/RedisSet.java
prettyClouds: yea, i may have been too ambitious in trying to get all the layers to conform to this interface...i think it's a good long-term goal to have, but we still need some changes to make this possible..As long as some commands go through the executor and some take a "side" route, it will be difficult to get this pattern implemented.

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/set/RedisSet.java
prettyClouds: ideally, i think we can remove the key from the `RedisSetCommands` interface.  I think that's a more OO approach where the key is the identity of the object and not something you can pass in.  Right now the `RedisSetCommands` interface is more like a bag of functions.  What do you think?  

I think that change is a little bit down the road, though, once all the set commands are   moved over.

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/set/RedisSetInRegion.java
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  public boolean del(ByteArrayWrapper key) {
+    return region.remove(key) != null;
+  }
+
+  @Override
+  public Set<ByteArrayWrapper> smembers(
+      ByteArrayWrapper key) {
+    RedisSet redisSet = region.getOrDefault(key, RedisSet.EMPTY);
```
dschneider-pivotal, line 71: Would it be worth refactoring this "region.getOrDefault(key, RedisSet.EMPTY)" into a local private method named "getRedisSet(key)" and then these methods can just be a one liner: "return getRedisSet(key).smembers()"

prettyClouds: sure, i can do that.

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/set/RedisSet.java
dschneider-pivotal: When JohnH and I worked on renaming RedisSetCommands (from RedisSet), we chose "Commands" because it seemed like a bag of functions. And from the point of view of the netty executors (like SAddExecutor) I think this may make sense. I like for the interface methods to have a close correspondence to the actual redis command line arguments. Since the REDIS SADD takes a "key" and "members" then the method on the Commands interface should have both a "key" and "members". It should not have "region" since the REDIS SADD command does not have a region. So that is why when you create a RedisSetCommandsFunctionExecutor you give it a Region but you do not give it the key. This also means you could use the same instance of RedisSetCommandsFunctionExecutor to do multiple commands on different keys.
But once you get down to having on instance of our data structure (RedisSet in this example) then you no longer need the key because the key was used to find it.
If I had to choose one or the other (methods with the key or methods without), I think I'd chose with the key since it corresponds to the redis apis.

dschneider-pivotal:
you should update the description of this PR. You got rid of the statics on RedisSet and RedisHash. You also introduced a singleton EMPTY for both of them.

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/set/RedisSet.java
prettyClouds: I can see both sides of this coin.  

I think it's nice that the `RedisSetCommands` match the Redis API. 
I also think that once we are in the executor we are in the Java code, we are now doing OO programming and don't need our interfaces to conform to Redis' binary protocol.

The "user" I imagine implementing this interface is a developer looking to implement the Redis data structures in a completely new way...in geode or in some other platform.  I don't have enough of that architecture worked out to be able to make a clear decision, so at this point I am okay either way.

PR #5102, opened 2020-05-12
GEODE-8112: Add --member option in query command.

pivotal-eshu:


geode-dunit/src/main/java/org/apache/geode/management/internal/cli/commands/QueryCommandDUnitTestBase.java
```diff
@@ -310,4 +323,40 @@ public void setValue1(Value1 value1) {
...
+
+    server1.invoke(() -> prepareDataForRegion(DATA_REGION_WITH_PROXY_NAME_PATH));
+
+    String member = getHostingMember();
+    Random random = new Random(System.nanoTime());
+    int randomInteger = random.nextInt(COUNT);
+    String query = "query --member=" + member
+        + " --query=\"select ID , status , createTime , pk, floatMinValue from "
+        + DATA_REGION_WITH_PROXY_NAME_PATH + " where ID <= " + randomInteger
+        + "\" --interactive=false";
```
DonalEvans, line 341: Could this use `CommandStringBuilder`? It makes things easier to read in my experience.

geode-dunit/src/main/java/org/apache/geode/management/internal/cli/commands/QueryCommandDUnitTestBase.java
```diff
@@ -310,4 +323,40 @@ public void setValue1(Value1 value1) {
...
+
+    CommandResult commandResult = gfsh.executeCommand(query);
```
DonalEvans, line 343: `executeCommand()` is deprecated. You can use `executeAndAssertThat(query).getCommandResult()` instead.

geode-dunit/src/main/java/org/apache/geode/management/internal/cli/commands/QueryCommandDUnitTestBase.java
```diff
@@ -310,4 +323,40 @@ public void setValue1(Value1 value1) {
...
+    validateSelectResult(commandResult, true, (randomInteger + 1),
+        new String[] {"ID", "status", "createTime", "pk", "floatMinValue"});
+  }
+
+  private String getHostingMember() {
+    String hostingMember = null;
+    String findMemberCommand = "describe region --name=" + DATA_REGION_WITH_PROXY_NAME;
+    CommandResult findMemberResult = gfsh.executeCommand(findMemberCommand);
```
DonalEvans, line 351: `executeCommand()` is deprecated. You can use `executeAndAssertThat(query).getCommandResult()` instead.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/QueryCommand.java
```diff
@@ -60,12 +60,17 @@ public ResultModel query(
       @CliOption(key = "file", help = "File in which to output the results.",
           optionContext = ConverterHint.FILE) final File outputFile,
       @CliOption(key = CliStrings.QUERY__INTERACTIVE, unspecifiedDefaultValue = "false",
-          help = CliStrings.QUERY__INTERACTIVE__HELP) final boolean interactive) {
-    DataCommandResult dataResult = select(query);
+          help = CliStrings.QUERY__INTERACTIVE__HELP) final boolean interactive,
+      @CliOption(key = CliStrings.MEMBER,
+          optionContext = ConverterHint.MEMBERIDNAME,
+          help = CliStrings.QUERY__MEMBER__HELP) final String memberNameOrId) {
```
DonalEvans, line 66: This new command option should be documented in `query.html.md.erb`. Also, I notice that that page has some inaccuracies that could be fixed while you're updating it. `[--step-name=value]` is listed as an argument, but not represented in the code,  and the `--file` option described in the docs seems to not be used at all in the code. The `--interactive` option is also not used anywhere in the code, or documented, so I wonder if it could just be removed.

geode-core/src/main/java/org/apache/geode/management/internal/i18n/CliStrings.java
```diff
@@ -1823,6 +1823,8 @@
       "Cannot find regions <{0}> in any of the members";
   public static final String QUERY__MSG__NOT_SUPPORTED_ON_MEMBERS = CliStrings.QUERY
       + " command should be used only from shell. Use QueryService API for running query inside Geode VMs";
+  public static final String QUERY__MEMBER__HELP =
+      "Name/Id of a member which hosts the regions to be queried.";
```
jujoramos, line 1827: Maybe change this to `Name/Id of the member on which to execute the query` instead?, I think is easier to understand and doesn't generate doubts on the operator/user executing the command.

geode-dunit/src/main/java/org/apache/geode/management/internal/cli/commands/QueryCommandDUnitTestBase.java
```diff
@@ -256,6 +260,15 @@ private static void setupReplicatedRegion(String regionName) {
...
 
+  private static void setupReplicatedProxyRegion(String regionName) {
+    InternalCache cache = ClusterStartupRule.getCache();
+    RegionFactory regionFactory = cache.createRegionFactory(RegionShortcut.REPLICATE_PROXY);
+
+    Region proxyRegion = regionFactory.create(regionName);
+    assertThat(proxyRegion).isNotNull();
+    assertThat(proxyRegion.getFullPath()).contains(regionName);
+  }
+
```
jujoramos, line 271: Small nitpicks: since the region name is always the same (`DATA_REGION_WITH_PROXY_NAME`), can we use the constant instead of the parameter?.
Also, to prevent introducing unnecessary warnings, can we avoid the usage of `rawtypes` in the method?:
```
RegionFactory<Integer, Portfolio> regionFactory = cache.createRegionFactory(RegionShortcut.REPLICATE_PROXY);
Region<Integer, Portfolio> proxyRegion = regionFactory.create(regionName);
```

geode-dunit/src/main/java/org/apache/geode/management/internal/cli/commands/QueryCommandDUnitTestBase.java
```diff
@@ -310,4 +323,40 @@ public void setValue1(Value1 value1) {
       this.value1 = value1;
     }
   }
+
+  @Test
+  public void testSimpleQueryWithProxyRegion() {
```
jujoramos, line 328: Would it make sense to add the same test for a `PARTITION_PROXY` region?, or the changes within this `PR` doesn't affect that use case?.

geode-dunit/src/main/java/org/apache/geode/management/internal/cli/commands/QueryCommandDUnitTestBase.java
jujoramos: +1

geode-dunit/src/main/java/org/apache/geode/management/internal/cli/commands/QueryCommandDUnitTestBase.java
jujoramos: +1

geode-dunit/src/main/java/org/apache/geode/management/internal/cli/commands/QueryCommandDUnitTestBase.java
jujoramos: +1

geode-core/src/main/java/org/apache/geode/management/internal/i18n/CliStrings.java
```diff
@@ -1823,6 +1823,8 @@
       "Cannot find regions <{0}> in any of the members";
   public static final String QUERY__MSG__NOT_SUPPORTED_ON_MEMBERS = CliStrings.QUERY
       + " command should be used only from shell. Use QueryService API for running query inside Geode VMs";
+  public static final String QUERY__MEMBER__HELP =
```
jinmeiliao, line 1826: you probably want to consult doc's team on this. To me something that says: "the member to start the query on, usually should be the member that holds the data of the region." would be more helpful.

geode-dunit/src/main/java/org/apache/geode/management/internal/cli/commands/QueryCommandDUnitTestBase.java
```diff
@@ -310,4 +323,40 @@ public void setValue1(Value1 value1) {
...
+    server1.invoke(() -> setupReplicatedProxyRegion(DATA_REGION_WITH_PROXY_NAME));
+    server2.invoke(() -> setupReplicatedRegion(DATA_REGION_WITH_PROXY_NAME));
+    locator.waitUntilRegionIsReadyOnExactlyThisManyServers(DATA_REGION_WITH_PROXY_NAME_PATH, 2);
+
+    server1.invoke(() -> prepareDataForRegion(DATA_REGION_WITH_PROXY_NAME_PATH));
+
+    String member = getHostingMember();
```
jinmeiliao, line 335: for this test, you know server1 has the data and server2 is the proxy, so you don't really need to use "describe region" to get this information, you can simply add "--member=server-1" in your query command

geode-gfsh/src/test/java/org/apache/geode/management/internal/cli/commands/QueryCommandTest.java
```diff
@@ -0,0 +1,72 @@
...
+
+    spyCommand.query(query, null, false, null);
+
+    verify(spyCommand).select(query, null);
+  }
+
+  @Test
+  public void targetMemberIsSetIfMemberOptionsIsUsed() {
+    QueryCommand spyCommand = spy(command);
+    DistributedMember member = mock(DistributedMember.class);
```
jinmeiliao, line 48: you can put these two lines in a @Before setup method so that you don't need to repeat that for every test.

geode-dunit/src/main/java/org/apache/geode/management/internal/cli/commands/QueryCommandDUnitTestBase.java
pivotal-eshu: It does not affect PR.

geode-core/src/main/java/org/apache/geode/management/internal/i18n/CliStrings.java
davebarnes97: +1

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/QueryCommand.java
davebarnes97: I can add --member=value to the gfsh query command description.
Not sure about removing --step-name and --file. Is it possible those are present for legacy support (like, they used to be supported in earlier versions, so need to be kept around?)

DonalEvans: I talked to Jens about it briefly and the `--file` argument is definitely used, it just gets picked up by the interceptor on the gfsh command, not used in the actual command itself. If you try to include the `--step-name` option, then gfsh throws an error: `Option 'step-name' is not available for this command.` The "step-name" string only appears in this one place too, with no other occurrences in the codebase or the docs, so I can't think of a reason to not remove it here.

jinmeiliao: `--step-name` and `--interactive` can be removed in the docs since they are not used by the command anymore, but they can NOT be removed in the option list for backward compatibility reasons (user with old scripts might break).

davebarnes97:
Update to the gfsh query command documentation has been pushed. Should we also add something about intended usage in the Developing -> Querying section of the User Guide? For example, “Querying a Partitioned Region on a Single Node”, https://geode.apache.org/docs/guide/112/developing/query_additional/query_on_a_single_node.html#concept_30B18A6507534993BD55C2C9E0544A97

geode-dunit/src/main/java/org/apache/geode/management/internal/cli/commands/QueryCommandDUnitTestBase.java
pivotal-eshu: Fixed.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/QueryCommand.java
pivotal-eshu: Thanks Dave for helping on the docs.

geode-dunit/src/main/java/org/apache/geode/management/internal/cli/commands/QueryCommandDUnitTestBase.java
pivotal-eshu: Fixed.

geode-dunit/src/main/java/org/apache/geode/management/internal/cli/commands/QueryCommandDUnitTestBase.java
```diff
@@ -332,31 +333,18 @@ public void testSimpleQueryWithProxyRegion() {
...
+    String member = "server-2";
     Random random = new Random(System.nanoTime());
     int randomInteger = random.nextInt(COUNT);
-    String query = "query --member=" + member
-        + " --query=\"select ID , status , createTime , pk, floatMinValue from "
-        + DATA_REGION_WITH_PROXY_NAME_PATH + " where ID <= " + randomInteger
-        + "\" --interactive=false";
-
-    CommandResult commandResult = gfsh.executeCommand(query);
+    StringBuilder queryString = new StringBuilder();
```
DonalEvans, line 339: I think there was some confusion about my original review comment. I meant to use the Geode `CommandStringBuilder` class, not the Java `StringBuilder`.

PR #5107, opened 2020-05-13
GEODE-8120: Make GeodeRedisServer internal

sabbeyPivotal:
GeodeRedisServer does not need to be externally accessible.


sabbeyPivotal:
Jira to fix JavaDocs that reference GeodeRedisServer: https://issues.apache.org/jira/browse/GEODE-8122

PR #5111, opened 2020-05-13
GEODE-8078: log and report error at the correct place.

jinmeiliao:
* For get operation, if cache is not ready, just don't report any runtime information
* We would still log errors in case the stack trace is useful.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/management/internal/functions/CacheRealizationFunction.java
+++ b/geode-core/src/main/java/org/apache/geode/management/internal/functions/CacheRealizationFunction.java
@@ -36,6 +36,7 @@ import org.apache.commons.io.IOUtils;
 import org.apache.logging.log4j.Logger;
 
 import org.apache.geode.annotations.Immutable;
+import org.apache.geode.cache.CacheClosedException;
 import org.apache.geode.cache.execute.FunctionContext;
 import org.apache.geode.internal.cache.InternalCache;
 import org.apache.geode.internal.cache.execute.InternalFunction;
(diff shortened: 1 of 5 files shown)


geode-web-management/src/distributedTest/java/org/apache/geode/management/client/GetStartingMemberTest.java
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  @Test
+  public void getStartingMember() throws Exception {
+    Member server0 = new Member();
+    server0.setId("server-0");
+    concurrency.add(() -> cluster.startServerVM(0, webContext.getLocator().getPort()));
+    concurrency.add(() -> {
+      ClusterManagementGetResult<Member, MemberInformation> result = null;
+      while (result == null) {
```
kirklund, line 78: I recommend using GeodeAwaitility instead of a while-loop without timeout (I realize that ConcurrencyRule is ultimately providing the cancellation):
```
await()
  .ignoreException(ClusterManagementException.class)
  .until(() -> client.get(server0) != null);
```

geode-web-management/src/distributedTest/java/org/apache/geode/management/client/GetStartingMemberTest.java
```diff
@@ -0,0 +1,89 @@
...
+        try {
+          result = client.get(server0);
+        } catch (ClusterManagementException e) {
+          // this is expected for the first several tries
+        }
+      }
+      return result;
+    });
+    concurrency.executeInParallel();
```
kirklund, line 87: ExecutorServiceRule would really clean this up and I think we should look into merging these two rules. In general, I think ExecutorServiceRule is preferable (I can explain outside of PR). If you switch to ExecutorServiceRule the test could use an assertion in await and seems much more readable to me:
```
import static java.util.concurrent.TimeUnit.MILLISECONDS;
import static org.apache.geode.test.awaitility.GeodeAwaitility.await;

  @Rule
  public ExecutorServiceRule executorServiceRule = new ExecutorServiceRule();

  @Test
  public void getStartingMember() throws Exception {
    Member server0 = new Member();
    server0.setId("server-0");

    Future<Void> startServer = executorServiceRule.submit(() -> {
      cluster.startServerVM(0, webContext.getLocator().getPort());
    });

    await()
        .ignoreException(ClusterManagementException.class)
        .untilAsserted(() -> assertThat(client.get(server0)).isNotNull());

    startServer.get(getTimeout().toMillis(), MILLISECONDS);
  }
```

jinmeiliao: Great, thank you so much!

geode-core/src/main/java/org/apache/geode/management/internal/functions/CacheRealizationFunction.java
```diff
@@ -78,24 +79,44 @@
...
-    InternalCache cache = (InternalCache) context.getCache();
-    try {
-      if (operation == CacheElementOperation.GET) {
+    // for get operation, caller is expecting RuntimeInfo
+    if (operation == CacheElementOperation.GET) {
+      try {
+        InternalCache cache = (InternalCache) context.getCache();
         context.getResultSender().lastResult(executeGet(context, cache, cacheElement));
-      } else {
+      } catch (CacheClosedException e) {
```
jchen21, line 89: I am not sure the `CacheClosedException` in the JIRA is thrown this function. This function does not call `ResultSender.sendException()` to the caller. How can the caller `LocatorClusterManagementService` get the `CacheClosedException` and then `ClassCastException`?

jinmeiliao: > I am not sure the `CacheClosedException` in the JIRA is thrown this function. This function does not call `ResultSender.sendException()` to the caller. How can the caller `LocatorClusterManagementService` get the `CacheClosedException` and then `ClassCastException`?

the exception is thrown by `context.getCache()` method, (https://github.com/apache/geode/blob/ba8a5528f462c86104deb531b18245db64efe324/geode-core/src/main/java/org/apache/geode/internal/cache/execute/FunctionContextImpl.java#L109). Before the fix, the exception is thrown and caught by the function executor and then it directly gave the exception back to the caller, but caller is expecting a `RuntimeInfo` object and tried to cast the exception to `RuntimeInfo`, thus the error.

PR #5114, opened 2020-05-13
Introduce RedisResponse to SET executors

jdeppe-pivotal:

### For all changes:

### Note:


geode-redis/src/main/java/org/apache/geode/redis/internal/RedisCommandType.java
```diff
@@ -349,9 +350,17 @@ private RedisCommandType(Executor executor, ParameterRequirements parameterRequi
     this.parameterRequirements = parameterRequirements;
   }
 
-  public void executeCommand(Command command, ExecutionHandlerContext executionHandlerContext) {
+  public RedisResponse executeCommand(Command command,
+      ExecutionHandlerContext executionHandlerContext) {
     parameterRequirements.checkParameters(command, executionHandlerContext);
+
+    if (executor instanceof ResponseExecutor) {
```
dschneider-pivotal, line 357: Could you change "Executor" to have "executeCommandsWithResponse"? For all the old executors we have not converted yet, executeCommandsWithResponse could just call executeCommand and then return null.
That would get rid of this instanceof check.

jdeppe-pivotal: Excellent! Good call.

geode-redis/src/main/java/org/apache/geode/redis/internal/ExecutionHandlerContext.java
```diff
@@ -203,43 +205,42 @@ public void channelInactive(ChannelHandlerContext ctx) {
...
+      response = RedisResponse.error(RedisConstants.ERROR_NOT_AUTH);
+    }
+
+    if (response != null) {
       writeToChannel(response);
     } else {
-      ByteBuf r = Coder.getNoAuthResponse(this.byteBufAllocator, RedisConstants.ERROR_NOT_AUTH);
-      writeToChannel(r);
+      // PUBLISH responses are always deferred
+      if (command.getCommandType() != RedisCommandType.PUBLISH) {
```
dschneider-pivotal, line 234: It seems like this if expression should use "!command.isOfType(RedisCommandType.PUBLISH)".

You could get this down to a single "writeToChannel" call like so:
if (response == null && !command.isOfType(PUBLISH) {
  response = command.getResponse();
}
if (response != null) {
  writeToChannel(response);
}

geode-redis/src/main/java/org/apache/geode/redis/internal/Executor.java
```diff
@@ -27,9 +27,13 @@
    * @param command The command to be executed
    * @param context The execution context by which this command is to be executed
    */
-  void executeCommand(Command command, ExecutionHandlerContext context);
+  default void executeCommand(Command command, ExecutionHandlerContext context) {
+    executeCommandWithResponse(command, context);
```
dschneider-pivotal, line 31: It looks like these two default impls call each other. executeCommand calls executeCommandWithResponse and then it calls back to executeCommand. Seems like it would do an infinite recursion until the stack overflows.

geode-redis/src/main/java/org/apache/geode/redis/internal/Executor.java
```diff
@@ -27,9 +27,13 @@
...
+  }
 
-  default void execute(Command command, ExecutionHandlerContext context) {
+  default RedisResponse executeCommandWithResponse(Command command,
+      ExecutionHandlerContext context) {
     executeCommand(command, context);
+    return null;
```
dschneider-pivotal, line 37: One other idea. I think the plan is to eventually have every executeCommand have a non-null response. But if we are going to continue to have some return "null" could they instead return an instance of RedisResponse to represents null?

geode-redis/src/main/java/org/apache/geode/redis/internal/ExecutionHandlerContext.java
jdeppe-pivotal: I made some small updates and I think it's better - YMMV :)

geode-redis/src/main/java/org/apache/geode/redis/internal/Executor.java
jdeppe-pivotal: Yes, that is a danger. Once all commands are converted we can fix this.

geode-redis/src/main/java/org/apache/geode/redis/internal/Executor.java
jdeppe-pivotal: Will do once we switch the `PUBLISH` command which is the only one (I can think of) that would return null.

geode-redis/src/main/java/org/apache/geode/redis/internal/ExecutionHandlerContext.java
```diff
@@ -56,7 +58,7 @@
   private static final Logger logger = LogService.getLogger();
   private static final int WAIT_REGION_DSTRYD_MILLIS = 100;
   private static final int MAXIMUM_NUM_RETRIES = (1000 * 60) / WAIT_REGION_DSTRYD_MILLIS; // 60
```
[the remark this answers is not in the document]
jdeppe-pivotal: Done

PR #5122, opened 2020-05-15
GEODE-8122: Update GeodeRedisServer java doc and Redis API for Geode documentation

sabbeyPivotal:
- Java doc for GeodeRedisServer updated for current GA release plans
- Java doc for ConfigurationProperties updated so it doesn't reference internal GeodeRedisServer
- Geode docs for Redis API for Geode updated


geode-core/src/main/java/org/apache/geode/distributed/ConfigurationProperties.java
```diff
@@ -1916,11 +1916,10 @@
   /**
    * The static String definition of the <i>"redis-port"</i> property <a name="redis-port"/a>
    * </p>
-   * <U>Description</U>: Specifies the port used by {@code GeodeRedisServer} which enables redis
-   * clients to connect and store data in GemFire distributed system. see {@code GeodeRedisServer}
-   * for other configuration options. A value of "-1" will select a random port.
+   * <U>Description</U>: Specifies the port used by Redis API for Geode which enables redis
+   * clients to connect and store data in GemFire distributed system.
```
dschneider-pivotal, line 1920: I think you should specify here what special values do. We had some of this text before which may have been lost in a merge. For example:
A value of "0" disables the Redis API for Geode.
A value of "-1" will select a random available port.

geode-redis/src/main/java/org/apache/geode/redis/internal/GeodeRedisServer.java
```diff
@@ -93,65 +93,11 @@
...
- * number set compared to overridden.</li>
- * <li>Transactions work just as they would on a Redis instance, they are local transaction.
- * Transactions cannot be executed on data that is not local to the executing server, that is on a
- * partitioned region in a different server instance or on a persistent region that does not have
- * transactions enabled. Also, you cannot watch or unwatch keys as all keys within a GemFire
- * transaction are watched by default.</li>
- * </ul>
+ *
+ * @See <a href=
+ *      "https://geode.apache.org/docs/guide/12/tools_modules/redis_api_for_geode.html">Redis API
```
dschneider-pivotal, line 98: I'm not sure I like this. It will grow stale over time since it is a reference to a specific version of the docs. And it is not really needed since this class is now internal. I'd vote for just dropping this reference.

sabbeyPivotal: Sounds good, I'll remove it.

geode-core/src/main/java/org/apache/geode/distributed/ConfigurationProperties.java
sabbeyPivotal: good catch, I'll update that.

pivotal-jbarrett: These selections are consistent with network services though. Zero typically mean automatically pick an ephemeral port. Given that ports should not be negative, can we use another property or API to enable service, or the use of `null`.

jdeppe-pivotal: @pivotal-jbarrett That's true. `-1` was chosen because `0` is already used to indicate that we don't want the service to be enabled. However we could (should?) switch that notion now while things are still `Experimental`.

sabbeyPivotal: Created a Jira for it: https://issues.apache.org/jira/browse/GEODE-8139
Will either do it now or add a story in the backlog

geode-docs/tools_modules/redis_api_for_geode.html.md.erb
```diff
@@ -0,0 +1,87 @@
...
+
+``` pre
+start server --name=<serverName> --locators=<locatorPort> --redis-port=<redisPort> --redis-bind-address=<redisBindAddress>
+```
+
+Replace `<serverName>` with the name of your server.
+
+Replace `<locatorPort>` with your locator port.
+
+Replace `<redisPort>` to specify the port the <%=vars.product_name%> server will listen on for Redis commands (the typical default Redis port is 6379).
```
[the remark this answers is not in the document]
sabbeyPivotal: Thanks for the suggestion Karen. I just updated it!

PR #5124, opened 2020-05-15
GEODE-7678 (2nd PR) - Support for cache-listener and client-notification for Partitioned Region Clear operation 

agingade:
The changes are made to PR clear messaging and locking mechanism to preserve
cache-listener and client-events ordering during concurrent cache operation
while clear in progress.
This also handles any partial clear due to member departure, by throwing PartitionedRegionPartialClearException.

### For all changes:

- [N] Is your initial contribution a single, squashed commit?

### Note:


geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionClearMessage.java
```diff
@@ -0,0 +1,289 @@
...
+
+  private Object cbArg;
+
+  private OperationType op;
+
+  private EventID eventID;
+
+  private PartitionedRegion partitionedRegion;
+
+  private Set<InternalDistributedMember> recipients;
```
jchen21, line 60: This seems to be duplicate. The super class already has an `ArrayList` of recipients.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionClearMessage.java
```diff
@@ -0,0 +1,289 @@
...
+    @Override
+    public void toData(DataOutput out,
+        SerializationContext context) throws IOException {
+      super.toData(out, context);
+      out.writeByte(op.ordinal());
+      DataSerializer.writeArrayList(bucketsCleared, out);
+    }
+
+    @Override
+    public String toString() {
```
jchen21, line 280: The implementation can use `super.toString()` and append fields specific to this class.

geode-core/src/test/java/org/apache/geode/internal/cache/PartitionedRegionTest.java
```diff
@@ -221,22 +220,6 @@ public void clearShouldNotThrowUnsupportedOperationException() {
     spyPartitionedRegion.clear();
   }
 
-  @Test(expected = CacheClosedException.class)
-  public void clearShouldThrowCacheClosedExceptionIfShutdownAll() {
```
jchen21, line 225: Is the deleted test case covered else where? Or it no longer valid?

geode-core/src/test/java/org/apache/geode/internal/cache/PartitionedRegionTest.java
```diff
@@ -249,28 +232,6 @@ public void createClearPRMessagesShouldCreateMessagePerBucket() {
     assertThat(msgs.size()).isEqualTo(3);
   }
 
-  @Test
-  public void sendEachMessagePerBucket() {
```
jchen21, line 253: Is the deleted test case covered else where? Or it no longer valid?

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionClearMessage.java
```diff
@@ -0,0 +1,289 @@
...
+    }
+    return null;
+  }
+
+  @Override
+  protected boolean operateOnPartitionedRegion(ClusterDistributionManager dm,
+      PartitionedRegion partitionedRegion,
+      long startTime) throws CacheException {
+
+    if (partitionedRegion == null) {
```
jchen21, line 110: This method never returns `false`. It always returns `true`. So there is no need to do the `if` check.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionClearMessage.java
```diff
@@ -0,0 +1,289 @@
...
+      return true;
+    }
+
+    if (partitionedRegion.isDestroyed()) {
```
jchen21, line 114: This method never returns `false`. It always returns `true`. So there is no need to do the `if` check.

geode-core/src/distributedTest/java/org/apache/geode/cache/PRCacheListenerDistributedTest.java
```diff
@@ -38,28 +45,60 @@
...
 public class PRCacheListenerDistributedTest extends ReplicateCacheListenerDistributedTest {
 
-  @Parameters(name = "{index}: redundancy={0}")
-  public static Iterable<Integer> data() {
-    return Arrays.asList(0, 3);
+  @Parameters
+  public static Collection<Object[]> data() {
+    return Arrays.asList(new Object[][] {
+        {1, Boolean.FALSE},
+        // {3, Boolean.TRUE},
```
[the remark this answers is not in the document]
agingade: Good catch. Have uncommented the test param.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionClearMessage.java
agingade: Thats not true; if the region is not destroyed, it returns false. line#1892 in LocalRegion.java.

geode-core/src/test/java/org/apache/geode/internal/cache/PartitionedRegionTest.java
agingade: This is testing the old way of messaging; where message is sent per bucket. That is not true anymore.

geode-core/src/test/java/org/apache/geode/internal/cache/PartitionedRegionTest.java
agingade: This is based on the earlier clear messaging. Thats not valid anymore.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionClearMessage.java
agingade: Thats not right; the class extending should be overriding the method and call super if any super fields needs to be included. If you look into other inheriting classes they have implemented their own toString().

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearWithExpirationDUnitTest.java
```diff
@@ -408,22 +394,21 @@ public void clearShouldFailWhenCoordinatorMemberIsBouncedAndExpirationTasksShoul
   @TestCaseName("[{index}] {method}(Coordinator:{0}, RegionType:{1})")
   public void clearShouldSucceedAndRemoveRegisteredExpirationTasksWhenNonCoordinatorMemberIsBounced(
       TestVM coordinatorVM, RegionShortcut regionShortcut) {
-    final int entries = 1500;
+    final int entries = 500;
+
+    RegionShortcut rs = regionShortcut;
```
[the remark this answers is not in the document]
agingade: Fixed the way it was getting used.

geode-core/src/main/java/org/apache/geode/internal/cache/BucketRegion.java
```diff
@@ -575,16 +575,23 @@ public void cmnClearRegion(RegionEventImpl regionEvent, boolean cacheWrite, bool
...
+      }
       // no need to dominate my own rvv.
       // Clear is on going here, there won't be GII for this member
       clearRegionLocally(regionEvent, cacheWrite, null);
       distributeClearOperation(regionEvent, rvv, participants);
 
       // TODO: call reindexUserDataRegion if there're lucene indexes
     } finally {
-      releaseWriteLocksForClear(regionEvent, participants);
+      if (!isLockedAlready) {
```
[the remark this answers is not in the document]
agingade: If you look into the line where the lock is taken its based on the "if(!isLockedAlready)" the same logic is used to unlock the lock.

geode-core/src/main/java/org/apache/geode/cache/Region.java
```diff
@@ -1302,7 +1302,10 @@ Object selectValue(String queryPredicate) throws FunctionDomainException, TypeMi
    * @see java.util.Map#clear()
    * @see CacheListener#afterRegionClear
    * @see CacheWriter#beforeRegionClear
-   * @throws UnsupportedOperationException If the region is a partitioned region
+   * @throws PartitionedRegionPartialClearException when data is partially cleared on partitioned
+   *         region. Its caller responsibility to handle the partial data clear either by retrying
```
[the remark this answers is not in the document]
agingade: Fixed it.

geode-core/src/main/java/org/apache/geode/internal/cache/BucketRegion.java
```diff
@@ -2493,4 +2500,11 @@ void updateSenderIdMonitor() {
   void checkSameSenderIdsAvailableOnAllNodes() {
     // nothing needed on a bucket region
   }
+
+  @Override
+  protected void basicClear(RegionEventImpl regionEvent) {
+    getDataView().checkSupportsRegionClear();
```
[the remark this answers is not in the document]
agingade: For bucket region, we don't need it. Have removed it.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
```diff
@@ -2148,69 +2150,71 @@ public void writeToDisk() {
...
-      regionEvent.setEventType(EnumListenerEvent.AFTER_REGION_CLEAR);
-      boolean hasListener = hasListener();
-      if (hasListener) {
-        dispatchListenerEvent(EnumListenerEvent.AFTER_REGION_CLEAR, regionEvent);
-      }
-      notifyBridgeClients(regionEvent);
-      logger.info("Partitioned region {} finsihed clear operation.", this.getFullPath());
-    }
-  }
+  /* @Override */
```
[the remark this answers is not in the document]
agingade: removed.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionClearMessage.java
jchen21: This `toString()` actually is missing the `sender` field. 

The `super.toString()` already contain the class name, `procesorId` and `exception`. Only `bucketsCleared` has to be added here.  I was trying to say something like:
```
    @Override
    public String toString() {
      StringBuffer sb = new StringBuffer();
      sb.append(super.toString());
      sb.append(" bucketsCleared ").append(this.bucketsCleared);
      return sb.toString();
    }
```

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionClearMessage.java
```diff
@@ -0,0 +1,286 @@
...
+        SerializationContext context) throws IOException {
+      super.toData(out, context);
+      out.writeByte(op.ordinal());
+      DataSerializer.writeArrayList(bucketsCleared, out);
+    }
+
+    @Override
+    public String toString() {
+      StringBuffer sb = new StringBuffer();
+      sb.append("PartitionedRegionClearReplyMessage ").append("processorid=")
```
jchen21, line 279: The `sender` field is missing.

agingade: Fixed.

PR #5125, opened 2020-05-16
GEODE-8170: Change all hash and set commands to use function+delta

dschneider-pivotal:
In addition to the hash commands this branch also changes smove to use functions.

### For all changes:

### Note:

### DIFF
--- a/geode-redis/src/main/java/org/apache/geode/redis/internal/executor/CommandFunction.java
+++ b/geode-redis/src/main/java/org/apache/geode/redis/internal/executor/CommandFunction.java
@@ -76,6 +76,7 @@ public class CommandFunction extends SingleResultRedisFunction {
   protected Object compute(Region localRegion, ByteArrayWrapper key,
       RedisCommandType command, Object[] args) {
     Callable<Object> callable;
+    boolean useStripedExecutor = true;
     switch (command) {
       case SADD: {
         ArrayList<ByteArrayWrapper> membersToAdd = (ArrayList<ByteArrayWrapper>) args[1];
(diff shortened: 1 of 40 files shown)


geode-redis/src/main/java/org/apache/geode/redis/internal/executor/EmptyRedisHash.java
```diff
@@ -26,41 +25,26 @@
...
-  @Override
-  public synchronized int hdel(Region<ByteArrayWrapper, RedisData> region, ByteArrayWrapper key,
-      List<ByteArrayWrapper> fieldsToRemove) {
-    return 0;
-  }
-
-  @Override
-  public synchronized Collection<ByteArrayWrapper> hgetall() {
-    return emptyList();
+  public EmptyRedisHash() {
```
jdeppe-pivotal, line 28: Do we still need this class? I'm not sure I fully understand it's usage since when I switched `RedisHash.EMPTY` to `new RedisHash(Collections.emptyList())`, `HashesIntegrationTest` still passes.

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/CommandFunction.java
```diff
@@ -117,6 +118,32 @@ protected Object compute(Region localRegion, ByteArrayWrapper key,
             () -> new RedisSetInRegion(localRegion).sscan(key, matchPattern, count, cursor);
         break;
       }
+      case SCREATE: {
```
jdeppe-pivotal, line 121: How is this used / invoked?

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/EmptyRedisHash.java
dschneider-pivotal: It is just an optimization. It gives us a single instance of RedisHash that represents an empty hash instead of having to create a new instance every time we want to represent an empty hash. We do this alot because we have a common pattern of not wanting to return null when we do the initial lookup of a RedisHash from the region.

dschneider-pivotal: We do have to be careful that our canonical instance is immutable

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/CommandFunction.java
dschneider-pivotal: I think the place I was using it was refactored and now it is not used. I'll remove it.

PR #5128, opened 2020-05-18
GEODE-8147: DELETE and EXISTS implement Function-Delta

prettyClouds:

### For all changes:

### Note:

### DIFF
--- a/geode-redis/src/main/java/org/apache/geode/redis/internal/RegionProvider.java
+++ b/geode-redis/src/main/java/org/apache/geode/redis/internal/RegionProvider.java
@@ -225,15 +225,7 @@ public class RegionProvider implements Closeable {
           return hLLRegion.remove(key) != null;
         } else if (type == RedisDataType.REDIS_LIST || type == RedisDataType.REDIS_SORTEDSET) {
           return destroyRegion(key, type);
-        } else if (type == RedisDataType.REDIS_SET) {
-          RedisSetCommands redisSetCommands =
-              new RedisSetCommandsFunctionExecutor(dataRegion);
-          return redisSetCommands.del(key);
-        } else if (type == RedisDataType.REDIS_HASH) {
-          RedisHashCommands redisHashCommands =
-              new RedisHashCommandsFunctionExecutor(dataRegion);
-          return redisHashCommands.del(key);
-        } else {
+        }else {
           return false;
         }
       } catch (Exception exc) {
--- a/geode-redis/src/main/java/org/apache/geode/redis/internal/RegionProvider.java
+++ b/geode-redis/src/main/java/org/apache/geode/redis/internal/RegionProvider.java
@@ -47,6 +47,10 @@ import org.apache.geode.management.internal.cli.result.model.ResultModel;
 import org.apache.geode.redis.internal.executor.ExpirationExecutor;
 import org.apache.geode.redis.internal.executor.ListQuery;
 import org.apache.geode.redis.internal.executor.SortedSetQuery;
+import org.apache.geode.redis.internal.executor.hash.RedisHashCommands;
+import org.apache.geode.redis.internal.executor.hash.RedisHashCommandsFunctionExecutor;
+import org.apache.geode.redis.internal.executor.set.RedisSetCommands;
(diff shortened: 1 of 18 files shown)


geode-redis/src/main/java/org/apache/geode/redis/internal/executor/RedisDataInRegion.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+package org.apache.geode.redis.internal.executor;
+
+import org.apache.geode.cache.Region;
+import org.apache.geode.redis.internal.ByteArrayWrapper;
+import org.apache.geode.redis.internal.RedisData;
+import org.apache.geode.redis.internal.RegionProvider;
+
+class RedisDataInRegion {
```
dschneider-pivotal, line 23: Think about making this classes name consistent with RedisKeyCommands. Should they both be "Key" or both be "Data"?

geode-redis/src/main/java/org/apache/geode/redis/internal/RegionProvider.java
```diff
@@ -218,21 +216,16 @@ public boolean removeKey(ByteArrayWrapper key, RedisDataType type, boolean cance
       if (!typeStoresDataInKeyRegistrar(type)) {
         keyRegistrar.unregister(key);
       }
+      RedisKeyCommands redisKeyCommands = new RedisKeyCommandsFunctionExecutor(dataRegion);
       try {
         if (type == RedisDataType.REDIS_STRING) {
```
dschneider-pivotal, line 221: consider converting this if/elseif chain to a switch statement

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/RedisKeyCommandsFunctionExecutor.java
```diff
@@ -0,0 +1,40 @@
...
+  private Region<ByteArrayWrapper, RedisData> region;
+
+  public RedisKeyCommandsFunctionExecutor(
+      Region<ByteArrayWrapper, RedisData> region) {
+    this.region = region;
+  }
+
+  @Override
+  public boolean del(ByteArrayWrapper key) {
+    return (boolean) CommandFunction.execute(RedisCommandType.DEL, key, new Object[] {}, region);
```
dschneider-pivotal, line 33: Something I find slightly confusing on CommandFunction is that is has a static method named execute. When looking at a Function I tend to look for its "instance" method named execute. Something we could consider is having all our *FunctionExecutor classes, subclass a common class that has the code from the static CommandFunction execute. I think that could also simplify how much we need to pass to this method (we would no longer need to pass the region since we could access it from the instance). Let me know what you think. No need to do this on this PR

prettyClouds: yea, the static execute method looks a little weird to me as well.
I like the idea of having a simple interface other than the FunctionService to expose the function call, but the name `execute` is definitely confusing.
Some kind of code sharing in the *FunctionExecutor classes is a good idea.

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/ExistsExecutor.java
```diff
@@ -20,22 +20,19 @@
...
-      return;
-    }
+    RedisKeyCommands redisKeyCommands =
+        new RedisKeyCommandsFunctionExecutor(context.getRegionProvider().getDataRegion());
 
     long existsCount = commandElems
         .subList(1, commandElems.size())
         .stream()
-        .filter(key -> context.getKeyRegistrar().isRegistered(key))
+        .filter(key -> redisKeyCommands.exists(key))
```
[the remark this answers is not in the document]
prettyClouds: lol...that would have been very passive aggressive

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/RedisDataInRegion.java
prettyClouds: made it RedisKeyInRegion

PR #5131, opened 2020-05-19
GEODE-8144: endpoint identification in servers is not working

bschuchardt:
Set the SNI server-name field in SSL parameters for p2p communications,
allowing endpoint identification to work properly.

I modified one of the SNI haproxy tests to have keystores with the
proper subject-alternative-names for p2p communications in the docker
containers and for client/server off-platform communications and enabled
endpoint identification in the test.  I used Sai's keystore/truststore construction
CertificateMaterial/CertStores classes to generate the stores...

      .sanDnsName("geode") // for inside the docker container
      .sanDnsName("localhost") // for inside the docker container
      .sanIpAddress(InetAddress.getByName("0.0.0.0")) // for inside the docker container
      .sanDnsName(certName) // for client endpoint validation (locator-maeve for instance)

### For all changes:

### Note:


Bill:
Bruce's idea yesterday, of eliminating the numeric IP referencing the local locator (up) in locator launcher seems like a good one. Jake thought that the `--bind-address` arg to `gfsh locator start` should do the trick: if that's a name (not an IP number) it should be passed right through as the way to reach the local locator.

I spent a little time experimenting with that this morning to try and find exactly where that reference to the local locator was being generated, and how to fix it. Here's a test I've been experimenting with (added to `LocatorLauncherIntegrationTest`):

```
 @Test
  public void symbolicBindAddressUsedToStartLocator() throws Exception {
    // given: a new builder
    Builder builder = new Builder();

    // when: parsing a symbolic (not-an-IP-number) bind address
    builder.parseArguments("start", "--bind-address", "foo");

    assertThat(builder.getBindAddress()).isEqualTo(InetAddress.getByName("foo"));

    final LocatorLauncher launcher = builder.build();
    final Properties dsps = launcher.getDistributedSystemProperties();

    // fails: properties has only one property: "name"="foo" ugh
    //assertThat(dsps).containsKey(BIND_ADDRESS);

    final DistributionConfigImpl distributionConfig = new DistributionConfigImpl(dsps);

    // then: no numeric bind addy's in start command args
    // fails: no start locator!
    //assertThat(distributionConfig.getStartLocator()).isEqualTo("foo");

    // fails: no locators at all
    assertThat(distributionConfig.getLocators()).isEqualTo("foo");
  }
```

It'll take some more chasing to find it.

bschuchardt:
The locator setting is inserted in `InternalLocator.startDistributedSystem()`

```
private void startDistributedSystem() throws IOException {
    InternalDistributedSystem existing = InternalDistributedSystem.getConnectedInstance();
    if (existing != null) {
      // LOG: changed from config to info
      logger.info("Using existing distributed system: {}", existing);
      startCache(existing);
    } else {
      StringBuilder sb = new StringBuilder(100);
      if (bindAddress != null) {
        sb.append(bindAddress.getHostAddress()); <<< HERE
      } else {
        sb.append(LocalHostUtil.getLocalHost().getHostAddress());  <<< and HERE
      }
      sb.append('[').append(getPort()).append(']');
      String thisLocator = sb.toString();
```

geode-core/src/main/java/org/apache/geode/internal/net/SocketCreator.java
```diff
@@ -791,7 +792,19 @@ private boolean setServerNames(SSLParameters modifiedParams, HostAndPort addr) {
...
 
-    serverNames.add(new SNIHostName(addr.getHostName()));
+    String hostName = addr.getHostName();
+    if (this.sslConfig.doEndpointIdentification()
+        && InetAddressValidator.getInstance().isValid(hostName)) {
+      // endpoint validation typically uses a hostname in the sniServer parameter that the handshake
+      // will compare against the subject alternative addresses in the server's certificate. Here
+      // we attempt to get a hostname instead of the proffered numeric address
+      try {
+        hostName = InetAddress.getByName(hostName).getCanonicalHostName();
```
pivotal-jbarrett, line 802: This strikes me as a security issue. Let's say I issued a start server with `locators=1.2.3.4`, which is bad but stick with me. Then here we lookup 1.2.3.4's PTR record. What if someone has maliciously inserted themselves into my DNS resolution path, easy to do with out greater adoption of DNSSEC. So my reverse lookup finds `1.2.3.4 PTR locator1.malicious.com` and adds "locator1.malicious.com" to the SNI and proceeds to connect. Then connection is established to what our network stack thinks is 1.2.3.4, but is really our malicious router redirecting the connection. Now the malicious host will respond with a valid signed certificate for "locator1.malicious.com" and we will happily verify it. Of course that assumes also that the malicious certificate is signed by someone in our chain of trust. So, not a likely attack vector since we don't trust the root signers by default but not something a service that does trust the roots should ever do.

bschuchardt: @pivotal-jbarrett if you will look at the implementation of getCanonicalHostName, I think you will find that it already addresses your concerns.

pivotal-jbarrett: As you mentioned offline, the same malicious entity could inject the IP into their SAN and we would validate that. I don't think this code makes anything any less secure from that standpoint so I am removing my block.

PR #5132, opened 2020-05-19
GEODE-8131: reader thread blocked attempting to issue an alert

bschuchardt:
This PR does not solve the problem of the alert system causing a P2P
message reader to block but instead ensures that the reader thread will
write deserialization information to the log before issuing a fatal
alert.  If the alert level is set to "info" this won't help, but no-one
would set the alert level to that level.

Along the way I removed some duplicate strings.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/tcp/Connection.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/tcp/Connection.java
@@ -3027,17 +3027,13 @@ public class Connection implements Runnable {
         failureMsg = "ClassNotFound deserializing message";
         failureEx = ex;
         rpId = md.getRPid();
-        // log at info level first in case fatal-level alert notification becomes blocked
-        logger.info(failureMsg, failureEx);
-        logger.fatal(failureMsg, failureEx.toString());
+        logAtInfoAndFatal(failureMsg, failureEx);
       } catch (IOException ex) {
         owner.getConduit().getStats().decMessagesBeingReceived(md.size());
         failureMsg = "IOException deserializing message";
         failureEx = ex;
         rpId = md.getRPid();
-        // log at info level first in case fatal-level alert notification becomes blocked
-        logger.info(failureMsg, failureEx);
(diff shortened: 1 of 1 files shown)


geode-core/src/main/java/org/apache/geode/internal/tcp/Connection.java
```diff
@@ -3054,7 +3058,9 @@ private void readMessage(ByteBuffer peerDataBuffer) {
         failureMsg = "Unexpected failure deserializing message";
         failureEx = ex;
         rpId = md.getRPid();
-        logger.fatal("Unexpected failure deserializing message", failureEx);
+        // log at info level first in case fatal-level alert notification becomes blocked
+        logger.info(failureMsg, failureEx);
+        logger.fatal(failureMsg, failureEx);
```
Bill, line 3063: Instead of repeating this code in three places, comment-and-all, it would be better to have the three sites call a new method e.g. `logFatal(…)`. The method could have a block comment telling about what it's contract is etc.

geode-core/src/main/java/org/apache/geode/internal/tcp/Connection.java
```diff
@@ -3027,13 +3027,17 @@ private void readMessage(ByteBuffer peerDataBuffer) {
         failureMsg = "ClassNotFound deserializing message";
         failureEx = ex;
         rpId = md.getRPid();
-        logger.fatal("ClassNotFound deserializing message: {}", ex.toString());
+        // log at info level first in case fatal-level alert notification becomes blocked
+        logger.info(failureMsg, failureEx);
+        logger.fatal(failureMsg, failureEx.toString());
```
Bill, line 3032: why is `toString()` called explicitly in this case but not in the other two?

bschuchardt: Beats me why the toString is there - I didn't add it.  I do want to see the full call stack when triaging a problem like this.  Maybe it has to do with alert listeners wanting one-liners?

PR #5130, opened 2020-05-19
GEODE-8145: Add Redis configuration properties to gemfire_properties.html

sabbeyPivotal:
redis-password, redis-bind-address, and redis-port need to be added to this page

PR #5142, opened 2020-05-20
GEODE-8138: Improve semantics of the redis-port option

sabbeyPivotal:
- Introduce a Geode parameter - `redis-enabled` which is `false` by default. This parameter determines whether the service is started or not.
- Set the default `redis-port` value to `6379`
- `redis-port` range is 0..65535
- When `redis-port=0`, and `redis-enabled=true` the service will start on an ephemeral port.
- When `gfsh` is used with `--redis-port`, `--redis-password`, or `--redis-bind-address`, it should automatically set `redis-enabled=true`

`gemfire_properties.html.md.erb` updated as part of this PR: https://github.com/apache/geode/pull/5130

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/distributed/ConfigurationProperties.java
+++ b/geode-core/src/main/java/org/apache/geode/distributed/ConfigurationProperties.java
@@ -1914,17 +1914,6 @@ public interface ConfigurationProperties {
    */
   String OFF_HEAP_MEMORY_SIZE = "off-heap-memory-size";
   /**
-   * The static String definition of the <i>"redis-port"</i> property <a name="redis-port"/a>
-   * </p>
-   * <U>Description</U>: Specifies the port on which the server listens for Redis API for Geode
-   * connections. A value of 0 selects a random port.</td>
-   * </p>
-   * <U>Default</U>: 6379
-   * </p>
-   * <U>Allowed values</U>: 0..65535
-   */
-  String REDIS_PORT = "redis-port";
-  /**
    * The static String definition of the <i>"redis-bind-address"</i> property <a
    * name="redis-bind-address"/a>
    * </p>
@@ -1935,16 +1924,6 @@ public interface ConfigurationProperties {
    * <U>Default</U>: ""
    */
   String REDIS_BIND_ADDRESS = "redis-bind-address";
-  /**
-   * The static String definition of the <i>"redis-password"</i> property <a
-   * name="redis-password"/a>
-   * </p>
-   * <U>Description</U>: Specifies the password that the server uses when a client attempts to
-   * authenticate.
-   * </p>
-   * <U>Default</U>: no password set
-   */
-  String REDIS_PASSWORD = "redis-password";
   /**
    * The static String definition of the <i>"redis-enabled"</i> property <a
    * name="redis-enabled"/a>
@@ -1963,6 +1942,27 @@ public interface ConfigurationProperties {
    * </p>
    */
   String REDIS_ENABLED = "redis-enabled";
+  /**
+   * The static String definition of the <i>"redis-password"</i> property <a
+   * name="redis-password"/a>
+   * </p>
+   * <U>Description</U>: Specifies the password that the server uses when a client attempts to
+   * authenticate.
+   * </p>
+   * <U>Default</U>: no password set
+   */
+  String REDIS_PASSWORD = "redis-password";
+  /**
+   * The static String definition of the <i>"redis-port"</i> property <a name="redis-port"/a>
+   * </p>
+   * <U>Description</U>: Specifies the port on which the server listens for Redis API for Geode
+   * connections. A value of 0 selects a random port.</td>
+   * </p>
+   * <U>Default</U>: 6379
+   * </p>
+   * <U>Allowed values</U>: 0..65535
+   */
+  String REDIS_PORT = "redis-port";
   /**
    * The static String definition of the <i>"lock-memory"</i> property <a name="lock-memory"/a>
    * </p>
--- a/geode-core/src/main/java/org/apache/geode/distributed/ConfigurationProperties.java
+++ b/geode-core/src/main/java/org/apache/geode/distributed/ConfigurationProperties.java
@@ -1946,6 +1946,24 @@ public interface ConfigurationProperties {
    */
   String REDIS_PASSWORD = "redis-password";
   /**
+   * The static String definition of the <i>"redis-enabled"</i> property <a
+   * name="redis-enabled"/a>
+   * </p>
+   * <U>Description</U>: When the default value of false, the Redis API for Geode is not available.
+   * Set to true to enable the Redis API for Geode.</td>
+   * </p>
+   * <U>Default</U>: false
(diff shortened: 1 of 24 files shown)


geode-core/src/main/java/org/apache/geode/distributed/internal/AbstractDistributionConfig.java
```diff
@@ -1312,11 +1313,13 @@ static Class _getAttributeType(String attName) {
     m.put(MEMCACHED_BIND_ADDRESS,
         "The address the GemFireMemcachedServer will listen on for remote connections. Default is \"\" which causes the GemFireMemcachedServer to listen on the host's default address. This property is ignored if memcached-port is \"0\".");
     m.put(REDIS_PORT,
-        "The port GeodeRedisServer will listen on. Default is 0. Set to zero to disable GeodeRedisServer.");
+        "The port Redis API for Geode will listen on. Default is 0");
```
dschneider-pivotal, line 1316: Is the default now 6379? Also I think you want a "." at the end of the sentence.

geode-core/src/main/java/org/apache/geode/management/internal/i18n/CliStrings.java
```diff
@@ -2557,14 +2557,14 @@
       "The gemfire.properties file for configuring the Cache Server's distributed system. The file's path can be absolute or relative to the gfsh working directory.";
   public static final String START_SERVER__REDIS_PORT = ConfigurationProperties.REDIS_PORT;
   public static final String START_SERVER__REDIS_PORT__HELP =
-      "Sets the port that the Geode Redis service listens on for Redis clients.";
+      "Sets the port that the Redis API for Geode server listens on for Redis clients. The default is an ephemeral port";
```
dschneider-pivotal, line 2560: For gfsh start server I thought the default would be 6379

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/StartServerCommand.java
```diff
@@ -301,6 +301,13 @@ ResultModel doStartServer(String memberName, Boolean assignBuckets, String bindA
         ConfigurationProperties.HTTP_SERVICE_PORT, httpServicePort);
     StartMemberUtils.setPropertyIfNotNull(gemfireProperties,
         ConfigurationProperties.HTTP_SERVICE_BIND_ADDRESS, httpServiceBindAddress);
+
+    // if redis-port or redis-bind-address are specified in the command line,
+    // REDIS_SERVICE_ENABLED should be set to true
+    if (StringUtils.isNotBlank(redisPassword) || StringUtils.isNotBlank(redisBindAddress)) {
```
dschneider-pivotal, line 307: I see this checking the password and bindAddress setting. But shouldn't it also be checking the redisPort? The comment says port and bind-address. Perhaps any redis option (including password) should enable redis.

geode-redis/src/main/java/org/apache/geode/redis/internal/GeodeRedisServer.java
```diff
@@ -241,7 +241,7 @@ private int setNumWorkerThreads() {
    * to the first non-loopback address
    */
   public GeodeRedisServer() {
-    this(null, -1, null);
+    this(null, 0, null);
```
dschneider-pivotal, line 244: Should this use RANDOM_PORT_INDICATOR instead of 0?

geode-redis/src/main/java/org/apache/geode/redis/internal/GeodeRedisServer.java
```diff
@@ -275,7 +275,8 @@ public GeodeRedisServer(String bindAddress, int port) {
    *
    * @param bindAddress The address to which the server will attempt to bind to
    * @param port The port the server will bind to, will use {@value #DEFAULT_REDIS_SERVER_PORT}
-   *        by default if argument is less than -1. If the port is {@value #RANDOM_PORT_INDICATOR}
+   *        by default if argument is less than {@value #RANDOM_PORT_INDICATOR}. If the port is
```
dschneider-pivotal, line 278: Is it correct that if it is < RANDOM_PORT_INDICATOR it will use the default? I couldn't find the code that made that happen but it seems odd if I say -33456 it will listen on 6379. I think I'd prefer an IllegalArgumentException

geode-core/src/main/java/org/apache/geode/distributed/internal/AbstractDistributionConfig.java
sabbeyPivotal: Yes. Thank you for finding this!

geode-core/src/main/java/org/apache/geode/management/internal/i18n/CliStrings.java
sabbeyPivotal: Yes!

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/StartServerCommand.java
sabbeyPivotal: You are right. What do you think? Maybe it would make sense to enable it if any redis option is specified?

geode-redis/src/main/java/org/apache/geode/redis/internal/GeodeRedisServer.java
sabbeyPivotal: Definitely.

geode-redis/src/main/java/org/apache/geode/redis/internal/GeodeRedisServer.java
sabbeyPivotal: That is currently what happens. I like what you're suggesting though.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/StartServerCommand.java
sabbeyPivotal: Updated it!

geode-core/src/main/java/org/apache/geode/distributed/ConfigurationProperties.java
```diff
@@ -1943,6 +1931,36 @@
...
+   * <td>redis-service-enabled</td>
+   * <td>When the default value of false, the Redis API for <%=vars.product_name%> is not available.
+   * Set
+   * to true to enable the Redis API for <%=vars.product_name%>.</td>
+   * <td>S</td>
+   * <td>false</td>
+   * </tr>
+   * </p>
+   */
+  String REDIS_SERVICE_ENABLED = "redis-service-enabled";
```
pivotal-jbarrett, line 1963: Why add in the word "service" here? Why not just "redis-enabled". None of the other redis properties use the word "service". I think consistent naming conventions are really important to avoid confusion.

sabbeyPivotal: Yeah, we can make it `redis-enabled`

PR #5146, opened 2020-05-22
GEODE-8175: removed list, sorted set, hyperLog and transaction redis commands and code

dschneider-pivotal:

### For all changes:

### Note:


geode-redis/src/distributedTest/java/org/apache/geode/redis/RedisDistDUnitTest.java
```diff
@@ -209,18 +209,6 @@ public void run() throws InterruptedException {
             } else {
               jedis.del(hKey);
             }
-          } else if (n == 1) {
```
ringles, line 212: Can we get rid of the @Ignore on this test now that we're only testing hashes and sets?

geode-redis/src/main/java/org/apache/geode/redis/internal/KeyRegistrar.java
```diff
@@ -201,11 +172,7 @@ private void throwDataTypeException(ByteArrayWrapper key, RedisData data) {
   }
 
   private void throwDataTypeException(ByteArrayWrapper key, RedisDataType dataType) {
-    if (RedisDataType.REDIS_PROTECTED.equals(dataType)) {
-      throw new RedisDataTypeMismatchException("The key name \"" + key + "\" is protected");
-    } else {
-      throw new RedisDataTypeMismatchException(
-          RedisConstants.ERROR_WRONG_TYPE);
-    }
+    throw new RedisDataTypeMismatchException(
```
ringles, line 175: Doesn't look like the arguments are needed anymore. In fact, the whole function may be eliminated...

geode-redis/src/main/java/org/apache/geode/redis/internal/RegionProvider.java
```diff
@@ -205,281 +98,27 @@ private boolean typeStoresDataInKeyRegistrar(RedisDataType type) {
...
-  private boolean destroyRegion(ByteArrayWrapper key, RedisDataType type) {
-    Region<?, ?> r = dynamicRegions.get(key);
-    if (r != null) {
-      try {
-        r.destroyRegion();
-      } catch (Exception e) {
-        return false;
-      } finally {
-        removeRegionState(key, type);
+      if (cancelExpiration) {
```
ringles, line 116: So, because we don't care whether or not these calls return true or false, we don't really need the lock anymore? If two threads try to remove the same key, at least one will succeed so no harm, no foul?

dschneider-pivotal: The locks on this method had to do with dynamic regions which no longer exist because they were only used by lists a zset. I do think we care about what we return from this and we end up being protected by the StripedExecutor now. I think once Strings have moved to the new data model this "del" code can be cleaned up some more. Also once everything is on the new expiration model then expiration cleanup this method does will not be needed.

geode-redis/src/distributedTest/java/org/apache/geode/redis/RedisDistDUnitTest.java
dschneider-pivotal: done

geode-redis/src/main/java/org/apache/geode/redis/internal/KeyRegistrar.java
dschneider-pivotal: done

PR #5149, opened 2020-05-22
GEODE-8108: Remove System.out.println calls from geode-redis

alb3rtobr:
Ticket description, reported by @dschneider-pivotal :
> 
> I noticed in SMoveExecutor a number of System.out.println calls. This should not be done. 
> Either remove the calls (that is what I recommend for SMoveExecutor) or use a logger. 
> I don't know if SMoveExecutor is the only place this happens so a code scan should be done.

`SMoveExecutor` class was already fixed but I searched for other calls to `System.out.println` in `geode-redis` and I have removed them.
I also updated the logger of `GeodeRedisServer` class.


dschneider-pivotal:
Thanks for this work. Can you resolve the conflicts in GeodeRedisServer?

PR #5155, opened 2020-05-23
Revert "GEODE-8175: removed list, sorted set, hyperLog and transaction redis commands and code"

onichols-pivotal:
Reverts apache/geode#5146 which caused compile error


onichols-pivotal:
easier to just fix, see #5156

PR #5159, opened 2020-05-25
GEODE-7956: correct legal region names

mkevo:
With existing regex we supported also: [, ], ^, `, \ in region name.
With a new regex we supported only characters described in the documentation.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/RegionNameValidation.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/RegionNameValidation.java
@@ -24,7 +24,7 @@ import org.apache.geode.cache.Region;
 
 public class RegionNameValidation {
 
-  private static final Pattern NAME_PATTERN = Pattern.compile("[aA-zZ0-9-_.]+");
+  private static final Pattern NAME_PATTERN = Pattern.compile("[a-zA-Z0-9-_.]+");
 
   static Pattern getNamePattern() {
     return NAME_PATTERN;
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/RegionNameValidation.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/RegionNameValidation.java
@@ -24,7 +24,7 @@ import org.apache.geode.cache.Region;
 
 public class RegionNameValidation {
 
-  private static final Pattern NAME_PATTERN = Pattern.compile("[a-zA-Z0-9-_.]+");
+  private static final Pattern NAME_PATTERN = Pattern.compile("[a-zA-Z\\[\\]0-9-_.]+");
 
   static Pattern getNamePattern() {
     return NAME_PATTERN;
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/RegionNameValidation.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/RegionNameValidation.java
@@ -24,7 +24,7 @@ import org.apache.geode.cache.Region;
 
 public class RegionNameValidation {
 
-  private static final Pattern NAME_PATTERN = Pattern.compile("[a-zA-Z\\[\\]0-9-_.]+");
+  private static final Pattern NAME_PATTERN = Pattern.compile("[a-zA-Z0-9-_.^`\\[\\]\\\\]+");
 
   static Pattern getNamePattern() {
     return NAME_PATTERN;
--- a/geode-docs/basic_config/data_regions/region_naming.html.md.erb
+++ b/geode-docs/basic_config/data_regions/region_naming.html.md.erb
@@ -24,7 +24,7 @@ follow these region naming guidelines.
 
 -   Characters permitted in region names are alphanumeric characters
 (`ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789`),
-period (`.`), underscore (`_`), square brackets (`[, ]`) and hyphen (`-`).
+period (`.`), underscore (`_`), square brackets (`[, ]`), hyphen (`-`), caret (`^`) and backqoute (```).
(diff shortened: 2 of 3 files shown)


geode-docs/basic_config/data_regions/region_naming.html.md.erb
```diff
@@ -24,7 +24,7 @@ follow these region naming guidelines.
 
 -   Permitted characters within region names are alphanumeric characters
 (`ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789`),
-the underscore character (`_`), and the hyphen character (`-`).
+the dot character (`.`), the underscore character (`_`), the square brackets characters (`[, ]`) and the hyphen character (`-`).
```
davebarnes97, line 27: Good clarification. An editorial suggestion: Since you've already established that you're talking about characters, you can simplify the language for readability without loss of precision.
Please change "the dot character" to "period", "the underscore character" to "underscore", "the square brackets characters" to "square brackets" and "the hyphen character" to "hyphen". Also for readability, "Permitted characters within region names" could be "Characters permitted in region names are".
The result would be:
Characters permitted in region names are alphanumeric characaters (`ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789`), period (`.`), underscore (`_`), square brackets (`[, ]`) and hyphen (`-`).

mkevo: Done! Thanks @davebarnes97!

geode-core/src/main/java/org/apache/geode/internal/cache/RegionNameValidation.java
```diff
@@ -24,7 +24,7 @@
 
 public class RegionNameValidation {
 
-  private static final Pattern NAME_PATTERN = Pattern.compile("[aA-zZ0-9-_.]+");
+  private static final Pattern NAME_PATTERN = Pattern.compile("[a-zA-Z\\[\\]0-9-_.]+");
```
dschneider-pivotal, line 27: I don't understand why you changed this Pattern. I think GEODE-7956 was just about correcting the docs. It looks to me like you are expanding what a legal region name is to include square brackets. Did this get approved in a RFC? If not I think you should do so before this gets approved. In we end up keeping square brackets in region names then update GEODE-7956 to say we should and why it is valuable.

dschneider-pivotal: I see your comment now about the old regex allowing extra characters. Can you enhance the RegionNameValidationTest to demonstrate that these were allowed before. I think it would be safest to just document all the characters already supported in a region name by the product and not change it. If someone already put one of those characters in their region name we done want to break them. I don't understand how the old regex let those sneak in. Did aA-zZ include ^, `, [, ], and backslash somehow?

mkevo: I didn't find RFC for this, but old pattern include this characters and you can see it if you add logger for names in test `startingWithMatchingCharactersAreOk` in RegionNameValidationTest where it validates all characters which is allowed by this pattern. 
From ASCII table A is 65 and z is 122, and this part **A-z** include all characters between these two values, and this characters(^, `, [, ], and backslash) are between 91 and 96.
Old pattern **[aA-zZ0-9-_.]+** is the same as **[A-z0-9-_.]+**.\
So we can go with change pattern to **[A-z0-9-_.]+** and be sure that if someone is using it we don't break them. And also document all charactes which is allowed.
Do you agree with that?

dschneider-pivotal: When I read aA-zZ I thought it was just a special regex pattern to match a-zA-Z. I bet whoever wrote this thought that. But since the product was already released with a pattern that allows those other characters it would be safest to continue to allow those characters. To make a breaking change I think we need an RFC. For now it is best to change the docs to describe the behavior of the existing geode release. We could schedule removing these special characters from region names in a future release. I think it would be good to rewrite the regex to still match what it did before but to do it more explicitly (something like a-zA-Z0-9-_.^`[]\\) or add comments to make clear what A-z matches. I also think it would be good to have some explicit test methods that show that these special characters are allowed in region names.
Thanks for figuring this out!

mkevo: Yes, I agree. I also think that aA-zZ is the same as a-zA-Z. 
The new regex will be **[a-zA-Z0-9-_.^`\[\]\\]+**. Is it ok with you?
This more clear to see what is included. I will also add separate tests for all of these characters.

dschneider-pivotal: Sounds perfect!

mkevo:
@davebarnes97, please review new document changes. Thanks!

geode-docs/basic_config/data_regions/region_naming.html.md.erb
```diff
@@ -24,7 +24,7 @@ follow these region naming guidelines.
 
 -   Characters permitted in region names are alphanumeric characters
 (`ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789`),
-period (`.`), underscore (`_`), square brackets (`[, ]`) and hyphen (`-`).
+period (`.`), underscore (`_`), square brackets (`[, ]`), hyphen (`-`), caret (`^`) and backqoute (```).
```
dschneider-pivotal, line 27: type: change qoute to quote

mkevo: Done. Thanks!

PR #5165, opened 2020-05-27
GEODE-8182: change String commands to use one region

dschneider-pivotal:

### For all changes:

### Note:


geode-redis/src/main/java/org/apache/geode/redis/internal/executor/string/SetNXExecutor.java
```diff
@@ -36,27 +36,23 @@
...
 
     ByteArrayWrapper key = command.getKey();
-    checkAndSetDataType(key, context);
-    byte[] value = commandElems.get(VALUE_INDEX);
+    ByteArrayWrapper value = new ByteArrayWrapper(commandElems.get(VALUE_INDEX));
 
-    Object oldValue =
-        region.putIfAbsent(key, (RedisData) new RedisString(new ByteArrayWrapper(value)));
+    RedisStringCommands stringCommands = getRedisStringCommands(context);
+    SetOptions setOptions = new SetOptions(NX, 0L, null, false);
```
jdeppe-pivotal, line 48: This should be `SetOptions(NX, 0L, ExpireUnit.NONE, false);`

dschneider-pivotal: Thanks for pointing this out. At this time, all that is used on the SetOptions when we call RedisStringCommands.set is the first parameter (NX in this case) so I just went with defaults on all the others. Some of these apply to expiration so they will be reviewed when the string expiration work is done.

PR #5167, opened 2020-05-27
[DO NOT REVIEW]: Debug 4 Regression Introduced Through GEODE-8119

jujoramos:
Hello @mkevo,
This is just a draft `PR` to reproduce the regression introduced through [GEODE-8119](https://issues.apache.org/jira/browse/GEODE-8119). The [first commit](https://github.com/apache/geode/pull/5167/commits/cbc5d9ce0a60b2ae929a016b398b5624a999c86b) contains only your original changes, the [second commit](https://github.com/apache/geode/pull/5167/commits/ae7a5817c39744b4b9b437b2cebbd1c602ff0ab2) is the distributed test showing the regression failure.
We can use this `PR` or the ticket for communication purposes, whatever you prefer.
Thanks in advance!.

### DIFF
--- a/geode-core/src/upgradeTest/java/org/apache/geode/internal/cache/Geode8119RegressionDistributedTest.java
+++ b/geode-core/src/upgradeTest/java/org/apache/geode/internal/cache/Geode8119RegressionDistributedTest.java
@@ -1,0 +1,170 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package org.apache.geode.internal.cache;
+
+import static org.apache.geode.distributed.ConfigurationProperties.LOCATORS;
+import static org.apache.geode.distributed.ConfigurationProperties.MCAST_PORT;
+import static org.apache.geode.distributed.ConfigurationProperties.START_LOCATOR;
+import static org.apache.geode.internal.AvailablePortHelper.getRandomAvailableTCPPortsForDUnitSite;
+import static org.apache.geode.test.awaitility.GeodeAwaitility.await;
+import static org.apache.geode.test.dunit.VM.getVM;
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.Serializable;
+import java.util.Arrays;
+import java.util.Properties;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+
+import org.apache.geode.cache.DiskStoreFactory;
+import org.apache.geode.cache.Region;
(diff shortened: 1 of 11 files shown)


mkevo:
I tried to run it and it failed with:
```
[Could not find: "/tmp/junit5454972723589376526/diskDir2/BACKUPtestDisk.if, /tmp/junit5454972723589376526/diskDir3/BACKUPtestDisk.if"
] expected:<[OK]> but was:<[ERROR]>
```

Also I tried to run it locally on the gfsh(without my changes) and see that if you create disk-store with multiple dirs added it will create this .if file only in first dir, and when doing validate it can't found it in other dirs.

geode-core/src/upgradeTest/java/org/apache/geode/internal/cache/Geode8119RegressionDistributedTest.java
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  @Test
+  public void regressionTest() throws IOException {
+    VM locator, server;
+    final int ENTRIES = 100000;
+    final Host host = Host.getHost(0);
+    int site1Port = getRandomAvailableTCPPortsForDUnitSite(1)[0];
+    File diskStoreDirectory1 = temporaryFolder.newFolder("diskDir1");
+    File diskStoreDirectory2 = temporaryFolder.newFolder("diskDir2");
```
mkevo, line 131: Use only one dir work. So when you change this part to:
```
    File[] diskStoreDirectories = new File[] {temporaryFolder.newFolder("diskDir1")};

    String diskDirs = Arrays.stream(diskStoreDirectories).map(File::getAbsolutePath)
        .collect(Collectors.joining(","));
```
test is executed successfully.

jujoramos:
@mkevo 
I know that having one single directory per `disk-store` avoids the issue entirely, **_but_** having multiple directories per `disk-store` is a supported configuration and it's also what one of our internal test scenarios is doing, that's why your commit for `GEODE-8119` was flagged in the first place, it caused all those test to fail with the exception:
```
[Could not find: "/tmp/junit5454972723589376526/diskDir2/BACKUPtestDisk.if, /tmp/junit5454972723589376526/diskDir3/BACKUPtestDisk.if"
] expected:<[OK]> but was:<[ERROR]>
```

I added the `DistributedTest` to this `PR` so you can reproduce the issue we internally see. In fact, if you take the same test and run it on the current `develop` branch (which doesn't include your changes for `GEODE-8119`) everything works just fine.

mkevo:
Yes, It was my mistake, I run it on develop but my local develop branch has my changes. :(
What I found that the problem is in change of this method: `validatedDirectoriesAndFile`.
Before it only checks if dir name is valid, but if you set wrong name thread will not shutdown.
So I add check name also and don't start thread if it name is not valid.

mkevo:
I will change this `validatedDirectoriesAndFile` method to check if dirs exist and if yes check if .if file exist in one of them. I tried some proof of concept and it works. Is it ok?
This is what I tried, but need to do refactoring on it:
```
  static String validatedDirectoriesAndFile(String[] diskDirs, String name) {
    String invalidDirectories = null;
    StringBuilder builder = null;
    File diskDir;
    File diskFile = null;
    int counter = 0;
    for (String diskDirPath : diskDirs) {
      diskDir = new File(diskDirPath);
      diskFile = new File(diskDirPath, "BACKUP" + name + IF_FILE_EXT);
      if (!diskDir.exists()) {
        if (builder == null) {
          builder = new StringBuilder();
        } else if (builder.length() != 0) {
          builder.append(", ");
        }
        builder.append(diskDir);
      } else if (diskFile.exists()) {
        counter++;
      }
    }
    if (counter == 0) {
      if (builder == null) {
        builder = new StringBuilder();
      } else if (builder.length() != 0) {
        builder.append(", ");
      }
      builder.append(diskFile);
    }
    if (builder != null) {
      invalidDirectories = builder.toString();
    }

    return invalidDirectories;
  }
```

jujoramos:
Hello @mkevo ,

I'd say that the best approach would be to open a new `PR` with your original changes + the fix + extra tests (the original `PR` didn't have much testing added) so members of the community can review it.
Once the `PR` is opened, also, I can go ahead and run some tests using our internal scenarios to see how it behaves compared to the original commit that caused the failiures.
Hope that works for you.

PR #5178, opened 2020-05-28
GEODE-8179: gfsh query cmd returns incorrect results if '=' sign is missing

alb3rtobr:
@davebarnes97 please take a look. As we cannot upgrade `spring shell` version I agreed with @pivotal-eshu to add a note in the `--query` option.


alb3rtobr:
Im going to close this PR by the moment, I want to check the code again. I tested with a simple spring-shell 1.2.0 app trying to confirm the problem exists there, but I saw that the `=` sign between the attribute name and value seems not been accepted. Thanks for the suggestion anyway @agingade

PR #5182, opened 2020-05-29
GEODE-7591: Fix for hang in ClusterDistributionManager

jvarenina:
Added timeout after which it will be periodically checked whether
distribution manager is disconnecting while waiting on the membership
view installation to avoid hanging

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/distributedTest/java/org/apache/geode/distributed/internal/ClusterDistributionManagerDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/distributed/internal/ClusterDistributionManagerDUnitTest.java
@@ -382,7 +382,8 @@ public class ClusterDistributionManagerDUnitTest extends CacheTestCase {
    * while waiting for the latest membership view to install
    */
   @Test
-  public void testWaitForViewInstallationDisconnectDS() {
+  public void testWaitForViewInstallationDisconnectDS()
+      throws InterruptedException, TimeoutException, BrokenBarrierException, ExecutionException {
     InternalDistributedSystem system = getSystem();
     ClusterDistributionManager dm = (ClusterDistributionManager) system.getDM();
     MembershipView<InternalDistributedMember> view = dm.getDistribution().getView();
@@ -397,14 +398,9 @@ public class ClusterDistributionManagerDUnitTest extends CacheTestCase {
       }
     });
 
-    try {
-      cyclicBarrier.await(getTimeout().toMillis(), TimeUnit.MILLISECONDS);
-      system.disconnect();
-      future.get(getTimeout().toMillis(), TimeUnit.MILLISECONDS);
-    } catch (InterruptedException | TimeoutException | ExecutionException
-        | BrokenBarrierException e) {
-      errorCollector.addError(e);
-    }
+    cyclicBarrier.await(getTimeout().toMillis(), TimeUnit.MILLISECONDS);
+    system.disconnect();
+    future.get(getTimeout().toMillis(), TimeUnit.MILLISECONDS);
   }
 
   private CacheListener<String, String> getSleepingListener(final boolean playDead) {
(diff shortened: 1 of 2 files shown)


geode-core/src/distributedTest/java/org/apache/geode/distributed/internal/ClusterDistributionManagerDUnitTest.java
```diff
@@ -370,6 +371,33 @@ public void testWaitForViewInstallation() {
...
+    executorService.submit(() -> {
+      try {
+        dm.waitForViewInstallation(view.getViewId() + 1);
+        waitForViewInstallationDone.set(true);
+      } catch (InterruptedException e) {
+        errorCollector.addError(e);
+      }
+    });
+
+    await().timeout(2000, TimeUnit.MILLISECONDS);
```
bschuchardt, line 394: pauses like this often cause tests to periodically fail in CI.  Waiting for a semaphore to be triggered by the executor thread might be better.

kirklund: I don't think await() without an until of some sort does anything. It's the until clauses that start the waiting. I recommend using a CyclicBarrier to sync up the actions of the two threads and a Future to wait for completion:
```
CyclicBarrier cyclicBarrier = new CyclicBarrier(2);

Future<Void> future = executorService.submit(() -> {
  try {
    cyclicBarrier.await(getTimeout().toMillis(), MILLISECONDS);
    dm.waitForViewInstallation(view.getViewId() + 1);
    errorCollector.addError(e);
  }
});

cyclicBarrier.await(getTimeout().toMillis(), MILLISECONDS);
system.disconnect();

future.getTimeout().toMillis(), MILLISECONDS);
```
There's still no guarantee that the executor thread will actually be deep inside dm.waitForViewInstallation when the main thread calls disconnect, but there will be some percentage of CI runs that are hitting this scenario as desired. So as long as the test remains passing 100% of the time, we're confident enough that the scenario is tested and passing.

To really guarantee the exact combination of two threads being in specific places would require some crazy test hooks down in dm and/or system which would probably require some unnatural acts of coding (not recommended). Another approach is to write an IntegrationTest or UnitTest in which some of the classes involved are mocks or spies but even then it's nice to have the dunit coverage as well.

jvarenina: Thank you both for the comments and elaborate proposal! I will update test as Kirk suggested.

geode-core/src/distributedTest/java/org/apache/geode/distributed/internal/ClusterDistributionManagerDUnitTest.java
```diff
@@ -370,6 +377,36 @@ public void testWaitForViewInstallation() {
...
+      } catch (InterruptedException | BrokenBarrierException | TimeoutException e) {
+        errorCollector.addError(e);
+      }
+    });
+
+    try {
+      cyclicBarrier.await(getTimeout().toMillis(), TimeUnit.MILLISECONDS);
+      system.disconnect();
+      future.get(getTimeout().toMillis(), TimeUnit.MILLISECONDS);
+    } catch (InterruptedException | TimeoutException | ExecutionException
```
kirklund, line 404: You can delete the catch-block and use of errorCollector in the main thread. You only need to do that inside the Runnable that you submitted. From line 400 on, it's better to just let the Exceptions percolate out of the test method and let junit catch it.

jvarenina: You are right! Done.

PR #5187, opened 2020-05-29
GEODE-8179: gfsh query cmd returns incorrect results if '=' sign is missing

alb3rtobr:
When parsing the user input in gfsh, the first `=` character position was used to split strings which were expected to follow the pattern `--parameter=value`. But if the `=` sign is not introduced, the `=` is deleted in `--parameter` and `value` (if exists).

In the case of `--query` parameter, the value of the parameter (the query itself) is very likely to contain a `=`, which could be removed from the query.

With this modification, if a query is introduced without the `=` between `--query` and the query statement, the `=` will not be removed from the query.

I also have included the `--J` parameter in this check, because its value is always containing a `=`.

### DIFF
--- a/geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/GfshParser.java
+++ b/geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/GfshParser.java
@@ -123,26 +123,24 @@ public class GfshParser extends SimpleParser {
 
     List<String> furtherSplitWithEquals = new ArrayList<>();
     for (String token : splitWithWhiteSpaces) {
-      int indexOfFirstEqual = token.indexOf('=');
-      int indexOfFirstDoubleQuote = token.indexOf('"');
-      int indexOfFirstSingleQuote = token.indexOf('\'');
-      if (indexOfFirstEqual < 0 || token.startsWith("-D")) {
+      // do not split with "=" if this part starts with quotes or is part of -D
+      if (token.startsWith("'") || token.startsWith("\"") || token.startsWith("-D")) {
         furtherSplitWithEquals.add(token);
-      } else {
-        if (indexOfFirstDoubleQuote == 0 || indexOfFirstSingleQuote == 0) {
-          furtherSplitWithEquals.add(token);
-        } else if ((indexOfFirstDoubleQuote < 0 && indexOfFirstSingleQuote < 0)
-            || (indexOfFirstEqual < indexOfFirstDoubleQuote)
-            || (indexOfFirstEqual < indexOfFirstSingleQuote)) {
-          String left = token.substring(0, indexOfFirstEqual);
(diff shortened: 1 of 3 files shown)


alb3rtobr:
I have rebased the PR now that LGTM Java Analysis seems working again.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/GfshParser.java
```diff
@@ -122,21 +122,30 @@ else if (insideQuoteOf == c) {
...
+          String right = token.substring(indexOfFirstEqual + 1);
+          if (left.length() > 0) {
+            furtherSplitWithEquals.add(left);
+          }
+          if (right.length() > 0) {
+            furtherSplitWithEquals.add(right);
+          }
         }
       }
+      if (token.equals("--query") || token.equals("--J")) {
```
jinmeiliao, line 146: anyway to not hardcode these two conditions? maybe based the decision on "do not remove = sign if it's in the quotes"?

alb3rtobr: good point, I have changed the solution. Thanks!

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/GfshParser.java
```diff
@@ -123,18 +123,25 @@ else if (insideQuoteOf == c) {
...
+          String right = token.substring(indexOfFirstEqual + 1);
+          if (left.length() > 0) {
+            furtherSplitWithEquals.add(left);
+          }
+          if (right.length() > 0) {
+            furtherSplitWithEquals.add(right);
+          }
         }
       }
     }
```
jinmeiliao, line 147: this entire for loop can be simplified to:

```
for (String token : splitWithWhiteSpaces) {
      // do not split with "=" if this part starts with quotes or is part of -D
      if(token.startsWith("'") || token.startsWith("\"") || token.startsWith("-D")) {
        furtherSplitWithEquals.add(token);
        continue;
      }
      // if this token has equal sign, split around the first occurrence of it
      int indexOfFirstEqual = token.indexOf('=');
      if (indexOfFirstEqual < 0) {
        furtherSplitWithEquals.add(token);
        continue;
      }
      String left = token.substring(0, indexOfFirstEqual);
      String right = token.substring(indexOfFirstEqual + 1);
      if (left.length() > 0) {
        furtherSplitWithEquals.add(left);
      }
      if (right.length() > 0) {
        furtherSplitWithEquals.add(right);
      }
    }
```

alb3rtobr: thanks!

PR #5188, opened 2020-05-30
GEODE-8099: add dlock around cms create/delete operations.

jinmeiliao:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/integrationTest/java/org/apache/geode/distributed/internal/InternalLocatorClusterManagementServiceIntegrationTest.java
+++ b/geode-core/src/integrationTest/java/org/apache/geode/distributed/internal/InternalLocatorClusterManagementServiceIntegrationTest.java
@@ -28,6 +28,7 @@ import java.net.URISyntaxException;
 import java.util.Optional;
 
 import org.junit.After;
+import org.junit.Before;
 import org.junit.Test;
 
 import org.apache.geode.cache.RegionShortcut;
@@ -43,11 +44,11 @@ import org.apache.geode.management.internal.BaseManagementService;
 public class InternalLocatorClusterManagementServiceIntegrationTest {
 
   private InternalLocator internalLocator; // the instance under test
-  private DistributionConfigImpl distributionConfig = mock(DistributionConfigImpl.class);
-  private InternalCacheForClientAccess cache = mock(InternalCacheForClientAccess.class);
-  private BaseManagementService managementService = mock(BaseManagementService.class);
-  private AgentUtil agentUtil = mock(AgentUtil.class);
-  private HttpService httpService = mock(HttpService.class);
+  private DistributionConfigImpl distributionConfig;
+  private InternalCacheForClientAccess cache;
+  private BaseManagementService managementService;
+  private AgentUtil agentUtil;
+  private HttpService httpService;
 
   @After
   public void tearDown() {
@@ -56,13 +57,38 @@ public class InternalLocatorClusterManagementServiceIntegrationTest {
     }
   }
 
+  @Before
+  public void setup() throws URISyntaxException {
+    distributionConfig = mock(DistributionConfigImpl.class);
+    cache = mock(InternalCacheForClientAccess.class);
+    managementService = mock(BaseManagementService.class);
+    agentUtil = mock(AgentUtil.class);
+    httpService = mock(HttpService.class);
+
+    LoggingSession loggingSession = mock(LoggingSession.class);
+    when(distributionConfig.getJmxManager()).thenReturn(true);
+    when(distributionConfig.getJmxManagerPort())
+        .thenReturn(AvailablePortHelper.getRandomAvailableTCPPort());
+    when(distributionConfig.getLocators()).thenReturn("");
+    when(distributionConfig.getSecurableCommunicationChannels())
+        .thenReturn(new SecurableCommunicationChannel[] {});
+    when(distributionConfig.getSecurityAuthTokenEnabledComponents()).thenReturn(new String[] {});
+    internalLocator = new InternalLocator(0, loggingSession, null, null, null, null,
+        null, null, distributionConfig, null);
+    InternalRegionFactory regionFactory = mock(InternalRegionFactory.class);
+    when(cache.createInternalRegionFactory(RegionShortcut.REPLICATE)).thenReturn(regionFactory);
+    when(cache.getOptionalService(HttpService.class))
+        .thenReturn(Optional.of(httpService));
+    when(cache.getCacheForProcessingClientRequests()).thenReturn(cache);
+    BaseManagementService.setManagementService(cache, managementService);
+    URI uri = new URI("file", "/management.war", null);
+    when(agentUtil.findWarLocation("geode-web-management")).thenReturn(uri);
+  }
+
   @Test
   public void startClusterManagementServiceWithRestServiceEnabledInvokesStartManager()
       throws Exception {
-    createInternalLocator();
-    setupForStartClusterManagementService();
     when(distributionConfig.getEnableManagementRestService()).thenReturn(true);
-
     internalLocator.startClusterManagementService(cache, agentUtil);
 
     verify(managementService).startManager();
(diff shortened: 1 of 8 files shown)


geode-core/src/main/java/org/apache/geode/management/internal/api/LocatorClusterManagementService.java
```diff
@@ -105,7 +107,12 @@
 import org.apache.geode.management.runtime.OperationResult;
 import org.apache.geode.management.runtime.RuntimeInfo;
 
+/**
+ * each locator will have one instance of this running if enabled
+ */
 public class LocatorClusterManagementService implements ClusterManagementService {
+  @VisibleForTesting
+  static final String CMS_NAME = "CMS_LOCK_SERVICE";
```
jchen21, line 115: It would be better to add some comment for this constant.

jchen21, line 115: It would be better to add some comments for this constant.

geode-core/src/test/java/org/apache/geode/management/internal/api/LocatorClusterManagementServiceTest.java
```diff
@@ -125,22 +127,48 @@ public void before() throws Exception {
...
+    verify(dLockService).lock(LocatorClusterManagementService.CMS_NAME, -1, -1);
+    verify(dLockService).unlock(LocatorClusterManagementService.CMS_NAME);
+  }
+
   @Test
   public void create_persistenceIsNull() {
     org.apache.geode.cache.Region<Object, Object> region =
         mock(org.apache.geode.cache.Region.class);
     when(cache.getRegion(any())).thenReturn(region);
-    service = new LocatorClusterManagementService(cache, null);
```
jchen21, line 143: This change removes the code coverage for the constructor `LocatorClusterManagementService(InternalCache cache, InternalConfigurationPersistenceService persistenceService)`. I don't see it is covered anywhere else.

geode-core/src/integrationTest/java/org/apache/geode/distributed/internal/InternalLocatorClusterManagementServiceIntegrationTest.java
```diff
@@ -56,13 +57,38 @@ public void tearDown() {
     }
   }
 
+  @Before
+  public void setup() throws URISyntaxException {
```
kirklund, line 61: I recommend moving all the field and var creations at the top, then move all the when blocks together, order both blocks in some way (I alphabetize), and then consider moving the subject(s) under test to the bottom of the setup method.

I also recommend cleaning up any statics in tearDown that you've configured in setUp. For example: how to undo this statement in tearDown:
```
BaseManagementService.setManagementService(cache, managementService);
```
The statics will still have this configured after each test completes.

geode-core/src/main/java/org/apache/geode/management/internal/api/LocatorClusterManagementService.java
```diff
@@ -114,6 +122,7 @@
   private final MemberValidator memberValidator;
   private final CommonConfigurationValidator commonValidator;
   private final InternalCache cache;
+  private DistributedLockService cmsDlockService;
```
kirklund, line 125: Let's move away from lazy-initializing and mutable fields and just make it final.

I'm assuming you made it mutable for testing. This is where you typically want to chain more than one constructor. One of the constructors needs `@VisibleForTesting` and pass in the DistributedLockService (as a mock or whatever). The main public constructor can then create the DistributedLockService and pass it as a parameter to the next constructor. The test would use the next constructor that accepts a DistributedLockService.

geode-core/src/integrationTest/java/org/apache/geode/distributed/internal/InternalLocatorClusterManagementServiceIntegrationTest.java
```diff
@@ -56,13 +57,38 @@ public void tearDown() {
...
+    distributionConfig = mock(DistributionConfigImpl.class);
+    cache = mock(InternalCacheForClientAccess.class);
+    managementService = mock(BaseManagementService.class);
```
mhansonp, line 64: With these changes does this test need to become a unit test rather than an integration test?

geode-core/src/main/java/org/apache/geode/management/internal/api/LocatorClusterManagementService.java
jinmeiliao: I tried your approach, but that affects lots of other integration tests that uses some API in InternalLocator that would start the cms service and hit error when they can't get the dlockservice to inject to it.

geode-core/src/main/java/org/apache/geode/management/internal/BaseManagementService.java
```diff
@@ -94,6 +94,13 @@ public static void setManagementService(InternalCacheForClientAccess cache,
     }
   }
 
+  @VisibleForTesting
+  public static void clearManagementService(InternalCacheForClientAccess cache) {
+    synchronized (instances) {
+      instances.remove(cache);
```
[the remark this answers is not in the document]
jinmeiliao: yeah, this is legacy code. for testing purpose, I had to add this here.

geode-core/src/test/java/org/apache/geode/management/internal/api/LocatorClusterManagementServiceTest.java
```diff
@@ -125,16 +127,41 @@ public void before() throws Exception {
...
     regionConfig.setName("region1");
 
     rebalanceOperation = new RebalanceOperation();
   }
 
+  @Test
+  public void lockAndUnlockCalledAtCreate() {
+    try {
+      service.create(regionConfig);
+    } catch (Exception ignore) {
```
kirklund, line 147: Is create expected to throw some Exception every time this test runs? If not, then let's remove the try-catch and add `throws Exception` clause to the method.

geode-core/src/test/java/org/apache/geode/management/internal/api/LocatorClusterManagementServiceTest.java
```diff
@@ -125,16 +127,41 @@ public void before() throws Exception {
...
+
+    verify(dLockService).lock(LocatorClusterManagementService.CMS_DLOCK_SERVICE_NAME, -1, -1);
+    verify(dLockService).unlock(LocatorClusterManagementService.CMS_DLOCK_SERVICE_NAME);
+  }
+
+  @Test
+  public void lockAndUnlockCalledAtDelete() {
+    try {
+      service.delete(regionConfig);
+    } catch (Exception ignore) {
```
kirklund, line 158: Can we remove try-block and use `throws Exception`?

geode-core/src/main/java/org/apache/geode/management/internal/api/LocatorClusterManagementService.java
```diff
@@ -154,6 +163,24 @@ public LocatorClusterManagementService(
     this.operationManager = operationManager;
   }
 
+  @VisibleForTesting
+  DistributedLockService getCmsDlockService() {
```
kirklund, line 167: This method should be `synchronized` since cmsDlockService is lazy initialized.

geode-core/src/test/java/org/apache/geode/management/internal/api/LocatorClusterManagementServiceTest.java
jinmeiliao: in this test setup, a NPE will already be thrown. This test is to make sure the two calls always get called when any exception happens. I changed the test method to be `lockAndUnlockCalledAtDeleteWithException`. And added the verification to another test setup when deletion is successful.

geode-core/src/main/java/org/apache/geode/management/internal/api/LocatorClusterManagementService.java
jinmeiliao: good catch.

geode-core/src/test/java/org/apache/geode/management/internal/api/LocatorClusterManagementServiceTest.java
jinmeiliao: same as "delete" case.

PR #5195, opened 2020-06-02
GEODE-8095 create restore redundancy

mhansonp:
First cut of REST restore redundancy
More tests need to be added, but current testing is pretty solid
because it is largely based on previous work done by others.


geode-management/src/main/java/org/apache/geode/management/operation/RestoreRedundancyRequest.java
```diff
@@ -0,0 +1,130 @@
...
+@Experimental
+public class RestoreRedundancyRequest
+    implements ClusterManagementOperation<RestoreRedundancyResponse> {
+
+  /**
+   * see {@link #getEndpoint()}
+   */
+  public static final String RESTORE_REDUNDANCY_REBALANCE_ENDPOINT =
+      "/operations/restoreRedundancy";
+  public static final String RESTORE_REDUNDANCY_OPERATOR = "restore redundancy";
```
jinmeiliao, line 39: I don't think we need this. Operator is whoever initiated this operation. I also think  this should be in the OperationState instead of the Operation request itself (but this could be another issue).

geode-management/src/main/java/org/apache/geode/management/internal/operation/RestoreRedundancyResponseImpl.java
```diff
@@ -0,0 +1,116 @@
...
+package org.apache.geode.management.internal.operation;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.apache.geode.management.api.ClusterManagementResult;
+import org.apache.geode.management.runtime.RegionRedundancyStatusSerializable;
+import org.apache.geode.management.runtime.RestoreRedundancyResponse;
+
+public class RestoreRedundancyResponseImpl
```
jinmeiliao, line 25: let's just make RestoreReduncyResponse a concrete class. I really don't see much benefit in having an interface.

jinmeiliao: also you don't have to call it Serializable in the class name, it fact that it implements OperationResult is already Serializable

geode-core/src/main/resources/org/apache/geode/internal/sanctioned-geode-core-serializables.txt
```diff
@@ -457,6 +457,12 @@ org/apache/geode/management/internal/functions/CacheRealizationFunction,false
 org/apache/geode/management/internal/functions/CliFunctionResult$StatusState,false
 org/apache/geode/management/internal/functions/GetMemberInformationFunction,true,1404642539058875565
 org/apache/geode/management/internal/functions/RebalanceFunction,true,1
+org/apache/geode/management/internal/functions/RestoreRedundancyFunction,true,1
+org/apache/geode/management/runtime/RestoreRedundancyResponse,true,1
```
jinmeiliao, line 461: I believe some of these belongs to sanctioned-geode-management-serializables.txt

geode-management/src/main/java/org/apache/geode/management/internal/operation/RestoreRedundancyResponseImpl.java
mhansonp: I used serializable in the name because the class is basically a copy of an existing one with the serializability being a distinguishing feature. This is one of those points of question where, should I just modify the RegionRedundancyStatus class to make it serializable?

mhansonp: > let's just make RestoreReduncyResponse a concrete class. I really don't see much benefit in having an interface.

I agree. It adds a layer of isolation that is not necessary.

geode-core/src/main/resources/org/apache/geode/internal/sanctioned-geode-core-serializables.txt
mhansonp: Great catch. This is one of the changes that I am just not sure about. I put in a placeholder values that allow it to work, but I need to really consider the actual values.

geode-management/src/main/java/org/apache/geode/management/internal/operation/RestoreRedundancyResponseImpl.java
jinmeiliao: > I used serializable in the name because the class is basically a copy of an existing one with the serializability being a distinguishing feature. This is one of those points of question where, should I just modify the RegionRedundancyStatus class to make it serializable?

that one is already DataSerializable, so you shouldn't have to make it Serializable. I think you can just use whatever they had used as the result before, just move them to the management module. If the object are used in locator/server function calls, then they must have been Serializable/DataSerializable already.

jinmeiliao:
also I would expect a bulk of the work is already done by the restoreredundancy command, so we are just reusing their code instead of adding a bunch of new code.

mhansonp:
Taking a different path.

PR #5196, opened 2020-06-02
GEODE-8197: Add launcher acceptance testing for custom logging config

kirklund:


kirklund:
@demery-pivotal Yep, I minimized the number of System.outs but I left them in on purpose to print out the command being executed. Two tests are currently disabled due to bug GEODE-8197 and will fail until it's fixed.

geode-assembly/src/acceptanceTest/java/org/apache/geode/launchers/LocatorLauncherWithPulseAcceptanceTest.java
```diff
@@ -0,0 +1,113 @@
...
+    System.out.println("Launching command: " + processBuilder.command());
+
+    locator = processBuilder
+        .start();
+
+    assertThat(locator.isAlive()).isTrue();
+
+    await().untilAsserted(() -> {
+      LogFileAssert.assertThat(locatorLogFile.toFile())
+          .exists()
```
jchen21, line 102: Why `locatorLogFile` exits for this test, while ``locatorLogFile` doesn't exist for the other tests in this pull request? Can you share what configuration makes this difference?

kirklund:
Looks like Docker blew up in AcceptanceTest. Retriggering...

geode-assembly/src/acceptanceTest/java/org/apache/geode/launchers/LocatorLauncherWithPulseAcceptanceTest.java
kirklund: LocatorLauncherWithPulseAcceptanceTest is the only test in this PR that doesn't provide a custom log4j2 xml file. So that means, it'll create a locator log file.

The others tests all use a custom log4j2 xml file which turns off logging to a file and turns on logging to stdout. These tests assert that the locator log file doesn't exist to help verify that the custom log4j2 xml file for that test was actually used.

geode-assembly/src/acceptanceTest/java/org/apache/geode/launchers/LocatorLauncherWithPulseAndCustomLogConfigAcceptanceTest.java
```diff
@@ -0,0 +1,268 @@
...
+          .contains("Located war: geode-pulse")
+          .contains("Adding webapp /pulse")
+          .contains("Starting server location for Distribution Locator")
+          .doesNotContain("geode-pulse war file was not found")
+          .doesNotContain("java.lang.IllegalStateException: No factory method found for class");
+    });
+  }
+
+  @Test
+  @Ignore("GEODE-8197")
```
jchen21, line 186: My understanding is that once GEODE-8197 related bug is fixed, this `@Ignore` annotation will be removed.

geode-assembly/src/acceptanceTest/java/org/apache/geode/launchers/LocatorLauncherWithPulseAndCustomLogConfigAcceptanceTest.java
```diff
@@ -0,0 +1,268 @@
...
```
jchen21, line 226: My understanding is that once GEODE-8197 related bug is fixed, this `@Ignore` annotation will be removed.

kirklund: Yes. This PR is only creating new tests. I'll follow up in the future with a fix.

geode-assembly/src/acceptanceTest/java/org/apache/geode/launchers/LocatorLauncherWithCustomLogConfigAcceptanceTest.java
```diff
@@ -0,0 +1,175 @@
...
+    }
+  }
+
+  @Test
+  public void locatorLauncherUsesLog4jConfigurationFile() throws Exception {
+    ProcessBuilder processBuilder = new ProcessBuilder()
+        .redirectErrorStream(true)
+        .redirectOutput(stdoutFile.toFile())
+        .directory(temporaryFolder.getRoot())
+        .command("java",
```
onichols-pivotal, line 96: If system property java.home is defined, please use `System.getProperty("java.home")+"/bin/java"` instead.  Only use `java` if system property java.home is not defined

kirklund: Thanks Owen!

kirklund:
@onichols-pivotal Yay, JAVA_HOME/bin/java worked! AcceptanceTests all passed.

onichols-pivotal:
> @onichols-pivotal Yay, JAVA_HOME/bin/java worked! AcceptanceTests all passed.

Does this also pass locally for you?  For developers that do not have JAVA_HOME set but do have `java` on their system path, is it reasonable to require them to set JAVA_HOME in order to run these tests locally?

PR #5203, opened 2020-06-02
GEODE-8186: UG add example for query --member option

alb3rtobr:


alb3rtobr:
> @alb3rtobr Nice work. I have a small grammar edit and a larger proposal: let's add the example as a new topic under Advanced Querying.
> Since I don't have permission to push to your fork, I'll email you a tarfile with the proposed changes and let you incorporate them.

Thanks Dave, I have added your changes. Its ok if you send a PR to my branch next time :)

PR #5207, opened 2020-06-03
GEODE-8218: Update docs to reflect changes to default redis-bind-address behavior

sabbeyPivotal:
Geode Redis API now listens on all local addresses when bind-address is not specified. The docs should be updated to reflect this change.

### DIFF
--- a/geode-docs/reference/topics/gemfire_properties.html.md.erb
+++ b/geode-docs/reference/topics/gemfire_properties.html.md.erb
@@ -489,7 +489,7 @@ off-heap-memory-size=120g</code></pre></td>
 <tr>
 <td>redis-bind-address</td>
 <td>Specifies the address on which the Redis API for <%=vars.product_name%> is listening. If set to the empty string or this property is not
-specified, the server will listen on all local addresses.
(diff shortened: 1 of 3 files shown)


geode-docs/reference/topics/gemfire_properties.html.md.erb
```diff
@@ -489,7 +489,7 @@ off-heap-memory-size=120g</code></pre></td>
 <tr>
 <td>redis-bind-address</td>
 <td>Specifies the address on which the Redis API for <%=vars.product_name%> is listening. If set to the empty string or this property is not
-specified, localhost is requested from the operating system.</td>
+specified, the server will listen on all local addresses.
```
davebarnes97, line 492: For consistency and ease of understanding, I recommend present tense:

...the server listens on all local addresses.

I'd like to see this change adopted for the doc (.erb) file, at least. I recommend it for the .java files, as well.

sabbeyPivotal: Just updated the doc and Java docs!  Thank you, Dave.

PR #5226, opened 2020-06-08
GEODE-8205: Feature flag unsupported Redis commands

sabbeyPivotal:
Added log message and redis --enable-unsupported-commands gfsh command.

Co-authored-by: Murtuza Boxwala <mboxwala@pivotal.io>
Co-authored-by: Jens Deppe <jdeppe@pivotal.io>


geode-redis/src/main/java/org/apache/geode/redis/internal/gfsh/RedisCommand.java
```diff
@@ -0,0 +1,51 @@
...
+  public static final String REDIS = "redis";
+  public static final String REDIS__HELP =
+      "Commands related to the Redis API for Geode.";
+  public static final String REDIS__ENABLE_UNSUPPORTED_COMMANDS = "enable-unsupported-commands";
+  public static final String REDIS__ENABLE_UNSUPPORTED_COMMANDS__HELP =
+      "Boolean value to determine "
+          + "whether or not to enable unsupported commands. Unsupported commands have not been fully tested.";
+
+  @CliCommand(value = {REDIS}, help = REDIS__HELP)
+  public ResultModel enableUnsupportedCommands(
```
jdeppe-pivotal, line 39: We should also add some security meta data here:
```
@ResourceOperation(resource = ResourcePermission.Resource.DATA, operation = ResourcePermission.Operation.MANAGE);
```

geode-redis/src/main/java/org/apache/geode/redis/internal/gfsh/RedisCommandFunction.java
```diff
@@ -0,0 +1,57 @@
...
+package org.apache.geode.redis.internal.gfsh;
+
+import org.apache.geode.cache.execute.Function;
+import org.apache.geode.cache.execute.FunctionContext;
+import org.apache.geode.cache.execute.FunctionService;
+import org.apache.geode.internal.cache.InternalCache;
+import org.apache.geode.management.internal.functions.CliFunctionResult;
+import org.apache.geode.redis.internal.GeodeRedisService;
+
+public class RedisCommandFunction implements Function<Boolean> {
```
jdeppe-pivotal, line 24: For consistency with other cli-related functions we should `extend CliFunction` instead of `implements Function`. That class exposes an `execute` that just needs to return a `CliFunctionResult`. `CliFunction` also extends `InternalFunction` which is more secure (it cannot be user-invoked in any way - for example via gfsh).

geode-redis/src/main/java/org/apache/geode/redis/internal/gfsh/RedisCommand.java
sabbeyPivotal: Updated it! Thank you!

geode-redis/src/main/java/org/apache/geode/redis/internal/gfsh/RedisCommandFunction.java
sabbeyPivotal: I updated it, let me know what you think.

PR #5230, opened 2020-06-10
Revert "GEODE-8234: redis functions now implement InternalFunction"

onichols-pivotal:
Reverts apache/geode#5229


onichols-pivotal:
This commit broke https://concourse.apachegeode-ci.info/teams/main/pipelines/apache-develop-main/jobs/UpgradeTestOpenJDK8/builds/254 if you can fix it that would be great too, but I believe our policy is revert-then-fix when you break the pipeline...

dschneider-pivotal:
Perhaps it would be better to disable this flaky test until someone works on it. But normally we just look at failures and find open tickets saying that they fail intermittently and ignore that failure. I'm not sure why that did not happen in this case. A ticket already existed but did not seem to help.

onichols-pivotal:
sounds like the consensus is not to revert

PR #5231, opened 2020-06-10
GEODE-8237: Add note about 'alter region' & cluster conf service

alb3rtobr:


alb3rtobr:
@jinmeiliao this PR adds a note in the documentation about my question you answered in the users lists. Could you take a look? Thanks

geode-docs/configuring/cluster_config/gfsh_persist.html.md.erb
```diff
@@ -88,6 +88,8 @@ The following `gfsh` commands cause the configuration to be written to all locat
 
 **\*** Note that the `configure pdx` command must be executed *before* starting your data members. This command does not affect any currently running members in the system. Data members (with cluster configuration enabled) that are started after running this command will pick up the new PDX configuration.
 
+**\*\*** The region this command is applied to must have been created via `gfsh`. If that is not the case, the command will fail saying the region does not exist.
```
jinmeiliao, line 91: Maybe change it to "the region this command is applied needs to exist in the cluster configuration. If that is not the case...."

alb3rtobr: done!

alb3rtobr:
> One change, please: change "needs to exist" to "must exist". Thanks.

done

PR #5236, opened 2020-06-11
GEODE-8241: Locator observes locator-wait-time

aaronlindsey:
In the case where a locator starts up and is unable to connect to any
other locators, it may decide to become the membership coordinator even
if locator-wait-time has not elapsed.

This change addresses this issue by requiring a locator to wait for
locator-wait-time before deciding to become the coordinator.

Co-authored-by: Aaron Lindsey <alindsey@vmware.com>
Co-Authored-By: Vincent Ford <vford@pivotal.io>
Co-authored-by: Bill Burcham <bburcham@pivotal.io>

### DIFF
--- a/geode-membership/src/integrationTest/java/org/apache/geode/distributed/internal/membership/gms/MembershipIntegrationTest.java
+++ b/geode-membership/src/integrationTest/java/org/apache/geode/distributed/internal/membership/gms/MembershipIntegrationTest.java
@@ -184,7 +184,7 @@ public class MembershipIntegrationTest {
 
   @Test
   public void secondMembershipPausesForLocatorWaitTime()
-      throws IOException, MemberStartupException {
+      throws IOException, MemberStartupException, InterruptedException {
 
     /*
      * Start a locator for the coordinator (membership) so we have a port for it.
@@ -219,6 +219,7 @@ public class MembershipIntegrationTest {
 
     final int[] locatorPorts = new int[] {coordinatorLocatorPort, lateJoiningLocatorPort};
 
+    // minimum duration a locator waits to become the coordinator, regardless of locatorWaitTime
     final Duration minimumJoinWaitTime = Duration
         // amount of sleep time per retry in GMSJoinLeave.join()
         .ofMillis(JOIN_RETRY_SLEEP + FIND_LOCATOR_RETRY_SLEEP)
@@ -226,12 +227,12 @@ public class MembershipIntegrationTest {
         .multipliedBy(getMinimumRetriesBeforeBecomingCoordinator(locatorPorts.length));
 
     /*
-     * By setting locatorWaitTime to 3x the minimumJoinWaitTime, we are trying to make sure the
+     * By setting locatorWaitTime to 10x the minimumJoinWaitTime, we are trying to make sure the
      * locatorWaitTime is sufficiently larger than the minimum so we can reliably detect whether
      * the lateJoiningMembership is waiting for the full locatorWaitTime and not just the minimum
      * wait time.
      */
-    final int locatorWaitTime = (int) (3 * minimumJoinWaitTime.getSeconds());
+    final int locatorWaitTime = (int) (10 * minimumJoinWaitTime.getSeconds());
 
     final MembershipConfig lateJoiningMembershipConfig =
         createMembershipConfig(true, locatorWaitTime, locatorPorts);
@@ -247,27 +248,19 @@ public class MembershipIntegrationTest {
     });
 
     /*
+     * By sleeping for 2x the minimumJoinWaitTime, we are trying to make sure we sleep for
+     * longer than the minimum but shorter than the locatorWaitTime so we can detect whether the
+     * lateJoiningMembership is waiting for the full locatorWaitTime and not just the minimum
+     * wait time.
+     */
+    Thread.sleep(2 * minimumJoinWaitTime.toMillis());
+
+    /*
      * Now start the coordinator (membership), after waiting longer than the minimum wait time for
      * connecting to a locator but shorter than the locator-wait-time.
      */
+    start(coordinatorMembership);
 
-    CompletableFuture<Void> coordinatorMembershipStartup = executorServiceRule.runAsync(() -> {
-      try {
-        /*
-         * By sleeping for 2x the minimumJoinWaitTime, we are trying to make sure we sleep for
-         * longer than the minimum but shorter than the locatorWaitTime so we can detect whether the
-         * lateJoiningMembership is waiting for the full locatorWaitTime and not just the minimum
-         * wait time.
-         */
-        Thread.sleep(2 * minimumJoinWaitTime.toMillis());
-        start(coordinatorMembership);
-      } catch (InterruptedException ignored) {
-      } catch (MemberStartupException e) {
-        throw new RuntimeException(e);
-      }
-    });
-
-    await().untilAsserted(() -> assertThat(coordinatorMembershipStartup).isCompleted());
     await().untilAsserted(() -> assertThat(lateJoiningMembershipStartup).isCompleted());
 
     await().untilAsserted(
--- a/geode-membership/src/integrationTest/java/org/apache/geode/distributed/internal/membership/gms/MembershipIntegrationTest.java
+++ b/geode-membership/src/integrationTest/java/org/apache/geode/distributed/internal/membership/gms/MembershipIntegrationTest.java
@@ -234,7 +234,7 @@ public class MembershipIntegrationTest {
       try {
         start(lateJoiningMembership);
       } catch (MemberStartupException e) {
-        e.printStackTrace();
+        throw new RuntimeException(e);
       }
     });
 
@@ -247,13 +247,14 @@ public class MembershipIntegrationTest {
       try {
         Thread.sleep(2 * minimumJoinWaitTime.toMillis());
         start(coordinatorMembership);
-      } catch (MemberStartupException | InterruptedException e) {
-        e.printStackTrace();
+      } catch (InterruptedException ignored) {
+      } catch (MemberStartupException e) {
+        throw new RuntimeException(e);
       }
     });
 
-    await().untilAsserted(() -> assertThat(coordinatorMembershipStartup).isDone());
-    await().untilAsserted(() -> assertThat(lateJoiningMembershipStartup).isDone());
+    await().untilAsserted(() -> assertThat(coordinatorMembershipStartup).isCompleted());
+    await().untilAsserted(() -> assertThat(lateJoiningMembershipStartup).isCompleted());
 
     await().untilAsserted(
         () -> assertThat(coordinatorMembership.getView().getMembers()).hasSize(2));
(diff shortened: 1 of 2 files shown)


aaronlindsey:
@vfordpivotal please review

aaronlindsey:
FWIW, I ran the test 100x on my machine and it passed each time. When I reverted the change to `GMSJoinLeave.java`, I saw the test fail about 30x consistently.

vfordpivotal:
I can further support this code change by having run this change multiple times in another test context and it addressed our start-up issues around start-up of locators when DNS resolution was slow. I feel very confident this code change addresses our issues, but would like additional feedback on the test code provided.

geode-membership/src/integrationTest/java/org/apache/geode/distributed/internal/membership/gms/MembershipIntegrationTest.java
```diff
@@ -172,6 +179,91 @@ public void secondMembershipCanJoinUsingTheSecondLocatorToStart()
...
+    final MembershipConfig lateJoiningMembershipConfig =
+        createMembershipConfig(true, locatorWaitTime, lateJoiningMembershipLocatorPorts);
+    final Membership<MemberIdentifier> lateJoiningMembership =
+        createMembership(lateJoiningMembershipConfig, lateJoiningLocator);
+
+    CompletableFuture<Void> lateJoiningMembershipStartup = executorServiceRule.runAsync(() -> {
+      try {
+        start(lateJoiningMembership);
+      } catch (MemberStartupException e) {
+        e.printStackTrace();
```
bschuchardt, line 237: There are a couple of these auto-generated try/catch statements in the test.  If they're failure conditions shouldn't the test fail?  If they're not, what good are these stack traces?

geode-membership/src/integrationTest/java/org/apache/geode/distributed/internal/membership/gms/MembershipIntegrationTest.java
```diff
@@ -172,6 +179,91 @@ public void secondMembershipCanJoinUsingTheSecondLocatorToStart()
...
+
+    final MembershipLocator<MemberIdentifier> lateJoiningLocator = createLocator(0);
+    lateJoiningLocator.start();
+    final int lateJoiningLocatorPort = lateJoiningLocator.getPort();
+
+    final int[] lateJoiningMembershipLocatorPorts =
+        new int[] {coordinatorLocatorPort, lateJoiningLocatorPort};
+
+    final Duration minimumJoinWaitTime = Duration
+        .ofMillis(2_000) // expected amount of sleep time per loop in GMSJoinLeave.join()
```
aaronlindsey, line 224: It seems like it would be better to reference the same constants that are used in `GMSJoinLeave` instead of hard-coding this value.

geode-membership/src/integrationTest/java/org/apache/geode/distributed/internal/membership/gms/MembershipIntegrationTest.java
aaronlindsey: Updated in latest commits

geode-membership/src/integrationTest/java/org/apache/geode/distributed/internal/membership/gms/MembershipIntegrationTest.java
aaronlindsey: Updated in latest commits

geode-membership/src/integrationTest/java/org/apache/geode/distributed/internal/membership/gms/MembershipIntegrationTest.java
```diff
@@ -172,6 +181,92 @@ public void secondMembershipCanJoinUsingTheSecondLocatorToStart()
...
+    lateJoiningLocator.start();
+    final int lateJoiningLocatorPort = lateJoiningLocator.getPort();
+
+    final int[] lateJoiningMembershipLocatorPorts =
+        new int[] {coordinatorLocatorPort, lateJoiningLocatorPort};
+
+    final Duration minimumJoinWaitTime = Duration
+        .ofMillis(JOIN_RETRY_SLEEP + FIND_LOCATOR_RETRY_SLEEP) // amount of sleep time per retry
+        .multipliedBy(lateJoiningMembershipLocatorPorts.length * 2); // expected number of retries
+    final int locatorWaitTime = (int) (3 * minimumJoinWaitTime.getSeconds());
```
Bill, line 228: where does `3` come from?

geode-membership/src/integrationTest/java/org/apache/geode/distributed/internal/membership/gms/MembershipIntegrationTest.java
```diff
@@ -172,6 +181,92 @@ public void secondMembershipCanJoinUsingTheSecondLocatorToStart()
...
+    final MembershipLocator<MemberIdentifier> lateJoiningLocator = createLocator(0);
+    lateJoiningLocator.start();
+    final int lateJoiningLocatorPort = lateJoiningLocator.getPort();
+
+    final int[] lateJoiningMembershipLocatorPorts =
+        new int[] {coordinatorLocatorPort, lateJoiningLocatorPort};
+
+    final Duration minimumJoinWaitTime = Duration
+        .ofMillis(JOIN_RETRY_SLEEP + FIND_LOCATOR_RETRY_SLEEP) // amount of sleep time per retry
+        .multipliedBy(lateJoiningMembershipLocatorPorts.length * 2); // expected number of retries
```
Bill, line 227: where does `2` come from? even if you don't want to explicitly reference an internal constant, it'd be nice to document where that constant lives

geode-membership/src/integrationTest/java/org/apache/geode/distributed/internal/membership/gms/MembershipIntegrationTest.java
```diff
@@ -172,6 +181,92 @@ public void secondMembershipCanJoinUsingTheSecondLocatorToStart()
...
+    });
+
+    /*
+     * Now start the coordinator (membership), after waiting longer than the minimum wait time for
+     * connecting to a locator but shorter than the locator-wait-time.
+     */
+
+    CompletableFuture<Void> coordinatorMembershipStartup = executorServiceRule.runAsync(() -> {
+      try {
+        Thread.sleep(2 * minimumJoinWaitTime.toMillis());
```
Bill, line 250: where does this `2` come from?

geode-membership/src/integrationTest/java/org/apache/geode/distributed/internal/membership/gms/MembershipIntegrationTest.java
aaronlindsey: There is a minimum amount of time a locator will wait to become the coordinator, regardless of the value for `locator-wait-time`. The variable, `minimumJoinWaitTime`, is meant to capture that value. By setting `locator-wait-time` to 3x the `minimumJoinWaitTime`, we are just trying to make sure the `locator-wait-time` is sufficiently larger than the minimum so we can reliably detect whether the `lateJoiningMembership` is waiting for the full `locator-wait-time` and not just the minimum wait time.

geode-membership/src/integrationTest/java/org/apache/geode/distributed/internal/membership/gms/MembershipIntegrationTest.java
aaronlindsey: It comes from [this line](https://github.com/apache/geode/blob/e5426f79870e3217fc7c66572d2dddf44dbddd43/geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/membership/GMSJoinLeave.java#L327). I could assign the `2` to a constant and reference that if it would be clearer.

geode-membership/src/integrationTest/java/org/apache/geode/distributed/internal/membership/gms/MembershipIntegrationTest.java
aaronlindsey: Again, `minimumJoinWaitTime` is the minimum amount of time a locator will wait to become a coordinator. We are just trying to make sure we sleep for longer than the minimum so we can detect whether the `lateJoiningMembership` is waiting for the full `locator-wait-time` and not just the minimum wait time.

aaronlindsey:
> Once those are answered, I feel like we are still left with a test that will be flaky because of all the timing dependencies here.

I understand your concern. We tried to choose the 2x and 3x values so that there is a sufficient buffer between (a) the minimum wait time, (b) the time at which the coordinator membership becomes "reachable", and (c) the locator wait time. We saw that with these values the test did pass and fail reliably. We also tried writing a test using `GMSJoinLeaveJUnitTest` but it had the same type of timing dependencies.

The difficulty here is that we're trying to write "black box" style tests for a timeout. I don't see a way to avoid the timing dependencies without a large refactor, which I was trying to avoid since this is otherwise a small change with very little risk.

aaronlindsey:
@Bill what do you think of mocking out calls to `System.currentTimeMillis()` in `GMSJoinLeave`? Allowing the test to control the clock might allow a more deterministic test. I'm not sure if it would work, though, because there are a lot of usages of `System.currentTimeMillis()` in `GMSJoinLeave` that our test doesn't care about which would also be affected.

geode-membership/src/integrationTest/java/org/apache/geode/distributed/internal/membership/gms/MembershipIntegrationTest.java
aaronlindsey: I added a comment to the code in the latest commit.

geode-membership/src/integrationTest/java/org/apache/geode/distributed/internal/membership/gms/MembershipIntegrationTest.java
aaronlindsey: I added a comment to the code in the latest commit.

geode-membership/src/integrationTest/java/org/apache/geode/distributed/internal/membership/gms/MembershipIntegrationTest.java
aaronlindsey: I extracted a method for calculating the minimum number of retries before becoming coordinator in the latest commit.

aaronlindsey:
I've been running this test in the background while I work and so far it has passed 300x in a row without failing. I will continue letting it run until it reaches 1000 iterations.

aaronlindsey:
@Bill @bschuchardt with the previous changes I noticed a failure after 400-something runs. The chosen `locatorWaitTime` was not quite long enough. So I greatly increased the `locatorWaitTime` in the latest commit and then I was able to run the test 1000x successfully. I'm pretty confident that after the latest commit the chance of flakiness is negligible.

aaronlindsey:
The failing `AcceptanceTestOpenJDK11` check is also failing on develop.

aaronlindsey:
I rebased on develop to pick up the fix for the failing `AcceptanceTestOpenJDK11` check.

PR #5237, opened 2020-06-11
GEODE-8268: clean up ExecutionHandlerContext

dschneider-pivotal:
The ShutdownExecutor now implements the shutdown command
instead of it being embedded in the ExecutionHandlerContext.
The SubscribeExecutor now takes care of changing its channel to a different thread group.

### For all changes:

### Note:

### DIFF
--- a/geode-redis/src/main/java/org/apache/geode/redis/internal/netty/ExecutionHandlerContext.java
+++ b/geode-redis/src/main/java/org/apache/geode/redis/internal/netty/ExecutionHandlerContext.java
@@ -33,7 +33,6 @@ import io.netty.channel.EventLoopGroup;
 import io.netty.handler.codec.DecoderException;
 import org.apache.logging.log4j.Logger;
 
-import org.apache.geode.cache.Cache;
 import org.apache.geode.cache.CacheClosedException;
 import org.apache.geode.cache.execute.FunctionException;
 import org.apache.geode.logging.internal.log4j.api.LogService;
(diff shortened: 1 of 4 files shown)


geode-redis/src/main/java/org/apache/geode/redis/internal/netty/ExecutionHandlerContext.java
```diff
@@ -264,7 +238,7 @@ private void moveSubscribeToNewEventLoopGroup(ChannelHandlerContext ctx, Command
     if (command.isOfType(SUBSCRIBE)) {
       CountDownLatch latch = new CountDownLatch(0);
       ctx.channel().deregister().addListener((ChannelFutureListener) future -> {
-        subscriberEventLoopGroup.register(ctx.channel()).sync();
+        server.getSubscriberGroup().register(ctx.channel()).sync();
```
jdeppe-pivotal, line 241: I wonder if there is some 'feature-envy' creep happening here. What do you think about pushing the registration functionality onto the `server`? I'm also fine with this approach.

geode-redis/src/main/java/org/apache/geode/redis/internal/netty/ExecutionHandlerContext.java
```diff
@@ -264,7 +238,7 @@ private void moveSubscribeToNewEventLoopGroup(ChannelHandlerContext ctx, Command
     if (command.isOfType(SUBSCRIBE)) {
       CountDownLatch latch = new CountDownLatch(0);
```
jdeppe-pivotal, line 239: This looks like a bug in that the latch should be initialized with `1`. This is my bug :(.

geode-redis/src/main/java/org/apache/geode/redis/internal/netty/ExecutionHandlerContext.java
dschneider-pivotal: What I would like to see here is that this code (moveSubscribeToNewEventLoopGroup) to the SubscribeExecutor. Is that possible? I don't really understand what this is about and maybe it is important to do it after the executor has completed.
I'd also like to see all the netty specific code get moved out of GeodeRedisServer into something like NettyServer in our netty package just to have all the interactions with netty be in that one package. So then this code would interact with the NettyServer for this registration. But then this code would need to get access to the NettyServer (probably from the GeodeRedisServer). If we could break this class's dependency on GeodeRedisServer that might be a good thing but for now we need the GeodeRedisServer anyway (I think so we can ask it to shutdown and if unsupported commands are allowed).

PR #5249, opened 2020-06-15
GEODE-8272 Refactor Restore Redundancy Command

mhansonp:
This set of changes is to refactor the Restore Redundancy Command in GFSH through to the actual Restore Redundancy Operation that is performed.


geode-core/src/test/java/org/apache/geode/internal/cache/control/RestoreRedundancyOperationImplTest.java
```diff
@@ -38,19 +38,19 @@
...
+import org.apache.geode.management.runtime.RegionRedundancyStatus;
+import org.apache.geode.management.runtime.RestoreRedundancyResults;
 
 public class RestoreRedundancyOperationImplTest {
   InternalCache cache;
   InternalResourceManager manager;
   ResourceManagerStats stats;
   RestoreRedundancyOperationImpl operation;
-  RestoreRedundancyResultsImpl emptyResults;
+  SerializableRestoreRedundancyResultsImpl emptyResults;
```
kirklund, line 53: Let's make all of these fields `private`. There's an inspection that finds and automates this.

geode-core/src/test/java/org/apache/geode/internal/cache/control/RestoreRedundancyResultsImplTest.java
```diff
@@ -102,7 +108,7 @@ public void getMessageReturnsStatusForAllRegionsAndPrimaryInfo() {
 
     results.addPrimaryReassignmentDetails(details);
 
-    String message = results.getMessage();
+    String message = results.getRegionOperationMessage();
     List<String> messageLines = Arrays.asList(message.split("\n"));
```
kirklund, line 112: This is untouched code, but we should change all `"\n"` uses with System.lineSeparator() to ensure it is platform independent.

geode-core/src/test/java/org/apache/geode/internal/cache/control/RestoreRedundancyResultsImplTest.java
```diff
@@ -68,30 +68,36 @@ public void setUp() {
     when(zeroRedundancyRegionResult.getRegionName()).thenReturn(zeroRedundancyRegionName);
     when(details.getPrimaryTransfersCompleted()).thenReturn(transfersCompleted);
     when(details.getPrimaryTransferTime()).thenReturn(transferTime);
-    results = new RestoreRedundancyResultsImpl();
+    results = new SerializableRestoreRedundancyResultsImpl();
+  }
+
+  @Test
+  public void initialStateIsSuccess() throws Exception {
```
kirklund, line 75: You can delete the `throws Exception`.

geode-core/src/test/java/org/apache/geode/management/internal/operation/RestoreRedundancyPerformerTest.java
```diff
@@ -0,0 +1,263 @@
...
+import org.apache.geode.management.DistributedSystemMXBean;
+import org.apache.geode.management.internal.BaseManagementService;
+import org.apache.geode.management.internal.functions.RestoreRedundancyFunction;
+import org.apache.geode.management.operation.RestoreRedundancyRequest;
+import org.apache.geode.management.runtime.RegionRedundancyStatus;
+import org.apache.geode.management.runtime.RestoreRedundancyResults;
+
+public class RestoreRedundancyPerformerTest {
+
+  public static final String DS_MEMBER_NAME_SERVER1 = "server1";
```
kirklund, line 51: Please make all the constants in this test `private`.

geode-core/src/test/java/org/apache/geode/management/internal/operation/RestoreRedundancyPerformerTest.java
```diff
@@ -0,0 +1,263 @@
...
+    BaseManagementService baseManagementService = mock(BaseManagementService.class);
+    DistributedSystemMXBean distributedSystemMXBean = mock(DistributedSystemMXBean.class);
+    DistributedRegionMXBean distributedRegionMXBean = mock(DistributedRegionMXBean.class);
+    server1 = mock(InternalDistributedMember.class);
+    server2 = mock(InternalDistributedMember.class);
+    internalCacheForClientAccess = mock(InternalCacheForClientAccess.class);
+    InternalDistributedSystem internalDistributedSystem = mock(InternalDistributedSystem.class);
+    DistributionManager distributionManager = mock(DistributionManager.class);
+    when(baseManagementService.getDistributedSystemMXBean()).thenReturn(distributedSystemMXBean);
+    when(baseManagementService.getDistributedRegionMXBean(Mockito.anyString()))
```
kirklund, line 72: All of the `Mockito.*****` matchers in this class are actually owned by `org.mockito.ArgumentMatchers`. I would recommend just converting them all to static import which should switch them to importing from ArgumentMatchers.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/RedundancyCommand.java
```diff
@@ -148,81 +126,37 @@ void populateLists(List<RebalanceOperationPerformer.MemberPRInfo> membersForEach
...
-      }
-      functionResults.add(functionResult);
-      completedMembers.add(targetMember);
-    }
-    return functionResults;
-  }
-
-  ResultModel buildResultModelFromFunctionResults(List<CliFunctionResult> functionResults,
-      List<String> includedRegionsWithNoMembers, boolean isStatusCommand) {
+  ResultModel buildResultModelFromFunctionResults(RestoreRedundancyResults results,
```
kirklund, line 129: buildResultModelFromFunctionResults should be private until something outside the class needs to reference it.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/RedundancyCommand.java
```diff
@@ -15,35 +15,33 @@
...
 import org.apache.geode.management.internal.cli.result.model.InfoResultModel;
 import org.apache.geode.management.internal.cli.result.model.ResultModel;
-import org.apache.geode.management.internal.functions.CliFunctionResult;
 import org.apache.geode.management.internal.operation.RebalanceOperationPerformer;
+import org.apache.geode.management.internal.operation.RestoreRedundancyPerformer;
+import org.apache.geode.management.operation.RestoreRedundancyRequest;
+import org.apache.geode.management.runtime.RegionRedundancyStatus;
+import org.apache.geode.management.runtime.RestoreRedundancyResults;
 
 public class RedundancyCommand extends GfshCommand {
```
kirklund, line 46: Optional: RedundancyCommand has unused methods and one constant that can be deleted. It also has constants and other methods that should be changed to private sometime. Most of this is untouched code.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/RedundancyCommand.java
```diff
@@ -148,81 +126,37 @@ void populateLists(List<RebalanceOperationPerformer.MemberPRInfo> membersForEach
...
-  }
-
-  ResultModel buildResultModelFromFunctionResults(List<CliFunctionResult> functionResults,
-      List<String> includedRegionsWithNoMembers, boolean isStatusCommand) {
+  ResultModel buildResultModelFromFunctionResults(RestoreRedundancyResults results,
+      boolean isStatusCommand) {
     // No members hosting partitioned regions were found, but no regions were explicitly included,
     // so return OK status
-    if (functionResults.size() == 0 && includedRegionsWithNoMembers.size() == 0) {
+    if (results.getRegionResults().size() == 0
```
kirklund, line 133: This class has `size()` calls that would be more appropriate as `sEmpty()` calls.

geode-management/src/main/java/org/apache/geode/management/internal/operation/RegionRedundancyStatusImpl.java
```diff
@@ -0,0 +1,103 @@
...
+package org.apache.geode.management.internal.operation;
+
+import org.apache.geode.management.runtime.RegionRedundancyStatus;
+
+/**
+ * result object used by the cms that only needs to be json serializable
+ */
+public class RegionRedundancyStatusImpl implements RegionRedundancyStatus {
+
+  public static final String OUTPUT_STRING =
```
kirklund, line 24: OUTPUT_STRING should be `protected` instead of `public`.

geode-management/src/main/java/org/apache/geode/management/internal/operation/RegionRedundancyStatusImpl.java
```diff
@@ -0,0 +1,103 @@
...
+   * The {@link RedundancyStatus} of the region used to create this object at time of creation.
+   */
+  protected RedundancyStatus status;
+
+  /**
+   * Default constructor used for serialization
+   */
+  public RegionRedundancyStatusImpl() {}
+
+  public RegionRedundancyStatusImpl(int configuredRedundancy, int actualRedundancy,
```
kirklund, line 52: The non-default constructor should be package-private (no qualifier) with the annotation `@VisibleForTesting`

geode-management/src/main/java/org/apache/geode/management/internal/operation/RegionRedundancyStatusImpl.java
```diff
@@ -0,0 +1,103 @@
...
+
+  /**
+   * Determines the {@link RedundancyStatus} for the region. If redundancy is not configured (i.e.
+   * configured redundancy = 0), this always returns {@link RedundancyStatus#SATISFIED}.
+   *
+   * @param desiredRedundancy The configured redundancy of the region.
+   * @param actualRedundancy The actual redundancy of the region.
+   * @return The {@link RedundancyStatus} for the region.
+   */
+  private RedundancyStatus determineStatus(int desiredRedundancy, int actualRedundancy) {
```
kirklund, line 89: determineStatus is unused.

geode-management/src/main/java/org/apache/geode/management/internal/operation/RestoreRedundancyResultsImpl.java
```diff
@@ -12,29 +12,22 @@
...
-
-public class RestoreRedundancyResultsImpl
-    implements RestoreRedundancyResults, DataSerializableFixedID {
+import org.apache.geode.management.runtime.RegionRedundancyStatus;
+import org.apache.geode.management.runtime.RestoreRedundancyResults;
+
+/**
+ * result object used by the cms that only needs to be json serializable
+ */
+public class RestoreRedundancyResultsImpl implements RestoreRedundancyResults {
```
kirklund, line 30: Optional: This class has some `size()` calls that could be changed to use `isEmpty()` if you want. Also some unused code and overly public constants and methods.

geode-management/src/main/java/org/apache/geode/management/operation/RestoreRedundancyRequest.java
```diff
@@ -0,0 +1,123 @@
...
+@Experimental
+public class RestoreRedundancyRequest
+    implements ClusterManagementOperation<RestoreRedundancyResults> {
+
+  /**
+   * see {@link #getEndpoint()}
+   */
+  public static final String RESTORE_REDUNDANCY_REBALANCE_ENDPOINT =
+      "/operations/restoreRedundancy";
+  // null means all regions included
```
kirklund, line 37: This is a dangling comment which can be orphaned easily if you move code around. I would make it a javadoc so that it "sticks" to the field:
```
/** null means all regions included */
private List<String> includeRegions;
```
Same thing for the comment above `excludeRegions`.

geode-management/src/main/java/org/apache/geode/management/operation/RestoreRedundancyRequest.java
```diff
@@ -0,0 +1,123 @@
...
+ * Defines a distributed system request to optimize bucket allocation across members.
+ */
+@Experimental
+public class RestoreRedundancyRequest
+    implements ClusterManagementOperation<RestoreRedundancyResults> {
+
+  /**
+   * see {@link #getEndpoint()}
+   */
+  public static final String RESTORE_REDUNDANCY_REBALANCE_ENDPOINT =
```
kirklund, line 35: RESTORE_REDUNDANCY_REBALANCE_ENDPOINT should be private.

geode-management/src/main/java/org/apache/geode/management/operation/RestoreRedundancyRequest.java
```diff
@@ -0,0 +1,123 @@
...
+   * by default, requests all partitioned regions to be rebalanced
+   */
+  public RestoreRedundancyRequest() {}
+
+  /**
+   * copy constructor
+   */
+  public RestoreRedundancyRequest(
+      RestoreRedundancyRequest other) {
+    this.setExcludeRegions(other.getExcludeRegions());
```
kirklund, line 54: Lots of unnecessary `this.` qualifiers.

geode-management/src/main/java/org/apache/geode/management/operation/RestoreRedundancyRequest.java
```diff
@@ -0,0 +1,123 @@
...
+  public String getEndpoint() {
+    return RESTORE_REDUNDANCY_REBALANCE_ENDPOINT;
+  }
+
+  @Override
+  public String getOperator() {
+    return operator;
+  }
+
+  public void setOperator(String operator) {
```
kirklund, line 110: setOperator is unused.

geode-core/src/test/java/org/apache/geode/management/internal/functions/RestoreRedundancyFunctionTest.java
```diff
@@ -0,0 +1,163 @@
...
+
+    SerializableRestoreRedundancyResultsImpl result = argumentCaptor.getValue();
+    verify(result).setSuccess(true);
+    assertThat(result.getRegionOperationStatus())
+        .isEqualTo(RestoreRedundancyResults.Status.SUCCESS);
+    assertThat(result).isSameAs(mockResults);
+  }
+
+  @Test
+  public void whenFunctionThrowException() throws Exception {
```
DonalEvans, line 155: This test name could be a little more descriptive, saying what the expected behaviour is given the test conditions, such as "executeFunctionReturnsFailureResultWhenExceptionIsThrownDuringOperation". Also, an exception is never thrown from this method, so the `throws Exception` can be removed from the method signature.

geode-core/src/main/java/org/apache/geode/management/internal/operation/RestoreRedundancyPerformer.java
```diff
@@ -0,0 +1,179 @@
...
+    finalResult.addIncludedRegionsWithNoMembers(includedRegionsWithNoMembers);
+    for (RestoreRedundancyResults functionResult : functionResults) {
+      finalResult.addRegionResults(functionResult);
+      finalResult.setSuccess(functionResult.getSuccess());
+      finalResult.setStatusMessage(functionResult.getStatusMessage());
+    }
+    return finalResult;
+  }
+
+  // this returns either an Exception or RestoreRedundancyResults
```
DonalEvans, line 115: This comment does not seem entirely accurate. The method either returns `null` or a `RestoreRedundancyResults` object.

geode-management/src/main/java/org/apache/geode/management/operation/RestoreRedundancyRequest.java
mhansonp: I believe this is used by serialization.

geode-core/src/main/java/org/apache/geode/management/internal/operation/RestoreRedundancyPerformer.java
```diff
@@ -0,0 +1,179 @@
...
+      functionResults.add(functionResult);
+      completedMembers.add(targetMember);
+    }
+
+    RestoreRedundancyResultsImpl finalResult = new RestoreRedundancyResultsImpl();
+    finalResult.addIncludedRegionsWithNoMembers(includedRegionsWithNoMembers);
+    for (RestoreRedundancyResults functionResult : functionResults) {
+      finalResult.addRegionResults(functionResult);
+      finalResult.setSuccess(functionResult.getSuccess());
+      finalResult.setStatusMessage(functionResult.getStatusMessage());
```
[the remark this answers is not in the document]
mhansonp: Yes, but we are setting the OperationResult implementation.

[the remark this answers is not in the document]
DonalEvans: I'm not sure I follow. `setSuccess()` and `setStatusMessage()` are just setting fields on the `finalResult` object, which is then returned once we exit the loop. Why do we need to set them multiple times, overwriting the previous value each time? Can't we just set them once?

geode-core/src/test/java/org/apache/geode/management/internal/functions/RestoreRedundancyFunctionTest.java
```diff
@@ -0,0 +1,163 @@
...
+  private RestoreRedundancyFunction function;
+  private ResultSender resultSender;
+  private ArgumentCaptor<SerializableRestoreRedundancyResultsImpl> argumentCaptor;
+  private RestoreRedundancyRequest request;
+
+  @Before
+  public void setUp() throws InterruptedException, ExecutionException {
+    function = new RestoreRedundancyFunction();
+    when(mockContext.getCache()).thenReturn(mockCache);
+    request = new RestoreRedundancyRequest();
```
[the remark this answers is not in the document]
mhansonp: I was looking at that and it does seem like its possible, but it doesn't seem worth it since the request is just a flat data class and doesn't do much in it.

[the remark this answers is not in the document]
DonalEvans: If we can be reasonably sure that this will remain true, and that no additional logic will be added to `RestoreRedundancyRequest` in the future, then this is fine to leave as it is.

geode-core/src/main/java/org/apache/geode/management/internal/operation/RestoreRedundancyPerformer.java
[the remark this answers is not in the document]
mhansonp: Agreed. Since we no longer getting errors, this can be simplified.

geode-core/src/main/java/org/apache/geode/management/internal/functions/RestoreRedundancyFunction.java
```diff
@@ -0,0 +1,105 @@
...
+        throw e;
+      }
+      results.setSuccess(true);
+      results.setStatusMessage("Success"); // MLH change this
+    } catch (Exception e) {
+      results =
+          new SerializableRestoreRedundancyResultsImpl();
+      results.setSuccess(false);
+      results.setStatusMessage(e.getMessage());
+    }
```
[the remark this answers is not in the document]
mhansonp: I don't believe this is required either.

geode-core/src/test/java/org/apache/geode/management/internal/functions/RestoreRedundancyFunctionTest.java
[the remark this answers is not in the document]
mhansonp: I think it is reasonable at this point to believe that. It is basically like a work order being passed in...

geode-core/src/main/java/org/apache/geode/management/internal/operation/RestoreRedundancyPerformer.java
```diff
@@ -112,14 +114,14 @@ public RestoreRedundancyResults perform(Cache cache, RestoreRedundancyRequest op
     return finalResult;
   }
 
-  // this returns either an Exception or RestoreRedundancyResults
+  // this returns RestoreRedundancyResults or null based on
```
DonalEvans, line 118: Incomplete comment here?

geode-management/src/main/java/org/apache/geode/management/internal/operation/RegionRedundancyStatusImpl.java
mhansonp: Its necessary for serialization.

mhansonp: Sorry misread that... I agree.

geode-core/src/main/java/org/apache/geode/management/internal/operation/RestoreRedundancyPerformer.java
mhansonp: Good catch. Task switching is bad. The comment just restates what is visible in the code so I deleted it.

mhansonp: Thanks for your attention to detail.

geode-management/src/main/java/org/apache/geode/management/operation/RestoreRedundancyRequest.java
jinmeiliao: I believe you will need to set the operator in the controller, just like what the rebalance did

jinmeiliao:
> I have a question about the packages and modules of the classes. e.g. `RestoreRedundancyResultsImpl` is in package `org.apache.geode.management.internal.operation`. And it is in `geode-management` module. Its subclass `SerializableRestoreRedundancyResultsImpl` is in a different package `org.apache.geode.internal.cache.control` and a different module `geode-core`. Why these two classes are in different packages and different modules?
> Also some classes have been moved to a different package. e.g. `RegionRedundancyStatus` and `RestoreRedundancyResults`. Is there a reason for that?

RestoreRedundancyResults is an interface and is external api, RestoreRedundancyResultsImpl is a direct implementation of it. These two are in the geode-management module because cms needs them to convey result data back to the client. "SerializableRestoreRedundancyResultsImpl", however, is needed to convey function call results from locator to servers, it needs to support rolling upgrade, so it needs to implement "FixedDataSerializableID" interface, that's why we have to have it extends RestoreRedundancyResultsImpl and implement the additional interfaces in geode-core.

geode-core/src/main/java/org/apache/geode/management/internal/functions/RestoreRedundancyFunction.java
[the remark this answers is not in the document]
jinmeiliao: Generally `SerializableRestoreRedundancyResultsImpl` is used as the data object that's flowing from server to locator in order to support rolling upgrade. It would be nice that the result of the function call are all in that type.

geode-management/src/main/java/org/apache/geode/management/runtime/RestoreRedundancyResults.java
```diff
@@ -18,28 +18,28 @@
...
    * {@link #SUCCESS} is defined as every included region having fully satisfied redundancy.
    * {@link #FAILURE} is defined as at least one region that is configured to have redundant copies
    * having fewer than its configured number of redundant copies.
-   * {@link #ERROR} is for cases when the restore redundancy operation was unable to begin or threw
-   * an exception.
    */
   enum Status {
     SUCCESS,
-    FAILURE,
-    ERROR
```
jinmeiliao, line 36: @DonalEvans Please take a look at this change. We get rid of the ERROR status here because we found it's never used. No one is able to set the status to that state, and the getter of it would only return either SUCCESS or FAILURE. Even on develop branch, this state seems unreachable.

DonalEvans: You're correct, this was originally intended to be a way to capture the case that an exception was thrown during the  `doRestoreRedundancy()` method in `RestoreRedundancyOperationImpl`, but in order to keep the implementation as similar to that found in `RebalanceOperationImpl` as possible, this was not implemented. It should be fine to remove it.

PR #5257, opened 2020-06-16
GEODE-8251: make sure Configuration can be deserialized post 1.12.

jinmeiliao:
Co-authoried-by: Alberto Gomez <alberto.gomez@est.tech>

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/upgradeTest/java/org/apache/geode/internal/cache/rollingupgrade/RollingUpgrade2DUnitTestBase.java
+++ b/geode-core/src/upgradeTest/java/org/apache/geode/internal/cache/rollingupgrade/RollingUpgrade2DUnitTestBase.java
@@ -1185,6 +1185,8 @@ public abstract class RollingUpgrade2DUnitTestBase extends JUnit4DistributedTest
     props.setProperty(DistributionConfig.LOCATORS_NAME, locatorsString);
     props.setProperty(DistributionConfig.LOG_LEVEL_NAME, DUnitLauncher.logLevel);
     props.setProperty(DistributionConfig.ENABLE_CLUSTER_CONFIGURATION_NAME, enableCC + "");
+    // do not start http service to avoid port conflict between upgrade tests
+    props.setProperty(DistributionConfig.HTTP_SERVICE_PORT_NAME, "0");
     return props;
   }
 
--- a/geode-core/src/upgradeTest/java/org/apache/geode/internal/cache/rollingupgrade/RollingUpgradeWithGfshDUnitTest.java
+++ b/geode-core/src/upgradeTest/java/org/apache/geode/internal/cache/rollingupgrade/RollingUpgradeWithGfshDUnitTest.java
@@ -14,21 +14,6 @@
  */
 package org.apache.geode.internal.cache.rollingupgrade;
 
-/*
- * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
- * agreements. See the NOTICE file distributed with this work for additional information regarding
- * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
- * "License"); you may not use this file except in compliance with the License. You may obtain a
- * copy of the License at
- *
- * http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software distributed under the License
- * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
- * or implied. See the License for the specific language governing permissions and limitations under
- * the License.
- */
-
 import static org.assertj.core.api.Assertions.assertThat;
 
 import java.io.File;
@@ -139,14 +124,14 @@ public class RollingUpgradeWithGfshDUnitTest {
     return "deploy --dir=" + jarsDir.getAbsolutePath();
   }
 
-  protected String startServerCommand(String name, int port, int connectedLocatorPort) {
+  private String startServerCommand(String name, int port, int connectedLocatorPort) {
     String command = "start server --name=" + name
         + " --server-port=" + port
         + " --locators=localhost[" + connectedLocatorPort + "]";
     return command;
   }
 
-  protected String startLocatorCommand(String name, int port, int jmxPort,
+  private String startLocatorCommand(String name, int port, int jmxPort,
       int connectedLocatorPort) {
     String command = "start locator --name=" + name
         + " --port=" + port
--- a/geode-core/src/upgradeTest/java/org/apache/geode/internal/cache/rollingupgrade/RollingUpgradeWithGfshDUnitTest.java
+++ b/geode-core/src/upgradeTest/java/org/apache/geode/internal/cache/rollingupgrade/RollingUpgradeWithGfshDUnitTest.java
@@ -1,0 +1,160 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.apache.geode.internal.cache.rollingupgrade;
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+import static org.assertj.core.api.Assertions.assertThat;
+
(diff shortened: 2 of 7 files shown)


jinmeiliao:
@albertogpz

geode-core/src/upgradeTest/java/org/apache/geode/internal/cache/rollingupgrade/RollingUpgradeWithGfshDUnitTest.java
```diff
@@ -0,0 +1,160 @@
...
+@RunWith(Parameterized.class)
+@Parameterized.UseParametersRunnerFactory(CategoryWithParameterizedRunnerFactory.class)
+public class RollingUpgradeWithGfshDUnitTest {
+  private final UniquePortSupplier portSupplier = new UniquePortSupplier();
+  private final String oldVersion;
+
+  @Parameterized.Parameters(name = "{0}")
+  public static Collection<String> data() {
+    List<String> result = VersionManager.getInstance().getVersionsWithoutCurrent();
+    result.removeIf(s -> TestVersion.compare(s, "1.10.0") < 0);
```
albertogpz, line 70: Why start from this version? Is the upgrade from previous versions not supported?

jinmeiliao: There is no change in the the Configuration object except in 1.12, so only to test a few version before it and after it. Start from 1.10 just to save a few test cycles, but we can test the previous ones.

jinmeiliao:
> What's the point for these changes?
> I have ran the test case without them and it passes.
> Also, I have searched in the Geode code for the use of this file and have only found it in `ObjectInputStreamFilterWrapper.java`. From what I see there, only the first part of every line, the class name, is used.
> What am I missing?

the previous version of Deployment has variable name "jarFileName" instead of "fileName", when we are deserializing a previous version bytes, if the variable name doesn't match, it won't take the value. We should have a test to verify that, but I saw the behavior in the debugger.

jinmeiliao:
@albertogpz the pipeline tests failed because we added GEODE_HOME environment variable in all tests environment, some tests are able to find the war files thus starting the web services, thus getting the address bind exception. Is there any way we can add that environment variable only to one test?

albertogpz:
> > What's the point for these changes?
> > I have ran the test case without them and it passes.
> > Also, I have searched in the Geode code for the use of this file and have only found it in `ObjectInputStreamFilterWrapper.java`. From what I see there, only the first part of every line, the class name, is used.
> > What am I missing?
> 
> the previous version of Deployment has variable name "jarFileName" instead of "fileName", when we are deserializing a previous version bytes, if the variable name doesn't match, it won't take the value. We should have a test to verify that, but I saw the behavior in the debugger.

I understand that. What I do not get are the changes in the 

> > What's the point for these changes?
> > I have ran the test case without them and it passes.
> > Also, I have searched in the Geode code for the use of this file and have only found it in `ObjectInputStreamFilterWrapper.java`. From what I see there, only the first part of every line, the class name, is used.
> > What am I missing?
> 
> the previous version of Deployment has variable name "jarFileName" instead of "fileName", when we are deserializing a previous version bytes, if the variable name doesn't match, it won't take the value. We should have a test to verify that, but I saw the behavior in the debugger.

I understand the change in the members of the class. What I do not get is why it needs to be reflected in sanctioned-geode-management-serializables.txt.
Who is accessing that file?

jinmeiliao:
> > > What's the point for these changes?
> > > I have ran the test case without them and it passes.
> > > Also, I have searched in the Geode code for the use of this file and have only found it in `ObjectInputStreamFilterWrapper.java`. From what I see there, only the first part of every line, the class name, is used.
> > > What am I missing?
> > 
> > 
> > the previous version of Deployment has variable name "jarFileName" instead of "fileName", when we are deserializing a previous version bytes, if the variable name doesn't match, it won't take the value. We should have a test to verify that, but I saw the behavior in the debugger.
> 
> I understand that. What I do not get are the changes in the
> 
> > > What's the point for these changes?
> > > I have ran the test case without them and it passes.
> > > Also, I have searched in the Geode code for the use of this file and have only found it in `ObjectInputStreamFilterWrapper.java`. From what I see there, only the first part of every line, the class name, is used.
> > > What am I missing?
> > 
> > 
> > the previous version of Deployment has variable name "jarFileName" instead of "fileName", when we are deserializing a previous version bytes, if the variable name doesn't match, it won't take the value. We should have a test to verify that, but I saw the behavior in the debugger.
> 
> I understand the change in the members of the class. What I do not get is why it needs to be reflected in sanctioned-geode-management-serializables.txt.
> Who is accessing that file?

the `AnalyzeManagementSerializablesJUnitTest` is making sure that we don't make random changes in those serialized classes. Every change in the sanction list poses a potential upgrade issue. So the fact that the test is broken will make the developer think twice about the changes he/she is making to the class.

albertogpz:
> > > > What's the point for these changes?
> > > > I have ran the test case without them and it passes.
> > > > Also, I have searched in the Geode code for the use of this file and have only found it in `ObjectInputStreamFilterWrapper.java`. From what I see there, only the first part of every line, the class name, is used.
> > > > What am I missing?
> > > 
> > > 
> > > the previous version of Deployment has variable name "jarFileName" instead of "fileName", when we are deserializing a previous version bytes, if the variable name doesn't match, it won't take the value. We should have a test to verify that, but I saw the behavior in the debugger.
> > 
> > 
> > I understand that. What I do not get are the changes in the
> > > > What's the point for these changes?
> > > > I have ran the test case without them and it passes.
> > > > Also, I have searched in the Geode code for the use of this file and have only found it in `ObjectInputStreamFilterWrapper.java`. From what I see there, only the first part of every line, the class name, is used.
> > > > What am I missing?
> > > 
> > > 
> > > the previous version of Deployment has variable name "jarFileName" instead of "fileName", when we are deserializing a previous version bytes, if the variable name doesn't match, it won't take the value. We should have a test to verify that, but I saw the behavior in the debugger.
> > 
> > 
> > I understand the change in the members of the class. What I do not get is why it needs to be reflected in sanctioned-geode-management-serializables.txt.
> > Who is accessing that file?
> 
> the `AnalyzeManagementSerializablesJUnitTest` is making sure that we don't make random changes in those serialized classes. Every change in the sanction list poses a potential upgrade issue. So the fact that the test is broken will make the developer think twice about the changes he/she is making to the class.

I see now. Thanks for the clarification.

geode-core/src/upgradeTest/java/org/apache/geode/internal/cache/rollingupgrade/RollingUpgradeWithGfshDUnitTest.java
```diff
@@ -0,0 +1,160 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.apache.geode.internal.cache.rollingupgrade;
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
```
kirklund, line 18: You accidentally have the license header in this class twice.

geode-core/src/upgradeTest/java/org/apache/geode/internal/cache/rollingupgrade/RollingUpgradeWithGfshDUnitTest.java
```diff
@@ -0,0 +1,160 @@
...
+    ClassBuilder classBuilder = new ClassBuilder();
+    File jarsDir = tempFolder.newFolder();
+    String jarName1 = "DeployCommandsDUnit1.jar";
+    File jar1 = new File(jarsDir, jarName1);
+    String class1 = "DeployCommandsDUnitA";
+    classBuilder.writeJarFromName(class1, jar1);
+    return "deploy --dir=" + jarsDir.getAbsolutePath();
+  }
+
+  protected String startServerCommand(String name, int port, int connectedLocatorPort) {
```
kirklund, line 142: Is there a subclass? I'd like to see these `protected` methods be `private` if possible.

geode-management/src/main/java/org/apache/geode/management/configuration/AbstractConfiguration.java
```diff
@@ -47,7 +47,7 @@
 @Experimental
 public abstract class AbstractConfiguration<R extends RuntimeInfo>
     implements Identifiable<String>, JsonSerializable {
-
+  private static final long serialVersionUID = -6612840641128145954L;
```
kirklund, line 50: Optional: all JDK classes put serialVersionUID at the bottom of the class. I usually do this to try and be consistent but obviously doesn't matter much.

geode-core/src/upgradeTest/java/org/apache/geode/internal/cache/rollingupgrade/RollingUpgradeWithGfshDUnitTest.java
```diff
@@ -0,0 +1,145 @@
...
+    verifyListDeployed(locatorPort);
+
+    oldGfsh.stopLocator(startupExecution, "loc2");
+    GfshScript.of(startLocatorCommand("loc2", locator2Port, locator2JmxPort, locatorPort))
+        .execute(currentGfsh);
+    verifyListDeployed(locator2Port);
+
+    // make sure servers can do rolling upgrade too
+    oldGfsh.stopServer(startupExecution, "server1");
+    GfshScript.of(startServerCommand("server1", server1Port, locatorPort)).execute(currentGfsh);
```
jchen21, line 103: Better `verifyListDeployed` after rolling upgrade the servers.

geode-core/src/upgradeTest/java/org/apache/geode/internal/cache/rollingupgrade/RollingUpgrade2DUnitTestBase.java
```diff
@@ -1185,6 +1185,8 @@ Properties getLocatorProperties(String locatorsString, boolean enableCC) {
     props.setProperty(DistributionConfig.LOCATORS_NAME, locatorsString);
     props.setProperty(DistributionConfig.LOG_LEVEL_NAME, DUnitLauncher.logLevel);
     props.setProperty(DistributionConfig.ENABLE_CLUSTER_CONFIGURATION_NAME, enableCC + "");
+    // do not start http service to avoid port conflict between upgrade tests
+    props.setProperty(DistributionConfig.HTTP_SERVICE_PORT_NAME, "0");
```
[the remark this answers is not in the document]
jinmeiliao: here 0 means do not start.

PR #5261, opened 2020-06-17
first full cut of REST restore redundancy command 

mhansonp:
Checking for unexpected failures...


mhansonp:
Changing to a new branch.

PR #5269, opened 2020-06-17
GEODE-8240: first try at a fix—not good

Bill:
[GEODE-8240](https://issues.apache.org/jira/browse/GEODE-8240)

**Not ready for review**

This was a first attempt at a fix for GEODE-8240. It'll be abandoned after the ultimate fix (that merged to `develop` on 6/30) makes it through the CI pipeline. Bear with me for a little longer please.

This is a minimalist approach to trying to fix the bug, by allowing `GMSMemberData` to hold unknown (ordinal) product versions received through serialization. Rather than having `GMSMemberData` store a `Version` it stores a `short` to represent the ordinal version. When a caller needs the `Version` it can get it from get it.

A problem arises when a caller needs to recover a `Version` from the ordinal (`short1`) though. It leaves us with two bad options when that ordinal value represents an unknown product version:

1. collapse all unknown product versions into `Version.CURRENT` (silently)
2. allow `UnsupportedSerializationVersionException` to escape `getVersion()`

The former (1) is another version of the very bug we seek to fix. We have just moved it from the set-side (eager) to the get side (lazy). The latter (2) is bad because some callers actually might want (like `GMSMemberData`) to be able to handle unknown product versions. That exception does not differentiate between impossible product version ordinals e.g. -5 and possible-but-unknown ones e.g. 12467.

A separate PR: https://github.com/apache/geode/pull/5273 seeks to resolve the mess by introducing a new type: `VersionOrdinal`.

### For all changes:

### DIFF
--- a/geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSMemberData.java
+++ b/geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSMemberData.java
@@ -79,7 +79,14 @@ public class GMSMemberData implements MemberData, Comparable<GMSMemberData> {
    * displayable.
    */
   private String uniqueTag = null;
-  private transient Version versionObj = Version.CURRENT;
+
+  /**
+   * versionOrdinal is stored here as an ordinal and not a Version proper, because
+   * GMSMemberData needs to sometimes store the version of a new product version,
+   * e.g. during rolling upgrade members with old versions receive member identifiers
+   * from members with new (unknown) versions.
+   */
+  private transient short versionOrdinal = Version.CURRENT.ordinal();
 
   /**
    * whether this is a partial member ID (without roles, durable attributes). We use partial IDs in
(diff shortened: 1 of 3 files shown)


geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSMemberData.java
```diff
@@ -220,12 +220,16 @@ public InetAddress getInetAddress() {
...
-    return this.versionObj.ordinal();
+    return versionOrdinal;
   }
 
   @Override
   public Version getVersion() {
-    return versionObj;
+    try {
+      return Version.fromOrdinal(versionOrdinal);
+    } catch (final UnsupportedSerializationVersionException e) {
```
Bill, line 230: **This right here** is a big fat problem. We are silently throwing away information when the version ordinal represents an unknown version. One approach is to let the exception escape (as a checked exception.) But that entails fixing the 53 callers of this method. An alternative would be to "tunnel" the checked exception through an unchecked one.

Bill:
I didn't end up using this PR—I took a different approach. Closing this one because it's not needed.

PR #5224, opened 2020-06-08
DRAFT: Modify upgrade test case to manifest problem after 1.12

albertogpz:
Thank you for submitting a contribution to Apache Geode.

In order to streamline the review of the contribution we ask you
to ensure the following steps have been taken:

### For all changes:
- [ ] Is there a JIRA ticket associated with this PR? Is it referenced in the commit message?

- [ ] Has your PR been rebased against the latest commit within the target branch (typically `develop`)?

- [ ] Is your initial contribution a single, squashed commit?

- [ ] Does `gradlew build` run cleanly?

- [ ] Have you written or updated unit tests to verify your changes?

- [ ] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)?

### Note:
Please ensure that once the PR is submitted, check Concourse for build issues and
submit an update to your PR as soon as possible. If you need help, please send an
email to dev@geode.apache.org.

PR #5273, opened 2020-06-18
GEODE-8240: solve rolling upgrade bug with new VersionOrdinal interface

Bill:
[GEODE-8240](https://issues.apache.org/jira/browse/GEODE-8240)

This is an alternative solution to the one explored in https://github.com/apache/geode/pull/5269. This PR introduces a new interface `VersionOrdinal` to model not just versions we know about (as `Version` does) but also future versions. Formerly Geode modeled these as `short` but that presented difficulties in the many places where we need to compare versions. By unifying known and unknown versions under one interface, capable of providing generic comparison (`compareTo()`, `equals()`, and `hashCode()`) my hope is that we can eliminate this egregious loss of information (in that other PR's `Version` implementation):

```java
  @Override
  public Version getVersion() {
    try {
      return Version.fromOrdinal(versionOrdinal);
    } catch (final UnsupportedSerializationVersionException e) {
      return Version.CURRENT;
    } 
```

That code right there is essentially a mirror-image of the `GMSMemberData.setVersionObject()` method which is the original culprit in [GEODE-8240](https://issues.apache.org/jira/browse/GEODE-8240). Rather than moving the wrongness from the eager/set side to the lazy/get side, this PR seeks to eliminate the wrongness entirely.

## Highlights

Introduced new `VersionOrdinal` type (interface). It carries all the equality, comparison, hash code functionality formerly down in `Version`. Made a `VersionOrdinalImpl` class to implement that interface and for use in places where we want an instance of just `VersionOrdinal` without the restrictions in `Version` (mainly that we can only have instances of known versions.)

`GMSMemberData`, which is the state holder for a `MemberIdentifier` now has a `VersionOrdinal` where it used to have a `Version`. `getVersion()` has been renamed `getVersionOrdinalObject()` and now returns a `VersionOrdinal` where it used to return a `Version`. Since _all_ the callers needed only comparison ops on that return value, the changes to call sites were trivial, if widespread. (Most of the files affected by this PR have those call sites and in every case the change was merely to change a method parameter, field, or local variable declaration from `Version` to `VersionOrdinal`.

Also since `Version` isa `VersionOrdinal` they are cross-comparable.

I think this new `VersionOrdinal` type is more in the spirit of the old code (from before the bug was introduced.) That code carried `short` version ordinals around. The old `GMSMember` class (which later became `GMSMembeData`) had a `short versionOrdinal` field. This new `VersionOrdinal` class does that one better by giving us a place to hang the comparsion and hashing ops and gives us a type/subtype relationship between `VersionOrdinal` (any version ordinal, known or unknown) and `Version` (a known version.)

It would be nice if, when you asked for a `VersionOrdinal`, if that represented a known version, i.e. a `Version`, that you'd actually get the pertinent `Version` object back. A factory, `Versioning.getVersionOrdinal(short ordinal)` was introduced fore this purpose. Only that factory is allowed to construct `VersionOrdinalImpl`. As was the case before this PR, `Version` has no public constructor at all—it is essentially an enumerated type.

## Testing

The original PR that pointed out this bug https://github.com/apache/geode/pull/5224 enhanced `RollingUpgradeDUnitTest` to check the versions (of each member) in the view reported by gfsh after a rolling upgrade. The fix to `GMSMemberData` was backported to a feature branch off of `support/1.12` and the modified `RollingUpgradeDUnitTest` passed ✓

A new unit test `GMSMemberDataVersionDUnitTest` was added to verify that it's now possible to set an unknown version (ordinal) on `GMSMemberData` without loss of information.

So much for the bug itself. But what about 1.12.0 coordinators running in the wild with the bug? 

During an upgrade a locator will receive a bad view from one of those. "Compensating" code was added to `GMSJoinLeave.installView()` to patch up bad member version data arriving in a view from a compromised coordinator. See the call to `GMSMembershipView.correctWrongVersionIn(final ID memberID)` in `GMSJoinLeave.installView()`.

A new unit test, `GMSJoinLeaveJUnitTest.repairWrongVersionInView()` was added to test the compensation logic. As an additional test, the new `RollingUpgradeDUnitTest` was verified passing for all old Geode versions, including an un-repaired 1.12.0.

## Miscellaneous Things I Hope You'll Check Me On

* `sanctionedDataSerializables.txt` in `geode-core` changed. It's odd to me that there were apparently a bunch of `geode-membership` classes in that file before this PR. I don't think they should have been there. Nevertheless it's a mystery as to why they were and I'd like another opinion on that file. Also please double-check the (non-deletion) changes to that file—I checked 'em but they need a double-check.

## For all changes:

### DIFF
--- a/geode-core/src/integrationTest/resources/org/apache/geode/codeAnalysis/sanctionedDataSerializables.txt
+++ b/geode-core/src/integrationTest/resources/org/apache/geode/codeAnalysis/sanctionedDataSerializables.txt
@@ -282,89 +282,13 @@ toData,12
 toDataPre_GFE_7_1_0_0,12
 toDataPre_GFE_9_0_0_0,12
 
-org/apache/geode/distributed/internal/membership/adapter/LocalViewMessage,2
-fromData,8
-toData,8
-
-org/apache/geode/distributed/internal/membership/gms/GMSMembershipView,2
-fromData,121
-toData,72
-
-org/apache/geode/distributed/internal/membership/gms/locator/FindCoordinatorRequest,2
-fromData,112
-toData,132
-
-org/apache/geode/distributed/internal/membership/gms/locator/FindCoordinatorResponse,2
-fromData,130
-toData,114
-
-org/apache/geode/distributed/internal/membership/gms/locator/GetViewRequest,2
-fromData,1
-toData,1
-
-org/apache/geode/distributed/internal/membership/gms/locator/GetViewResponse,2
-fromData,20
-toData,17
-
-org/apache/geode/distributed/internal/membership/gms/messages/FinalCheckPassedMessage,2
-fromData,20
-toData,17
-
-org/apache/geode/distributed/internal/membership/gms/messages/HeartbeatMessage,2
-fromData,11
-toData,11
-
-org/apache/geode/distributed/internal/membership/gms/messages/HeartbeatRequestMessage,2
-fromData,30
-toData,27
-
-org/apache/geode/distributed/internal/membership/gms/messages/InstallViewMessage,2
-fromData,60
-toData,56
-
-org/apache/geode/distributed/internal/membership/gms/messages/JoinRequestMessage,2
-fromData,63
-toData,60
-
-org/apache/geode/distributed/internal/membership/gms/messages/JoinResponseMessage,2
-fromData,63
-toData,57
-
-org/apache/geode/distributed/internal/membership/gms/messages/LeaveRequestMessage,2
-fromData,28
-toData,25
-
-org/apache/geode/distributed/internal/membership/gms/messages/NetworkPartitionMessage,2
-fromData,1
-toData,1
-
-org/apache/geode/distributed/internal/membership/gms/messages/RemoveMemberMessage,2
-fromData,28
-toData,25
-
-org/apache/geode/distributed/internal/membership/gms/messages/SuspectMembersMessage,2
-fromData,63
-toData,92
-
-org/apache/geode/distributed/internal/membership/gms/messages/ViewAckMessage,2
-fromData,40
-toData,37
-
 org/apache/geode/distributed/internal/streaming/StreamingOperation$RequestStreamingMessage,2
 fromData,17
 toData,17
 
 org/apache/geode/distributed/internal/streaming/StreamingOperation$StreamingReplyMessage,2
-fromData,417
+fromData,419
 toData,86
-
-org/apache/geode/distributed/internal/tcpserver/InfoResponse,2
-fromData,9
-toData,9
-
-org/apache/geode/distributed/internal/tcpserver/VersionResponse,2
-fromData,11
-toData,11
 
 org/apache/geode/internal/DSFIDFactory,2
 fromData,8
@@ -577,9 +501,6 @@ toData,7
 org/apache/geode/internal/admin/remote/LicenseInfoResponse,2
 fromData,18
 toData,15
-
-org/apache/geode/internal/admin/remote/MissingPersistentIDsRequest,1
-fromData,7
 
 org/apache/geode/internal/admin/remote/MissingPersistentIDsResponse,2
 fromData,129
@@ -981,7 +902,7 @@ toData,22
 org/apache/geode/internal/cache/EventID,4
 fromData,53
 fromDataPre_GFE_8_0_0_0,33
-toData,106
+toData,108
 toDataPre_GFE_8_0_0_0,24
 
 org/apache/geode/internal/cache/EvictionAttributesImpl,2
@@ -1274,8 +1195,8 @@ fromData,38
 toData,35
 
 org/apache/geode/internal/cache/TXRegionLockRequestImpl,2
-fromData,97
-toData,55
+fromData,99
+toData,57
 
 org/apache/geode/internal/cache/TXRemoteCommitMessage$TXRemoteCommitReplyMessage,2
 fromData,18
@@ -1376,8 +1297,8 @@ fromData,22
 toData,19
 
 org/apache/geode/internal/cache/execute/FunctionRemoteContext,2
-fromData,124
-toData,102
+fromData,126
+toData,104
 
 org/apache/geode/internal/cache/ha/HARegionQueue$DispatchedAndCurrentEvents,2
 fromData,37
@@ -1636,8 +1557,8 @@ fromData,74
 toData,68
 
 org/apache/geode/internal/cache/partitioned/PartitionMessage,2
-fromData,60
-toData,113
+fromData,62
+toData,115
 
 org/apache/geode/internal/cache/partitioned/PartitionedRegionFunctionStreamingMessage,2
 fromData,18
@@ -1680,7 +1601,7 @@ fromData,54
 toData,35
 
 org/apache/geode/internal/cache/partitioned/RemoveAllPRMessage,2
-fromData,194
+fromData,187
 toData,201
 
 org/apache/geode/internal/cache/partitioned/RemoveAllPRMessage$RemoveAllReplyMessage,2
@@ -1978,13 +1899,13 @@ fromData,214
 toData,245
 
 org/apache/geode/internal/cache/versions/VersionTag,2
-fromData,225
+fromData,227
 toData,254
 
 org/apache/geode/internal/cache/wan/GatewaySenderAdvisor$GatewaySenderProfile,4
-fromData,283
+fromData,285
 fromDataPre_GFE_8_0_0_0,188
-toData,271
+toData,273
 toDataPre_GFE_8_0_0_0,236
 
 org/apache/geode/internal/cache/wan/GatewaySenderEventCallbackArgument,2
@@ -2103,9 +2024,9 @@ toData,25
 
 org/apache/geode/pdx/internal/PdxField,2
 fromData,99
-toData,105
+toData,107
 
 org/apache/geode/pdx/internal/PdxType,2
 fromData,109
-toData,124
+toData,126
 
--- a/geode-core/src/integrationTest/resources/org/apache/geode/codeAnalysis/sanctionedDataSerializables.txt
+++ b/geode-core/src/integrationTest/resources/org/apache/geode/codeAnalysis/sanctionedDataSerializables.txt
@@ -287,7 +287,7 @@ fromData,17
 toData,17
 
 org/apache/geode/distributed/internal/streaming/StreamingOperation$StreamingReplyMessage,2
-fromData,419
+fromData,417
 toData,86
 
 org/apache/geode/internal/DSFIDFactory,2
@@ -872,7 +872,7 @@ org/apache/geode/internal/cache/DistributedPutAllOperation$PutAllEntryData,1
 toData,252
 
 org/apache/geode/internal/cache/DistributedPutAllOperation$PutAllMessage,2
-fromData,217
+fromData,214
 toData,188
 
 org/apache/geode/internal/cache/DistributedRegionFunctionStreamingMessage,2
@@ -880,7 +880,7 @@ fromData,177
 toData,174
 
 org/apache/geode/internal/cache/DistributedRemoveAllOperation$RemoveAllMessage,2
-fromData,217
+fromData,194
 toData,188
 
 org/apache/geode/internal/cache/DistributedTombstoneOperation$TombstoneMessage,2
@@ -902,7 +902,7 @@ toData,22
 org/apache/geode/internal/cache/EventID,4
 fromData,53
 fromDataPre_GFE_8_0_0_0,33
-toData,108
+toData,106
 toDataPre_GFE_8_0_0_0,24
 
 org/apache/geode/internal/cache/EvictionAttributesImpl,2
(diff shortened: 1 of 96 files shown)


geode-serialization/src/main/java/org/apache/geode/internal/serialization/internal/DSFIDSerializerImpl.java
```diff
@@ -220,20 +220,6 @@ public void invokeToData(Object ds, DataOutput out) throws IOException {
...
-   * zero if the version is same as this member's.
-   */
-  public Version getVersionForDataStreamOrNull(DataOutput out) {
-    // check if this is a versioned data output
-    if (out instanceof VersionedDataStream) {
-      return ((VersionedDataStream) out).getVersion();
-    } else {
-      return null;
-    }
-  }
```
Bill, line 235: dead code

geode-serialization/src/main/java/org/apache/geode/internal/serialization/Version.java
```diff
@@ -635,48 +569,4 @@ public boolean isPre65() {
         .collect(Collectors.toList());
   }
 
-  public boolean isCurrentVersion() {
-    return this.ordinal == CURRENT.ordinal;
-  }
```
Bill, line 640: this was used in exactly two places and is better accomplished via `Version.CURRENT.isEqual(someVersionOrdinal)`

geode-core/src/integrationTest/resources/org/apache/geode/codeAnalysis/sanctionedDataSerializables.txt
```diff
@@ -578,9 +502,6 @@ org/apache/geode/internal/admin/remote/LicenseInfoResponse,2
 fromData,18
 toData,15
 
-org/apache/geode/internal/admin/remote/MissingPersistentIDsRequest,1
```
bschuchardt, line 581: this removal of MissingPersistentIDsRequest surprised me.  The class still exists in my checkout and is a DSFID.

geode-core/src/main/java/org/apache/geode/cache/client/internal/ClientSideHandshakeImpl.java
```diff
@@ -270,7 +271,7 @@ public ServerQueueStatus handshakeWithServer(Connection conn, ServerLocation loc
   private InternalDistributedMember readServerMember(DataInputStream p_dis) throws IOException {
 
     byte[] memberBytes = DataSerializer.readByteArray(p_dis);
-    Version v = StaticSerialization.getVersionForDataStreamOrNull(p_dis);
+    final VersionOrdinal v = StaticSerialization.getVersionForDataStreamOrNull(p_dis);
```
bschuchardt, line 274: This should continue to use Version, not VersionOrdinal.  Using Version ensures that we never create a datastream with an unknown ordinal.  None of our serialization code will know what to do if that should happen.

geode-core/src/main/java/org/apache/geode/distributed/internal/membership/InternalDistributedMember.java
```diff
@@ -549,7 +550,7 @@ public void setVersionObjectForTest(Version v) {
   }
 
   @Override
-  public Version getVersionObject() {
+  public VersionOrdinal getVersionObject() {
```
bschuchardt, line 553: This should continue to return a Version, not a VersionOrdinal.  We already have a method for getting the ordinal.

geode-core/src/main/java/org/apache/geode/internal/HeapDataOutputStream.java
```diff
@@ -48,7 +49,7 @@
     org.apache.geode.internal.serialization.BufferDataOutputStream
     implements ObjToByteArraySerializer, ByteBufferWriter {
 
-  public HeapDataOutputStream(Version version) {
+  public HeapDataOutputStream(VersionOrdinal version) {
```
bschuchardt, line 52: same comment about retaining the use of Version in all serialization code.  We need assurance in place that we will never encounter an unknown ordinal during serialization/deserialization.

Bill:
> Is there any reason for this commented out line to be left here=

@albertogpz for some reason I'm not able to request you as a reviewer. Thanks for the comments. Keep them coming!

Allowing all the old versions again revealed that pre-1.9.0 versions are failing to upgrade with these changes. Once I've fixed that I'll push a new commit with your comments addressed.

geode-serialization/src/main/java/org/apache/geode/internal/serialization/VersionOrdinalImpl.java
```diff
@@ -0,0 +1,132 @@
...
+  public String toString() {
+    return toString(ordinal);
+  }
+
+  /**
+   * TODO: eliminate this legacy method in favor of requiring callers to construct a
+   * VersionOrdinalImpl. Inline this logic up in toString().
+   */
+  public static String toString(short ordinal) {
+    return "UNKNOWN[ordinal=" + ordinal + ']';
```
bschuchardt, line 84: could this find the name of a Version and use it?  Alternatively just use "VersionOrdinal" instead of "UNKNOWN"

geode-core/src/main/java/org/apache/geode/distributed/internal/membership/InternalDistributedMember.java
Bill: Since `IDM` isa `MemberIdentifier` this method has to return a `VersionOrdinal` not a `Version` since, in general, a `MemberIdentifier` can be for a member that is running an unknown software version.

I'm renaming `MemberIdentifier.getVersionObject()` => `MemberIdentifier.getVersionOrdinalObject()` and adding a method `IDM.getVersionObject()` to really return the version for callers that really need it.

What's odd about the latter is that I don't see how that method could/should ever return anything but `Version.getCurrentVersion()` since the version, in this case, should be the version of the software running in the local JVM—which is always `Version.getCurrentVersion()`.

geode-core/src/integrationTest/resources/org/apache/geode/codeAnalysis/sanctionedDataSerializables.txt
Bill: I re-ran `./gradlew :geode-core:integrationTest --tests=AnalyzeCoreSerializablesJUnitTest` and `sanctionedDataSerializables.txt` did not change. `MissingPersistentIDsRequest` is still not listed there. I see it ultimately derives from `DSFID`.

bschuchardt: I see why - that class has no fromData/toData methods.  The test just searches for those in parsed class files.  It doesn't care if they're DataSerializable/DSFID.

geode-serialization/src/main/java/org/apache/geode/internal/serialization/VersionOrdinalImpl.java
```diff
@@ -0,0 +1,132 @@
...
+  public int compareTo(final short other) {
+    // short min/max can't overflow int, so use (a-b)
+    final int thisOrdinal = this.ordinal;
+    final int otherOrdinal = other;
+    return thisOrdinal - otherOrdinal;
+  }
+
+  @Override
+  public boolean equals(final Object other) {
+    if (other == this)
```
luissson, line 53: we should have a null check on 'other' here

geode-serialization/src/main/java/org/apache/geode/internal/serialization/internal/AbstractSerializationContext.java
```diff
@@ -25,10 +26,10 @@
  */
 public abstract class AbstractSerializationContext {
 
-  <IO> Version getVersionForDataStream(final IO in) {
+  <IO> VersionOrdinal getVersionForDataStream(final IO in) {
     // check if this is a versioned data input
     if (in instanceof VersionedDataStream) {
-      final Version v = ((VersionedDataStream) in).getVersion();
+      final VersionOrdinal v = ((VersionedDataStream) in).getVersion();
       return v != null ? v : Version.getCurrentVersion();
```
luissson, line 33: the return here may be either Version or VersionOrdinal, do we want to upcast or downcast to ensure returning a single type? Doesn't appear to be a problem currently though

geode-core/src/integrationTest/resources/org/apache/geode/codeAnalysis/sanctionedDataSerializables.txt
Bill: I wonder if it used to have fromData/toData and now does not? why else would it have showed up in this file previously?

geode-core/src/main/java/org/apache/geode/cache/client/internal/ClientSideHandshakeImpl.java
Bill: fixed

geode-core/src/main/java/org/apache/geode/internal/HeapDataOutputStream.java
Bill: fixed

geode-serialization/src/main/java/org/apache/geode/internal/serialization/VersionOrdinalImpl.java
Bill: fixed

geode-serialization/src/main/java/org/apache/geode/internal/serialization/internal/AbstractSerializationContext.java
Bill: "fixed" this code by reverting it to deal exclusively with `Version` as before the PR

geode-serialization/src/main/java/org/apache/geode/internal/serialization/VersionOrdinalImpl.java
Bill: I think we are ok. If `other` is `null` then it cannot be equal to `this`.

if `other` is `null` the `instanceof` check will return `false`

geode-core/src/upgradeTest/java/org/apache/geode/internal/cache/rollingupgrade/RollingUpgradeDUnitTest.java
```diff
@@ -191,20 +207,26 @@ void doTestRollAll(String regionType, String objectType, String startingVersion)
...
       putAndVerify(objectType, server1, regionName, 5, 15, server2);
       putAndVerify(objectType, server2, regionName, 10, 20, server1);
 
       server2 = rollServerToCurrentAndCreateRegion(server2, regionType, testingDirs[1],
-          shortcutName, regionName, locatorPorts);
+          shortcutName, regionName, ports);
       verifyValues(objectType, regionName, 0, 10, server2);
       putAndVerify(objectType, server2, regionName, 15, 25, server1);
 
+      String shutDownCommand = "shutdown --include-locators=true";
```
albertogpz, line 225: The shutdown command had also issues. That's why the variable was there initially in the test case.
Could you add the execution of the shutdown command after the ```list members``` in order to verify that with your solution it works?
Otherwise, the variable must be removed.

Bill: My latest commit eliminates `gfsh` entirely so the issue with `shutDownCommand` is moot.

I eliminated use of `gfsh` because I found the test had become flaky on versions 1.3.0-1.8.0, at least on macOS. On Linux (CI) I did not see the flakiness. But rather than figuring that out, I decided to grab the view more directly: by accessing `InternalDistributedSystem` in the locator JVM.

geode-serialization/src/main/java/org/apache/geode/internal/serialization/VersionOrdinal.java
```diff
@@ -0,0 +1,83 @@
...
+ *
+ * Unlike Version (a subtype of VersionOrdinal which acts like an
+ * enumerated type), VersionOrdinal does not, in general, guarantee
+ * that if vo1.equals(vo2) then vo1 == vo2.
+ *
+ * Use the Versioning factory class to construct objects implementing
+ * this interface. All instances of known versions are defined as
+ * constants in the Version class, e.g. Version.GEODE_1_11_0
+ */
+public interface VersionOrdinal extends Comparable<VersionOrdinal> {
```
bschuchardt, line 39: VersionOrdinal's methods should take VersionOrdinal arguments.

Bill: Corrected this and went one better: added a new unit test for `VersionOrdinalImpl`.

geode-core/src/upgradeTest/java/org/apache/geode/internal/cache/rollingupgrade/RollingUpgradeDUnitTest.java
albertogpz: Overall, great work. Thanks, Bill.

PR #5282, opened 2020-06-22
GEODE-8288: Match Redis's glob-style pattern

sabbeyPivotal:
Redis has their own glob-style pattern (https://redis.io/commands/keys). Change our implementation to match that of Redis

### DIFF
--- a/geode-redis/src/main/java/org/apache/geode/redis/internal/executor/key/KeysExecutor.java
+++ b/geode-redis/src/main/java/org/apache/geode/redis/internal/executor/key/KeysExecutor.java
@@ -15,14 +15,15 @@
  */
 package org.apache.geode.redis.internal.executor.key;
 
-import static org.apache.geode.redis.internal.RedisConstants.ERROR_ILLEGAL_GLOB;
-
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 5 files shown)


geode-redis/src/main/java/org/apache/geode/redis/internal/executor/key/KeysExecutor.java
```diff
@@ -45,7 +47,10 @@ public RedisResponse executeCommand(Command command,
     try {
       pattern = GlobPattern.compile(glob);
     } catch (PatternSyntaxException e) {
-      return RedisResponse.error(ERROR_ILLEGAL_GLOB);
+      logger.debug(
```
dschneider-pivotal, line 50: Instead of debug, it seems like this should be a warning. Also the log message should mention that it was a pattern given to the KEYS command. Maybe the last sentence could be changed to "KEYS will return an empty list."
Also it seems like you should have some quotes around the curlies for the e.getMessage() and a period after the closing quote.

sabbeyPivotal: updated it!

PR #5285, opened 2020-06-23
GEODE-8099: make those gfsh commands that updates cluster configurati…

jinmeiliao:
…on thead safe.

* command executor will acquire the lock when executing commands that affects cluster configuration.
* clean up commands that doesn't need to extends implement SingleGfshCommand
* SingleGfshCommand are for those commands that need to update cluster configuration

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/management/internal/api/LocatorClusterManagementService.java
+++ b/geode-core/src/main/java/org/apache/geode/management/internal/api/LocatorClusterManagementService.java
@@ -113,7 +113,7 @@ import org.apache.geode.management.runtime.RuntimeInfo;
 public class LocatorClusterManagementService implements ClusterManagementService {
   @VisibleForTesting
   // the dlock service name used by the CMS
-  static final String CMS_DLOCK_SERVICE_NAME = "CMS_DLOCK_SERVICE";
+  public static final String CMS_DLOCK_SERVICE_NAME = "CMS_DLOCK_SERVICE";
   private static final Logger logger = LogService.getLogger();
   private final InternalConfigurationPersistenceService persistenceService;
   private final Map<Class, ConfigurationManager> managers;
--- a/geode-gfsh/src/main/java/org/apache/geode/management/cli/GfshCommand.java
+++ b/geode-gfsh/src/main/java/org/apache/geode/management/cli/GfshCommand.java
@@ -59,6 +59,14 @@ public abstract class GfshCommand implements CommandMarker {
     return gfsh.isConnectedAndReady();
   }
 
+  /**
+   * For those commands that could possibly change the cluster configuration, they need to
+   * override this method to return true.
+   */
+  public boolean affectsClusterConfiguration() {
+    return false;
+  }
+
   public void authorize(ResourcePermission.Resource resource,
       ResourcePermission.Operation operation, ResourcePermission.Target target) {
     cache.getSecurityService().authorize(resource, operation, target);
--- a/geode-gfsh/src/main/java/org/apache/geode/management/cli/SingleGfshCommand.java
+++ b/geode-gfsh/src/main/java/org/apache/geode/management/cli/SingleGfshCommand.java
@@ -22,7 +22,10 @@ import org.apache.geode.cache.configuration.CacheConfig;
 
 /**
  * Command class that extends this class can only have one single command method,
- * i.e only one method that is annotated with @CliCommand.
+ * * i.e only one method that is annotated with @CliCommand.
+ *
+ * this is also specific for commands that will need to update cluster configuration. Child classes
+ * are required to implement the "updateConfigForGroup" method.
  */
 @Experimental
 public abstract class SingleGfshCommand extends GfshCommand {
@@ -39,7 +42,11 @@ public abstract class SingleGfshCommand extends GfshCommand {
    *        return value of your command method.
    * @return a boolean indicating whether a change to the cluster configuration was persisted.
    */
-  public boolean updateConfigForGroup(String group, CacheConfig config, Object configObject) {
-    return false;
+  public abstract boolean updateConfigForGroup(String group, CacheConfig config,
+      Object configObject);
+
+  @Override
+  public boolean affectsClusterConfiguration() {
+    return true;
   }
 }
--- a/geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/remote/CommandExecutor.java
+++ b/geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/remote/CommandExecutor.java
@@ -14,6 +14,8 @@
  */
 package org.apache.geode.management.internal.cli.remote;
 
+import static org.apache.geode.management.internal.api.LocatorClusterManagementService.CMS_DLOCK_SERVICE_NAME;
+
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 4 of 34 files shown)


geode-core/src/main/java/org/apache/geode/management/internal/api/LocatorClusterManagementService.java
```diff
@@ -113,7 +113,7 @@
 public class LocatorClusterManagementService implements ClusterManagementService {
   @VisibleForTesting
   // the dlock service name used by the CMS
-  static final String CMS_DLOCK_SERVICE_NAME = "CMS_DLOCK_SERVICE";
+  public static final String CMS_DLOCK_SERVICE_NAME = "CMS_DLOCK_SERVICE";
```
agingade, line 116: Since "LocatorClusterManagementService" is central to the ClusterConfigurationService; will it be better to have the cms-dlock service created in this class; and have helper method to lock and unlock. Assumption is when the lock is requested the Locator Management service will be up and running.

lockConfigForUpdate() {}
unLockConfigForUpdate() {}

geode-gfsh/src/main/java/org/apache/geode/management/cli/GfshCommand.java
```diff
@@ -59,6 +59,14 @@ public boolean isOnlineCommandAvailable() {
     return gfsh.isConnectedAndReady();
   }
 
+  /**
+   * For those commands that could possibly change the cluster configuration, they need to
+   * override this method to return true.
+   */
+  public boolean affectsClusterConfiguration() {
```
agingade, line 66: How about naming it as "updatesClusterConfiguration" (); they both are same, just another name to consider, its left to you.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/remote/CommandExecutor.java
```diff
@@ -207,4 +219,32 @@ protected Object invokeCommand(Object command, GfshParseResult parseResult) {
 
     return resultModel;
   }
+
+  @VisibleForTesting
+  boolean lockCMS(Object command) {
+    if (cmsDlockService == null) {
+      return false;
+    }
+    if (!(command instanceof GfshCommand)) {
```
agingade, line 228: How about combining all of the if into an AND condition.

if (cmsDlockService == null && !(command instanceof GfshCommand) &&...)

jinmeiliao: I think this reads better and I plan to add comments in each if block

geode-core/src/main/java/org/apache/geode/management/internal/api/LocatorClusterManagementService.java
jinmeiliao: I believe it is created by this class, see `getCmsDlockService` in this class

agingade: In this PR. the Dlock is getting created in "OnlineCommandProcessor.java"; I don't have latest develop, my guess it may be getting created in rest endpoint too...

jinmeiliao: they both call `DLockService.getOrCreateService`, if it's already created, it will just get it. So, I guess whoever get to it first will create it.

jinmeiliao:
> I have some questions about `SingleGfshCommand`, which is extended by commands that will update cluster config. Now in this pull request, quite a few classes used to be subclasses of `SingleGfshCommand` now become the subclasses of `GfshCommand`. Does it mean that these classes don't update cluster config any more?

They don't update cluster config at all. They don't need to extends SingleGfshCommand in the first place and they didn't even implement the `updateGroupConfig` method at all.

> 
> By default, the subclasses of `GfshCommand` returns `false` for `affectsClusterConfiguration`. But some class, e.g. `AlterRuntimeConfigCommand` that extends `GfshCommand` returns `true` for `affectsClusterConfiguration`. Why is that?

Because AlterRuntimConfigComand does affect cluster configuration. It updates the properties part of the cluster configuration.

geode-gfsh/src/main/java/org/apache/geode/management/cli/SingleGfshCommand.java
```diff
@@ -39,7 +42,11 @@
    *        return value of your command method.
    * @return a boolean indicating whether a change to the cluster configuration was persisted.
    */
-  public boolean updateConfigForGroup(String group, CacheConfig config, Object configObject) {
-    return false;
+  public abstract boolean updateConfigForGroup(String group, CacheConfig config,
+      Object configObject);
+
+  @Override
+  public boolean affectsClusterConfiguration() {
```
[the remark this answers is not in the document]
jinmeiliao: It's marked as @Experimental

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/remote/OnlineCommandProcessor.java
```diff
@@ -53,14 +57,13 @@
...
 
-  private InternalCache cache;
-
   public OnlineCommandProcessor() {}
 
   @VisibleForTesting
-  public OnlineCommandProcessor(Properties cacheProperties, SecurityService securityService,
-      CommandExecutor commandExecutor, InternalCache cache) {
-    this.gfshParser = new GfshParser(new CommandManager(cacheProperties, cache));
+  public OnlineCommandProcessor(GfshParser gfshParser,
```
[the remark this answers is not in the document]
jinmeiliao: the change is necessary for the test to pass. All this change is because public methods in DlockService are static.

geode-gfsh/src/main/java/org/apache/geode/management/cli/SingleGfshCommand.java
```diff
@@ -39,7 +42,11 @@
    *        return value of your command method.
    * @return a boolean indicating whether a change to the cluster configuration was persisted.
    */
-  public boolean updateConfigForGroup(String group, CacheConfig config, Object configObject) {
-    return false;
+  public abstract boolean updateConfigForGroup(String group, CacheConfig config,
```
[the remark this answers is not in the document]
jinmeiliao: this class is @experimental

geode-core/src/main/java/org/apache/geode/management/internal/api/LocatorClusterManagementService.java
agingade: Can that creating (getOrCreateService ) be done in one place, in "LocatorClusterManagementService" which seems to be right place for it...

jinmeiliao: Then OnlineCommandProcessor needs to get ahold of `LocatorClusterManagementService` and get the dLock service from there? What if CMS is not started? Gfsh still needs to have this dLockService.

agingade: My understanding was, we have centralized configuration-management class, that is used by both gfsh and rest end-point to make the configuration changes, and I thought it was LocatorClusterManagementService, I guess I am wrong...

Is it "InternalConfigurationPersistenceService" where both gfsh and end-point make the configuration changes...If so, can the dlock creation and unlocking/unlocking helper method be added there.

I am trying to see if we can have the lock creation in one place, rather than in many clients/apis that make the configuration changes...

jinmeiliao: Yes, the `InternalConfigurationPersistenceService` is used by both rest and gfsh to persist the configuration change in the region, and it already has its own dlockService to make sure update to the region is thread safe. 

What are make thread-safe in here is not just "update to the region", it's the "update to the server sand then update the region", which is a level above the persistence layer.  Both rest and gfsh does this and so both needs the same dlockService.

PR #5288, opened 2020-06-23
GEODE-8283: Provide REST interface for disk-store creation

jmelchio:
- Provides create, get, list and delete operations
- Can create with locator only running

Co-Authored-By: Jason Huynh <jhuynh@vmware.com>

### For all changes:

### Note:


geode-management/src/main/java/org/apache/geode/management/configuration/DiskStore.java
```diff
@@ -0,0 +1,166 @@
...
+
+import org.apache.geode.annotations.Experimental;
+import org.apache.geode.management.runtime.DiskStoreInfo;
+
+@Experimental
+public class DiskStore extends GroupableConfiguration<DiskStoreInfo> {
+  public static final String DISK_STORE_CONFIG_ENDPOINT = "/diskstores";
+
+  @JsonIgnore
+  private String id;
```
jinmeiliao, line 35: Does it need both id and name? Are all the attributes necessary in the rest interface? Rest interface is supposed to be a more simplified/opinionated interface than the xml. It doesn't have to repeat everything that xml object has. Rather only focus on the attributes that users would need to configure most. If having it there would confuse users rather than help them, better not provide it in the rest interface.

geode-management/src/main/java/org/apache/geode/management/configuration/DiskDir.java
```diff
@@ -0,0 +1,54 @@
...
+
+package org.apache.geode.management.configuration;
+
+import java.io.Serializable;
+
+import org.apache.geode.management.api.JsonSerializable;
+
+public class DiskDir implements Serializable, JsonSerializable {
+  private String name;
+  private String dirSize;
```
jinmeiliao, line 28: would an int/long be better?

geode-core/src/main/java/org/apache/geode/management/internal/configuration/mutators/DiskStoreManager.java
```diff
@@ -0,0 +1,76 @@
...
+  private final DiskStoreConverter diskStoreConverter = new DiskStoreConverter();
+
+  public DiskStoreManager(ConfigurationPersistenceService service) {
+    super(service);
+  }
+
+  @Override
+  public void add(DiskStore config, CacheConfig existing) {
+    List<DiskStoreType> diskStoreTypes = existing.getDiskStores();
+    if (diskStoreTypes.stream().noneMatch(diskStoreType -> diskStoreType.getName()
```
jinmeiliao, line 43: I don't think you need the check here. the pipeline should already make sure that there is no existing one that has the same id exists before it's calling this `add`

geode-core/src/main/java/org/apache/geode/management/internal/configuration/mutators/DiskStoreManager.java
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  @Override
+  public void update(DiskStore config, CacheConfig existing) {
+    throw new IllegalStateException("Not implemented");
+  }
+
+  @Override
+  public void delete(DiskStore config, CacheConfig existing) {
+    existing.getDiskStores().stream()
```
jinmeiliao, line 56: same as add, no need to check

geode-management/src/main/java/org/apache/geode/management/configuration/DiskStore.java
```diff
@@ -0,0 +1,166 @@
...
+import java.util.List;
+import java.util.Objects;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+
+import org.apache.geode.annotations.Experimental;
+import org.apache.geode.management.runtime.DiskStoreInfo;
+
+@Experimental
+public class DiskStore extends GroupableConfiguration<DiskStoreInfo> {
```
jinmeiliao, line 31: If `DiskStoreInfo` is an empty class (no runtime info to gather for this configuration object), then you can declare this `DiskStore` as `public class DiskStore extends GroupableConfiguration<RuntimeInfo>`. This way, for get/list, the cms will not even issue the functions to the servers to try to retrieve an empty object.

jmelchio: In its current configuration it does list the servers to which the `DiskStore` has been applied. I'll try and see if this continues when we roll it into the `DiskStore` class. If that is the case we can drop the `DiskStoreInfo` class.

geode-core/src/main/java/org/apache/geode/management/internal/configuration/mutators/DiskStoreManager.java
jmelchio: Okay, we'll check if we can drop these checks without side effects.

geode-management/src/main/java/org/apache/geode/management/configuration/DiskDir.java
jmelchio: Yes, probably better.

geode-management/src/main/java/org/apache/geode/management/configuration/DiskStore.java
jmelchio: We weren't sure what would and would not be used often. If we have someone who can sound off on what to include I'm happy to trim the list of properties and revert to defaults where we don't have them.

geode-management/src/main/java/org/apache/geode/management/configuration/DiskStore.java
jmelchio: It appears the `DiskStoreRealizer` needs a concrete implementation of `RuntimeInfo`.

jinmeiliao: If it returns null in the `get` method?

jmelchio: That would work but then we wouldn't even get a list of machines where the disk-stores live at run time.

jinmeiliao: Ah, I see, you don't need further diskstore info on each server, but you do need the server id. I guess you can leave it as is.

geode-management/src/main/java/org/apache/geode/management/configuration/DiskStore.java
jinmeiliao: What's your use case for needing these rest api? For now I would suggest just expose those attributes that you needed.  It's always easy to add attributes to the config objects

jmelchio: We need it for setting up disk-stores in k8s.

jinmeiliao: Do you need to configure all the attributes in k8s? I am asking because of what we did for Region config object, we only expose what's critical to configure the region to simplify user experience. I would hope Diskstore would do the same. Maybe talk to Gang to find out what are the most used attributes when configuring diskstore and what are the user's pain point given the current set of options in gfsh

PR #5289, opened 2020-06-23
GEODE-8295: refactor GeodeRedisServer

dschneider-pivotal:
1. removed main thread from GeodeRedisServer
2. removed NamedThreadFactory in favor of geode's LoggingThreadFactory and LoggingExecutors
3. removed logLevel since it is configured by gemfire
4. A cache is now passed to the GeodeRedisServer constructor
5. region initialization now done in RegionProvider
6. extracted passive expiration code
7. extracted netty server code

### For all changes:

### Note:


geode-redis/src/main/java/org/apache/geode/redis/internal/PassiveExpirationManager.java
```diff
@@ -0,0 +1,80 @@
...
+
+
+  public PassiveExpirationManager(Region<ByteArrayWrapper, RedisData> dataRegion) {
+    this.dataRegion = dataRegion;
+    expirationExecutor = newSingleThreadScheduledExecutor("GemFireRedis-PassiveExpiration-");
+  }
+
+  public void start() {
+    int INTERVAL = 1;
+    expirationExecutor.scheduleAtFixedRate(() -> doDataExpiration(dataRegion), INTERVAL, INTERVAL,
```
jdeppe-pivotal, line 51: I think this would be potentially better with `scheduleWithFixedDelay` so that there's a guaranteed pause between runs. I'm concerned that a large data set would result in `doDataExpiration` taking longer than 1 second and putting pressure on the system by constantly running back-to-back.

geode-redis/src/main/java/org/apache/geode/redis/internal/netty/NettyRedisServer.java
```diff
@@ -0,0 +1,317 @@
...
+  }
+
+  private Class<? extends ServerChannel> initializeEventLoopGroups() {
+    ThreadFactory selectorThreadFactory =
+        new LoggingThreadFactory("GeodeRedisServer-SelectorThread-", false);
+
+    ThreadFactory workerThreadFactory =
+        new LoggingThreadFactory("GeodeRedisServer-WorkerThread-", true);
+
+    if (singleThreadPerConnection) {
```
jdeppe-pivotal, line 223: What do you think about removing this altogether? We do no testing with this mode and I wouldn't expect anyone to actually use it.

dschneider-pivotal: I like that idea. It uses deprecated apis so I'd really want us to test with it before allowing customers to use it.
But I think that should be on its own PR. We have one about numThreads sys prop; seems like we could do it on that one.

geode-redis/src/main/java/org/apache/geode/redis/internal/PassiveExpirationManager.java
dschneider-pivotal: good idea; it is done

geode-redis/src/main/java/org/apache/geode/redis/internal/GeodeRedisService.java
```diff
@@ -77,7 +78,7 @@ private void startRedisServer(InternalCache cache) {
           String.format("Starting GeodeRedisServer on bind address %s on port %s",
```
jdeppe-pivotal, line 78: Since you're working on this, could you move the log message after the start so that we log the actual port used in the event that it is started on an ephemeral port. (Otherwise it just says `Starting... on port 0`).

dschneider-pivotal: We currently have two log messages. One says "starting" and shows the requested configuration. The other says "started" and show the actual host and port. It seems like both are worth having. Does that resolve this conversation?

geode-redis/src/main/java/org/apache/geode/redis/internal/netty/NettyRedisServer.java
dschneider-pivotal: I added this to the numThreads sys prop story

jdeppe-pivotal: Sounds good.

PR #5291, opened 2020-06-23
[DRAFT] GEODE-8066: Pull up transitive dependency to clean up javax.activation and mail licenses.

rhoughton-pivot:


rhoughton-pivot:
Tried to get Spring to handle this, but they punted: https://github.com/spring-projects/spring-security/issues/8733

rhoughton-pivot:
@jinmeiliao @onichols-pivotal I have not updated the source-distribution license file as part of the draft. I'd be happy to pair with someone on how to do that, once I get signoff that this is a working dependency change for us.

onichols-pivotal:
> @jinmeiliao @onichols-pivotal I have not updated the source-distribution license file as part of the draft. I'd be happy to pair with someone on how to do that, once I get signoff that this is a working dependency change for us.

You can run the license-review script to point you to anything that might need fixing `./dev-tools/release/license_review.sh -v HEAD`

metatype:
IIUC, this removes javax.mail and javax.activation from the bundled dependencies.  I believe that we need those dependencies until we remove the admin API entirely.

rhoughton-pivot:
@metatype The module that pulls those in was also getting the symbols transitively. I'm not convinced that anything we can call can hit any symbols from these particular `javax` jars. If this change to drop the duplication of license in our source-license file doesn't work, can we close this PR as will-not-fix?

onichols-pivotal:
@metatype this PR is in response to the ticket you asked me to create during the last license review: GEODE-8066 Geode should not be using both activation 1.1 and javax.activation 1.2

So we are trying to go from 2 versions of a dependency to just one (not trying to get rid of it entirely).

onichols-pivotal:
@rhoughton-pivot don't forget to remove activation 1.1 (lines 269-270) from geode-assembly/src/main/dist/LICENSE as part of this PR, and also please add the ticket to the front of the PR title.

onichols-pivotal (approved):

PR #5299, opened 2020-06-24
GEODE-8200: enhance GfshRule to specify a working dir

jinmeiliao:
* improve some backward compatibility test to cover more versions.

### DIFF
--- a/geode-assembly/src/upgradeTest/java/org/apache/geode/management/DeploymentManagementUpgradeTest.java
+++ b/geode-assembly/src/upgradeTest/java/org/apache/geode/management/DeploymentManagementUpgradeTest.java
@@ -16,26 +16,55 @@
 package org.apache.geode.management;
 
 import static org.apache.geode.test.junit.assertions.ClusterManagementListResultAssert.assertManagementListResult;
+import static org.apache.geode.test.junit.rules.gfsh.GfshRule.startLocatorCommand;
 
 import java.io.File;
 import java.io.IOException;
+import java.util.Collection;
+import java.util.List;
 
 import org.junit.BeforeClass;
 import org.junit.ClassRule;
 import org.junit.Rule;
 import org.junit.Test;
+import org.junit.experimental.categories.Category;
 import org.junit.rules.TemporaryFolder;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
 
 import org.apache.geode.internal.AvailablePortHelper;
 import org.apache.geode.management.api.ClusterManagementService;
 import org.apache.geode.management.client.ClusterManagementServiceBuilder;
 import org.apache.geode.management.configuration.Deployment;
 import org.apache.geode.test.compiler.JarBuilder;
+import org.apache.geode.test.junit.categories.BackwardCompatibilityTest;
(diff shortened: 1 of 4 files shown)


geode-assembly/src/upgradeTest/java/org/apache/geode/management/DeploymentManagementUpgradeTest.java
```diff
@@ -55,16 +84,16 @@ public static void beforeClass() throws Exception {
...
   public void newLocatorCanReadOldConfigurationData() throws IOException {
-    File workingDir = tempFolder.newFolder();
     int[] ports = AvailablePortHelper.getRandomAvailableTCPPorts(3);
-    oldGfsh.execute("start locator --name=test --port=" + ports[0] + " --http-service-port="
-        + ports[1] + " --dir=" + workingDir.getAbsolutePath() + " --J=-Dgemfire.jmx-manager-port="
-        + ports[2],
-        "deploy --jar=" + clusterJar.getAbsolutePath(),
-        "shutdown --include-locators");
+    GfshExecution execute =
+        GfshScript.of(startLocatorCommand("test", ports[0], ports[2], ports[1], 0))
```
jchen21, line 89: For code readability, it is better to have some meaningful names for the ports. Before the code change, it is easy to identify the purpose of a port with the gfsh option in the code. Now it is not that straight forward to tell ports[0] is locator port. port[2] is JMX port and ports[1] is HTTP port, unless you see to the definition of `startLocatorCommand`.

geode-junit/src/main/java/org/apache/geode/test/junit/rules/gfsh/GfshRule.java
```diff
@@ -199,4 +220,23 @@ private void stopMembers(GfshExecution gfshExecution) {
     }
     execute(GfshScript.of(stopMemberScripts).withName("Stop-Members"));
   }
+
+  public static String startServerCommand(String name, int port, int connectedLocatorPort) {
```
kirklund, line 224: These new static methods don't really fit well in GfshRule. I think you should probably move them to two new classes (probably not a rule) named something like ServerCommandBuilder and LocatorCommandBuilder, and explode the parameters to each have their own setter type method:
```
String startServerCommand = new ServerCommandBuilder()
    .withPort(port)
    .withLocator(locatorPort)
    .create(name);
```

kirklund: There are actually a couple existing classes like this that we could move to geode-dunit and make any necessary changes:
```
geode-core/src/integrationTest/java/org/apache/geode/distributed/LocatorCommand.java
geode-core/src/integrationTest/java/org/apache/geode/distributed/ServerCommand.java
```

kirklund: We could commit what you have ready and then make further changes. I'll go ahead and approve and let you decide.

jinmeiliao: Thanks! I will merge this now and we can make improvements on it when we get to consolidating all these methods into one place.

PR #5300, opened 2020-06-24
GEODE-8095: Changes to make GEODE Respond to Restore Redundancy REST Command

mhansonp:
These are the core changes that will allow the invocation of Restore Redundancy from REST

### DIFF
--- a/geode-web-management/src/main/java/org/apache/geode/management/internal/rest/controllers/RestoreRedundancyOperationController.java
+++ b/geode-web-management/src/main/java/org/apache/geode/management/internal/rest/controllers/RestoreRedundancyOperationController.java
@@ -33,7 +33,6 @@ import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RestController;
 
 import org.apache.geode.internal.security.SecurityService;
-import org.apache.geode.logging.internal.log4j.api.LogService;
 import org.apache.geode.management.api.ClusterManagementListOperationsResult;
 import org.apache.geode.management.api.ClusterManagementOperationResult;
 import org.apache.geode.management.operation.RestoreRedundancyRequest;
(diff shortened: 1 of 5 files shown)


geode-web-management/src/main/java/org/apache/geode/management/internal/rest/controllers/RestoreRedundancyOperationController.java
```diff
@@ -0,0 +1,74 @@
...
+        Optional.ofNullable(securityService).map(SecurityService::getSubject).map(Object::toString)
+            .orElse(null));
+    ClusterManagementOperationResult<RestoreRedundancyRequest, RestoreRedundancyResults> result =
+        clusterManagementService
+            .start(operation);
+    return new ResponseEntity<>(result, HttpStatus.ACCEPTED);
+  }
+
+  @ApiOperation(value = "list restore-redundancy")
+  @PreAuthorize("@securityService.authorize('DATA', 'MANAGE')")
```
jinmeiliao, line 59: Is this the right authorization code we need to  use?

geode-web-management/src/main/java/org/apache/geode/management/internal/rest/controllers/RestoreRedundancyOperationController.java
```diff
@@ -0,0 +1,74 @@
...
+  @GetMapping(RESTORE_REDUNDANCY_ENDPOINT)
+  public ClusterManagementListOperationsResult<RestoreRedundancyRequest, RestoreRedundancyResults> listRestoreRedundancies() {
+    return clusterManagementService.list(new RestoreRedundancyRequest());
+  }
+
+  @ApiOperation(value = "get restore-redundancy")
+  @PreAuthorize("@securityService.authorize('DATA', 'MANAGE')")
```
jinmeiliao, line 66: check this as well

geode-web-management/src/main/java/org/apache/geode/management/internal/rest/controllers/RestoreRedundancyOperationController.java
mhansonp: How do I check it?

mhansonp: We are agreed that this is correct.

geode-web-management/src/main/java/org/apache/geode/management/internal/rest/controllers/RestoreRedundancyOperationController.java
mhansonp: We are agreed that this is correct.

PR #5303, opened 2020-06-25
GEODE-8176: Fix for flaky testPingWrongServer

alb3rtobr:
The DistributedPingMessage is handled asynchronously: I have check that if I introduce a sleep `process()` method of `DistributedPingServer`, the test case fails because the `ClientHealthMonitor` has not received the ping.
I suspect that in the ocassions where the test failed was due to a race condition between the check of the `ClientHealthMonitor` content and the process of the message. So Im introducing a wait when checking the `ClientHealthMonitor` to avoid this.

### DIFF
--- a/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/tier/sockets/ClientServerMiscDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/tier/sockets/ClientServerMiscDUnitTest.java
@@ -20,6 +20,8 @@ import static org.apache.geode.test.awaitility.GeodeAwaitility.await;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.Assert.assertEquals;
 
+import java.time.Duration;
+
 import org.junit.Test;
 
 import org.apache.geode.cache.Operation;
@@ -117,7 +119,6 @@ public class ClientServerMiscDUnitTest extends ClientServerMiscDUnitTestBase {
     InternalDistributedMember server2ID = server2.invoke("get ID", () -> cache.getMyId());
     pool = (PoolImpl) createClientCache(NetworkUtils.getServerHostName(), PORT1);
     // send the ping to server1 but use server2's identifier so the ping will be forwarded
-
     ClientProxyMembershipID proxyID = server1.invoke(
         () -> CacheClientNotifier.getInstance().getClientProxies().iterator().next().getProxyID());
     logger.info("ProxyID is : " + proxyID);
@@ -129,7 +130,9 @@ public class ClientServerMiscDUnitTest extends ClientServerMiscDUnitTestBase {
     PingOp.execute(pool, new ServerLocation(NetworkUtils.getServerHostName(), PORT1), server2ID);
     // if the ping made it to server2 it will have the client's ID in its health monitor
     server2.invoke(() -> {
-      assertEquals(1, ClientHealthMonitor.getInstance().getClientHeartbeats().keySet().size());
+      await("For heartbeat to be received").timeout(Duration.ofMinutes(1))
+          .untilAsserted(() -> assertEquals(1,
+              ClientHealthMonitor.getInstance().getClientHeartbeats().keySet().size()));
       ClientProxyMembershipID proxyIDFound =
           ClientHealthMonitor.getInstance().getClientHeartbeats().keySet().iterator().next();
       logger.info("ProxyID found in clientHealthMonitor: " + proxyIDFound);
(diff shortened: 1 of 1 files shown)


geode-core/src/distributedTest/java/org/apache/geode/internal/cache/tier/sockets/ClientServerMiscDUnitTest.java
```diff
@@ -129,7 +130,9 @@ public void testPingWrongServer() {
     PingOp.execute(pool, new ServerLocation(NetworkUtils.getServerHostName(), PORT1), server2ID);
     // if the ping made it to server2 it will have the client's ID in its health monitor
     server2.invoke(() -> {
-      assertEquals(1, ClientHealthMonitor.getInstance().getClientHeartbeats().keySet().size());
+      await("For heartbeat to be received").timeout(Duration.ofMinutes(1))
```
jujoramos, line 133: Please use the default timeout (5 minutes) set by `GeodeAwaitility` instead: `await().untilAsserted()` (without the `timeout`).

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/tier/sockets/ClientServerMiscDUnitTest.java
```diff
@@ -20,6 +20,8 @@
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.Assert.assertEquals;
```
jujoramos, line 21: The current preferred approach is to always use `org.assertj.core.api.Assertions` instead of `org.junit.Assert`, can you change it?.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/tier/sockets/ClientServerMiscDUnitTest.java
alb3rtobr: I thought 5 minutes were too much for this case but if its ok for you, then I can change it, no problem.

jujoramos: I agree in that 5 minutes might be "too much" for this test, but it's the current standard and there are only a handful of tests (8 I think) that overrides it, so I think is better to keep using the defaults for everything new.
As a side note, the poll occurs every 100 milliseconds anyway, so the condition should be met way sooner than 5 minutes.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/tier/sockets/ClientServerMiscDUnitTest.java
```diff
@@ -117,19 +116,21 @@ public void testPingWrongServer() {
     InternalDistributedMember server2ID = server2.invoke("get ID", () -> cache.getMyId());
     pool = (PoolImpl) createClientCache(NetworkUtils.getServerHostName(), PORT1);
     // send the ping to server1 but use server2's identifier so the ping will be forwarded
-
     ClientProxyMembershipID proxyID = server1.invoke(
         () -> CacheClientNotifier.getInstance().getClientProxies().iterator().next().getProxyID());
     logger.info("ProxyID is : " + proxyID);
```
jujoramos, line 121: This logging message can be removed.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/tier/sockets/ClientServerMiscDUnitTest.java
```diff
@@ -117,19 +116,21 @@ public void testPingWrongServer() {
...
     PingOp.execute(pool, new ServerLocation(NetworkUtils.getServerHostName(), PORT1), server2ID);
     // if the ping made it to server2 it will have the client's ID in its health monitor
     server2.invoke(() -> {
-      assertEquals(1, ClientHealthMonitor.getInstance().getClientHeartbeats().keySet().size());
+      await("For heartbeat to be received").untilAsserted(() -> assertThat(
+          ClientHealthMonitor.getInstance().getClientHeartbeats().keySet().size())
+              .isEqualTo(1));
       ClientProxyMembershipID proxyIDFound =
           ClientHealthMonitor.getInstance().getClientHeartbeats().keySet().iterator().next();
       logger.info("ProxyID found in clientHealthMonitor: " + proxyIDFound);
```
jujoramos, line 136: This logging message can be removed.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/tier/sockets/ClientServerMiscDUnitTest.java
alb3rtobr: done!

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/tier/sockets/ClientServerMiscDUnitTest.java
alb3rtobr: done!

PR #5305, opened 2020-06-25
GEODE-8095 backport to 1.13

mhansonp:


jinmeiliao:
I am not sure if merge it here would be the right way to do it. You probably would trigger some commit checks if you don't add some specific commit message saying it's cherry picked from what commit on develop

PR #5306, opened 2020-06-25
GEODE-8195: ConcurrentModificationException from LocatorMembershipLis…

bschuchardt:
…tenerImpl

I've replaced the "for" loop using an implicit Iterator with one using an
explicit Iterator so that its safe "remove()" method can be used.  The
Iterator method is stated as being the only safe way to modify the
collection while iterating over its contents.

I've also modified a test to validate the fix.  The test forces a
failure to send two messages to an address.  The failures are then
handled in the code that was throwing the
ConcurrentModificationException and, since there are two failures,
it causes two removals to be performedon the failedMessages collection.

### For all changes:

### Note:

### DIFF
--- a/geode-wan/src/main/java/org/apache/geode/cache/client/internal/locator/wan/LocatorMembershipListenerImpl.java
+++ b/geode-wan/src/main/java/org/apache/geode/cache/client/internal/locator/wan/LocatorMembershipListenerImpl.java
@@ -18,6 +18,7 @@ package org.apache.geode.cache.client.internal.locator.wan;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


geode-wan/src/main/java/org/apache/geode/cache/client/internal/locator/wan/LocatorMembershipListenerImpl.java
```diff
@@ -328,6 +331,7 @@ public void run() {
                   Thread.currentThread().interrupt();
                   logger.warn(
                       "Locator Membership listener permanently failed to exchange locator information due to interruption.");
+                  return;
```
Bill, line 334: If a sleep is interrupted we now bail out of `run()` entirely, rather than continuing on to attempt delivering the message to the next target. Why the change?

bschuchardt: The exception handler was leaving the interrupt bit set.  Nothing useful could happen after that.  Any I/O operations that it attempted would fail, resulting in the thread hanging around sleeping until it gave up;.

bschuchardt:
DistributedTests failed due to GEODE-8172, which we're seeing in a lot of runs.

PR #5309, opened 2020-06-25
GEODE-8250: Create new custom log config acceptance tests

kirklund:
Resubmit new acceptance tests after reverting initial commit.

ServerWithCustomLogConfigAcceptanceTest was failing on Windows because it was attempting to use `locators=""` in the command to launch the Server. Empty string is the default so explicitly specifying this is not necessary. I deleted it and the test started passing on Windows.


geode-assembly/src/acceptanceTest/java/org/apache/geode/logging/ServerWithCustomLogConfigAcceptanceTest.java
```diff
@@ -0,0 +1,209 @@
...
+    serverName = testName.getMethodName();
+
+    workingDir = temporaryFolder.getRoot().toPath().toAbsolutePath();
+    serverLogFile = workingDir.resolve(serverName + ".log");
+    customLogFile = workingDir.resolve("custom.log");
+  }
+
+  @Test
+  public void serverLauncherUsesDefaultLoggingConfig() {
+    String startLocatorCommand = String.join(" ",
```
jchen21, line 81: `startServerCommand`?

geode-assembly/src/acceptanceTest/java/org/apache/geode/management/internal/cli/commands/StartLocatorAcceptanceTest.java
```diff
@@ -14,43 +14,84 @@
...
+  private Path geodeDependencies;
+  private Path stdoutFile;
+  private Path locatorLogFile;
+  private Path pulseLogFile;
+  private Path javaBin;
+
   @Rule
   public GfshRule gfshRule = new GfshRule();
+  @Rule
+  public RequiresGeodeHome requiresGeodeHome = new RequiresGeodeHome();
```
kirklund, line 49: Delete RequiresGeodeHome

geode-assembly/src/acceptanceTest/java/org/apache/geode/management/internal/cli/commands/StartLocatorAcceptanceTest.java
```diff
@@ -14,43 +14,84 @@
...
+
+  @Before
+  public void setUpRandomPorts() {
+    int[] ports = AvailablePortHelper.getRandomAvailableTCPPorts(2);
+    locatorPort = ports[0];
+    httpServicePort = ports[1];
+  }
+
+  @After
+  public void stopLocator() throws Exception {
```
kirklund, line 61: Delete stopLocator(). I must've only removed these from the other test.

geode-junit/src/main/java/org/apache/geode/test/junit/rules/gfsh/GfshRule.java
```diff
@@ -145,9 +145,7 @@ public GfshExecution execute(GfshScript gfshScript, File workingDir) {
 
   public GfshExecution execute(GfshScript gfshScript) {
     try {
-      File workingDir = new File(temporaryFolder.getRoot(), gfshScript.getName());
-      workingDir.mkdirs();
-      return execute(gfshScript, workingDir);
+      return execute(gfshScript, temporaryFolder.getRoot());
```
kirklund, line 148: Note: The workingDir for gfshScript is the workingDir for the GFSH process spawned by the JUnit test. GFSH then spawns the Locator and Server which is running in a subfolder. This change makes it easy for GfshRule to stop ALL Locators and Servers anywhere in the TemporaryFolder. Since only one gfshScript is executed at a time, the GFSH workingDir should be ok as temporaryFolder.getRoot(). 

On the off-chance that this does cause a problem, it would have to be for other GFSH commands that write something to the current workingDir of GFSH. I'm fairly confident we don't have tests that will have disk artifacts colliding just because of this little change.

geode-assembly/src/acceptanceTest/java/org/apache/geode/management/internal/cli/commands/StartLocatorAcceptanceTest.java
```diff
@@ -14,43 +14,84 @@
...
 import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
 
+import org.apache.geode.internal.AvailablePortHelper;
+import org.apache.geode.test.junit.rules.RequiresGeodeHome;
 import org.apache.geode.test.junit.rules.gfsh.GfshExecution;
 import org.apache.geode.test.junit.rules.gfsh.GfshRule;
 import org.apache.geode.test.junit.rules.gfsh.GfshScript;
 
 public class StartLocatorAcceptanceTest {
```
kirklund, line 35: I didn't mean to edit this class. Sigh.

geode-assembly/src/acceptanceTest/java/org/apache/geode/management/internal/cli/commands/StartLocatorAcceptanceTest.java
kirklund: Wrong test

geode-assembly/src/acceptanceTest/java/org/apache/geode/management/internal/cli/commands/StartLocatorAcceptanceTest.java
kirklund: I reverted the unintended changes to this test.

geode-junit/src/main/java/org/apache/geode/test/junit/rules/gfsh/GfshRule.java
jinmeiliao: so basically all scripts executed under the same gfshRule would use the same gfsh working dir instead of each execution has its own. I think this is a good change.

kirklund:
AcceptanceTest job hit a SIGABRT for invalid pointer. StatusLocatorExitCodeAcceptanceTest also failed -- this PR doesn't change that test.

geode-junit/src/main/java/org/apache/geode/test/junit/rules/gfsh/GfshRule.java
jinmeiliao: Looks like your change made some other tests that uses gfshRule fail. Can I offer a different approach to fix your test only? This way you don't need to touch other tests and the gfshRule. I had one of your tests modified and put it on my repo so that it can compare with yours:

https://github.com/kirklund/geode/commit/e88006e8f1c9f5924f7daacbc1a7ef5668d72212

kirklund:
:geode-assembly:docker SIGABRT. I think we just need to rerun tests again and hopefully Docker will behave.

PR #5313, opened 2020-06-26
GEODE-8302: Fixed 'events not queued conflated' stats when group-tran…

albertogpz:
…saction-events is true

Batch conflation is not compatible with group-transaction-events.
It must be prevented that both are enabled at the same time for
a given gateway sender.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/cache/wan/GatewaySenderFactory.java
+++ b/geode-core/src/main/java/org/apache/geode/cache/wan/GatewaySenderFactory.java
@@ -39,8 +39,9 @@ public interface GatewaySenderFactory {
    * Indicates whether events belonging to the same transaction must be
    * delivered inside the same batch, i.e. they cannot be spread across different
    * batches.
-   * Can only be enabled on serial gateway senders with just one dispatcher
+   * It can only be enabled on serial gateway senders with just one dispatcher
    * thread or on parallel gateway senders.
+   * Besides, it cannot be enabled if batch conflation is enabled.
    *
    * @param groupTransactionEvents boolean to indicate whether events from
    *        the same transaction must be delivered inside
--- a/geode-core/src/main/java/org/apache/geode/cache/wan/GatewaySenderFactory.java
+++ b/geode-core/src/main/java/org/apache/geode/cache/wan/GatewaySenderFactory.java
@@ -29,7 +29,7 @@ public interface GatewaySenderFactory {
 
   /**
    * Indicates whether all VMs need to distribute events to remote site. In this case only the
-   * events originating in a particular VM will be in dispatched in order.
+   * events originating in a particular VM will be dispatched in order.
    *
    * @param isParallel boolean to indicate whether distribution policy is parallel
    */
@@ -39,9 +39,9 @@ public interface GatewaySenderFactory {
    * Indicates whether events belonging to the same transaction must be
    * delivered inside the same batch, i.e. they cannot be spread across different
    * batches.
-   * It can only be enabled on serial gateway senders with just one dispatcher
-   * thread or on parallel gateway senders.
-   * Besides, it cannot be enabled if batch conflation is enabled.
+   * <code>groupTransactionEvents</code> can be enabled only on parallel gateway senders
+   * or on serial gateway senders with just one dispatcher thread.
+   * It cannot be enabled if batch conflation is enabled.
    *
    * @param groupTransactionEvents boolean to indicate whether events from
    *        the same transaction must be delivered inside
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/wan/serial/SerialGatewaySenderQueue.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/wan/serial/SerialGatewaySenderQueue.java
@@ -106,8 +106,8 @@ public class SerialGatewaySenderQueue implements RegionQueue {
   private final Deque<Long> peekedIds = new LinkedBlockingDeque<Long>();
 
   /**
-   * Contains the peekedIds but not including those that were peeked to complete a transaction
-   * inside a batch when groupTransactionEvents is set.
+   * Contains the peekedIds, excepting those peeked to complete transactions
+   * inside a batch.
    */
   private final Deque<Long> peekedIdsWithoutExtra = new LinkedBlockingDeque<Long>();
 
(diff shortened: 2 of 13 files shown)


DonalEvans:
When the following diff is applied, the modified test `testPartitionedSerialPropagationHA()` also fails with incorrectly conflated events. Since this test does not set `groupTransactionEvents` to true, it seems like there is more to this issue.

`diff --git a/geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialWANPropagation_PartitionedRegionDUnitTest.java b/geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialWANPropagation_PartitionedRegionDUnitTest.java
index 9e80bab9aa..59b5b7a903 100644
--- a/geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialWANPropagation_PartitionedRegionDUnitTest.java
+++ b/geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialWANPropagation_PartitionedRegionDUnitTest.java
@@ -330,6 +330,8 @@ public class SerialWANPropagation_PartitionedRegionDUnitTest extends WANTestBase
     inv.join();
     vm2.invoke(() -> WANTestBase.validateRegionSize(getTestMethodName() + "_PR", 1000));
     vm3.invoke(() -> WANTestBase.validateRegionSize(getTestMethodName() + "_PR", 1000));
+    vm4.invoke(() -> WANTestBase.checkConflatedStats("ln", 0));
+    vm5.invoke(() -> WANTestBase.checkConflatedStats("ln", 0));
   }
 
   @Test
`

albertogpz:
> @albertogpz Good material - thanks for writing the docs for this feature.
> I have some edits for clarification and consistency, in both user guide files and the javadoc comments in class files. I will send these to you in email for you to incorporate as you see fit. Thanks!

Thanks for the review, Dave. I have accepted all  your proposals with the exception of the javadoc on ```GatewaySenderQueue.java``` which had not ended up describing correctly the ```peekedIdsWithoutExtra``` member variable. I changed it to a simpler sentence. Let me know if you are ok with the change.

albertogpz:
In the last commit I reverted most of the code added in the second [4ff9b1d](https://github.com/apache/geode/pull/5313/commits/4ff9b1d236771d80bbf6f613d1e13b131519743d) and just added a simpler fix.
The comment from @davebarnes97 about the javadoc in ```GatewaySenderQueue.java``` would not apply anymore as that javadoc would not change.

albertogpz:
> Thank you for incorporating my previous suggestions. Just two more items:
> 
> In geode-docs/tools_modules/gfsh/command-pages/create.html.md.erb:
> Please change "gatewaysenders" to "gateway senders".
> 
> In CreateGatewaySenderCommand.java and four other files, please reword the error message
> 
> "GatewaySender %s cannot be created with group transaction events set to true and batch conflation enabled"
> 
> to clarify that the two properties are mutually exclusive. Suggestion, add "both":
> 
> "GatewaySender %s cannot be created with both group transaction events set to true and batch conflation enabled"

Thanks for the comments, Dave.
Changes done on the erb file and on four java files. I hope you meant four files counting with CreateGatewaySenderCommand.java. I could not find the string in a fifth file.

DonalEvans:
It appears that the failing StressNewTest seen in this PR is caused by https://issues.apache.org/jira/browse/GEODE-8320 and has not been introduced by these changes, so with a clean pre-checkin, this should be okay to merge.

albertogpz:
> It appears that the failing StressNewTest seen in this PR is caused by https://issues.apache.org/jira/browse/GEODE-8320 and has not been introduced by these changes, so with a clean pre-checkin, this should be okay to merge, although ideally, it would be better to fix the underlying issue and then allow the StressNewTest to pass 100% of the time before merging this.

I'd rather have this PR merged first because fixing the flaky test case will probably take longer and I plan to do it on top of this solution.

nabarunnag:
Hi Alberto, I would suggest splitting up the commit for GEODE-8320 fix separate from GEODE-8302.
You can still work on multiple commits for GEODE-8320
- This helps to maintain clear records in the release notes for releases.
- In case we have to revert the fix for GEODE-8320 we don't have to also revert the fix for GEODE-8302 and vice versa,

albertogpz:
> Hi Alberto, I would suggest splitting up the commit for GEODE-8320 fix separate from GEODE-8302.
> You can still work on multiple commits for GEODE-8320
> 
>     * This helps to maintain clear records in the release notes for releases.
> 
>     * In case we have to revert the fix for GEODE-8320 we don't have to also revert the fix for GEODE-8302 and vice versa, as these two fixes are clubbed together

Hi Naba. What you are proposing makes a lot of sense. Nevertheless, in this case, I do not know if it is worth it as the solution for both tickets is shared, i.e., the solution for GEODE-8320 is a subset of the solution for GEODE-8302.
If I split the solution into two PRs I would have to repeat the solution on both PRs, have it reviewed twice, do changes if needed on both...
Do you still think I should do it like that or could we just have the PR for GEODE-8302 solve both GEODE-8302  and GEODE-8320?

alb3rtobr:
@nabarunnag I agree that is not recommended to include more than one ticket solution in the same commit, but in this case one ticket is a subset of the other one. Luckily Alberto has solved the bigger one, that as a side effect solves the other. Or even you can think that it is invalidating the other one, as it cannot be reproduced after merging GEODE-8302.

Anyway independently of the decision taken, GEODE-8302 can be merged into develop if there is no other problem with the review. Then Alberto could prepare a separate PR with the code of GEODE-8320 that will result on an empty commit in develop because the code is already there.

albertogpz:
> > It appears that the failing StressNewTest seen in this PR is caused by https://issues.apache.org/jira/browse/GEODE-8320 and has not been introduced by these changes, so with a clean pre-checkin, this should be okay to merge, although ideally, it would be better to fix the underlying issue and then allow the StressNewTest to pass 100% of the time before merging this.
> 
> I'd rather have this PR merged first because fixing the flaky test case will probably take longer and I plan to do it on top of this solution.

@DonalEvans Could you please review my latest commit? It was targeted at fixing a failing test case that was actually not the same as the one in GEODE-8320 but this last commit fixes the problem in both test cases.

mkevo:
As it has two approvals I will merge it. Thank you all!

PR #5316, opened 2020-06-26
GEODE-8312: Improve Redis pub/sub capabilities

jdeppe-pivotal:
- Add ability to [P]UNSUBSCRIBE from multiple channels.
- Match Redis semantics when subscribing to empty channel name - (`""`)
- Interact with plain channel names as `byte[]`s instead of `String`s - (fixes issues on Windows when subscribing with arbitrary byte[] names).

### DIFF
--- a/geode-redis/src/main/java/org/apache/geode/redis/internal/netty/ExecutionHandlerContext.java
+++ b/geode-redis/src/main/java/org/apache/geode/redis/internal/netty/ExecutionHandlerContext.java
@@ -188,13 +188,6 @@ public class ExecutionHandlerContext extends ChannelInboundHandlerAdapter {
       return;
     }
 
-    if (command.isOfType(RedisCommandType.SELECT)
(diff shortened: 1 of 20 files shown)


geode-redis/src/main/java/org/apache/geode/redis/internal/netty/ExecutionHandlerContext.java
```diff
@@ -188,6 +188,13 @@ private void executeCommand(ChannelHandlerContext ctx, Command command) throws E
       return;
     }
 
+    if (command.isOfType(RedisCommandType.SELECT)
```
dschneider-pivotal, line 191: I don't understand why this code was added. These commands are on our unimplemented list. Was this added to this PR by accident?

dschneider-pivotal: Jens said that code was mistakenly added. It has now been removed

PR #5329, opened 2020-06-30
GEODE-8029: Allow OplogEntryIdSet to Overflow

jujoramos:
Do not delete drf files during member startup as that should be only
done by the compactor thread. Instead, allow the OplogEntryIdSet to
grow over the default capacity and log a warning message instructing
the user to manually compact the disk-stores.

- Added unit tests.
- Replaced usages of 'junit.Assert' by 'assertj'.
- Modified DiskStoreImpl.deadRecordCount to return long instead of int.
- Added internal overflow implementation to the OplogEntryIdSet so it can
  grow above the default limit.

### For all changes:

### Note:


geode-core/src/main/java/org/apache/geode/internal/cache/Oplog.java
```diff
@@ -937,17 +937,11 @@ void initAfterRecovery(boolean offline) {
         // this.crf.raf.seek(this.crf.currSize);
       } else if (!offline) {
         // drf exists but crf has been deleted (because it was empty).
+        // I don't think the drf needs to be opened. It is only used during
+        // recovery.
+        // At some point the compacter my identify that it can be deleted.
```
dschneider-pivotal, line 942: change "my" to "may"

geode-core/src/main/java/org/apache/geode/internal/cache/DiskStoreImpl.java
```diff
@@ -3517,33 +3518,86 @@ public String toString() {
...
+        if (id > 0 && id <= 0x00000000FFFFFFFFL) {
+          this.currentInts.get().add((int) id);
+        } else {
+          this.currentLongs.get().add(id);
+        }
+      } catch (IllegalArgumentException illegalArgumentException) {
+        // See GEODE-8029.
+        // Too many entries on the accumulated drf files, overflow and continue.
+        logger.warn(
+            "There are too many entries within the disk-store, please execute an offline compaction.",
```
agingade, line 3570: "Too many entries", this is referring to deleted entries, right? In that case how about changing it 
to "large number of deleted entries"?

agingade: Do we need to pass "illegalArgumentExcpetion" to warning message? It may not be useful for end user.

geode-core/src/main/java/org/apache/geode/internal/cache/DiskStoreImpl.java
```diff
@@ -3517,33 +3518,86 @@ public String toString() {
...
+            illegalArgumentException);
+
+        // Overflow to the next [Int|Long]OpenHashSet and continue.
+        if (id > 0 && id <= 0x00000000FFFFFFFFL) {
```
agingade, line 3574: @dschneider-pivotal 
Is there a scenario where the same ID is added many times? In that case, earlier we had only one id set, and there was no chance of duplicate. Now with the new approach, the same ID could be present in multiple id sets.
One of the caller of add is "offlineCompact()" its iterating over live entries and calling add...Will there be two live entries with the same id?

dschneider-pivotal: I don't think so. The ids are unique and should only be deleted once from a drf. If this did happen it would just cause the size to be larger than it should be. So you could review the code that uses the size of this class.
You could prevent this from having each add first call contains which shouldn't slow down the add too much.

geode-core/src/main/java/org/apache/geode/internal/cache/Oplog.java
jujoramos: Thanks @dschneider-pivotal!. I'll fix the type and create another ticket for the long term solution.

geode-core/src/main/java/org/apache/geode/internal/cache/DiskStoreImpl.java
jujoramos: 👍

jujoramos:
@pivotal-eshu 
> If we no longer need the OplogEntryIdSet after the recovery, shall we clear these sets once it is done?

There are no references to the `OplogEntryIdSet` instance once we finish the recovery, so the class itself (along with all internal structures) will eventually be removed by the garbage collector.

> With these many entryIds in drf files, it will also take longer time to recover. I think Darrel's idea for removing unnecessary drf files will help in this case as well.

Agreed, I'll create another ticket for that.

geode-core/src/main/java/org/apache/geode/internal/cache/DiskStoreImpl.java
jujoramos: @agingade: I think we're good on this regard, `ids` are unique and the actual size of the `OplogEntryIdSet` is _**only used**_ by `offline-validate` / `offline-compact` commands. That is, if we ever report a bigger number for `deadRecordCount` (set through `OplogEntryIdSet.size()`) because there are duplicated `ids` (shouldn't happen), the user would know that the `disk-store` should be compacted anyways.
The real problem would happen if we ever report there are not compact-able entries when there actually are, which is not the case here.

agingade: Thanks for the clarification.

PR #5334, opened 2020-07-01
GEODE-8240: Back-Port to 1.12

Bill:
Back-port to `support/1.12` branch, of https://github.com/apache/geode/pull/5273 fix for [GEODE-8240](https://issues.apache.org/jira/browse/GEODE-8240)

I used `git cherry-pick -x bfe1ca113124ea958dc02dd9be90cdb0aabe5c4d`

In resolving conflicts I mostly accepted all non-conflicting changes from 1.12. An exception was that in many places where we were using `version.compareTo(other) < 0` or `version.compareTo(other) >= 0`, I accepted the changes from the cherry-pick commit to replace those with `version.isOlderThan(other)` or `version.isNotOlderThan(other)`. Those changes were all made by @pivotal-jbarrett on `develop` and they improve readability.

~The other change I had to make was to fix the `equals(other)` method on `InternalDistributedMember`. That method didn't work when comparing a `InternalDistributedMember` to a `MemberIdentifier`. As a result the 1.12 version of `GMSJoinLeaveJUnitTest` was failing. Fixing `equals(other)` fixed the test. The alternative would have been a lot of surgery on that test (mocking.). Rather than perturb that test further, I left it alone and fixed `IDM.equals(other)`.~

I did not change `InternalDistributedMember.equals(other)`—instead, I pulled in a couple minor changes to `GMSJoinLeaveJUnitTest` from the `develop` branch, to get that test running with the (wrong/unchanged) `IDM.equals(other)` functionality. Since this ticket is not about fixing `IDM` this was the better way to go.


geode-core/src/main/java/org/apache/geode/distributed/internal/membership/InternalDistributedMember.java
```diff
@@ -440,15 +441,21 @@ public boolean isPartial() {
...
-    if (o == null || getClass() != o.getClass()) {
+    if (o == null) {
+      return false;
+    }
+    if (o instanceof InternalDistributedMember) {
+      final InternalDistributedMember that = (InternalDistributedMember) o;
+      return memberIdentifier.equals(that.memberIdentifier);
+    } else if (o instanceof MemberIdentifier) {
+      return memberIdentifier.equals(o);
+    } else {
```
Bill, line 456: this `equals()` method needed repair to get `GMSJoinLeaveJUnitTest` passing

geode-core/src/integrationTest/java/org/apache/geode/distributed/internal/membership/gms/membership/GMSJoinLeaveJUnitTest.java
```diff
@@ -1589,6 +1590,36 @@ public void testMulticastDiscoveryNotAllowed() {
...
+
+    viewWithWrongVersion.remove(gmsJoinLeaveMemberId);
+    viewWithWrongVersion.add(myMemberIDWithWrongVersion);
+
+    gmsJoinLeave.installView(viewWithWrongVersion);
+
+    assertThat(
+        gmsJoinLeave.getView().getCanonicalID(gmsJoinLeaveMemberId).getVersionOrdinalObject())
+            .isEqualTo(Version.getCurrentVersion());
+  }
```
Bill, line 1621: this is the new unit test for GEODE-8240

geode-core/src/upgradeTest/java/org/apache/geode/internal/cache/rollingupgrade/RollingUpgradeDUnitTest.java
```diff
@@ -205,6 +210,22 @@ void doTestRollAll(String regionType, String objectType, String startingVersion)
...
+        final InternalDistributedSystem ids =
+            (InternalDistributedSystem) distributedSystem;
+        final DistributionManager distributionManager = ids.getDistributionManager();
+        final MembershipView<InternalDistributedMember> view =
+            distributionManager.getDistribution().getView();
+
+        for (final InternalDistributedMember member : view.getMembers()) {
+          assertThat(member.getVersionOrdinal()).isEqualTo(versionOrdinalAfterUpgrade);
+        }
+      });
```
Bill, line 227: This is the new rolling upgrade test logic that verifies that after the upgrade, all the `MemberIdentifiers` in the view, report the current version (not an old version.)

This is adapted from the original Ericsson PR that showed us the bug that became GEODE-8240. Rather than use gfsh, we are delving into the `IDM` directly here.

Bill:
> I trust this backport is pretty similar to the GEODE-8240 fix on develop and your PR checks and testing are good.
> 
> I am curious what the implications are for users currently running Geode 1.12.0. Should we advise that they upgrade to 1.12.1 to get this fix before attempting to upgrade to 1.13.0?

This fix will be ported to `support/1.13` next. So 1.12.x users will be able to upgrade to the soon-to-be-released 1.13.0 version directly.

> 
> What about upgrades from 1.11.0 to 1.12.0 or 1.11.0 to 1.12.1?

The bug exists in the 1.12 line before this PR. I think this PR will go into a 1.12.1 or 1.12.2? 

If that's right, then a 1.11.0 to 1.12.0 upgrade will see the bug, and a 1.11.0 to 1.12.1 (or whatever release this PR goes into) will not see the bug.

Incidentally, I don't really know what versions of 1.12 actually exist after 1.12.0. I see a bunch of four-part version numbers here:

image

I don't understand those four-part version numbers.

But I can say that a 1.11.x upgrade to a 1.12.x version that has this fix, will work and the bug will be gone.

> In general does this bug only affect upgrades between minors, or between patch versions too?

My understanding is that Geode usually does not change `Version.CURRENT` between patch releases. On the other hand, I think I have heard that it has been done at least once.

If you upgrade from one patch version to another patch version in the same line and `Version.CURRENT` doesn't change then you won't see this bug.

If you upgrade from one patch version to another patch version in the same line and `Version.CURRENT` _does_ change then you _will_ see this bug.

Bill:
> I don't think InternalDistributedMember should be changed in order to make this new unit test pass. Instead, the unit test should be changed to use MemberIdentifierImplFactory and MemberIdentifierImpl. That's what the test is doing in the develop branch. That was done a while ago & it just didn't make it over to the support branch.

Right you are! I undid my change to `IDM.equals(other)` in favor of small changes to `GMSJoinLeaveJUnitTest`.

Good advice!
 
> Also, shouldn't this be ported to support/1.13 before it's ported to support/1.12?

Ooh yeah. I'll go do that…

Bill:
OK after https://github.com/apache/geode/pull/5336 is merged to 1.13, I'll merge this one to 1.12

PR #5348, opened 2020-07-04
DRAFT: Add flag in gateway sender to not store dropped events while stopped …

albertogpz:
…set via gfhs

### For all changes:

### Note:

DO NOT REVIEW.
This PR contains a preliminary solution for the problem of queuing dropped events when the gateway sender is stopped.

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySender.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySender.java
@@ -180,6 +180,9 @@ public abstract class AbstractGatewaySender implements InternalGatewaySender, Di
 
   protected volatile ConcurrentLinkedQueue<EntryEventImpl> tmpDroppedEvents =
       new ConcurrentLinkedQueue<>();
+
+  private volatile boolean mustQueueDroppedEvents = true;
+
   /**
    * The number of seconds to wait before stopping the GatewaySender. Default is 0 seconds.
    */
@@ -911,6 +914,19 @@ public abstract class AbstractGatewaySender implements InternalGatewaySender, Di
   }
 
   @Override
+  public boolean mustQueueDroppedEvents() {
+    if (this.eventProcessor == null) {
+      return false;
+    }
+    return mustQueueDroppedEvents;
+  }
+
+  @Override
+  public void setMustQueueDroppedEvents(boolean mustQueueDroppedEvents) {
+    this.mustQueueDroppedEvents = mustQueueDroppedEvents;
+  }
+
+  @Override
   public AbstractGatewaySenderEventProcessor getEventProcessor() {
     return this.eventProcessor;
   }
@@ -1033,7 +1049,7 @@ public abstract class AbstractGatewaySender implements InternalGatewaySender, Di
 
       // If this gateway is not running, return
       if (!isRunning()) {
-        if (this.isPrimary()) {
+        if (isPrimary() && mustQueueDroppedEvents()) {
           tmpDroppedEvents.add(clonedEvent);
           if (isDebugEnabled) {
             logger.debug("add to tmpDroppedEvents for evnet {}", clonedEvent);
@@ -1136,6 +1152,7 @@ public abstract class AbstractGatewaySender implements InternalGatewaySender, Di
    * ParallelGatewaySenderQueue.addPartitionedRegionForRegion
    */
   public void enqueueTempEvents() {
+
     if (this.eventProcessor != null) {// Fix for defect #47308
       // process tmpDroppedEvents
       EntryEventImpl droppedEvent = null;
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySender.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySender.java
@@ -181,7 +181,7 @@ public abstract class AbstractGatewaySender implements InternalGatewaySender, Di
   protected volatile ConcurrentLinkedQueue<EntryEventImpl> tmpDroppedEvents =
       new ConcurrentLinkedQueue<>();
 
-  private volatile boolean mustQueueDroppedEvents = true;
+  private volatile boolean allInstancesStopped = false;
 
   /**
    * The number of seconds to wait before stopping the GatewaySender. Default is 0 seconds.
@@ -259,7 +259,7 @@ public abstract class AbstractGatewaySender implements InternalGatewaySender, Di
     this.alertThreshold = attrs.getAlertThreshold();
     this.manualStart = attrs.isManualStart();
     if (manualStart) {
-      mustQueueDroppedEvents = false;
+      allInstancesStopped = true;
     }
     this.isParallel = attrs.isParallel();
     this.groupTransactionEvents = attrs.mustGroupTransactionEvents();
@@ -917,13 +917,13 @@ public abstract class AbstractGatewaySender implements InternalGatewaySender, Di
   }
 
   @Override
-  public boolean mustQueueDroppedEvents() {
-    return mustQueueDroppedEvents;
+  public boolean allInstancesStopped() {
+    return allInstancesStopped;
   }
 
   @Override
-  public void setMustQueueDroppedEvents(boolean mustQueueDroppedEvents) {
-    this.mustQueueDroppedEvents = mustQueueDroppedEvents;
+  public void setAllInstancesStopped(boolean allInstancesStopped) {
+    this.allInstancesStopped = allInstancesStopped;
   }
 
   @Override
@@ -1049,7 +1049,7 @@ public abstract class AbstractGatewaySender implements InternalGatewaySender, Di
 
       // If this gateway is not running, return
       if (!isRunning()) {
-        if (isPrimary() && mustQueueDroppedEvents()) {
+        if (isPrimary() && !allInstancesStopped()) {
           tmpDroppedEvents.add(clonedEvent);
           if (isDebugEnabled) {
             logger.debug("add to tmpDroppedEvents for evnet {}", clonedEvent);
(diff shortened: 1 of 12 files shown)


geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/StartGatewaySenderCommand.java
```diff
@@ -155,4 +160,38 @@ public ResultModel startGatewaySender(@CliOption(key = CliStrings.START_GATEWAYS
 
     return resultModel;
   }
+
+  private boolean sendAllStartingMessage(Set<DistributedMember> dsMembers, Cache cache,
```
jinmeiliao, line 164: I wonder why you are doing it through JMX, have you tried having a function to turn on/off this flag and just execute that function on the other members? That's a common way to tell members to do something

geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySender.java
```diff
@@ -180,6 +180,9 @@
 
   protected volatile ConcurrentLinkedQueue<EntryEventImpl> tmpDroppedEvents =
       new ConcurrentLinkedQueue<>();
+
+  private volatile boolean mustQueueDroppedEvents = true;
```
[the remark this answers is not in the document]
albertogpz: Yep, I agree

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/StartGatewaySenderCommand.java
albertogpz: Do you have an example from other part of the code where it is done as you are suggesting?
Do you mean executing the function once for each member or having a function that executes on all members (the latter would seem more convenient)?

jinmeiliao: You can check here for an example: https://github.com/apache/geode/blob/eb8668ad7caa6e20d9be242ad719e173ebf244d8/geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/DestroyRegionCommand.java#L77

albertogpz: I have added a new commit with a change to function invocation instead of doing it through JMX.

geode-core/src/main/java/org/apache/geode/management/GatewaySenderMXBean.java
```diff
@@ -273,6 +273,8 @@
    */
   int getEventsExceedingAlertThreshold();
 
-
-
+  /**
```
jinmeiliao, line 276: you don't need these methods in the mbeans anymore

geode-core/src/main/java/org/apache/geode/management/internal/beans/GatewaySenderMBean.java
```diff
@@ -241,4 +241,9 @@ public boolean isConnected() {
   public int getEventsExceedingAlertThreshold() {
     return bridge.getEventsExceedingAlertThreshold();
   }
+
```
jinmeiliao, line 244: no need for this

geode-core/src/main/java/org/apache/geode/management/internal/beans/GatewaySenderMBeanBridge.java
```diff
@@ -334,4 +334,8 @@ public boolean isConnected() {
   public int getEventsExceedingAlertThreshold() {
     return getStatistic(StatsKey.GATEWAYSENDER_EVENTS_EXCEEDING_ALERT_THRESHOLD).intValue();
   }
+
```
jinmeiliao, line 337: no need

geode-core/src/main/java/org/apache/geode/management/internal/configuration/functions/GatewaySenderManageFunction.java
```diff
@@ -0,0 +1,148 @@
...
+      GatewaySender sender = cache.getGatewaySender(senderId);
+
+      if (sender == null) {
+        context.getResultSender().lastResult(new CliFunctionResult(memberName, false,
+            String.format("Sender '%s' does not exist", senderId)));
+        return;
+      }
+
+      sender.setMustQueueDroppedEvents(!stop);
+      XmlEntity xmlEntity = new XmlEntity(CacheXml.GATEWAY_SENDER, "name", senderId);
```
jinmeiliao, line 82: I don't think you need to modify the cache xml for this flag, there is no need to send this back to the command. All this function does is to turn on/off this flag and report the sucess/failure back to the command.

jinmeiliao: a good example to use is `CreateDiskStoreFunction`, that one extends CliFunction which handles the exceptions for you.

geode-core/src/main/java/org/apache/geode/management/GatewaySenderMXBean.java
albertogpz: true

geode-core/src/main/java/org/apache/geode/management/internal/beans/GatewaySenderMBean.java
albertogpz: true

geode-core/src/main/java/org/apache/geode/management/internal/beans/GatewaySenderMBeanBridge.java
albertogpz: true

geode-core/src/main/java/org/apache/geode/management/internal/configuration/functions/GatewaySenderManageFunction.java
albertogpz: I did not want to extend CliFunction because I intended this function to be used by users of the Java API as @gesterzhou was after.

In a new commit:

- I have simplified the function a bit because the first level try/catch was already taking care of any exception.

- I have also changed the name of the setter/flag from setMustQueueDroppedEvents to setAllInstancesStopped. I think it is a more understandable name without knowing the low level details about the dropped events.

@gesterzhou Would this solution be acceptable for you having fixed the problem in gfsh and at the same time having a function that could be called by Java API users?

geode-core/src/main/java/org/apache/geode/management/internal/beans/GatewaySenderMBean.java
```diff
@@ -214,11 +214,6 @@ public boolean isParallel() {
     return bridge.isParallel();
   }
 
-  @Override
-  public boolean mustGroupTransactionEvents() {
-    return bridge.mustGroupTransactionEvents();
-  }
```
gesterzhou, line 220: why you remove these parts?

jinmeiliao: these are mBean methods, he is not using jmx to turn off the flags on the servers, he is using function calls now.

geode-core/src/main/java/org/apache/geode/management/internal/configuration/functions/GatewaySenderManageFunction.java
albertogpz: @jinmeiliao Any idea why org.apache.geode.internal.cache.wan.wancommand.StopGatewaySenderCommandDUnitTest > testStopGatewaySender_Group tests are failing?

jinmeiliao: > @jinmeiliao Any idea why org.apache.geode.internal.cache.wan.wancommand.StopGatewaySenderCommandDUnitTest > testStopGatewaySender_Group tests are failing?

The function you wrote `GatewaySenderManageFunction` needs to be in the sanctioned-serializable list, because it is not, Geode rejected it when it's been serialized/deserialized across the wire (when you try to call it from the locator on all the servers). That's why "org.apache.geode.codeAnalysis.AnalyzeCoreSerializablesJUnitTest > testSerializables " is also failing with a new serializable class detected but not in the list. You should add that. Then both these test failures would go away.

albertogpz: @jinmeiliao Thanks a lot for your support!

geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySender.java
```diff
@@ -180,6 +180,9 @@
 
   protected volatile ConcurrentLinkedQueue<EntryEventImpl> tmpDroppedEvents =
       new ConcurrentLinkedQueue<>();
+
+  private volatile boolean allInstancesStopped = false;
```
gesterzhou, line 184: This name seems not correct. What you really enable/disable is actually the tmpDroppedEvents. 

I suggest to use disableTmpDroppedEvents.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/StartGatewaySenderCommand.java
```diff
@@ -77,6 +82,9 @@ public ResultModel startGatewaySender(@CliOption(key = CliStrings.START_GATEWAYS
       return ResultModel.createError(CliStrings.NO_MEMBERS_FOUND_MESSAGE);
     }
 
+    // TODO handle error situations when invoking sendAllStartingMessage
+    setNotAllGatewaySenderInstancesStopped(dsMembers, id);
```
gesterzhou, line 86: This method and the latter setAllGatewaySenderInstancesStopped() should be merged into one. And the name is confusing.

PR #5349, opened 2020-07-06
GEODE-8333: Fix PUBSUB hang

sabbeyPivotal:
PUBSUB hangs with concurrent publishers and subscribers on multiple
servers. Changed Publish executor to execute publish on background
thread (the DistributionManager's waiting pool is used). Removed separate separate subscriber group.  Added DUnit test to
recreate failure.
To preserve the order of incoming commands, even when some of them are running asynchronously in background threads, a command queue now exists. When an async commands comes in (currently only PUBLISH is async) then it goes in the queue and is started. Any other commands that come in while it is in progress are added to the a queue. Once the async command completes it also drains the command queue down to the next async command it finds in it.

Co-authored-by: Murtuza Boxwala <mboxwala@pivotal.io>

### DIFF
--- a/geode-redis/src/main/java/org/apache/geode/redis/internal/executor/pubsub/PublishExecutor.java
+++ b/geode-redis/src/main/java/org/apache/geode/redis/internal/executor/pubsub/PublishExecutor.java
@@ -34,7 +34,6 @@ public class PublishExecutor extends AbstractExecutor {
   @Override
   public RedisResponse executeCommand(Command command,
       ExecutionHandlerContext context) {
-
     executorService.submit(new PublishingRunnable(context, command));
     return null;
   }
--- a/geode-redis/src/main/java/org/apache/geode/redis/internal/executor/pubsub/PublishExecutor.java
+++ b/geode-redis/src/main/java/org/apache/geode/redis/internal/executor/pubsub/PublishExecutor.java
@@ -16,23 +16,50 @@
 package org.apache.geode.redis.internal.executor.pubsub;
 
 import java.util.List;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+
+import io.netty.buffer.ByteBuf;
 
 import org.apache.geode.redis.internal.executor.AbstractExecutor;
 import org.apache.geode.redis.internal.executor.RedisResponse;
+import org.apache.geode.redis.internal.netty.Coder;
 import org.apache.geode.redis.internal.netty.Command;
 import org.apache.geode.redis.internal.netty.ExecutionHandlerContext;
 
 public class PublishExecutor extends AbstractExecutor {
+
+  private ExecutorService executorService = Executors.newSingleThreadExecutor();
 
   @Override
   public RedisResponse executeCommand(Command command,
       ExecutionHandlerContext context) {
     List<byte[]> args = command.getProcessedCommand();
 
-    long publishCount =
-        context.getPubSub().publish(getDataRegion(context), args.get(1), args.get(2));
-
-    return RedisResponse.integer(publishCount);
+    executorService.submit(new PublishingRunnable(context, args.get(1), args.get(2)));
+    return null;
   }
 
+
+  public static class PublishingRunnable implements Runnable {
+
+    private final ExecutionHandlerContext context;
+    private final byte[] channelName;
+    private final byte[] message;
+
+    public PublishingRunnable(ExecutionHandlerContext context, byte[] channelName, byte[] message) {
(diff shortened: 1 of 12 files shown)


geode-redis/src/main/java/org/apache/geode/redis/internal/netty/ExecutionHandlerContext.java
```diff
@@ -191,38 +199,121 @@ public void channelInactive(ChannelHandlerContext ctx) {
...
-    if (command.isUnsupported() && !allowUnsupportedCommands()) {
-      writeToChannel(
-          RedisResponse.error(command.getCommandType() + RedisConstants.ERROR_UNSUPPORTED_COMMAND));
-      return;
+  public void endAsyncCommandExecution(Command command, ByteBuf response) {
+    synchronized (commandQueue) {
+      Command head = takeFromCommandQueue();
+      if (head != command) {
+        throw new IllegalStateException(
+            "expected " + command + " but found " + head + " in the queue");
```
ringles, line 216: Is there a test for this exception?

dschneider-pivotal: No. This is basically a check to make sure the implementation is correct. It should never happen if it is.
So the only way I could think of getting it to happen under test would be to have a unit test that mocks takeFromCommandQueue to return a bogus command. But I don't see much value in adding that test.

geode-redis/src/main/java/org/apache/geode/redis/internal/netty/ExecutionHandlerContext.java
```diff
@@ -113,17 +114,19 @@ public ChannelFuture writeToChannel(RedisResponse response) {
   @Override
   public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
     Command command = (Command) msg;
-    try {
-      if (logger.isDebugEnabled()) {
-        logger.debug("Executing Redis command: {}", command);
+    command.setChannelHandlerContext(ctx);
```
jdeppe-pivotal, line 117: I think this might read a bit better like this:
```
    synchronized (commandQueue) {
      boolean isEmpty = commandQueue.isEmpty();
      boolean isAsync = command.getCommandType().isAsync();
      if (!isEmpty) {
        commandQueue.offer(command);
        return;
      }
      if (!isAsync) {
        executeCommand(command);
      } else {
        commandQueue.offer(command);
        startAsyncCommandExecution(command);
      }
    }
```

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/pubsub/PublishExecutor.java
```diff
@@ -27,12 +30,34 @@
...
+    @Override
+    public void run() {
+      List<byte[]> args = command.getProcessedCommand();
+      byte[] channelName = args.get(1);
+      byte[] message = args.get(2);
+      try {
+        long publishCount =
+            context.getPubSub()
+                .publish(context.getRegionProvider().getDataRegion(), channelName, message);
+        ByteBuf response = Coder.getIntegerResponse(context.getByteBufAllocator(), publishCount);
```
jdeppe-pivotal, line 56: I'd prefer to create a `RedisResponse` so that we keep netty specifics out of the executors.

geode-redis/src/main/java/org/apache/geode/redis/internal/netty/ExecutionHandlerContext.java
dschneider-pivotal: Yes, and since isEmpty and isAsync are only used once I ended up changing it to this:
`    synchronized (commandQueue) {
      if (!commandQueue.isEmpty()) {
        commandQueue.offer(command);
        return;
      }
      if (command.getCommandType().isAsync()) {
        commandQueue.offer(command);
        startAsyncCommandExecution(command);
        return;
      }
      executeCommand(command);
    }
`

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/pubsub/PublishExecutor.java
dschneider-pivotal: Done. I also found one other place (in AbstractSubscription) that was not using RedisResponse. I change it also to do so and deleted the old writeToChannel(ByteBuf) which was no longer used after these changes.

dschneider-pivotal:
the dunit failure is GEODE-7710 and unrelated to this pull request

PR #5350, opened 2020-07-07
GEODE-8200: Rebalance operations stuck in "IN_PROGRESS" state forever

jchen21:

### For all changes:

### Note:


geode-core/src/main/java/org/apache/geode/management/internal/operation/OperationState.java
```diff
@@ -28,12 +28,25 @@
...
   private Throwable throwable;
+  private String locator;
+
+  public String getLocator() {
+    return this.locator;
+  }
+
+  public void setLocator(
+      String locator) {
+    synchronized (this) {
```
jinmeiliao, line 46: this is just one line operation, is this not atomic? If not, can we put the synchronize on the method?

geode-core/src/main/java/org/apache/geode/management/internal/operation/OperationStateStore.java
```diff
@@ -53,6 +53,8 @@
    */
   <V extends OperationResult> void recordEnd(String opId, V result, Throwable exception);
 
+  void recordLocator(String opId, String locator);
```
jinmeiliao, line 56: instead of adding this interface, you can change the method of recordStart() to add the a locator id parameter, since when started, we should always know what locator started this operation.

geode-core/src/main/java/org/apache/geode/management/internal/operation/OperationState.java
```diff
@@ -28,12 +28,25 @@
...
 public class OperationState<A extends ClusterManagementOperation<V>, V extends OperationResult>
     implements Identifiable<String> {
+  private static final long serialVersionUID = 8212319653561969588L;
   private final String opId;
   private final A operation;
   private final Date operationStart;
   private Date operationEnd;
   private V result;
   private Throwable throwable;
+  private String locator;
```
agingade, line 38: Can this be DistributedID than a String ID. That way we can avoid converting to string in other places?

geode-core/src/main/java/org/apache/geode/management/internal/operation/OperationHistoryManager.java
```diff
@@ -90,6 +95,27 @@ private static boolean isExpired(long expirationTime, OperationState<?, ?> opera
...
+              + operationState.getLocator()));
+    }
+
+    return operationState;
+  }
+
+  private boolean isLocatorOffline(OperationState operationState) {
+    if (operationState.getOperationEnd() == null
+        && (operationState.getLocator() != null)
+        && cache.getMyId().toString().compareTo(operationState.getLocator()) != 0
```
agingade, line 111: Does it need to be compared? can it be changed to "equals"....

jchen21: Good point!

geode-core/src/main/java/org/apache/geode/management/internal/operation/OperationState.java
jchen21: Do you mean `InternalDistributedMember` when you talk about `DistributedID`? I was using the `InternalDistributedMember` as the type for `locator` field. But it is a large object with many fields. We only need to identify the locator, so a String type of a member's ID should be good.

PR #5363, opened 2020-07-09
GEODE-8349: Reintroduce use of SSLSocket in cluster communications

bschuchardt:
Use of SSLEngine with NIO has proven to be less robust than using
"old-io" SSLSockets.  This PR reintroduces SSLSockets into cluster
communications and removes the use of "new IO" SSLEngine as an option.

The structure of the org.apache.geode.internal.tcp package is left
intact.  None of the old code including Connection.runOioReader() has
been brought back.  Code differences between use of NIO and OIO are
minimal and are mostly at the socket-read/socket-write level.  The
MsgReader class is modified to handle buffer manipulation that was in
the previous "old IO" implementation but was delegated to the "filter"
implementation prior to this PR.

Performance testing has shown that serialization and transmission of
messages works best with "direct" memory buffers while the reading of
messages with SSL functions best with "heap" memory buffers.
DirectChannel and DirectReplyProcessor reflect the choice of "direct"
memory buffers for writing while ClusterConnection's inputBuffer reflects
the choice of "heap" memory buffers for reading.

The cluster communications class has been renamed to ClusterConnection
because I'm very tired of the client/server Connection class having the
same name.  Every time I try to open the class with IntelliJ it
recommends the client/server class as the first option.  No more!

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/net/BufferPool.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/net/BufferPool.java
@@ -24,7 +24,7 @@ import org.apache.logging.log4j.Logger;
 import org.apache.geode.distributed.internal.DMStats;
 import org.apache.geode.distributed.internal.DistributionConfig;
 import org.apache.geode.internal.Assert;
-import org.apache.geode.internal.tcp.Connection;
+import org.apache.geode.internal.tcp.ClusterConnection;
 import org.apache.geode.logging.internal.log4j.api.LogService;
 import org.apache.geode.util.internal.GeodeGlossary;
 
@@ -65,7 +65,7 @@ public class BufferPool {
   private final ConcurrentLinkedQueue<BBSoftReference> bufferLargeQueue =
       new ConcurrentLinkedQueue<>();
 
-  private final int SMALL_BUFFER_SIZE = Connection.SMALL_BUFFER_SIZE;
+  private final int SMALL_BUFFER_SIZE = ClusterConnection.SMALL_BUFFER_SIZE;
 
 
   private final int MEDIUM_BUFFER_SIZE = DistributionConfig.DEFAULT_SOCKET_BUFFER_SIZE;
@@ -253,7 +253,7 @@ public class BufferPool {
     return newBuffer;
   }
 
-  ByteBuffer acquireDirectBuffer(BufferPool.BufferType type, int capacity) {
+  public ByteBuffer acquireDirectBuffer(BufferPool.BufferType type, int capacity) {
     switch (type) {
       case UNTRACKED:
         return ByteBuffer.allocate(capacity);
@@ -265,7 +265,7 @@ public class BufferPool {
     throw new IllegalArgumentException("Unexpected buffer type " + type.toString());
   }
 
-  ByteBuffer acquireNonDirectBuffer(BufferPool.BufferType type, int capacity) {
+  public ByteBuffer acquireNonDirectBuffer(BufferPool.BufferType type, int capacity) {
     switch (type) {
       case UNTRACKED:
         return ByteBuffer.allocate(capacity);
@@ -277,7 +277,7 @@ public class BufferPool {
     throw new IllegalArgumentException("Unexpected buffer type " + type.toString());
   }
 
-  void releaseBuffer(BufferPool.BufferType type, ByteBuffer buffer) {
+  public void releaseBuffer(BufferPool.BufferType type, ByteBuffer buffer) {
     switch (type) {
       case UNTRACKED:
         return;
--- a/geode-core/src/main/java/org/apache/geode/internal/net/NioFilter.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/net/NioFilter.java
@@ -15,8 +15,8 @@
 package org.apache.geode.internal.net;
 
 import java.io.IOException;
+import java.net.Socket;
 import java.nio.ByteBuffer;
-import java.nio.channels.SocketChannel;
 
 /**
  * Prior to transmitting a buffer or processing a received buffer
@@ -53,7 +53,7 @@ public interface NioFilter {
    * wrappedBuffer = filter.ensureWrappedCapacity(amount, wrappedBuffer, etc.);<br>
    * unwrappedBuffer = filter.readAtLeast(channel, amount, wrappedBuffer, etc.)
    */
-  ByteBuffer readAtLeast(SocketChannel channel, int amount, ByteBuffer wrappedBuffer)
+  ByteBuffer readAtLeast(int amount, ByteBuffer wrappedBuffer, Socket socket)
       throws IOException;
 
   /**
@@ -77,11 +77,8 @@ public interface NioFilter {
 
   /**
    * invoke this method when you are done using the NioFilter
-   *
    */
-  default void close(SocketChannel socketChannel) {
-    // nothing by default
-  }
+  void close(Socket socket);
 
   /**
    * returns the unwrapped byte buffer associated with the given wrapped buffer
--- a/geode-core/src/main/java/org/apache/geode/internal/net/NioFilter.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/net/NioFilter.java
@@ -15,8 +15,8 @@
 package org.apache.geode.internal.net;
 
 import java.io.IOException;
-import java.net.Socket;
 import java.nio.ByteBuffer;
+import java.nio.channels.SocketChannel;
 
 /**
  * Prior to transmitting a buffer or processing a received buffer
@@ -53,7 +53,7 @@ public interface NioFilter {
    * wrappedBuffer = filter.ensureWrappedCapacity(amount, wrappedBuffer, etc.);<br>
    * unwrappedBuffer = filter.readAtLeast(channel, amount, wrappedBuffer, etc.)
    */
-  ByteBuffer readAtLeast(int amount, ByteBuffer wrappedBuffer, Socket socket)
+  ByteBuffer readAtLeast(SocketChannel channel, int amount, ByteBuffer wrappedBuffer)
       throws IOException;
 
   /**
@@ -77,8 +77,11 @@ public interface NioFilter {
 
   /**
    * invoke this method when you are done using the NioFilter
+   *
    */
-  void close(Socket socket);
+  default void close(SocketChannel socketChannel) {
+    // nothing by default
+  }
 
   /**
    * returns the unwrapped byte buffer associated with the given wrapped buffer
--- a/geode-core/src/main/java/org/apache/geode/internal/net/NioPlainEngine.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/net/NioPlainEngine.java
@@ -17,24 +17,31 @@ package org.apache.geode.internal.net;
 
 import java.io.EOFException;
 import java.io.IOException;
+import java.io.InputStream;
+import java.net.Socket;
 import java.nio.ByteBuffer;
-import java.nio.channels.SocketChannel;
 
 import org.apache.geode.internal.Assert;
 
 /**
  * A pass-through implementation of NioFilter. Use this if you don't need
- * secure communications.
+ * secure communications. This filter can handle both old-IO and new-IO (NIO)
+ * sockets.
  */
 public class NioPlainEngine implements NioFilter {
   private final BufferPool bufferPool;
+  private final boolean useDirectBuffers;
+  private final InputStream inputStream;
 
   int lastReadPosition;
   int lastProcessedPosition;
 
 
-  public NioPlainEngine(BufferPool bufferPool) {
+  public NioPlainEngine(BufferPool bufferPool, boolean useDirectBuffers,
+      InputStream inputStream) {
     this.bufferPool = bufferPool;
+    this.useDirectBuffers = useDirectBuffers;
+    this.inputStream = inputStream;
   }
 
   @Override
@@ -54,7 +61,8 @@ public class NioPlainEngine implements NioFilter {
     ByteBuffer buffer = wrappedBuffer;
 
     if (buffer == null) {
-      buffer = bufferPool.acquireDirectBuffer(bufferType, amount);
+      buffer = useDirectBuffers ? bufferPool.acquireDirectBuffer(bufferType, amount)
+          : bufferPool.acquireNonDirectBuffer(bufferType, amount);
       buffer.clear();
       lastProcessedPosition = 0;
       lastReadPosition = 0;
@@ -71,7 +79,8 @@ public class NioPlainEngine implements NioFilter {
       ByteBuffer oldBuffer = buffer;
       oldBuffer.limit(lastReadPosition);
       oldBuffer.position(lastProcessedPosition);
-      buffer = bufferPool.acquireDirectBuffer(bufferType, amount);
+      buffer = useDirectBuffers ? bufferPool.acquireDirectBuffer(bufferType, amount)
+          : bufferPool.acquireNonDirectBuffer(bufferType, amount);
       buffer.clear();
       buffer.put(oldBuffer);
       bufferPool.releaseBuffer(bufferType, oldBuffer);
@@ -82,9 +91,8 @@ public class NioPlainEngine implements NioFilter {
   }
 
   @Override
-  public ByteBuffer readAtLeast(SocketChannel channel, int bytes, ByteBuffer wrappedBuffer)
+  public ByteBuffer readAtLeast(int bytes, ByteBuffer buffer, Socket socket)
       throws IOException {
-    ByteBuffer buffer = wrappedBuffer;
 
     Assert.assertTrue(buffer.capacity() - lastProcessedPosition >= bytes);
 
@@ -93,7 +101,7 @@ public class NioPlainEngine implements NioFilter {
     buffer.position(lastReadPosition);
 
     while (buffer.position() < (lastProcessedPosition + bytes)) {
-      int amountRead = channel.read(buffer);
+      int amountRead = SocketUtils.readFromSocket(socket, buffer, inputStream);
       if (amountRead < 0) {
         throw new EOFException();
       }
@@ -111,13 +119,9 @@ public class NioPlainEngine implements NioFilter {
     return buffer;
   }
 
-  public void doneReading(ByteBuffer unwrappedBuffer) {
-    if (unwrappedBuffer.position() != 0) {
-      unwrappedBuffer.compact();
-    } else {
-      unwrappedBuffer.position(unwrappedBuffer.limit());
-      unwrappedBuffer.limit(unwrappedBuffer.capacity());
-    }
+  @Override
+  public void close(Socket socket) {
+    // nothing needs to be done in this implementation
   }
 
   @Override
--- a/geode-core/src/main/java/org/apache/geode/internal/net/NioPlainEngine.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/net/NioPlainEngine.java
@@ -17,31 +17,24 @@ package org.apache.geode.internal.net;
 
 import java.io.EOFException;
 import java.io.IOException;
-import java.io.InputStream;
-import java.net.Socket;
 import java.nio.ByteBuffer;
+import java.nio.channels.SocketChannel;
 
 import org.apache.geode.internal.Assert;
 
 /**
  * A pass-through implementation of NioFilter. Use this if you don't need
- * secure communications. This filter can handle both old-IO and new-IO (NIO)
- * sockets.
+ * secure communications.
  */
 public class NioPlainEngine implements NioFilter {
   private final BufferPool bufferPool;
-  private final boolean useDirectBuffers;
-  private final InputStream inputStream;
 
   int lastReadPosition;
   int lastProcessedPosition;
 
 
-  public NioPlainEngine(BufferPool bufferPool, boolean useDirectBuffers,
-      InputStream inputStream) {
+  public NioPlainEngine(BufferPool bufferPool) {
     this.bufferPool = bufferPool;
-    this.useDirectBuffers = useDirectBuffers;
-    this.inputStream = inputStream;
   }
 
   @Override
@@ -61,8 +54,7 @@ public class NioPlainEngine implements NioFilter {
     ByteBuffer buffer = wrappedBuffer;
 
     if (buffer == null) {
-      buffer = useDirectBuffers ? bufferPool.acquireDirectBuffer(bufferType, amount)
-          : bufferPool.acquireNonDirectBuffer(bufferType, amount);
+      buffer = bufferPool.acquireDirectBuffer(bufferType, amount);
       buffer.clear();
       lastProcessedPosition = 0;
       lastReadPosition = 0;
@@ -79,8 +71,7 @@ public class NioPlainEngine implements NioFilter {
       ByteBuffer oldBuffer = buffer;
       oldBuffer.limit(lastReadPosition);
       oldBuffer.position(lastProcessedPosition);
-      buffer = useDirectBuffers ? bufferPool.acquireDirectBuffer(bufferType, amount)
-          : bufferPool.acquireNonDirectBuffer(bufferType, amount);
+      buffer = bufferPool.acquireDirectBuffer(bufferType, amount);
       buffer.clear();
       buffer.put(oldBuffer);
       bufferPool.releaseBuffer(bufferType, oldBuffer);
@@ -91,8 +82,9 @@ public class NioPlainEngine implements NioFilter {
   }
 
   @Override
-  public ByteBuffer readAtLeast(int bytes, ByteBuffer buffer, Socket socket)
+  public ByteBuffer readAtLeast(SocketChannel channel, int bytes, ByteBuffer wrappedBuffer)
       throws IOException {
+    ByteBuffer buffer = wrappedBuffer;
 
     Assert.assertTrue(buffer.capacity() - lastProcessedPosition >= bytes);
 
@@ -101,7 +93,7 @@ public class NioPlainEngine implements NioFilter {
     buffer.position(lastReadPosition);
 
     while (buffer.position() < (lastProcessedPosition + bytes)) {
-      int amountRead = SocketUtils.readFromSocket(socket, buffer, inputStream);
+      int amountRead = channel.read(buffer);
       if (amountRead < 0) {
         throw new EOFException();
       }
@@ -119,9 +111,13 @@ public class NioPlainEngine implements NioFilter {
     return buffer;
   }
 
-  @Override
-  public void close(Socket socket) {
-    // nothing needs to be done in this implementation
+  public void doneReading(ByteBuffer unwrappedBuffer) {
+    if (unwrappedBuffer.position() != 0) {
+      unwrappedBuffer.compact();
+    } else {
+      unwrappedBuffer.position(unwrappedBuffer.limit());
+      unwrappedBuffer.limit(unwrappedBuffer.capacity());
+    }
   }
 
   @Override
--- a/geode-core/src/main/java/org/apache/geode/internal/net/SocketUtils.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/net/SocketUtils.java
@@ -16,8 +16,16 @@
 package org.apache.geode.internal.net;
 
 import java.io.IOException;
+import java.io.InputStream;
 import java.net.ServerSocket;
 import java.net.Socket;
+import java.nio.ByteBuffer;
+
+import javax.net.ssl.SSLSocket;
+
+import org.apache.logging.log4j.Logger;
+
+import org.apache.geode.logging.internal.log4j.api.LogService;
 
 /**
  * The SocketUtils class is utility class for performing operations on Sockets and ServerSockets.
@@ -28,6 +36,7 @@ import java.net.Socket;
  * @since GemFire 7.0
  */
 public abstract class SocketUtils {
+  private static final Logger logger = LogService.getLogger();
 
   /**
    * Closes the specified Socket silently ignoring any IOException, guarding against null Object
@@ -72,5 +81,50 @@ public abstract class SocketUtils {
 
     return true;
   }
+
+  /**
+   * Read data from the given socket into the given ByteBuffer. If NIO is supported
+   * we use Channel.read(ByteBuffer). If not we use byte arrays to read available
+   * bytes or buffer.remaining() bytes, whichever is smaller. If buffer.limit is zero
+   * and buffer.remaining is also zero the limit is changed to be buffer.capacity
+   * before reading.
+   *
+   * @return the number of bytes read, which may be -1 for EOF
+   */
+  public static int readFromSocket(Socket socket, ByteBuffer inputBuffer,
+      InputStream socketInputStream) throws IOException {
+    int amountRead;
+    inputBuffer.limit(inputBuffer.capacity());
+    if (socket instanceof SSLSocket) {
+      amountRead = readFromStream(socketInputStream, inputBuffer);
+    } else {
+      amountRead = socket.getChannel().read(inputBuffer);
+    }
+    return amountRead;
+  }
+
+  private static int readFromStream(InputStream stream, ByteBuffer inputBuffer) throws IOException {
+    int amountRead;
+    // if bytes are available we read that number of bytes. Otherwise we do a blocking read
+    // of buffer.remaining() bytes
+    int amountToRead = inputBuffer.remaining();
+    // stream.available() > 0 ? Math.min(stream.available(), inputBuffer.remaining())
+    // : inputBuffer.remaining();
+    if (inputBuffer.hasArray()) {
+      amountRead = stream.read(inputBuffer.array(),
+          inputBuffer.arrayOffset() + inputBuffer.position(), amountToRead);
+      if (amountRead > 0) {
+        inputBuffer.position(inputBuffer.position() + amountRead);
+      }
+    } else {
+      byte[] bytesRead = new byte[amountToRead];
+      amountRead = stream.read(bytesRead);
+      if (amountRead > 0) {
+        inputBuffer.put(bytesRead, 0, amountRead);
+      }
+    }
+    return amountRead;
+  }
+
 
 }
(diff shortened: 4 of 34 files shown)


geode-core/src/main/java/org/apache/geode/internal/net/NioPlainEngine.java
```diff
@@ -17,24 +17,31 @@
 
 import java.io.EOFException;
```
echobravopapa, line 18: should these classes be renamed since we are using OiO now?

geode-core/src/main/java/org/apache/geode/internal/net/BufferPool.java
```diff
@@ -253,7 +253,7 @@ ByteBuffer expandWriteBufferIfNeeded(BufferType type, ByteBuffer existing,
     return newBuffer;
   }
 
-  ByteBuffer acquireDirectBuffer(BufferPool.BufferType type, int capacity) {
+  public ByteBuffer acquireDirectBuffer(BufferPool.BufferType type, int capacity) {
```
echobravopapa, line 256: what required the addition of `public`?

geode-core/src/main/java/org/apache/geode/internal/tcp/ClusterConnection.java
```diff
@@ -94,11 +96,14 @@
  *
  * @since GemFire 2.0
  */
-public class Connection implements Runnable {
+public class ClusterConnection implements Runnable {
```
echobravopapa, line 99: +1 on the rename

geode-core/src/main/java/org/apache/geode/internal/tcp/ClusterConnection.java
```diff
@@ -140,7 +145,7 @@
   private final ConnectionTable owner;
 
   private final TCPConduit conduit;
-  private NioFilter ioFilter;
+  // private NioFilter ioFilter;
```
echobravopapa, line 148: dead code...

geode-core/src/main/java/org/apache/geode/internal/tcp/ClusterConnection.java
```diff
@@ -1142,31 +1155,47 @@ private Connection(ConnectionTable t, boolean preserveOrder, InternalDistributed
...
-    owner.addConnectingSocket(channel.socket(), addr.getAddress());
-
-    try {
-      channel.socket().setTcpNoDelay(true);
-      channel.socket().setKeepAlive(SocketCreator.ENABLE_TCP_KEEP_ALIVE);
 
+    int connectTime = getP2PConnectTimeout(conduit.getDM().getConfig());
+    boolean useSSL = getConduit().useSSL();
+    if (useSSL) {
+      // int socketBufferSize = -1;
```
echobravopapa, line 1162: dead code...

geode-core/src/main/java/org/apache/geode/internal/tcp/ClusterConnection.java
```diff
@@ -1142,31 +1155,47 @@ private Connection(ConnectionTable t, boolean preserveOrder, InternalDistributed
...
-
-        createIoFilter(channel, true);
+        if (!useSSL) {
+          // haven't connected yet
+          socket.connect(addr, connectTime);
+        }
+        configureInputStream(socket, true);
 
       } catch (NullPointerException e) {
         // jdk 1.7 sometimes throws an NPE here
```
echobravopapa, line 1201: out of scope, but kinda reads like tech debt...

geode-core/src/main/java/org/apache/geode/internal/tcp/ClusterConnection.java
```diff
@@ -2588,23 +2596,44 @@ void writeFully(SocketChannel channel, ByteBuffer buffer, boolean forceAsync,
...
+            if (socket instanceof SSLSocket) {
+              OutputStream output = socket.getOutputStream();
+              if (buffer.hasArray()) {
+                output.write(buffer.array(), buffer.arrayOffset(),
+                    buffer.limit() - buffer.position());
+                buffer.position(buffer.limit());
+              } else {
+                // socket output streams are FileOutputStreams and have a writeable Channel.
+                // This code merely fetches that channel and writes to it.
+                // Channels.newChannel(output).write(buffer);
```
echobravopapa, line 2612: dead code...

geode-core/src/main/java/org/apache/geode/internal/tcp/ClusterConnection.java
```diff
@@ -2722,9 +2753,10 @@ public void readAck(final DirectReplyProcessor processor)
    * deserialized and passed to TCPConduit for further processing
    */
   private void processInputBuffer() throws ConnectionException, IOException {
+    // BRUCE: simplify this
```
echobravopapa, line 2756: do we need a story for this TODO?

echobravopapa: or a git-hook ;)

geode-core/src/main/java/org/apache/geode/internal/tcp/ClusterConnection.java
```diff
@@ -2909,13 +2950,13 @@ private boolean readMessageHeader(ByteBuffer peerDataBuffer) throws IOException
       readerShuttingDown = true;
       requestClose(String.format("Unknown P2P message type: %s",
           nioMessageTypeInteger));
-      return true;
+      return false;
```
echobravopapa, line 2953: not obvious why this bool is flipped...

geode-core/src/main/java/org/apache/geode/internal/tcp/ClusterConnection.java
```diff
@@ -3006,7 +3047,7 @@ private void readMessage(ByteBuffer peerDataBuffer) {
       } catch (IOException ex) {
         // ignored
       }
-    } else /* (nioMessageType == END_CHUNKED_MSG_TYPE) */ {
+    } else /* (messageType == END_CHUNKED_MSG_TYPE) */ {
```
echobravopapa, line 3050: looks dead, maybe there was `else if` once upon a time...

geode-core/src/main/java/org/apache/geode/internal/tcp/MsgReader.java
```diff
@@ -14,70 +14,66 @@
...
-
-      boolean directAck = (nioMessageType & Connection.DIRECT_ACK_BIT) != 0;
-      if (directAck) {
-        nioMessageType &= ~Connection.DIRECT_ACK_BIT; // clear the ack bit
-      }
+    Assert.assertTrue(buffer.remaining() >= ClusterConnection.MSG_HEADER_BYTES);
 
-      header.setFields(nioMessageLength, nioMessageType, nioMsgId);
+    int messageLength = buffer.getInt();
+    /* nioMessageVersion = */
```
echobravopapa, line 63: dead?

geode-core/src/main/java/org/apache/geode/internal/tcp/TCPConduit.java
```diff
@@ -202,6 +202,8 @@
 
   private final Stopper stopper = new Stopper();
 
+  private boolean enableTLSOverNIO = true; // Boolean.getBoolean("geode.enable-tls-nio");
+
```
echobravopapa, line 206: dead code hanging out...

geode-core/src/main/java/org/apache/geode/internal/tcp/TCPConduit.java
```diff
@@ -945,6 +962,10 @@ public boolean useSSL() {
     return useSSL;
   }
 
+  public boolean useDirectReceiveBuffers() {
+    return !useSSL();
+  }
+
```
echobravopapa, line 968: previous question might be resolved following this accessor... for ssl/tls we use NIO and !ssl we use the directbuffers, cool and with 6 you get eggroll

geode-core/src/main/java/org/apache/geode/internal/tcp/ClusterConnection.java
bschuchardt: I'll remove that

geode-core/src/main/java/org/apache/geode/internal/net/BufferPool.java
bschuchardt: it's now accessed by the org.apache.geode.internal.tcp package

geode-core/src/main/java/org/apache/geode/internal/net/NioPlainEngine.java
bschuchardt: Those classes are no longer used by the tcp package, but will be used by client/server.  I don't think they should be renamed.

geode-core/src/main/java/org/apache/geode/internal/tcp/ClusterConnection.java
bschuchardt: I'm removing that

geode-core/src/main/java/org/apache/geode/internal/tcp/ClusterConnection.java
bschuchardt: The old meaning didn't make a lot of sense.  Returning _false_ when the header is successfully read?  No.  That's why I changed it.

geode-core/src/main/java/org/apache/geode/internal/tcp/MsgReader.java
bschuchardt: hmm, that was already dead code.  I'll remove that but I think I should rename the calcMsgVersion method

geode-core/src/main/java/org/apache/geode/internal/tcp/TCPConduit.java
bschuchardt: ooh - thanks

pivotal-jbarrett:
Can you please post the results the geode-benchmarks running this branch against develop.

bschuchardt:
> org.apache.geode.benchmark.tests.PartitionedGetBenchmark
>              average ops/second  Baseline:    331221.55  Test:    342201.08  Difference:   +3.3%
>       ops/second standard error  Baseline:       521.34  Test:       555.60  Difference:   +6.6%
>   ops/second standard deviation  Baseline:      9014.73  Test:      9607.16  Difference:   +6.6%
>      YS 99th percentile latency  Baseline:     20071.00  Test:     20071.00  Difference:   +0.0%
>                  median latency  Baseline:    362239.00  Test:    361471.00  Difference:   -0.2%
>         90th percentile latency  Baseline:   7036927.00  Test:   6995967.00  Difference:   -0.6%
>         99th percentile latency  Baseline:  24428543.00  Test:  23085055.00  Difference:   -5.5%
>       99.9th percentile latency  Baseline:  34963455.00  Test:  33259519.00  Difference:   -4.9%
>                 average latency  Baseline:   2171050.56  Test:   2101421.48  Difference:   -3.2%
>      latency standard deviation  Baseline:   5334288.43  Test:   5076754.60  Difference:   -4.8%
>          latency standard error  Baseline:       535.20  Test:       501.11  Difference:   -6.4%


>org.apache.geode.benchmark.tests.PartitionedPutBenchmark
>              average ops/second  Baseline:    178773.53  Test:    182095.99  Difference:   +1.9%
>       ops/second standard error  Baseline:       599.58  Test:       835.58  Difference:  +39.4%
>   ops/second standard deviation  Baseline:     10367.72  Test:     14448.49  Difference:  +39.4%
>      YS 99th percentile latency  Baseline:     20083.00  Test:     20082.80  Difference:   -0.0%
>                  median latency  Baseline:    765951.00  Test:    767999.00  Difference:   +0.3%
>         90th percentile latency  Baseline:  13459455.00  Test:  13107199.00  Difference:   -2.6%
>         99th percentile latency  Baseline:  28983295.00  Test:  27787263.00  Difference:   -4.1%
>       99.9th percentile latency  Baseline: 107151359.00  Test: 109379583.00  Difference:   +2.1%
>                 average latency  Baseline:   4024484.55  Test:   3951012.79  Difference:   -1.8%
>      latency standard deviation  Baseline:   9229899.71  Test:   8986494.06  Difference:   -2.6%
>          latency standard error  Baseline:      1260.59  Test:      1216.02  Difference:   -3.5%


>org.apache.geode.benchmark.tests.ReplicatedGetBenchmark
>              average ops/second  Baseline:    339091.57  Test:    351035.30  Difference:   +3.5%
>       ops/second standard error  Baseline:       450.59  Test:       371.17  Difference:  -17.6%
>   ops/second standard deviation  Baseline:      7791.42  Test:      6418.07  Difference:  -17.6%
>      YS 99th percentile latency  Baseline:     20071.00  Test:     20071.33  Difference:   +0.0%
>                  median latency  Baseline:    355839.00  Test:    357375.00  Difference:   +0.4%
>         90th percentile latency  Baseline:   6524927.00  Test:   5992447.00  Difference:   -8.2%
>         99th percentile latency  Baseline:  24526847.00  Test:  23691263.00  Difference:   -3.4%
>       99.9th percentile latency  Baseline:  34013183.00  Test:  32948223.00  Difference:   -3.1%
>                 average latency  Baseline:   2120692.40  Test:   2048843.90  Difference:   -3.4%
>      latency standard deviation  Baseline:   5263600.14  Test:   5054339.01  Difference:   -4.0%
>          latency standard error  Baseline:       521.94  Test:       492.61  Difference:   -5.6%


>org.apache.geode.benchmark.tests.ReplicatedPutBenchmark
>              average ops/second  Baseline:    192443.22  Test:    194403.34  Difference:   +1.0%
>       ops/second standard error  Baseline:       726.17  Test:       788.60  Difference:   +8.6%
>   ops/second standard deviation  Baseline:     12556.67  Test:     13636.14  Difference:   +8.6%
>      YS 99th percentile latency  Baseline:     20060.00  Test:     20072.00  Difference:   +0.1%
>                  median latency  Baseline:    809471.00  Test:    781823.00  Difference:   -3.4%
>         90th percentile latency  Baseline:  10690559.00  Test:  11599871.00  Difference:   +8.5%
>         99th percentile latency  Baseline:  23724031.00  Test:  24379391.00  Difference:   +2.8%
>       99.9th percentile latency  Baseline: 106692607.00  Test: 110166015.00  Difference:   +3.3%
>                 average latency  Baseline:   3739364.07  Test:   3701351.14  Difference:   -1.0%
>      latency standard deviation  Baseline:   8255312.29  Test:   8420063.75  Difference:   +2.0%
>          latency standard error  Baseline:      1086.76  Test:      1102.91  Difference:   +1.5%

geode-core/src/main/java/org/apache/geode/internal/tcp/ClusterConnection.java
bschuchardt: I think that's just an informative comment.  It sets the context for understanding that block of code, so I don't think it should be removed.

pivotal-jbarrett:
In response to the benchmark results, the values are within the margin of error for benchmarks. The get benchmarks would not much if any P2P operations. Put benchmarks would have 1 P2P op per client/server put op and likely bottlenecked by the p2p operation. There are currently no benchmarks that measure P2P operations in isolation.

geode-core/src/main/java/org/apache/geode/internal/tcp/ClusterConnection.java
```diff
@@ -1142,31 +1154,46 @@ private Connection(ConnectionTable t, boolean preserveOrder, InternalDistributed
...
       try {
 
-        channel.socket().connect(addr, connectTime);
-
-        createIoFilter(channel, true);
+        if (!useSSL) {
+          // haven't connected yet
+          socket.connect(addr, connectTime);
+        }
+        configureInputStream(socket, true);
```
Bill, line 1196: my read of `configureInputStream()` is that it initializes the `inputStream` field and does the TLS handshake if one is needed. That implies that a connect call happens in that case.

inasmuch as `configureInputStream()` causes connect for the TLS case, would it make sense for it to also handle the connect for the non-TLS case too? just thinking of ways to reduce the number of branches on `getConduit().useSSL()`

ugh but now I see `configureInputStream()` only initiates the TLS handshake if `!clientSocket`. Why does it only initiate the handshake for "server" sockets?

geode-core/src/main/java/org/apache/geode/internal/net/NioFilter.java
```diff
@@ -53,7 +53,7 @@ ByteBuffer ensureWrappedCapacity(int amount, ByteBuffer wrappedBuffer,
    * wrappedBuffer = filter.ensureWrappedCapacity(amount, wrappedBuffer, etc.);<br>
    * unwrappedBuffer = filter.readAtLeast(channel, amount, wrappedBuffer, etc.)
    */
-  ByteBuffer readAtLeast(SocketChannel channel, int amount, ByteBuffer wrappedBuffer)
+  ByteBuffer readAtLeast(int amount, ByteBuffer wrappedBuffer, Socket socket)
```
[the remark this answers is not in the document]
bschuchardt: that's explained in my initial comment

geode-core/src/main/java/org/apache/geode/internal/tcp/MsgReader.java
```diff
@@ -14,70 +14,65 @@
...
-
-  protected final Connection conn;
+  protected final ClusterConnection conn;
+  private final BufferPool bufferPool;
   protected final Header header = new Header();
-  private final NioFilter ioFilter;
   private ByteBuffer peerNetData;
+  private final InputStream inputStream;
   private final ByteBufferInputStream byteBufferInputStream;
+  private int lastProcessedPosition;
```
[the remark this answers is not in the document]
bschuchardt: It is the position of the last processed byte in the buffer and lets us know which byte should be processed next.

geode-core/src/main/java/org/apache/geode/internal/tcp/MsgReader.java
```diff
@@ -14,70 +14,65 @@
...
+  private int lastReadPosition;
```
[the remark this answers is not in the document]
bschuchardt: Both of those variables are state used in readAtLeast()
```
    // keep track of how much of the buffer contains valid data with lastReadPosition
    lastReadPosition = buffer.position();

    // set up the buffer for reading and keep track of how much has been consumed with
    // lastProcessedPosition
    buffer.limit(lastProcessedPosition + bytes);
    buffer.position(lastProcessedPosition);
    lastProcessedPosition += bytes;
```

geode-core/src/main/java/org/apache/geode/internal/net/SocketUtils.java
```diff
@@ -73,4 +82,49 @@ public static boolean close(final ServerSocket serverSocket) {
...
 
+  /**
+   * Read data from the given socket into the given ByteBuffer. If NIO is supported
+   * we use Channel.read(ByteBuffer). If not we use byte arrays to read available
+   * bytes or buffer.remaining() bytes, whichever is smaller. If buffer.limit is zero
+   * and buffer.remaining is also zero the limit is changed to be buffer.capacity
+   * before reading.
+   *
+   * @return the number of bytes read, which may be -1 for EOF
+   */
```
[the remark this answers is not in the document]
bschuchardt: ok

geode-core/src/main/java/org/apache/geode/internal/net/SocketUtils.java
```diff
@@ -73,4 +82,49 @@ public static boolean close(final ServerSocket serverSocket) {
...
+  public static int readFromSocket(Socket socket, ByteBuffer inputBuffer,
+      InputStream socketInputStream) throws IOException {
+    int amountRead;
+    inputBuffer.limit(inputBuffer.capacity());
+    if (socket instanceof SSLSocket) {
+      amountRead = readFromStream(socketInputStream, inputBuffer);
+    } else {
+      amountRead = socket.getChannel().read(inputBuffer);
+    }
```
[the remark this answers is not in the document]
bschuchardt: yes, performance is worse for non-TLS if the stream is used

geode-core/src/main/java/org/apache/geode/internal/net/SocketUtils.java
```diff
@@ -73,4 +82,49 @@ public static boolean close(final ServerSocket serverSocket) {
...
+    int amountToRead = inputBuffer.remaining();
+    // stream.available() > 0 ? Math.min(stream.available(), inputBuffer.remaining())
+    // : inputBuffer.remaining();
+    if (inputBuffer.hasArray()) {
+      amountRead = stream.read(inputBuffer.array(),
+          inputBuffer.arrayOffset() + inputBuffer.position(), amountToRead);
+      if (amountRead > 0) {
+        inputBuffer.position(inputBuffer.position() + amountRead);
+      }
+    } else {
```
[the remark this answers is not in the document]
bschuchardt: HeapByteBuffer has an array.  DirectByteBuffer does not have an array.

geode-core/src/main/java/org/apache/geode/internal/tcp/ClusterConnection.java
```diff
@@ -1142,31 +1154,46 @@ private Connection(ConnectionTable t, boolean preserveOrder, InternalDistributed
...
 
+    int connectTime = getP2PConnectTimeout(conduit.getDM().getConfig());
+    boolean useSSL = getConduit().useSSL();
+    if (useSSL) {
+      int socketBufferSize =
+          sharedResource ? SMALL_BUFFER_SIZE : this.owner.getConduit().tcpBufferSize;
+      socket = getConduit().getSocketCreator().forAdvancedUse().connect(
+          new HostAndPort(remoteID.getHostName(), remoteID.getDirectChannelPort()),
+          0, null, false, socketBufferSize, true);
+      setSocketBufferSize(this.socket, false, socketBufferSize, true);
```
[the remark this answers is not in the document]
bschuchardt: I'll add a comment there.  Performance degrades if you mess with that buffer size in a SSLSocket, but we want to set the local buffer-size variable for message streamers.

geode-core/src/main/java/org/apache/geode/internal/tcp/ClusterConnection.java
bschuchardt: The TLS handshake doesn't imply a connect().  The socket is already connected at this point but hasn't performed the TLS handshake.  This method is used both for "sender" connections and "receiver" connections.  In the latter case the socket is already connected so it wouldn't make sense to move the non-TLS connect into this method.

geode-core/src/main/java/org/apache/geode/internal/net/NioFilter.java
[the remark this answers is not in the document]
bschuchardt: on second thought, I'm going to undo those changes.  My first cut at a solution modified the filters to allow their use with an SSLSocket and its input stream, but I decided that it was better to stop using them at all.

bschuchardt:
I've written some geode-benchmark tests to vet this PR.  The tests add a new ClusterTopology to the benchmarks so that there are more servers and workload operations are distributed among the servers.  I wrote a replicated region `get` benchmark but since all data was local it was pretty meaningless and so I removed it.

```
org.apache.geode.benchmark.tests.PartitionedGetBenchmark
              average ops/second  Baseline:    415568.87  Test:    421152.13  Difference:   +1.3%
       ops/second standard error  Baseline:      1137.52  Test:      1202.93  Difference:   +5.7%
   ops/second standard deviation  Baseline:     34087.75  Test:     36047.68  Difference:   +5.7%
      YS 99th percentile latency  Baseline:     20007.00  Test:     20007.33  Difference:   +0.0%
                  median latency  Baseline:      3619.00  Test:      3539.00  Difference:   -2.2%
         90th percentile latency  Baseline:    822271.00  Test:    823295.00  Difference:   +0.1%
         99th percentile latency  Baseline:  41058303.00  Test:  41746431.00  Difference:   +1.7%
       99.9th percentile latency  Baseline:  96010239.00  Test:  95289343.00  Difference:   -0.8%
                 average latency  Baseline:   1732279.46  Test:   1708445.36  Difference:   -1.4%
      latency standard deviation  Baseline:   8040574.12  Test:   8054977.95  Difference:   +0.2%
          latency standard error  Baseline:       416.09  Test:       413.94  Difference:   -0.5%
org.apache.geode.benchmark.tests.PartitionedPutBenchmark
              average ops/second  Baseline:     65766.06  Test:     68103.19  Difference:   +3.6%
       ops/second standard error  Baseline:       280.40  Test:       337.96  Difference:  +20.5%
   ops/second standard deviation  Baseline:      8379.10  Test:     10093.67  Difference:  +20.5%
      YS 99th percentile latency  Baseline:     20080.54  Test:     20080.31  Difference:   -0.0%
                  median latency  Baseline:    628735.00  Test:    647679.00  Difference:   +3.0%
         90th percentile latency  Baseline:  31834111.00  Test:  29949951.00  Difference:   -5.9%
         99th percentile latency  Baseline: 173539327.00  Test: 176160767.00  Difference:   +1.5%
       99.9th percentile latency  Baseline: 251789311.00  Test: 271581183.00  Difference:   +7.9%
                 average latency  Baseline:  11008644.65  Test:  10651543.96  Difference:   -3.2%
      latency standard deviation  Baseline:  31209391.24  Test:  30946703.39  Difference:   -0.8%
          latency standard error  Baseline:      4070.65  Test:      3970.15  Difference:   -2.5%
org.apache.geode.benchmark.tests.ReplicatedPutBenchmark
              average ops/second  Baseline:     36802.56  Test:     36020.12  Difference:   -2.1%
       ops/second standard error  Baseline:       579.41  Test:       638.69  Difference:  +10.2%
   ops/second standard deviation  Baseline:     16842.76  Test:     18599.04  Difference:  +10.4%
      YS 99th percentile latency  Baseline:     20085.94  Test:     20085.82  Difference:   -0.0%
                  median latency  Baseline:    563711.00  Test:    641535.00  Difference:  +13.8%
         90th percentile latency  Baseline:  32653311.00  Test:  29753343.00  Difference:   -8.9%
         99th percentile latency  Baseline: 507772927.00  Test: 596639743.00  Difference:  +17.5%
       99.9th percentile latency  Baseline: 758644735.00  Test: 882376703.00  Difference:  +16.3%
                 average latency  Baseline:  20738470.30  Test:  20930653.55  Difference:   +0.9%
      latency standard deviation  Baseline:  82082668.05  Test:  90329383.31  Difference:  +10.0%
          latency standard error  Baseline:     14695.56  Test:     16252.34  Difference:  +10.6%

```

geode-core/src/test/java/org/apache/geode/internal/net/SocketUtilsJUnitTest.java
```diff
@@ -88,4 +95,86 @@ public void testCloseServerSocketThrowsIOException() throws IOException {
...
+    assertThat(stream.available()).isEqualTo(bytes.length);
+  }
+
+
+  @Test
+  public void readFromSocketWithDirectBuffer() throws IOException {
+    Socket socket = mock(Socket.class);
+    SocketChannel channel = mock(SocketChannel.class);
+    when(socket.getChannel()).thenReturn(channel);
+    final ByteBuffer buffer = ByteBuffer.allocateDirect(100); // non-heap buffer
```
Bill, line 127: Great to see this increased code coverage!

From scanning this method and the previous one I think the only two differences are 1) the allocation statement and 2) the verification of `buffer.hasArray()`. It would aid maintenance, I think, if the parts that are identical were factored out.

geode-core/src/test/java/org/apache/geode/internal/net/SocketUtilsJUnitTest.java
```diff
@@ -88,4 +95,86 @@ public void testCloseServerSocketThrowsIOException() throws IOException {
...
+    byte[] bytes = new byte[100];
+    InputStream stream = new ByteArrayInputStream(bytes);
+    assertThat(stream.available()).isEqualTo(bytes.length);
+    SocketUtils.readFromSocket(socket, buffer, stream);
+    // the channel was not used
+    verify(channel, times(0)).read(buffer);
+    // the buffer was filled
+    assertThat(buffer.position()).isEqualTo(bytes.length);
+    // the stream was used to fill the buffer
+    assertThat(stream.available()).isZero();
```
Bill, line 178: I recommend factoring common code out of these two test methods. In fact all four methods have a common shape.

geode-core/src/main/java/org/apache/geode/internal/net/SocketUtils.java
```diff
@@ -89,6 +89,10 @@ public static boolean close(final ServerSocket serverSocket) {
    * and buffer.remaining is also zero the limit is changed to be buffer.capacity
    * before reading.
    *
+   * @param socket the socket to read from
+   * @param inputBuffer the buffer into which data should be stored
+   * @param socketInputStream the socket's inputStream, included as a parameter so it can be a
+   *        buffered stream, if desired
```
Bill, line 95: "if desired" by whom—by what part of the code? The comment implies the caller controls whether or not the stream is used, perhaps by making it `null` or not. Reality is this parameter is dependent on the kind of object referenced by the `socket` parameter.

Any time I see a dependency between parameters like this, especially when it comes with a switch-on-type it makes me wonder if there is a mis-assignment of responsibility somewhere. Is there already some object that could hold the stream? If so, could we delegate this behavior to that object?

geode-core/src/main/java/org/apache/geode/internal/tcp/MsgReader.java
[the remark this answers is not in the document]
Bill: My comment came out of frustration, perhaps born of laziness. I'm trying to decide whether or not I think the new code in `MsgReader` is correct—by inspection. That's very hard for me.

Earlier in the day I went and read the NIO `Buffer` code since I needed to understand that code to understand these changes. That class has [Jon Bentley](https://www.goodreads.com/book/show/52084.Programming_Pearls)-level comments detailing the purpose of each field along with the field invariants. The documentation of those invariants are a huge aid in visual inspection. 

https://docs.oracle.com/javase/8/docs/api/java/nio/Buffer.html

So with this in mind:

1. my hunch is that `Buffer` already has all the state we need to solve this problem—so I am suspicious of these added fields
2. it feels like we're doing a lot of poking into the `Buffer` object to control its details—setting limit and position explicitly rather than putting and getting and flipping

If the extra fields are really needed then it would help to see a block comment at the point of declaration, tying together these variables with their invariants. Rather than saying things like "keep track of", it would be helpful to know the relationships between these two variables and between them and the various state variables inside the `Buffer`.

geode-core/src/main/java/org/apache/geode/internal/net/NioFilter.java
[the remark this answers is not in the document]
Bill: ah I failed to understand that both these things are true at once:

"This PR … removes the use of "new IO" SSLEngine as an option."

and

"There are a few changes to the NIO "engine" implementations that I made
while testing performance that I decided to keep because they allow the
use of NioPlainEngine with SSLSockets. That could be an advantage when
performing similar work in client/server communications"

If I understand: implication of the second bit is that we didn't remove SSLEngine from the codebase—we only eliminated it in the P2P messaging path.

geode-core/src/main/java/org/apache/geode/internal/tcp/ClusterConnection.java
Bill: In `configureInputStream(Socket socket, boolean clientSocket)` is `clientSocket` a misnomer then? Does that really mean `receiverSocket`?

Remind me: are these sockets full-duplex or half-duplex?

geode-core/src/main/java/org/apache/geode/internal/tcp/MsgReader.java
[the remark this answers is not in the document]
bschuchardt: I'm not going to change that logic.  It's what used to be in the MsgReader subclasses before I moved it to the NioFilter classes.  This PR just restores that logic to MsgReader.

geode-core/src/main/java/org/apache/geode/internal/net/SocketUtils.java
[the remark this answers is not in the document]
bschuchardt: I've added tests for this, so I'm resolving this thread.

geode-core/src/main/java/org/apache/geode/internal/tcp/ClusterConnection.java
bschuchardt: "client" socket is a term from TLS, where one side must be the "client" and the other side is the "server".
The sockets in this class may be one-way or may be bidirectional.  "Shared" connections (those shared between threads) are one-way.  Thread-owned connections are bidirectional.  A P2P "reader" thread reads from the socket and DirectReplyProcessor writes responses.

geode-core/src/test/java/org/apache/geode/internal/net/SocketUtilsJUnitTest.java
bschuchardt: I don't like to factor out code in tests too much.  It leads to tests that are difficult to understand.  Short tests like this should be easy to read and stand on their own.

geode-core/src/main/java/org/apache/geode/internal/net/SocketUtils.java
bschuchardt: The older Old IO code base used a BufferedInputStream wrapper around the socket's input stream on recommendation from StackOverflow and other sources.  I did not find that it increased performance so I removed that wrapper, but I left this parameter in place for flexibility since this is a "Util" class.  This parameter also makes the methods easier to test.

geode-core/src/main/java/org/apache/geode/internal/tcp/MsgReader.java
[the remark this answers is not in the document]
bschuchardt: Let me say a little more about these variables - my understanding of doing this external book-keeping on the buffer is to avoid a compaction when there is still room in the buffer to read more data.  The code in ClusterConnection that reads messages still does that compaction and we should eventually remove it and use MsgReader to read all messages.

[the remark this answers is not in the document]
Bill: thanks for that clarification @bschuchardt. I thought maybe this code was old (and therefore vetted) but it wasn't obvious from the PR.

geode-core/src/test/java/org/apache/geode/internal/net/SocketUtilsJUnitTest.java
Bill: fair.

geode-core/src/main/java/org/apache/geode/internal/net/SocketUtils.java
Bill: thanks

PR #5364, opened 2020-07-09
GEODE-8351: DUnit tests for Delta Propagation

sabbeyPivotal:
Need to confirm that when deltas are propagated, the data is correctly stored on the secondary

### DIFF
--- a/geode-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/DeltaDUnitTest.java
+++ b/geode-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/DeltaDUnitTest.java
@@ -45,6 +45,7 @@ import org.apache.geode.test.dunit.rules.ClusterStartupRule;
 import org.apache.geode.test.dunit.rules.MemberVM;
 import org.apache.geode.test.dunit.rules.RedisClusterStartupRule;
 
+@SuppressWarnings("unchecked")
 public class DeltaDUnitTest {
 
   @ClassRule
--- a/geode-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/DeltaDUnitTest.java
+++ b/geode-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/DeltaDUnitTest.java
@@ -1,0 +1,339 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package org.apache.geode.redis.internal.data;
+
+import static org.apache.geode.distributed.ConfigurationProperties.MAX_WAIT_TIME_RECONNECT;
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Properties;
+import java.util.Set;
+
+import org.junit.AfterClass;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Test;
+import redis.clients.jedis.Jedis;
+
+import org.apache.geode.cache.Region;
+import org.apache.geode.cache.partition.PartitionRegionHelper;
+import org.apache.geode.internal.cache.InternalCache;
+import org.apache.geode.test.awaitility.GeodeAwaitility;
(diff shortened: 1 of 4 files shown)


geode-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/DeltaDUnitTest.java
```diff
@@ -0,0 +1,339 @@
...
+    server2.stop();
+  }
+
+  @Test
+  public void shouldCorrectlyPropagateDeltaToSecondaryServer_whenAppending() {
+    String key = "key";
+    String baseValue = "value-";
+    jedis1.set(key, baseValue);
+    for (int i = 0; i < SET_SIZE; i++) {
+      jedis1.set(key, String.valueOf(i));
```
dschneider-pivotal, line 101: I think you want this to be "append" instead of "set".

geode-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/DeltaDUnitTest.java
```diff
@@ -0,0 +1,339 @@
...
+    server1.stop();
+    server2.stop();
+  }
+
+  @Test
+  public void shouldCorrectlyPropagateDeltaToSecondaryServer_whenAppending() {
+    String key = "key";
+    String baseValue = "value-";
+    jedis1.set(key, baseValue);
+    for (int i = 0; i < SET_SIZE; i++) {
```
dschneider-pivotal, line 100: instead of SET_SIZE in this context should be something like APPEND_COUNT?

geode-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/DeltaDUnitTest.java
```diff
@@ -0,0 +1,339 @@
...
+      jedis1.set(key, String.valueOf(i));
+
+      String server1LocalValue = server1.invoke(() -> {
```
dschneider-pivotal, line 103: extract this chunk of code that into a single method that checks that the primary and secondary are consistent.

geode-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/DeltaDUnitTest.java
```diff
@@ -0,0 +1,339 @@
...
+        InternalCache cache = ClusterStartupRule.getCache();
+        Region<ByteArrayWrapper, RedisData> region = cache.getRegion("__REDIS_DATA");
+        Region<ByteArrayWrapper, RedisData> localRegion =
+            PartitionRegionHelper.getLocalData(region);
+
+        RedisData localValue = localRegion.get(new ByteArrayWrapper(key.getBytes()));
+        return localValue.toString();
```
dschneider-pivotal, line 110: Instead of using the RedisData's toString (which might not show all fields) try converting it to a byte array. You can use BlobHelper.serializeToBlob and then have this return the byte array. Then use the JDK Arrays.equals(byte[], byte[]) to see if they are equal. If they are not you will want to use BlobHelper.deserializeBlob(byte[]) to report the objects that are not equal.

geode-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/DeltaDUnitTest.java
dschneider-pivotal: It seems like you could write it in a way that it is the same for each test. You really don't care in this test what is in each server; just that whatever one server has the other does also.

geode-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/DeltaDUnitTest.java
sabbeyPivotal: Yes, thank you!

geode-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/DeltaDUnitTest.java
sabbeyPivotal: Thank you, meant to change that.

geode-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/DeltaDUnitTest.java
sabbeyPivotal: good call, updating it!

geode-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/DeltaDUnitTest.java
sabbeyPivotal: That is true. I made a generic method for getting the local region and getting the correct data.  I'm not sure how much further we could go without sacrificing readability?

geode-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/DeltaDUnitTest.java
```diff
@@ -0,0 +1,315 @@
...
+          return null;
+        }
+        return localSet.getExpirationTimestamp();
+      }));
+
+      assertThat(server1LocalExpirtionTimestamp).isEqualTo(server2LocalExpirationTimestamp);
+    }
+  }
+
+  private static Object getLocalData(String key,
```
jdeppe-pivotal, line 289: `key` is unused since it is captured by the lambdas and not used anywhere else other than the function being passed in. Don't worry about changing it unless there is something else you happen to also change before we merge in.

PR #5371, opened 2020-07-14
GEODE-8359

mhansonp:
Fix to an NPE in RegisterInterestList66, needed to validate that a we were not getting a null.

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/command/RegisterInterestList66.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/command/RegisterInterestList66.java
@@ -27,6 +27,7 @@ import org.apache.geode.internal.cache.tier.Command;
 import org.apache.geode.internal.cache.tier.InterestType;
 import org.apache.geode.internal.cache.tier.MessageType;
 import org.apache.geode.internal.cache.tier.sockets.BaseCommand;
+import org.apache.geode.internal.cache.tier.sockets.CacheClientProxy;
 import org.apache.geode.internal.cache.tier.sockets.ChunkedMessage;
 import org.apache.geode.internal.cache.tier.sockets.Message;
 import org.apache.geode.internal.cache.tier.sockets.Part;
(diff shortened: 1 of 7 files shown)


geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/command/RegisterInterestList66.java
```diff
@@ -163,29 +162,40 @@ public void cmdExecute(final Message clientMessage, final ServerConnection serve
...
       serverConnection.setAsTrue(RESPONDED);
       return;
     }
 
-    boolean isPrimary = serverConnection.getAcceptor().getCacheClientNotifier()
-        .getClientProxy(serverConnection.getProxyID()).isPrimary();
+    CacheClientProxy ccp = serverConnection.getAcceptor().getCacheClientNotifier()
+        .getClientProxy(serverConnection.getProxyID());
+
+    if (ccp == null) {
```
mhansonp, line 189: The important chunk here copied from RegisterInterest61 which already had a bug fix for this issue.

mhansonp: The core issue was an NPE, because there was no cache client proxy returned during shutdown.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/command/RegisterInterestList66.java
```diff
@@ -163,29 +162,40 @@ public void cmdExecute(final Message clientMessage, final ServerConnection serve
...
+      IOException ioException = new IOException(
+          "CacheClientProxy for this client is no longer on the server , so registerInterest operation is unsuccessful");
```
agingade, line 191: We could change the messaging part...

mhansonp: done

PR #5379, opened 2020-07-16
GEODE-8361: Use Set instead of List to track cleared buckets

DonalEvans:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionClear.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionClear.java
@@ -14,11 +14,9 @@
  */
 package org.apache.geode.internal.cache;
 
-import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashSet;
-import java.util.Iterator;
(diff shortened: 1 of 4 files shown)


geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionClear.java
```diff
@@ -401,7 +395,7 @@ protected void handleClearFromDepartedMember(InternalDistributedMember departedM
     }
   }
 
-  class LockForListenerAndClientNotification {
+  static class LockForListenerAndClientNotification {
```
agingade, line 398: Any reason for changing it to static? The PRClear is separate for each PR.

DonalEvans: This was just a suggestion from the IDE. I'm happy to change it back if it's better to do so.

agingade: It should not be static.

PR #5382, opened 2020-07-17
GEODE-8368: Upgrade ClassGraph from 4.8.52 to 4.8.87

kirklund:
Upgrade ClassGraph dependency from 4.8.52 to 4.8.87.

See GEODE-8150. There were performance issues with 4.8.78 that caused us to rollback to 4.8.52. We'll need to do extra testing as well as review of ClassGraph release notes and bugs.

The intention is to review and fully test any new release of ClassGraph with bug fixes to determine if it's worth upgrading the dependency in Geode. If it introduces new problems or doesn't offer any improvements, we won't upgrade. If it's more stable and fixes bugs in ClassGraph that Geode is vulnerable to, then I will recommend upgrading.

ClassGraph release notes: https://github.com/classgraph/classgraph/releases
ClassGraph closed bugs: https://github.com/classgraph/classgraph/issues?q=is%3Aissue+is%3Aclosed

CAUTION: Please don't merge this PR for me!


onichols-pivotal:
Out of curiosity what are we changing this?  Was there a specific issue with 4.8.52 you are trying to solve?  I would much rather go back to 4.0.6 and fix the log4j issue with an exclusion, than mess around with any of these newfangled versions.

kirklund:
@onichols-pivotal ClassGraph has nothing to do with Log4j. I analyzed a test run in which a thread hung in ClassGraph#scan. ClassGraph has had many bugs identified and fixed between 4.8.52 and 4.8.87. I believe that a dependency like this should be kept up to date especially when important bugs have been fixed.

kirklund:
I don't have time for this PR.

PR #5385, opened 2020-07-20
GEODE-8370: Add test for maxInactiveInterval

sabbeyPivotal:
Write test verifying that maxInactiveInterval propagates correctly to all buckets

### DIFF
--- a/geode-redis/src/acceptanceTest/java/session/NativeRedisSessionExpirationAcceptanceTest.java
+++ b/geode-redis/src/acceptanceTest/java/session/NativeRedisSessionExpirationAcceptanceTest.java
@@ -16,7 +16,6 @@ package session;
 
 import org.junit.BeforeClass;
 import org.junit.ClassRule;
-import org.junit.Ignore;
 import org.junit.Test;
 import org.junit.rules.TestRule;
 import org.testcontainers.containers.GenericContainer;
@@ -46,13 +45,11 @@ public class NativeRedisSessionExpirationAcceptanceTest extends SessionExpiratio
   }
 
   @Test
-  @Ignore
   public void sessionShouldTimeout_whenAppFailsOverToAnotherRedisServer() {
     // Only using one server for Native Redis
   }
 
   @Test
-  @Ignore
   public void sessionShouldNotTimeout_whenPersisted() {
     // Only using one server for Native Redis, no need to compare redundant copies
   }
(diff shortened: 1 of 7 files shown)


sabbeyPivotal:
Failure in Distributed Test unrelated to this PR

geode-redis/src/acceptanceTest/java/session/NativeRedisSessionExpirationAcceptanceTest.java
```diff
@@ -50,4 +50,10 @@ public static void setup() {
   public void sessionShouldTimeout_whenAppFailsOverToAnotherRedisServer() {
     // Only using one server for Native Redis
   }
+
+  @Test
+  @Ignore
```
dschneider-pivotal, line 55: Normally when I see an ignore on a test I want to know why the test is being ignored. I think it is something we could fix and then have this test. In this case I think we are overriding a test that we will never do with native redis. It seems like the empty method impl takes care of that and you would not need the ignore. This is just my opinion, I'd be okay if you leave the ignore if you think that is best.

geode-redis/src/distributedTest/java/org/apache/geode/redis/session/SessionExpirationDUnitTest.java
```diff
@@ -86,15 +100,78 @@ public void sessionShouldTimeout_whenAppFailsOverToAnotherRedisServer() {
...
+        .postForEntity(
+            "http://localhost:" + ports.get(sessionApp) + "/setMaxInactiveInterval",
+            request,
+            Integer.class)
+        .getHeaders();
+  }
+
+  private void compareMaxInactiveIntervals() {
+    cluster.getVM(1).invoke(() -> {
+      for (int j = 0; j < 113; j++) {
```
dschneider-pivotal, line 137: instead of the hard coded "113", ask the PartitionedRegion for its max bucket. See PartitionedRegion.getTotalNumberOfBuckets. This will help in the future when we allow the redis bucket count to be configured

geode-redis/src/acceptanceTest/java/session/NativeRedisSessionExpirationAcceptanceTest.java
sabbeyPivotal: Just updated it and took out the ignores!

geode-redis/src/distributedTest/java/org/apache/geode/redis/session/SessionExpirationDUnitTest.java
sabbeyPivotal: Good suggestion! Just updated it!

PR #5391, opened 2020-07-21
GEODE-7846: Adding Stats for Partitioned Region Clear

BenjaminPerryRoss:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearDUnitTest.java
@@ -152,6 +152,24 @@ public class PartitionedRegionClearDUnitTest implements Serializable {
     dataStore3.invoke(() -> verifyRegionSize(false, expectedNum));
   }
 
+  private void verifyDatastoreStats(MemberVM datastore) {
+    datastore.invoke(() -> {
+      PartitionedRegion region = (PartitionedRegion) getRegion(false);
+      long clearCount = 0L;
+
+      for (BucketRegion bucket : region.getDataStore().getAllLocalBucketRegions()) {
+        if (clearCount == 0) {
+          clearCount = bucket.getCachePerfStats().getBucketClearCount();
+        }
+        assertThat(bucket.getCachePerfStats().getBucketClearCount()).isEqualTo(clearCount)
+            .isNotEqualTo(0);
+      }
+
+      assertThat(region.getCachePerfStats().getRegionClearCount()).isEqualTo(1);
+      assertThat(region.getCachePerfStats().getPartitionedRegionClearDuration()).isGreaterThan(0);
+    });
+  }
+
   private void verifyClientRegionSize(int expectedNum) {
     client1.invoke(() -> verifyRegionSize(true, expectedNum));
     // TODO: notify register clients
@@ -168,13 +186,6 @@ public class PartitionedRegionClearDUnitTest implements Serializable {
     int destroys =
         destroysByRegion.get(REGION_NAME) == null ? 0 : destroysByRegion.get(REGION_NAME).get();
     return destroys;
-  };
-
-  SerializableCallableIF<Integer> getStatClearCount = () -> {
-    Region region = getRegion(false);
-    long clearCount = ((LocalRegion) region).getCachePerfStats().getRegionClearCount();
-
-    return (int) clearCount;
   };
 
   SerializableCallableIF<Integer> getBucketRegionWriterClears = () -> {
@@ -359,51 +370,9 @@ public class PartitionedRegionClearDUnitTest implements Serializable {
     verifyServerRegionSize(0);
 
     // Verify the stats were properly updated for the bucket regions
-    dataStore1.invoke(() -> {
-      PartitionedRegion region = (PartitionedRegion) getRegion(false);
-      long clearCount = 0L;
-
-      for (BucketRegion bucket : region.getDataStore().getAllLocalBucketRegions()) {
-        if (clearCount == 0) {
-          clearCount = bucket.getCachePerfStats().getBucketClearCount();
-        }
-        assertThat(bucket.getCachePerfStats().getBucketClearCount()).isEqualTo(clearCount)
-            .isNotEqualTo(0);
-      }
-
-      assertThat(region.getCachePerfStats().getRegionClearCount()).isEqualTo(1);
-      assertThat(region.getCachePerfStats().getPartitionedRegionClearDuration()).isGreaterThan(0);
-    });
-
-    dataStore2.invoke(() -> {
-      PartitionedRegion region = (PartitionedRegion) getRegion(false);
-      long clearCount = 0L;
-
-      for (BucketRegion bucket : region.getDataStore().getAllLocalBucketRegions()) {
-        if (clearCount == 0) {
-          clearCount = bucket.getCachePerfStats().getBucketClearCount();
-        }
-        assertThat(bucket.getCachePerfStats().getBucketClearCount()).isEqualTo(clearCount)
-            .isNotEqualTo(0);
-      }
-
-      assertThat(region.getCachePerfStats().getRegionClearCount()).isEqualTo(1);
-    });
-
-    dataStore3.invoke(() -> {
-      PartitionedRegion region = (PartitionedRegion) getRegion(false);
-      long clearCount = 0L;
-
-      for (BucketRegion bucket : region.getDataStore().getAllLocalBucketRegions()) {
-        if (clearCount == 0) {
-          clearCount = bucket.getCachePerfStats().getBucketClearCount();
-        }
-        assertThat(bucket.getCachePerfStats().getBucketClearCount()).isEqualTo(clearCount)
-            .isNotEqualTo(0);
-      }
-
-      assertThat(region.getCachePerfStats().getRegionClearCount()).isEqualTo(1);
-    });
+    verifyDatastoreStats(dataStore1);
+    verifyDatastoreStats(dataStore2);
+    verifyDatastoreStats(dataStore3);
 
 
     // The accessor shouldn't increment the region clear count
--- a/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearDUnitTest.java
@@ -20,13 +20,11 @@ import static org.apache.geode.test.dunit.rules.ClusterStartupRule.getClientCach
 import static org.assertj.core.api.Assertions.assertThat;
 
 import java.io.Serializable;
-import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.Properties;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.IntStream;
 
-import com.sun.tools.javac.util.List;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.junit.Before;
@@ -174,7 +172,7 @@ public class PartitionedRegionClearDUnitTest implements Serializable {
 
   SerializableCallableIF<Integer> getStatClearCount = () -> {
     Region region = getRegion(false);
-    long clearCount = ((LocalRegion)region).getCachePerfStats().getClearCount();
+    long clearCount = ((LocalRegion)region).getCachePerfStats().getRegionClearCount();
 
     return(int)clearCount;
   };
@@ -350,7 +348,7 @@ public class PartitionedRegionClearDUnitTest implements Serializable {
       PartitionedRegion region = (PartitionedRegion) getRegion(false);
 
       for(BucketRegion bucket : region.getDataStore().getAllLocalBucketRegions()) {
-        long clearCount = bucket.getCachePerfStats().getClearCount();
+        long clearCount = bucket.getCachePerfStats().getRegionClearCount();
         assertThat(clearCount).isEqualTo(0);
       }
     });
@@ -363,12 +361,53 @@ public class PartitionedRegionClearDUnitTest implements Serializable {
     //Verify the stats were properly updated for the bucket regions
     dataStore1.invoke(() -> {
       PartitionedRegion region = (PartitionedRegion) getRegion(false);
+      long clearCount = 0L;
 
       for(BucketRegion bucket : region.getDataStore().getAllLocalBucketRegions()) {
-        long clearCount = bucket.getCachePerfStats().getClearCount();
-        assertThat(clearCount).isEqualTo(1);
+        if(clearCount == 0) {
+          clearCount = bucket.getCachePerfStats().getBucketClearCount();
+        }
+        assertThat(bucket.getCachePerfStats().getBucketClearCount()).isEqualTo(clearCount).isNotEqualTo(0);
       }
+
+      assertThat(region.getCachePerfStats().getRegionClearCount()).isEqualTo(1);
+      assertThat(region.getCachePerfStats().getPartitionedRegionClearDuration()).isGreaterThan(0);
     });
+
+    dataStore2.invoke(() -> {
+      PartitionedRegion region = (PartitionedRegion) getRegion(false);
+      long clearCount = 0L;
+
+      for(BucketRegion bucket : region.getDataStore().getAllLocalBucketRegions()) {
+        if(clearCount == 0) {
+          clearCount = bucket.getCachePerfStats().getBucketClearCount();
+        }
+        assertThat(bucket.getCachePerfStats().getBucketClearCount()).isEqualTo(clearCount).isNotEqualTo(0);
+      }
+
+      assertThat(region.getCachePerfStats().getRegionClearCount()).isEqualTo(1);
+    });
+
+    dataStore3.invoke(() -> {
+      PartitionedRegion region = (PartitionedRegion) getRegion(false);
+      long clearCount = 0L;
+
+      for(BucketRegion bucket : region.getDataStore().getAllLocalBucketRegions()) {
+        if(clearCount == 0) {
+          clearCount = bucket.getCachePerfStats().getBucketClearCount();
+        }
+        assertThat(bucket.getCachePerfStats().getBucketClearCount()).isEqualTo(clearCount).isNotEqualTo(0);
+      }
+
+      assertThat(region.getCachePerfStats().getRegionClearCount()).isEqualTo(1);
+    });
+
+
+ //   The accessor shouldn't increment the region clear count
+    accessor.invoke(() -> {
+      PartitionedRegion region = (PartitionedRegion) getRegion(false);
+      assertThat(region.getCachePerfStats().getRegionClearCount()).isEqualTo(0);
+        });
 
     // do the region destroy to compare that the same callbacks will be triggered
     dataStore1.invoke(() -> {
--- a/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearDUnitTest.java
@@ -156,13 +156,13 @@ public class PartitionedRegionClearDUnitTest implements Serializable {
     datastore.invoke(() -> {
       PartitionedRegion region = (PartitionedRegion) getRegion(false);
       long clearCount = 0L;
+      int bucketCount = region.getDataStore().getAllLocalBucketRegions().size();
 
       for (BucketRegion bucket : region.getDataStore().getAllLocalBucketRegions()) {
         if (clearCount == 0) {
           clearCount = bucket.getCachePerfStats().getBucketClearCount();
         }
-        assertThat(bucket.getCachePerfStats().getBucketClearCount()).isEqualTo(clearCount)
-            .isNotEqualTo(0);
+        assertThat(bucket.getCachePerfStats().getBucketClearCount()).isEqualTo(bucketCount);
       }
 
       assertThat(region.getCachePerfStats().getRegionClearCount()).isEqualTo(1);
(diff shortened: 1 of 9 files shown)


geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionClear.java
```diff
@@ -162,6 +162,7 @@ protected void waitForPrimary(PartitionedRegion.RetryTimeKeeper retryTimer) {
         doAfterClear(regionEvent);
       }
     }
+    incClearCount();
```
DonalEvans, line 165: I feel like the calls to `incClearCount()` and `incClearDuration()` should be made at the same time, since there should be a consistent approach to when a clear is "done."

geode-core/src/main/java/org/apache/geode/internal/cache/CachePerfStats.java
```diff
@@ -442,7 +451,10 @@
             f.createIntCounter("retries",
                 "Number of times a concurrent destroy followed by a create has caused an entry operation to need to retry.",
                 "operations"),
-            f.createLongCounter("clears", clearsDesc, "operations"),
+            f.createLongCounter("regionClears", regionClearsDesc, "operations"),
```
DonalEvans, line 454: If the name of this statistic is changed, then `geode-docs/reference/statistics_list.html.md.erb` should be updated to reflect the new name. Also, it might be good to document the newly added statistics too.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionClear.java
```diff
@@ -367,12 +370,38 @@ void doClear(RegionEventImpl regionEvent, boolean cacheWrite) {
...
+    }
+  }
+
+  void incClearCount() {
+    if (partitionedRegion != null && partitionedRegion.getDataStore() != null
+        && partitionedRegion.getDataStore().getAllLocalBucketRegions() != null
+        && partitionedRegion.getDataStore().getAllLocalBucketRegions().size() != 0) {
+      CachePerfStats stats = partitionedRegion.getCachePerfStats();
+      if (stats != null) {
+        logger.info("BR inc PR Region count:" + stats.getClass().getName() + ":"
```
DonalEvans, line 385: Remove this logging.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionClear.java
```diff
@@ -367,12 +370,38 @@ void doClear(RegionEventImpl regionEvent, boolean cacheWrite) {
...
+        stats.incClearCount();
+      }
     }
   }
 
+  void incClearDuration(long durationNanos) {
+    if (partitionedRegion != null && partitionedRegion.getTotalNumberOfBuckets() != 0) {
+      CachePerfStats stats = partitionedRegion.getCachePerfStats();
+      if (stats != null) {
+        logger.info("BR inc PR Duration by + " + durationNanos + " ns:" + stats.getClass().getName()
```
DonalEvans, line 396: Remove this logging.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionClear.java
```diff
@@ -367,12 +370,38 @@ void doClear(RegionEventImpl regionEvent, boolean cacheWrite) {
...
     } finally {
       releaseDistributedClearLock(lockName);
+      incClearDuration(System.nanoTime() - clearStartTime);
+    }
+  }
+
+  void incClearCount() {
+    if (partitionedRegion != null && partitionedRegion.getDataStore() != null
+        && partitionedRegion.getDataStore().getAllLocalBucketRegions() != null
+        && partitionedRegion.getDataStore().getAllLocalBucketRegions().size() != 0) {
```
DonalEvans, line 382: I'm not sure that all these null checks are required. I would think that if the region is null at this point, then we would expect that something has gone wrong and that we should throw an exception rather than just silently not incrementing the stats. Also, I'm not sure I understand why we only increment the stats if `getAllLocalBucketRegions()` returns a non-empty set. Are there scenarios where we could call clear, it could complete without an exception, bringing us to this method, and then `getAllLocalBucketRegions()` returns empty?

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionClear.java
```diff
@@ -367,12 +370,38 @@ void doClear(RegionEventImpl regionEvent, boolean cacheWrite) {
...
+      if (stats != null) {
+        logger.info("BR inc PR Region count:" + stats.getClass().getName() + ":"
+            + partitionedRegion.getFullPath(), new Exception());
+        stats.incClearCount();
+      }
     }
   }
 
+  void incClearDuration(long durationNanos) {
+    if (partitionedRegion != null && partitionedRegion.getTotalNumberOfBuckets() != 0) {
```
DonalEvans, line 393: See my other comment about the possibility of hitting the scenario being checked for. Is it actually possible that we get to where this method is called and `getTotalNumberOfBuckets()` is zero? If it is, do we want to throw an exception instead of just quietly not incrementing the stats?

geode-core/src/main/java/org/apache/geode/internal/cache/RegionPerfStats.java
```diff
@@ -519,9 +519,15 @@ public void incEvictWorkTime(long delta) {
...
+  @Override
+  public void incRegionClearCount() {
+    stats.incLong(regionClearsId, 1L);
+    cachePerfStats.incClearCount();
+  }
+
   @Override
   public void incClearCount() {
-    stats.incLong(clearsId, 1L);
+    stats.incLong(bucketClearsId, 1L);
```
DonalEvans, line 530: I think this may possibly be incorrect. `incRegionClearCount()` is being overridden from `RegionStats` but every other method in this class is overriding methods from `CachePerfStats`. I think the correct methods might be `incClearCount()` which calls `cachePerfStats.incClearCount()` and `incBucketClearCount()` which calls `cachePerfStats.incBucketClearCount()`.

geode-core/src/main/java/org/apache/geode/internal/cache/RegionStats.java
```diff
@@ -135,7 +135,9 @@
 
   void incEvictWorkTime(long delta);
 
-  void incClearCount();
+  void incBucketClearCount();
+
+  void incRegionClearCount();
```
DonalEvans, line 140: This method should have remained as `incClearCount()`, I think, to remain consistent with `CachePerfStats`.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearDUnitTest.java
```diff
@@ -170,6 +170,13 @@ private void verifyClientRegionSize(int expectedNum) {
     return destroys;
   };
 
+  SerializableCallableIF<Integer> getStatClearCount = () -> {
```
DonalEvans, line 173: This method is never used.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearDUnitTest.java
```diff
@@ -330,6 +337,101 @@ public void normalClearFromAccessorWithoutWriterButWithWriterOnDataStore() {
...
+
+    accessor.invoke(() -> feed(false));
+    verifyServerRegionSize(NUM_ENTRIES);
+    dataStore1.invoke(() -> getRegion(false).clear());
+    verifyServerRegionSize(0);
+
+    // Verify the stats were properly updated for the bucket regions
+    dataStore1.invoke(() -> {
+      PartitionedRegion region = (PartitionedRegion) getRegion(false);
+      long clearCount = 0L;
```
DonalEvans, line 364: Rather than just comparing the clear counts of the buckets between eachother, it might be better to compare all of them to the expected clear count. The current check just asserts that the clear counts for all the buckets are equal, but not that that value is necessarily correct.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearDUnitTest.java
```diff
@@ -330,6 +337,101 @@ public void normalClearFromAccessorWithoutWriterButWithWriterOnDataStore() {
...
+
+      for (BucketRegion bucket : region.getDataStore().getAllLocalBucketRegions()) {
+        if (clearCount == 0) {
+          clearCount = bucket.getCachePerfStats().getBucketClearCount();
+        }
+        assertThat(bucket.getCachePerfStats().getBucketClearCount()).isEqualTo(clearCount)
+            .isNotEqualTo(0);
+      }
+
+      assertThat(region.getCachePerfStats().getRegionClearCount()).isEqualTo(1);
```
DonalEvans, line 390: This invocation is executed on all three datastores. It might simplify the test to extract it to a method.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearDUnitTest.java
```diff
@@ -330,6 +337,101 @@ public void normalClearFromAccessorWithoutWriterButWithWriterOnDataStore() {
...
+    assertThat(dataStore2.invoke(getWriterDestroys)).isEqualTo(dataStore2.invoke(getWriterClears))
+        .isEqualTo(0);
+    assertThat(dataStore3.invoke(getWriterDestroys)).isEqualTo(dataStore3.invoke(getWriterClears))
+        .isEqualTo(0);
+    assertThat(accessor.invoke(getWriterDestroys)).isEqualTo(accessor.invoke(getWriterClears))
+        .isEqualTo(1);
+
+    assertThat(accessor.invoke(getBucketRegionWriterDestroys))
+        .isEqualTo(accessor.invoke(getBucketRegionWriterClears))
+        .isEqualTo(0);
```
DonalEvans, line 432: I'm not sure what this section has to do with stats. I think it can safely be removed from this test case.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionClear.java
BenjaminPerryRoss: This is part of a larger discussion about what we're actually trying to record with the duration. It's something Gester and I have touched on in the operations standup and I'll probably put something up regarding it in the slack channel but basically the duration in the initial version (the one you commented on) was measuring the total duration of the entire clear operation from the perspective of the coordinator. The way it's been changed to (in line with your suggestion) is to move it into the actual local clear. This makes the stat present on all members which have the region (as long as they have data) but it means that this value ONLY captures the actual process of clearing the region and not any of the stuff that happens in between (messaging, any work done on the coordinator other than the local clear, etc). For now it's been changed to something that matches with your thought process here, but if you have ideas on this I would love to hear them.

geode-core/src/main/java/org/apache/geode/internal/cache/CachePerfStats.java
BenjaminPerryRoss: Agreed. I recall there was a documentation story for PR clear, but I'm not sure what happened to it. I'll hunt it down and if this documentation isn't covered by it I'll make sure to add it to this PR.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionClear.java
BenjaminPerryRoss: Done

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionClear.java
BenjaminPerryRoss: Done.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionClear.java
BenjaminPerryRoss: Checking that the partitioned region is non null might be overkill, but the others are needed to determine whether or not this member has data. If there's no data on the member then no clear is actually taking place and we don't want to log a count. I added this many checks because I saw inconsistent behavior in my tests where sometimes the PR would have a list which was empty, sometimes it would have a list which was null for my Accessor. I don't think it's necessary to throw an exception in this case.

geode-core/src/main/java/org/apache/geode/internal/cache/RegionPerfStats.java
BenjaminPerryRoss: It took me awhile to understand what you meant here - but you were correct. There was some weird inheretance going on here stemming from the fact that I didn't understand the relationship between RegionStats, RegionPerfStats, and CachePerfStats. These should be fixed now.

geode-core/src/main/java/org/apache/geode/internal/cache/RegionStats.java
BenjaminPerryRoss: RegionPerfStats and other classes which used "incClearCount()" have been replaced with "incRegionClearCount()" as that is more descriptive and you're right - they should be consistent. This was being blocked previously by the error you identified in your previous comment.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearDUnitTest.java
BenjaminPerryRoss: Removed.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearDUnitTest.java
BenjaminPerryRoss: The trick here is that we don't know exactly how many buckets will be in each datastore for sure. In the current iteration of the test, 2 members will have 7 and one member will have 6 buckets (although I'd imagine you could end up with slightly different balance in specific scenarios). Originally I changed the test to have a number of total buckets across all members that was evenly divisible by the number of datastores - but this felt like a hardcoded trap for anyone trying to modify the test in the future. In the end I decided that it was good enough to determine that the count was incrementing to a consistent non-zero value even if I couldn't make sure that value matched up to a hardcoded expectation - but if you feel this is insufficient I can take another shot at this check.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearDUnitTest.java
BenjaminPerryRoss: Good call - done.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearDUnitTest.java
BenjaminPerryRoss: These assertions mostly just check that the cache writer behaved as expected. While they're not specifically necessary, they're present in every other test in this suite so I opted to leave them in.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionClear.java
BenjaminPerryRoss: Same reason as above, but I wanted to add that this was simpler because in the original PR this call was only happening on the coordinator. Now this can happen on any member so I've changed it to check everything that was being checked for the clear count.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearDUnitTest.java
DonalEvans: If it's safe to assume that buckets aren't moving around during the clear (which in this test seems to be the case) then it should be possible to get the number of buckets on each member with `region.getDataStore().getAllLocalBucketRegions()` and use that for verification.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionClear.java
DonalEvans: Might it be possible to instead move the call to this method (and `incClearDuration()`, if it ends up staying where it currently is) inside the `if (partitionedRegion.getDataStore() != null)` block in `clearRegionLocal()`, just after the `doAfterClear()`, and guarded with a check on the size of `clearedBuckets`? i.e. if no buckets are cleared, do not increment the stats? That should avoid having to do quite so much checking in the method itself and instead place the focus on calling it in the appropriate place, where it will always increment the stats.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearDUnitTest.java
DonalEvans: I think that the rest of the tests in the class are specifically testing clear's interaction with writers (at least judging by test names) so it's appropriate for these checks to be present in them, but in a test specifically about stats, they risk causing confusion.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionClear.java
BenjaminPerryRoss: It could, but that pattern would have to be followed if this clear count was ever incremented from anywhere else in the code base, as it would be reliant upon being called within a conditional. Is that preferable? Personally, I feel like having the conditions - while a little messy - communicates the requirements of the method pretty clearly and while I wouldn't necessarily promote that approach for a more complex method which is already very busy with functionality, this method is simple enough to read even with the conditions. Aside from that I think what you're suggesting is certainly possible.

DonalEvans: If all the checks were removed from the method, then it would turn into just 'partitionedRegion.getCachePerfStats().incRegionClearCount()` which could be called inline rather than having a separate method, especially considering that it's only called once. Looking at how other stats are incremented, this pattern of directly calling the method on the stats class inline when it's appropriate to do so seems to be used almost everywhere, and using it here would simplify the class a fair bit.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearDUnitTest.java
BenjaminPerryRoss: Done.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearDUnitTest.java
BenjaminPerryRoss: Done.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionClear.java
```diff
@@ -123,6 +123,7 @@ protected void waitForPrimary(PartitionedRegion.RetryTimeKeeper retryTimer) {
 
   public Set<Integer> clearRegionLocal(RegionEventImpl regionEvent) {
     Set<Integer> clearedBuckets = new HashSet<>();
+    Long clearStartTime = System.nanoTime();
```
DonalEvans, line 126: Changing this from `Long` to `long` will fix the LGTM alert introduced by this PR.

geode-core/src/main/java/org/apache/geode/internal/cache/AbstractRegionMap.java
```diff
@@ -344,7 +344,12 @@ void incClearCount(LocalRegion lr) {
     if (lr != null && !(lr instanceof HARegion)) {
       CachePerfStats stats = lr.getCachePerfStats();
       if (stats != null) {
-        stats.incClearCount();
+        if (lr instanceof BucketRegion) {
```
gesterzhou, line 347: Usually we use:  if (owner.isUsedForPartitionedRegionBucket()) { instead.

geode-core/src/main/java/org/apache/geode/internal/cache/CachePerfStats.java
```diff
@@ -274,7 +278,14 @@
         "Current number of regions configured for reliablity that are missing required roles with Limited access";
     final String reliableRegionsMissingNoAccessDesc =
         "Current number of regions configured for reliablity that are missing required roles with No access";
-    final String clearsDesc = "The total number of times a clear has been done on this cache.";
+    final String regionClearsDesc =
+        "The total number of times a clear has been done on this cache.";
```
gesterzhou, line 282: on this cache? or on this region?

geode-core/src/main/java/org/apache/geode/internal/cache/CachePerfStats.java
```diff
@@ -274,7 +278,14 @@
...
+    final String bucketClearsDesc =
+        "The total number of times a clear has been done on this region and it's bucket regions";
+    final String partitionedRegionClearLocalDurationDesc =
+        "The time in nanoseconds partitioned region clear has been running for the region on this member";
```
gesterzhou, line 286: we have to follow other duration stats's timeunit. If other duration stats used nanoseconds, we use it too. Otherwise, we should not.

BenjaminPerryRoss: I actually switch this from the original metric I used to nanoseconds because other stats in the file seem to be using them and I wanted to be consistent (such as EventQueueThrottleTime or any of the tx lifetime stats in that class).

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionClear.java
```diff
@@ -378,8 +378,11 @@ void doClear(RegionEventImpl regionEvent, boolean cacheWrite) {
     } finally {
       releaseDistributedClearLock(lockName);
       if (partitionedRegion.getCachePerfStats() != null) {
```
DonalEvans, line 380: I think that the null check on `partitionedRegion.getCachePerfStats()` is no longer necessary with this change, since it's not them we're updating.

PR #5400, opened 2020-07-24
GEODE-7864: Fix several LGTM warnings

DonalEvans:
- This commit addresses language-features, logic, useless-code and
readability warnings
- The change to CompactRangeIndex may be fixing a bug/incorrectness in
the code

### For all changes:

### Note:


geode-core/src/main/java/org/apache/geode/cache/query/internal/index/CompactRangeIndex.java
```diff
@@ -939,7 +939,7 @@ protected boolean evaluateEntry(IndexInfo indexInfo, ExecutionContext context, O
       // either of them is null and operator is other than == or !=
       if (result == QueryService.UNDEFINED) {
         // Undefined is added to results for != conditions only
-        if (operator != OQLLexerTokenTypes.TOK_NE || operator != OQLLexerTokenTypes.TOK_NE_ALT) {
```
nabarunnag, line 942: This seems like a behavior change, sorry if i am missing something.

nabarunnag: judging by this comment. 
`// Undefined is added to results for != conditions only`

DonalEvans: It is a change in behaviour, but I think it's a change that introduces the originally intended behaviour. The previous logic always evaluated to true, which made me think that there might have been a mistake in it. The comment about "for != conditions only" I think is referring to the `TOK_NE` and `TOK_NE_ALT` types (not equal and not equal alt), not that the operator should be not equal to them.

Previously, any time the result was `UNDEFINED` it would be added to the results, but with the change, it's only added to the results if the operator is `TOK_NE` or `TOK_NE_ALT`, which sounds from the comment to be what was originally intended. All of the tests are passing here with this change, and I ran some other tests privately that also hit no issues, so I'm fairly confident that this isn't breaking anything.

PR #5403, opened 2020-07-27
GEODE-8385: hang recovering from disk with cyclic dependencies

bschuchardt:
This restores the point at which we notify membership listeners of
departures.  We used to do this (in 1.12 and earlier) when a ShutdownMessage
is received instead of waiting for a new membership view announcing the departure.
Membership views can take some time to form and install, which can cause
persistent (disk store) views to be updated later than they used to be.

In the case of this ticket the disk store of one member was being
closed while another was shutting down.  The member closing its disk
store did not see the view announcing that shutdown until most of its
disk store regions had closed their persistence advisors.  This left the
disk store thinking that the other member was still up at the time it
was closed.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/distributed/internal/ClusterDistributionManager.java
+++ b/geode-core/src/main/java/org/apache/geode/distributed/internal/ClusterDistributionManager.java
@@ -1784,19 +1784,15 @@ public class ClusterDistributionManager implements DistributionManager {
   /**
    * Returns true if id was removed. Returns false if it was not in the list of managers.
    */
-  private boolean removeManager(InternalDistributedMember theId, boolean crashed, String p_reason) {
-    String reason = p_reason;
-
-    reason = prettifyReason(reason);
+  private void removeManager(InternalDistributedMember theId, boolean crashed, String p_reason) {
     if (logger.isDebugEnabled()) {
+      String reason = prettifyReason(p_reason);
       logger.debug("DistributionManager: removing member <{}>; crashed {}; reason = {}", theId,
           crashed, reason);
     }
     removeHostedLocators(theId);
 
     redundancyZones.remove(theId);
-
-    return true;
   }
 
   /**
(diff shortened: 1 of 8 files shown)


geode-core/src/main/java/org/apache/geode/distributed/internal/ClusterDistributionManager.java
```diff
@@ -2366,10 +2375,10 @@ public void memberDeparted(InternalDistributedMember theId, boolean crashed, Str
           message.setReason(reason); // added for #37950
           handleIncomingDMsg(message);
         }
-        dm.handleManagerDeparture(theId, crashed, reason);
       } catch (DistributedSystemDisconnectedException se) {
         // let's not get huffy about it
       }
+      dm.handleManagerDeparture(theId, crashed, reason, true);
```
Bill, line 2381: why did this statement move? should it be in a `finally` block?

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSMemberData.java
```diff
@@ -399,11 +399,11 @@ public int hashCode() {
   public String toString() {
     StringBuilder sb = new StringBuilder(100);
 
-    sb.append("MemberData[");
+    sb.append("GMSMember[");
```
Bill, line 402: Why did the name change to `"GMSMember"` here? It was `"MemberData"` which kinda made sense because `GMSMemberData` isa `MemberData`. I could see leaving it alone or maybe making it `GMSMemberData`.

bschuchardt: I think that's an artifact of my doing this work on an old SHA and then rebasing onto develop.  I'll revert that change

geode-core/src/main/java/org/apache/geode/distributed/internal/ClusterDistributionManager.java
bschuchardt: that change was unnecessary - I'll revert it.

geode-core/src/main/java/org/apache/geode/distributed/internal/ClusterDistributionManager.java
```diff
@@ -1784,19 +1784,15 @@ private String prettifyReason(String r) {
   /**
    * Returns true if id was removed. Returns false if it was not in the list of managers.
    */
-  private boolean removeManager(InternalDistributedMember theId, boolean crashed, String p_reason) {
-    String reason = p_reason;
-
-    reason = prettifyReason(reason);
+  private void removeManager(InternalDistributedMember theId, boolean crashed, String p_reason) {
```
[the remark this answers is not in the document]
bschuchardt: I've in-lined this method.  Its name no longer made sense and it was only used by handleManagerDeparture()

geode-core/src/main/java/org/apache/geode/distributed/internal/ClusterDistributionManager.java
```diff
@@ -1866,41 +1848,45 @@ public void handleConsoleShutdown(InternalDistributedMember theId, boolean crash
...
 
   @Override
-  public void handleManagerDeparture(InternalDistributedMember theId, boolean p_crashed,
-      String p_reason) {
-
+  public void handleManagerDeparture(InternalDistributedMember theId, boolean memberCrashed,
+      String reason, boolean fromViewChange) {
     alertingService.removeAlertListener(theId);
 
+    removeUnfinishedStartup(theId, true);
```
Bill, line 1859: nice

geode-core/src/main/java/org/apache/geode/distributed/internal/ClusterDistributionManager.java
```diff
@@ -1866,41 +1848,45 @@ public void handleConsoleShutdown(InternalDistributedMember theId, boolean crash
...
-            "Member at {} gracefully left the distributed cache: {}";
-        addMemberEvent(new MemberDepartedEvent(theId, p_reason));
-      }
-      logger.info(msg, new Object[] {theId, prettifyReason(p_reason)});
+    if (logger.isDebugEnabled()) {
+      logger.debug("DistributionManager: removing member <{}>; crashed {}; reason = {}", theId,
+          memberCrashed, prettifyReason(reason));
+    }
+    removeHostedLocators(theId);
+    redundancyZones.remove(theId);
```
Bill, line 1872: ✓ inlined `removeManager()`

geode-core/src/main/java/org/apache/geode/distributed/internal/ClusterDistributionManager.java
```diff
@@ -1866,39 +1848,56 @@ public void handleConsoleShutdown(InternalDistributedMember theId, boolean crash
   void shutdownMessageReceived(InternalDistributedMember theId, String reason) {
     removeHostedLocators(theId);
     distribution.shutdownMessageReceived(theId, reason);
+    handleManagerDeparture(theId, false, reason);
```
jchen21, line 1851: Why this is not `handleManagerDeparture(theId, false, reason, true)`? `handleManagerDeparture(theId, false, reason)` will not update the `stats` in line 1887.

geode-core/src/main/java/org/apache/geode/distributed/internal/ClusterDistributionManager.java
```diff
@@ -1866,39 +1848,56 @@ public void handleConsoleShutdown(InternalDistributedMember theId, boolean crash
...
+      msg =
+          "Member at {} unexpectedly left the distributed cache: {}";
+      addMemberEvent(new MemberCrashedEvent(theId, reason));
+    } else {
+      msg =
+          "Member at {} gracefully left the distributed cache: {}";
+      addMemberEvent(new MemberDepartedEvent(theId, reason));
+    }
+    if (fromViewChange) {
+      logger.info(msg, new Object[] {theId, prettifyReason(reason)});
```
jchen21, line 1900: Would it better to move the `logger.info` out of the `if` statement? So regardless the value of `fromViewChange`, it logs the member departure. It will be helpful when analyzing the logs.

bschuchardt: If I move it out of the `if` statement it will log the departure on every invocation of this method, and I don't want to do that.  I think it will be confusing.

geode-core/src/main/java/org/apache/geode/distributed/internal/ClusterDistributionManager.java
bschuchardt: Sorry, I don't understand the question.  We want to decrement the node count once, but this method will be invoked multiple times for the same departing node.  If we restrict the decrement to view-changes it will only be decremented once because we only notify memberDeparted once for a member that's left the view.

geode-core/src/main/java/org/apache/geode/distributed/internal/ClusterDistributionManager.java
bschuchardt: I've simplified the method now, so it no longer has fromViewChange.  We'll be notifying listeners of departures when we get a shutdown message.  The subsequent view change won't trigger listener invocation.

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSMembership.java
```diff
@@ -688,7 +688,11 @@ private void removeWithViewLock(ID dm, boolean crashed, String reason) {
       return; // Explicit deletion, no upcall.
     }
 
-    listener.memberDeparted(dm, crashed, reason);
+    if (!shutdownMembers.containsKey(dm)) {
```
jchen21, line 691: 👍

PR #5407, opened 2020-07-28
Revert "GEODE-8373: For CI test runs, split the assemble and test invocation, to set max-…"

rhoughton-pivot:
Reverts apache/geode#5389


rhoughton-pivot:
The concatenation of args to the `bash -c` is not evaluated as expected, so the actual test portion does not run. Revert and look for a better answer to the JVM args for testing.

PR #5411, opened 2020-07-29
GEODE-8372: Configure CMS to send UTF-8 regardless of JVM default

jinmeiliao:


geode-management/src/main/java/org/apache/geode/management/api/RestTemplateClusterManagementServiceTransport.java
```diff
@@ -73,6 +75,27 @@
...
+      messageConverters.add(jackson2HttpMessageConverter);
+    }
+
+    jackson2HttpMessageConverter.setPrettyPrint(false);
+    // the client should use a mapper that would ignore unknown properties in case the server
+    // is a newer version than the client
+    jackson2HttpMessageConverter
+        .setObjectMapper(GeodeJsonMapper.getMapperIgnoringUnknownProperties());
+    // if we don't set the default charset here, the request will use ServletRequest's default
+    // charset which is ISO-8859
```
demery-pivotal, line 97: I think this comment is overly specific. If we don't set the charset, the request will use some default, but it may not be ISO-8859. I've been able to change the default on my Mac and in docker images by:

- setting the `LC_ALL` shell variable.
- specifying the `file.encoding` JVM option (which I think overrides `LC_ALL`, but I'm not sure).

Consider removing the comment. To me, `setDefaultCharset(UTF_8)` is clear enough without it.

But if you want a comment, consider: "... which may not be UTF-8."

PR #5427, opened 2020-08-05
GEODE-8406: Enable early-return in CI scripts for CI-only changes

rhoughton-pivot:
Creates a function that tests the code changes to be:
1) From a pull request in Geode
2) Those changes are more than ci,  geode-book, geode-docs
  ie: code changes, not non-testable changes

### For all changes:

### Note:


ci/scripts/delete_instance.sh
```diff
@@ -29,6 +29,11 @@ while [ -h "$SOURCE" ]; do # resolve $SOURCE until the file is no longer a symli
 done
 SCRIPTDIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
 
+. ${SCRIPTDIR}/short_circuit_pr_tasks.sh
+cd geode
+  is_source_from_pr_testable "ci" "dev-tools" "etc" "geode-book" "geode-docs" || exit 0
```
onichols-pivotal, line 34: what about README.md, LICENSE, or NOTICE in the top level? or other LICENSE/NOTICE files?

ci/scripts/execute_tests.sh
```diff
@@ -34,6 +34,11 @@ if [[ -z "${GRADLE_TASK}" ]]; then
   exit 1
 fi
 
+. ${SCRIPTDIR}/short_circuit_pr_tasks.sh
+cd geode
+  is_source_from_pr_testable "ci" "dev-tools" "etc" "geode-book" "geode-docs" || exit 0
```
onichols-pivotal, line 39: any way to avoid duplicating this list two places?

rhoughton-pivot: If you can think of one, I'm happy to. We could move the directory change into the function itself to make this a one-liner

rhoughton-pivot:
> does this affect PRs only or regular pipelines as well? Do I understand correctly that this will cause the jobs to run super-fast? That will mess with timing statistics for each job...but I guess that's my problem. The whole reason for doing it this way instead of as git resource exclusions is because of required PR checks, yes?

This makes all the jobs run quick, without starting a heavy-lifter, in the event that the PR is non-testable. We could experiment with twiddling the PR resource to two versions, one for "code" and one for "no-code". Code would run as-is, no-code would only set "status:passed". We'd have to be careful about our GitHub API quotas if we did that; the resource uses a lot more API calls if we start filtering by file/directory.

ci/scripts/delete_instance.sh
rhoughton-pivot: Yup, we can add those to the list. But based on number of times those are touched, they were low-priority for testing (and I didn't think about them)

ci/scripts/execute_tests.sh
pivotal-jbarrett: Is it necessary to have the arguments at all given that they are the same in every call? Seems you could whittle this down to a single line `${SCRIPTDIR}/short_circuit_pr_tasks.sh && exit 0`, where `short_circuit_pr_tasks.sh` returns true if it should short circuit.

rhoughton-pivot:
@pivotal-jbarrett I could have, but chose instead to have it be a modular function that takes input. Makes it clear from the outside what is being tested. Or, call it without any args, and the default is only the [ci] directory, which is honestly the only one I care about

onichols-pivotal:
keep the modular thing, fine, but can we have a wrapper script that encapsulates the magic list, so we just call that script 6 places?

ci/scripts/execute_tests.sh
pivotal-jbarrett: I have little skin in this game but if it doesn't matter than I would make it simple but if its not going to change over time then its probably fine as it is.

PurelyApplied: I also have little skin in the game, but here are some options that jump out at me:

* Like Jake said, let `short_circuit_pr_tasks` be the source of truth for what directories are excluded.
* Take an arg list, and specify those args in the task's `yml`.  Jinja or yml anchoring could reduce duplication.
* Get exceedingly convoluted by having this check be an independent Concourse job which outputs a sentinel file, which test jobs use to determine whether or not to short-circuit to green.

I feel like the first option is best for the first iteration.

PR #5436, opened 2020-08-07
GEODE-7672: add dunit test to verify OQL index after PR clear.

jinmeiliao:
* allow ClusterStartupRule to launch dunit default locator

### DIFF
--- a/geode-core/src/distributedTest/java/org/apache/geode/cache/query/partitioned/PRClearQueryIndexDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/cache/query/partitioned/PRClearQueryIndexDUnitTest.java
@@ -20,9 +20,11 @@ import static org.apache.geode.test.awaitility.GeodeAwaitility.await;
 import static org.apache.geode.test.junit.rules.VMProvider.invokeInEveryMember;
 import static org.assertj.core.api.Assertions.assertThat;
 
+import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.IntStream;
 
+import org.junit.After;
 import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.ClassRule;
@@ -41,21 +43,28 @@ import org.apache.geode.cache.query.SelectResults;
 import org.apache.geode.cache.query.data.City;
 import org.apache.geode.internal.cache.InternalCache;
 import org.apache.geode.test.dunit.AsyncInvocation;
+import org.apache.geode.test.dunit.DUnitBlackboard;
 import org.apache.geode.test.dunit.rules.ClusterStartupRule;
 import org.apache.geode.test.dunit.rules.MemberVM;
 import org.apache.geode.test.junit.rules.ClientCacheRule;
+import org.apache.geode.test.junit.rules.ExecutorServiceRule;
 
 public class PRClearQueryIndexDUnitTest {
   public static final String MUMBAI_QUERY = "select * from /cities c where c.name = 'MUMBAI'";
   public static final String ID_10_QUERY = "select * from /cities c where c.id = 10";
   @ClassRule
-  public static ClusterStartupRule cluster = new ClusterStartupRule(2, true);
+  public static ClusterStartupRule cluster = new ClusterStartupRule(4, true);
 
   private static MemberVM server1;
   private static MemberVM server2;
 
+  private static DUnitBlackboard blackboard;
+
   @Rule
   public ClientCacheRule clientCacheRule = new ClientCacheRule();
+
+  @Rule
+  public ExecutorServiceRule executor = ExecutorServiceRule.builder().build();
 
   private ClientCache clientCache;
   private Region cities;
@@ -64,10 +73,10 @@ public class PRClearQueryIndexDUnitTest {
   @BeforeClass
   public static void beforeClass() {
     int locatorPort = ClusterStartupRule.getDUnitLocatorPort();
-    server1 = cluster.startServerVM(0, s -> s.withConnectionToLocator(locatorPort)
+    server1 = cluster.startServerVM(1, s -> s.withConnectionToLocator(locatorPort)
         .withProperty(SERIALIZABLE_OBJECT_FILTER, "org.apache.geode.cache.query.data.*")
         .withRegion(RegionShortcut.PARTITION, "cities"));
-    server2 = cluster.startServerVM(1, s -> s.withConnectionToLocator(locatorPort)
+    server2 = cluster.startServerVM(2, s -> s.withConnectionToLocator(locatorPort)
         .withProperty(SERIALIZABLE_OBJECT_FILTER, "org.apache.geode.cache.query.data.*")
         .withRegion(RegionShortcut.PARTITION, "cities"));
 
@@ -113,8 +122,8 @@ public class PRClearQueryIndexDUnitTest {
   }
 
   @Test
-  public void createAndRemoveIndexWhileClear() throws Exception {
-    IntStream.range(0, 20).forEach(i -> cities.put(i, new City(i)));
+  public void createIndexWhileClear() throws Exception {
+    IntStream.range(0, 100).forEach(i -> cities.put(i, new City(i)));
 
     // create index while clear
     AsyncInvocation createIndex = server1.invokeAsync("create index", () -> {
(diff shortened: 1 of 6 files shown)


geode-core/src/distributedTest/java/org/apache/geode/cache/query/partitioned/PRClearQueryIndexDUnitTest.java
```diff
@@ -0,0 +1,157 @@
...
+
+    server1.invokeAsync(() -> {
+      Cache cache = ClusterStartupRule.getCache();
+      Region region = cache.getRegion("cities");
+      IntStream.range(0, 100).forEach(i -> region.put(i, new City(i)));
+      region.clear();
+    });
+
+    await().pollDelay(5, TimeUnit.MILLISECONDS).untilAsserted(() -> {
+      assertThat(((SelectResults) query.execute()).size()).isEqualTo(0);
```
jchen21, line 152: The assertions here could be executed before `region.clear()` in line 148, or even before the region is populated. There is no guarantee when the code block in `invokeAsync` is executed.

geode-core/src/distributedTest/java/org/apache/geode/cache/query/partitioned/PRClearQueryIndexDUnitTest.java
```diff
@@ -0,0 +1,157 @@
...
+  public void before() throws Exception {
+    int locatorPort = ClusterStartupRule.getDUnitLocatorPort();
+    clientCache = clientCacheRule.withLocatorConnection(locatorPort).createCache();
+  }
+
+  @Test
+  public void verifyQuerySucceedsAfterClear() throws Exception {
+    Region cities = clientCacheRule.createProxyRegion("cities");
+    // put in some data
+    IntStream.range(0, 100).forEach(i -> cities.put(i, new City(i)));
```
jchen21, line 99: The `City` class is defined in another package used by other tests. Any changes in `City` could potentially break the test. I would recommend using a class defined within the test itself. It is easier to maintain the code.

jinmeiliao: Yes, the PR is also to make sure this change won't break other tests. It's better not to make a different class if we can just add an id field and also not breaking others.

geode-core/src/distributedTest/java/org/apache/geode/cache/query/partitioned/PRClearQueryIndexDUnitTest.java
jinmeiliao: yes, it will keep on executing the query until it's returning 0 results.

There is a slight chance that this could get executed even before the region is populated, that's why I put a pollDelay there to reduce this chance.

jchen21: I am not sure I understand the purpose of this test. Depends on the thread scheduling, the query result size could be non-zero. e.g. if the query is executed after populating the region, but before clearing the region. 

The test result is non-deterministic. This test is flaky.

jinmeiliao: the `await().untilAsserted()` method will keep on executing the code block inside until it's successful or until it timed out. This test is trying to execute `clear()` and `query()` at the same time, and make sure that query will eventually return 0 results with no errors.

geode-core/src/distributedTest/java/org/apache/geode/cache/query/partitioned/PRClearQueryIndexDUnitTest.java
```diff
@@ -0,0 +1,369 @@
...
+        .withProperty(SERIALIZABLE_OBJECT_FILTER, "org.apache.geode.cache.query.data.*")
+        .withRegion(RegionShortcut.PARTITION, "cities"));
+
+    server1.invoke(() -> {
+      Cache cache = ClusterStartupRule.getCache();
+      Region region = cache.getRegion("cities");
+      // create indexes
+      QueryService queryService = cache.getQueryService();
+      queryService.createKeyIndex("cityId", "c.id", "/cities c");
+      queryService.createIndex("cityName", "c.name", "/cities c");
```
agingade, line 90: Another most commonly used index is range index; which gets created when multiple iterators are used in index expression, can you please add that one into the test.
qs.createIndex("sIndex", "pos.secId",
        "portfolio.values val, val.positions.values pos");

geode-core/src/distributedTest/java/org/apache/geode/cache/query/partitioned/PRClearQueryIndexDUnitTest.java
```diff
@@ -0,0 +1,369 @@
...
+    // create index while clear
+    AsyncInvocation createIndex = server1.invokeAsync("create index", () -> {
+      Cache cache = ClusterStartupRule.getCache();
+      QueryService queryService = cache.getQueryService();
+      Index cityZip = queryService.createIndex("cityZip", "c.zip", "/cities c");
+      assertThat(cityZip).isNotNull();
+    });
+
+    // do clear at the same time
+    cities.clear();
```
agingade, line 138: If the idea is to get clear and index to be happening parallel, this may not guarantee that. There is a IndexManager.testHook that can be used to control index creation. And having more entries will help chances of occurring clear and create-index in parallel.

geode-core/src/distributedTest/java/org/apache/geode/cache/query/partitioned/PRClearQueryIndexDUnitTest.java
```diff
@@ -0,0 +1,369 @@
...
+
+    // do clear at the same time
+    replicateCities.clear();
+    createIndex.await();
+
+    QueryService queryService = clientCache.getQueryService();
+    Query query =
+        queryService
+            .newQuery("select * from /replicateCities c where c.zip < " + (City.ZIP_START + 10));
+    assertThat(((SelectResults) query.execute()).size()).isEqualTo(0);
```
agingade, line 181: This is a way to see how many entries are there in index. With this we need to make sure the query has used index. 
Along with that it will be nice to validate the index storage size: it could be done by calling  isEmpty() on Index class or using the index stats (as done in other places in this test)

geode-core/src/distributedTest/java/org/apache/geode/cache/query/partitioned/PRClearQueryIndexDUnitTest.java
jinmeiliao: I believe even with just one iteration, it's still a range index. There are only 2 kinds of index now: range or key. Hash is deprecated.

geode-core/src/distributedTest/java/org/apache/geode/cache/query/partitioned/PRClearQueryIndexDUnitTest.java
jinmeiliao: I would prefer not to use testHooks if I can help it. I increased the region size to 1000 and do clear for multiple times to increase the chance of concurrency.

mhansonp:
I made some updates on mhansonp minorcleanup to resolve warnings.

geode-core/src/distributedTest/java/org/apache/geode/cache/query/partitioned/PRClearQueryIndexDUnitTest.java
```diff
@@ -0,0 +1,374 @@
...
+        } catch (ServerOperationException e) {
+          assertThat(e.getCause().getMessage())
+              .contains("Unable to clear all the buckets from the partitioned region cities")
+              .contains("either data (buckets) moved or member departed");
+        }
+        QueryService queryService = clientCache.getQueryService();
+        Query query = queryService.newQuery(MUMBAI_QUERY);
+        Query query2 = queryService.newQuery(ID_10_QUERY);
+        assertThat(((SelectResults) query.execute()).size()).isEqualTo(0);
+        assertThat(((SelectResults) query2.execute()).size()).isEqualTo(0);
```
mhansonp, line 350: The asserts here should probably be await.until's. Judging from the DUnit failure.

geode-core/src/distributedTest/java/org/apache/geode/cache/query/partitioned/PRClearQueryIndexDUnitTest.java
```diff
@@ -0,0 +1,374 @@
...
+
+  private static void verifyIndexesAfterClear(String regionName, String... indexes) {
+    InternalCache internalCache = ClusterStartupRule.getCache();
+    QueryService qs = internalCache.getQueryService();
+    Region region = internalCache.getRegion(regionName);
+    assertThat(region.size()).isEqualTo(0);
+    for (String indexName : indexes) {
+      Index index = qs.getIndex(region, indexName);
+      IndexStatistics statistics = index.getStatistics();
+      assertThat(statistics.getNumberOfKeys()).isEqualTo(0);
```
mhansonp, line 266: Would this be better as an await.untilAsserted?

jinmeiliao: this method is called AFTER clear is done, so it should be 0 at this point all the time.

PR #5447, opened 2020-08-11
Changes to help reduce the occurence of GEODE-8422

mhansonp:

### DIFF
--- a/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/GIIDeltaDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/GIIDeltaDUnitTest.java
@@ -497,7 +497,7 @@ public class GIIDeltaDUnitTest extends JUnit4CacheTestCase {
 
     // If fullGII, the key size in gii chunk is 3, i.e. key1,key3,key5. key2 is GCed.
     // If delta GII, the key size should be 1 (key5(T) which is unfinished operation)
-    verifyDeltaSizeFromStats(R, 3, 0);
+    verifyDeltaSizeFromStats(R, 1, 1);
   }
 
   /**
@@ -1534,9 +1534,9 @@ public class GIIDeltaDUnitTest extends JUnit4CacheTestCase {
     assertEquals(0, count);
     verifyDeltaSizeFromStats(R, 1, 1); // deltaGII, key1 in delta
 
-    // tombstone key2, key5 should be GCed at R
-    verifyTombstoneExist(R, "key2", false, false);
-    verifyTombstoneExist(R, "key5", false, false);
+    // tombstone key2, key5 should still exist and expired at R
+    verifyTombstoneExist(R, "key2", true, true);
+    verifyTombstoneExist(R, "key5", true, true);
 
     // tombstone key2, key5 should still exist and expired at P
     verifyTombstoneExist(P, "key2", true, true);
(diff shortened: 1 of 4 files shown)


geode-core/src/test/java/org/apache/geode/internal/cache/TombstoneServiceTest.java
```diff
@@ -0,0 +1,79 @@
...
+  RegionMap regionMap = mock(RegionMap.class);
+  RegionEntry entry = mock(RegionEntry.class);
+  DistributedRegion region = mock(DistributedRegion.class);
+  VersionTag destroyedVersion = mock(VersionTag.class);
+
+  @Before
+  public void setUp() throws Exception {}
+
+  @After
+  public void tearDown() throws Exception {}
```
DonalEvans, line 34: This method is empty and can be removed.

geode-core/src/test/java/org/apache/geode/internal/cache/TombstoneServiceTest.java
```diff
@@ -0,0 +1,79 @@
...
+  CachePerfStats stats = mock(CachePerfStats.class);
+  CancelCriterion cancelCriterion = mock(CancelCriterion.class);
+  ExecutorService executor = mock(ExecutorService.class);
+  RegionMap regionMap = mock(RegionMap.class);
+  RegionEntry entry = mock(RegionEntry.class);
+  DistributedRegion region = mock(DistributedRegion.class);
+  VersionTag destroyedVersion = mock(VersionTag.class);
+
+  @Before
+  public void setUp() throws Exception {}
```
DonalEvans, line 31: The creation and assignment of the mocks should be moved to the `setUp()` method to ensure that fresh mocks are created for each test method and hence prevent multiple tests acting on the same instance of a mock and producing unexpected results.

geode-core/src/test/java/org/apache/geode/internal/cache/TombstoneServiceTest.java
```diff
@@ -0,0 +1,79 @@
...
+        new TombstoneService.ReplicateTombstoneSweeper(cacheTime, stats,
+            cancelCriterion, executor);
+    TombstoneService.Tombstone tombstone =
+        new TombstoneService.Tombstone(entry, region, destroyedVersion);
+    tombstone.entry = entry;
+
+    replicateTombstoneSweeper.expireTombstone(tombstone);
+    replicateTombstoneSweeper.expireBatch();
+    verify(region, Mockito.atLeastOnce()).getVersionVector();
+    verify(regionVersionVector, Mockito.atLeastOnce()).recordGCVersion(tombstone.getMemberID(),
```
DonalEvans, line 74: For these verifications, would it be reasonable to assert the specific number of invocations of `getVersionVector()` and `recordGCVersion()` we expect to see using `Mockito.times(1)` rather than just that it's at least one?

geode-core/src/test/java/org/apache/geode/internal/cache/TombstoneServiceTest.java
```diff
@@ -0,0 +1,79 @@
+package org.apache.geode.internal.cache;
```
onichols-pivotal, line 1: please include the standard apache copyright header

geode-core/src/test/java/org/apache/geode/internal/cache/TombstoneServiceTest.java
mhansonp: Done

geode-core/src/test/java/org/apache/geode/internal/cache/TombstoneServiceTest.java
mhansonp: Done

geode-core/src/test/java/org/apache/geode/internal/cache/TombstoneServiceTest.java
mhansonp: changed the code to actually check a call to removeTombstone which is what we really care about. Using times 1 now.

geode-core/src/test/java/org/apache/geode/internal/cache/TombstoneServiceTest.java
mhansonp: done.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/GIIDeltaDUnitTest.java
```diff
@@ -497,7 +497,7 @@ public void run() {
 
     // If fullGII, the key size in gii chunk is 3, i.e. key1,key3,key5. key2 is GCed.
     // If delta GII, the key size should be 1 (key5(T) which is unfinished operation)
-    verifyDeltaSizeFromStats(R, 3, 0);
+    verifyDeltaSizeFromStats(R, 1, 1);
```
jchen21, line 500: Does the comment above have to be updated?

PR #5452, opened 2020-08-13
DRAFT: fix create region with gw-sender vs alter region --gw-sender d…

albertogpz:
…ifference

This is a draft PR to see if it fixes, without any undesired effect, the difference in behavior in Geode between creating a region with a gateway sender or creating a region without it and then altering the region to add the gateway sender. In the second case, when executing transactions on partitioned regions, all members in the region are notified and not only those that host the entries corresponding to the transaction. 
The notification in the members who do not host entries for the transaction provokes that they try to queue the event in the gateway sender queue but as they do not host a bucket for the entry in the event, they are queued in ParallelGatewaySenderQueue.tempQueue. These events are removed from there when the event is sent by the primary gateway sender but if the remote site is stopped, these events stored in tempQueue could end up filling up the heap of the member.

### For all changes:

### Note:


onichols-pivotal:
last activity was some while ago, please close this PR if no longer needed

albertogpz:
See the fix in GEODE-8455

PR #5462, opened 2020-08-18
GEODE-8435: restore ability to connect gfsh by serialization version

onichols-pivotal:
This used to work but was broken recently by changes to rely exclusively on product version.

Although Geode releases were never impacted, loss of this feature causes quite a headache for developers that don't typically pass a product version in their gradle build command, leading to unintuitive errors like `Cannot use a 0.0.0 gfsh client to connect to a 0.0.0 cluster`

Serialization version is the surest way to check gfsh compatibility, since in general it is unwise to make assumptions about product version because it is not intrinsic to Geode.

This PR restores the serialization version check, which was lost during GEODE-8331.  However, serialization version was only exposed in Geode 1.12, and since we are now interoperable back to Geode 1.10, the product version range check is retained as a fallback.

### DIFF
--- a/geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/ConnectCommand.java
+++ b/geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/ConnectCommand.java
@@ -171,6 +171,17 @@ public class ConnectCommand extends OfflineGfshCommand {
     // since 1.14, only allow gfsh to connect to cluster that's older than 1.10
     String remoteVersion = null;
     String gfshVersion = gfsh.getVersion();
+    // First: see if serialization versions matches (only works with Geode 1.12 or later cluster)
+    try {
+      String gfshGeodeSerializationVersion = gfsh.getGeodeSerializationVersion();
+      String remoteGeodeSerializationVersion = invoker.getRemoteGeodeSerializationVersion();
+      if (gfshGeodeSerializationVersion.equals(remoteGeodeSerializationVersion)) {
+        return result;
+      }
+    } catch (Exception ignore) {
+    }
+
+    // Fallback: make assumptions based on the product version
     try {
       remoteVersion = invoker.getRemoteVersion();
       int minorVersion = Integer.parseInt(versionComponent(remoteVersion, VERSION_MINOR));


geode-gfsh/src/test/java/org/apache/geode/management/internal/cli/commands/ConnectCommandTest.java
```diff
@@ -417,4 +417,21 @@ public void connectToManagerBefore1_10() {
...
+    when(operationInvoker.isConnected()).thenReturn(true);
+
+    ResultModel resultModel = new ResultModel();
+    when(connectCommand.jmxConnect(any(), anyBoolean(), any(), any(), anyBoolean()))
+        .thenReturn(resultModel);
+
+    gfshParserRule.executeAndAssertThat(connectCommand, "connect --locator=localhost:4040")
+        .statusIsSuccess()
+        .doesNotContainOutput("Cannot use a 0.0.0 gfsh client to connect to a 0.0.0 cluster");
+  }
```
Bill, line 436: This is my understanding of the equivalence classes for testing are (rsv = remote serialization version read via JMX, rmv = remote marketing version read via JMX)

```
rsv >= 1.12 && rmv == dont-care
   compatible, covered by connectToManagerBySerializationVersion()
rsv < 1.12
   impossible
rsv unknown
   rmv >= 1.10
     compatible, covered by connectToManagerOlderThan1_10() sic
   rmv < 1.10
      incompatible
   rmv unknown
      incompatible
```

By the way `connectToManagerOlderThan1_10()` is mis-named: it's actually connecting to version 1.10—not a version _older_ than 1.10. Please fix this name.

But then I also see these tests:

```
connectToManagerWithDifferentMajorVersion()
connectToManagerWithDifferentMinorVersion()
```

So major and minor versions seem to be important too. But only for certain version ranges? Actually `ConnectCommand` does not, in general respect semver at all. I think those tests need to be removed.

onichols-pivotal: The different-minor-version test was a holdover from the old days when gfsh was locked to same major&minor as the cluster.  I removed it.  I think major versions still play a role, as we need to carefully consider how far into the future we are willing to guarantee forward compatibility.  It seems safest for now to presume that gfsh 1.x won't be interoperable with a future Geode 2.x.  The reverse might be more plausible but we can unlock backward major version compatibility if and when the time comes.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/ConnectCommand.java
```diff
@@ -186,6 +186,16 @@ public ResultModel connect(
...
+    // fallback: see if serialization versions matches (Geode 1.12 or later cluster)
+    try {
+      String gfshGeodeSerializationVersion = gfsh.getGeodeSerializationVersion();
+      String remoteGeodeSerializationVersion = invoker.getRemoteGeodeSerializationVersion();
+      if (gfshGeodeSerializationVersion.equals(remoteGeodeSerializationVersion)) {
+        return result;
+      }
+    } catch (Exception ignore) {
+    }
+
```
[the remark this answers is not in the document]
onichols-pivotal: I've moved the logic for determining compatible clusters into a central method to decouple from the network calls needed to get the necessary info.  It's a lot easier now to test just this method in isolation.

Rather than add even further assumptions about product name (even if it were available from the remote cluster) I've constrained the logic that uses product version to the narrowest possible scenario (specific known versions and only used if the remote cluster does not report serialization version)

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/ConnectCommand.java
```diff
@@ -168,26 +168,30 @@ public ResultModel connect(
...
     try {
       remoteVersion = invoker.getRemoteVersion();
-      int minorVersion = Integer.parseInt(versionComponent(remoteVersion, VERSION_MINOR));
-      if (versionComponent(remoteVersion, VERSION_MAJOR).equals("1") && minorVersion >= 10 ||
-          versionComponent(remoteVersion, VERSION_MAJOR).equals("9") && minorVersion >= 9) {
-        InfoResultModel versionInfo = result.addInfo("versionInfo");
-        versionInfo.addLine("You are connected to a cluster of version: " + remoteVersion);
-        return result;
+      try {
+        remoteGeodeSerializationVersion = invoker.getRemoteGeodeSerializationVersion();
```
jinmeiliao, line 176: I think we should try to get the geodeSeriliazationVersion first since that's introduced later than the remoteVersion. If we can't get the remoteVersion, we definitely would fail at getting the remoteGeodeSerilizationVersion

onichols-pivotal: both are needed before calling `isCompatible`. so the only optimization is to skip the 2nd call to get remote serialization version if the first call to get product version failed (as this would indicate a pre-1.5 cluster or some network issue).  I can remove this optimization if you think it would make the code more readable.

onichols-pivotal: I can't tell what change you are asking for.  This PR already gets the remote version first, because there would be no reason to try to get remote serialization version if that fails...

jinmeiliao: I think I misread your code.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/ConnectCommand.java
```diff
@@ -197,6 +201,32 @@ public ResultModel connect(
...
+   * because remote serialization version was not exposed until 1.12, but we are compatible back to
+   * 1.10, then any 1.x remote serialization version implies compatibility; otherwise make some
+   * narrow assumptions about product version numbers known to be associated with 1.10/1.11 to fill
+   * the gap.
+   *
+   * It seems reasonable to commit to future compatibility with any future Geode of the same major,
+   * but we should probably draw the line somewhere and not promise that gfsh 1.x will be eternally
+   * compatible with Geode 2.x and later
+   */
+  static boolean isCompatible(String ourSerializationVersion, String remoteVersion,
```
jinmeiliao, line 214: Maybe call this method shouldConnect?

Your logic is correct, but maybe to make this more readable and have the code reflect our train of thought better, you can consider this logic flow:

```
// client should connect to cluster version newer than 1.10
boolean shouldConnect (String ourSerializationVersion, String remoteVersion, String remoteSerializationVersion) {
  if (remoteVersion == null) { // pre 1.5
     return false;
  }
  // post 1.12, this should always return true, since it's definitely after 1.10.  I think this should fix the hydra test 
  // not sure if we need the ourSerilizationVersion to make this decision
  if (remoteSerializationVersion !=null) {
      return true;
  }
  // after 1.5 but before 1.12, use remoteVersion to deteremin if it's after 1.10 or 9.9
  ......

}
```

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/ConnectCommand.java
jinmeiliao: Can you put these two remote call in the same try block, this "try" inside "try" seems confusing:

```
try {
      // serialization version is  introduced later than remoteVersion, have to call them in the right order
      remoteVersion = invoker.getRemoteVersion();
      remoteSerializationVersion = invoker.getRemoteGeodeSerializationVersion();
    }
    catch (Exception ex) {
      // if unable to get the remote version, we are certainly talking to a pre-1.5 cluster
      // if unable to get the remote serilization version, we are certainly talking to a pre-1.12 cluster
      gfsh.logInfo("failed to get the the remote version.", ex);
    }
```

onichols-pivotal: combining like this would cause gfsh to print "failed to get the the remote version." when connecting to 1.10 or 1.11

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/ConnectCommand.java
onichols-pivotal: it sounds like you want to guarantee future compatibility with ALL Geode majors.  Are we sure we can promise that today's gfsh should connect to some future Geode 2.0 or 3.0 cluster?  I included ourSerializationVersion to limit that promise to same-major.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/ConnectCommand.java
jinmeiliao: Oh, we use gfsh to log, maybe instead of calling `gfsh.logInfo` we should just call `gfsh.logToFile`

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/ConnectCommand.java
jinmeiliao: sure, if we want to limit that promise to same-major, we can add the compare major in the 2nd if block

onichols-pivotal: I have renamed the method to `shouldConnect` and arranged it as you suggested to make it more readable, thanks

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/ConnectCommand.java
onichols-pivotal: factored out a method to make this more readable (and add the comment on ignored exception that Bill requested)

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/ConnectCommand.java
```diff
@@ -168,26 +168,27 @@ public ResultModel connect(
...
     try {
       remoteVersion = invoker.getRemoteVersion();
-      int minorVersion = Integer.parseInt(versionComponent(remoteVersion, VERSION_MINOR));
-      if (versionComponent(remoteVersion, VERSION_MAJOR).equals("1") && minorVersion >= 10 ||
-          versionComponent(remoteVersion, VERSION_MAJOR).equals("9") && minorVersion >= 9) {
-        InfoResultModel versionInfo = result.addInfo("versionInfo");
-        versionInfo.addLine("You are connected to a cluster of version: " + remoteVersion);
-        return result;
-      }
+      remoteGeodeSerializationVersion = getRemoteSerializationVersion(invoker);
```
jinmeiliao, line 175: This is fine, but if you just call the invoker method directly, and user gfsh.logToFile in the catch block, things should be fine too. The difference is that user won't see this "failed to get the remote version" message when they connect to pre-1.5 cluster, but they would still see "you can't use a xxx version gfsh to connect to this cluster"

PR #5464, opened 2020-08-18
GEODE-8432: use regionPath directly instead of getRegion when put eve…

gesterzhou:
…nt into parallelGatewaySenderQueue

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
@@ -693,15 +693,19 @@ public class ParallelGatewaySenderQueue implements RegionQueue {
 
     boolean isDREvent = isDREvent(sender.getCache(), value);
 
-    Region region = value.getRegion();
-    String regionPath = null;
-    if (isDREvent) {
-      regionPath = region.getFullPath();
-    } else {
+    String regionPath = value.getRegionPath();
+    if (!isDREvent) {
+      Region region = sender.getCache().getRegion(regionPath);
+      if (region == null) {
+        if (isDebugEnabled) {
+          logger.debug("The PR " + regionPath + " has not finished initializing.");
+        }
+        region = value.getRegion();
+      }
       regionPath = ColocationHelper.getLeaderRegion((PartitionedRegion) region).getFullPath();
     }
     if (isDebugEnabled) {
-      logger.debug("Put is for the region {}", region);
+      logger.debug("Put is for the region {}", regionPath);
     }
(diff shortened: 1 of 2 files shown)


geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
```diff
@@ -693,15 +693,19 @@ public boolean put(Object object) throws InterruptedException, CacheException {
...
-      regionPath = region.getFullPath();
-    } else {
+    String regionPath = value.getRegionPath();
+    if (!isDREvent) {
+      Region region = sender.getCache().getRegion(regionPath);
+      if (region == null) {
+        if (isDebugEnabled) {
+          logger.debug("The PR " + regionPath + " has not finished initializing.");
+        }
+        region = value.getRegion();
```
pivotal-eshu, line 703: Should not use value.getRegion() here as it will still cause the hang.

geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
```diff
@@ -693,15 +693,19 @@ public boolean put(Object object) throws InterruptedException, CacheException {
...
     boolean isDREvent = isDREvent(sender.getCache(), value);
 
-    Region region = value.getRegion();
-    String regionPath = null;
-    if (isDREvent) {
-      regionPath = region.getFullPath();
-    } else {
+    String regionPath = value.getRegionPath();
+    if (!isDREvent) {
+      Region region = sender.getCache().getRegion(regionPath);
```
pivotal-eshu, line 698: sender.getCache().getRegion(regionPath) should not be null due to initialization as it invokes region.waitOnInitialization(). If region is null, I think it indicates region is destroyed.

gesterzhou: I will change to sender.getCache().getRegion(regionPath,true). 

It will be null. In the test, the region is destroyed. But we still need to get the root region. The event will be ignored later by if (!this.userRegionNameToShadowPRMap.containsKey(regionPath)) {

geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
gesterzhou: After I use getRegion(path, true), I don't need to call value.getRegion(). 

Otherwise, value.getRegion() is still necessary. When code arrived here, it's normal enqueue (not syncWith enqueue), the event.getRegion() will never be null.

PR #5469, opened 2020-08-21
GEODE-8447: QueryResultFormatter should show dates in localized strings

jinmeiliao:

### DIFF
--- a/boms/geode-all-bom/src/test/resources/expected-pom.xml
+++ b/boms/geode-all-bom/src/test/resources/expected-pom.xml
@@ -68,6 +68,12 @@
         <scope>compile</scope>
       </dependency>
       <dependency>
+        <groupId>com.fasterxml.jackson.datatype</groupId>
+        <artifactId>jackson-datatype-jsr310</artifactId>
+        <version>2.11.2</version>
+        <scope>compile</scope>
+      </dependency>
+      <dependency>
         <groupId>com.fasterxml.jackson.module</groupId>
         <artifactId>jackson-module-scala_2.10</artifactId>
         <version>2.10.0</version>
--- a/geode-core/build.gradle
+++ b/geode-core/build.gradle
@@ -366,6 +366,7 @@ dependencies {
 
     integrationTestRuntimeOnly('org.apache.derby:derby')
     integrationTestRuntimeOnly('xerces:xercesImpl')
+    integrationTestRuntimeOnly('com.fasterxml.jackson.datatype:jackson-datatype-jsr310')
 
     distributedTestImplementation(project(':geode-gfsh'))
     distributedTestImplementation(project(':geode-junit')) {
--- a/geode-core/src/integrationTest/java/org/apache/geode/management/internal/beans/DistributedSystemMBeanIntegrationTest.java
+++ b/geode-core/src/integrationTest/java/org/apache/geode/management/internal/beans/DistributedSystemMBeanIntegrationTest.java
@@ -1,0 +1,149 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package org.apache.geode.management.internal.beans;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.text.SimpleDateFormat;
+import java.time.LocalDate;
+import java.util.Date;
+
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Rule;
+import org.junit.Test;
+
+import org.apache.geode.cache.Region;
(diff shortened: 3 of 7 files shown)


geode-core/src/integrationTest/java/org/apache/geode/management/internal/beans/DistributedSystemMBeanIntegrationTest.java
```diff
@@ -0,0 +1,149 @@
...
+import org.apache.geode.management.DistributedSystemMXBean;
+import org.apache.geode.management.internal.json.QueryResultFormatter;
+import org.apache.geode.test.junit.assertions.TabularResultModelAssert;
+import org.apache.geode.test.junit.rules.GfshCommandRule;
+import org.apache.geode.test.junit.rules.MBeanServerConnectionRule;
+import org.apache.geode.test.junit.rules.ServerStarterRule;
+
+public class DistributedSystemMBeanIntegrationTest {
+
+  public static final String SELECT = "select * from /testRegion r where r.id=1";
```
agingade, line 40: Does it also handles the date format used by the application/user in the query.
(From test SelectToDateJUnitTest.java):
 "select * from /test p where p.createDate = to_date('05202012100559', 'MMddyyyyHHmmss')

jinmeiliao: It formats the the date in the select results.

geode-gfsh/src/test/java/org/apache/geode/management/internal/cli/json/QueryResultFormatterTest.java
```diff
@@ -117,15 +119,19 @@ public void testPrimitives() throws Exception {
...
+    String expectedString =
+        new SimpleDateFormat(QueryResultFormatter.DATE_FORMAT_PATTERN).format(date);
     QueryResultFormatter javaDateResult =
-        new QueryResultFormatter(100).add(RESULT, new java.util.Date(0));
+        new QueryResultFormatter(100).add(RESULT, date);
     checkResult(javaDateResult,
-        "{\"result\":[[\"java.util.Date\",\"1970-01-01T00:00:00.000+0000\"]]}");
+        "{\"result\":[[\"java.util.Date\",\"" + expectedString + "\"]]}");
 
+    java.sql.Date sqlDatee = new java.sql.Date(0);
```
jchen21, line 130: Is `sqlDatee` a typo?

boms/geode-all-bom/src/test/resources/expected-pom.xml
```diff
@@ -67,6 +67,12 @@
         <version>2.9.8</version>
         <scope>compile</scope>
       </dependency>
+      <dependency>
+        <groupId>com.fasterxml.jackson.datatype</groupId>
```
jchen21, line 71: Can you point me where this new dependency is used in the test? I don't see it is used in the tests of this pull request.

geode-core/src/main/java/org/apache/geode/management/internal/json/AbstractJSONFormatter.java
```diff
@@ -94,6 +94,8 @@ public AbstractJSONFormatter(int maxCollectionElements, int serializationDepth,
 
     // register the custom module
     mapper.registerModule(mapperModule);
+    // to support jdk8 java.time if jackson-datatype-jsr310 is included in the classpath
```
jchen21, line 97: Can you elaborate a bit why `jackson-datatype-jsr310` is needed? I am not sure why it is necessary to add it.

geode-gfsh/src/test/java/org/apache/geode/management/internal/cli/json/QueryResultFormatterTest.java
jinmeiliao: yeah, should be.e

geode-core/src/main/java/org/apache/geode/management/internal/json/AbstractJSONFormatter.java
jinmeiliao: the test shows an example using java.time.LocalTime object. If without this line and without the jackson-datatype-jsr310.jar in the classpath, the java.time.LocalTime will not be formatted correctly.

boms/geode-all-bom/src/test/resources/expected-pom.xml
jinmeiliao: The test uses a java.time.LocalTime as one of the values in the domain object to show it will be displayed correctly.

geode-core/src/main/java/org/apache/geode/management/internal/json/AbstractJSONFormatter.java
jchen21: That means if the user uses ` java.time.LocalTime`, it also requires `jackson-datatype-jsr310` to format it properly. However, this dependency is for `integrationTestRuntimeOnly` in `build.gradle`. In `expected-pom.xml`, its scope is `compile`.  We might have to check the open source license requirements for new third party dependency.

jinmeiliao: It's only required for testing. If user's domain object includes that type and they would like it to be formatted correctly, then they will need to include the jar in the classpath (Geode is not shipping with it). We are just calling the `findAndRegisterModuel` in case users have the jar in the classpath.

geode-core/build.gradle
```diff
@@ -366,6 +366,7 @@ dependencies {
 
     integrationTestRuntimeOnly('org.apache.derby:derby')
     integrationTestRuntimeOnly('xerces:xercesImpl')
+    integrationTestRuntimeOnly('com.fasterxml.jackson.datatype:jackson-datatype-jsr310')
```
onichols-pivotal, line 369: this seems like the wrong scope.  don't we need jackson-datatype-jdr310 jar to be included in the shipping product, not just available to this one test?

PR #5473, opened 2020-08-24
GEODE-7864: Fix some LGTM alerts and suppress some false positives

DonalEvans:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/cache/query/internal/NullToken.java
+++ b/geode-core/src/main/java/org/apache/geode/cache/query/internal/NullToken.java
@@ -34,6 +34,8 @@ import org.apache.geode.internal.serialization.SerializationContext;
 public class NullToken implements DataSerializableFixedID, Comparable {
 
   @SuppressWarnings("lgtm[java/useless-null-check]")
+  // The "useless null check" alert is suppressed here as the first time the constructor is called,
+  // QueryService.NULL is actually null, but subsequent times it is not
   public NullToken() {
     Support.assertState(IndexManager.NULL == null, "NULL constant already instantiated");
   }
--- a/geode-core/src/main/java/org/apache/geode/cache/query/internal/NullToken.java
+++ b/geode-core/src/main/java/org/apache/geode/cache/query/internal/NullToken.java
@@ -33,6 +33,7 @@ import org.apache.geode.internal.serialization.SerializationContext;
  */
 public class NullToken implements DataSerializableFixedID, Comparable {
 
+  @SuppressWarnings("lgtm[java/useless-null-check]")
   public NullToken() {
     Support.assertState(IndexManager.NULL == null, "NULL constant already instantiated");
   }
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/Oplog.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/Oplog.java
@@ -1748,29 +1748,33 @@ public class Oplog implements CompactableOplog, Flushable {
                         oplogKeyId));
               }
             }
-            DiskEntry de = drs.getDiskEntry(key);
-            if (de == null) {
-              if (logger.isTraceEnabled(LogMarker.PERSIST_RECOVERY_VERBOSE)) {
-                logger.trace(LogMarker.PERSIST_RECOVERY_VERBOSE,
-                    "readNewEntry oplogKeyId=<{}> drId={} userBits={} oplogOffset={} valueLen={}",
-                    oplogKeyId, drId, userBits, oplogOffset, valueLength);
-              }
-              DiskEntry.RecoveredEntry re = createRecoveredEntry(valueBytes, valueLength, userBits,
-                  getOplogId(), oplogOffset, oplogKeyId, false, version, in);
-              if (tag != null) {
-                re.setVersionTag(tag);
-              }
-              initRecoveredEntry(drs.getDiskRegionView(), drs.initializeRecoveredEntry(key, re));
-              drs.getDiskRegionView().incRecoveredEntryCount();
-              this.stats.incRecoveredEntryCreates();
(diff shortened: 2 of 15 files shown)


geode-core/src/main/java/org/apache/geode/internal/cache/Oplog.java
```diff
@@ -3583,7 +3597,7 @@ private void basicCreate(DiskRegion dr, DiskEntry entry, ValueWrapper value, byt
             logger.trace(LogMarker.PERSIST_WRITES_VERBOSE,
                 "basicCreate: id=<{}> key=<{}> valueOffset={} userBits={} valueLen={} valueBytes={} drId={} versionTag={} oplog#{}",
                 abs(id.getKeyId()), entry.getKey(), startPosForSynchOp, userBits,
-                (value != null ? value.getLength() : 0), value.getBytesAsString(), dr.getId(), tag,
```
mhansonp, line 3586: Did you decrease protection here?

geode-core/src/main/java/org/apache/geode/internal/cache/Oplog.java
```diff
@@ -5653,7 +5667,7 @@ private void deleteFile(final OplogFile olf) {
       if (!olf.f.exists())
         return;
       assert olf.RAFClosed;
-      if (!olf.RAFClosed || olf.raf != null) {
+      if (!olf.RAFClosed && olf.raf != null) {
```
mhansonp, line 5670: Is the first part of the test required? "if (!olf.RAFClosed" It sits below an assert that in order to pass must be true.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
```diff
@@ -2641,9 +2640,7 @@ private VersionedObjectList sendMsgByBucket(final Integer bucketId, PutAllPRMess
             retryTime.waitToRetryNode();
           }
           event.setPossibleDuplicate(true);
-          if (prMsg != null) {
-            prMsg.setPossibleDuplicate(true);
-          }
+          prMsg.setPossibleDuplicate(true);
```
mhansonp, line 2643: Is there a reason you reduced protection after adding it in elsewhere?

geode-core/src/main/java/org/apache/geode/internal/cache/Oplog.java
mhansonp: Actually, with the new code, this if statement should never be true because of the assert above.

geode-core/src/main/java/org/apache/geode/internal/cache/Oplog.java
DonalEvans: In a sense, yes. The warning was being generated because checking if `value` was null here implied that it could be null before/after this check, and that calling methods on it could produce an NPE. However, when examining the code path, `value` should never be null at this point, so the check was redundant.

geode-core/src/main/java/org/apache/geode/internal/cache/Oplog.java
DonalEvans: Good catch. I'll remove the first part.

mhansonp: Other than that, it all looks good.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
DonalEvans: As with the other removed null check in Oplog.java, `prMsg` should never be null in this method, so checking if it is null here simply causes LGTM to assume that it may be null everywhere and complain any time a method is called on it without first checking if it's null.

geode-core/src/main/java/org/apache/geode/cache/query/internal/NullToken.java
```diff
@@ -33,6 +33,7 @@
  */
 public class NullToken implements DataSerializableFixedID, Comparable {
 
+  @SuppressWarnings("lgtm[java/useless-null-check]")
```
moleske, line 36: Since lgtm is not a "normal" warning to suppress, it may be worth explaining this somewhere why we are suppressing it (rather than fixing it).  Also it seems like we're using both `@SuppressWarnings("lgtm[])` and comments `// lgtm[]` should be using one consistently or is there certain use cases that require one or the other?

DonalEvans: Good call on explaining the suppression here and other places the annotation is used. As for the different approaches to suppressing, [LGTM requires that inline suppressions contain no line breaks](https://lgtm.com/help/lgtm/alert-suppression#where-in-code-do-i-add-suppression-comments), which would not be possible for some of the lines due to the spotless plugin automatically wrapping comments that extend too far to the right. In those cases, the @SuppressWarnings annotation is used, but is not used throughout because it would potentially mask other, legitimate alerts in methods that are currently present or that get introduced at a later date.

moleske: Ok, the comments in the code help help.  I do wonder if we should make the link you posted available somewhere (like the readme or contributing.)

And as long as we put `@SuppressWarnings` at the lowest possible level that should prevent the accident masking of other suppressions so that looks good.

DonalEvans: I'll try to write up and add a section about LGTM alerts in the appropriate section of the wiki/docs.

I've tried to use the inline comment approach wherever possible, since that suppresses only the specified warning on the single line with the comment, so it's much more surgical than the annotation, which on a large method could have a fairly big blanket effect.

PR #5476, opened 2020-08-25
GEODE-8455: Fix difference between create region with gw sender and a…

albertogpz:
…lter region with gw sender

Geode behaves differently with respect to transactions when
creating a partitioned region with a parallel gateway sender to
when first the partitioned region is created and then the
parallel gateway sender is added by altering the region.

In the first case (create region with gateway sender), when
sending a transaction for the region, an event is sent
to each server hosting the bucket for the data in the transaction.

In the second case (create region + alter region),
when sending a transaction for the region, an event
is sent to every cache member and not only to those hosting
the bucket for the data in the transaction. This leads to,
in the servers not hosting the bucket for the data,
that the events are stored in the
bucketToTempQueueMap member variable of the
ParallelGatewaySenderQueue.

The behavior in the first case is the right one.

The wrong behavior is provoked by the setting to true of the
requiresNotification variable of the PartitionedRegion
by the distributeUpdatedProfileOnSenderCreation method.
This method should only set this variable to true
in case the sender is serial as it is done when
a partitioned region is initialized.

### For all changes:

### Note:

### DIFF
--- a/geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/WANTestBase.java
+++ b/geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/WANTestBase.java
@@ -1725,13 +1725,12 @@ public class WANTestBase extends DistributedTestCase {
       Integer maxMemory, Integer batchSize, boolean isConflation, boolean isPersistent,
       GatewayEventFilter filter, boolean isManualStart) {
     createSender(dsName, remoteDsId, isParallel, maxMemory, batchSize, isConflation, isPersistent,
-        filter, isManualStart, false, -1);
+        filter, isManualStart, false);
   }
 
   public static void createSender(String dsName, int remoteDsId, boolean isParallel,
       Integer maxMemory, Integer batchSize, boolean isConflation, boolean isPersistent,
-      GatewayEventFilter filter, boolean isManualStart,
-      boolean groupTransactionEvents, int batchTimeInterval) {
+      GatewayEventFilter filter, boolean isManualStart, boolean groupTransactionEvents) {
     final IgnoredException exln = IgnoredException.addIgnoredException("Could not connect");
     try {
       File persistentDirectory =
@@ -1745,9 +1744,6 @@ public class WANTestBase extends DistributedTestCase {
           GatewaySender.DEFAULT_SOCKET_BUFFER_SIZE);
       gateway.setGroupTransactionEvents(groupTransactionEvents);
       gateway.create(dsName, remoteDsId);
-      if (batchTimeInterval > 0) {
-        gateway.setBatchTimeInterval(batchTimeInterval);
-      }
     } finally {
       exln.remove();
     }
--- a/geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/WANTestBase.java
+++ b/geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/WANTestBase.java
@@ -1141,7 +1141,7 @@ public class WANTestBase extends DistributedTestCase {
   }
 
   public static void enableConflation(String senderId) {
-    AbstractGatewaySender sender = getAbstractGatewaySender(senderId);
+    AbstractGatewaySender sender = (AbstractGatewaySender) getGatewaySender(senderId);
     sender.test_setBatchConflationEnabled(true);
   }
 
@@ -3100,7 +3100,7 @@ public class WANTestBase extends DistributedTestCase {
     IgnoredException exp1 =
         IgnoredException.addIgnoredException(ForceReattemptException.class.getName());
     try {
-      AbstractGatewaySender sender = getAbstractGatewaySender(senderId);
+      AbstractGatewaySender sender = (AbstractGatewaySender) getGatewaySender(senderId);
       if (sender.isPrimary()) {
         logger.info("Gateway sender is killed by a test");
         cache.getDistributedSystem().disconnect();
@@ -3112,18 +3112,6 @@ public class WANTestBase extends DistributedTestCase {
       exp1.remove();
       exln.remove();
     }
-  }
-
-  private static AbstractGatewaySender getAbstractGatewaySender(String senderId) {
-    Set<GatewaySender> senders = cache.getGatewaySenders();
-    AbstractGatewaySender sender = null;
-    for (GatewaySender s : senders) {
-      if (s.getId().equals(senderId)) {
-        sender = (AbstractGatewaySender) s;
-        break;
-      }
-    }
-    return sender;
   }
 
   public static void killSender() {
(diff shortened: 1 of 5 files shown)


geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/parallel/ParallelWANPropagationDUnitTest.java
```diff
@@ -190,6 +190,16 @@ protected SerializableRunnableIF createPartitionedRegionRedundancy1Runnable() {
...
+  protected SerializableRunnableIF createPartitionedRegionRedundancy1RunnableNoSenders() {
+    return () -> WANTestBase.createPartitionedRegion(getTestMethodName() + "_PR", null, 1, 100,
+        isOffHeap());
+  }
+
+  protected SerializableRunnableIF createPartitionedRegionRedundancy0RunnableNoSenders() {
+    return () -> WANTestBase.createPartitionedRegion(getTestMethodName() + "_PR", null, 0, 100,
+        isOffHeap());
+  }
+
```
DonalEvans, line 202: These methods are very similar. Would it be possible to replace them with one method which takes an `int` argument for the desired redundancy level?

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/WANTestBase.java
```diff
@@ -3599,21 +3444,42 @@ public static Long getNumberOfEntriesInVM(final String senderId) {
...
-  public static void verifyQueueSize(String senderId, int size) {
+  /**
+   * Checks that the bucketToTempQueueMap for a partitioned region
+   * that holds events for buckets that are not available locally, is empty.
+   */
+  public static void validateEmptyBucketToTempQueueMap(String senderId) {
+    GatewaySender sender = getGatewaySender(senderId);
+
+    int size = 0;
+    Set queues = ((AbstractGatewaySender) sender).getQueues();
```
DonalEvans, line 3462: This should be `Set<RegionQueue>` to avoid compiler warnings.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/WANTestBase.java
```diff
@@ -3599,21 +3444,42 @@ public static Long getNumberOfEntriesInVM(final String senderId) {
...
+    for (Object queue : queues) {
+      PartitionedRegion region =
+          (PartitionedRegion) ((ConcurrentParallelGatewaySenderQueue) queue).getRegion();
+      int buckets = region.getTotalNumberOfBuckets();
+      for (int bucket = 0; bucket < buckets; bucket++) {
+        BlockingQueue newQueue =
```
DonalEvans, line 3468: This should be `BlockingQueue<GatewaySenderEventImpl>` to avoid compiler warnings.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/WANTestBase.java
```diff
@@ -3599,21 +3444,42 @@ public static Long getNumberOfEntriesInVM(final String senderId) {
...
+            ((ConcurrentParallelGatewaySenderQueue) queue).getBucketTmpQueue((int) bucket);
```
DonalEvans, line 3469: The cast to `int` here is unnecessary.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/parallel/ParallelWANPropagationDUnitTest.java
```diff
@@ -1131,6 +1141,111 @@ public void testParallelPropagationWithFilter_AfterAck() throws Exception {
...
+   * Test that, when a parallel gateway sender is added to a partitioned region through attributes
+   * mutator, transaction events are not sent to all region members but only to those who are
+   * hosting the bucket for the event and thus, events are not stored in the bucketToTempQueueMap
+   * member of the ParallelGatewaySenderQueue.
+   * Redundancy = 1 in the partitioned region.
+   *
+   */
+  @Test
+  public void testParallelPropagationTxNotificationsNotSentToAllRegionMembersWhenAddingParallelGatewaySenderThroughAttributesMutator()
+      throws Exception {
```
DonalEvans, line 1154: No exception is thrown from this method, so this can be removed.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/parallel/ParallelWANPropagationDUnitTest.java
```diff
@@ -1131,6 +1141,111 @@ public void testParallelPropagationWithFilter_AfterAck() throws Exception {
...
+    Integer lnPort = (Integer) vm0.invoke(() -> WANTestBase.createFirstLocatorWithDSId(1));
+    Integer nyPort = (Integer) vm1.invoke(() -> WANTestBase.createFirstRemoteLocator(2, lnPort));
```
DonalEvans, line 1156: These casts to `Integer` are redundant and can be removed.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/parallel/ParallelWANPropagationDUnitTest.java
```diff
@@ -1131,6 +1141,111 @@ public void testParallelPropagationWithFilter_AfterAck() throws Exception {
...
+   * Test that, when a parallel gateway sender is added to a partitioned region through attributes
+   * mutator, transaction events are not sent to all region members but only to those who are
+   * hosting the bucket for the event and thus, events are not stored in the bucketToTempQueueMap
+   * member of the ParallelGatewaySenderQueue.
+   * No redundancy in the partitioned region.
+   *
+   */
+  @Test
+  public void testParallelPropagationTxNotificationsNotSentToAllRegionMembersWhenAddingParallelGatewaySenderThroughAttributesMutatorNoRedundancy()
+      throws Exception {
```
DonalEvans, line 1211: No exception is thrown from this method, so this can be removed.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/parallel/ParallelWANPropagationDUnitTest.java
```diff
@@ -1131,6 +1141,111 @@ public void testParallelPropagationWithFilter_AfterAck() throws Exception {
...
+    Integer lnPort = (Integer) vm0.invoke(() -> WANTestBase.createFirstLocatorWithDSId(1));
+    Integer nyPort = (Integer) vm1.invoke(() -> WANTestBase.createFirstRemoteLocator(2, lnPort));
```
DonalEvans, line 1213: These casts to `Integer` are redundant and can be removed.

albertogpz:
> A few small changes to clean up the code a bit. I also notice that there are several DUnit test failures associated with this change (more than we would expect just due to the known flakiness of some WAN DUnit tests) so I would prefer not to merge this change until they have been diagnosed and remedied.
> 
> Also, while it's not strictly within the scope of this PR, I notice that in `WANTestBase.createSender()` line 1747, the call to `gateway.create(dsName, remoteDsId);` comes before the call to `gateway.setBatchTimeInterval(batchTimeInterval);` which means the second call has no effect on the batch time interval. If fixing this has no effect on existing tests, then it would be okay to include it as part of this PR, I think. If fixing it causes tests to start failing, a separate JIRA ticket should be created to address the changes.

@DonalEvans Thanks a lot for your comments. You have an incredible eye to find every error and identify items to improve.
- I have added a fix for the distributed tests that were failing.
- I have incorporated all the fixes you have suggested in your comments.
- I have also tackled the issue you mentioned regarding `WANTestBase.createSender()` when specifying a `batchTimeInterval`. The way to address it has been to remove the parameter from the method as it had no effect anyway. I think it was me the one who introduced it in the first place when I implemented the "GEODE-7971" and was trying to fix some flaky tests. But obviously, that parameter did not have anything to do to the fix of the test cases.

mkevo (approved):
Thanks @albertogpz! It looks good to me too.

PR #5483, opened 2020-08-27
GEODE-8462: Make geode server startup fail if geode redis server has a port conflict

jhutchison:

### For all changes:

### Note:

### DIFF
--- a/geode-redis/src/main/java/org/apache/geode/redis/internal/netty/NettyRedisServer.java
+++ b/geode-redis/src/main/java/org/apache/geode/redis/internal/netty/NettyRedisServer.java
@@ -101,11 +101,16 @@ public class NettyRedisServer {
     if (port < RANDOM_PORT_INDICATOR) {
       throw new IllegalArgumentException("Redis port cannot be less than 0");
     }
-    this.bindAddress = getBindAddress(requestedAddress);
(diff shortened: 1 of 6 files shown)


geode-redis/src/main/java/org/apache/geode/redis/internal/netty/NettyRedisServer.java
```diff
@@ -99,7 +99,7 @@ public NettyRedisServer(Supplier<DistributionConfig> configSupplier,
       throw new IllegalArgumentException("Redis port cannot be less than 0");
     }
     this.bindAddress = getBindAddress(requestedAddress);
-    selectorGroup = createEventLoopGroup("Selector", false, 1);
+    selectorGroup = createEventLoopGroup("Selector", true, 1);
```
dschneider-pivotal, line 102: Was this non-daemon thread still hanging around and that caused problems? It might be okay for this to be daemon threads (that allows the JVM to exit without shutting down the redis server) but we should probably still do some cleanup of these groups before we throw the ManagementException. We should try calling the "stop" method before throwing the ManagementException

sabbeyPivotal: Good call, Darrel! Just updated it.  We decided to leave the selectorGroup as a daemon thread.

PR #5485, opened 2020-08-27
GEODE-7864: Always true check removed.

nabarunnag:
* removed the check for id less than membershipViewIdAcknowledged
	* check is already done the beginning of method

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/distributed/internal/ClusterDistributionManager.java
+++ b/geode-core/src/main/java/org/apache/geode/distributed/internal/ClusterDistributionManager.java
@@ -1387,7 +1387,7 @@ public class ClusterDistributionManager implements DistributionManager {
       return;
     }
     synchronized (membershipViewIdGuard) {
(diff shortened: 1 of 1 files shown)


geode-core/src/main/java/org/apache/geode/distributed/internal/ClusterDistributionManager.java
```diff
@@ -1387,7 +1387,7 @@ void waitForViewInstallation(long id) throws InterruptedException {
       return;
     }
     synchronized (membershipViewIdGuard) {
-      while (membershipViewIdAcknowledged < id && !stopper.isCancelInProgress()) {
+      while (!stopper.isCancelInProgress()) {
```
DonalEvans, line 1390: I think that this LGTM alert is incorrect, since it fails to take into account the value of `membershipViewIdAcknowledged` being modified by another thread, which seems to be what this method expects to happen. This false positive alert has been suppressed in this PR: https://github.com/apache/geode/pull/5473 to prevent it from showing up in future.

PR #5486, opened 2020-08-28
DRAFT: Do not store dropped events in stopped primary gateway sender when po…

albertogpz:
…ssible

### For all changes:

### Note:

This Draft PR tries to provide an alternative solution to the ones offered in RFC https://cwiki.apache.org/confluence/display/GEODE/Avoid+the+queuing+of+dropped+events+by+the+primary+gateway+sender+when+the+gateway+sender+is+stopped.

The high level idea consists of instead of trying not to store dropped events in tmpDroppedEvents to later send batch removal messages when the primary gateway sender is not started. Instead try to send the batch removal message when the event to be dropped is received. That way, when the sender is stopped for a long time and there are events coming, the memory of the AbstractGatewaySender will not grow with entries in the tmpDroppedEvents member.

In order to send the batch removal message directly, the eventProcessor for the AbstractGatewaySender must have been created. If it is not yet created because the sender was created with manual start set to true, while receiving events to be dropped, they will be stored in tmpDroppedEvents as there is no other choice. Nevertheless, in order to consume less memory, the event stored will be a simplified event containing only the necessary information to handle it.

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySender.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySender.java
@@ -1034,10 +1034,7 @@ public abstract class AbstractGatewaySender implements InternalGatewaySender, Di
       // If this gateway is not running, return
       if (!isRunning()) {
         if (this.isPrimary()) {
-          tmpDroppedEvents.add(clonedEvent);
-          if (isDebugEnabled) {
-            logger.debug("add to tmpDroppedEvents for evnet {}", clonedEvent);
-          }
+          dropEvent(clonedEvent);
         }
         if (isDebugEnabled) {
           logger.debug("Returning back without putting into the gateway sender queue:" + event);
@@ -1118,6 +1115,24 @@ public abstract class AbstractGatewaySender implements InternalGatewaySender, Di
     }
   }
 
+  private void dropEvent(EntryEventImpl event) {
+    final boolean isDebugEnabled = logger.isDebugEnabled();
+    if (this.eventProcessor != null) {
+      this.eventProcessor.registerEventDroppedInPrimaryQueue(event);
+    } else {
+      // Add empty event so that in case the event stays for long in
+      // tmpDroppedEvents it takes as little space as possible.
+      // No need to have all the contents of the event for a dropped one.
+      EntryEventImpl emptyEvent = new EntryEventImpl(event.getKey(), false);
+      emptyEvent.setRegion(event.getRegion());
+      emptyEvent.setTailKey(event.getTailKey());
+      tmpDroppedEvents.add(emptyEvent);
+      if (isDebugEnabled) {
+        logger.debug("added to tmpDroppedEvents event: {}", event);
+      }
+    }
+  }
+
   @VisibleForTesting
   int getTmpDroppedEventSize() {
     return tmpDroppedEvents.size();
@@ -1138,7 +1153,7 @@ public abstract class AbstractGatewaySender implements InternalGatewaySender, Di
   public void enqueueTempEvents() {
     if (this.eventProcessor != null) {// Fix for defect #47308
       // process tmpDroppedEvents
-      EntryEventImpl droppedEvent = null;
+      EntryEventImpl droppedEvent;
       while ((droppedEvent = tmpDroppedEvents.poll()) != null) {
         this.eventProcessor.registerEventDroppedInPrimaryQueue(droppedEvent);
       }
--- a/geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/WANTestBase.java
+++ b/geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/WANTestBase.java
@@ -1715,6 +1715,20 @@ public class WANTestBase extends DistributedTestCase {
     }
   }
 
+  public static void stopSenderInVMsAsync(String senderId, VM... vms) {
+    List<AsyncInvocation> tasks = new LinkedList<>();
+    for (VM vm : vms) {
+      tasks.add(vm.invokeAsync(() -> stopSender(senderId)));
+    }
+    for (AsyncInvocation invocation : tasks) {
+      try {
+        invocation.await();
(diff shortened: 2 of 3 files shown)


jinmeiliao (commented):
LGTM

albertogpz:
> LGTM

Thanks for the review!

geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySender.java
```diff
@@ -1034,10 +1034,7 @@ public void distribute(EnumListenerEvent operation, EntryEventImpl event,
       // If this gateway is not running, return
       if (!isRunning()) {
         if (this.isPrimary()) {
-          tmpDroppedEvents.add(clonedEvent);
-          if (isDebugEnabled) {
-            logger.debug("add to tmpDroppedEvents for evnet {}", clonedEvent);
-          }
+          dropEvent(clonedEvent);
```
gesterzhou, line 1037: this should be recordDroppedEvent or similar name.

albertogpz: I agree. Thanks

geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySender.java
```diff
@@ -1118,6 +1115,24 @@ public void distribute(EnumListenerEvent operation, EntryEventImpl event,
     }
   }
 
+  private void dropEvent(EntryEventImpl event) {
+    final boolean isDebugEnabled = logger.isDebugEnabled();
+    if (this.eventProcessor != null) {
+      this.eventProcessor.registerEventDroppedInPrimaryQueue(event);
+    } else {
+      // Add empty event so that in case the event stays for long in
```
gesterzhou, line 1123: the event is already a cloned. No need to clone again here. 

The if-else will reduce some chance to save into tmpDroppedEvents. But if the sender is purposely shutdown, the events will still be added into tmpDroppedEvents. This is your original problem. I don't think your current version will resolve this problem. Can you add a test case to prove?

albertogpz: The newly cloned event does not contain the same informacion as the cloned event. Just the minimal needed to drain the queue of the secondary. This is just an optimization so, in case events need to be stored in tmpDroppedEvents, they take less space (for example, the value is not stored). With it, the memory for the original clone could be freed by the garbage collector.

With regards to the if-else you mention, when the gateway sender is started, an instance of an `AbstractGatewaySenderEventProcessor` is created and stored in the `eventProcessor member`. If the sender is stopped, the eventProcessor is not deleted. Therefore, events will not be stored in tmpDroppedEvents but will be sent to the other gateway senders by means of the call to `eventProcessor.registerEventDroppedInPrimaryQueue(event)`.
The only case in which events could be stored in tmpDroppedEvents is when the sender is created with `manualStart` set to true while the sender is started.

In the last two test cases added `stoppedSenderShouldNotAddEventsToTmpDroppedEventsButStillDrainQueuesWhenStarted` (line 581) and `stoppedPrimarySenderShouldNotAddEventsToTmpDroppedEventsButStillDrainQueuesWhenStarted` (line 655) you can see that events are not added to tmpDroppedEvents and prove that the solution solves the problem.

gesterzhou: I see. That makes sense. But please don't clone a new event again. The cloneEvent is created the same way.

geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySender.java
```diff
@@ -1138,7 +1153,7 @@ int getTmpDroppedEventSize() {
   public void enqueueTempEvents() {
     if (this.eventProcessor != null) {// Fix for defect #47308
       // process tmpDroppedEvents
-      EntryEventImpl droppedEvent = null;
+      EntryEventImpl droppedEvent;
```
gesterzhou, line 1156: this code change is unnecessary

geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySender.java
gesterzhou: I reviewed the my change history, I was using 
 if (this.eventProcessor != null) {
      this.eventProcessor.registerEventDroppedInPrimaryQueue(event);
    } 
in version one. 

Then I found a bug that a newly started sender has a race to miss some event. So I change to use tmpDroppedEvents. However, we need to keep both.

albertogpz: I am ok with not cloning again the event. Anyway, if you look at the cloned event it would contain the newValue while the newly cloned event will not. If the value is big, tmpDroppedEvents will take much more memory.

Apart from the above, do you agree with this proposal to be implemented in a pull request?

geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySender.java
albertogpz: The original statement is redundant (as pointed by the IDE) as the variable will be reassigned in the next statement (while) 
anyway.

geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySender.java
gesterzhou: yes, I think it's good.

albertogpz: Thanks. I have written a JIRA https://issues.apache.org/jira/browse/GEODE-8491 and here is the PR: https://github.com/apache/geode/pull/5509

gesterzhou:
This is fixed in GEODE-8491

PR #5488, opened 2020-08-31
GEODE-8333: Change Redis adapter threading model - fixes pubsub issues

jdeppe-pivotal:
- Do not use Netty threads for the entire request lifecycle. Each
  instance of `ExecutionHandlerContext` (essentially each client
  connection) uses a command queue which is actioned by a single thread
  taken from the Geode 'waiting pool'.
- Every `SUBSCRIBE`ed client is moved to a separate `EventGroupLoop` so that
  `PUBLISH`ed messages are not sent back using the 'normal' Worker
  `EventLoopGroup`. This avoids a deadlock issue where `PUBLISH`ed messages
  need to be sent using the same thread that the `PUBLISH` response needs
  to happen on.
- Fix issues with PubSub where switching the `EventLoopGroup` may fail
  (because a client has already closed the connection) thus resulting in
  hanging `PUBLISH`ers waiting for a `Subscription` to be marked as
  `readyForPublish`.
- Consolidate various `MockSubscriber` classes

### For all changes:

### Note:


geode-redis/src/main/java/org/apache/geode/redis/internal/netty/ExecutionHandlerContext.java
```diff
@@ -61,17 +64,22 @@
...
   private final ByteBufAllocator byteBufAllocator;
   private final byte[] authPassword;
   private final Supplier<Boolean> allowUnsupportedSupplier;
   private final Runnable shutdownInvoker;
   private final RedisStats redisStats;
+  private final EventLoopGroup subscriberGroup;
+  private final int MAX_QUEUED_COMMANDS =
+      Integer.getInteger("geode.redis.commandQueueSize", 1000);
+  private final LinkedBlockingQueue<Command> commandQueue =
+      new LinkedBlockingQueue<>(100);
```
dschneider-pivotal, line 82: Instead of "100" should this be "MAX_QUEUED_COMMANDS"?

geode-redis/src/main/java/org/apache/geode/redis/internal/netty/ExecutionHandlerContext.java
```diff
@@ -139,6 +164,33 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) {
     }
   }
 
+  public EventLoopGroup getSubscriberGroup() {
+    return subscriberGroup;
+  }
+
+  public synchronized void changeChannelEventLoopGroup(EventLoopGroup newGroup,
```
dschneider-pivotal, line 171: I'd like to discuss what this should synchronize on. The method syncs on this (ExecutionHandlerContext) but the future it creates syncs on the channel. I think it would be better to sync on "channel" on this method instead of "this"

geode-redis/src/main/java/org/apache/geode/redis/internal/netty/ExecutionHandlerContext.java
```diff
@@ -247,25 +306,16 @@ private RedisResponse handleUnAuthenticatedCommand(Command command) {
...
+  private void logResponse(RedisResponse response, String extraMessage, Throwable cause) {
     if (logger.isDebugEnabled() && response != null) {
       ByteBuf buf = response.encode(new UnpooledByteBufAllocator(false));
-      logger.debug("Redis command returned: {}",
-          Command.getHexEncodedString(buf.array(), buf.readableBytes()));
+      if (cause == null) {
+        logger.debug("Redis command returned: {} - {}",
+            Command.getHexEncodedString(buf.array(), buf.readableBytes()), extraMessage);
+      } else {
+        logger.warn("Redis command FAILED to return: {} - {}",
```
dschneider-pivotal, line 316: Does this warning message make sense? It might but I thought it would just say it FAILED I'm not sure about "to return"

geode-redis/src/main/java/org/apache/geode/redis/internal/netty/ExecutionHandlerContext.java
```diff
@@ -192,43 +244,50 @@ public void channelInactive(ChannelHandlerContext ctx) {
     if (logger.isDebugEnabled()) {
       logger.debug("GeodeRedisServer-Connection closing with " + ctx.channel().remoteAddress());
     }
+    commandQueue.offer(TERMINATE_COMMAND);
```
dschneider-pivotal, line 247: I think we should review the commandQueue code and see if we might continue to process commands in the queue after this happens. For example this code adds TERMINATE_COMMAND to the end of the queue so do we end up still processing any commands already queued? We should also check that the "quit" command does not allow commands queued after it to still be processed on the server

geode-redis/src/main/java/org/apache/geode/redis/internal/netty/ExecutionHandlerContext.java
jdeppe-pivotal: Switched method synchronization to synchronize on `channel`

geode-redis/src/main/java/org/apache/geode/redis/internal/netty/ExecutionHandlerContext.java
jdeppe-pivotal: Changed to `debug`

geode-redis/src/main/java/org/apache/geode/redis/internal/netty/ExecutionHandlerContext.java
jdeppe-pivotal: This will be addressed in a separate PR.

PR #5490, opened 2020-08-31
GEODE-8467: server fails to notify of a ForcedDisconnect and fails to tear down the cache

bschuchardt:
Catch exceptions that occur during XML generation and disable auto
reconnect.

Ensure that the DisconnectThread is launched by placing it in a
"finally" block.

@kamilla1201 @Bill

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/GemFireCacheImpl.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/GemFireCacheImpl.java
@@ -57,6 +57,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.OutputStreamWriter;
+import java.io.PrintWriter;
 import java.io.Reader;
 import java.io.StringBufferInputStream;
 import java.io.StringWriter;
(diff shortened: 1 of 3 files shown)


geode-core/src/main/java/org/apache/geode/internal/cache/GemFireCacheImpl.java
```diff
@@ -1177,34 +1178,53 @@ public MeterRegistry getMeterRegistry() {
...
+        getCacheConfig().setCacheServerCreation(list);
+        logger.info("cache server configuration saved");
       }
-      getCacheConfig().setCacheServerCreation(list);
-      logger.info("CacheServer configuration saved");
+    } catch (Throwable throwable) {
+      logger.info("Saving of cache configuration for auto-reconnect has failed.  "
+          + "Auto-reconnect will be disabled since the cache cannot be rebuilt.", throwable);
+      getInternalDistributedSystem().getConfig().setDisableAutoReconnect(true);
+
```
Bill, line 1227: I guess it's ok to not rethrow the `Throwable` here. `saveConfg()` is called from two places in `GMSMembership`:

1. `ManagerImpl.forceDisconnect()`
2. `requestMemberRemoval()`

In both those cases it looks like `setDisableAutoReconnect(true)` will set us on a permanent shutdown path. Since we've logged the `Throwable` here we've fulfilled our obligation to inform the operator.

PR #5500, opened 2020-09-09
GEODE-8486: record TransactionDataRebalancedException if tx put failed

pivotal-eshu:
This is used to handle a retry of transactional put.


geode-core/src/main/java/org/apache/geode/internal/cache/TXState.java
```diff
@@ -146,7 +146,9 @@
   /** keeps track of events, so as not to re-apply events */
   protected Set<EventID> seenEvents = new HashSet<EventID>();
   /** keeps track of results of txPutEntry */
-  private Map<EventID, Boolean> seenResults = new HashMap<EventID, Boolean>();
+  private Map<EventID, Boolean> seenResults = new HashMap<>();
+  /** keeps track of TransactionDataRebalancedException during txPutEntry */
+  private Map<EventID, TransactionDataRebalancedException> failedExceptions = new HashMap<>();
```
DonalEvans, line 151: This might be better named "failedEvents" or "eventExceptions."

DonalEvans (approved):

PR #5501, opened 2020-09-09
GEODE-8339: Redis Rename hangs when servers are killed and revived

ringles:
The Redis RENAME command could hang during a rebalance, if the old key was stored in a bucket on one server, and the new key was in a different bucket on a separate server. Rename would read-lock the buckets, but rebalance would wait for a write lock, and the Rename region put/destroy would then wait on the write lock.

Now Rename passes a callback argument that indicates it has already locked the primary, and will not attempt to lock the primary again when doing the put/destroy.

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/LocalRegion.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/LocalRegion.java
@@ -5718,6 +5718,7 @@ public class LocalRegion extends AbstractRegion implements LoaderHelperFactory,
   @Override
   public long basicPutPart2(EntryEventImpl event, RegionEntry entry, boolean isInitialized,
       long lastModified, boolean clearConflict) {
+
     final boolean isNewKey = event.getOperation().isCreate();
 
     // Invoke callbacks only if we are not creating a tombstone
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/LocalRegion.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/LocalRegion.java
@@ -5718,7 +5718,6 @@ public class LocalRegion extends AbstractRegion implements LoaderHelperFactory,
   @Override
   public long basicPutPart2(EntryEventImpl event, RegionEntry entry, boolean isInitialized,
       long lastModified, boolean clearConflict) {
-
     final boolean isNewKey = event.getOperation().isCreate();
 
     // Invoke callbacks only if we are not creating a tombstone
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
@@ -611,8 +611,7 @@ public class PartitionedRegion extends LocalRegion
     return redundancyTracker;
   }
 
-  public boolean computeWithPrimaryLocked(Object key, Runnable r)
-      throws PrimaryBucketLockException {
+  public void computeWithPrimaryLocked(Object key, Runnable r) throws PrimaryBucketLockException {
     int bucketId = PartitionedRegionHelper.getHashKey(this, null, key, null, null);
 
     BucketRegion br;
@@ -623,9 +622,7 @@ public class PartitionedRegion extends LocalRegion
     }
 
     try {
-      if (!br.doLockForPrimary(false)) {
-        return false;
-      }
+      br.doLockForPrimary(false);
     } catch (PrimaryBucketException e) {
       throw new PrimaryBucketLockException("retry since primary lock failed: " + e);
     }
@@ -635,7 +632,6 @@ public class PartitionedRegion extends LocalRegion
     } finally {
       br.doUnlockForPrimary();
     }
-    return true;
   }
 
 
@@ -3399,10 +3395,11 @@ public class PartitionedRegion extends LocalRegion
    * @throws PrimaryBucketException if the remote bucket was not the primary
    * @throws ForceReattemptException if the peer is no longer available
    */
-
   public boolean putRemotely(final DistributedMember recipient, final EntryEventImpl event,
       boolean ifNew, boolean ifOld, Object expectedOldValue, boolean requireOldValue)
       throws PrimaryBucketException, ForceReattemptException {
+    // boolean forceAck = basicGetWriter() != null
+    // || getDistributionAdvisor().adviseNetWrite().size() > 0;
     long eventTime = event.getEventTime(0L);
     PutMessage.PutResponse response = (PutMessage.PutResponse) PutMessage.send(recipient, this,
         event, eventTime, ifNew, ifOld, expectedOldValue, requireOldValue);
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
@@ -611,7 +611,8 @@ public class PartitionedRegion extends LocalRegion
     return redundancyTracker;
   }
 
-  public void computeWithPrimaryLocked(Object key, Runnable r) throws PrimaryBucketLockException {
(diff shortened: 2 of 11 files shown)


geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
```diff
@@ -611,7 +611,8 @@ PartitionedRegionRedundancyTracker getRedundancyTracker() {
     return redundancyTracker;
   }
 
-  public void computeWithPrimaryLocked(Object key, Runnable r) throws PrimaryBucketLockException {
+  public boolean computeWithPrimaryLocked(Object key, Runnable r, boolean tryLock)
```
dschneider-pivotal, line 614: My understanding is that we no longer use this new "tryLock" parameter. So should we remove it from this method so that this PR does not change this code?

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/key/RedisKeyCommandsFunctionInvoker.java
```diff
@@ -74,18 +73,18 @@ public String type(ByteArrayWrapper key) {
 
 
   @Override
-  @SuppressWarnings("unchecked")
   public boolean rename(ByteArrayWrapper oldKey, ByteArrayWrapper newKey) {
+    while (true) {
```
dschneider-pivotal, line 77: I think the "while (true)" loop should be removed. I think it is a remnant of the tryLock code. As it is now I don't see that it would ever loop.

dschneider-pivotal (changes requested):

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
sabbeyPivotal: You are correct, Darrel, we no longer need it.  Jens and I talked about it a bit this. morning and thought it might be beneficial to have the option in the method, but since it's not needed for this PR it probably makes sense to remove it.

jdeppe-pivotal (approved):
+1  Hopefully if we discover other uses for this pattern we can provide a better abstraction for general use.

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/key/RedisKeyCommandsFunctionInvoker.java
sabbeyPivotal: Truth, Darrel. Just removed it!

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
```diff
@@ -611,7 +611,8 @@ PartitionedRegionRedundancyTracker getRedundancyTracker() {
     return redundancyTracker;
   }
 
-  public void computeWithPrimaryLocked(Object key, Runnable r) throws PrimaryBucketLockException {
+  public boolean computeWithPrimaryLocked(Object key, Runnable r)
```
dschneider-pivotal, line 614: this should no longer return boolean (since we got rid of the tryLock option)

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
```diff
@@ -622,7 +623,9 @@ public void computeWithPrimaryLocked(Object key, Runnable r) throws PrimaryBucke
     }
 
     try {
-      br.doLockForPrimary(false);
+      if (!br.doLockForPrimary(false)) {
```
dschneider-pivotal, line 626: remove the "if" and just call br.doLockForPrimary(false);

geode-core/src/main/java/org/apache/geode/internal/cache/PrimaryBucketLockException.java
```diff
@@ -19,7 +19,7 @@
 import org.apache.geode.GemFireException;
 
 /**
- * Thrown by {@link PartitionedRegion#computeWithPrimaryLocked(Object, Runnable)}
+ * Thrown by {@link PartitionedRegion#computeWithPrimaryLocked(Object, Runnable, boolean)}
```
dschneider-pivotal, line 22: this javadoc needs to be updated for the signature with no boolean

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
```diff
@@ -3395,11 +3399,10 @@ private InternalDistributedMember getNodeForBucketReadOrLoad(int bucketId) {
    * @throws PrimaryBucketException if the remote bucket was not the primary
    * @throws ForceReattemptException if the peer is no longer available
    */
+
```
dschneider-pivotal, line 3402: take out this new blank line

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
```diff
@@ -3395,11 +3399,10 @@ private InternalDistributedMember getNodeForBucketReadOrLoad(int bucketId) {
...
   public boolean putRemotely(final DistributedMember recipient, final EntryEventImpl event,
       boolean ifNew, boolean ifOld, Object expectedOldValue, boolean requireOldValue)
       throws PrimaryBucketException, ForceReattemptException {
-    // boolean forceAck = basicGetWriter() != null
-    // || getDistributionAdvisor().adviseNetWrite().size() > 0;
```
dschneider-pivotal, line 3402: restore this comment; basically get this PR to the point that it does not modify PartitionedRegion.java

geode-core/src/main/java/org/apache/geode/internal/cache/LocalRegion.java
```diff
@@ -5718,7 +5718,6 @@ RegionEntry basicPutEntry(final EntryEventImpl event, final long lastModified)
   @Override
   public long basicPutPart2(EntryEventImpl event, RegionEntry entry, boolean isInitialized,
       long lastModified, boolean clearConflict) {
-
```
dschneider-pivotal, line 5721: put this blank line back so that this PR does not modify LocalRegion.java

dschneider-pivotal (changes requested):

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
sabbeyPivotal: Good call, just updated it!

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
sabbeyPivotal: just updated it!

geode-core/src/main/java/org/apache/geode/internal/cache/PrimaryBucketLockException.java
sabbeyPivotal: Truth. Thanks, Darrel!

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
sabbeyPivotal: updated.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
sabbeyPivotal: updated

geode-core/src/main/java/org/apache/geode/internal/cache/LocalRegion.java
sabbeyPivotal: updated.

dschneider-pivotal (approved):

PR #5504, opened 2020-09-10
GEODE-8393: Add sequence number to RedisString to support delta and APPEND

jdeppe-pivotal:
- There seems to be an intermittent problem when Deltas arrive and a GII
  is in progress. This fix introduces a sequence number that is used for
  APPEND operations only. Since APPEND is NOT idempotent, duplicate
  deltas can be a problem and will cause data issues if applied when they
  should not be.

Co-Authored-By: Darrel Schneider <darrel@vmware.com>


sabbeyPivotal (approved):
Seems like a good solution!

dschneider-pivotal (approved):

PR #5507, opened 2020-09-11
GEODE-8489: "Select *" query by DistributedSystemMBean should not hon…

jinmeiliao:
…or jackson json annotations

* null value fields are serialized as well
* move QueryResultFormatterTest to core

### DIFF
--- a/geode-core/src/integrationTest/java/org/apache/geode/management/internal/beans/DistributedSystemMBeanIntegrationTest.java
+++ b/geode-core/src/integrationTest/java/org/apache/geode/management/internal/beans/DistributedSystemMBeanIntegrationTest.java
@@ -113,7 +113,7 @@ public class DistributedSystemMBeanIntegrationTest {
   // gfsh doesn't attempt to format the date objects as of now, and it respect the json annotations
   // when listing out the headers
   @Test
-  public void queryUsingGfshDoesNotFormatDate() throws Exception {
+  public void queryAllUsingGfshDoesNotFormatDate() throws Exception {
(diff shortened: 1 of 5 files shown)


geode-core/src/integrationTest/java/org/apache/geode/management/internal/beans/DistributedSystemMBeanIntegrationTest.java
```diff
@@ -30,14 +31,16 @@
...
+import org.apache.geode.management.model.Employee;
 import org.apache.geode.test.junit.assertions.TabularResultModelAssert;
 import org.apache.geode.test.junit.rules.GfshCommandRule;
 import org.apache.geode.test.junit.rules.MBeanServerConnectionRule;
 import org.apache.geode.test.junit.rules.ServerStarterRule;
 
 public class DistributedSystemMBeanIntegrationTest {
 
-  public static final String SELECT = "select * from /testRegion r where r.id=1";
+  public static final String SELECT_ALL = "select * from /testRegion r where r.id=1";
```
agingade, line 42: Can we add a query that returns 0 results. And verify the results are shown as expected.

jinmeiliao: the test is to verify the results are formatted correctly. If returning 0 results, no formats needed, that's not the purpose of the tests. I am sure there are other tests that tests 0 results.

jinmeiliao: done

agingade (approved):

PR #5509, opened 2020-09-14
GEODE-8491: Do not store dropped events in stopped primary gateway se…

albertogpz:
…nder when possible

Instead of storing dropped events in tmpDroppedEvents to later send batch
removal messages when the primary gateway sender is not started, try to send
the batch removal message when the event to be dropped is received.
That way, when the sender is stopped for a long time and there are events
coming, the memory of the AbstractGatewaySender will not grow with entries
in the tmpDroppedEvents member.

In order to send the batch removal message directly, the eventProcessor for
the AbstractGatewaySender must have been created.
If it is not yet created because the sender was created with manual start
set to true, while receiving events to be dropped, they will be stored in
tmpDroppedEvents as there is no other choice. Nevertheless,
in order to consume less memory, the event stored could be a simplified event
containing only the necessary information to handle it.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySender.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySender.java
@@ -1116,12 +1116,11 @@ public abstract class AbstractGatewaySender implements InternalGatewaySender, Di
   }
 
   private void recordDroppedEvent(EntryEventImpl event) {
-    final boolean isDebugEnabled = logger.isDebugEnabled();
     if (this.eventProcessor != null) {
       this.eventProcessor.registerEventDroppedInPrimaryQueue(event);
     } else {
       tmpDroppedEvents.add(event);
-      if (isDebugEnabled) {
+      if (logger.isDebugEnabled()) {
         logger.debug("added to tmpDroppedEvents event: {}", event);
       }
     }
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySender.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySender.java
@@ -1120,13 +1120,7 @@ public abstract class AbstractGatewaySender implements InternalGatewaySender, Di
     if (this.eventProcessor != null) {
       this.eventProcessor.registerEventDroppedInPrimaryQueue(event);
     } else {
-      // Add empty event so that in case the event stays for long in
-      // tmpDroppedEvents it takes as little space as possible.
-      // No need to have all the contents of the event for a dropped one.
-      EntryEventImpl emptyEvent = new EntryEventImpl(event.getKey(), false);
-      emptyEvent.setRegion(event.getRegion());
-      emptyEvent.setTailKey(event.getTailKey());
-      tmpDroppedEvents.add(emptyEvent);
+      tmpDroppedEvents.add(event);
       if (isDebugEnabled) {
         logger.debug("added to tmpDroppedEvents event: {}", event);
       }
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySender.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySender.java
@@ -1034,10 +1034,7 @@ public abstract class AbstractGatewaySender implements InternalGatewaySender, Di
       // If this gateway is not running, return
       if (!isRunning()) {
         if (this.isPrimary()) {
-          tmpDroppedEvents.add(clonedEvent);
-          if (isDebugEnabled) {
-            logger.debug("add to tmpDroppedEvents for evnet {}", clonedEvent);
-          }
+          recordDroppedEvent(clonedEvent);
         }
         if (isDebugEnabled) {
           logger.debug("Returning back without putting into the gateway sender queue:" + event);
@@ -1118,6 +1115,24 @@ public abstract class AbstractGatewaySender implements InternalGatewaySender, Di
     }
   }
 
+  private void recordDroppedEvent(EntryEventImpl event) {
+    final boolean isDebugEnabled = logger.isDebugEnabled();
+    if (this.eventProcessor != null) {
+      this.eventProcessor.registerEventDroppedInPrimaryQueue(event);
+    } else {
+      // Add empty event so that in case the event stays for long in
+      // tmpDroppedEvents it takes as little space as possible.
+      // No need to have all the contents of the event for a dropped one.
+      EntryEventImpl emptyEvent = new EntryEventImpl(event.getKey(), false);
+      emptyEvent.setRegion(event.getRegion());
+      emptyEvent.setTailKey(event.getTailKey());
+      tmpDroppedEvents.add(emptyEvent);
+      if (isDebugEnabled) {
+        logger.debug("added to tmpDroppedEvents event: {}", event);
+      }
+    }
+  }
+
   @VisibleForTesting
   int getTmpDroppedEventSize() {
     return tmpDroppedEvents.size();
@@ -1138,7 +1153,7 @@ public abstract class AbstractGatewaySender implements InternalGatewaySender, Di
   public void enqueueTempEvents() {
     if (this.eventProcessor != null) {// Fix for defect #47308
       // process tmpDroppedEvents
-      EntryEventImpl droppedEvent = null;
+      EntryEventImpl droppedEvent;
       while ((droppedEvent = tmpDroppedEvents.poll()) != null) {
         this.eventProcessor.registerEventDroppedInPrimaryQueue(droppedEvent);
       }
--- a/geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/WANTestBase.java
+++ b/geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/WANTestBase.java
@@ -1646,6 +1646,20 @@ public class WANTestBase extends DistributedTestCase {
     }
   }
 
+  public static void stopSenderInVMsAsync(String senderId, VM... vms) {
+    List<AsyncInvocation> tasks = new LinkedList<>();
+    for (VM vm : vms) {
+      tasks.add(vm.invokeAsync(() -> stopSender(senderId)));
+    }
+    for (AsyncInvocation invocation : tasks) {
+      try {
+        invocation.await();
+      } catch (InterruptedException e) {
+        fail("Stopping senders was interrupted");
+      }
+    }
+  }
+
   public static void stopSender(String senderId) {
     final IgnoredException exln = IgnoredException.addIgnoredException("Could not connect");
     IgnoredException exp =
--- a/geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/WANTestBase.java
+++ b/geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/WANTestBase.java
@@ -1647,7 +1647,7 @@ public class WANTestBase extends DistributedTestCase {
   }
 
   public static void stopSenderInVMsAsync(String senderId, VM... vms) {
-    List<AsyncInvocation> tasks = new LinkedList<>();
+    List<AsyncInvocation<Void>> tasks = new LinkedList<>();
     for (VM vm : vms) {
       tasks.add(vm.invokeAsync(() -> stopSender(senderId)));
     }
--- a/geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/parallel/ParallelWANPropagationLoopBackDUnitTest.java
+++ b/geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/parallel/ParallelWANPropagationLoopBackDUnitTest.java
@@ -18,6 +18,7 @@ import org.junit.Test;
 import org.junit.experimental.categories.Category;
 
 import org.apache.geode.internal.cache.wan.WANTestBase;
+import org.apache.geode.test.dunit.AsyncInvocation;
 import org.apache.geode.test.dunit.Wait;
 import org.apache.geode.test.junit.categories.WanTest;
 
@@ -370,8 +371,8 @@ public class ParallelWANPropagationLoopBackDUnitTest extends WANTestBase {
    */
   @Test
   public void unstartedSenderShouldNotAddReceivedEventsIntoTmpDropped() throws Exception {
-    Integer lnPort = (Integer) vm0.invoke(() -> WANTestBase.createFirstLocatorWithDSId(2));
-    Integer nyPort = (Integer) vm1.invoke(() -> WANTestBase.createFirstRemoteLocator(1, lnPort));
+    Integer lnPort = vm0.invoke(() -> WANTestBase.createFirstLocatorWithDSId(2));
+    Integer nyPort = vm1.invoke(() -> WANTestBase.createFirstRemoteLocator(1, lnPort));
 
     // create receiver on site-ln and site-ny
     createCacheInVMs(lnPort, vm2, vm4);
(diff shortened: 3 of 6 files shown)


geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySender.java
```diff
@@ -1118,6 +1115,24 @@ public void distribute(EnumListenerEvent operation, EntryEventImpl event,
...
 
+  private void recordDroppedEvent(EntryEventImpl event) {
+    final boolean isDebugEnabled = logger.isDebugEnabled();
+    if (this.eventProcessor != null) {
+      this.eventProcessor.registerEventDroppedInPrimaryQueue(event);
+    } else {
+      // Add empty event so that in case the event stays for long in
+      // tmpDroppedEvents it takes as little space as possible.
+      // No need to have all the contents of the event for a dropped one.
+      EntryEventImpl emptyEvent = new EntryEventImpl(event.getKey(), false);
```
gesterzhou, line 1126: Again, don't clone a new event. In the new code change, the "else" will only be executed in very very small time windows. 

You have proved that when the sender is shutdown, the eventProcessor will be kept. So enqueue will not go to the "else". We will never worry about the tmpDroppedEvents become too big.

albertogpz: ok. I will remove it.
Anyway, if the gateway sender is created with manualStart=true then the eventProcessor will not be yet instantiated. In that case, and until the gateway sender is started we could see dropped events put in tmpDroppedEvents. It cannot be guaranteed that it will be a small time window.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/WANTestBase.java
```diff
@@ -1646,6 +1646,20 @@ public static void resumeSender(String senderId) {
     }
   }
 
+  public static void stopSenderInVMsAsync(String senderId, VM... vms) {
+    List<AsyncInvocation> tasks = new LinkedList<>();
+    for (VM vm : vms) {
+      tasks.add(vm.invokeAsync(() -> stopSender(senderId)));
+    }
+    for (AsyncInvocation invocation : tasks) {
```
DonalEvans, line 1654: The IDE warnings here can be resolved by using `List<AsyncInvocation<Void>>`.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/parallel/ParallelWANPropagationLoopBackDUnitTest.java
```diff
@@ -370,69 +371,312 @@ public void testParallelPropagationLoopBack3SitesNtoNTopologyPutFromOneDS() {
...
+        isOffHeap()));
+    vm4.invoke(() -> WANTestBase.createPartitionedRegion(getTestMethodName() + "_PR", "ny", 1, 100,
+        isOffHeap()));
+
+    vm3.invoke(() -> WANTestBase.createPartitionedRegion(getTestMethodName() + "_PR", "ln", 1, 100,
+        isOffHeap()));
+    vm5.invoke(() -> WANTestBase.createPartitionedRegion(getTestMethodName() + "_PR", "ln", 1, 100,
+        isOffHeap()));
+
+    AsyncInvocation inv =
```
DonalEvans, line 471: The IDE warning here can be resolved by using `List<AsyncInvocation<Void>>`.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/parallel/ParallelWANPropagationLoopBackDUnitTest.java
```diff
@@ -370,69 +371,312 @@ public void testParallelPropagationLoopBack3SitesNtoNTopologyPutFromOneDS() {
...
```
DonalEvans, line 522: The IDE warning here can be resolved by using `List<AsyncInvocation<Void>>`.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/parallel/ParallelWANPropagationLoopBackDUnitTest.java
```diff
@@ -370,69 +371,312 @@ public void testParallelPropagationLoopBack3SitesNtoNTopologyPutFromOneDS() {
...
+        vm2.invokeAsync(() -> WANTestBase.doPuts(getTestMethodName() + "_PR", 1000));
+    startSenderInVMsAsync("ny", vm2, vm4);
+    inv.join();
```
DonalEvans, line 474: The `AsyncInvocation.join()` method is deprecated. It should be replaced with `AsyncInvocation.await()`.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/parallel/ParallelWANPropagationLoopBackDUnitTest.java
```diff
@@ -370,69 +371,312 @@ public void testParallelPropagationLoopBack3SitesNtoNTopologyPutFromOneDS() {
...
+   * NY site's sender's manual-start=false
+   * LN site's sender's manual-start=false
+   *
+   * put some events from LN and stop the sender in NY simultaneously
+   * Start the sender in NY.
+   * Make sure there are no events in tmpDroppedEvents and the queues are drained.
+   */
+  @Test
+  public void stoppedSenderShouldNotAddEventsToTmpDroppedEventsButStillDrainQueuesWhenStarted()
+      throws Exception {
```
DonalEvans, line 552: An exception is never thrown from this method, so this can be removed.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/parallel/ParallelWANPropagationLoopBackDUnitTest.java
```diff
@@ -370,69 +371,312 @@ public void testParallelPropagationLoopBack3SitesNtoNTopologyPutFromOneDS() {
...
+   * NY site's sender's manual-start=false
+   * LN site's sender's manual-start=false
+   *
+   * put some events from LN and stop one instance of the sender in NY simultaneously
+   * Start the stopped instance of the sender in NY.
+   * Make sure there are no events in tmpDroppedEvents and the queues are drained.
+   */
+  @Test
+  public void stoppedPrimarySenderShouldNotAddEventsToTmpDroppedEventsButStillDrainQueuesWhenStarted()
+      throws Exception {
```
DonalEvans, line 626: An exception is never thrown from this method, so this can be removed.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialWANPropagationsFeatureDUnitTest.java
```diff
@@ -297,4 +298,322 @@ public void testReplicatedSerialPropagationWithFilter_AfterAck() throws Exceptio
...
+   * Site-LN: dsid=2: senderId="ny": vm2, vm4
+   * Site-NY: dsid=1: senderId="ln": vm3, vm6
+   * NY site's sender's manual-start=true
+   *
+   * Make sure the events are sent from LN to NY and will not be added into tmpDroppedEvents
+   * while normal events put from NY site can still be added to tmpDroppedEvents
+   * Start the sender, make sure the events in tmpDroppedEvents are sent to LN finally
+   */
+  @Test
+  public void unstartedSenderShouldNotAddReceivedEventsIntoTmpDropped() throws Exception {
```
DonalEvans, line 313: An exception is never thrown from this method, so this "throws" can be removed.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialWANPropagationsFeatureDUnitTest.java
```diff
@@ -297,4 +298,322 @@ public void testReplicatedSerialPropagationWithFilter_AfterAck() throws Exceptio
...
+        isOffHeap()));
+    vm4.invoke(() -> WANTestBase.createPartitionedRegion(getTestMethodName() + "_PR", "ny", 1, 100,
+        isOffHeap()));
+
+    vm3.invoke(() -> WANTestBase.createPartitionedRegion(getTestMethodName() + "_PR", "ln", 1, 100,
+        isOffHeap()));
+    vm5.invoke(() -> WANTestBase.createPartitionedRegion(getTestMethodName() + "_PR", "ln", 1, 100,
+        isOffHeap()));
+
+    AsyncInvocation inv =
```
DonalEvans, line 407: The IDE warning here can be resolved by using `List<AsyncInvocation<Void>>`.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialWANPropagationsFeatureDUnitTest.java
```diff
@@ -297,4 +298,322 @@ public void testReplicatedSerialPropagationWithFilter_AfterAck() throws Exceptio
...
+        vm2.invokeAsync(() -> WANTestBase.doPuts(getTestMethodName() + "_PR", 1000));
+    startSenderInVMsAsync("ny", vm2, vm4);
+    inv.join();
```
DonalEvans, line 410: The `AsyncInvocation.join()` method is deprecated. It should be replaced with `AsyncInvocation.await()`.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialWANPropagationsFeatureDUnitTest.java
```diff
@@ -297,4 +298,322 @@ public void testReplicatedSerialPropagationWithFilter_AfterAck() throws Exceptio
...
+
+    vm3.invoke(() -> WANTestBase.createPartitionedRegion(getTestMethodName() + "_PR", "ln", 1, 100,
+        isOffHeap()));
+    vm5.invoke(() -> WANTestBase.createPartitionedRegion(getTestMethodName() + "_PR", "ln", 1, 100,
+        isOffHeap()));
+
+    AsyncInvocation inv =
+        vm2.invokeAsync(() -> WANTestBase.doPuts(getTestMethodName() + "_PR", 1000));
+    stopSenderInVMsAsync("ny", vm2, vm4);
+    inv.join();
```
DonalEvans, line 461: The `AsyncInvocation.join()` method is deprecated. It should be replaced with `AsyncInvocation.await()`.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialWANPropagationsFeatureDUnitTest.java
```diff
@@ -297,4 +298,322 @@ public void testReplicatedSerialPropagationWithFilter_AfterAck() throws Exceptio
...
+        isOffHeap()));
+    vm4.invoke(() -> WANTestBase.createPartitionedRegion(getTestMethodName() + "_PR", "ny", 1, 100,
+        isOffHeap()));
+
+    vm3.invoke(() -> WANTestBase.createPartitionedRegion(getTestMethodName() + "_PR", "ln", 1, 100,
+        isOffHeap()));
+    vm5.invoke(() -> WANTestBase.createPartitionedRegion(getTestMethodName() + "_PR", "ln", 1, 100,
+        isOffHeap()));
+
+    AsyncInvocation inv =
```
DonalEvans, line 458: The IDE warning here can be resolved by using `List<AsyncInvocation<Void>>`.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialWANPropagationsFeatureDUnitTest.java
```diff
@@ -297,4 +298,322 @@ public void testReplicatedSerialPropagationWithFilter_AfterAck() throws Exceptio
...
+   * NY site's sender's manual-start=false
+   * LN site's sender's manual-start=false
+   *
+   * put some events from LN and stop the sender in NY simultaneously
+   * Start the sender in NY.
+   * Make sure there are no events in tmpDroppedEvents and the queues are drained.
+   */
+  @Test
+  public void stoppedSenderShouldNotAddEventsToTmpDroppedEventsButStillDrainQueuesWhenStarted()
+      throws Exception {
```
DonalEvans, line 488: An exception is never thrown from this method, so this "throws" can be removed.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialWANPropagationsFeatureDUnitTest.java
```diff
@@ -297,4 +298,322 @@ public void testReplicatedSerialPropagationWithFilter_AfterAck() throws Exceptio
...
+   * NY site's sender's manual-start=false
+   * LN site's sender's manual-start=false
+   *
+   * put some events from LN and stop one instance of the sender in NY simultaneously
+   * Start the stopped instance of the sender in NY.
+   * Make sure there are no events in tmpDroppedEvents and the queues are drained.
+   */
+  @Test
+  public void stoppedPrimarySenderShouldNotAddEventsToTmpDroppedEventsButStillDrainQueuesWhenStarted()
+      throws Exception {
```
DonalEvans, line 562: An exception is never thrown from this method, so this "throws" can be removed.

geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySender.java
```diff
@@ -1118,6 +1115,18 @@ public void distribute(EnumListenerEvent operation, EntryEventImpl event,
     }
   }
 
+  private void recordDroppedEvent(EntryEventImpl event) {
+    final boolean isDebugEnabled = logger.isDebugEnabled();
```
DonalEvans, line 1119: This line could be moved to immediately before the `if (isDebugEnabled)` check so that it's only executed if necessary rather than unnecessarily in the case that `this.eventProcessor != null` evaluates to true.

DonalEvans (changes requested):
A few code cleanup requests, but otherwise looks good. An elegant solution.

albertogpz:
> A few code cleanup requests, but otherwise looks good. An elegant solution.

@Donal, thanks for the comments.

geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySender.java
```diff
@@ -1118,6 +1115,17 @@ public void distribute(EnumListenerEvent operation, EntryEventImpl event,
     }
   }
 
+  private void recordDroppedEvent(EntryEventImpl event) {
+    if (this.eventProcessor != null) {
+      this.eventProcessor.registerEventDroppedInPrimaryQueue(event);
```
boglesby, line 1120: In ConcurrentParallelGatewaySenderEventProcessor.registerEventDroppedInPrimaryQueue instead of calling sendQueueRemovalMesssageForDroppedEvent to send a ParallelQueueRemovalMessage inline, can you call ParallelGatewaySenderQueue.addRemovedEvent and let the BatchRemovalThread handle sending the message?

Something like:
```
...
int bucketId = PartitionedRegionHelper.getHashKey((EntryOperation) droppedEvent);
boolean isPrimary = prQ.getRegionAdvisor().getBucketAdvisor(bucketId).isPrimary();
if (isPrimary) {
  long shadowKey = droppedEvent.getTailKey();
  ParallelGatewaySenderQueue pgsq = (ParallelGatewaySenderQueue) cpgsq.getQueueByBucket(bucketId);
  pgsq.addRemovedEvent(prQ, bucketId, shadowKey);
  this.sender.getStatistics().incEventsDroppedDueToPrimarySenderNotRunning();
...
``
I didn't try this, but it seems like it should work.

I'm not sure if the same idea would work for SerialGatewaySenderEventProcessor.sendBatchDestroyOperationForDroppedEvent since it has a different implementation.

btw - sendQueueRemovalMesssageForDroppedEvent has an extra 's'.

albertogpz: @boglesby I have made the change you have proposed for the parallel gateway sender case. You can check it in the new commit. Nevertheless, this change it requires that the BatchRemoval thread is not shutdown when the gateway sender is stopped. Is there any problem with not stopping the BatchRemoval thread when the sender is stopped?

Regarding the SerialGatewaySender, I think I found a way of doing it which consists of changing the following call in SerialGatewaySenderEventProcessor.registerEventDroppedInPrimaryQueue():

``
this.processors.get(index).sendBatchDestroyOperationForDroppedEvent(droppedEvent, index);
``
to this one:

``
this.processors.get(index).sendBatchDestroyOperationForDroppedEvent(droppedEvent, index);
``

And adding the following method to SerialGatewaySenderQueue:
``
  public void addRemovedEvent(EntryEventImpl droppedEvent) {
    lock.writeLock().lock();
    lastDispatchedKey = droppedEvent.getTailKey();
    notifyAll();
    lock.writeLock().unlock();
  }
``

The problem is that I did not find any test case to verify that it worked correctly. I ran the tests in SerialWANPropagationsFeatureDUnitTest and it did not matter if I used the original solution, the new solution or even removed the sending of batch destroy operation. I looks as if it were not necessary. That's why I did not dare to add the change. Any idea?

DonalEvans (approved):

geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySender.java
boglesby: Oh, I didn't think about the BatchRemoval thread not being there. I'm not really sure the consequences of keeping that thread around. I just seems better for performance to not send a message on every operation, and instead batch them up. Maybe the behavior can be something like: If that cache is closing, stop the BRM; otherwise keep it around to process these messages.

albertogpz: How could I know if the cache is closing from inside the stop method of the ParallelGatewaySender? Is there a way?

albertogpz: @boglesby I reverted the change about letting the batch removal thread remove the dropped events because I did not find a way to stop the thread when the cache was closing.

@boglesby and @gesterzhou , could you please approve the PR if you are ok with the changes? Otherwise, please let me know what else should be changed.

albertogpz: @boglesby and @gesterzhou, friendly reminder :-)

geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySender.java
gesterzhou: For this part, it's unnecessary to enhance registerEventDroppedInPrimaryQueue() since it's used for a very small window (sender is restarting). 

I feel your previous solution is good enough.

gesterzhou (approved):

mkevo (approved):

PR #5510, opened 2020-09-14
GEODE-8492: Redis 'clients' statistic goes negative

sabbeyPivotal:
When running a long running app, we noticed that our "clients" statistic was negative. It should always be greater than or equal to 0.

It seems like the call to decrement the number of clients was being invoked multiple times for each client.

We created an atomic boolean to manage the client's closed state.

### DIFF
--- a/geode-redis/src/main/java/org/apache/geode/redis/internal/RedisStats.java
+++ b/geode-redis/src/main/java/org/apache/geode/redis/internal/RedisStats.java
@@ -177,6 +177,10 @@ public class RedisStats {
     stats.incLong(clientId, -1);
   }
 
+  public long getClients() {
+    return stats.getLong(clientId);
+  }
+
   public long startPassiveExpirationCheck() {
     return getTime();
   }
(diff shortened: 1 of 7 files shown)


geode-redis/src/main/java/org/apache/geode/redis/internal/RedisStats.java
```diff
@@ -177,6 +177,10 @@ public void removeClient() {
     stats.incLong(clientId, -1);
   }
 
+  public long getClients() {
```
dschneider-pivotal, line 180: add visible for testing

dschneider-pivotal (changes requested):

geode-redis/src/main/java/org/apache/geode/redis/internal/RedisStats.java
sabbeyPivotal: Just added it!

dschneider-pivotal (approved):

PR #5511, opened 2020-09-14
GEODE-8493: Redis idle clients can cause server stuck thread warning

sabbeyPivotal:
Changes type of executor service we used so that it does not log a stuck threads warning when Redis clients are idle


dschneider-pivotal (approved):

jdeppe-pivotal (changes requested):
I'd like to see this change run through `PubSubDUnitTest` under stress. To that end, I think we can de-tune the `PubSubDUnitTest.testPubSubWithMoreSubscribersThanNettyWorkerThreads` to only spin up 200 clients (no need for 1000). I've done some out-of-band testing and it looks like this will also avoid the thread starvation issue we see when running this test under stress.

jdeppe-pivotal (approved):

PR #5518, opened 2020-09-16
GEODE-8499: Redis subscriptions leak if they are not explicitly unsubscribed

sabbeyPivotal:
Subscription structures leak if clients disconnect without unsubscribing. Normally subscriptions should be cleaned up when a PUBLISH is attempted but fails. We should try and detect clients disconnecting and remove the necessary subscription structures.

This code will properly shutdown a subscription when the connection is closed.


dschneider-pivotal (approved):

sabbeyPivotal:
Distributed test failure is unrelated to this PR

PR #5532, opened 2020-09-22
GEODE-8513: Remove (de)serialization of local sessions.

pivotal-jbarrett:
Removes query that resulted in casting issue in antlr.
Clears all locally cached entries.

### For all changes:

### Note:


metatype:
I don't have any concerns about removing the `doUnload`.

metatype (approved):

jdeppe-pivotal:
IIRC, the reason this functionality was added was so that webapps could be hot-deployed which would require sessions to be 'rebuilt' since the classloader would have changed. I'm not sure if this is still a valid scenario.

pivotal-jbarrett:
Two interesting things were happing in there. First was the serialization of the session to disk, which really only made sense in peer mode since in client server the persistence is off JVM already. In peer mode it makes more sense, and is more tested, to use region persistence to achieve this hot deployment behavior. The second thing going on was the removal from the local region. This was to avoid class loader casting issues. This was just simplified to a local clear operation rather than iterating over all the sessions specific to this web context. The upside is that it takes a lot less time and doesn't invoke any queries. The downside is that other web contexts will have to refresh their local caches too but this scenario should be very small.

PR #5533, opened 2020-09-22
GEODE-8498: Redis messages written to Netty channel sometimes delivered out of order

sabbeyPivotal:
Error from .tcl tests:
```
[err]: Mix SUBSCRIBE and PSUBSCRIBE in tests/unit/pubsub.tcl
Expected 'pmessage foo.* foo.bar hello' to be equal to 'message foo.bar hello'

*** [err]: Mix SUBSCRIBE and PSUBSCRIBE in tests/unit/pubsub.tcl
Expected 'pmessage foo.* foo.bar hello' to be equal to 'message foo.bar hello'
```

When writing responses to a Netty channel, the messages may get written out of order in the `writeAndFlush` call. To verify this bug, we used a single thread to call `writeAndFlush` and added logging right before calling `writeAndFlush` and after `afterWrite` in the callback listener. We saw the messages get logged in the proper order before `writeAndFlush` was called, but the messages were logged out of order in the callback listener and arrived in the wrong order on the client.


sabbeyPivotal:
https://github.com/apache/geode/pull/5550

PR #5536, opened 2020-09-22
GEODE-8520: GCStatsMonitor should sum up all the GC stats to get the …

jinmeiliao:
…total GC count and GC time.

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/management/internal/beans/stats/GCStatsMonitor.java
+++ b/geode-core/src/main/java/org/apache/geode/management/internal/beans/stats/GCStatsMonitor.java
@@ -15,12 +15,17 @@
 package org.apache.geode.management.internal.beans.stats;
 
 
+import java.util.HashMap;
+import java.util.Map;
+
 import org.apache.geode.StatisticDescriptor;
 import org.apache.geode.Statistics;
+import org.apache.geode.annotations.VisibleForTesting;
 import org.apache.geode.distributed.ConfigurationProperties;
 import org.apache.geode.internal.statistics.StatisticId;
 import org.apache.geode.internal.statistics.StatisticNotFoundException;
 import org.apache.geode.internal.statistics.StatisticsNotification;
+import org.apache.geode.internal.statistics.ValueMonitor;
 
 /**
  * This class acts as a monitor and listen for GC statistics updates on behalf of MemberMBean.
@@ -40,29 +45,49 @@ import org.apache.geode.internal.statistics.StatisticsNotification;
  */
 public class GCStatsMonitor extends MBeanStatsMonitor {
   // this class uses these volatile variables to make sure reads are reading the latest values
-  // it is not using the parent's siteMap which is not volatile to keep the stats values.
-  private volatile long collections = 0;
-  private volatile long collectionTime = 0;
+  // it is not using the parent's siteMap
+  // this stores each stat's uniqueId and its collection count
+  private volatile Map<Long, Number> collections;
+
+  // this stors eaech stat's uniqueId and its collectionTime
+  private volatile Map<Long, Number> collectionTime;
 
   public GCStatsMonitor(String name) {
-    super(name);
+    this(name, new ValueMonitor());
+  }
+
+  @VisibleForTesting
+  public GCStatsMonitor(String name, ValueMonitor valueMonitor) {
+    super(name, valueMonitor);
+    collections = new HashMap<>();
+    collectionTime = new HashMap<>();
+  }
+
+  long getCollections() {
+    return collections.values().stream().mapToLong(Number::longValue).sum();
+  }
+
+  long getCollectionTime() {
+    return collectionTime.values().stream().mapToLong(Number::longValue).sum();
   }
 
   @Override
-  // this will be called multiple times with different collector's stats
+  // this will be called multiple times initially with different collector's stats
   public void addStatisticsToMonitor(Statistics stats) {
     monitor.addListener(this);// if already listener is added this will be a no-op
     monitor.addStatistics(stats);
 
-    // stats map should keep the sum of all the GC stats
     StatisticDescriptor[] descriptors = stats.getType().getStatistics();
     for (StatisticDescriptor d : descriptors) {
       String name = d.getName();
       Number value = stats.get(d);
+      if (value == null) {
+        continue;
+      }
       if (name.equals(StatsKey.VM_GC_STATS_COLLECTIONS)) {
-        collections += value.longValue();
+        collections.put(stats.getUniqueId(), value);
       } else if (name.equals(StatsKey.VM_GC_STATS_COLLECTION_TIME)) {
-        collectionTime += value.longValue();
+        collectionTime.put(stats.getUniqueId(), value);
       }
     }
   }
@@ -70,20 +95,19 @@ public class GCStatsMonitor extends MBeanStatsMonitor {
   @Override
   public Number getStatistic(String statName) {
     if (statName.equals(StatsKey.VM_GC_STATS_COLLECTIONS)) {
-      return collections;
+      return getCollections();
     }
 
     if (statName.equals(StatsKey.VM_GC_STATS_COLLECTION_TIME)) {
-      return collectionTime;
+      return getCollectionTime();
     }
+
     return 0;
   }
 
   @Override
+  // this will be called each time a collector's stats changed
   public void handleNotification(StatisticsNotification notification) {
-    // sum up all the count and all the time in the stats included in this notification
-    long totalCount = 0;
-    long totalTime = 0;
     for (StatisticId statId : notification) {
       StatisticDescriptor descriptor = statId.getStatisticDescriptor();
       String name = descriptor.getName();
@@ -95,15 +119,12 @@ public class GCStatsMonitor extends MBeanStatsMonitor {
       }
       log(name, value);
       if (name.equals(StatsKey.VM_GC_STATS_COLLECTIONS)) {
-        totalCount += value.longValue();
+        collections.put(statId.getStatistics().getUniqueId(), value);
       }
 
       else if (name.equals(StatsKey.VM_GC_STATS_COLLECTION_TIME)) {
-        totalTime += value.longValue();
+        collectionTime.put(statId.getStatistics().getUniqueId(), value);
       }
     }
-
-    collections = totalCount;
-    collectionTime = totalTime;
   }
 }
--- a/geode-core/src/main/java/org/apache/geode/management/internal/beans/stats/GCStatsMonitor.java
+++ b/geode-core/src/main/java/org/apache/geode/management/internal/beans/stats/GCStatsMonitor.java
@@ -14,9 +14,9 @@
  */
 package org.apache.geode.management.internal.beans.stats;
 
-import java.util.Map;
 
 import org.apache.geode.StatisticDescriptor;
+import org.apache.geode.Statistics;
 import org.apache.geode.distributed.ConfigurationProperties;
 import org.apache.geode.internal.statistics.StatisticId;
 import org.apache.geode.internal.statistics.StatisticNotFoundException;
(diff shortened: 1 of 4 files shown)


geode-core/src/test/java/org/apache/geode/management/bean/stats/MemberLevelStatsTest.java
```diff
@@ -339,6 +342,41 @@ public void testRegionCounters() {
...
+    assertThat(memberMBeanBridge.getGarbageCollectionCount()).isEqualTo(newTotals[0]);
+    assertThat(memberMBeanBridge.getGarbageCollectionTime()).isEqualTo(newTotals[1]);
+  }
+
+  private long[] modifyStatsAndReturnTotalCountAndTime(
+      long baseCount, long baseTime,
+      Statistics[] modifiedStats) {
+    long[] totalCountAndTime = {0, 0};
+    for (Statistics gcStat : modifiedStats) {
+      StatisticDescriptor[] statistics = gcStat.getType().getStatistics();
```
mhansonp, line 364: Why did we switch from getCollections and getCollectionTime to the string based approach?

mhansonp (approved):
Other than the string lookups, things look fine. I recognize some of that was there from before.

geode-core/src/test/java/org/apache/geode/management/bean/stats/MemberLevelStatsTest.java
jinmeiliao: those two methods are removed from the GCMonitor. They were package protected anyway, no usage besides this tests. production code is using getStatistics method.

geode-core/src/test/java/org/apache/geode/management/bean/stats/MemberLevelStatsTest.java
```diff
@@ -339,6 +342,41 @@ public void testRegionCounters() {
...
+      for (StatisticDescriptor d : statistics) {
+        if ("collections".equals(d.getName())) {
```
kirklund, line 366: I think you should either reference a constant instead of referencing "collections" and "collectionTime" in multiple classes.

geode-core/src/main/java/org/apache/geode/management/internal/beans/stats/GCStatsMonitor.java
```diff
@@ -39,69 +39,71 @@
...
+    monitor.addListener(this);// if already listener is added this will be a no-op
+    monitor.addStatistics(stats);
+
+    // stats map should keep the sum of all the GC stats
+    StatisticDescriptor[] descriptors = stats.getType().getStatistics();
+    for (StatisticDescriptor d : descriptors) {
+      String name = d.getName();
+      Number value = stats.get(d);
+      if (name.equals(StatsKey.VM_GC_STATS_COLLECTIONS)) {
+        collections += value.longValue();
```
kirklund, line 63: Using += on a volatile field is not thread safe. I recommend changing collections and collectionTime to be AtomicLong instances. Then you can use incrementAndGet() instead of +=. Even if multiple threads do not currently invoke this method, maybe they will in the future.

kirklund (dismissed):

geode-core/src/main/java/org/apache/geode/management/internal/beans/stats/GCStatsMonitor.java
```diff
@@ -39,69 +39,71 @@
...
+      }
+
+      else if (name.equals(StatsKey.VM_GC_STATS_COLLECTION_TIME)) {
+        totalTime += value.longValue();
+      }
     }
+
+    collections = totalCount;
+    collectionTime = totalTime;
   }
```
demery-pivotal, line 108: I think a notification includes only those stats whose values have changed. If that's correct, what we're summing here is only the _changed_ GC stats, not _all_ of the GC stats. This will work only if we're absolutely certain that every stat will change on every sample.

See `ValueMonitor.monitorStatistics()`, which builds the notification only from updated stats.

See `SampleCollector.sample()`, which adds a stats to the updated stats list only if the value of the stat changed from the previous sample.

demery-pivotal: Please write unit tests for `GCSTatsMonitor` directly, rather than testing only through `MemberLevelStatsTest`.

demery-pivotal (changes requested):
This code assumes that every GC stat will change on every sample. Please validate that assumption (I'm not sure how to validate it), or change the code so that it doesn't make that assumption.

Test `GCStatsMonitor` directly.

geode-core/src/main/java/org/apache/geode/management/internal/beans/stats/GCStatsMonitor.java
jinmeiliao: > I think a notification includes only those stats whose values have changed. If that's correct, what we're summing here is only the _changed_ GC stats, not _all_ of the GC stats. This will work only if we're absolutely certain that every stat will change on every sample.
> 
> See `ValueMonitor.monitorStatistics()`, which builds the notification only from updated stats.
> 
> See `SampleCollector.sample()`, which adds a stats to the updated stats list only if the value of the stat changed from the previous sample.

then what would those count/time mean though. It's not right to report just a single one of them. It's not right to sum them up either then....

jinmeiliao: > Please write unit tests for `GCSTatsMonitor` directly, rather than testing only through `MemberLevelStatsTest`.

the unit test will be just like the MemberLevelStatsTest, only by mocking all those variables.

demery-pivotal: > then what would those count/time mean though. It's not right to report just a single one of them. It's not right to sum them up either then....

The stats notification stuff is specifically designed to report only changes. As a result, old gen and new gen CGs are almost always reported independently (unless they both happen to change during a single sample).

I think that's why the old code was doing the weird stuff. It was trying to track the previous value for each kind of GC, so that when a new value arrived, it could add in just the change in that kind of GC. Decrementing and incrementing was a puzzling way to do that, but that was the goal.

Here's a way that seems simpler to me than decrementing/incrementing. I'll use collections as an example, but it's the same for collectionTime.
- Create a map that records the latest value for each GC type.
- Create a field that records the sum of values for all types. This is what gets reported via getCollections()
- When a new kind of GC is added via addStatisticsToMonitor()…
    1.  Store the initial value for that GC type in the map.
    2. Sum the values of all GC types in the map, and store the sum in the field.
- When a notification arrives…
    1. Store the new value for that GC type in the map.
    2. Sum the values of all GC types in the map, and store the sum in the field.

demery-pivotal: Let me correct my terminology a bit… I keep saying "GC type," and that's not the right distinction. What distinguishes one batch of GC stats from another is which GarbageCollectorMXBean the values come from. I think (but I'm not sure) that each GC bean reports about a particular memory space.

So where I say "GC type," a better term might be "memory space".

geode-core/src/main/java/org/apache/geode/management/internal/beans/stats/GCStatsMonitor.java
```diff
@@ -39,69 +39,71 @@
...
-      collections += value.longValue();
-      return;
-    }
-
-    if (name.equals(StatsKey.VM_GC_STATS_COLLECTION_TIME)) {
-      collectionTime += value.longValue();
-      return;
+  @Override
+  // this will be called multiple times with different collector's stats
+  public void addStatisticsToMonitor(Statistics stats) {
```
jchen21, line 53: The `addStatisticsToMonitor` method ignores the descriptors other than `VM_GC_STATS_COLLECTIONS` and `VM_GC_STATS_COLLECTION_TIME`. Can we guarantee that there is no other descriptor?

demery-pivotal: Those are currently the only two descriptors in the GC statistics type, and the only two values reported by GarbageCollectorMXBean (the ultimate source of the values). If another descriptor were to be added to the statistics type in the future, we'd want to review it before deciding how `GCStatsMonitor` should handle it. So I think it's appropriate to ignore any descriptors we don't recognize.

geode-core/src/main/java/org/apache/geode/management/internal/beans/stats/GCStatsMonitor.java
jinmeiliao: > > then what would those count/time mean though. It's not right to report just a single one of them. It's not right to sum them up either then....
> 
> The stats notification stuff is specifically designed to report only changes. As a result, old gen and new gen CGs are almost always reported independently (unless they both happen to change during a single sample).
> 
> I think that's why the old code was doing the weird stuff. It was trying to track the previous value for each kind of GC, so that when a new value arrived, it could add in just the change in that kind of GC. Decrementing and incrementing was a puzzling way to do that, but that was the goal.
> 
> Here's a way that seems simpler to me than decrementing/incrementing. I'll use collections as an example, but it's the same for collectionTime.
> 
> * Create a map that records the latest value for each GC type.
> * Create a field that records the sum of values for all types. This is what gets reported via getCollections()
> * When a new kind of GC is added via addStatisticsToMonitor()…
>   
>   1. Store the initial value for that GC type in the map.
>   2. Sum the values of all GC types in the map, and store the sum in the field.
> * When a notification arrives…
>   
>   1. Store the new value for that GC type in the map.
>   2. Sum the values of all GC types in the map, and store the sum in the field.

Yes, this sounds like a better idea. I will implement it.

geode-core/src/main/java/org/apache/geode/management/internal/beans/stats/GCStatsMonitor.java
jinmeiliao: I think we are only interested in these two.

geode-core/src/main/java/org/apache/geode/management/internal/beans/stats/GCStatsMonitor.java
jinmeiliao: Not using += anymore.

geode-core/src/test/java/org/apache/geode/management/internal/beans/stats/GCStatsMonitorTest.java
```diff
@@ -33,16 +45,94 @@
...
+    StatisticDescriptor d1 = mock(StatisticDescriptor.class);
+    when(d1.getName()).thenReturn(StatsKey.VM_GC_STATS_COLLECTIONS);
+    StatisticDescriptor d2 = mock(StatisticDescriptor.class);
+    when(d2.getName()).thenReturn(StatsKey.VM_GC_STATS_COLLECTION_TIME);
+    StatisticDescriptor[] descriptors = {d1, d2};
+    StatisticsType type = mock(StatisticsType.class);
+    when(stats.getType()).thenReturn(type);
+    when(type.getStatistics()).thenReturn(descriptors);
+
+    when(stats.get(any(StatisticDescriptor.class))).thenReturn(8L, 300L);
```
demery-pivotal, line 74: This way of specifying return values unnecessarily couples the test to the current implementation—it insists that the implementation must call `stats.get(sd)` this many times, and in this specific order.

On this line and the two later ones, it would be better to associate each value with the appropriate statistic descriptor, the way the other test does.

demery-pivotal (approved):

geode-core/src/test/java/org/apache/geode/management/bean/stats/MemberLevelStatsTest.java
mhansonp: Thanks!

jchen21 (approved):

PR #5543, opened 2020-09-23
GEODE-8496: fix rest management test after dependency bump

onichols-pivotal:
I've uncommented the test that started failing after 37 dependencies were bumped for GEODE-8496.  It appears as if the test was relying on a quirk of an underlying library to generate the unusual exception text it is expecting...seems as if we should really be enforcing this restriction in Geode code, if it's important...


onichols-pivotal:
Fixed by #5547 thanks!

PR #5544, opened 2020-09-23
GEODE-8515: Redis PING should respond appropriately when called from within a SUBSCRIBE/PSUBSCRIBE

sabbeyPivotal:
From PING documentation (https://redis.io/commands/ping):

If the client is subscribed to a channel or a pattern, it will instead return a multi-bulk with a "pong" in the first position and an empty bulk in the second position, unless an argument is provided in which case it returns a copy of the argument.

### DIFF
--- a/geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/pubsub/SubscriptionsIntegrationTest.java
+++ b/geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/pubsub/SubscriptionsIntegrationTest.java
@@ -39,7 +39,7 @@ public class SubscriptionsIntegrationTest {
 
   @Test
   public void pingWhileSubscribed() {
-    Jedis client = new Jedis("localhost", server.getPort(), 1000000000);
+    Jedis client = new Jedis("localhost", server.getPort());
     MockSubscriber mockSubscriber = new MockSubscriber();
 
     executor.submit(() -> client.subscribe(mockSubscriber, "same"));
--- a/geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/pubsub/SubscriptionsIntegrationTest.java
+++ b/geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/pubsub/SubscriptionsIntegrationTest.java
@@ -38,6 +38,42 @@ public class SubscriptionsIntegrationTest {
   @ClassRule
   public static ExecutorServiceRule executor = new ExecutorServiceRule();
 
+  // @Test
+  // public void pingWith
+  @Test
+  public void pingWhileSubscribed() {
+    Jedis client = new Jedis("localhost", server.getPort(), 1000000000);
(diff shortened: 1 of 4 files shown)


sabbeyPivotal:
Distributed test failure is unrelated to this PR.

geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/pubsub/SubscriptionsIntegrationTest.java
```diff
@@ -39,9 +38,8 @@
   public static ExecutorServiceRule executor = new ExecutorServiceRule();
 
   @Test
-  @Ignore("GEODE-8515")
   public void pingWhileSubscribed() {
-    Jedis client = new Jedis("localhost", server.getPort());
+    Jedis client = new Jedis("localhost", server.getPort(), 1000000000);
```
dschneider-pivotal, line 42: why a long timeout on this one but not on pingWithArgumentWhileSubscribed?

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/connection/PingExecutor.java
```diff
@@ -31,9 +32,17 @@ public RedisResponse executeCommand(Command command,
       ExecutionHandlerContext context) {
     List<byte[]> commandElems = command.getProcessedCommand();
     byte[] result = PING_RESPONSE.getBytes();
+    byte[] subscribeResult = "".getBytes();
     if (commandElems.size() > 1) {
       result = commandElems.get(1);
+      subscribeResult = result;
     }
+
+    if (!context.getPubSub().findSubscribedChannels(context.getClient()).isEmpty()) {
```
dschneider-pivotal, line 41: consider refactoring this so that we only compute the RedisResponse for subscribe if we find a client and then otherwise with (an else) compute the RedisResponse when not subscribed. I don't like all the logic we have initializing two different results and then we only end up using one or the other. I'd prefer that we only have state at the top that would be shared by either branch (for example commandElems). Once the RedisResponse is done we can then fall through to a command return.

geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/pubsub/SubscriptionsIntegrationTest.java
sabbeyPivotal: Thank you for calling this out.  I had a long timeout for debugging and forgot to remove it.

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/connection/PingExecutor.java
sabbeyPivotal: I refactored it. Check it out and let me know what you think.

dschneider-pivotal (approved):

dschneider-pivotal (approved):

PR #5546, opened 2020-09-23
GEODE-8523: Session state document update

nabarunnag:
* geode-tcp and geode-membership added to the list
	* removed geode-json as it does not exist anymore.


jhuynh1 (approved):

jhuynh1:
This might have to be backported to previous releases based on when these jars were introduced...

PR #5549, opened 2020-09-24
GEODE-8533: Docs - compaction-threshold mechanism description are wrong

davebarnes97:
Questions for reviewers:
- Have the explanations satisfactorily stated that the compaction-threshold is a percentage of live data (non-garbage), as opposed to a percentage of garbage?
- Is it clear that the threshold is activated when the live data percentage drops below the threshold, as opposed to exceeding it?
- Is the modified formula for calculating required disk space correct?
Thanks!


upthewaterspout (approved):

upthewaterspout (approved):

karensmolermiller (approved):
The prose changes look good to me. Thanks for the clarifications.

PR #5557, opened 2020-09-25
GEODE-8540: Create new DistributedBlackboard Rule

kirklund:
Package up DUnitBlackboard as a JUnit Rule named DistributedBlackboard.

PR #5558, opened 2020-09-25
GEODE-8539: Update FixedPartitioningDUnitTest with Rules

kirklund:
Depends on #5557


demery-pivotal (approved):

mhansonp (approved):
Nice cleanup. Looks good. I ran than these tests and everything passed.

PR #5559, opened 2020-09-25
GEODE-8544: Making VM class start versioned VM

mhansonp:
Fixing an issue where the VM class will not properly start a versioned server.


kirklund (approved):
Great catch! Thanks.

PR #5560, opened 2020-09-25
GEODE-7845: adding PRClear protection code

mhansonp:
- prevent PRClear messages from going to older servers.


mhansonp:
I am not a fan of this solution at all. I am hoping people can provide some useful feedback to make this way better.

PR #5567, opened 2020-09-29
GEODE-8547: Added impacts to show missing disk-stores

mivanac:

### For all changes:

### Note:

### DIFF
--- a/geode-gfsh/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/ShowMissingDiskStoreCommandDUnitTest.java
+++ b/geode-gfsh/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/ShowMissingDiskStoreCommandDUnitTest.java
@@ -28,15 +28,18 @@ import org.junit.rules.TestName;
 import org.apache.geode.cache.CacheFactory;
 import org.apache.geode.cache.Region;
 import org.apache.geode.cache.RegionShortcut;
+import org.apache.geode.distributed.DistributedSystemDisconnectedException;
 import org.apache.geode.management.internal.cli.result.CommandResult;
 import org.apache.geode.management.internal.cli.result.model.ResultModel;
 import org.apache.geode.management.internal.cli.result.model.TabularResultModel;
 import org.apache.geode.management.internal.cli.util.CommandStringBuilder;
 import org.apache.geode.management.internal.i18n.CliStrings;
+import org.apache.geode.test.dunit.IgnoredException;
 import org.apache.geode.test.dunit.rules.ClusterStartupRule;
 import org.apache.geode.test.dunit.rules.MemberVM;
 import org.apache.geode.test.junit.categories.PersistenceTest;
 import org.apache.geode.test.junit.rules.GfshCommandRule;
+import org.apache.geode.test.junit.rules.LocatorStarterRule;
 
 @Category({PersistenceTest.class})
 public class ShowMissingDiskStoreCommandDUnitTest {
@@ -60,6 +63,7 @@ public class ShowMissingDiskStoreCommandDUnitTest {
 
     // start a server so that we can execute data commands that requires at least a server running
   }
+
 
   @Test
   public void showMissingDiskStoresDoesNotDuplicateDiskStores() {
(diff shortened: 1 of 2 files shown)


kirklund (changes requested):
Please add some testing to cover the changes.

mivanac:
Test added.

mivanac:
@kirklund  is added test OK with you.

kohlmu-pivotal (changes requested):
I've run your test by only rolling back the ShowMissingDiskStoreCommand and it seems to pass either way. So I'm not sure what this test is actually supposed to test or resolve.

Please clarify what this fix is supposed to fix and have a test that proves that a problem exists. As I don't see any problem by rolling the Command code.

mivanac:
Hi, when I remove my changes in ShowMissingDiskStoreCommand, this TC fails. Problem here is that without these changes, command is rejected, with message "No caching members found".

jinmeiliao:
I believe it's as designed that if no members are up and running, then `show missing disk-store` will not work.  I don't quite understand what's the problem you are trying to solve by this fix?

mivanac:
We are solving a problem when all members are shutdown, and after some time we are trying to startup. Let say, that due to some problems locator with latest info, cannot start. In this case, remaining locator will not start, so we need to use command revoke missing-disk-store, which needs input from command show missing disk-store.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/ShowMissingDiskStoreCommand.java
```diff
@@ -53,16 +53,19 @@ public ResultModel showMissingDiskStore() {
     Set<DistributedMember> dataMembers =
         DiskStoreCommandsUtils.getNormalMembers((InternalCache) getCache());
 
-    if (dataMembers.isEmpty()) {
-      return ResultModel.createInfo(CliStrings.NO_CACHING_MEMBERS_FOUND_MESSAGE);
+    List<ColocatedRegionDetails> missingRegions = null;
+    boolean cacheMemberExist = false;
+
+    if (!dataMembers.isEmpty()) {
+      cacheMemberExist = true;
```
jinmeiliao, line 60: I believe you don't need this boolean. If no servers, `missingRegions` will be null, otherwise, it's empty or has data

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/ShowMissingDiskStoreCommand.java
```diff
@@ -113,7 +116,9 @@ private ResultModel toMissingDiskStoresTabularResult(
     }
 
     TabularResultModel missingRegionsSection = result.addTable(MISSING_COLOCATED_REGIONS_SECTION);
-    if (hasMissingColocatedRegions) {
+    if (!cacheMemberExist) {
```
jinmeiliao, line 119: here you can do something like:
```
if (missingColocatedRegions == null) {
  // set the header as "no caching members"
}
else if (missingColocatedRegions.isEmpty()) {
....
}
else{
....
}

File: geode-gfsh/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/ShowMissingDiskStoreCommandDUnitTest.java [hunk 3]
```diff
@@ -156,6 +160,71 @@ public void stoppingAndRestartingMemberDoesNotResultInMissingDiskStore() {
...
 
+
+  @Test
+  public void stopAllMembersAndStart2ndLocator() throws Exception {
+    IgnoredException.addIgnoredException(DistributedSystemDisconnectedException.class);
+
+    MemberVM locator1 = lsRule.startLocatorVM(1, locator.getPort());
+
+    MemberVM server1 = lsRule.startServerVM(2, locator.getPort(), locator1.getPort());
+    @SuppressWarnings("unused")
```
jinmeiliao, line 171: instead of suppress the warning, just don't assign it to a variable

geode-gfsh/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/ShowMissingDiskStoreCommandDUnitTest.java
```diff
@@ -156,6 +160,71 @@ public void stoppingAndRestartingMemberDoesNotResultInMissingDiskStore() {
...
+    MemberVM server2 = lsRule.startServerVM(3, locator.getPort(), locator1.getPort());
+
+    final String testRegionName = "regionA";
```
jinmeiliao, line 174: I believe you can just create a region without the diskstore, the same problem would still manifest. Make the test simpler

geode-gfsh/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/ShowMissingDiskStoreCommandDUnitTest.java
```diff
@@ -156,6 +160,71 @@ public void stoppingAndRestartingMemberDoesNotResultInMissingDiskStore() {
...
+        .addOption(CliStrings.CREATE_REGION__REGIONSHORTCUT,
+            RegionShortcut.PARTITION_REDUNDANT_PERSISTENT.toString());
+    await().untilAsserted(() -> gfshConnector.executeAndAssertThat(createRegion.getCommandString())
+        .statusIsSuccess());
+
+
+    lsRule.stop(1, false);
+
+    // Add data to the region
+    addData(server1, testRegionName);
```
jinmeiliao, line 193: you also don't need to add data and do rebalance

jinmeiliao (changes requested):

mivanac:
Thanks for the comments.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/ShowMissingDiskStoreCommand.java
```diff
@@ -95,7 +96,8 @@ private ResultModel toMissingDiskStoresTabularResult(
     ResultModel result = new ResultModel();
 
     boolean hasMissingDiskStores = missingDiskStores.length != 0;
-    boolean hasMissingColocatedRegions = !missingColocatedRegions.isEmpty();
+    boolean hasMissingColocatedRegions =
```
jinmeiliao, line 99: this variable is not used anymore.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/ShowMissingDiskStoreCommand.java
```diff
@@ -95,7 +96,8 @@ private ResultModel toMissingDiskStoresTabularResult(
     ResultModel result = new ResultModel();
 
     boolean hasMissingDiskStores = missingDiskStores.length != 0;
```
jinmeiliao, line 98: actually you can get rid of this variable and inline this too, to make it symetric.

jinmeiliao (approved):
Good work!

kohlmu-pivotal (approved):

PR #5569, opened 2020-09-29
GEODE-8555: SimpleDiskRegionJunitTest fails on Windows

sabbey37:
Co-authored-by: Ray Ingles <ringles@pivotal.io>

Updating Junit to 4.13, one test failed on Windows due to failure to delete some temporary test files.

https://concourse.apachegeode-ci.info/teams/main/pipelines/apache-develop-main/jobs/WindowsCoreIntegrationTestOpenJDK8/builds/455

```
org.apache.geode.internal.cache.SimpleDiskRegionJUnitTest > testBasicClose FAILED
    java.lang.AssertionError:  Exception in createOverflowOnly due to java.lang.IllegalStateException: The region "/testRegion" has been persisted to disk so it can not be recreated on the same disk store without persistence. Either destroy the persistent region, recreate it as overflow and persistent, or create the overflow only region on a different disk store.
        at org.junit.Assert.fail(Assert.java:89)
        at org.apache.geode.internal.cache.SimpleDiskRegionJUnitTest.testBasicClose(SimpleDiskRegionJUnitTest.java:75)
```

### DIFF
--- a/geode-junit/src/main/java/org/apache/geode/internal/cache/DiskRegionTestingBase.java
+++ b/geode-junit/src/main/java/org/apache/geode/internal/cache/DiskRegionTestingBase.java
@@ -30,6 +30,7 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.Properties;
 
+import org.apache.commons.io.FileUtils;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Rule;
@@ -46,6 +47,7 @@ import org.apache.geode.cache.Region;
 import org.apache.geode.cache.RegionDestroyedException;
 import org.apache.geode.distributed.DistributedSystem;
 import org.apache.geode.internal.cache.versions.VersionTag;
+import org.apache.geode.logging.internal.log4j.api.LogService;
 
 /**
  * DiskRegionTestingBase: This class is extended to write more JUnit tests for Disk Regions.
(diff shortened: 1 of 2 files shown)


sabbey37:
@onichols-pivotal @kirklund

geode-junit/src/main/java/org/apache/geode/internal/cache/DiskRegionTestingBase.java
```diff
@@ -200,6 +202,19 @@ protected void deleteFiles() {
...
 
+  protected void forceDeleteFiles() {
+    closeDiskStores();
+    File file = tempDir.getRoot();
+    File[] files = file.listFiles();
+    for (File each : files) {
+      try {
+        FileUtils.forceDelete(each);
+      } catch (IOException e) {
+        LogService.getLogger().error("Caught", e);
```
kirklund, line 213: Instead of logging the error, can we add and use ErrorCollector? I would just add this to DiskRegionTestingBase up where the other Rules are defined.
```
  @Rule
  public ErrorCollector errorCollector = new ErrorCollector();
```
And then:
```
      try {
        FileUtils.forceDelete(each);
      } catch (IOException e) {
        errorCollector.addError(e);
      }
```
After each test finishes, (if any were thrown) ErrorCollector will fail the test and print out every stack trace.

Since SimpleDiskRegionJUnitTest is an integration test, logging the stack traces won't cause the test to fail like a distributed test would.

kirklund (changes requested):

geode-junit/src/main/java/org/apache/geode/internal/cache/DiskRegionTestingBase.java
sabbey37: I'd definitely like to do something more than just logging the error, but not all of the files are deleted (like `statsstats.gfs` file).  The `statsstats.gfs` file doesn't have to be deleted, though.  We wrapped the `FileUtils.forceDelete(each);` method line with a `if (!each.getName().contains(".gfs"))` since that file doesn't have to be deleted.  Let me know what you think!

kirklund (approved):

PR #5570, opened 2020-09-29
GEODE-8550: Create CloseableReference

kirklund:
Create CloseableReference JUnit Rule for unit and integration tests.

### DIFF
--- a/geode-junit/src/main/java/org/apache/geode/test/junit/rules/CloseableReference.java
+++ b/geode-junit/src/main/java/org/apache/geode/test/junit/rules/CloseableReference.java
@@ -1,0 +1,208 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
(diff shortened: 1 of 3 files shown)


geode-junit/src/main/java/org/apache/geode/test/junit/rules/CloseableReference.java
```diff
@@ -0,0 +1,208 @@
...
+
+/**
+ * JUnit Rule that provides automated tearDown for an atomic reference. If the referenced value is
+ * an {@code AutoCloseable} or {@code Closeable} then it will be auto-closed and set to null during
+ * tear down.
+ *
+ * <p>
+ * If the referenced value is not an {@code AutoCloseable} or {@code Closeable}, the
+ * {@code CloseableReference} will use reflection to invoke any method named {@code close} or
+ * {@code disconnect} regardless of what interfaces are implemented.
```
demery-pivotal, line 35: Also: stop()

demery-pivotal (approved):

PR #5571, opened 2020-09-29
GEODE-8558: query input by users should trim newlines and comments.

jinmeiliao:


mhansonp (approved):
I like that there are a good number of tests. Nicely done.

kirklund (approved):

PR #5574, opened 2020-09-30
GEODE-8561: Add support to PR pipeline for running Windows checks

onichols-pivotal:
fix for jinja template variable that isn't always present for windows


sabbey37 (approved):
Looks good, seems like you've included conditions for all the jinja variables that might/might not be there for Windows tests.

PR #5585, opened 2020-10-01
GEODE-8533: Docs - compaction-threshold description refinements

davebarnes97:
One more garbage vs. non-garbage clarification, and added a link to making current oplog eligible for compaction by forcing oplog rotation


karensmolermiller (approved):
Looks good to me.

PR #5595, opened 2020-10-05
GEODE-8572: Make LogExporter not read dirs

demery-pivotal:
Changed LogExporter to refrain from attempting to read directories, even
when a directory's path matches the exporter's file selector predicates.

Changed LogExporterIntegrationTest to configure the server to write
files in a unique subdirectory of the current working directory. Some
tests were configuring the server to use a temporary folder that could
be reused by other tests.

### DIFF
--- a/geode-gfsh/src/integrationTest/java/org/apache/geode/management/internal/cli/util/LogExporterIntegrationTest.java
+++ b/geode-gfsh/src/integrationTest/java/org/apache/geode/management/internal/cli/util/LogExporterIntegrationTest.java
@@ -18,21 +18,19 @@ import static org.apache.geode.distributed.ConfigurationProperties.LOG_FILE;
 import static org.apache.geode.distributed.ConfigurationProperties.STATISTIC_ARCHIVE_FILE;
 import static org.assertj.core.api.Assertions.assertThat;
 
-import java.io.File;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.util.Arrays;
+import java.nio.file.Paths;
 import java.util.List;
-import java.util.Properties;
 import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipFile;
 
-import com.google.common.collect.Sets;
 import org.apache.commons.io.FileUtils;
 import org.apache.logging.log4j.Level;
+import org.junit.After;
 import org.junit.Before;
 import org.junit.Ignore;
 import org.junit.Rule;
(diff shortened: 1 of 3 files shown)


jinmeiliao (approved):
LGTM

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/util/LogExporter.java
```diff
@@ -181,12 +180,13 @@ private long filterAndSize(Path originalLogFile) throws IOException {
...
   private List<Path> findFiles(Path workingDir, Predicate<Path> fileSelector) throws IOException {
-    Stream<Path> selectedFiles/* = null */;
     if (!workingDir.toFile().isDirectory()) {
       return Collections.emptyList();
     }
-    selectedFiles = Files.list(workingDir).filter(fileSelector).filter(this.logFilter::acceptsFile);
-
-    return selectedFiles.collect(toList());
+    return Files.list(workingDir)
+        .filter(Files::isRegularFile)
```
jchen21, line 187: By default, this follows the symbolic link. Just make sure this is expected.

geode-gfsh/src/integrationTest/java/org/apache/geode/management/internal/cli/util/LogExporterIntegrationTest.java
```diff
@@ -50,127 +48,146 @@
...
+
+    List<Path> statsFiles = logExporter.findStatFiles(serverFilesDir);
+    assertThat(statsFiles)
+        .as("stat files")
+        .hasSize(1);
+    assertThat(statsFiles.get(0)).hasFileName(statsFileName);
   }
 
   @Test
   @Ignore("GEODE-2574: fix .gz suffix")
```
jchen21, line 126: Not sure why this test is @Ignore. Any change to the test is not executed.

jchen21 (approved):

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/util/LogExporter.java
demery-pivotal: I hadn't thought about that. Is there a way to ensure that the path represents a regular file without following symbolic links?

geode-gfsh/src/integrationTest/java/org/apache/geode/management/internal/cli/util/LogExporterIntegrationTest.java
demery-pivotal: Though the test is ignored, I updated it to match the style of the tests I changed. I left it ignored because the system fails the test.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/util/LogExporter.java
jchen21: You might want [`NOFOLLOW_LINKS`](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#isRegularFile-java.nio.file.Path-java.nio.file.LinkOption...-).

demery-pivotal: Here's a scenario: Let's say the path is a symbolic link, and it links to a regular file.

My understanding is:
- If I use `NOFOLLOW_LINKS`, the exporter will skip the regular file.
- If I omit `NOFOLLOW_LINKS`, the exporter will export the regular file.

Is my understanding correct? If so, are you saying that it's best to ignore the regular file if the path we're testing is a symbolic link?

jchen21: If you use `NOFOLLOW_LINKS`, for the scenario, `Files::isRegularFile` returns `false`. 
If you omit `NOFOLLOW_LINKS`, `Files::isRegularFile` returns `true`. 
I haven't tried it with the exporter though.

demery-pivotal: Yes, that's what I thought, and I verified with some experiments (which did not involve the exporter).

My question remains: If the path we're testing is a symbolic link, are you saying that you _want_ to ignore the file it links to? If so, what makes you want to skip the file in that case?

One possibility: The link links to a file in the same dir, which would lead us to export the content twice (once via the file, and once via the link to the file). We could solve this by resolving the real path to the file, and collecting the matching files into a Set instead of a List.

Another possibility: The link links to a file in some other directory, and some concern makes you not want to read the file from the other directory.

jchen21: No, I am not saying I want to ignore the file it links to. Actually, I am not sure when it should follow the link, or when it should not. Based on my understanding, most likely, it should follow the link. You probably have more context than I have. What I am trying to say is that `Files::isRegularFile` has an option `NOFOLLOW_LINKS` for symbolic link. And symbolic link could be a tricky case. As long as it works for the scenario as you expect, it should be good.

demery-pivotal: Okay. I'll merge this, because it fixes the problem of trying to read directories, but will otherwise behave as before. In particular, before and after this fix it follows symbolic links when exporting files.

As for whether we _ought_ to follow symbolic links when we're exporting, I'm confident I have no more context than you do about that ;-)

PR #5596, opened 2020-10-05
GEODE-8574: ClusterManagementService should not throw ClassCastExcept…

jinmeiliao:
…ion when function execution throws an Exception.

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/management/internal/api/LocatorClusterManagementService.java
+++ b/geode-core/src/main/java/org/apache/geode/management/internal/api/LocatorClusterManagementService.java
@@ -663,7 +663,7 @@ public class LocatorClusterManagementService implements ClusterManagementService
       }
       if (functionResult instanceof Throwable) {
         // log the exception and continue
-        logger.info("Error executing CacheRealizationFunction.", (Throwable) functionResult);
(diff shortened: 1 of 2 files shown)


geode-core/src/main/java/org/apache/geode/management/internal/api/LocatorClusterManagementService.java
```diff
@@ -653,7 +651,23 @@ public void close() {
...
+  @VisibleForTesting
+  <R> List<R> cleanResults(List<?> functionResults) {
+    List<R> results = new ArrayList<>();
+    for (Object functionResult : functionResults) {
+      if (functionResult == null) {
+        continue;
+      }
+      if (functionResult instanceof Throwable) {
+        // log the exception and continue
+        logger.info("Error executing CacheRealizationFunction.", (Throwable) functionResult);
```
jchen21, line 666: A `warn` level maybe more appropriate than `info` level for Throwable. Some more information in the log might help the user to diagnose in case of an error. e.g. Function name etc. I would probably consider logging the stack trace.

jchen21 (approved):

kirklund (approved):
I like the idea of changing the log-level flagged by Jianxia from info to warn.

PR #5601, opened 2020-10-07
GEODE-8603: Potentially expand classes identified for CI stressing to include subclasses

jdeppe-pivotal:
- Expand any classes targeted by StressNewTest to also include sub classes.
  This is so that we don't miss coverage if an abstract class, containing tests, is really
  executed through a concrete class which did not also change.


geode-junit/src/main/java/org/apache/geode/test/util/ClassScanner.java
```diff
@@ -0,0 +1,98 @@
...
+  public List<String> whatExtends(String classOrFile) {
+    Set<String> classesToConsider = new HashSet<>();
+    classesToConsider.add(pathToClass(classOrFile));
+
+    if (!classOrFile.contains(".")) {
+      classesToConsider.addAll(fullyQualifyClass(classOrFile));
+    }
+
+    return classesToConsider.stream()
+        .flatMap(x -> scanResult.getSubclasses(x).stream().map(ClassInfo::getSimpleName))
```
upthewaterspout, line 53: Is this possibly missing running concrete classes that also have subclasses? Like someone extended a real test with more subclasses?

upthewaterspout (approved):
Conceptually looks fine! I had one question, inline.

ci/scripts/repeat-new-tests.sh
```diff
@@ -46,26 +46,56 @@ function changes_for_path() {
...
-INTEGRATION_TEST_CHANGES=$(changes_for_path '*/src/integrationTest/java') || exit $?
-DISTRIBUTED_TEST_CHANGES=$(changes_for_path '*/src/distributedTest/java') || exit $?
-ACCEPTANCE_TEST_CHANGES=$(changes_for_path '*/src/acceptanceTest/java') || exit $?
-UPGRADE_TEST_CHANGES=$(changes_for_path '*/src/upgradeTest/java') || exit $?
+function save_classpath() {
+  echo "Building and saving classpath"
+  pushd geode >> /dev/null
+    # Do this twice since devBuild still dumps a warning string to stdout.
+    ./gradlew --console=plain -q devBuild 2>/dev/null
+    ./gradlew --console=plain -q printTestClasspath 2>/dev/null >/tmp/classpath.txt
```
rhoughton-pivot, line 54: What is the behavior if devBuild has not been run first?

geode-junit/src/main/java/org/apache/geode/test/util/ClassScanner.java
jdeppe-pivotal: In that case the selected subclass would end up executing tests in the super. A test could be missed if the subclass overrides a test in the super.

ci/scripts/repeat-new-tests.sh
jdeppe-pivotal: No code is compiled which is needed for the subsequent `StressNewTestHelper` to be executed.

jdeppe-pivotal:
After some rework I ended up having the java code take on building the full gradle test task command line. This is because a subclass does not necessarily exist within the same source set than its superclass. This situation then requires executing the subclass under a different gradle task.

rhoughton-pivot:
I wish that the creation of what to run was encapsulated entirely in build with less `bash` needed, but this seems like a good step in the right direction. Eventually, I think the StressNewTestHelper can move to `buildSrc` as a plugin, feeding to the `repeatTest` task for auto-determining what to run.

rhoughton-pivot (approved):

rhoughton-pivot (approved):

onichols-pivotal (approved):

geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/pubsub/AbstractPubSubIntegrationTest.java
```diff
@@ -51,6 +51,7 @@
 import org.apache.geode.test.junit.rules.ExecutorServiceRule;
 
 public abstract class AbstractPubSubIntegrationTest implements RedisPortSupplier {
+  /* Trigger a stress test run */
```
onichols-pivotal, line 54: was this intended to be left in?

jdeppe-pivotal: Thanks for pointing that out - I'll remove it.

PR #5605, opened 2020-10-07
GEODE-8584: Message transmission fails with IllegalStateException in socket i/o code

bschuchardt:
Add appropriate synchronization when using ioFilter's buffers.

Originally NioFilters were accessed only by a single thread at a time.  Eventually this changed, with Connection.asyncClose now invoking the close() method on the filter.  In the NioSslEngine this modifies the encrypted buffer, and that conflicted with the use of the buffer in Connection.writeFully().

Rather than just add additional synchronization to Connection.writeFully() I've examined all uses of NioSslEngine's buffers and have added synchronization around all of these uses.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/net/NioSslEngine.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/net/NioSslEngine.java
@@ -40,6 +40,7 @@ import javax.net.ssl.SSLSession;
 import org.apache.logging.log4j.Logger;
 
 import org.apache.geode.GemFireIOException;
+import org.apache.geode.annotations.internal.MakeImmutable;
 import org.apache.geode.internal.net.BufferPool.BufferType;
 import org.apache.geode.logging.internal.log4j.api.LogService;
 
@@ -52,6 +53,9 @@ import org.apache.geode.logging.internal.log4j.api.LogService;
 public class NioSslEngine implements NioFilter {
   private static final Logger logger = LogService.getLogger();
 
+  // this variable requires the MakeImmutable annotation but the buffer is empty and
+  // not really modifiable
+  @MakeImmutable
   private static final ByteBuffer EMPTY_BUFFER = ByteBuffer.allocate(0);
 
   private final BufferPool bufferPool;
--- a/geode-core/src/main/java/org/apache/geode/internal/net/NioSslEngine.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/net/NioSslEngine.java
@@ -404,9 +404,9 @@ public class NioSslEngine implements NioFilter {
     } finally {
       logger.info("NioSSLEngine releasing two buffers myNetData={}({}), peerAppData={}({})",
           Integer.toHexString(System.identityHashCode(myNetData)),
-          (myNetData.isDirect()? "direct":"heap"),
+          (myNetData.isDirect() ? "direct" : "heap"),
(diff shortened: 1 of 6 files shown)


geode-core/src/main/java/org/apache/geode/internal/tcp/Connection.java
```diff
@@ -2588,17 +2592,20 @@ void writeFully(SocketChannel channel, ByteBuffer buffer, boolean forceAsync,
...
-          } finally {
-            stats.endSocketWrite(true, start, amtWritten, 0);
+        // synchronize on the ioFilter while using its network buffer
+        synchronized (ioFilter) {
+          ByteBuffer wrappedBuffer = ioFilter.wrap(buffer);
+          while (wrappedBuffer.remaining() > 0) {
+            int amtWritten = 0;
+            long start = stats.startSocketWrite(true);
+            try {
+              amtWritten = channel.write(wrappedBuffer);
```
agingade, line 2604: Is there anything that needs to be done, if write throws exception...Sorry I may be asking wrong question as I don't have much context here.

geode-core/src/main/java/org/apache/geode/internal/net/NioSslEngine.java
```diff
@@ -396,9 +402,17 @@ public synchronized void close(SocketChannel socketChannel) {
     } catch (IOException e) {
       throw new GemFireIOException("exception closing SSL session", e);
     } finally {
-      bufferPool.releaseBuffer(TRACKED_SENDER, myNetData);
-      bufferPool.releaseBuffer(TRACKED_RECEIVER, peerAppData);
+      logger.info("NioSSLEngine releasing two buffers myNetData={}({}), peerAppData={}({})",
```
agingade, line 405: This should be info or debug level log? just checking...

bschuchardt: thanks for catching that Anil - that's debug output that I meant to delete

geode-core/src/main/java/org/apache/geode/internal/tcp/Connection.java
bschuchardt: handlers for declared exceptions are already in place

geode-core/src/main/java/org/apache/geode/internal/net/NioSslEngine.java
```diff
@@ -52,6 +53,11 @@
 public class NioSslEngine implements NioFilter {
   private static final Logger logger = LogService.getLogger();
 
+  // this variable requires the MakeImmutable annotation but the buffer is empty and
+  // not really modifiable
+  @MakeImmutable
+  private static final ByteBuffer EMPTY_BUFFER = ByteBuffer.allocate(0);
```
dschneider-pivotal, line 59: did you consider adding ".asReadOnlyBuffer()" so that this would be immutable?

dschneider-pivotal (approved):

geode-core/src/main/java/org/apache/geode/internal/net/NioSslEngine.java
Bill: That won't satisfy PMD (since it returns a `ByteBuffer` which is not considered an immutable object by PMD) but it still might be valuable.

PR #5618, opened 2020-10-12
Improve project resolution of nested subprojects for jar manifests

rhoughton-pivot:


gradle/publish-java.gradle
```diff
@@ -34,40 +34,46 @@ gradle.taskGraph.whenReady({ graph ->
...
         if (projectDependencies.contains(geodeProject)) {
           try {
-            def parentProject = project(":$geodeProject" - "-$version")
+            def parentProject = geodeProject
             def collect = parentProject.configurations.runtimeClasspath.collect { it.name }
             runtimeList.removeAll(collect)
-            projectDependencies.removeAll(collect.collect {it-".jar"})
+//            projectDependencies.removeAll(collect.collect {it-".jar"})
           } catch (UnknownProjectException ignore) {
+            throw ignore
```
kohlmu-pivotal, line 69: If we throw the exception again, without any action, why not take out the try/catch block?

pivotal-jbarrett: I think that exception isn't possible there anyway. Pull it.

gradle/publish-java.gradle
```diff
@@ -34,40 +34,46 @@ gradle.taskGraph.whenReady({ graph ->
...
       // within it's parent's project dependencies.
       projectDependencies.clone().each { projectDependency ->
-        def geodeProject = projectDependency - "-${version}.jar"
+        def geodeProject = projectDependency.getDependencyProject()
         if (projectDependencies.contains(geodeProject)) {
           try {
-            def parentProject = project(":$geodeProject" - "-$version")
+            def parentProject = geodeProject
             def collect = parentProject.configurations.runtimeClasspath.collect { it.name }
             runtimeList.removeAll(collect)
```
pivotal-jbarrett, line 66: I really think we don't want to be removing these dependencies. If we discovered them in the block above as direct dependencies they should be included. Because we aren't iterating over transitive dependencies we shouldn't have any to filter out.

pivotal-jbarrett (changes requested):

gradle/publish-java.gradle
rhoughton-pivot: Because I forgot, while we were pairing.

gradle/publish-java.gradle
```diff
@@ -34,40 +34,48 @@ gradle.taskGraph.whenReady({ graph ->
...
+        allProjectDeps.addAll(project.configurations.getByName(conf).getDependencies())
+      }
+
+//      // Iterate over runtime classpath dependencies and separate project dependencies from library
+//      // dependencies.
+      allProjectDeps.each { dependency ->
+        if ( dependency instanceof ProjectDependency )  {
+          def platformAttribute = Attribute.of("org.gradle.category", org.gradle.api.attributes.Category.class)
+          def foundAttribute = dependency.attributes.getAttribute(platformAttribute)
+          if (! foundAttribute || 'platform'.equals(foundAttribute)) {
```
pivotal-jbarrett, line 50: Order of operations is not clear here. Is it:
```
(!(foundArttribute || 'platform'.equals(foundAttribute)))
```
or:
```
((!foundArttribute) || 'platform'.equals(foundAttribute))
```

gradle/publish-java.gradle
```diff
@@ -34,40 +34,48 @@ gradle.taskGraph.whenReady({ graph ->
...
         if (projectDependencies.contains(geodeProject)) {
-          try {
-            def parentProject = project(":$geodeProject" - "-$version")
+            def parentProject = geodeProject
             def collect = parentProject.configurations.runtimeClasspath.collect { it.name }
             runtimeList.removeAll(collect)
-            projectDependencies.removeAll(collect.collect {it-".jar"})
-          } catch (UnknownProjectException ignore) {
-          }
+//            projectDependencies.removeAll(collect.collect {it-".jar"})
```
pivotal-jbarrett, line 72: Dead code?

pivotal-jbarrett (changes requested):

gradle/publish-java.gradle
rhoughton-pivot: @kohlmu-pivotal Can we get consensus on these remove statements before converting this draft to a final PR?

gradle/publish-java.gradle
rhoughton-pivot: yes

gradle/publish-java.gradle
kohlmu-pivotal: I believe this to be correct for the state of the project at this stage. Until libraries are loaded as modules and ensures that there is ever only one copy of a class at a particular version, this should stay.

We can always change in the future, but for now, this is sufficient.

kohlmu-pivotal (approved):

rhoughton-pivot:
@pivotal-jbarrett Is that logic better?

rhoughton-pivot:
Creating non-draft PR -> https://github.com/apache/geode/pull/5660

PR #5622, opened 2020-10-13
GEODE-7845: Adding a cleaner simpler test.

mhansonp:
Added a cleaner test class that used ClusterStartupRule to make it easier to understand and less cumbersome


geode-core/src/upgradeTest/java/org/apache/geode/internal/cache/rollingupgrade/RollingUpgradePartitionRegionClearServerVersionMismatch.java
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  /**
+   * testServer_ServerVersionMismatchException - validates that when a partitioned region clear is
+   * invoked on a cluster where one server is running an unsupported version for this feature we
+   * return a ServerVersionMismatchException
+   */
+  @Test
+  public void testServer_ServerVersionMismatchException() {
+    IgnoredException.addIgnoredException(ServerOperationException.class);
```
jinmeiliao, line 156: every test will shut down all the locator/servers and restart them again, making the tests taking longer. If you would like make the tests run faster, you can lump these two tests into one. first all region.clear vm and then in a server vm. And would be nice to verify that after these two clears, the region still has those 2 records in them (clear didn't go through on any servers).

mhansonp: Sure, I agree with the record check, but I prefer to have my tests test one thing at time. I realize there is speed efficiency to be had, but I understand that depends on the type of efficiency you are looking for. I am looking at what the test tells me. It certainly feels weird to me though.

jinmeiliao (approved):

PR #5623, opened 2020-10-14
GEODE-8607: Adding some test API to get oldest tombstone

mhansonp:
- Need an API to tell how many how long until the oldest tombstone expires.

### DIFF
--- a/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/versions/TombstoneDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/versions/TombstoneDUnitTest.java
@@ -15,12 +15,14 @@
 package org.apache.geode.internal.cache.versions;
 
 import static org.apache.geode.test.awaitility.GeodeAwaitility.await;
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.Assert.assertEquals;
 
 import java.io.Serializable;
 import java.util.Properties;
 import java.util.concurrent.CountDownLatch;
 
+import org.apache.logging.log4j.Logger;
 import org.junit.Test;
 
 import org.apache.geode.cache.Region;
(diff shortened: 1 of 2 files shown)


geode-core/src/main/java/org/apache/geode/internal/cache/TombstoneService.java
```diff
@@ -450,6 +450,11 @@ boolean testHook_forceExpiredTombstoneGC(int count, long timeout, TimeUnit unit)
 
     @Override
     protected void beforeSleepChecks() {}
+
+    @Override
+    public long getOldestTombstoneTimeDelta() {
+      return tombstones.peek().getVersionTimeStamp() - cacheTime.cacheTimeMillis();
```
agingade, line 456: The peek could return null...
Will it be a good enough to return version timestamp instead of delta...The caller can calculate delta with different time mechanism (using cacheTime or currentTime).

Also, will it be helpful if the tombstone.toString() info is returned...It has all the information associated with tombstone.

mhansonp: I figured that it was one extra step that will always be done, but I will change it.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/versions/TombstoneDUnitTest.java
```diff
@@ -15,127 +15,226 @@
...
+      assertEquals(0, ((InternalCache) cache).getCachePerfStats().getTombstoneCount());
     });
   }
 
   @Test
-  public void testTombstonesWithLowerVersionThanTheRecordedVersionGetsGCed() throws Exception {
-    Host host = Host.getHost(0);
-    VM vm0 = host.getVM(0);
-    VM vm1 = host.getVM(1);
+  public void TestGetOldestTombstoneTimeReplicate() {
```
kirklund, line 110: Change the test name to start with lower-case `test`?

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/versions/TombstoneDUnitTest.java
```diff
@@ -15,127 +15,226 @@
...
+      assertThat(tombstoneSweeper.getOldestTombstoneTime()).isEqualTo(0);
+    });
+  }
+
+  @Test
+  public void testTombstonesWithLowerVersionThanTheRecordedVersionGetsGCed() throws Exception {
+    VM vm0 = VM.getVM(0);
+    VM vm1 = VM.getVM(1);
+    Properties props = DistributedRule.getDistributedSystemProperties();
+    props.put("conserve-sockets", "false");
```
kirklund, line 176: `props.setProperty` is more correct than using the Map interface method `put`.

kirklund (approved):

PR #5627, opened 2020-10-14
GEODE-8585: Redis SCAN, HSCAN, and SSCAN do not detect illegal parameters

sabbey37:
Add integration tests and functionality to match native Redis.

### DIFF
--- a/geode-redis/src/main/java/org/apache/geode/redis/internal/executor/key/ScanExecutor.java
+++ b/geode-redis/src/main/java/org/apache/geode/redis/internal/executor/key/ScanExecutor.java
@@ -15,15 +15,21 @@
  */
 package org.apache.geode.redis.internal.executor.key;
 
-import static org.apache.geode.redis.internal.RedisConstants.ERROR_ILLEGAL_GLOB;
+import static org.apache.geode.redis.internal.RedisConstants.ERROR_CURSOR;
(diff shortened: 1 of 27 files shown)


geode-redis/src/main/java/org/apache/geode/redis/internal/executor/key/ScanExecutor.java
```diff
@@ -36,91 +39,87 @@
...
         cursor);
+    context.setScanCursor(new BigInteger((String) returnList.get(0)));
 
     return RedisResponse.scan(returnList);
   }
 
-  private List<String> getIteration(Collection<ByteArrayWrapper> list, Pattern matchPattern,
-      int count, int cursor) {
-    List<String> returnList = new ArrayList<>();
+  private List<Object> scan(Collection<ByteArrayWrapper> list, Pattern matchPattern,
```
jdeppe-pivotal, line 102: I think it would be better to have this return a `Pair<BigInteger, List<Object>>` - that way it's not hiding the fact that the zero-th element is special.

sabbey37: Good idea! I just changed implemented it. Let me know what you think.

jdeppe-pivotal (approved):
Looks good!

PR #5631, opened 2020-10-15
GEODE-8584: avoid locking when closing connection

Bill:
The initial fix for [GEODE-8584](https://issues.apache.org/jira/browse/GEODE-8584) has a hang when `Connection.notifyHandshakeWaiter()` tries to acquire the monitor on the `ByteBuffer` but that monitor is already held by another thread that is trying to read an ack.

The fix here is to simply not try to acquire the monitor if we are closing the connection.


Bill:
this was a quick-and-dirty attempt to fix what became known as GEODE-8652. See that ticket and its final PR for the real solution.

PR #5638, opened 2020-10-18
GEODE-8584: AutoCloseable for ByteBuffer Locking

Bill:
A third attempt at fixing GEODE-8584 locking granularity.

In this PR we have fine-grained locking (read has its own lock, write has its own lock, and `NioSslEngine.close()` has a completely separate lock). Also in the place where we acquire the write lock in `close()` we have a finite wait time (1 minute.) so we don't prevent `Connection` from closing the socket.

This is a refinement of the [Finer-Grained Locking PR](https://github.com/apache/geode/pull/5632). That PR split the single lock into two (but does not have the timeout on write lock acquisition that this PR has.)  With that other PR, client code in e.g. .`MsgReader` had to do all this ceremony:

```
synchronized (ioFilter.getInputSyncObject()) {
  ioFilter.setInputInUse(true);
  try {
    ByteBuffer unwrappedBuffer = readAtLeast(Connection.MSG_HEADER_BYTES);

    // do stuff with unwrappedBuffer

  } finally {
	ioFilter.setInputInUse(false);
  }
}
```

@dschneider-pivotal noticed that this followed the `AutoCloseable` pattern. So this PR makes good on that observation and introduces the auto-closeable `ByteBufferSharing`. We now return a `ByteBufferSharing` from methods that used to return the input `ByteBuffer`. Callers acquire it in a try-with-resources and rely on the `AutoCloseable` protocol to take care of lock closing and reference counting (decrement).

With this PR that same `MsgReader` code looks like this:

```
try (final ByteBufferSharing sharedBuffer = readAtLeast(Connection.MSG_HEADER_BYTES)) {
  ByteBuffer unwrappedBuffer = sharedBuffer.getBuffer();

  // do stuff with unwrappedBuffer
  
}
```      

The value of this approach goes beyond mere concision. It helped me find places where we were failing to properly follow the locking and ref-counting protocol in the previous PR. Now the methods that require the special treatment are much more self-documenting. There is no danger of failing to lock now. The only risk is a failure to acquire the resource in a try-with-resources, resulting in a failure to release the lock and release the reference.

# TODO

- [ ] in `NioSslEngine` use `Lock` for `myNetData` (output) sync (shall we reuse `ByteBufferSharing`+`ByteBufferReferencing` for this?)
- [ ] change `close()` sync to time out if `Lock` can't be acquired so that `close()` is no longer blocked on readers

# Boilerplate TODO

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/net/ByteBufferSharingImpl.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/net/ByteBufferSharingImpl.java
@@ -1,0 +1,101 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package org.apache.geode.internal.net;
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+
+import org.apache.logging.log4j.Logger;
+
+import org.apache.geode.logging.internal.log4j.api.LogService;
+
+/**
+ * An {@link AutoCloseable} meant to be acquired in a try-with-resources statement. The resource (a
+ * {@link ByteBuffer}) is available (for reading and modification) in the scope of the
+ * try-with-resources. Collaborates with {@link ByteBufferReferencing} to ensure final dereference
+ * returns the {@link ByteBuffer} to the pool.
+ */
+class ByteBufferSharingImpl implements ByteBufferSharing {
+  private static final Logger logger = LogService.getLogger();
+
+  static class LockAttemptTimedOut extends Exception {
+  }
+
+  private final Lock lock;
+  private final ByteBufferReferencing referencing;
+
+  /**
+   * This constructor is for use only by the owner of the shared resource (a {@link ByteBuffer}).
+   *
+   * A resource owner must invoke {@link #alias()} once for each reference that escapes (is passed
+   * to an external object or is returned to an external caller.)
+   *
+   * This constructor acquires no lock and does not modify the reference count through the {@link
+   * ByteBufferReferencing} (passed in here.) Usually, the referencing object will have been set to
+   * a reference of 1 before calling this method.
+   */
+  ByteBufferSharingImpl(final ByteBufferReferencing referencing) {
+    this(new ReentrantLock(), referencing);
+  }
+
+  /**
+   * This method is for use only by the owner of the shared resource. It's used for handing out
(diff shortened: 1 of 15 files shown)


geode-core/src/main/java/org/apache/geode/internal/net/ByteBufferSharingImpl.java
```diff
@@ -0,0 +1,91 @@
...
+  @Override
+  public void close() {
+    final int usages = referencing.dropReference();
+    if (usages > 0) {
+      /*
+       * We don't unlock when the very last usage goes away. The resource owner holds the first
+       * usage and there is no lock associated with that one. Subsequent usages have a lock and
+       * so require a corresponding unlock.
+       */
+      lock.unlock();
```
dschneider-pivotal, line 87: I think this introduces a bug in the case when "close" finds the reference still in use. In that case "close" will call unlock but that will fail because ReentrantLock requires that unlock be called by the same thread that called lock (I think). In general it wold be better if the thread that called lock at the beginning of the "try-with-resource" was also the caller of "unlock"

geode-core/src/main/java/org/apache/geode/internal/net/ByteBufferSharingImpl.java
```diff
@@ -0,0 +1,91 @@
...
+
+  /**
+   * This constructor is for use only by the owner of the shared resource. It's used for handing out
+   * references to the shared resource. So it does reference counting and also acquires a lock.
+   *
+   * Resource owners will typically return the result of this constructor, to a caller. That caller
+   * binds that reference to a variable in a try-with-resources statement and relies on the
+   * AutoCloseable protocol to invoke close() on the object at the end of the block.
+   */
+  ByteBufferSharingImpl(final ByteBufferSharingImpl other) {
```
dschneider-pivotal, line 59: Always creating a new instance at the start of try-with-resources is not needed. In the case of this object I think it would be okay for the same instance to be used by multiple threads concurrently and each one to call close concurrently on the same instance. If so this would address Jakes concern about object allocation.

geode-core/src/main/java/org/apache/geode/internal/net/ByteBufferSharingImpl.java
```diff
@@ -0,0 +1,91 @@
...
+
+/**
+ * An {@link AutoCloseable} meant to be acquired in a try-with-resources statement. The resource (a
+ * {@link ByteBuffer}) is available (for reading and modification) in the scope of the
+ * try-with-resources. Collaborates with {@link ByteBufferReferencing} to ensure final dereference
+ * returns the {@link ByteBuffer} to the pool.
+ */
+class ByteBufferSharingImpl implements ByteBufferSharing {
+
+  private final ByteBuffer buffer;
```
dschneider-pivotal, line 30: I think it would be better to obtain the buffer from "referencing" instead of having an extra copy of it here.

geode-core/src/main/java/org/apache/geode/internal/net/ByteBufferSharingImpl.java
Bill: Talked to @dschneider-pivotal and we agreed this is not an issue.

geode-core/src/main/java/org/apache/geode/internal/net/ByteBufferSharingImpl.java
Bill: This is addressed in the latest commit. This copy-constructor was replaced by this method:

```
ByteBuffer alias()
```

The resource-owner returns the result of invoking `alias()` everywhere it hands out an alias. No more construction!

geode-core/src/main/java/org/apache/geode/internal/net/ByteBufferSharingImpl.java
Bill: This is addressed in the latest commit. Now the "referencing" owns the `ByteBuffer` referencing entirely!

Bill:
These ideas made it into the first and second PRs for GEODE-8652 and were ported everywhere.

PR #5639, opened 2020-10-19
GEODE-8627: Redis not unsubscribing and punsubscribing correctly when no channel/pattern provided

sabbey37:
Geode Redis should unsubscribe from all channels (not including patterns) when no channel is provided and punsubscribe from all patterns (not including channels) when no pattern is provided.

### DIFF
--- a/geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/pubsub/AbstractSubscriptionsIntegrationTest.java
+++ b/geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/pubsub/AbstractSubscriptionsIntegrationTest.java
@@ -19,6 +19,11 @@ package org.apache.geode.redis.internal.executor.pubsub;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
+import java.util.List;
+import java.util.stream.Collectors;
+
+import org.junit.After;
+import org.junit.Before;
 import org.junit.ClassRule;
 import org.junit.Test;
 import redis.clients.jedis.Jedis;
@@ -31,12 +36,25 @@ import org.apache.geode.test.junit.rules.ExecutorServiceRule;
 
 public abstract class AbstractSubscriptionsIntegrationTest implements RedisPortSupplier {
 
+  public static final int REDIS_CLIENT_TIMEOUT =
+      Math.toIntExact(GeodeAwaitility.getTimeout().toMillis());
+  private Jedis client;
+
   @ClassRule
   public static ExecutorServiceRule executor = new ExecutorServiceRule();
 
+  @Before
+  public void setUp() {
+    client = new Jedis("localhost", getPort(), REDIS_CLIENT_TIMEOUT);
+  }
+
+  @After
+  public void tearDown() {
+    client.close();
+  }
+
   @Test
   public void pingWhileSubscribed() {
-    Jedis client = new Jedis("localhost", getPort());
     MockSubscriber mockSubscriber = new MockSubscriber();
 
     executor.submit(() -> client.subscribe(mockSubscriber, "same"));
@@ -46,12 +64,10 @@ public abstract class AbstractSubscriptionsIntegrationTest implements RedisPortS
         .untilAsserted(() -> assertThat(mockSubscriber.getReceivedPings().size()).isEqualTo(1));
     assertThat(mockSubscriber.getReceivedPings().get(0)).isEqualTo("");
     mockSubscriber.unsubscribe();
-    client.close();
   }
 
   @Test
   public void pingWithArgumentWhileSubscribed() {
-    Jedis client = new Jedis("localhost", getPort());
     MockSubscriber mockSubscriber = new MockSubscriber();
 
     executor.submit(() -> client.subscribe(mockSubscriber, "same"));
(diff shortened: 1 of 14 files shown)


geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/pubsub/AbstractSubscriptionsIntegrationTest.java
```diff
@@ -88,18 +111,91 @@ public void multiSubscribe() {
         .untilAsserted(() -> assertThat(mockSubscriber.getReceivedPMessages()).hasSize(1));
     assertThat(mockSubscriber.getReceivedEvents()).containsExactly("message", "pmessage");
     mockSubscriber.unsubscribe();
-    client.close();
   }
 
   @Test
-  public void unallowedCommandsWhileSubscribed() {
-    Jedis client = new Jedis("localhost", getPort());
+  public void unsubscribingImplicitlyFromAllChannels_doesNotUnsubscribeFromPatterns() {
```
sabbey37, line 117: It might make sense to move these test to `AbstractPubSubIntegrationTest.java` once [8577](https://issues.apache.org/jira/projects/GEODE/issues/GEODE-8577?filter=allopenissues) is resolved.

jdeppe-pivotal: Do we not already have similar tests in `AbstractPubSubIntegrationTest`? Perhaps `testUnsubscribingImplicitlyFromAllChannels` and `testPunsubscribingImplicitlyFromAllChannels`.

geode-redis/src/main/java/org/apache/geode/redis/internal/pubsub/Subscription.java
```diff
@@ -53,11 +53,15 @@ void publishMessage(byte[] channel, byte[] message,
...
-   * Return the subscription name. In the case of a pattern the string representation of the
-   * pattern is returned.
+   * Return the channel name. In the case of a pattern null is returned.
    */
   byte[] getChannelName();
 
+  /**
+   * Return the pattern name. In the case of a channel null is returned.
+   */
+  byte[] getPatternName();
```
jdeppe-pivotal, line 63: I can see that you want the method name to be more expressive, but I don't think it's good practice to have an API where some methods return null depending on what type of object you're using. That opens the door for a dev to easily make a mistake. I'd prefer to keep a single method but give it a more descriptive name. Perhaps `getSubscriptionName()` or `getChannelOrPattern()`.

jdeppe-pivotal (changes requested):

geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/pubsub/AbstractSubscriptionsIntegrationTest.java
jdeppe-pivotal: Do we not already have similar tests in `AbstractPubSubIntegrationTest`? Perhaps `testUnsubscribingImplicitlyFromAllChannels` and `testPunsubscribingImplicitlyFromAllChannels`

sabbey37: We do have a test that only subscribes to channels and then unsubscribes without an argument. We did not have a test that subscribes to channels and psubscribes to patterns, then unsubscribes without an argument (or punsubscribes without an argument).  We were incorrectly unsubscribing from everything (channels and patterns) when punsubscribe or unsubscribe were called without an argument.  Native Redis will only unsubscribe from channels (not patterns) if unsubscribe is called without an argument or punsubscribe only from patterns (not channels) if punsubscribe is called without an argument.

geode-redis/src/main/java/org/apache/geode/redis/internal/pubsub/Subscription.java
sabbey37: I'm definitely open to suggestions on this, maybe we could have another method that's called `getChannelOrPattern()` or some other way of distinguishing whether a channel/pattern name was returned?

jdeppe-pivotal: I see what the problem is now. I think it would be better to provide a `Type` enum into the `Subscription` interface:
```
  enum Type {
    CHANNEL,
    PATTERN;
  }

  Type getType();
```

And then you should be able to filter by that when you need to have methods distinguish the type of subscription.

In the `PubSub` interface you can add:
```
List<byte[]> findSubscribedChannels(Client client, Subscription.Type type);
```

sabbey37: Good idea, I just pushed a change. Let me know what you think!

geode-redis/src/main/java/org/apache/geode/redis/internal/pubsub/PubSub.java
```diff
@@ -81,19 +81,19 @@ long publish(
...
   /**
-   * Return a list of pattern names that a client has subscribed to
+   * Return a list of channel names and patterns that a client has subscribed to
    *
    * @param client the Client which is to be queried
-   * @return the list of patterns
+   * @return the list of channels and patterns
    */
-  List<byte[]> findSubscribedPatterns(Client client);
+  List<byte[]> findSubscriptionsNames(Client client);;
```
jdeppe-pivotal, line 97: Only if you need to push another change: I think this should be `findSubscriptionNames` and the double `;` at the end of the line. No biggy.

geode-redis/src/main/java/org/apache/geode/redis/internal/pubsub/PubSubImpl.java
```diff
@@ -140,18 +139,17 @@ public long punsubscribe(GlobPattern pattern, Client client) {
...
         .collect(Collectors.toList());
   }
 
   @Override
-  public List<byte[]> findSubscribedPatterns(Client client) {
+  public List<byte[]> findSubscriptionsNames(Client client, Subscription.Type type) {
     return subscriptions.findSubscriptions(client).stream()
-        .map(Subscription::getPatternName)
-        .filter(Objects::nonNull)
+        .filter(s -> s.getType().equals(type))
```
jdeppe-pivotal, line 151: Also no biggy, but for enum equality it's slightly better to use `==` since it's also null safe.

jdeppe-pivotal (approved):

geode-redis/src/main/java/org/apache/geode/redis/internal/pubsub/PubSub.java
sabbey37: Yes!  Thank you for pointing this out!

geode-redis/src/main/java/org/apache/geode/redis/internal/pubsub/PubSubImpl.java
sabbey37: Good to know! I'll update it.

PR #5645, opened 2020-10-20
GEODE-7845: Now behaving with clients of various versions.

mhansonp:


geode-core/src/upgradeTest/java/org/apache/geode/internal/cache/rollingupgrade/RollingUpgradePartitionRegionClearServerVersionMismatch.java
```diff
@@ -109,20 +111,28 @@ public void before() {
...
     });
+  }
+
+  @After
+  public void after() {
+    locator.stop();
+    serverNew.stop();
+    serverOld.stop();
 
   }
```
gesterzhou, line 123: Change all the ServerOperationException  to be UnsupportedOperationException in this class, including comments and method names.

geode-core/src/upgradeTest/java/org/apache/geode/internal/cache/rollingupgrade/RollingUpgradePartitionRegionClearServerVersionMismatch.java
```diff
@@ -109,20 +111,28 @@ public void before() {
...
   @Test
-  public void testClient_ServerVersionMismatchException() throws Exception {
+  public void testClient_ServerOperationExceptionCurrentServerVersion() throws Exception {
     IgnoredException.addIgnoredException(ServerOperationException.class);
-    final int locatorPort = locator.getPort();
+
     // Get a client VM
-    ClientVM clientVM = cluster.startClientVM(3, c -> c.withLocatorConnection(locatorPort));
+    int serverPort = serverNew.getPort();
+    clientVM = cluster.startClientVM(3, oldVersion, c -> c.withServerConnection(serverPort));
```
gesterzhou, line 135: it's better to have one test to run pr clear from an old version and another test to run pr clear from a new version. Two test methods.

mhansonp: Not sure I follow here, I have it broken out into two different client side tests. client -> old server, client -> new server

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionClear.java
```diff
@@ -349,13 +348,14 @@ public void allServerVersionsSupportPartitionRegionClear() {
...
           memberNames.add(internalDistributedMember.getName());
-          logger.info("MLH adding " + internalDistributedMember.getName());
         }
       }
     }
     if (!memberNames.isEmpty()) {
-      throw new ServerVersionMismatchException(memberNames, "Partitioned Region Clear",
-          KnownVersion.GEODE_1_14_0.toString());
+      throw new UnsupportedOperationException(
+          "A server's " + memberNames + " version was too old (< "
```
kirklund, line 356: This message reads weird. Maybe change to:
```
"Server(s) " + memberNames + " are running a version older than (<" + KnownVersion.GEODE_1_14_0 + ">) which does not support partitioned region clear.");
```

geode-core/src/upgradeTest/java/org/apache/geode/internal/cache/rollingupgrade/RollingUpgradePartitionRegionClearServerVersionMismatch.java
kirklund: I think Gester is recommending expanding the testing to include:

1. new client --> new server
2. new client --> old server
3. old client --> new server
4. old client --> old server

If (3) and (4) throw UnsupportedOperationException without sending any message to the server, then they could probably be collapsed into a single test.

kirklund (approved):

gesterzhou (approved):

mhansonp:
The test requirement I think is met. All clients will talk to the new server, then all clients will talk to the old server.

PR #5654, opened 2020-10-22
GEODE-8662: add integration tests for strlen

nonbinaryprogrammer:

### For all changes:

### Note:

### DIFF
--- a/geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/string/AbstractIncrIntegrationTest.java
+++ b/geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/string/AbstractIncrIntegrationTest.java
@@ -104,7 +104,7 @@ public abstract class AbstractIncrIntegrationTest implements RedisPortSupplier {
     try {
       jedis.incr(key);
     } catch (JedisDataException e) {
-      assertThat(e.getMessage()).contains(RedisConstants.ERROR_NOT_INTEGER);
+      assertThat(e.getMessage()).contains("out of range");
     }
     assertThat(jedis.get(key)).isEqualTo(nonIntegerValue);
   }
--- a/geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/string/AbstractIncrIntegrationTest.java
+++ b/geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/string/AbstractIncrIntegrationTest.java
@@ -14,6 +14,8 @@
  */
 package org.apache.geode.redis.internal.executor.string;
 
+import static org.apache.geode.redis.internal.RedisConstants.ERROR_NOT_INTEGER;
+import static org.apache.geode.redis.internal.RedisConstants.ERROR_OVERFLOW;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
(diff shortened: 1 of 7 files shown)


geode-redis/src/main/java/org/apache/geode/redis/internal/RedisConstants.java
```diff
@@ -42,6 +42,7 @@
   public static final String ERROR_WRONG_TYPE =
       "Operation against a key holding the wrong kind of value";
   public static final String ERROR_NOT_INTEGER = "value is not an integer or out of range";
+  public static final String ERROR_NOT_LONG = "value is not a long or out of range";
```
sabbey37, line 45: I don't think Redis ever returns this error.

sabbey37 (changes requested):
Requesting a few changes so we can continue to match errors returned by Redis.  We are trying to match Redis's responses as much as possible.  Also, we need a Geode Jira ticket for this PR.

Thank you for adding all the tests around the RedisString class!  Those are great!

geode-redis/src/test/java/org/apache/geode/redis/internal/data/RedisStringTest.java
```diff
@@ -41,16 +42,228 @@ public static void beforeClass() {
...
+
+  @Test
+  public void decrErrorsWhenOverflows() {
+    // allows unchecked cast of mock to Region<ByteArrayWrapper, RedisData>
+    @SuppressWarnings("unchecked")
+    Region<ByteArrayWrapper, RedisData> region = mock(Region.class);
+    ByteArrayWrapper byteArrayWrapper = new ByteArrayWrapper(new byte[] {0});
+    RedisString string = new RedisString(byteArrayWrapper);
+    assertThatThrownBy(() -> string.decr(region, byteArrayWrapper))
+        .isInstanceOf(NumberFormatException.class);
```
sabbey37, line 219: This seems to be an error due to the inability to parse the value as a long rather than due to overflow.  We could change the test name or make it error due to overflow instead.

geode-redis/src/test/java/org/apache/geode/redis/internal/data/RedisStringTest.java
```diff
@@ -41,16 +42,228 @@ public static void beforeClass() {
...
+  @Test
+  public void decrbyErrorsWhenOverflows() {
+    // allows unchecked cast of mock to Region<ByteArrayWrapper, RedisData>
+    @SuppressWarnings("unchecked")
+    Region<ByteArrayWrapper, RedisData> region = mock(Region.class);
+    ByteArrayWrapper byteArrayWrapper = new ByteArrayWrapper(new byte[] {1});
+    RedisString string = new RedisString(byteArrayWrapper);
+    assertThatThrownBy(() -> string.decrby(region, byteArrayWrapper, 2))
+        .isInstanceOf(NumberFormatException.class);
+  }
```
sabbey37, line 242: This also seems to be an error due to the inability to parse the value as a long rather than due to overflow. We could change the test name or make it error due to overflow instead.

geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/string/AbstractIncrIntegrationTest.java
```diff
@@ -104,7 +104,7 @@ public void testIncr_whenWrongType_shouldReturnError() {
     try {
       jedis.incr(key);
     } catch (JedisDataException e) {
-      assertThat(e.getMessage()).contains(RedisConstants.ERROR_NOT_INTEGER);
+      assertThat(e.getMessage()).contains("out of range");
```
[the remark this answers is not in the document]
dschneider-pivotal: If our messages are out of sync with native redis shouldn't we be seeing our native acceptance tests failing on this branch? It looks like they all passed. Does this indicate something is wrong with our native redis testing?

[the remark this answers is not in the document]
sabbey37: This one was not out of sync until this PR.  The reason it still passes is because it was changed to only check that the error message contains "out of range" rather than the full message "value is not an integer or out of range".  I'm sure there are other messages throughout our code that are not the same as Redis, but we haven't found them yet because the commands are not fully tested.

sabbey37 (approved):

PR #5655, opened 2020-10-22
GEODE-8645: Reduce netty threads used by PubSubIntegrationTest

jdeppe-pivotal:
- Allows this test not to run out of memory when run under StressNewTest

Authored-by: Jens Deppe <jdeppe@vmware.com>

### For all changes:

### Note:

### DIFF
--- a/geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/pubsub/PubSubIntegrationTest.java
+++ b/geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/pubsub/PubSubIntegrationTest.java
@@ -20,6 +20,11 @@ import org.junit.ClassRule;
 import org.apache.geode.redis.GeodeRedisServerRule;
 
 public class PubSubIntegrationTest extends AbstractPubSubIntegrationTest {
+
+  static {
+    System.setProperty("io.netty.eventLoopThreads", "10");
+  }
+
   @ClassRule
   public static GeodeRedisServerRule server = new GeodeRedisServerRule();
 
--- a/geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/pubsub/PubSubIntegrationTest.java
+++ b/geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/pubsub/PubSubIntegrationTest.java
@@ -16,17 +16,19 @@
 package org.apache.geode.redis.internal.executor.pubsub;
 
 import org.junit.ClassRule;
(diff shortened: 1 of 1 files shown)


sabbey37 (approved):

geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/pubsub/PubSubIntegrationTest.java
```diff
@@ -20,6 +20,11 @@
 import org.apache.geode.redis.GeodeRedisServerRule;
 
 public class PubSubIntegrationTest extends AbstractPubSubIntegrationTest {
+
+  static {
+    System.setProperty("io.netty.eventLoopThreads", "10");
+  }
+
```
kohlmu-pivotal, line 27: Why could you not use `RestoreSystemProperties` here? 

This test does not clean up itself after setting the property. What would the effect be if you don't clean up this property to the "default" value?

kohlmu-pivotal (changes requested):

geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/pubsub/PubSubIntegrationTest.java
jdeppe-pivotal: Good call. Should be fixed now.

geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/pubsub/PubSubIntegrationTest.java
```diff
@@ -16,13 +16,20 @@
...
 public class PubSubIntegrationTest extends AbstractPubSubIntegrationTest {
-  @ClassRule
+
   public static GeodeRedisServerRule server = new GeodeRedisServerRule();
 
+  @ClassRule
+  public static RuleChain chain = RuleChain.outerRule(
+      new ProvideSystemProperty("io.netty.eventLoopThreads", "10"))
+      .around(server);
+
```
kohlmu-pivotal, line 32: @kirklund would you be happy with the instead of the `RestoreSystemProperties` and the changing of the properties in a `@BeforeClass` block?

jdeppe-pivotal:
The problem I'm trying to address here appeared to be the test running out of memory. Turns out that it's actually the __*gradle*__ daemon running out of memory because of too much logging being produced by the test. I'm going to open a new PR.

PR #5664, opened 2020-10-23
GEODE-8650: Support multiple instances of DistributedReference

kirklund:
Includes additional improvements for usage of DistributedReference
in tests.

### DIFF
--- a/geode-dunit/src/main/java/org/apache/geode/test/dunit/rules/DistributedReference.java
+++ b/geode-dunit/src/main/java/org/apache/geode/test/dunit/rules/DistributedReference.java
@@ -18,8 +18,6 @@ package org.apache.geode.test.dunit.rules;
 
 import static org.apache.geode.test.dunit.VM.DEFAULT_VM_COUNT;
 import static org.apache.geode.util.internal.CompletionUtils.close;
-import static org.apache.geode.util.internal.CompletionUtils.openLatch;
-import static org.apache.geode.util.internal.CompletionUtils.toFalse;
 import static org.apache.geode.util.internal.UncheckedUtils.uncheckedCast;
 
 import java.lang.reflect.InvocationTargetException;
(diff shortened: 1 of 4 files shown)


geode-dunit/src/main/java/org/apache/geode/test/dunit/rules/DistributedReference.java
```diff
@@ -187,47 +197,76 @@ public V get() {
...
-      invokeMethod(value, "disconnect");
+    } else if (object instanceof CountDownLatch) {
+      openLatch((CountDownLatch) object);
 
-    } else if (hasMethod(value.getClass(), "stop")) {
-      invokeMethod(value, "stop");
-    }
-  }
+    } else if (hasMethod(object.getClass(), "close")) {
+      invokeMethod(object, "close");
```
kohlmu-pivotal, line 263: Could we not leverage method overloading here.
Instead of `toFalse((AtomicBoolean)object)` we use `close(object)` and then have a `close(AtomicBoolean)` in the CompletionUtils?

kirklund: Do we really want a method called `close(AtomicBoolean)` that sets the specified value to false? I started out with multiple `close` methods and then I changed them from that to:
* toFalse(AtomicBoolean)
* openLatch(CountDownLatch)

And then what about opening `CountDownLatch`? Do we change that to `close` as well even tho close is the opposite of open?

I was thinking that my creation of `CompletionUtils` might be slightly more controversial because it's of dubious value outside `DistributedReference`.

kohlmu-pivotal: I imagine that the effort to write `toFalse` would be the same as `close(AtomicBoolean)`. 
I don't mind the `CompletionUtils` class. 
It is test only code anyway. I just thought the code would be cleaner with just a single `close(*)` call. Then all the logic for closing, even all the method look ups, could be contained in the `CompletionUtils` class.

kirklund: New commit includes "Rename all CompletionUtils to close". Please review again. Thanks!

kohlmu-pivotal (approved):

PR #5665, opened 2020-10-23
GEODE-8651: MsgReader's readHeader and readMessage should be synchron…

gesterzhou:
…ized

    Co-authored-by: Darrel Schneider <darrel@vmware.com>
    Co-authored-by: Ernie Burghardt <eburghardt@pivotal.io>

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/test/java/org/apache/geode/internal/tcp/ConnectionTest.java
+++ b/geode-core/src/test/java/org/apache/geode/internal/tcp/ConnectionTest.java
@@ -19,6 +19,7 @@ import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.ArgumentMatchers.isNull;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.spy;
 import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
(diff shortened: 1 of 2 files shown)


kohlmu-pivotal (changes requested):
Do we have any tests that reproduce the behavior that we are trying to fix with this synchronization?

Are there any existing tests that reproduce this error?

If none of the above are true, could you please update ticket with tests that prove that this problem has been resolved.

kohlmu-pivotal:
I would like to also point out that it does not make sense that any of the people that were added to review (except @bschuchardt ) to review this code. As the rest have all be names co-authors, which in reality does sort of defeat the point if they review what they authored

gesterzhou:
We have some tests to verify it. However, we are adding unit test, on the way.

bschuchardt (commented):
I don't think the additional synchronization is a problem, but it looks like the double-notification check would be sufficient to keep the handshake thread from unnecessarily modifying the input buffer a second time.
Would it be simpler to move the buffer-clearing to the readAck method?  Then the handshake thread double-notification wouldn't be a problem.
If you would respond to these questions I'll feel better about approving this PR.

dschneider-pivotal:
@bschuchardt when you said to move the buffer-clearing to readAck did you mean the code in notifyHandshakeWaiter that sets the position and limit to 0 if SSL? If so I don't understand what that has todo with readAck. notifyHandshakeWaiter is called to notify the thread that is stuck in waitForHandshake that it no longer needs to wait. I don't understand why in SSL we need to clear the buffer before doing this (we thought you might know why) but it does not seem to be connected with reading a direct ack which is what readAck does.

dschneider-pivotal:
As far as the extra synchronization goes in readAck, I agree that if the only place that was concurrently modifying this buffer was the extra notifyHandshakeWaiter then this sync is not needed. But from the viewpoint of readAck, it currently calls readHeader and readMessage or readChunk. All these methods end up synchronizing to protect the buffer. If that sync is needed then it is also needed at the readAck level. readAck is incorrect to allow another thread to modify the buffer between the readHeader and readMessage calls. It is hard to prove that no other threads could concurrently modify the buffer which is why I think we added the new ioFilter.getSynchObject. I'm arguing that if we are going to have getSynchObject then readAck should synchronize on it.

bschuchardt (approved):

gesterzhou:
> Do we have any tests that reproduce the behavior that we are trying to fix with this synchronization?
> 
> Are there any existing tests that reproduce this error?
> 
> If none of the above are true, could you please update ticket with tests that prove that this problem has been resolved.

The test is added.

kirklund (approved):

kirklund:
Ooops, I meant "The test has been added"

bschuchardt (approved):

geode-core/src/test/java/org/apache/geode/internal/tcp/ConnectionTest.java
```diff
@@ -113,4 +114,35 @@ public void connectTimeoutIsShortWhenAlerting() throws UnknownHostException {
...
+    when(tcpConduit.getStats()).thenReturn(dmStats);
+
+    SocketChannel channel = SocketChannel.open();
+
+    Connection connection = new Connection(connectionTable, channel.socket());
+    connection = spy(connection);
+    connection.setSharedUnorderedForTest();
+    connection.run();
+
+    verify(connection, times(1)).getConduit();
```
dschneider-pivotal, line 146: I don't like that we verify that getConduit is called 1 time. I understand after looking at the code but it seems rather indirect logically and brittle. I would prefer that you extract the block of code in notifyHandshakeWaiter starting at line 810 into a method named "clearSSLInputBuffer". Then this test can assert that clearSSLInputBuffer is only called 1 time. This would be much clearer and less brittle and would not increase the diffs much

dschneider-pivotal (changes requested):

geode-core/src/test/java/org/apache/geode/internal/tcp/ConnectionTest.java
gesterzhou: fixed.

dschneider-pivotal (approved):

kohlmu-pivotal (approved):

PR #5666, opened 2020-10-24
GEODE-8652: Allow NioSslEngine.close() to Bypass Locks

Bill:
This PR for [GEODE-8652](https://issues.apache.org/jira/browse/GEODE-8652) works around the situation where a thread wants to close a `Connection` while other threads are reading or writing on it.

- NioSslEngine.close() proceeds even if readers (or writers) are
  operating on its ByteBuffers, allowing Connection.close() to close its
  socket and proceed.

- NioSslEngine.close() needed a lock only on the output buffer, so 
  we split what was a single lock into two. Also instead of using 
  synchronized we use a ReentrantLock so we can 
  call tryLock() and time out if needed in NioSslEngine.close()

- Since readers/writers may hold locks on these input/output buffers
   when NioSslEngine.close() is called we needed to reference count
   the buffers and return them to the pool only when the last user 
   was finished.

- To manage the locking and reference counting a new AutoCloseable
   ByteBufferSharing interface is introduced with a trivial implementation:
   ByteBufferSharingNoOp and a real implementation:
   ByteBufferSharingImpl.

Co-authored-by: Ernie Burghardt <burghardte@vmware.com>


geode-core/src/main/java/org/apache/geode/internal/net/SocketCloser.java
```diff
@@ -41,6 +44,7 @@
  * This max threads can be configured using the "p2p.ASYNC_CLOSE_POOL_MAX_THREADS" system property.
  */
 public class SocketCloser {
+  private static final Logger logger = LogService.getLogger();
```
dschneider-pivotal, line 47: no need to modify SocketCloser.java in this pr

bschuchardt (changes requested):
I really like how you've separated the input/output buffer acquisition and added the auto-closable interface.  Very nice work.

geode-core/src/main/java/org/apache/geode/internal/net/ByteBufferSharingImpl.java
```diff
@@ -0,0 +1,142 @@
...
+  public void close() {
+    dropReference();
+    lock.unlock();
+  }
+
+  private int addReference() {
+    return counter.incrementAndGet();
+  }
+
+  private int dropReference() {
```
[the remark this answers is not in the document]
Bill: This is a `private` method. We know, in this method, that if `usages == 0` that `destruct()` has been called (in the past), since: the constructor started us with `counter.get() == 1` and `addReference()` is called for every `alias` call.

Because we know `destruct` has been called we know that `isClosed.get()` is already true.

geode-core/src/main/java/org/apache/geode/internal/net/SocketCloser.java
Bill: removed this unneeded log line

geode-core/src/main/java/org/apache/geode/internal/net/ByteBufferSharingImpl.java
```diff
@@ -0,0 +1,142 @@
...
+
+  /**
+   * This method is for use only by the owner of the shared resource. It's used for handing out
+   * references to the shared resource. So it does reference counting and also acquires a lock.
+   *
+   * Resource owners call this method as the last thing before returning a reference to the caller.
+   * That caller binds that reference to a variable in a try-with-resources statement and relies on
+   * the AutoCloseable protocol to invoke close() on the object at the end of the block.
+   */
+  ByteBufferSharing alias() {
```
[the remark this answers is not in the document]
Bill: renamed `open()` for symmetry with `close()`

geode-core/src/main/java/org/apache/geode/internal/net/NioSslEngine.java
```diff
@@ -315,50 +317,48 @@ public ByteBuffer ensureWrappedCapacity(int amount, ByteBuffer wrappedBuffer,
...
-   */
-  public ByteBuffer ensureUnwrappedCapacity(int amount) {
-    // for TTLS the app-data buffers do not need to be tracked direct-buffers since we
-    // do not use them for I/O operations
-    peerAppData =
-        bufferPool.expandReadBufferIfNeeded(TRACKED_RECEIVER, peerAppData, amount);
-    return peerAppData;
+  public ByteBufferSharing getUnwrappedBuffer(ByteBuffer wrappedBuffer) {
+    /*
+     * TODO: it can't be right that we ignore the wrappedBuffer parameter here!
```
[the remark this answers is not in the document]
Bill: Resolved and removed. We eliminated the parameter entirely.

geode-core/src/main/java/org/apache/geode/internal/net/NioSslEngine.java
```diff
@@ -368,17 +368,15 @@ public void doneReadingDirectAck(ByteBuffer unwrappedBuffer) {
     // read-operations
   }
 
-  @Override
```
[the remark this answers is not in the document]
Bill: thanks!

geode-core/src/main/java/org/apache/geode/internal/net/NioPlainEngine.java
```diff
@@ -121,8 +121,12 @@ public void doneReading(ByteBuffer unwrappedBuffer) {
   }
 
   @Override
-  public ByteBuffer getUnwrappedBuffer(ByteBuffer wrappedBuffer) {
-    return wrappedBuffer;
+  public ByteBufferSharing getUnwrappedBuffer() {
+    return shareBuffer(null);
```
bschuchardt, line 125: This method should throw an exception, not return a ByteBufferSharing with a null buffer.

bschuchardt (changes requested):

geode-core/src/main/java/org/apache/geode/internal/net/NioPlainEngine.java
Bill: This suggestion seems intuitive enough, however, where this is called from `Connection.readAck()` we rely on the current "no-op" behavior. If we were to throw here then it would break `readAck()` using the `NioPlainEngine`.

With the previous interface (`getUnwrappedBuffer(ByteBuffer)`) we were always sending in `null` when using the `NioPlainEngine` so the net result was the same in that case (we returned a `ByteBufferSharing` which held a `null` buffer.)

The only other place we call `getUnwrappedBuffer()` is from `clearSSLInputBuffer()` and in that case we know we are dealing with the `NioSslEngine` (not the `NioPlainEngine`.) We considered moving this method from the interface to the (`NioSslEngine`) implementation (and leaving it off `NioPlainEngine` entirely.) That would have been fine for `clearSSLInputBuffer` but would have left us with ugliness in the `readAck()` call site.

Bill:
That distributed test failure is [GEODE-8573](https://issues.apache.org/jira/browse/GEODE-8573)—it's not related to this PR so I'm re-running the suite.

geode-core/src/main/java/org/apache/geode/internal/net/NioPlainEngine.java
bschuchardt: can you have it wrap an empty buffer instead of null?

geode-core/src/main/java/org/apache/geode/internal/net/NioSslEngine.java
```diff
@@ -368,17 +365,15 @@ public void doneReadingDirectAck(ByteBuffer unwrappedBuffer) {
...
-  public synchronized boolean isClosed() {
-    return closed;
-  }
-
   @Override
   public synchronized void close(SocketChannel socketChannel) {
```
gesterzhou, line 369: Will this "synchronized" sync with wrap() method's shareOutputBuffer?

wrap() and close() have to sync with each other, otherwise it will encounter bug Gem-3116.

geode-core/src/main/java/org/apache/geode/internal/net/NioPlainEngine.java
Bill: What @bschuchardt's comment made me concerned about, is the prospect of a "client" of a `ByteBufferSharing` (`ByteBufferSharingImpl` particularly), calling `close()` too many times i.e. more than once.

Turns our there was a bug there. In the version of the code he commented on, if a client called `close()` twice then the reference count was decremented on the second call, even though it was already `0` (leaving it at `-1`.) You can imagine the bad things that happen in that state.

The latest commit adds a test for this bug (`ByteBufferSharingImplTest`) and also fixes the bug.

geode-core/src/main/java/org/apache/geode/internal/net/ByteBufferSharingImpl.java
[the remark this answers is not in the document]
Bill: What @bschuchardt's comment made me concerned about, is the prospect of a "client" of a `ByteBufferSharing` (`ByteBufferSharingImpl` particularly), calling `close()` too many times i.e. more than once.

Turns our there was a bug there. In the version of the code he commented on, if a client called `close()` twice then the reference count was decremented on the second call, even though it was already `0` (leaving it at `-1`.) You can imagine the bad things that happen in that state.

The latest commit adds a test for this bug (`ByteBufferSharingImplTest`) and also fixes the bug.

geode-core/src/main/java/org/apache/geode/internal/net/NioPlainEngine.java
Bill: done!

geode-core/src/main/java/org/apache/geode/internal/net/NioSslEngine.java
Bill: The `close()` method is a critical section. And after a thread has exited that critical section, the next thread entering it will skip all the actual close logic.

There is no possible interference between the close logic and e.g. `wrap()`, specifically in their use of the output buffer a.k.a. `myNetData` (the `ByteBuffer` managed by `outputSharing`) because both `wrap()` and `close()` acquire a reference to that buffer via `ByteBufferSharingImpl.open()` and `ByteBufferSharingImpl.open(long time, TimeUnit unit)` respectively. Those methods call `lock()`/`tryLock(long time, TimeUnit unit)` on the `ReentrantLock` before returning.

Does that help?

geode-core/src/main/java/org/apache/geode/internal/net/NioSslEngine.java
```diff
@@ -40,48 +40,48 @@
...
-
   private final BufferPool bufferPool;
 
   private boolean closed;
 
   SSLEngine engine;
 
   /**
-   * myNetData holds bytes wrapped by the SSLEngine
+   * holds bytes wrapped by the SSLEngine; a.k.a. myNetData
```
echobravopapa, line 63: i appreciate the buffer naming with their respective directions
maybe in the future this decoration can be extended to the uses of `peerAppData` and 'myNetData`  just to reduce cognitive load and increase readability

bschuchardt (approved):
I like the auto-closable approach.  It's much cleaner and easier to understand than the old code.

geode-core/src/main/java/org/apache/geode/internal/net/NioSslEngine.java
```diff
@@ -368,17 +365,15 @@ public void doneReadingDirectAck(ByteBuffer unwrappedBuffer) {
...
-
   @Override
   public synchronized void close(SocketChannel socketChannel) {
     if (closed) {
       return;
     }
-    try {
+    closed = true;
+    inputSharing.destruct();
+    try (final ByteBufferSharing outputSharing = shareOutputBuffer(1, TimeUnit.MINUTES)) {
```
echobravopapa, line 375: this probably warrants a comment for posterity...

geode-core/src/main/java/org/apache/geode/internal/net/NioSslEngine.java
```diff
@@ -405,14 +400,13 @@ public synchronized void close(SocketChannel socketChannel) {
...
+      }
     } finally {
-      ByteBuffer netData = myNetData;
-      ByteBuffer appData = peerAppData;
-      myNetData = null;
-      peerAppData = EMPTY_BUFFER;
-      bufferPool.releaseBuffer(TRACKED_SENDER, netData);
-      bufferPool.releaseBuffer(TRACKED_RECEIVER, appData);
-      this.closed = true;
+      outputSharing.destruct();
```
echobravopapa, line 409: much cleaner, like it!

echobravopapa (approved):

PR #5677, opened 2020-10-27
GEODE-7858: PR.clear notify client should let the queue holder member to notify

jinmeiliao:


gesterzhou (approved):

geode-cq/src/distributedTest/java/org/apache/geode/internal/cache/PartitionRegionClearMessageQueueDUnitTest.java
```diff
@@ -0,0 +1,163 @@
...
+    // verify that before register for interest, client region has no data
+    assertThat(testRegion).hasSize(0);
+
+    // verify that after register for interest, client region has data now.
+    testRegion.registerInterestForAllKeys();
+    assertThat(testRegion).hasSize(10);
+
+    // do PR clear on a server that has no client proxy, hence no HARegionQueue
+    Boolean cleared = false;
+    for (int i = 0; i < servers.size(); i++) {
```
mhansonp, line 92: Should this clear be invoked on each server? If it successfully clears on one server, it seems it will be happy. and finish the loop. One could just introduce a break if cleared == true rather than continuing to loop.

The logic here is a bit odd either way.

mhansonp (changes requested):
The logic is not clear in the for loops in the tests. Please clarify the for loops with comments or code changes..

geode-cq/src/distributedTest/java/org/apache/geode/internal/cache/PartitionRegionClearMessageQueueDUnitTest.java
jinmeiliao: done

mhansonp (approved):
much better

geode-cq/src/distributedTest/java/org/apache/geode/internal/cache/PartitionRegionClearMessageQueueDUnitTest.java
```diff
@@ -0,0 +1,165 @@
...
+    // verify that clear is called on a server
+    assertThat(cleared).isTrue();
+    // verify that the PR clear message is still delivered to the client
+    assertThat(clearEventReceived.get()).isTrue();
+  }
+
+  private static boolean clearRegionOnNoClientProxyMember() {
+    Collection<CacheClientProxy> clientProxies =
+        CacheClientNotifier.getInstance().getClientProxies();
+    if (clientProxies.isEmpty()) {
```
jchen21, line 155: I see in the tests the clear operations are invoked from a server directly, rather than from a client. If a server's `clientProxies` is not empty, can we invoke clear from that server? Will it work? I am not sure why clear has to be invoked from a server with empty `clientProxies`. Can you explain a bit?

jinmeiliao: the story is to verify that clear message is sent on the region level, i.e. even if the clear is done on a member that has no message dispatcher, the message should still be sent to the client. if a server has no `clientProxy` then it definitely won't have the `messageDispatcher`.

PR #5680, opened 2020-10-28
GEODE-8623: Retry getting local host if it fails.

pivotal-jbarrett:

### For all changes:

### Note:

### DIFF
--- a/geode-common/src/main/java/org/apache/geode/annotations/internal/SuppressDeprecationForBackwardsCompatibilityTesting.java
+++ b/geode-common/src/main/java/org/apache/geode/annotations/internal/SuppressDeprecationForBackwardsCompatibilityTesting.java
@@ -1,0 +1,46 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package org.apache.geode.annotations.internal;
+
+import static java.lang.annotation.ElementType.CONSTRUCTOR;
+import static java.lang.annotation.ElementType.FIELD;
+import static java.lang.annotation.ElementType.LOCAL_VARIABLE;
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.ElementType.PACKAGE;
+import static java.lang.annotation.ElementType.PARAMETER;
+import static java.lang.annotation.ElementType.TYPE;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Indicates suppression of deprecation warning because the element is required for backwards
+ * compatibility testing. The element being suppressed should have a
+ * {@link DeprecatedButRequiredForBackwardsCompatibilityTesting} annotation.
+ *
+ * Should be paired with {@link SuppressWarnings}("deprecation") to actually suppress warning.
+ */
+@Documented
+@Retention(RetentionPolicy.SOURCE)
+@Target(value = {CONSTRUCTOR, FIELD, LOCAL_VARIABLE, METHOD, PACKAGE, PARAMETER, TYPE})
+public @interface SuppressDeprecationForBackwardsCompatibilityTesting {
+
+  /** Optional description */
+  String value() default "";
+
+}
--- a/geode-common/src/main/java/org/apache/geode/internal/Retry.java
+++ b/geode-common/src/main/java/org/apache/geode/internal/Retry.java
@@ -1,0 +1,73 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.apache.geode.internal;
+
+import static java.util.concurrent.TimeUnit.NANOSECONDS;
+
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.function.Predicate;
+import java.util.function.Supplier;
+
+/**
+ * Utility class for retrying operations.
+ */
+public class Retry {
+
+  interface Clock {
+    long nanoTime();
+  }
+
+  private static class SteadyClock implements Clock {
+    @Override
+    public long nanoTime() {
+      return System.nanoTime();
+    }
+  }
+
+  private static final SteadyClock steadyClock = new SteadyClock();
+
+  /**
+   * Try the supplier function until the predicate is true or timeout occurs.
+   *
+   * @param supplier to execute until predicate is true or times out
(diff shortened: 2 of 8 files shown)


geode-common/src/main/java/org/apache/geode/internal/inet/LocalHostUtil.java
```diff
@@ -54,70 +59,80 @@
...
       && Boolean.getBoolean("java.net.preferIPv6Addresses");
 
-  static {
-    InetAddress inetAddress = null;
+  /**
+   * Resolves local host. Will retry if resolution fails.
+   *
+   * @return local host if resolved otherwise null.
+   */
+  private static InetAddress resolveLocalHost() {
```
dschneider-pivotal, line 74: Could you add some type of test for the new "resolveLocalHost" method? If you had this method call a local method that then calls InetAddress.getLocalHost you could then mock that method and control when it would throw UnknownHostException to show that this new code will do retries.

geode-common/src/main/java/org/apache/geode/internal/inet/LocalHostUtil.java
```diff
@@ -54,70 +59,80 @@
...
-            useIPv6Addresses = false;
-          } else if (ipv6Fallback != null) {
-            inetAddress = ipv6Fallback;
-            useIPv6Addresses = true;
-          }
-        }
-      }
-    } catch (UnknownHostException ignored) {
+        return inetAddress;
+      }, Objects::nonNull, 1, MINUTES);
```
dschneider-pivotal, line 126: Why is 1 minute enough? Instead of hard coding it to 1 minute should that be the default that can be overridden by setting a system property. 
Did you intend this to be a hot retry loop? If something is trying to get DNS up it might be better not to bombard it with requests during this minute.

dschneider-pivotal (commented):
In reading the ticket description I thought you would be making some change so that we don't end up with a "permanent unknown host exception". But it seems like we still will but now whatever is wrong has to persist for a minute for this to happen which seems much better than the previous single attempt. I think it would be helpful to rewrite the ticket description to make this clear (or you could just do it in an explanation when you mark the ticket fixed).

geode-common/src/main/java/org/apache/geode/annotations/internal/SuppressDeprecationForBackwardsCompatibilityTesting.java
```diff
@@ -0,0 +1,46 @@
...
+ * Indicates suppression of deprecation warning because the element is required for backwards
+ * compatibility testing. The element being suppressed should have a
+ * {@link DeprecatedButRequiredForBackwardsCompatibilityTesting} annotation.
+ *
+ * Should be paired with {@link SuppressWarnings}("deprecation") to actually suppress warning.
+ */
+@Documented
+@Retention(RetentionPolicy.SOURCE)
+@Target(value = {CONSTRUCTOR, FIELD, LOCAL_VARIABLE, METHOD, PACKAGE, PARAMETER, TYPE})
+public @interface SuppressDeprecationForBackwardsCompatibilityTesting {
```
jinmeiliao, line 41: i wonder if we can get rid of this annotation by just not using the deprecated method in the tests. I searched for the usage, there are only two places that uses `SocketCreator.getLocalhost`

geode-common/src/main/java/org/apache/geode/internal/inet/LocalHostUtil.java
jinmeiliao: that method has to be a static method and we have to use powermock to mock a static method. We are trying to steer away from using power mock as far as I understand.

There are tests for Retry.

geode-common/src/main/java/org/apache/geode/internal/inet/LocalHostUtil.java
jinmeiliao: you mean we should have some interval between retires?

geode-common/src/main/java/org/apache/geode/internal/inet/LocalHostUtil.java
demery-pivotal: If you want to test, you can create an overload that takes suppliers, functions, and predicates to call instead of calling static methods, then mock those things to drive the desired test conditions. Then `resolveLocalHost()` could simply call the overloaded one, passing references to the static methods, and we would leave that untested.

geode-common/src/main/java/org/apache/geode/internal/inet/LocalHostUtil.java
pivotal-jbarrett: It certainly could. I am not to concerned about the one in thousand times this might be hit in a managed environment though. A hot loop here probably isn't the end of the world. The request will go to the local resolver first then to the DNS so its not going to be all that hot.

geode-common/src/main/java/org/apache/geode/internal/inet/LocalHostUtil.java
pivotal-jbarrett: There is a test that already asserts the retry behavior outside the context of the thing being retried. Testing this method specifically must be an integration test since it requires the use of a resolver and DNS. There are several tests that are tangentially dependent on this method given that nothing in Geode starts if this method returns `UnknownHostException`.

geode-common/src/main/java/org/apache/geode/internal/Retry.java
```diff
@@ -0,0 +1,73 @@
...
+   * Try the supplier function until the predicate is true or timeout occurs.
+   *
+   * @param supplier to execute until predicate is true or times out
+   * @param predicate to test for retry
+   * @param timeout to retry for
+   * @param timeUnit to retry for
+   * @param <T> type of return value
+   * @return value from supplier after it passes predicate or times out.
+   */
+  public static <T> T tryFor(final Supplier<T> supplier, final Predicate<T> predicate,
```
Bill, line 52: Looking at the call sites for these methods, I wonder if those would be easier to understand if the non-lambda args came first? Not adamant about that but sometimes it makes things easier to read.

I usually put functions last in the ruby (block) style when a method takes a single one. Since this method takes two functions it may be beyond help though.

geode-common/src/main/java/org/apache/geode/internal/Retry.java
```diff
@@ -0,0 +1,73 @@
...
+      throws TimeoutException {
+    long until = clock.nanoTime() + NANOSECONDS.convert(timeout, timeUnit);
+
+    T value;
+    do {
+      value = supplier.get();
+      if (predicate.test(value)) {
+        return value;
+      }
+    } while (clock.nanoTime() < until);
```
Bill, line 68: This is a hot loop unless `supplier` or `predicate` sleeps. On the one hand that's cool because it leave sleeping outside this class. On the other hand, um, it might be nice to have this structured to not have a hot loop by default.

Similar to the way the `Clock` implementation is injected, a sleeper could be injected, if you wanted to do that, maybe you could:

```
@FunctionalInterface
public interface Sleeper {
 void sleep() throw InterruptedException;
}
```
and after `predicate` evaluates to `false`:

```
sleeper.sleep(...)
```
Where the sleep time comes from parameters the way `timeout` et al. do now.

Note that some sophistication could be added to clamp the sleep time so as to try and not overflow `until` by too much. Then again, that might turn into a fools errand.

geode-common/src/main/java/org/apache/geode/internal/inet/LocalHostUtil.java
```diff
@@ -54,70 +59,80 @@
...
-          boolean ipv4 = addr instanceof Inet4Address;
-          if ((preferIPv6 && ipv6) || (!preferIPv6 && ipv4)) {
-            String addrName = reverseDNS(addr);
-            if (inetAddress.isLoopbackAddress()) {
-              inetAddress = addr;
-              lhName = addrName;
-            } else if (addrName != null) {
-              inetAddress = addr;
-              lhName = addrName;
+      return tryFor(() -> {
```
Bill, line 76: This looks like a hot loop. See comments above on `tryFor()`

geode-common/src/main/java/org/apache/geode/internal/Retry.java
Bill: after writing this I saw the comments about the hot loop by others below but decided to leave it here anyway because I'm obsessed with this kind of stuff.

geode-common/src/main/java/org/apache/geode/internal/Retry.java
```diff
@@ -0,0 +1,73 @@
...
+
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.function.Predicate;
+import java.util.function.Supplier;
+
+/**
+ * Utility class for retrying operations.
+ */
+public class Retry {
```
Bill, line 27: So many cool directions this could take, for use-cases beyond the current one. Thinking of back-off policies, for instance.

Bill (commented):
The annotations and the `Retry` abstraction are really cool! Some random comments…

jinmeiliao:
appreciate everyone's feedback. I am closing this PR and continue the work on another PR: https://github.com/apache/geode/pull/5743

jinmeiliao:
appreciate everyone's feedback. I am closing this PR and continue the work on another PR: #5743

PR #5682, opened 2020-10-28
GEODE-8668: Implement Redis SELECT command

sabbey37:
Co-authored-by: Hale Bales <hbales@pivotal.io>

We are implementing [SELECT](https://redis.io/commands/select) because some Redis analysis tools (like Redis Insight) require the `SELECT` command in order to function properly.  `SELECT` returns `ERR Only DB 0 supported` if anything other than 0 is passed in as a DB index argument.

### For all changes:

### Note:


sabbey37:
DUnit test failure is unrelated and captured in: https://issues.apache.org/jira/browse/GEODE-8669

geode-redis/src/acceptanceTest/java/org/apache/geode/redis/internal/executor/connection/SelectNativeRedisAcceptanceTest.java
```diff
@@ -0,0 +1,60 @@
...
+  @Before
+  public void setUp() {
+    // starting redis in cluster mode
+    redisContainer =
+        new GenericContainer<>("grokzen/redis-cluster:5.0.9")
+            .withExposedPorts(7000)
+            .withEnv("IP", "0.0.0.0");
+
+    redisContainer.start();
+    jedis = new Jedis("localhost", getPort(), REDIS_CLIENT_TIMEOUT);
```
jdeppe-pivotal, line 46: No need to do this here since the super class handles the client setup.

geode-redis/src/main/java/org/apache/geode/redis/internal/RedisConstants.java
```diff
@@ -30,6 +30,8 @@
   public static final String SERVER_ERROR_MESSAGE =
       "The server had an internal error please try again";
   public static final String SERVER_ERROR_SHUTDOWN = "The server is shutting down";
+  public static final String ERROR_SELECT_CLUSTER_MODE = "SELECT is not allowed in cluster mode";
```
jdeppe-pivotal, line 33: I'm afraid that this error will be a bit misleading since it mentions `cluster mode`. I think the `invalid DB index` is better and we should document the limitation of only supporting DB 0. I do realize that this is a bit inconsistent with how native redis reports things.

jdeppe-pivotal (changes requested):

geode-redis/src/acceptanceTest/java/org/apache/geode/redis/internal/executor/connection/SelectNativeRedisAcceptanceTest.java
sabbey37: Good point. I just changed it to a class level method for setup and teardown of the Redis container so it didn't start a new container before each test.

geode-redis/src/main/java/org/apache/geode/redis/internal/RedisConstants.java
sabbey37: I definitely understand what you mean.  I think it would be good for us to talk about this with the team.  There is also an `ERR DB index is out of range` message Redis returns if the user tries to select a DB that is not 0-15, but is still a long.  We could potentially return that instead.

sabbey37: After some discussion, we decided to return `ERR Only DB 0 supported` if anything other than 0 is passed in as a DB index argument.

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/connection/SelectExecutor.java
```diff
@@ -0,0 +1,38 @@
...
+
+  @Override
+  public RedisResponse executeCommand(Command command, ExecutionHandlerContext context) {
+    String dbIndexString = command.getStringKey();
+
+    if (!dbIndexString.equals("0")) {
+      return RedisResponse.error(ERROR_SELECT);
+    }
+
+    return RedisResponse.ok();
```
sabbey37, line 35: I changed this so we were no longer parsing the long and just comparing the string value.  Let me know if you think this is the right approach.

jdeppe-pivotal (approved):
Looks good

PR #5684, opened 2020-10-28
GEODE-8570: deprecate cluster-configuration-dir property

jinmeiliao:


demery-pivotal:
In addition to marking the property as deprecated, we should fix the documentation. The documentation gives the mistaken impression that the property affects the product. But the product does not use this property in any way, and so the documentation is incorrect.

geode-docs/reference/topics/gemfire_properties.html.md.erb
```diff
@@ -117,7 +117,7 @@ Valid values are in the range 0...2147483647</td>
 </tr>
 <tr>
 <td>cluster-configuration-dir</td>
-<td>This property specifies the directory in which the cluster configuration related disk-store and artifacts are stored. This property is only applicable to dedicated locators that have &quot;enable-cluster-configuration&quot; set to true.</td>
+<td><b>Deprecated.</b> Use import/export clutser-configuration command instead. This property specifies the directory in which the cluster configuration related disk-store and artifacts are stored. This property is only applicable to dedicated locators that have &quot;enable-cluster-configuration&quot; set to true.</td>
```
karensmolermiller, line 120: I only looked at the docs portion of this PR.

1. Please change "clutser" to "cluster"
2. Please add "gfsh" in to this sentence, since we're not substituting a different property, and I think that you mean to use a gfsh command instead.  Something like "Use the gfsh import or export cluster-configuration command instead."

karensmolermiller (changes requested):

jinmeiliao:
> In addition to marking the property as deprecated, we should fix the documentation. The documentation gives the mistaken impression that the property affects the product. But the product does not use this property in any way, and so the documentation is incorrect.

I am not sure about this. Should we treat it as a bug and try to fix it in the future or should we document it if it would never work

geode-docs/reference/topics/gemfire_properties.html.md.erb
kirklund: I also recommend mentioning that it currently does nothing. We've done this in the past for APIs that do nothing -- deprecate and document that it does nothing and what the user should do instead.

demery-pivotal (approved):

karensmolermiller (approved):
The documentation changes look good to me.  Thank you for the revision.

PR #5686, opened 2020-10-28
GEODE-8659: Consolidating locator-wait-time tests

upthewaterspout:
We had two different tests of this functionality that were testing almost the
same thing. The first test had a slightly odd configuration - only one of the
locators knew about the other.

Combining these into a single test that locators wait for locator-wait-time and
stop waiting when all locators are found.


bschuchardt (approved):
removing the test that used an odd configuration is a good idea

kohlmu-pivotal (approved):

PR #5689, opened 2020-10-29
GEODE-8667: Duplicate online Oplog compaction after offline Oplog compaction

jchen21:
When Oplog.totalCount == 0, no Oplog compaction is needed.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/Oplog.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/Oplog.java
@@ -5795,7 +5795,6 @@ public class Oplog implements CompactableOplog, Flushable {
     } else {
       return true;
     }
-
     return false;
   }
 
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/Oplog.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/Oplog.java
@@ -1780,6 +1780,9 @@ public class Oplog implements CompactableOplog, Flushable {
           keyBytes = DataSerializer.readByteArray(dis);
         } // while
         setRecoverNewEntryId(oplogKeyIdHWM);
+        if (totalCount.get() == 0 && totalLiveCount.get() > 0) {
+          totalCount.set(totalLiveCount.get());
+        }
       } catch (IOException ex) {
         try {
           fis.close();
@@ -5789,7 +5792,7 @@ public class Oplog implements CompactableOplog, Flushable {
       if (((rv / (double) rvHWMtmp) * 100) <= parent.getCompactionThreshold()) {
         return true;
       }
-    } else if (this.totalLiveCount.get() <= 0) {
+    } else {
       return true;
     }
 
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/Oplog.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/Oplog.java
@@ -1780,8 +1780,9 @@ public class Oplog implements CompactableOplog, Flushable {
           keyBytes = DataSerializer.readByteArray(dis);
         } // while
         setRecoverNewEntryId(oplogKeyIdHWM);
-        if (totalCount.get() == 0 && totalLiveCount.get() > 0) {
-          totalCount.set(totalLiveCount.get());
+        long tlc = totalLiveCount.get();
+        if (totalCount.get() == 0 && tlc > 0) {
+          totalCount.set(tlc);
         }
       } catch (IOException ex) {
         try {
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/Oplog.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/Oplog.java
@@ -5789,6 +5789,8 @@ public class Oplog implements CompactableOplog, Flushable {
       if (((rv / (double) rvHWMtmp) * 100) <= parent.getCompactionThreshold()) {
         return true;
       }
+    } else if (this.totalLiveCount.get() <= 0) {
+      return true;
     }
 
     return false;
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/Oplog.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/Oplog.java
@@ -5790,7 +5790,7 @@ public class Oplog implements CompactableOplog, Flushable {
         return true;
       }
     } else {
-      return true;
+      // return true;
     }
     return false;
   }
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/Oplog.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/Oplog.java
@@ -5789,9 +5789,8 @@ public class Oplog implements CompactableOplog, Flushable {
       if (((rv / (double) rvHWMtmp) * 100) <= parent.getCompactionThreshold()) {
         return true;
       }
-    } else {
-      // return true;
     }
+
     return false;
   }
 
--- a/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/CompactDiskStoreDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/CompactDiskStoreDUnitTest.java
@@ -1,0 +1,117 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.apache.geode.internal.cache;
+
+import static org.apache.geode.test.dunit.rules.ClusterStartupRule.getCache;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.Serializable;
+import java.util.Properties;
+import java.util.stream.IntStream;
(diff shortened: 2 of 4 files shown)


dschneider-pivotal (changes requested):
add unit test coverage of needsCompaction to OplogTest.

It seems like the only change you made was to say we don't need to compact if an Oplog has a totalCount == 0. I think totalCount == 0 means that the oplog is empty. This seems like something we should compact. Do you understand why offline compaction gave us an empty oplog? I'm concerned this change might cause us not to remove empty oplogs. Or maybe we say it has no need to compact but still do a check and remove it for being empty somewhere else?

upthewaterspout:
I agree with Darrel.  Maybe the problem this test is finding is that the compaction is leaving behind an empty oplog (perhaps it generates an empty oplog to write to when it starts or something)?

If that's the case though - it's not much of a bug, especially since it gets cleaned up on startup the next time. 

If I remember what you were showing me earlier, I thought the compaction on startup after the offline compact was taking a long time. I wouldn't expect that to be the case if all it's doing is removing empty oplogs?

geode-core/src/main/java/org/apache/geode/internal/cache/Oplog.java
```diff
@@ -5789,9 +5789,8 @@ boolean needsCompaction() {
       if (((rv / (double) rvHWMtmp) * 100) <= parent.getCompactionThreshold()) {
         return true;
       }
-    } else {
```
gesterzhou, line 5792: This fix is incorrect. It only caused the empty oplog (created by the offline compaction) is not deleted by the recovery's auto compaction. 

I modified the dunit test a little bit and proved it. 

Unfortunately, there's no flag or status to show that a compaction is just finished. Even we want to use the empty oplog created by offline compaction, we need to careful arrange the logic. Such as:

change  needsCompaction to return int. 1: need to compact, -1: no need to compact, 0: there's one empty file,  only need to delete it.

But maybe there's better ideas.

jchen21:
> add unit test coverage of needsCompaction to OplogTest.
> 
> It seems like the only change you made was to say we don't need to compact if an Oplog has a totalCount == 0. I think totalCount == 0 means that the oplog is empty. This seems like something we should compact. Do you understand why offline compaction gave us an empty oplog? I'm concerned this change might cause us not to remove empty oplogs. Or maybe we say it has no need to compact but still do a check and remove it for being empty somewhere else?

I have added a unit test in `OplogTest`. 

`totalCount` == 0 does not necessarily mean the oplog is empty. You remind me of the case of an empty oplog. There is another field `totalLiveCount` to check. What I see in the persistent recovery immediately after an offline compaction is, the offline compacted oplog is non-empty. During the recovery, `totalCount` is `0`. However, `totalLiveCount` has the actual number of live entries in the oplog. Therefore I add one more condition to check `totalLiveCount`. 

If `totalCount` == 0 and `totalLiveCount` > 0, such as the case of recovery after offline compaction, we don't need compaction again. Are there any other cases that need compaction, even though `totalCount` == 0 and `totalLiveCount` > 0?

If `totalCount` == 0 and `totalLiveCount` <= 0, then the oplog is empty, which needs compaction. Please correct me if I am wrong.

jchen21:
When compacting oplogs offline, it creates a new krf and persists `totalCount` with `0` in Oplog constructor. Then live entries are copied from old oplogs to the new oplog. `totalCount` of the new oplog is updated during this process. However, the updated `totalCount` is not persisted. Its persisted value is still `0`, although the oplog is non-empty after compaction. 

During recovery, the value of `totalCount` is recovered from krf as `0`.  `totalLiveCount` is incremented as the oplog is recovered. `totalCount` remains `0`. Therefore, if `totalCount` == 0 and `totalLiveCount` > 0, no compaction is needed. Since this condition shows that offline compaction is just done before recovery.

geode-core/src/main/java/org/apache/geode/internal/cache/Oplog.java
gesterzhou: The new fix is good. 

Should use "} else if (hasNoLiveValues()) {" instead. 

And it's better to add a few comments here.

gesterzhou: Darrel and I discussed again. We might need to re-fix it the other way: 

Since the trouble is caused by offline compaction wrote totalCount=0 into krf, there's a potential bug that we restart with totalCount=0, but totalLiveCount=1, then with your current code change this oplog will never get a chance to compact until the only live entry is destroyed. 

So a better fix maybe: when restart and found totalCount=0, then update totalCount to be totalLiveCount. With this fix, you don't need to change needsCompaction() any more.

jchen21: I have changed `readKrf()` function. So that at the end of this function, when `totalCount` and `totalLiveCount` is updated after reading krf file, it will check if `totalCount` is `0` and `totalLiveCount` is greater than `0`, and update `totalCount` accordingly.

gesterzhou:
I read your new fix and have different opinion. The root cause is the new oplog (i.e. #2) created during offline compaction wrote a krf with totalCount==0. 

You fix is to still read the krf and when found it's totalCount is 0, change it to be totalLiveCount. This is just a patch fix. The real issue is: we should NOT create such a krf with totalCount==0. Actually, when offline compaction, it is not allowed to create krf. 

Here is the better fix for this issue:

diff --git a/geode-core/src/main/java/org/apache/geode/internal/cache/Oplog.java b/geode-core/src/main/java/org/apache/geode/internal/cache/Oplog.java
index f6daa3a1d6..0a31af9386 100755
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/Oplog.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/Oplog.java
@@ -582,7 +582,7 @@ public class Oplog implements CompactableOplog, Flushable {
       createDrf(null);
       createCrf(null);
       // open krf for offline compaction
-      if (getParent().isOfflineCompacting()) {
+      if (getParent().isOfflineCompacting() && couldHaveKrf()) {
         krfFileCreate();
       }
     } catch (Exception ex) {
@@ -3980,7 +3980,7 @@ public class Oplog implements CompactableOplog, Flushable {
     } catch (IOException ignore) {
     }
 
-    if (this.krf.f.exists()) {
+    if (this.krf.f != null && this.krf.f.exists()) {
       this.krf.f.delete();
     }
   }
@@ -5789,7 +5789,7 @@ public class Oplog implements CompactableOplog, Flushable {
       if (((rv / (double) rvHWMtmp) * 100) <= parent.getCompactionThreshold()) {
         return true;
       }
-    } else if (this.totalLiveCount.get() <= 0) {
+    } else {
       return true;
     }


I don't know who introduced "if (getParent().isOfflineCompacting()" here. To be conservative, I still keep this check. Maybe we can remove the 3 lines. But I want to be conservative for now.

jchen21:
> ```
> I read your new fix and have different opinion. The root cause is the new oplog (i.e. #2) created during offline compaction wrote a krf with totalCount==0. 
> 
> You fix is to still read the krf and when found it's totalCount is 0, change it to be totalLiveCount. This is just a patch fix. The real issue is: we should NOT create such a krf with totalCount==0. Actually, when offline compaction, it is not allowed to create krf. 
> 
> Here is the better fix for this issue:
> 
> diff --git a/geode-core/src/main/java/org/apache/geode/internal/cache/Oplog.java b/geode-core/src/main/java/org/apache/geode/internal/cache/Oplog.java
> index f6daa3a1d6..0a31af9386 100755
> --- a/geode-core/src/main/java/org/apache/geode/internal/cache/Oplog.java
> +++ b/geode-core/src/main/java/org/apache/geode/internal/cache/Oplog.java
> @@ -582,7 +582,7 @@ public class Oplog implements CompactableOplog, Flushable {
>        createDrf(null);
>        createCrf(null);
>        // open krf for offline compaction
> -      if (getParent().isOfflineCompacting()) {
> +      if (getParent().isOfflineCompacting() && couldHaveKrf()) {
>          krfFileCreate();
>        }
>      } catch (Exception ex) {
> @@ -3980,7 +3980,7 @@ public class Oplog implements CompactableOplog, Flushable {
>      } catch (IOException ignore) {
>      }
>  
> -    if (this.krf.f.exists()) {
> +    if (this.krf.f != null && this.krf.f.exists()) {
>        this.krf.f.delete();
>      }
>    }
> @@ -5789,7 +5789,7 @@ public class Oplog implements CompactableOplog, Flushable {
>        if (((rv / (double) rvHWMtmp) * 100) <= parent.getCompactionThreshold()) {
>          return true;
>        }
> -    } else if (this.totalLiveCount.get() <= 0) {
> +    } else {
>        return true;
>      }
> 
> 
> I don't know who introduced "if (getParent().isOfflineCompacting()" here. To be conservative, I still keep this check. Maybe we can remove the 3 lines. But I want to be conservative for now.
> ```

I don't think we should avoid creating krf during offline compaction. Your proposed fix avoid creating the krf file during offline compaction. After the compaction, the original non-compacted krf is also deleted. Without a krf, there is a significant impact on the performance of persistent recovery.

gesterzhou (approved):

jchen21:
@dschneider-pivotal So far, I still consider the current fix is the best option we can have. We need a convenient and reliable way to tell that the Oplog is offline compacted. The persisted `0` value of `totalCount` in krf serves the purpose. And during persistence recovery, if `totalCount` is `0`, then it is updated with the value of `totalLiveCount`, which is the most accurate value of the number of live entries in the Oplog.

I have been thinking about whether we should use the local variable `krfEntryCount`, instead of `totalLiveCount`. At the end of `readKrf()`, the value of `totalLiveCount` and `krfEntryCount` are the same. Because at the beginning of `readKrf()`, `totalLiveCount` and `krfEntryCount` are both 0. And in the `while` loop inside `readKrf()`, both variables are incremented together: 

`totalLiveCount` is incremented here: https://github.com/apache/geode/blob/develop/geode-core/src/main/java/org/apache/geode/internal/cache/Oplog.java#L1765

`krfEntryCount` is incremented here: https://github.com/apache/geode/blob/develop/geode-core/src/main/java/org/apache/geode/internal/cache/Oplog.java#L1768

I'd prefer using `totalLiveCount` instead of `krfEntryCount` which is a local variable.

dschneider-pivotal (approved):
I agree with your reasons for using totalLiveCount.

PR #5690, opened 2020-10-29
GEODE-8675: Log at query start when <trace> is enabled

DonalEvans:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/cache/query/internal/DefaultQuery.java
+++ b/geode-core/src/main/java/org/apache/geode/cache/query/internal/DefaultQuery.java
@@ -783,7 +783,7 @@ public class DefaultQuery implements Query {
       } else {
         queryObserver = qo;
       }
-      this.cache.getLogger().info("Starting query: " + this.queryString);
+      logger.info("Starting query: " + this.queryString);
     }
     return queryObserver;
   }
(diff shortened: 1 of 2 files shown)


geode-core/src/main/java/org/apache/geode/cache/query/internal/DefaultQuery.java
```diff
@@ -783,6 +783,7 @@ public QueryObserver startTrace() {
       } else {
         queryObserver = qo;
       }
+      this.cache.getLogger().info("Starting query: " + this.queryString);
```
kohlmu-pivotal, line 786: why would you not use Logger defined https://github.com/apache/geode/blob/d8fc2e12ec02f8b3deeb063396d0060f5e8f3ed7/geode-core/src/main/java/org/apache/geode/cache/query/internal/DefaultQuery.java#L69

kohlmu-pivotal (changes requested):

geode-core/src/main/java/org/apache/geode/cache/query/internal/DefaultQuery.java
DonalEvans: I was using the same pattern found elsewhere in the class (specifically in the two `endTrace()` methods) for consistency. The defined logger appears to not be used at all in the class as things stand, but if it's better to do so, I can replace all calls to `this.cache.getLogger()` in the class with `logger`.

kohlmu-pivotal: Interesting pattern indeed. Wonder why the Cache would be exposing it's logger like that.
If you want, you could always set the logger in the constructor.

Wonder if there is a real reason why the cache is exposing it's logger this way.

But thank you for amending.

kohlmu-pivotal (approved):

karensmolermiller (approved):
The docs portion of this PR looks good to me.  I approve of the docs part. I haven't looked at the code.

PR #5712, opened 2020-11-05
GEODE-8652: Round Two DING!

Bill:
This is a second try fixing [GEODE-8652](https://issues.apache.org/jira/browse/GEODE-8652).

- Original PR (now reverted) was: [PR #5666](https://github.com/apache/geode/pull/5666)

We committed a fix for this ticket a week ago but then found other problems in some applications, so we reverted it today. 

We've included a brand new concurrency test, `ByteBufferConcurrencyTest` in this latest PR. That new test validates that the newly-discovered concurrency issue is resolved. That test joins a couple tests that were new in the reverted PR: `ConnectionCloseSSLTLSDUnitTest`, and `ByteBufferSharingImplTest`.

Main product differences in this PR vs the reverted one is that `ByteBufferSharingImpl` has code in `addReference()` to ensure that a reference cannot be added after the sharing object has been "destructed". This prevents the reference count from growing (above zero) if it ever reaches zero. That in turn ensures that:

```
  private int dropReference() {
    final int usages = counter.decrementAndGet();
    if (usages == 0) {
      bufferPool.releaseBuffer(bufferType, buffer);
    }
    return usages;
  } 
```

Will never see `usages == 0` again. So we can never release the buffer again.

It also ensures that attempts to call `open()` on this sharing object after a `destruct()` call, will throw an `IOException`.


bschuchardt (approved):
I like the new tests.  Is it necessary for them to be executed 100 times?

dschneider-pivotal (approved):

PR #5722, opened 2020-11-09
GEODE-8692: ArrayIndexOutOfBoundsException may be thrown in RegionAdvisor.processProfilesQueuedDuringInitialization

sabbey37:
If `RegionAdvisor.buckets == null` at the time the value of `serials` is generated, an `ArrayIndexOutOfBoundsException` may be thrown in `RegionAdvisor.processProfilesQueuedDuringInitialization` if buckets have been initialized at that point.

Code where `ArrayIndexOutOfBoundsException` might be thrown in `RegionAdvisor.processProfilesQueuedDuringInitialization()`:

```
for (int i = 0; i < buckets.length; i++) {
  BucketAdvisor ba = buckets[i].getBucketAdvisor();
  int serial = qbp.serials[i]; <<< Exception thrown here
  if (serial != ILLEGAL_SERIAL) {
    ba.removeIdWithSerial(qbp.memberId, serial, qbp.destroyed);
  }
}
```
We ultimately decided that if the buckets were not created at the time the value of `serials` is generated, there is no need to worry about their profiles getting deleted later.  To avoid the `ArrayIndexOutOfBoundsException`, `getBucketSerials()` was modified to return an array of `ILLEGAL_SERIAL` values if the buckets had not been created yet.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/test/java/org/apache/geode/internal/cache/partitioned/RegionAdvisorJUnitTest.java
+++ b/geode-core/src/test/java/org/apache/geode/internal/cache/partitioned/RegionAdvisorJUnitTest.java
@@ -17,7 +17,11 @@ package org.apache.geode.internal.cache.partitioned;
 import static org.apache.geode.distributed.internal.DistributionAdvisor.ILLEGAL_SERIAL;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyBoolean;
+import static org.mockito.ArgumentMatchers.anyInt;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
 import org.junit.Before;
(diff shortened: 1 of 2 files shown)


dschneider-pivotal (changes requested):
Looks good. Can you add a unit test and add coverage of the new code to it?

sabbey37:
> Looks good. Can you add a unit test and add coverage of the new code to it?

Just added a couple of unit tests @dschneider-pivotal.  Let me know what you think.

geode-core/src/test/java/org/apache/geode/internal/cache/partitioned/RegionAdvisorJUnitTest.java
```diff
@@ -0,0 +1,72 @@
...
+    DistributionManager distributionManager = mock(DistributionManager.class);
+    when(regionAdvisor.getDistributionManager()).thenReturn(distributionManager);
+    when(distributionManager.isCurrentMember(any())).thenReturn(true);
+    regionAdvisor.preInitQueue.add(qbp);
+
+    ProxyBucketRegion pbr = mock(ProxyBucketRegion.class);
+    when(pbr.getBucketAdvisor()).thenReturn(mock(BucketAdvisor.class));
+    regionAdvisor.buckets = new ProxyBucketRegion[] {pbr, pbr, pbr};
+
+    regionAdvisor.processProfilesQueuedDuringInitialization();
```
DonalEvans, line 70: Might it be worth adding a check to verify that `ProxyBucketRegion.removeIdWithSerial()` is never called for the contents of `regionAdvisor.buckets` since the serial is always `ILLEGAL_SERIAL`? Or would that be outside the scope of this test?

dschneider-pivotal (approved):

geode-core/src/test/java/org/apache/geode/internal/cache/partitioned/RegionAdvisorJUnitTest.java
```diff
@@ -0,0 +1,72 @@
...
+    regionAdvisor = new RegionAdvisor(partitionedRegion);
+  }
+
+  @Test
+  public void getBucketSerials_shouldReturnAnArrayOfIllegalSerials_whenBucketsAreNull() {
+    RegionAttributes regionAttributes = mock(RegionAttributes.class);
+    PartitionAttributes partitionAttributes = mock(PartitionAttributes.class);
+    when(partitionedRegion.getAttributes()).thenReturn(regionAttributes);
+    when(regionAttributes.getPartitionAttributes()).thenReturn(partitionAttributes);
+    when(partitionAttributes.getTotalNumBuckets()).thenReturn(3);
```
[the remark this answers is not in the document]
sabbey37: Definitely. Good call!

geode-core/src/test/java/org/apache/geode/internal/cache/partitioned/RegionAdvisorJUnitTest.java
```diff
@@ -0,0 +1,72 @@
...
+
+    assertThat(regionAdvisor.getBucketSerials()).containsExactly(serials);
+  }
+
+  @Test
+  public void processProfilesQueuedDuringInitialization_shouldNotThrowIndexOutOfBoundsException() {
```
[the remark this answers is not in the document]
sabbey37: Sounds good!

geode-core/src/test/java/org/apache/geode/internal/cache/partitioned/RegionAdvisorJUnitTest.java
sabbey37: It definitely couldn't hurt to add it.  I feel like this test needs something more, but I'm not sure what.  The intent was mainly to show that `processProfilesQueuedDuringInitialization()` wouldn't result in an `ArrayIndexOutOfBoundsException` when given the serials that would be returned if `buckets` were `null` when `getBucketSerials()` is called.  What do you think?

DonalEvans: I think that just showing there's no `ArrayIndexOutOfBoundsException` is probably a bit weak, since we're explicitly setting the contents of both arrays in the test, so we know that exception isn't possible here. Asserting that the method is behaving as expected for the given conditions (not calling `removeIdWithSerial()` for buckets that correspond to an `ILLEGAL_SERIAL` value) would make for a more meaningful test.

sabbey37: I just added that verification!

DonalEvans (approved):

PR #5725, opened 2020-11-10
GEODE-8664: Nest errors in DistributionImpl.start

gaussianrecurrence:
- If while calling DistributionImpl.start there was either a
   MembershipConfigurationException or MemberStartupException exceptions, its
   original cause was not propagated, therefore being unable to properly
   tackling issues on startup.
 - This commit propagates the cause for both exceptions.
 - Also 2 JUint test were added to make sure this is working.

---

### For all changes:

- [x] **(N/A)** If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)?

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/distributed/internal/DistributionImpl.java
+++ b/geode-core/src/main/java/org/apache/geode/distributed/internal/DistributionImpl.java
@@ -181,9 +181,9 @@ public class DistributionImpl implements Distribution {
       throw new GemFireSecurityException(e.getMessage(),
           e);
     } catch (MembershipConfigurationException e) {
-      throw new GemFireConfigException(e.getMessage());
+      throw new GemFireConfigException(e.getMessage(), e.getCause());
     } catch (MemberStartupException e) {
-      throw new SystemConnectException(e.getMessage());
+      throw new SystemConnectException(e.getMessage(), e.getCause());
     } catch (RuntimeException e) {
       logger.error("Unexpected problem starting up membership services", e);
       throw new SystemConnectException("Problem starting up membership services: " + e.getMessage()
--- a/geode-core/src/main/java/org/apache/geode/distributed/internal/DistributionImpl.java
+++ b/geode-core/src/main/java/org/apache/geode/distributed/internal/DistributionImpl.java
@@ -181,9 +181,9 @@ public class DistributionImpl implements Distribution {
       throw new GemFireSecurityException(e.getMessage(),
           e);
     } catch (MembershipConfigurationException e) {
-      throw new GemFireConfigException(e.getMessage(), e.getCause());
+      throw new GemFireConfigException("Problem configuring membership services", e);
     } catch (MemberStartupException e) {
-      throw new SystemConnectException(e.getMessage(), e.getCause());
+      throw new SystemConnectException("Problem starting up membership services", e);
     } catch (RuntimeException e) {
       logger.error("Unexpected problem starting up membership services", e);
       throw new SystemConnectException("Problem starting up membership services: " + e.getMessage()
--- a/geode-core/src/test/java/org/apache/geode/distributed/internal/DistributionTest.java
+++ b/geode-core/src/test/java/org/apache/geode/distributed/internal/DistributionTest.java
@@ -15,6 +15,7 @@
 package org.apache.geode.distributed.internal;
 
 import static org.apache.geode.distributed.internal.DistributionImpl.EMPTY_MEMBER_ARRAY;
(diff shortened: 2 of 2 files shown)


geode-core/src/test/java/org/apache/geode/distributed/internal/DistributionTest.java
```diff
@@ -204,4 +207,30 @@ public void testSendToEmptyListIsRejected() throws Exception {
     distribution.send(emptyList, m);
     verify(membership, never()).send(any(), any());
   }
+
+  @Test
+  public void testExceptionNestedOnStartConfigError() throws Exception {
+    Throwable cause = new RuntimeException("Exception cause");
+    Throwable exception = new MembershipConfigurationException("Test exception", cause);
+    doThrow(exception).when(membership).start();
+    try {
```
moleske, line 216: JUnit has [ExpectedException](https://junit.org/junit4/javadoc/4.12/org/junit/rules/ExpectedException.html) which I think works better than `try` `catch` statements.  There are some examples in the Geode code base if you want some inspiration

gaussianrecurrence: Thanks for the comment! TBH I’ve never used JUnit or mockito before, so any comments are welcome!

moleske: Always glad to provide new learning opportunities for folks 😄
Plus no time like the present to learn new things!

jdeppe-pivotal: AssertJ also has a very convenient `assertThatThrownBy()` which lets you assert on the type of exception thrown as well as specific messages, causes, etc. There are numerous examples throughout the test code.

gaussianrecurrence: Finally went for AssertJ choice. Now the test is more more clear. Thanks!

geode-core/src/main/java/org/apache/geode/distributed/internal/DistributionImpl.java
```diff
@@ -181,9 +181,9 @@ public void start() {
       throw new GemFireSecurityException(e.getMessage(),
           e);
     } catch (MembershipConfigurationException e) {
-      throw new GemFireConfigException(e.getMessage());
+      throw new GemFireConfigException(e.getMessage(), e.getCause());
```
jdeppe-pivotal, line 184: You should just use the exception here instead of `e.getCause()` otherwise you're discarding one of the stack frames in the exception chain.

geode-core/src/main/java/org/apache/geode/distributed/internal/DistributionImpl.java
```diff
@@ -181,9 +181,9 @@ public void start() {
...
     } catch (MemberStartupException e) {
-      throw new SystemConnectException(e.getMessage());
+      throw new SystemConnectException(e.getMessage(), e.getCause());
```
jdeppe-pivotal, line 186: Ditto

jdeppe-pivotal (approved):

geode-core/src/main/java/org/apache/geode/distributed/internal/DistributionImpl.java
gaussianrecurrence: Wouldn't the message then be duplicated? I am asking out of ignorance :S
I mean probably it should be something like:
```java
throw new GemFireConfigException("Configuration exception while starting up distribution", e);
```
?

jdeppe-pivotal: The actual _message_ (string) would be duplicated, but that's OK. You wouldn't be losing part of the full stack trace however. It's not a big deal since you're typically only concerned with the root cause, but it's always better to leave information which may be helpful for debugging.
And, yes your example `throw` is what I'd suggest. That's also what happens just a few lines higher up.

gaussianrecurrence:
As it seems concourse is stating Jetty9CachingClientServerTest distributedTest is failing. I've run it multiple times on my local machine and it seems to be actually working. This seems like an sporious failure right? Is there any way in which I can retrigger the concourse execution?

Thanks!

jdeppe-pivotal:
Failures are not related to this PR, so it is OK to merge (there is some other issue with `Jetty9CachingClientServerTest`. To re-trigger you can always push an empty commit to your branch: `get commit --allow-empty -m "Trigger CI"`.

PR #5732, opened 2020-11-11
GEODE-8694: use redis tests with patch

nonbinaryprogrammer:
use the redis/redis repo's test with a patch containing changes to tests
that fail on unimplemented commands when testing geode-redis.

### For all changes:

### Note:

### DIFF
--- a/ci/scripts/execute_redis_tests.sh
+++ b/ci/scripts/execute_redis_tests.sh
@@ -23,10 +23,10 @@ cd ..
 # Redis commands.  Once all commands needed to run relevant test files are implemented, we hope to
 # use Redis's repo without a patch.
 git clone --config transfer.fsckObjects=false https://github.com/redis/redis.git
-cp geode-redis/src/acceptanceTest/resources/0001-configure-redis-tests.patch redis
+REDIS_PATCH=${PWD}/geode-redis/src/acceptanceTest/resources/0001-configure-redis-tests.patch
 cd redis
 git checkout origin/5.0
-git am < 0001-configure-redis-tests.patch
+git apply ${REDIS_PATCH}
 
 export JAVA_HOME=${JAVA_TEST_PATH}
 
@@ -41,7 +41,7 @@ failCount=0
 
 ./runtest --host 127.0.0.1 --port 6380 --single unit/auth
 
-((failCount+=$?))
+((failCount += $?))
 
 
 ../geode-assembly/build/install/apache-geode/bin/gfsh start server \
@@ -55,6 +55,6 @@ failCount=0
 --single unit/type/hash --single unit/type/string \
 --single unit/quit --single unit/pubsub
 
-((failCount+=$?))
+((failCount += $?))
 
-exit $failCount
+exit ${failCount}
--- a/ci/scripts/execute_redis_tests.sh
+++ b/ci/scripts/execute_redis_tests.sh
@@ -19,12 +19,14 @@
 
 cd ..
 
-# We are currently using a personal fork for this repo because our code does not implement all
+# We are currently using a patched version of this repo because our code does not implement all
 # Redis commands.  Once all commands needed to run relevant test files are implemented, we hope to
-# use Redis's repo instead.
-git clone --config transfer.fsckObjects=false https://github.com/prettyClouds/redis.git
+# use Redis's repo without a patch.
+git clone --config transfer.fsckObjects=false https://github.com/redis/redis.git
+cp geode-redis/src/acceptanceTest/resources/0001-configure-redis-tests.patch redis
 cd redis
-git checkout tests-geode-redis
+git checkout origin/5.0
+git am < 0001-configure-redis-tests.patch
 
 export JAVA_HOME=${JAVA_TEST_PATH}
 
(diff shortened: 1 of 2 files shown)


onichols-pivotal (approved):
looks solid!

onichols-pivotal:
the failure in the Redis PR check is probably because the PR pipeline picked up only part of your change (the port number in the rule) but not the other half (this is intentional for security reasons).  Once merged all should be well.

nonbinaryprogrammer:
> the failure in the Redis PR check is probably because the PR pipeline picked up only part of your change (the port number in the rule) but not the other half (this is intentional for security reasons). Once merged all should be well.

Thanks for the context, Owen!

ci/scripts/execute_redis_tests.sh
```diff
@@ -49,8 +51,8 @@ failCount=0
   --redis-port=6379 \
   --redis-bind-address=127.0.0.1
 
-./runtest --host 127.0.0.1 --port 6379 --single unit/type/set --single unit/expire \
---single unit/type/hash --single unit/type/string --single unit/type/srandmember \
+./runtest --host 127.0.0.1 --port 6380 --single unit/type/set --single unit/expire \
```
sabbey37, line 54: We should change this back to 6379

geode-redis/src/commonTest/java/org/apache/geode/redis/GeodeRedisServerRule.java
```diff
@@ -43,7 +43,7 @@ public GeodeRedisServerRule() {
   @Override
   protected void before() {
     cache = cacheFactory.create();
-    server = new GeodeRedisServer("localhost", 0, (InternalCache) cache);
+    server = new GeodeRedisServer("localhost", 6380, (InternalCache) cache);
```
sabbey37, line 46: Change this back to `0`

geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/GeodeServerRunTest.java
```diff
@@ -29,7 +28,8 @@
   public static GeodeRedisServerRule server = new GeodeRedisServerRule();
 
   @Test
-  @Ignore("This is a no-op test to conveniently run redis api for geode server for local development/testing purposes")
```
sabbey37, line 32: This should be ignored.

sabbey37 (changes requested):
Just a few things that we altered while testing need to be changed back.

ci/scripts/execute_redis_tests.sh
```diff
@@ -19,12 +19,14 @@
...
-# We are currently using a personal fork for this repo because our code does not implement all
+# We are currently using a patched version of this repo because our code does not implement all
 # Redis commands.  Once all commands needed to run relevant test files are implemented, we hope to
-# use Redis's repo instead.
-git clone --config transfer.fsckObjects=false https://github.com/prettyClouds/redis.git
+# use Redis's repo without a patch.
+git clone --config transfer.fsckObjects=false https://github.com/redis/redis.git
+cp geode-redis/src/acceptanceTest/resources/0001-configure-redis-tests.patch redis
 cd redis
 git checkout tests-geode-redis
```
sabbey37, line 28: Correct me if I'm missing something, but I don't think this branch is in the `redis` repo.  I guess we could check out the 5.0 branch.  Have you tried this sequence locally?

nonbinaryprogrammer:
thanks @sabbey37 for all your suggested changes! I forgot to check what other changes I was committing. oops

ci/scripts/execute_redis_tests.sh
```diff
@@ -19,12 +19,14 @@
...
 # Redis commands.  Once all commands needed to run relevant test files are implemented, we hope to
-# use Redis's repo instead.
-git clone --config transfer.fsckObjects=false https://github.com/prettyClouds/redis.git
+# use Redis's repo without a patch.
+git clone --config transfer.fsckObjects=false https://github.com/redis/redis.git
+cp geode-redis/src/acceptanceTest/resources/0001-configure-redis-tests.patch redis
 cd redis
-git checkout tests-geode-redis
+git checkout origin/5.0
+git am < 0001-configure-redis-tests.patch
```
sabbey37, line 29: It seems like the application of the patch is failing.  Could be related to this error:
```
*** Please tell me who you are.

Run

  git config --global user.email "you@example.com"
  git config --global user.name "Your Name"

to set your account's default identity.
Omit --global to set the identity only in this repository.

fatal: empty ident name (for <geode@heavy-lifter-bc46dab9-e3b2-587d-b59e-a07a7b894615.c.apachegeode-ci.internal>) not allowed
```
...though I tried removing my identity and running it locally with a fresh clone of redis and it worked.  Maybe an identity is required to apply a patch on heavy-lifter?

sabbey37 (changes requested):

ci/scripts/execute_redis_tests.sh
sabbey37: Ok, after hijacking the pipeline, if you set
```
git config --global user.email "you@example.com"
git config --global user.name "Your Name"
```
before trying to apply the patch, it will be applied successfully.  Not sure what you should set it to though?

nonbinaryprogrammer: thanks @sabbey37! I'll do some more testing on that

sabbey37 (approved):

ci/scripts/execute_redis_tests.sh
jdeppe-pivotal: Switched to using `git apply` which obviates the need to configure an email.

PR #5743, opened 2020-11-12
GEODE-8623: Retry getting local host if it fails.

jinmeiliao:
This is a continuation of the work done by https://github.com/apache/geode/pull/5680

### DIFF
--- a/geode-common/src/main/java/org/apache/geode/internal/Retry.java
+++ b/geode-common/src/main/java/org/apache/geode/internal/Retry.java
@@ -14,7 +14,6 @@
  */
 package org.apache.geode.internal;
 
-import static java.util.concurrent.TimeUnit.MILLISECONDS;
 import static java.util.concurrent.TimeUnit.NANOSECONDS;
 
 import java.util.concurrent.TimeUnit;
@@ -32,18 +31,21 @@ public class Retry {
   interface Timer {
     long nanoTime();
 
-    void sleep(long sleepTime, TimeUnit sleepTimeUnit) throws InterruptedException;
+    void sleep(long sleepTimeInNano) throws InterruptedException;
   }
 
-  private static class SteadyTimer implements Timer {
+  static class SteadyTimer implements Timer {
     @Override
     public long nanoTime() {
       return System.nanoTime();
     }
 
     @Override
-    public void sleep(long sleepTime, TimeUnit sleepTimeUnit) throws InterruptedException {
-      Thread.sleep(MILLISECONDS.convert(sleepTime, sleepTimeUnit));
+    public void sleep(long sleepTimeInNano) throws InterruptedException {
+      // avoid throwing IllegalArgumentException
+      if (sleepTimeInNano > 0) {
+        Thread.sleep(NANOSECONDS.toMillis(sleepTimeInNano));
+      }
     }
   }
 
@@ -82,7 +84,9 @@ public class Retry {
       if (predicate.test(value)) {
         return value;
       } else {
-        timer.sleep(interval, intervalUnit);
+        long sleepTimeInNano =
+            Math.min(NANOSECONDS.convert(interval, intervalUnit), until - timer.nanoTime());
+        timer.sleep(sleepTimeInNano);
       }
     } while (timer.nanoTime() < until);
 
--- a/geode-common/src/main/java/org/apache/geode/internal/Retry.java
+++ b/geode-common/src/main/java/org/apache/geode/internal/Retry.java
@@ -42,9 +42,10 @@ public class Retry {
 
     @Override
     public void sleep(long sleepTimeInNano) throws InterruptedException {
+      long millis = NANOSECONDS.toMillis(sleepTimeInNano);
       // avoid throwing IllegalArgumentException
-      if (sleepTimeInNano > 0) {
-        Thread.sleep(NANOSECONDS.toMillis(sleepTimeInNano));
+      if (millis > 0) {
+        Thread.sleep(millis);
       }
     }
   }
--- a/geode-common/src/main/java/org/apache/geode/internal/Retry.java
+++ b/geode-common/src/main/java/org/apache/geode/internal/Retry.java
@@ -14,12 +14,15 @@
  */
 package org.apache.geode.internal;
 
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
 import static java.util.concurrent.TimeUnit.NANOSECONDS;
 
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.function.Predicate;
 import java.util.function.Supplier;
+
+import org.apache.geode.annotations.VisibleForTesting;
 
 /**
  * Utility class for retrying operations.
@@ -28,12 +31,19 @@ public class Retry {
 
   interface Clock {
     long nanoTime();
+
+    void sleep(long sleepTime, TimeUnit sleepTimeUnit) throws InterruptedException;
   }
 
   private static class SteadyClock implements Clock {
     @Override
     public long nanoTime() {
       return System.nanoTime();
+    }
+
+    @Override
+    public void sleep(long sleepTime, TimeUnit sleepTimeUnit) throws InterruptedException {
+      Thread.sleep(MILLISECONDS.convert(sleepTime, sleepTimeUnit));
     }
   }
 
@@ -42,21 +52,29 @@ public class Retry {
   /**
    * Try the supplier function until the predicate is true or timeout occurs.
    *
+   * @param timeout to retry for
+   * @param interval time between each try
+   * @param timeUnit to retry for
    * @param supplier to execute until predicate is true or times out
    * @param predicate to test for retry
-   * @param timeout to retry for
-   * @param timeUnit to retry for
    * @param <T> type of return value
    * @return value from supplier after it passes predicate or times out.
    */
-  public static <T> T tryFor(final Supplier<T> supplier, final Predicate<T> predicate,
-      final long timeout, final TimeUnit timeUnit) throws TimeoutException {
-    return tryFor(supplier, predicate, timeout, timeUnit, steadyClock);
+  public static <T> T tryFor(long timeout,
+      long interval,
+      TimeUnit timeUnit,
+      Supplier<T> supplier,
+      Predicate<T> predicate) throws TimeoutException, InterruptedException {
+    return tryFor(timeout, interval, timeUnit, supplier, predicate, steadyClock);
   }
 
-  static <T> T tryFor(final Supplier<T> supplier, final Predicate<T> predicate,
-      final long timeout, final TimeUnit timeUnit, final Clock clock)
-      throws TimeoutException {
+  @VisibleForTesting
+  static <T> T tryFor(long timeout,
+      long interval,
+      TimeUnit timeUnit,
+      Supplier<T> supplier,
+      Predicate<T> predicate,
+      Clock clock) throws TimeoutException, InterruptedException {
     long until = clock.nanoTime() + NANOSECONDS.convert(timeout, timeUnit);
 
     T value;
@@ -64,10 +82,11 @@ public class Retry {
       value = supplier.get();
       if (predicate.test(value)) {
         return value;
+      } else {
+        clock.sleep(interval, timeUnit);
       }
     } while (clock.nanoTime() < until);
 
     throw new TimeoutException();
   }
-
 }
--- a/geode-common/src/main/java/org/apache/geode/internal/Retry.java
+++ b/geode-common/src/main/java/org/apache/geode/internal/Retry.java
@@ -1,0 +1,73 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.apache.geode.internal;
+
+import static java.util.concurrent.TimeUnit.NANOSECONDS;
+
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.function.Predicate;
+import java.util.function.Supplier;
+
+/**
+ * Utility class for retrying operations.
+ */
+public class Retry {
+
+  interface Clock {
+    long nanoTime();
+  }
+
+  private static class SteadyClock implements Clock {
+    @Override
+    public long nanoTime() {
+      return System.nanoTime();
+    }
+  }
+
+  private static final SteadyClock steadyClock = new SteadyClock();
+
+  /**
+   * Try the supplier function until the predicate is true or timeout occurs.
+   *
+   * @param supplier to execute until predicate is true or times out
+   * @param predicate to test for retry
+   * @param timeout to retry for
+   * @param timeUnit to retry for
+   * @param <T> type of return value
+   * @return value from supplier after it passes predicate or times out.
+   */
+  public static <T> T tryFor(final Supplier<T> supplier, final Predicate<T> predicate,
+      final long timeout, final TimeUnit timeUnit) throws TimeoutException {
+    return tryFor(supplier, predicate, timeout, timeUnit, steadyClock);
+  }
+
+  static <T> T tryFor(final Supplier<T> supplier, final Predicate<T> predicate,
+      final long timeout, final TimeUnit timeUnit, final Clock clock)
+      throws TimeoutException {
+    long until = clock.nanoTime() + NANOSECONDS.convert(timeout, timeUnit);
+
+    T value;
+    do {
+      value = supplier.get();
+      if (predicate.test(value)) {
+        return value;
+      }
+    } while (clock.nanoTime() < until);
+
+    throw new TimeoutException();
+  }
+
+}
--- a/geode-common/src/main/java/org/apache/geode/internal/Retry.java
+++ b/geode-common/src/main/java/org/apache/geode/internal/Retry.java
@@ -29,13 +29,13 @@ import org.apache.geode.annotations.VisibleForTesting;
  */
 public class Retry {
 
-  interface Clock {
+  interface Timer {
     long nanoTime();
 
     void sleep(long sleepTime, TimeUnit sleepTimeUnit) throws InterruptedException;
   }
 
-  private static class SteadyClock implements Clock {
+  private static class SteadyTimer implements Timer {
     @Override
     public long nanoTime() {
       return System.nanoTime();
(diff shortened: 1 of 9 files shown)


demery-pivotal (approved):

geode-common/src/main/java/org/apache/geode/internal/Retry.java
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  private static class SteadyClock implements Clock {
+    @Override
+    public long nanoTime() {
+      return System.nanoTime();
+    }
+
+    @Override
+    public void sleep(long sleepTime, TimeUnit sleepTimeUnit) throws InterruptedException {
```
pivotal-jbarrett, line 45: Sleeping doesn't feel like the responsibility of a clock.

geode-common/src/main/java/org/apache/geode/internal/Retry.java
```diff
@@ -0,0 +1,92 @@
...
+      TimeUnit timeUnit,
+      Supplier<T> supplier,
+      Predicate<T> predicate) throws TimeoutException, InterruptedException {
+    return tryFor(timeout, interval, timeUnit, supplier, predicate, steadyClock);
+  }
+
+  @VisibleForTesting
+  static <T> T tryFor(long timeout,
+      long interval,
+      TimeUnit timeUnit,
```
pivotal-jbarrett, line 74: Is TimeUnit intended to cover both time values? I think its more readable to have TimeUnit param for each so it reads as "(1, SECOND, 4,  SECOND)" rather than "(1, 4, SECOND)".

pivotal-jbarrett (changes requested):

geode-common/src/main/java/org/apache/geode/internal/Retry.java
jinmeiliao: We can give it another name, like Timer? This is an internal interface made accessible for testing purposes, so I am not too concerned about the naming or break it into another interface.

geode-common/src/main/java/org/apache/geode/internal/Retry.java
jinmeiliao: It is meant to cover both time values, but if you think it's readable otherwise....

geode-common/src/main/java/org/apache/geode/internal/Retry.java
```diff
@@ -47,45 +47,44 @@ public void sleep(long sleepTime, TimeUnit sleepTimeUnit) throws InterruptedExce
...
+    long until = timer.nanoTime() + NANOSECONDS.convert(timeout, timeoutUnit);
 
     T value;
     do {
       value = supplier.get();
       if (predicate.test(value)) {
         return value;
       } else {
-        clock.sleep(interval, timeUnit);
+        timer.sleep(interval, intervalUnit);
```
pivotal-jbarrett, line 85: The sleep should happen for min(interval, until - now) otherwise you could sleep past the timeout.

pivotal-jbarrett (changes requested):

geode-common/src/main/java/org/apache/geode/internal/Retry.java
jinmeiliao: Yes, but this extra calculation only affects the last iteration, and either we sleep past the timeout or exactly at the timeout, we would exit the loop anyway. I wouldn't mind sleeping for some extra time in the last iteration if we can make the code simpler.

pivotal-jbarrett: Also, we sleep at the end only to delay the return of the failed predicate state. Sleeping min time and then continue after guards against sleeping past the timeout and gives the opportunity of the predicate to go pass before returning. No point in sleeping if we are just going to fall through the while block immediately, we aren't trying to delay the return but rather the execution of the code block.

pivotal-jbarrett (changes requested):

geode-common/src/main/java/org/apache/geode/internal/Retry.java
Bill: We've been using two separate interfaces in other places in Geode where we do this, and I'd recommend it here too. Lemme go find an example… Ah yes, have a look at `PrimaryHandler` which has `Sleeper` and `MillisecondProvider` interfaces. Hmm looks like `PrimaryHandler.processRequest()` might benefit from this `Retry` class eventually!

geode-common/src/main/java/org/apache/geode/internal/inet/LocalHostUtil.java
```diff
@@ -54,19 +59,27 @@
...
 
-  static {
+  /**
+   * Resolves local host. Will retry if resolution fails.
+   *
+   * @return local host if resolved otherwise null.
+   */
+  private static InetAddress tryToResolveLocalHost() {
+    try {
+      return tryFor(60, SECONDS, 1, SECONDS, LocalHostUtil::resolveLocalHost, Objects::nonNull);
```
Bill, line 76: so cool!

geode-common/src/main/java/org/apache/geode/internal/Retry.java
jinmeiliao: I am not sure I follow here. if we sleep past the timeout then we won't execute the predicate anymore, it would just end the while loop with TimeoutException.

pivotal-jbarrett: The last iteration is the problem, why sleep to delay the return. Either just return because it would timeout at its regular interval or reduce the interval and try again before timing out. Either way there is no point I delaying the results in the last iteration as this implementation does.

jinmeiliao: what I am saying is that even it reduces the interval it wouldn't try again, it will always just throw TimeoutException in the end.

Bill: The example @jinmeiliao gave covers use-cases where the number of iterations is potentially large. Those are cases where time to run predicate + time to run supplier + sleep _interval_ << _timeout_.

There are other use-cases where that relation does not hold. In those cases perhaps we will only execute the body of the loop once. Imagine time to run predicate + time to run supplier + sleep _interval_ === _timeout_ i.e. the timeout interval is roughly of the same magnitude as the other work to be done each time through the loop. In that case, @pivotal-jbarrett's concern about overshooting the timeout seem more important. This second scenario seems to me to be worth the effort of the added complexity of calculating `sleepTime = min(interval, until - now)` and returning immediately if that's less-than-or-equal-to zero.

jinmeiliao: @Bill Your example is a special case where the LAST iteration is the ONLY iteration. The extra computing only makes a difference in the LAST iteration and only when it throws a TimeoutException.

I will try to make my case one more time, if after this, you guys still insist, I will rest my case.

1. Without the extra calculation, we would NEVER delay a normal return, we will only delay a TimeoutException by a little bit (maximum the rest period). Since we are going to timeout anyway, I don't really care about this little extra wait time.
2. The extra calculation only make a difference in the last iteration when it's about to reach a timeout, for all the previous loops, it's a waste of time and an extra call to the timer.getTime().
3. `until - now` could be negative, so the calculation should really be something like `min(interval, (until - now) < 0 ? 0 : (until - now)` which completely take the beauty and readability out of the code. To me, the detriment overweights the benefit. 

But if you guys still think otherwise, do let me know. I will make the change. @Bill @pivotal-jbarrett

gesterzhou (approved):

geode-common/src/main/java/org/apache/geode/internal/Retry.java
Bill: I don't think any extra call is needed to implement the proposal @jinmeiliao. The call to `timer.nanoTime()` can be done in the body of the loop and used in the body and in the `while()` condition I think. I still feel like the added code complexity is worth it since the whole point of making a stand-alone `Retry` utility like this is to be able to use it in many situations.

jinmeiliao: No, we can't reuse `timer.nanoTime()` since it will be different time at the time of calculation.

I am still not convinced, but I will go ahead and make the change and let you guys decide.

Bill: would something like this work:

image

jinmeiliao: I think will run the supplier and predicate one more time than the previous iteration. 

Using my previous example of 10 seconds timeout and 3 seconds sleep (supplier and predicate execution time 0 and predicate always returns false), I am supposed to throw the TimeoutException after sleeping for the last 1 second , but your implementation will run supplier/predicate one more time and then throw.

Bill: Agreed. With this approach we never return directly after a sleep. (That's what @pivotal-jbarrett meant above when he said "No point in sleeping if we are just going to fall through the while block immediately, we aren't trying to delay the return but rather the execution of the code block") We always return immediately after testing the predicate. It is different, but I don't think it's wrong.

If we are bothering to sleep, it makes sense to give the predicate one last try before returning. Otherwise why did we sleep?

geode-common/src/main/java/org/apache/geode/internal/Retry.java
```diff
@@ -32,18 +31,21 @@
...
       return System.nanoTime();
     }
 
     @Override
-    public void sleep(long sleepTime, TimeUnit sleepTimeUnit) throws InterruptedException {
-      Thread.sleep(MILLISECONDS.convert(sleepTime, sleepTimeUnit));
+    public void sleep(long sleepTimeInNano) throws InterruptedException {
+      // avoid throwing IllegalArgumentException
+      if (sleepTimeInNano > 0) {
+        Thread.sleep(NANOSECONDS.toMillis(sleepTimeInNano));
```
Bill, line 47: I recommend testing the condition _after_ the conversion to milliseconds since e.g. `NANOSECONDS.toMillis(1) == 0`

geode-common/src/main/java/org/apache/geode/internal/Retry.java
```diff
@@ -82,7 +84,9 @@ public void sleep(long sleepTime, TimeUnit sleepTimeUnit) throws InterruptedExce
       if (predicate.test(value)) {
         return value;
       } else {
-        timer.sleep(interval, intervalUnit);
+        long sleepTimeInNano =
+            Math.min(NANOSECONDS.convert(interval, intervalUnit), until - timer.nanoTime());
```
Bill, line 88: I recommend converting this outside the loop since it never changes.

geode-common/src/main/java/org/apache/geode/internal/Retry.java
jinmeiliao: I think I get your point, we already wasted some time, why don't we give it another try? It all boils down to what you think is right or tolerable.

geode-common/src/main/java/org/apache/geode/internal/Retry.java
jinmeiliao: no, it changes, only in the last iteration though.

geode-common/src/main/java/org/apache/geode/internal/Retry.java
jinmeiliao: good point.

geode-common/src/main/java/org/apache/geode/internal/Retry.java
jinmeiliao: you know if you change your implementation a little bit, instead of `if (sleepNanos > 0)` we change it to `if(sleepNanos == intervalNanos)`, I think this would be best.

jinmeiliao: Ok, I think my last commit combines the good points of both implementation. Thanks for all the points and nudges.

geode-common/src/main/java/org/apache/geode/internal/Retry.java
```diff
@@ -0,0 +1,101 @@
...
+
+    T value;
+    for (;;) {
+      value = supplier.get();
+      if (predicate.test(value)) {
+        return value;
+      } else {
+        // if there is still more time left after we sleep for interval period, then sleep and retry
+        // otherwise break out and throw TimeoutException
+        if ((timer.nanoTime() + intervalNano) < until) {
```
Bill, line 91: There is no need to sleep for the full interval (`intervalNano`) if `until - timer.nanoTime()` is smaller.

I recommend (again), as have others, that the code sleep for the minimum time required. We understand that `Thread.sleep()` does not make real-time guarantees. Nevertheless, this code would be more robust and accurate if it was coded that way.

Something like this:
image

Bill (changes requested):
We're so close now. `Retry` is shaping up nicely. It's not quite there though.

geode-common/src/main/java/org/apache/geode/internal/Retry.java
```diff
@@ -0,0 +1,101 @@
...
+
+/**
+ * Utility class for retrying operations.
+ */
+public class Retry {
+
+  interface Timer {
+    long nanoTime();
+
+    void sleep(long sleepTimeInNano) throws InterruptedException;
```
Bill, line 34: By making `sleep()` take nanosecond units, every implementation (like the one below) has to make the conversion and handle any problems.

I recommend making this method take millisecond units since that's what `Thread.sleep()` takes. This simplifies every implementation, and, perhaps more importantly, moves the error-handling logic up into the `tryFor()` method, making that method more robust.

geode-common/src/main/java/org/apache/geode/internal/Retry.java
jinmeiliao: @Bill per our offline meetings, can I merge this now?

geode-common/src/main/java/org/apache/geode/internal/Retry.java
Bill: ah yes!

geode-common/src/main/java/org/apache/geode/internal/Retry.java
Bill: I think a user of the `Retry` class would (rightly) expect `tryFor()` to keep trying if `timeout` has not yet been reached.

Computing `sleepNanos` gives good results when `intervalUnit` is greater than (`timeoutUnit` - supplier time - predicate time). Without `sleepNanos`, that scenario results in _no retries at all_ which I think most users of the `Retry` class would find counterintuitive.

Failing to compute `sleepNanos` introduces an error. That error is directly proportional to `interval` and the relative error (error / `timeout`) grows as `interval` grows relative to `timeout`.

Computing `sleepNanos` is simple and cheap. It gives better accuracy and meets user expectations in all cases. Unless you foresee some problem computing `sleepNanos` I still recommend doing it.

jinmeiliao: > I think a user of the `Retry` class would (rightly) expect `tryFor()` to keep trying if `timeout` has not yet been reached.
> 
> Computing `sleepNanos` gives good results when `intervalUnit` is greater than (`timeoutUnit` - supplier time - predicate time). Without `sleepNanos`, that scenario results in _no retries at all_ which I think most users of the `Retry` class would find counterintuitive.

So if say timeout is 5 second, supplier and predicate takes 3 seconds, and interval is 3 seconds, I won't expect there would be any retries at all. I don't think this is counterintuitive.


> Failing to compute `sleepNanos` introduces an error. That error is directly proportional to `interval` and the relative error (error / `timeout`) grows as `interval` grows relative to `timeout`.
> 
> Computing `sleepNanos` is simple and cheap. It gives better accuracy and meets user expectations in all cases. Unless you foresee some problem computing `sleepNanos` I still recommend doing it.

Bill: A user of `tryFor()` does not know (usually cannot know) how much clock time their supplier and predicate will take. So they won't be able to do the reasoning you just did there. I think what they'll usually do is try to pick `timeout` much larger than the (supplier+predicate) time.

We can presume that they will usually try to set `timeout` long enough so that at least one retry (two tries) can happen. If we let`timeout` mean: keep trying until `timeout` has elapsed, then the "error" term I referred to in my previous comment is the difference between `timeout` and the actual time of the final attempt. As a user I'd expect `tryFor()` to make a reasonable effort to minimize that error.

Maybe it would help if, in my proposal,`interval` was interpreted as "typical sleep time". For all but the last iteration `interval` is the actual sleep time. Usually, in the last interval (under my proposal) `sleepNanos` will be less than `maximumSleepTime`.

Is there some reason why you object to (usually) sleeping a shorter time before the final try? Do you think that calling `min()` and capturing the result in a variable (`sleepNanos`) is overly burdensome? Do you see any value in making some effort to make the timing of the actual final attempt correspond to `timeout`?

jinmeiliao: what's the point of sleeping if we know we are going to reach timeout after sleeping and throw a TimeoutException?

Bill: After each `sleep()` (in my proposal) we try go back to the top of the loop and execute the supplier and the predicate again. We only `break` (and throw the exception) if `! (sleepNanos > 0)`)

jinmeiliao: > A user of `tryFor()` does not know (usually cannot know) how much clock time their supplier and predicate will take. So they won't be able to do the reasoning you just did there. I think what they'll usually do is try to pick `timeout` much larger than the (supplier+predicate) time.
> 
> We can presume that they will usually try to set `timeout` long enough so that at least one retry (two tries) can happen. If we let`timeout` mean: keep trying until `timeout` has elapsed, then the "error" term I referred to in my previous comment is the difference between `timeout` and the actual time of the final attempt. As a user I'd expect `tryFor()` to make a reasonable effort to minimize that error.

No matter how long the `timeout` to be or how short the `interval` to be, a user can NOT "reasonably" assume two tries should happen at all. Even if I set timeout to be 100 seconds, and interval to be 1 second, there are plenty of chances that I could try only one time and the timeout period has already elapsed.

> 
> Maybe it would help if, in my proposal,`interval` was interpreted as "typical sleep time". For all but the last iteration `interval` is the actual sleep time. Usually, in the last interval (under my proposal) `sleepNanos` will be less than `maximumSleepTime`.
> 
> Is there some reason why you object to (usually) sleeping a shorter time before the final try? Do you think that calling `min()` and capturing the result in a variable (`sleepNanos`) is overly burdensome? Do you see any value in making some effort to make the timing of the actual final attempt correspond to `timeout`?

My objection is : 
1. there is no point in sleeping if we know we are going to timeout at the end of sleep.
2. It feels even more counter-intuitive that we know we've already reached the end of the timeout period and then still go ahead and try again. It's actually wrong.

Bill: I don't understand (1). In my proposal we never time out at the end of a sleep. After a sleep we always try the supplier and predicate again. In cases where the predicate returns `false` forever, we would always sleep and then try the predicate and then `break` (and throw the exception.)

In cases where the predicate returns `false` forever, our final attempt is made as close to the timeout as possible (given the limitations of Java thread scheduling). This is exactly what a user expects.

jinmeiliao: > In cases where the predicate returns `false` forever, our final attempt is made as close to the timeout as possible (given the limitations of Java thread scheduling). This is exactly what a user expects.

In your proposal, at your final attempt, you calculated what's the time left till timeout, then you sleep till timeout. Our differences lie at this point afterwards, I think that since we are already at timeout point we should throw exception, but you think we should try again. I don't think "trying again" is "exactly what a user expects", since I certainly didn't expect that.

Let's just list out the choices here and decide what's the "most" reasonable thing to do. Say our interval is 3 seconds, and there are only 2 seconds left after our last call to the supplier and the predicate is still false,  At this point, we can do either of the following:
1. throw timeout immediately (my implementation)
2. sleep for 2 seconds till timeout and throw timeout (since 2 is what's left till timeout)
3. sleep for 3 seconds and throw timeout (since 3 is what user has specified as interval)
4. sleep for 2 seconds till timeout and then try again (your suggestion)

In the above 4, my opinion is 4 is the least reasonable, and 1 has some optimization among the first 3.

geode-common/src/main/java/org/apache/geode/internal/Retry.java
Bill: After talking to @jinmeiliao and thinking about this some more I think the current implementation (nanosecond units) is ok.

geode-common/src/main/java/org/apache/geode/internal/Retry.java
Bill: The usage of `Retry` (as it stands) in this PR, by `tryToResolveLocalHost()` is fine:

`tryFor(60, SECONDS, 1, SECONDS…`

The difference between (1) and (4) in this case is at most:

(1) last retry at 59 seconds
(4) last retry at 60 seconds

If the purpose of this PR were to make the best possible `Retry` utility, I might be inclined to continue the deliberation. But since the purpose of this PR is to retry resolving the localhost address, and the PR does that, my work is done here.

In the interest of forward progress, I withdraw my change request on this PR.

Bill (approved):

PR #5749, opened 2020-11-13
GEODE-8711: Enable SLOWLOG Redis command

ringles:
Enable SLOWLOG Redis command to help support Redis monitoring tools like Redis Insights and Datadog.


geode-redis/src/main/java/org/apache/geode/redis/internal/ParameterRequirements/SlowlogParameterRequirements.java
```diff
@@ -0,0 +1,47 @@
...
+        throw new RedisParametersMismatchException(ERROR_NOT_INTEGER);
+      }
+    }
+  }
+
+  public String unknownSubcommandOrWrongNumberOfArgumentsErrorMessage(String command) {
+    String result;
+    result =
+        String.format("Unknown subcommand or wrong number of arguments for '%s'. Try SLOWLOG HELP.",
+            command);
```
sabbey37, line 44: I think we'd want to put the `command.getStringKey()` here instead of the command (if we want to match what Redis does, not sure if this was discussed and we decided to do something different).

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/server/SlowlogExecutor.java
```diff
@@ -30,7 +27,21 @@
...
+      return RedisResponse.ok();
+    }
+    throw new RedisParametersMismatchException(wrongNumberOfArgumentsErrorMessage());
+  }
+
+  public String wrongNumberOfArgumentsErrorMessage() {
+    String result;
+    result = String
+        .format("Unknown subcommand or wrong number of arguments for 'len'. Try SLOWLOG HELP.");
+    return result;
```
sabbey37, line 45: It seems kind of weird that we have this hard coded to `len`... right now if I type in `slowlog sjls`, I get `ERR Unknown subcommand or wrong number of arguments for 'len'. Try SLOWLOG HELP.` instead of `ERR Unknown subcommand or wrong number of arguments for 'jsls'. Try SLOWLOG HELP.` It also might be good to add this Error to the `RedisConstants` class.

sabbey37 (changes requested):
Have y'all discussed whether or not we want to create specific parameter requirements for each command (I thought for some reason we decided against this route)?  Right now, the things we are handling in `SlowlogParameterRequirements` could be handled in the executor (like other commands), or if we want to go the route of specific parameter requirements, we could potentially avoid throwing the `RedisParametersMismatchException(wrongNumberOfArgumentsErrorMessage());` in the SlowlogExecutor and handle that in the parameter requirements instead.  Also, `SLOWLOG` needs to be added to the list of unsupportedCommands in the `SupportedCommandsJunitTest`

geode-redis/src/main/java/org/apache/geode/redis/internal/RedisCommandType.java
```diff
@@ -265,6 +267,9 @@
   FLUSHDB(new FlushAllExecutor(), UNSUPPORTED, new MaximumParameterRequirements(2, ERROR_SYNTAX)),
   INFO(new InfoExecutor(), UNSUPPORTED, new MaximumParameterRequirements(2, ERROR_SYNTAX)),
   SHUTDOWN(new ShutDownExecutor(), UNSUPPORTED, new MaximumParameterRequirements(2, ERROR_SYNTAX)),
+  SLOWLOG(new SlowlogExecutor(), UNSUPPORTED, new MinimumParameterRequirements(2)
+      .and(new MaximumParameterRequirements(3, ERROR_SYNTAX))
+      .and(new SlowlogParameterRequirements())),
```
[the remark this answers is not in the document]
ringles: Yeah, we refactored the validation considerably. I would agree that using validators for specific commands should be avoided in general. But for specific commands with unique requirements (SLOWLOG takes three subcommands, one of which takes an optional argument, etc.) that a specific validator helps make the code more readable. The validator has responsibility for examining the arguments, and the executor has the responsibility for carrying out the (pre-validated) command.

geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/server/AbstractSlowlogIntegrationTest.java
```diff
@@ -0,0 +1,117 @@
...
+import org.junit.Before;
+import org.junit.Test;
+import redis.clients.jedis.Jedis;
+import redis.clients.jedis.Protocol;
+import redis.clients.jedis.util.Slowlog;
+
+import org.apache.geode.test.awaitility.GeodeAwaitility;
+import org.apache.geode.test.dunit.rules.RedisPortSupplier;
+
+public abstract class AbstractSlowlogIntegrationTest implements RedisPortSupplier {
```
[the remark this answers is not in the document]
ringles: Yup, we've added more tests. Thanks!

geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/server/AbstractSlowlogIntegrationTest.java
```diff
@@ -0,0 +1,117 @@
...
+
+    assertThat(response).isEqualTo("OK");
+  }
+
+  @Test
+  public void shouldThrowException_IfResetGivenAParameter() {
+    assertThatThrownBy(
+        () -> jedis.sendCommand(
+            Protocol.Command.SLOWLOG, "RESET", "Superfluous"))
+                .hasMessageContaining("ERR Unknown subcommand or wrong number of arguments for");
```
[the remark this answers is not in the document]
ringles: Done! Good call!

geode-redis/src/main/java/org/apache/geode/redis/internal/ParameterRequirements/SlowlogParameterRequirements.java
ringles: This was fixed in the refactor.

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/server/SlowlogExecutor.java
ringles: Also fixed when we refactored. Thanks!

geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/server/AbstractSlowlogIntegrationTest.java
```diff
@@ -0,0 +1,146 @@
...
+    jedis.close();
+  }
+
+  abstract int getExposedPort();
+
+  @Test
+  public void shouldReturnEmptyArray_whenGetSubcommandSpecified() {
+    List<Slowlog> actualResult = jedis.slowlogGet();
+
+    assertThat(actualResult.size()).isEqualTo(0);
```
sabbey37, line 55: We could use `assertThat(actualResult).isEmpty()` here (and for the ones below) .  This might be preferable because if the result is not empty, the error message will display the contents of the array rather than just the length.  Sorry I didn't notice this in the first review.

geode-redis/src/main/java/org/apache/geode/redis/internal/ParameterRequirements/SlowlogParameterRequirements.java
```diff
@@ -0,0 +1,62 @@
...
+  public void checkParameters(Command command, ExecutionHandlerContext context) {
+    int numberOfArguments = command.getProcessedCommand().size();
+
+    if (numberOfArguments < 2) {
+      throw new RedisParametersMismatchException(command.wrongNumberOfArgumentsErrorMessage());
+    } else if (numberOfArguments == 2) {
+      confirmKnownSubcommands(command);
+    } else if (numberOfArguments == 3) {
+      confirmArgumentsToGetSubcommand(command);
+    } else if (numberOfArguments > 3) {
```
sabbey37, line 35: I think LGTM is suggesting that this could just be an `else`, instead of an `else if`

geode-redis/src/acceptanceTest/java/org/apache/geode/redis/internal/executor/server/SlowlogNativeRedisAcceptanceTest.java
```diff
@@ -0,0 +1,36 @@
...
+
+  @Override
+  public int getPort() {
+    return redis.getPort();
+  }
+
+  @Override
+  int getExposedPort() {
+    return redis.getExposedPort();
+  }
```
sabbey37, line 35: You can delete this since the method is never used.

geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/server/SlowlogIntegrationTest.java
```diff
@@ -0,0 +1,39 @@
...
+
+  @Override
+  public int getPort() {
+    return server.getPort();
+  }
+
+  @Override
+  public int getExposedPort() {
+    return getPort();
+  }
```
sabbey37, line 38: You can delete this since the method is never used.

geode-redis/README.md
```diff
@@ -200,16 +200,16 @@ start server \
...
-|                    	| SPOP                                               	| CLUSTER NODES                 	|
-|                    	| SRANDMEMBER                                        	| CLUSTER REPLICAS                 	|
-|                    	| SSCAN                                              	| CLUSTER REPLICATE              	|
-|                    	| STRLEN                                             	| CLUSTER RESET                  	|
-|                    	| SUNION                                             	| CLUSTER SAVECONFIG                |
-|                    	| SUNIONSTORE                                        	| CLUSTER SET-CONFIG-EPOCH          |
-|                    	| TIME                                               	| CLUSTER SETSLOT                 	|
-|                    	| UNLINK [1]                                         	| CLUSTER SLAVES                  	|
-|                    	|                                                    	| CLUSTER SLOTS                 	|
+|                    	| SLOWLOG                                            	| CLUSTER MYID                   	|
```
sabbey37, line 203: Good call adding this to the README! We also need to add it to the list of unsupported commands in this file: `redis_api_for_geode.html.md.erb`

geode-redis/src/main/java/org/apache/geode/redis/internal/ParameterRequirements/SlowlogParameterRequirements.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+  }
+
+  private void confirmArgumentsToGetSubcommand(Command command) {
+    if (!command.getStringKey().toLowerCase().equals("get")) {
+      throw new RedisParametersMismatchException(
+          String.format(ERROR_UNKNOWN_SLOWLOG_SUBCOMMAND, command.getStringKey()));
+    }
+    try {
+      Integer.parseInt(new String(command.getProcessedCommand().get(2)));
```
sabbey37, line 56: Looks like this is one of those situations where Redis uses a long instead of an integer. We could also add a test for this.

sabbey37 (changes requested):

geode-redis/src/main/java/org/apache/geode/redis/internal/ParameterRequirements/SlowlogParameterRequirements.java
ringles: Functionally it's the same. I agree with LGTM that it's technically redundant... but I think it's marginally more readable, making clear to a human reading it what the intent is. Perhaps we could add a comment instead?

geode-redis/src/main/java/org/apache/geode/redis/internal/ParameterRequirements/SlowlogParameterRequirements.java
ringles: Test added, and passing.

geode-redis/README.md
ringles: Done, thanks!

geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/server/SlowlogIntegrationTest.java
ringles: Also done.

geode-redis/src/acceptanceTest/java/org/apache/geode/redis/internal/executor/server/SlowlogNativeRedisAcceptanceTest.java
ringles: Done.

geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/server/AbstractSlowlogIntegrationTest.java
```diff
@@ -0,0 +1,146 @@
...
+  public void setUp() {
+    jedis = new Jedis("localhost", getPort(), REDIS_CLIENT_TIMEOUT);
+  }
+
+  @After
+  public void tearDown() {
+    jedis.close();
+  }
+
+  abstract int getExposedPort();
```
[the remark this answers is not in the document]
ringles: Done.

geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/server/AbstractSlowlogIntegrationTest.java
ringles: Done!

geode-redis/src/main/java/org/apache/geode/redis/internal/ParameterRequirements/SlowlogParameterRequirements.java
sabbey37: That totally makes sense. I'm not sure how much LGTM alerts matter? @jdeppe-pivotal do you know?

ringles: Demoted to comment, just to keep LGTM happy.

sabbey37 (approved):

PR #5750, opened 2020-11-14
GEODE-8706: getConnection should get readLock to sync with destroyCon…

gesterzhou:
…nection

### For all changes:

### Note:

### DIFF
--- a/geode-wan/src/main/java/org/apache/geode/internal/cache/wan/GatewaySenderEventRemoteDispatcher.java
+++ b/geode-wan/src/main/java/org/apache/geode/internal/cache/wan/GatewaySenderEventRemoteDispatcher.java
@@ -25,6 +25,7 @@ import org.apache.logging.log4j.Logger;
 
 import org.apache.geode.CancelException;
 import org.apache.geode.GemFireIOException;
+import org.apache.geode.annotations.VisibleForTesting;
 import org.apache.geode.cache.RegionDestroyedException;
 import org.apache.geode.cache.client.ServerConnectivityException;
 import org.apache.geode.cache.client.ServerOperationException;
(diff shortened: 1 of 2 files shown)


kohlmu-pivotal:
Whilst the change LGTM... I cannot help but feel that the whole logic of (un)/locking of the `connectionLifeCycleLock` should really be a little more hidden. This locking logic proliferates throughout the whole `GatewaySenderEventRemoteDispatcher`. Which makes it way too easy to actually miss logic where the connection lock needs to be used.
@boglesby @gesterzhou maybe we should be locking at moving all this logic to another location. 
There is also no indication of WHY this lock is required? Is it protecting the accessing of `connection` properties or is it avoiding the accidental setting of the `connection` field with another instance? It is not clear.

boglesby:
The lock is controlling access to the connection field. The destroyConnection and initializeConnection methods set it so they use the writeLock, and the readAcknowledgement and _dispatchBatch methods access it so they use the readLock. The getConnection method should also use the readLock, but it hasn't up until now with this change.

geode-wan/src/main/java/org/apache/geode/internal/cache/wan/GatewaySenderEventRemoteDispatcher.java
```diff
@@ -295,9 +301,16 @@ public Connection getConnection(boolean startAckReaderThread) throws GatewaySend
...
+      getConnectionLifeCycleLock().readLock().lock();
+      try {
+        needToReconnect = this.connection == null || this.connection.isDestroyed()
+            || this.connection.getServer() == null
+            || !this.connection.getServer().equals(this.sender.getServerLocation());
+      } finally {
+        getConnectionLifeCycleLock().readLock().unlock();
+      }
+      if (needToReconnect) {
         if (logger.isDebugEnabled()) {
```
boglesby, line 314: Does this debug statement need to be moved inside the readLock too since it calls `this.connection.getServer()`?

boglesby (approved):

geode-wan/src/main/java/org/apache/geode/internal/cache/wan/GatewaySenderEventRemoteDispatcher.java
gesterzhou: No need, you see the original author used:
(this.connection == null) ? "null" : ...

That means he realized this connection could be null here

boglesby (approved):

PR #5751, opened 2020-11-14
GEODE-8664: Nest errors in DistributionImpl.start

gaussianrecurrence:
- If while calling DistributionImpl.start there was either a
   MembershipConfigurationException or MemberStartupException exceptions, its
   original cause was not propagated, therefore being unable to properly tackle
   issues on startup.
 - This commit propagates both exceptions.
 - Also 2 junit test were added to make sure this is working.
 - Also 2 DUnit tests were modified in order to integrate with this
   change.

Note: After having to revert the original change in PR #5725 due to some
      issues with the CI system, tests have been sorted out and everything
      is working.

---

### For all changes:

- [ ] (N/A) If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)?

### Note:

### DIFF
--- a/geode-core/src/distributedTest/java/org/apache/geode/distributed/LocatorDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/distributed/LocatorDUnitTest.java
@@ -1001,8 +1001,7 @@ public class LocatorDUnitTest implements Serializable {
       // I guess it can throw this too...
 
     } catch (GemFireConfigException ex) {
-      String s = ex.getMessage();
-      assertThat(s.contains("Locator does not exist")).isTrue();
+      assertThat(ex.getCause().getMessage().contains("Locator does not exist")).isTrue();
     }
   }
 
--- a/geode-core/src/distributedTest/java/org/apache/geode/distributed/LocatorUDPSecurityDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/distributed/LocatorUDPSecurityDUnitTest.java
@@ -49,7 +49,7 @@ public class LocatorUDPSecurityDUnitTest extends LocatorDUnitTest {
       system = getConnectedDistributedSystem(props);
       fail("Should not have reached this line, it should have caught the exception.");
     } catch (GemFireConfigException e) {
-      assertThat(e.getMessage()).contains("Rejecting findCoordinatorRequest");
+      assertThat(e.getCause().getMessage()).contains("Rejecting findCoordinatorRequest");
     }
   }
 }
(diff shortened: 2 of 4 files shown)


kohlmu-pivotal (approved):
LGTM

echobravopapa (dismissed):
I think this violates the intent that `membership` exceptions should not escape the module

kohlmu-pivotal:
> I think this violates the intent that `membership` exceptions should not escape the module

@echobravopapa I love the idea that modules/components don't leak anything outside of itself. But if looking at the `DistribtionImpl` the exceptions that you don't want to leak "out", have already leaked into geode-core. So, not sure how we are trying to stop the leaking.

It is not like there is even a violation of public -> internal package. The usage are in 2 internal classes.

If the intent that exceptions do not escape the module, then the current implementation needs to be addressed. As the current implementation is a sieve. The exceptions have already leaked out!

kohlmu-pivotal:
> I think this violates the intent that `membership` exceptions should not escape the module

@echobravopapa if the intent is not for the exceptions leaking out, could you possibly suggest an alternate implementations on how to resolve the issue that @gaussianrecurrence has raised?

echobravopapa:
it's unclear what the actual issue this change is fixing...  here is an example of a recent bug we fixed caused by exceptions escaping... https://issues.apache.org/jira/browse/GEODE-8697

kohlmu-pivotal:
@gaussianrecurrence Would it be possible for the you to reproduce the issue that  you are seeing (and trying to fix) and update the JIRA with this information?

That will hopefully give the reviewers some context as to WHAT exactly you are trying to fix.

gaussianrecurrence:
Sorry @kohlmu-pivotal and @echobravopapa but I have been quite busy these days.
The rationale behind this change is described already in the Jira description, probably with not too much detail.
But quickly summarizing, while starting a locator if there is an exception having to do with JGroups, the cause is not included in the top exception, hence loosing the context on what's happening. Hopefully this clarifies the intention of this fix.

Regarding about whether or not to leak the exception, I don't have an strong point as Java is not my main language. But for what I can tell, there are these exceptions which are internal and we wouldn't want these to be exposed, that's why they are being wrapper as SystemConnectException and GemFireConfigException. If that's the case I don't have any problem with that, 

Still,  unless I am missing something important here, what I believe is that the exception cause should be there, otherwise troubleshooting these kind of issues could be a real pain.

bschuchardt (approved):
This looks fine.  The original intent of the DistributionImpl code was to throw known exceptions.  It's okay for these exceptions to have "causes" that are geode-membership or jgroups throwables, especially if this helps with diagnosing problems.

bschuchardt:
Don't forget to resolve the JIRA ticket

PR #5756, opened 2020-11-17
Geode 8717 info command return only specified section

ringles:
Update INFO command to allow specifying new fields individually.

### DIFF
--- a/geode-redis/src/acceptanceTest/java/org/apache/geode/redis/internal/executor/server/SlowlogNativeRedisAcceptanceTest.java
+++ b/geode-redis/src/acceptanceTest/java/org/apache/geode/redis/internal/executor/server/SlowlogNativeRedisAcceptanceTest.java
@@ -28,4 +28,9 @@ public class SlowlogNativeRedisAcceptanceTest extends AbstractSlowlogIntegration
   public int getPort() {
     return redis.getPort();
   }
+
+  @Override
+  int getExposedPort() {
+    return redis.getExposedPort();
+  }
 }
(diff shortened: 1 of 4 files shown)


jdeppe-pivotal (approved):

geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/server/AbstractInfoIntegrationTest.java
```diff
@@ -98,71 +152,130 @@ public void shouldReturnClusterEnabledProperty() {
...
+    allProperties.addAll(KEYSPACE_PROPERTIES);
+    allProperties.addAll(MEMORY_PROPERTIES);
+    allProperties.addAll(CLIENTS_PROPERTIES);
+    allProperties.addAll(CLUSTER_PROPERTIES);
+    allProperties.addAll(PERSISTENCE_PROPERTIES);
+    allProperties.addAll(SERVER_PROPERTIES);
+    allProperties.addAll(REPLICATION_PROPERTIES);
+
+    return allProperties;
   }
```
sabbey37, line 279: It seems like you could do away with this method and just have `allProperties` as a constant since the properties that are added to the list never change.

geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/server/AbstractInfoIntegrationTest.java
```diff
@@ -182,10 +295,9 @@ public void shouldNotReturnKeySpaceSection_givenServerWithNoKeys() {
...
 
   @Test
-  public void shouldThrowExceptionIfGivenMoreThanOneParameter() {
+  public void shouldThrowException_ifGivenMoreThanOneParameter() {
     assertThatThrownBy(
         () -> jedis.sendCommand(
             Protocol.Command.INFO, "Server", "Cluster")).hasMessageContaining("ERR syntax error");
```
sabbey37, line 301: If you end up making changes anyway, we could also change this to `hasMessage()` instead of `hasMessageContaining()` so we check the exact message.

sabbey37 (changes requested):
Looks like a bunch of files got changed that are unrelated to this PR? Maybe something weird happened in the merge?

jdeppe-pivotal (changes requested):
I don't think this PR has the changes you want.

PR #5762, opened 2020-11-19
GEODE-8732: Update Tomcat9 module to publish to Maven

sabbey37:
We need to publish geode-modules-tomcat9 to Maven so it can be utilized properly when consuming Geode via Maven.


kohlmu-pivotal:
Just for interest sake, what is the use case that we now have to publish these extensions, which previously we've never had to publish before?

pivotal-jbarrett:
> Just for interest sake, what is the use case that we now have to publish these extensions, which previously we've never had to publish before?

Anyone downstream that wanted to include the Tomcat session manager into their build process had to do so outside of the Maven repo coordinates.

pivotal-jbarrett (approved):

PR #5763, opened 2020-11-19
GEODE-8521: detect if a p2p reader thread is hung

dschneider-pivotal:
Each p2p reader thread now creates a single instance of AbstractExecutor
and registers it with the ThreadsMonitoring instance while it is "dispatching"
the message. It will not be monitored while it is waiting on the socket to read
a message. So this will only report stuck p2p readers that are stuck dispatch
which for inline processing includes processing the message and reading the
direct ack.

### For all changes:

### Note:


geode-core/src/main/java/org/apache/geode/internal/monitoring/ThreadsMonitoringImpl.java
```diff
@@ -96,39 +98,60 @@ public void updateThreadStatus() {
...
       default:
-        return false;
+        throw new IllegalStateException("Unhandled mode=" + mode);
     }
-    this.monitorMap.put(Thread.currentThread().getId(), absExtgroup);
+  }
+
+  @Override
+  public boolean startMonitoring(AbstractExecutor executor) {
+    executor.setStartTime(System.currentTimeMillis());
```
bschuchardt, line 138: I'm a little concerned about the impact of taking yet another millisecond clock reading on every message that we dispatch.  This should be run through extensive performance testing before merging it to develop.

bschuchardt (commented):
It would be nice to be able to detect stuck P2P Reader threads that are dispatching messages but I'm concerned about the need to take a millisecond clock reading at the beginning of every message-dispatch.
Have you thought about having the thread be monitored continuously and checking its connectionState to see if it's dispatching?  You'd have to add a new DISPATCHING state and have custom code in the monitor for the new P2PReaderExecutorGroup to do this checking.

geode-core/src/main/java/org/apache/geode/internal/monitoring/ThreadsMonitoringImpl.java
dschneider-pivotal: I had the same thought. I was even concerned about constantly adding and removing from the concurrent map. We could just use the frequency of the monitor thread to do the time work. For example if it finds an AbstractExecutor in the map that it has not seen before then the monitor thread could set the time it first saw it. If later monitor samples see it again it could use this time (i.e. the one the monitor set) to determine if it is stuck. This could even be done without timestamps. Just a simple counter the monitor incs and the p2p reader thread clears. If the monitor thread is waking up at a fixed interval then you know the thread has been stuck for at least that much time. I like your idea of having custom code in P2PReaderExecutorGroup for this. When I was thinking about it I thought I would need to change the behavior of all the existing threads being monitored. Thanks for the feedback!

dschneider-pivotal: I converted back to draft mode. Will rework this and submit it again.

bschuchardt: having the monitor thread set the time is a good idea.  Another thing you could do is have the Monitor thread record the current time when it wakes up and use that for new entries in the map.  GMSHealthMonitor's monitor thread does something like that.

dschneider-pivotal: the p2p reader now simply sets a volatile boolean before and after messageDispatch.
the startTime for a p2p reader will now be set by the monitor thread.
Let me know what you think

agingade (approved):

bschuchardt (approved):
Thanks for removing the clock probes Darrel.  I also like the change to keep the reader threads in the monitor map.  This will be great to have in diagnosing customer issues!

geode-core/src/main/java/org/apache/geode/internal/monitoring/ThreadsMonitoringProcess.java
```diff
@@ -33,73 +31,86 @@
...
 
-  private final Properties nonDefault = new Properties();
-  private final DistributionConfigImpl distributionConfigImpl =
-      new DistributionConfigImpl(nonDefault);
+  private ResourceManagerStats resourceManagerStats = null;
 
   protected ThreadsMonitoringProcess(ThreadsMonitoring tMonitoring,
-      InternalDistributedSystem iDistributedSystem) {
-    this.timeLimit = this.distributionConfigImpl.getThreadMonitorTimeLimit();
+      InternalDistributedSystem iDistributedSystem, int timeLimit) {
```
kirklund, line 43: What unit is timeLimit? I recommend either adding a TimeUnit parameter or renaming timeLimit to include Millis or something as a prefix.

geode-core/src/test/java/org/apache/geode/internal/monitoring/ThreadsMonitoringProcessJUnitTest.java
```diff
@@ -34,11 +31,13 @@
  */
 public class ThreadsMonitoringProcessJUnitTest {
 
+  private static final int TIME_LIMIT = 1000;
```
kirklund, line 34: Please indicate the units of measure in the name or an additional parameter.

geode-core/src/test/java/org/apache/geode/internal/monitoring/executor/P2PReaderExecutorGroupJUnitTest.java
```diff
@@ -0,0 +1,53 @@
...
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.apache.geode.internal.monitoring.executor;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.Test;
+
+public class P2PReaderExecutorGroupJUnitTest {
```
kirklund, line 21: I recommend just naming unit tests as P2PReaderExecutorGroupTest (no need to include JUnit)

kirklund (changes requested):
Looks good! I requested a few small changes mostly involving naming.

geode-core/src/test/java/org/apache/geode/internal/monitoring/executor/P2PReaderExecutorGroupJUnitTest.java
dschneider-pivotal: done

geode-core/src/main/java/org/apache/geode/internal/monitoring/ThreadsMonitoringProcess.java
dschneider-pivotal: I changed the name to timeLimitMillis. This unit is determined by an external geode property so it will always be milliseconds

geode-core/src/test/java/org/apache/geode/internal/monitoring/ThreadsMonitoringProcessJUnitTest.java
dschneider-pivotal: done

kirklund (approved):

dschneider-pivotal:
the dunit failure is the known issue: GEODE-8573

PR #5770, opened 2020-11-24
GEODE-8745: Closing the queue region when senders are stopped.

nabarunnag:
* cleanQueues are applicable only while using persistence
	* when we are closing the region, the disk files are not deleted.
	* hence the values will still be maintained on restart.
	* Advantage will be that we are creating the queue region and its cache listener together
	* previously the region was not closed and cache listener was attached using mutators.
	* this caused secondary events to be missed before the cache listener is activated

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/wan/serial/SerialGatewaySenderEventProcessor.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/wan/serial/SerialGatewaySenderEventProcessor.java
@@ -111,10 +111,9 @@ public class SerialGatewaySenderEventProcessor extends AbstractGatewaySenderEven
       ThreadsMonitoring tMonitoring, boolean cleanQueues) {
     super("Event Processor for GatewaySender_" + id, sender, tMonitoring);
 
+    initializeMessageQueue(id, cleanQueues);
     this.unprocessedEvents = new LinkedHashMap<EventID, EventWrapper>();
     this.unprocessedTokens = new LinkedHashMap<EventID, Long>();
-
-    initializeMessageQueue(id, cleanQueues);
   }
 
   @Override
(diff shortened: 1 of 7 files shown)


geode-core/src/main/java/org/apache/geode/internal/cache/wan/serial/SerialGatewaySenderEventProcessor.java
```diff
@@ -111,10 +111,9 @@ public SerialGatewaySenderEventProcessor(AbstractGatewaySender sender, String id
       ThreadsMonitoring tMonitoring, boolean cleanQueues) {
     super("Event Processor for GatewaySender_" + id, sender, tMonitoring);
 
+    initializeMessageQueue(id, cleanQueues);
```
nabarunnag, line 114: This change is because the CacheListener to handle the unprocessedEventsMap is set here, hence I wanted the listener to be up and running before the map is initialized so the no events are missed because they were put in the map before the listener was running.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialWANPersistenceEnabledGatewaySenderDUnitTest.java
```diff
@@ -603,27 +605,32 @@ public void testReplicatedRegionPersistentWanGateway_restartSenderWithCleanQueue
...
     logger.info("Completed puts in the region");
 
     vm4.invoke(() -> WANTestBase.stopSender("ln"));
     vm5.invoke(() -> WANTestBase.stopSender("ln"));
 
-    logger.info("Stopped all the senders. ");
 
-    // Create receiver on remote site
-    createReceiverInVMs(vm2, vm3);
+    logger.info("Stopped all the senders. ");
```
DonalEvans, line 619: Instead of using the logger in this test to track progress, you could instead use the `invoke()` method signature that passes a String description/name as the first parameter, which allows anyone debugging the test to see when the invocation in question starts and finishes. Just personal preference, but I think it makes things neater and provides a little more information to anyone looking at the test output.

DonalEvans (approved):

boglesby (approved):

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialWANPersistenceEnabledGatewaySenderDUnitTest.java
nabarunnag: A valid point Donal. I am creating another GEODE ticket to just remove all the logger.info in tests. There are a couple of more statements that went in other commits.

PR #5774, opened 2020-11-26
GEODE-8741_1: implementation of API changes

mivanac:
Added proposal of implementation, of new API for block/suppress CQ Create/Update/Destroy notifications.
Idea is that user can chose which notifications wants to suppress, with goal to  remove unnecessary notifications, and improve performanse.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/cache/query/CqQuery.java
+++ b/geode-core/src/main/java/org/apache/geode/cache/query/CqQuery.java
@@ -156,11 +156,27 @@ public interface CqQuery {
   boolean isDurable();
 
   /**
+   * This allows to check if the CQ use option to suppress CQ create notification.
+   *
+   * @return boolean true if create is suppressed, false otherwise
+   * @since Geode 14.0
+   */
+  boolean isCreateSuppressed();
+
+  /**
    * This allows to check if the CQ use option to suppress CQ update notification.
    *
    * @return boolean true if update is suppressed, false otherwise
    * @since Geode 14.0
    */
   boolean isUpdateSuppressed();
+
+  /**
+   * This allows to check if the CQ use option to suppress CQ destroy notification.
+   *
+   * @return boolean true if destroy is suppressed, false otherwise
+   * @since Geode 14.0
+   */
+  boolean isDestroySuppressed();
 
 }
--- a/geode-core/src/main/java/org/apache/geode/cache/query/CqQuery.java
+++ b/geode-core/src/main/java/org/apache/geode/cache/query/CqQuery.java
@@ -155,4 +155,12 @@ public interface CqQuery {
    */
   boolean isDurable();
 
+  /**
+   * This allows to check if the CQ use option to suppress CQ update notification.
+   *
+   * @return boolean true if update is suppressed, false otherwise
+   * @since Geode 14.0
+   */
+  boolean isUpdateSuppressed();
+
 }
--- a/geode-core/src/main/java/org/apache/geode/cache/query/CqQuery.java
+++ b/geode-core/src/main/java/org/apache/geode/cache/query/CqQuery.java
@@ -159,7 +159,7 @@ public interface CqQuery {
    * This allows to check if the CQ use option to suppress CQ create notification.
    *
    * @return boolean true if create is suppressed, false otherwise
-   * @since Geode 14.0
+   * @since Geode 1.15.0
    */
   boolean isCreateSuppressed();
 
@@ -167,7 +167,7 @@ public interface CqQuery {
    * This allows to check if the CQ use option to suppress CQ update notification.
    *
    * @return boolean true if update is suppressed, false otherwise
-   * @since Geode 14.0
+   * @since Geode 1.15.0
    */
   boolean isUpdateSuppressed();
 
@@ -175,7 +175,7 @@ public interface CqQuery {
    * This allows to check if the CQ use option to suppress CQ destroy notification.
    *
    * @return boolean true if destroy is suppressed, false otherwise
-   * @since Geode 14.0
+   * @since Geode 1.15.0
    */
   boolean isDestroySuppressed();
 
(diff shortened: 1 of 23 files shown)


geode-core/src/main/java/org/apache/geode/cache/query/CqQuery.java
```diff
@@ -155,4 +155,12 @@
    */
   boolean isDurable();
 
+  /**
+   * This allows to check if the CQ use option to suppress CQ update notification.
+   *
+   * @return boolean true if update is suppressed, false otherwise
+   * @since Geode 14.0
+   */
+  boolean isUpdateSuppressed();
```
kohlmu-pivotal, line 164: @mivanac I quite like the idea of possibly limiting what notifications one would like to receive. I think that the approach is maybe a little limiting, as it is specifically focussed on  "update" only. Could we achieve a much more encompassing solution for delete or create as well? Maybe an inclusions or exclusions list of events that the end result would like to receive?

Given that this is an external facing API, I would encourage changes to be more encompassing, in order to avoid many API changes/deprecations of external facing APIs.
@upthewaterspout @agingade @nabarunnag could anyone of you possibly be able to work with @mivanac to come up with an API that is possibly a little more future proofed.

kohlmu-pivotal (changes requested):

DonalEvans (commented):
Rather than using a bitmask and requiring users to specify a non-intuitive integer value to indicate which notifications are suppressed, might it be better to create a wrapper class with setter and getter methods for each notification type which can be passed to the `newCq()` method? This adds a little complexity in implementation, but would be far more user-friendly.

DonalEvans (commented):
Instead of storing the notification suppression internally as an integer, would it be simpler to just use the new `CqSuppressNotification` class everywhere, and just check the value of the boolean flags rather than doing bit arithmetic? If there's a specific reason for preferring the current implementation that's fine though.

mivanac:
Main motivation is to simplify implementation, and also simplify adding and getting part to/from message ExecuteCQ.

geode-core/src/main/java/org/apache/geode/cache/query/CqQuery.java
```diff
@@ -155,4 +155,28 @@
    */
   boolean isDurable();
 
+  /**
+   * This allows to check if the CQ use option to suppress CQ create notification.
+   *
+   * @return boolean true if create is suppressed, false otherwise
+   * @since Geode 14.0
```
kirklund, line 162: I think this is supposed to be Geode 1.14?

kirklund (commented):
Please make sure you're creating new unit tests (not just distributed/integration tests) for classes you touch that have not unit test. At a minimum, I'd recommend covering the method(s) you add or alter.

onichols-pivotal:
This PR has been open for some while with no recent activity.  Please keep Geode tidy by closing PRs you're not actively using.

mivanac:
This PR is waiting on review from CODEOWNERS.

mivanac:
Hi  @jdeppe-pivotal, @jinmeiliao, @agingade, @DonalEvans, @nabarunnag, @jchen21, and @kirklund, could you review this PR.
Thanks.

DonalEvans (commented):
After doing a little research, I think that the simplest and clearest implementation for the "which notification to suppress" information might be an `EnumSet`. This would prevent the need for doing bit masking and passing around an int, would remove the need for individual "isXSuppressed()" methods (since you can just call `set.contails()`), and would be more intuitive from a development point of view. Here's an example of what I'm thinking of:
```
class Scratch {
  enum SuppressedNotifications {
    CREATE, UPDATE, DESTROY;
  }
  public static void main(String[] args) {
    EnumSet<SuppressedNotifications> onlyAllowDestroy = EnumSet.of(SuppressedNotifications.CREATE, SuppressedNotifications.UPDATE);
    for (SuppressedNotifications type : SuppressedNotifications.values()) {
      System.out.println("Set contains type " + type + " = " + onlyAllowDestroy.contains(type));
    }
  }
}
```
which outputs:
```
Set contains type CREATE = true
Set contains type UPDATE = true
Set contains type DESTROY = false
```
It's possible that this implementation may be slightly more performant, as an `EnumSet` is backed by a bit array which is the size of the number of enum values it's based on (3 in this case) instead of being a 32-bit int. I apologise for not suggesting this initially, but I only just took the time to really do some digging and see if there was a better implementation for this case.

mivanac:
Thanks for the comments and proposal. I will apply it.

agingade:
I am having hard time understanding the requirement here...The jira ticket just mention block CQ notification, what kind of notification we are talking here...By design CQ does not raise any notification...Can you be more specific about the notification you are mentioning about...
Also, i don't think it should be managed by the create/destroy APIs or by adding entirely new set of APIs, thats redundant.

agingade (changes requested):
I have added my comment about the requirement of this as I am unable to find details in the Jira ticket or here.

agingade:
I have added following comments in the ticket also:
I have to look into the PR to see what your are really are trying to achieve in this ticket. It will be helpful if you can add more information with the example (depending on the issue).

E.g. 
Avoid sending CQ events for specific CQ query-operations (E.g. destroy Query operation) to client.
For a region.put() which results in CQ event with its query operation as destroy; avoid sending this event to client.

I am not seeing any value/scenario that needs this kind of behavior. Do you have any example can you please explain here.

The core CQ design aspect was to establish a data view representing a query result that can be expressed in the form of query.  And support event mechanism that will enable to manage/maintain that view as the change happens on the source data. The changes requested in this ticket, breaks the core design and can allow one to represent/establish a different data view that is not in-line with the source data (a query result view). I don't see need for supporting that kind of behavior in the server side. If Client is not interested to see specific event/operation it should be filtering this out on the client side. 

Why the need for this behavior?
Looking the Pull Request opened; its not like its avoiding processing of CQs, which is generally more expensive part of having CQ; only thing its avoiding sending the CQ event to client. If this is the behavior client/application is looking for; probably CQ is not the functionality to use.

Also, it will be nice to use Request For Change (RFC) process for any issues that has behavior/functional or new API requirements. It will help to get feedback during design phase itself.

agingade:
From an implementation point of view; this seems to be added/handled in CqQuery as a CqAttributes...
E.g:
interestedEventsWithQueryOperations = { ALL, CREATE, UPDATE, DESTROY}

mivanac:
Thanks for comments. I agree with you that there is no gain from this PR. I will close these PRs.

PR #5775, opened 2020-11-28
Revert "GEODE-8656: Fix ping only sent to one gateway receiver when s…

gesterzhou:
…everal r… (#5670)"

This reverts commit c92a91eae587800c9d24c973e536cdfe8da6f38a.

### For all changes:

### Note:


onichols-pivotal (changes requested):
Why is this revert necessary?  (I looked in the PR description, PR comments, GEODE-8656, and geode dev list and found no discussion.  I also checked the pipeline, and all tests are passing).
I would also suggest including the original committer as a reviewer.

gesterzhou:
Reverting does not need to find the root cause or find a better fix for the original checkin. As long as we confirmed this checkin broke current tests, we should revert first and re-fix in order not to block other people's work.

onichols-pivotal:
Agreed! Do you happen to have a link to the concourse.apachegeode-ci.info test result to confirm this checkin broke current tests and is blocking other people's work?

gesterzhou:
It blocked regression tests and if we are releasing 1.14, it's a release blocker. 

All the release blockers for 1.13 and earlier versions we resolved recently did not cause concourse.apachegeode-ci.info to fail.

onichols-pivotal:
It sounds like your concern involves private code that is not part of Geode.  In these cases it is especially important to communicate in as much detail as possible the nature of the problem you encountered, under what conditions, and most importantly, what the next steps will be. Will you try to add a Geode test demonstrating the problem so the original committer can submit a revised PR?  Can you offer to work with the original committer to improve their change once you have a better understanding of the root cause?  Can you offer to test iterations of their improved PR against your private code?  If you find a workaround in your private code would you consider a straight undo of the revert?

It’s all too easy to revert someone else’s fix and then forget about it.  An “emergency revert” incurs a debt; there needs to be a plan to repay it.

I’ll be happy to withdraw my -1, just looking for a few words about the problem area (compile failure? data loss? server hangs/crashes?), conditions (number of servers/locators? wan? upgrade? any special config settings?), and commitment to follow up after revert.

gesterzhou:
I reviewed the code. It's the phase 2 of a good feature. The phase 1 is GEODE-7565. At that time, there's already a potential issue to break backward compatibility. Base on my limited knowledge, I feel that introducing ServerLocationAndMemberId to replace ServerLocation might not be the correct way. The better way should be to introduce a new field "String memberId" into the ServerLocation class, then use toDataPre_GEODE_1_14_0_0 and fromDataPre_GEODE_1_14_0_0 in ServerLocation to handle the new field for rollingupgrade. 

As for how to find such issue earlier in geode, we do have rollingupgrade dunit tests. The pull request does not have this type of test, then it has to rely on hydra to find the issue. It's not hydra's fault though. We should feel happy that hydra protected us and found the issue today. 

As for how to cooperate with remote committer, it's a long story. I'd like to cooperate with remote committer to enhance the GEODE-8656. But even I am not skillful to know the root cause and how to fix, as the rule of regression analysis, we should still revert first then re-fix.

kohlmu-pivotal:
I, like @onichols-pivotal do not see any problem with the original [PR.(https://github.com/apache/geode/pull/5670). It was even signed off by arguably one of the SMEs in this code area.

If there is some private, non-Geode, related failure, then is this really a concern to Geode? ServerLocation is in the "internal" package space. Which means that any Geode committers/developers have fullest right and privilege to alter and modify its behavior without concern of breaking backward compatibility.

@gesterzhou It seems there is a growing concern that this change has broken a rolling upgrade and backward compatibility. I think there is a much larger concern now. How did an internal class leak out of its "internal" containment? How did we not catch THAT? WHY would an internal class now break rolling upgrades? 

I fully support the reverting of this PR if we can prove that rolling upgrades will suffer from this change AND then raise a corresponding ticket to fix the underlying root cause, which it seems is the leaking of internal classes.

I disagree with the proposed fix of "adding of the memberId String to the ServerLocation". Whilst it may "fix" the problem, it merely glosses over the fact that we seem to have a potentially larger problem here.

@bschuchardt would you mind looking at this problem? As it seems that there is a larger problem here that would possibly not have been considered when we initially approved the [PR](https://github.com/apache/geode/pull/5670).

kohlmu-pivotal (changes requested):
See comments already posted

bschuchardt (changes requested):
@kohlmu-pivotal Apparently my re-review of Alberto's changes are no longer needed.  Let me know if that changes.

gesterzhou:
Mark and Naba's fix in private test code looks like to have resolved the issue. I think we don't have to revert it. 

I agree that for internal class, geode develop should have more freedom to change. Private test code should comply with the change. I will close the PR.

PR #5797, opened 2020-12-01
GEODE-8752: Revoke using member's actual address

demery-pivotal:
When revoking a persistent member,
PersistentPartitionedRegionDistributedTest now uses the to-be-revoked
member's actual inet address in the revocation pattern.


jchen21 (commented):
This is a more reliable way to get the address, I believe. I have limited context on it though.

jinmeiliao (approved):

PR #5799, opened 2020-12-01
GEODE-8634: Fix AsyncInvocationTimeoutDistributedTest flakiness

kirklund:
Null out latch and threadId before each test.

The bug was caused by:

1. The previous test left threadId set to a non-zero value*
2. The await() in the second thread completes before the threadId.set() in the first thread executes
3. The second thread returns the wrong value for remoteThreadId
4. The wrong remoteThreadId is then compared to the caught exception's message which has a different thread id

*Previous test was await_callable_timeout_includesStackTraceAsCause in all three recorded test failures

Fix is simple: always set threadId to zero value before running each test.

### DIFF
--- a/geode-dunit/src/distributedTest/java/org/apache/geode/test/dunit/tests/AsyncInvocationTimeoutDistributedTest.java
+++ b/geode-dunit/src/distributedTest/java/org/apache/geode/test/dunit/tests/AsyncInvocationTimeoutDistributedTest.java
@@ -28,6 +28,7 @@ import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicReference;
 
 import org.junit.After;
+import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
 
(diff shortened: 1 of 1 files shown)


DonalEvans (approved):

kirklund:
I changed setUp to set threadId to zero instead of null to avoid NPE being thrown by this line:
{{{
await().until(() -> threadId.get() > 0);
}}}

geode-dunit/src/distributedTest/java/org/apache/geode/test/dunit/tests/AsyncInvocationTimeoutDistributedTest.java
```diff
@@ -40,12 +41,20 @@
...
+  private static final AtomicReference<Long> threadId = new AtomicReference<>(0L);
   private static final AtomicReference<CountDownLatch> latch = new AtomicReference<>();
 
   @Rule
   public DistributedRule distributedRule = new DistributedRule();
 
+  @Before
+  public void setUp() {
+    getVM(0).invoke(() -> {
+      latch.set(null);
```
jchen21, line 53: Is it better to initialize `latch` to `CountDownLatch(0)` instead of `null`? Also initialize it in line 45, when `latch` is declared?

kirklund: Yes, it would be better to avoid nulls and use CountDownLatch(0). Thanks!

kirklund: I think I've resolved this. Please take a look. Thanks!

jchen21 (approved):

kirklund:
ConcurrentSerialGatewaySenderOperationsDistributedTest failed.

jchen21:
> ConcurrentSerialGatewaySenderOperationsDistributedTest failed.

That is due to GEODE-8573, which fails more frequently recently.

PR #5800, opened 2020-12-01
GEODE-8755: Write content list files to integration test dir

demery-pivotal:
BundledJarsJUnitTest, AssemblyContentsIntegrationTest, and GeodeDependencyJarIntegrationTest once again write their output to the file paths indicated in their failure messages.

Authored-by: Dale Emery <demery@vmware.com>

Please review @yozaner1324

### DIFF
--- a/geode-assembly/src/integrationTest/java/org/apache/geode/BundledJarsJUnitTest.java
+++ b/geode-assembly/src/integrationTest/java/org/apache/geode/BundledJarsJUnitTest.java
@@ -67,7 +67,7 @@ public class BundledJarsJUnitTest {
         sortedJars.entrySet().stream().map(entry -> removeVersion(entry.getKey()));
     Set<String> bundledJarNames = new TreeSet<>(lines.collect(Collectors.toSet()));
 
-    Files.write(Paths.get("bundled_jars.txt"), bundledJarNames);
+    Files.write(Paths.get("..", "bundled_jars.txt"), bundledJarNames);
 
     Set<String> newJars = new TreeSet<>(bundledJarNames);
     newJars.removeAll(expectedJars);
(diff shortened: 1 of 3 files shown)


geode-assembly/src/integrationTest/java/org/apache/geode/BundledJarsJUnitTest.java
```diff
@@ -67,7 +67,7 @@ public void verifyBundledJarsHaveNotChanged() throws IOException {
         sortedJars.entrySet().stream().map(entry -> removeVersion(entry.getKey()));
     Set<String> bundledJarNames = new TreeSet<>(lines.collect(Collectors.toSet()));
 
-    Files.write(Paths.get("bundled_jars.txt"), bundledJarNames);
+    Files.write(Paths.get("..", "bundled_jars.txt"), bundledJarNames);
```
yozaner1324, line 70: This change looks good, however, BundledJarsJUnitTest is not the only test that has this problem; AssemblyContentsIntegrationTest and GeodeDependencyJarIntegrationTest also need essentially the same change.

yozaner1324 (changes requested):

geode-assembly/src/integrationTest/java/org/apache/geode/BundledJarsJUnitTest.java
demery-pivotal: Can you think of way to identify the tests that have this problem?

demery-pivotal: I did a scan of the integration test output to spot this kind of file. I found AssemblyContentsIntegrationTest and GeodeDependencyJarIntegrationTest. I will fix those. If you know of others, let me know.

yozaner1324: The three mentioned tests are the only ones I'm aware of that write files that need to be accessed after the test runs. I searched for all other occurrences of "Files.write" and none seemed important. I'm not sure there is a better way to identify which files are important vs random temp files.

yozaner1324 (approved):

kirklund (approved):

PR #5801, opened 2020-12-02
GEODE-8665: validate offline-disk-store command is missing information

jchen21:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ValidateOfflineDiskStoreDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ValidateOfflineDiskStoreDUnitTest.java
@@ -147,7 +147,7 @@ public class ValidateOfflineDiskStoreDUnitTest implements Serializable {
 
     populateRegions();
 
-    AssertRegionSizeAndDiskStore();
+    assertRegionSizeAndDiskStore();
 
     server.invoke(ValidateOfflineDiskStoreDUnitTest::stopServer);
 
@@ -216,7 +216,7 @@ public class ValidateOfflineDiskStoreDUnitTest implements Serializable {
     SERVER.get().stop();
   }
 
-  private void AssertRegionSizeAndDiskStore() {
+  private void assertRegionSizeAndDiskStore() {
     assertRegionSize();
     assertDiskStore(serverName);
   }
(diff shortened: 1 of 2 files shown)


geode-core/src/main/java/org/apache/geode/internal/cache/Oplog.java
```diff
@@ -2225,6 +2225,12 @@ long recoverCrf(OplogEntryIdSet deletedIds, boolean recoverValues, boolean recov
         if (getParent().isOfflineCompacting()) {
           getParent().incLiveEntryCount(getRecoveryMap().size());
         }
+        long tc = totalCount.get();
+        long tlc = totalLiveCount.get();
+        if (getParent().isValidating() && tlc >= 0
+            && tc > tlc) {
```
agingade, line 2231: Can there be a case where tc and tlc both are same number (including 0); in that case dead-record-counts are 0; is this valid?

jchen21: Yes. It is valid. It is very common for tc(totalCount) and tlc(totalLiveCount) to be the same value, including 0, which indicates an empty Oplog. When tc and tlc are the same, offline disk store validation will not show any message about  the number of records that would be removed if you compacted the store, since there is no deleted or skipped entry in this case.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ValidateOfflineDiskStoreDUnitTest.java
```diff
@@ -0,0 +1,281 @@
...
+        .build());
+
+    SERVER.get().start();
+  }
+
+  private static void stopServer() {
+    SERVER.get().stop();
+  }
+
+  private void AssertRegionSizeAndDiskStore() {
```
gesterzhou, line 219: method name should starts with lower case.

gesterzhou (approved):
Fix the typo in method name. Other parts look good.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ValidateOfflineDiskStoreDUnitTest.java
jchen21: Good catch!

PR #5811, opened 2020-12-03
GEODE-8730: DualServerSNIAcceptanceTest fails to start server because port is in use

Bill:
2 Dockerized cache servers (in test) were binding to ephemeral ports 40404, 40405. These can conflict with some ephemeral port that is bound in the container before the Geode cluster is started. They also can conflict with the RMI port bound by each cache server.

Changed the test to bind to ports 8501, 8502.


bschuchardt (approved):
moving the server's port out of the ephemeral port range should fix the problem

Bill:
Though DistributedTestOpenJDK11 had a couple failures, I'm sure they are unrelated to changes in the PR, which changed only test code for an acceptance test. Merging…

PR #5818, opened 2020-12-07
GEODE-8769: help ensure stats are settled before testing

ringles:

### For all changes:
- [ X] Is there a JIRA ticket associated with this PR? Is it referenced in the commit message?

- [ X] Has your PR been rebased against the latest commit within the target branch (typically `develop`)?

- [ X] Is your initial contribution a single, squashed commit?

- [ X] Does `gradlew build` run cleanly?

### Note:

This helps to mitigate the issue but does not completely solve it. The problem happens when multiple threads are opening multiple Redis connections, and the statistics are reset. Depending on the timing of the connection opening and closing, the operations can interleave and result in an inconsistent state for the statistics. The stats will be "eventually correct" unless their values are explicitly reset at an unlucky time.

### DIFF
--- a/geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/RedisStatsIntegrationTest.java
+++ b/geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/RedisStatsIntegrationTest.java
@@ -77,7 +77,7 @@ public class RedisStatsIntegrationTest {
 
     preTestKeySpaceHits = redisStats.getKeyspaceHits();
     preTestKeySpaceMisses = redisStats.getKeyspaceMisses();
-    preTestConnectionsReceived = redisStats.getConnectionsReceived();
+    preTestConnectionsReceived = redisStats.getTotalConnectionsReceived();
     preTestConnectedClients = redisStats.getConnectedClients();
   }
 
@@ -496,16 +496,16 @@ public class RedisStatsIntegrationTest {
   }
 
   @Test
-  public void connectionsReceivedStat_shouldIncrement_WhenNewConnectionOccurs() {
+  public void totalConnectionsReceivedStat_shouldIncrement_WhenNewConnectionOccurs() {
 
     Jedis jedis2 = new Jedis("localhost", server.getPort(), TIMEOUT);
     jedis2.ping();
 
-    assertThat(redisStats.getConnectionsReceived()).isEqualTo(preTestConnectionsReceived + 1);
+    assertThat(redisStats.getTotalConnectionsReceived()).isEqualTo(preTestConnectionsReceived + 1);
 
     jedis2.close();
 
-    assertThat(redisStats.getConnectionsReceived()).isEqualTo(preTestConnectionsReceived + 1);
+    assertThat(redisStats.getTotalConnectionsReceived()).isEqualTo(preTestConnectionsReceived + 1);
   }
 
   // ######################## Server Section ################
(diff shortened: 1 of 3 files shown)


dschneider-pivotal (commented):
I don't think the current fix makes sense. At least I would not be able to explain why these tests are waiting for the stat to be zero at the beginning of their run.
I'm not sure what is causing the problem but I see a couple of things I recommend get cleaned up and perhaps the test will then pass consistently:
1. When the test starts up the server have redis listen on a unique port instead of the default one. That way nothing else should be accidently connecting to it.
2. get rid of the "clearAllStats" method in the product that this test uses (if you end up keeping it mark it with the visibleForTesting annotation). This test should instead read the current value of a stat, do an operation, and then confirm that the stat changed by the expected amount. Instead of asserting a stat has some absolute value, all your stat assertions should be relative to the original value.
3. some operation may update the stat async (jedis.close() may be like this) so in those cases your validation should use an await.

jdeppe-pivotal:
I agree that this fix is probably not going to address the issue. It's more like the async nature of netty that's causing this. Since we do `jedis.close()` in the `after` call and then reset stats in the `before` the client count decrement 'event' might only happen after the stats are decremented. As it stands I think just `awaiting` for the client count to reach zero at the start of the before block should fix this issue.
@dschneider-pivotal I'd be hesitant to remove the `reset` method. Because of the async nature of some of these stats it would still be an issue if you grabbed the 'current' value of a stat at the start of the test. (Perhaps it's only client connections that are affected by this).

dschneider-pivotal:
@jdeppe-pivotal  for async ops you can always have the test wait to confirm the stat is updated right after close returns. This tells you the async close operation has completed. I don't see how the reset helps (it with async closes is probably the cause of the stat going negative). It would be better to test the product without adding product code if possible. If you do need to keep it then make sure and mark it for testing only. Also check that it actually resets all the stats (I think at least one was missing).

geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/RedisStatsIntegrationTest.java
```diff
@@ -468,33 +479,33 @@ public void NetworkKiloBytesReadOverLastSecond_shouldReturnCorrectData() {
...
+    jedis2.ping();
 
-    assertThat(redisStats.getConnectionsReceived()).isEqualTo(1);
+    assertThat(redisStats.getConnectionsReceived()).isEqualTo(preTestConnectionsReceived + 1);
 
-    jedis.close();
+    jedis2.close();
 
-    assertThat(redisStats.getConnectionsReceived()).isEqualTo(1);
+    assertThat(redisStats.getConnectionsReceived()).isEqualTo(preTestConnectionsReceived + 1);
```
jdeppe-pivotal, line 508: I don't think this assertion is correct. It should be doing what the previous test is doing:
```
GeodeAwaitility.await().atMost(Duration.ofSeconds(2))
        .untilAsserted(() -> assertThat(redisStats.getConnectedClients()).isEqualTo(preTestConnectedClients));
```

jdeppe-pivotal (changes requested):

geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/RedisStatsIntegrationTest.java
jhutchison: I'm not sure how we can make this more clear, but I see how this is confusing-  this test is testing the connectionsReceived parameter, which keeps the history of connections received, even after they have been closed.   (the getConnectedlClient param, in the test above is about current active connections only)

jdeppe-pivotal: OK - got it. That does make sense. I think the test method name does make it clearer, but I wasn't paying close attention to that :/

jdeppe-pivotal (approved):

jhutchison:
re-submitted under https://github.com/apache/geode/pull/5849

jdeppe-pivotal:
Closing in favor of: https://github.com/apache/geode/pull/5849

PR #5821, opened 2020-12-08
GEODE-8745: Handle race between batch destroy msg and sec event.

nabarunnag:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySender.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySender.java
@@ -1068,8 +1068,8 @@ public abstract class AbstractGatewaySender implements InternalGatewaySender, Di
           if (isDebugEnabled) {
             logger.debug("Returning back without putting into the gateway sender queue:" + event);
           }
-          if (this.eventProcessor != null) {
-            this.eventProcessor.registerEventDroppedInPrimaryQueue(event);
+          if (this.isPrimary()) {
+            recordDroppedEvent(clonedEvent);
           }
           return;
         }
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/wan/serial/SerialGatewaySenderEventProcessor.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/wan/serial/SerialGatewaySenderEventProcessor.java
@@ -641,14 +641,15 @@ public class SerialGatewaySenderEventProcessor extends AbstractGatewaySenderEven
         return true;
       } else if (addToUnprocessedTokens) {
         // Secondary event may not have arrived
+        if (logger.isTraceEnabled()) {
+          logger.trace("{}: fromPrimary destroy event {} : added to unprocessed token map",
+              sender.getId(), eventId);
+        }
         Long mapValue =
-            Long.valueOf(System.currentTimeMillis() + AbstractGatewaySender.TOKEN_TIMEOUT);
+            System.currentTimeMillis() + AbstractGatewaySender.TOKEN_TIMEOUT;
         Long oldv = this.unprocessedTokens.put(eventId, mapValue);
         if (oldv == null) {
           statistics.incUnprocessedTokensAddedByPrimary();
-        } else {
-          // its ok for oldv to be non-null
-          // this shouldn't happen anymore @todo add an assertion here
         }
       }
     }
@@ -682,13 +683,10 @@ public class SerialGatewaySenderEventProcessor extends AbstractGatewaySenderEven
         }
         {
           Long mapValue =
-              Long.valueOf(System.currentTimeMillis() + AbstractGatewaySender.TOKEN_TIMEOUT);
+              System.currentTimeMillis() + AbstractGatewaySender.TOKEN_TIMEOUT;
           Long oldv = this.unprocessedTokens.put(gatewayEvent.getEventId(), mapValue);
           if (oldv == null) {
             statistics.incUnprocessedTokensAddedByPrimary();
-          } else {
-            // its ok for oldv to be non-null
-            // this shouldn't happen anymore @todo add an assertion here
           }
         }
       } else {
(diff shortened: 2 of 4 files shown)


nabarunnag:
* Primary serial sender receives an event but the sender is not running yet. Hence it has to drop the event.
   * Primary has to inform the secondary to remove the event from the unprocessedEventsMap.
   * It has to do that because the event is dropped and the cache listener on the secondary queue will not be triggered to remove it.
   * Hence, the primary sends BatchDestroyOperation to the secondary with the tail key set to -1 to trigger removal from unprocessedEventMap.
   * There is a race in which the BatchDestroyOperation arrives before the secondary event. 
   * In this case the event is never removed from the unprocessedEventsMap.
   * In this fix, if this happens the event ID is placed in the unprocessedTokensMap.
   * So, when the secondary event arrives, it checks the unprocessedTokensMap and then realizes that the primary has seen the event and there is no need to place the event in the unprocessedEventsMap.

geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySender.java
```diff
@@ -1068,8 +1068,8 @@ public void distribute(EnumListenerEvent operation, EntryEventImpl event,
           if (isDebugEnabled) {
             logger.debug("Returning back without putting into the gateway sender queue:" + event);
           }
-          if (this.eventProcessor != null) {
-            this.eventProcessor.registerEventDroppedInPrimaryQueue(event);
+          if (this.eventProcessor != null && this.isPrimary()) {
+            recordDroppedEvent(clonedEvent);
           }
```
nabarunnag, line 1073: Refer line#1035

geode-core/src/main/java/org/apache/geode/internal/cache/wan/serial/SerialGatewaySenderEventProcessor.java
```diff
@@ -638,6 +639,17 @@ protected boolean basicHandlePrimaryDestroy(final EventID eventId) {
...
+        // Secondary event may not have arrived
+        Long mapValue =
+            Long.valueOf(System.currentTimeMillis() + AbstractGatewaySender.TOKEN_TIMEOUT);
+        Long oldv = this.unprocessedTokens.put(eventId, mapValue);
+        if (oldv == null) {
+          statistics.incUnprocessedTokensAddedByPrimary();
+        } else {
+          // its ok for oldv to be non-null
+          // this shouldn't happen anymore @todo add an assertion here
+        }
```
nabarunnag, line 652: Same logic as in basicHandlePrimaryEntry

DonalEvans, line 652: This empty block can probably be replaced with just a comment.

geode-core/src/main/java/org/apache/geode/internal/cache/wan/serial/SerialGatewaySenderEventProcessor.java
```diff
@@ -638,6 +639,17 @@ protected boolean basicHandlePrimaryDestroy(final EventID eventId) {
         ew.event.release();
         statistics.incUnprocessedEventsRemovedByPrimary();
         return true;
+      } else if (addToUnprocessedTokens) {
+        // Secondary event may not have arrived
+        Long mapValue =
+            Long.valueOf(System.currentTimeMillis() + AbstractGatewaySender.TOKEN_TIMEOUT);
```
DonalEvans, line 645: This can just be `Long mapValue = System.currentTimeMillis() + AbstractGatewaySender.TOKEN_TIMEOUT;`

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialWANPersistenceEnabledGatewaySenderDUnitTest.java
```diff
@@ -577,17 +577,15 @@ public void testReplicatedRegionPersistentWanGateway_restartSenderWithCleanQueue
...
-        false, 100, 10, false, true, null, null, true));
-
-    logger.info("The first ds is " + firstDStore);
-    logger.info("The second ds is " + secondDStore);
+    String firstDStore =
+        (String) vm4.invoke("Creating DS", () -> WANTestBase.createSenderWithDiskStore("ln", 2,
+            false, 100, 10, false, true, null, null, true));
+    String secondDStore =
+        (String) vm5.invoke("Creating DS", () -> WANTestBase.createSenderWithDiskStore("ln", 2,
+            false, 100, 10, false, true, null, null, true));
```
DonalEvans, line 588: These variables are never used, so I think it should be okay to change these to just 
```
vm4.invoke("Creating DS", () -> WANTestBase.createSenderWithDiskStore("ln", 2, false, 100, 10, 
false, true, null, null, true));
vm5.invoke("Creating DS", () -> WANTestBase.createSenderWithDiskStore("ln", 2, false, 100, 10, 
false, true, null, null, true));
```

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialWANPersistenceEnabledGatewaySenderDUnitTest.java
```diff
@@ -605,26 +603,21 @@ public void testReplicatedRegionPersistentWanGateway_restartSenderWithCleanQueue
...
 
-    logger.info("Stopped all the senders. ");
-
-    AsyncInvocation inv1 = vm4.invokeAsync(() -> WANTestBase.startSenderwithCleanQueues("ln"));
-    logger.info("Started the sender in vm 4");
+    createReceiverInVMs(vm2, vm3);
 
-    vm5.invoke(() -> WANTestBase.startSenderwithCleanQueues("ln"));
-    logger.info("Started the sender in vm 5");
+    AsyncInvocation inv1 = vm4.invokeAsync("Starting sender with clean queues",
```
DonalEvans, line 615: The compiler warning here can be fixed by using `AsyncInvocation<?> inv1`.

DonalEvans (changes requested):
Just a few code clean-up requests.

DonalEvans (approved):

PR #5823, opened 2020-12-08
GEODE-8771: invalidate should acquire the lock before initIndex

jinmeiliao:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/integrationTest/java/org/apache/geode/cache/query/partitioned/PRClearIntegrationTest.java
+++ b/geode-core/src/integrationTest/java/org/apache/geode/cache/query/partitioned/PRClearIntegrationTest.java
@@ -1,0 +1,73 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package org.apache.geode.cache.query.partitioned;
+
+
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.IntStream;
+
+import org.junit.Rule;
+import org.junit.Test;
+
+import org.apache.geode.cache.EntryNotFoundException;
+import org.apache.geode.cache.Region;
+import org.apache.geode.internal.cache.InternalCache;
+import org.apache.geode.test.junit.rules.ExecutorServiceRule;
+import org.apache.geode.test.junit.rules.ServerStarterRule;
+
+public class PRClearIntegrationTest {
+
+  @Rule
+  public ServerStarterRule server = new ServerStarterRule().withAutoStart();
+
+  @Rule
+  public ExecutorServiceRule executor = new ExecutorServiceRule();
+
+  @Test
+  public void shouldNotHang() throws Exception {
+    InternalCache cache = server.getCache();
+    Region<Object, Object> region = server.createPartitionRegion("regionA", f -> {
+    }, f -> f.setTotalNumBuckets(1));
+    cache.getQueryService().createIndex("indexA", "r", "/regionA r");
+    region.put(0, "value0");
+
+    CompletableFuture<Void> put = executor.runAsync(() -> {
+      Thread.currentThread().setName("put-Thread");
+      IntStream.range(0, 10).forEach(i -> region.put(i, "value" + i));
+    });
+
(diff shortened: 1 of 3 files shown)


mhansonp (approved):

geode-core/src/integrationTest/java/org/apache/geode/cache/query/partitioned/PRClearIntegrationTest.java
```diff
@@ -0,0 +1,73 @@
...
+public class PRClearIntegrationTest {
+
+  @Rule
+  public ServerStarterRule server = new ServerStarterRule().withAutoStart();
+
+  @Rule
+  public ExecutorServiceRule executor = new ExecutorServiceRule();
+
+  @Test
+  public void shouldNotHang() throws Exception {
```
demery-pivotal, line 41: I would like a name that better describes the intention of the test. It looks as if the concurrent puts and invalidates are important. So maybe something like `doesNotHangWhenClearedWithConcurrentPutsAndInvalidates()`.

demery-pivotal (approved):

geode-core/src/main/java/org/apache/geode/internal/cache/AbstractRegionMap.java
```diff
@@ -1282,14 +1282,18 @@ public boolean invalidate(EntryEventImpl event, boolean invokeCallbacks, boolean
     DiskRegion dr = owner.getDiskRegion();
     boolean ownerIsInitialized = owner.isInitialized();
 
+    // lock before waitForIndexInit so that we should wait
+    // till a concurrent clear to finish
+    lockForCacheModification(owner, event);
```
jchen21, line 1287: I am not sure I understand why the code change here can fix the hang, and not causing a different type of hang. I do see the two-line comment above. But it does not answer my question. It would be helpful for the reviewers to understand it, if there is some explanation of root cause analysis and the fix in the pull request description or in the JIRA. It will also help for future code maintenance.

jinmeiliao: It would bee nice to run a full regression with this change to see if it will cause another hang. Right now I don't think I am fully qualified to explain to you on that.

dschneider-pivotal, line 1287: Since you moved this up to be before "waitForIndexInit" is called you need to add a try/finally that started right after lockForCacheModification returns and the finally should be way down past the finally for the try that started right after "lockWhenRegionIsInitializing" returns. If you look at the finally for that try you will see it calls releaseCacheModificationLock. This needs to instead be called in your new finally block so that it corresponds to the lockForCacheModification call.
It should have probably been done this way already (two finally blocks; one that calls releaseCacheModificationLock and another that calls unlockWhenRegionIsInitializing) but it is needed now because waitForIndexInit can throw an exception because it calls checkCancelInProgress.

dschneider-pivotal (changes requested):

geode-core/src/main/java/org/apache/geode/internal/cache/AbstractRegionMap.java
agingade: @jchen21 
There is locking mechanism to co-ordinate between cache modification (index maintenance) and index initialization. During index initialization all the cache operations are blocked.
As part of index management; clear invokes index re-initialization. 
Also, during clear no cache operations are allowed; this is co-ordinated through lockForCacheModification (version lock).
The hang is observed due to:
Thread1 calling update, which blocks index initialization.
Thread2 calling clear, takes the cache modification lock, this blocks all cache operation; and clear triggers index re-initialization; this is now blocked for thread1 to finish update. But thread1 is waiting for cache modification lock which is held by thread2; resulting in dead-lock.

agingade (approved):
Approved; assuming darrel's suggestion of calling releaseCacheModification under new try/finally block is taken care of.

geode-core/src/integrationTest/java/org/apache/geode/cache/query/partitioned/PRClearIntegrationTest.java
```diff
@@ -0,0 +1,73 @@
...
+        }
+      });
+    });
+
+    CompletableFuture<Void> clear = executor.runAsync(() -> {
+      Thread.currentThread().setName("Clear-Thread");
+      region.clear();
+    });
+
+    put.get(5, TimeUnit.SECONDS);
```
agingade, line 69: Can we add some kind of validation to make sure put/invalidate and clear is performed as expected. By looking at region contents?

jinmeiliao: that's not the purpose of this test and if we want to do that, we would make the test far more complicated since it's multi-threading and when we do assert after clear it's not guaranteed this region content is empty. 

I am sure we have plenty of other tests that verifies this.

geode-core/src/main/java/org/apache/geode/internal/cache/AbstractRegionMap.java
jchen21: Thanks @agingade ! This is exactly what I was looking for. Your comment is automatically logged in the JIRA, so we have the context in JIRA now. Looking at the code change of #44431 and #45699, where `waitForIndexInit()` and `lockForCacheModification()` are introduced respectively, I don't see any side effect of swapping the order of them. So I expect it should be good to fix this hang, as long as Darrel's concern is addressed.

dschneider-pivotal (approved):

PR #5828, opened 2020-12-09
GEODE-8496: Bump jackson from 2.11.3 to 2.12.1

onichols-pivotal:
1 test fails after this bump.  it appears that the test was capturing a "bad" behavior that is now [prevented](https://github.com/FasterXML/jackson-databind/issues/2683) in Jackson 2.12

to reproduce: `./gradlew :geode-core:integrationTest --tests DistributedSystemMBeanIntegrationTest.queryAllUsingGfshDoesNotFormatDate`

### DIFF
--- a/boms/geode-all-bom/src/test/resources/expected-pom.xml
+++ b/boms/geode-all-bom/src/test/resources/expected-pom.xml
@@ -62,18 +62,6 @@
         <scope>compile</scope>
       </dependency>
       <dependency>
-        <groupId>com.fasterxml.jackson.datatype</groupId>
-        <artifactId>jackson-datatype-joda</artifactId>
-        <version>2.9.8</version>
-        <scope>compile</scope>
-      </dependency>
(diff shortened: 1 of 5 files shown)


rhoughton-pivot (approved):

onichols-pivotal:
@jinmei it looks like the test DistributedSystemMBeanIntegrationTest.queryAllUsingGfshDoesNotFormatDate was expecting Date objects to be serialized a POJOs (i.e. the test expects to see the internal serialized fields of a Date object).  I've updated it to instead expect the new failure message.  Is this an acceptable change in default behavior?  Perhaps better would be to start including the jsr310 jar in the shipping product, instead of as test-runtime-only dependency, so that date formatting always displays *something* out-of-the-box?

dickcav (approved):

jdeppe-pivotal (approved):

kirklund (approved):

PR #5832, opened 2020-12-09
GEODE-8778: Change conserve-sockets default value to false

DonalEvans:
- Update certain tests to explicitly set conserve-sockets=true to allow
them to pass with the new default
- Update docs to reflect new default

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/distributedTest/java/org/apache/geode/internal/tcp/CloseConnectionTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/internal/tcp/CloseConnectionTest.java
@@ -48,6 +48,10 @@ public class CloseConnectionTest implements Serializable {
   @Rule
   public DistributedRule distributedRule = new DistributedRule();
 
+  /**
+   * The test case in this class requires that conserve-sockets=true in order for the connections to
+   * be shared.
+   */
   @Rule
   public CacheRule cacheRule = new CacheRule.Builder().addConfig(CONSERVE_SOCKETS, "true").build();
 
--- a/geode-core/src/distributedTest/java/org/apache/geode/internal/tcp/CloseConnectionTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/internal/tcp/CloseConnectionTest.java
@@ -14,6 +14,7 @@
  */
 package org.apache.geode.internal.tcp;
 
+import static org.apache.geode.distributed.ConfigurationProperties.CONSERVE_SOCKETS;
 import static org.apache.geode.test.awaitility.GeodeAwaitility.await;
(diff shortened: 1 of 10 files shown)


pivotal-jbarrett (commented):
Why are the two tests modified in this PR setting the the value against the new default? If the default is `false` now it stands to reason that unless a tests is explicitly testing socket conservation that it should use the new default of `false`.

DonalEvans:
> Why are the two tests modified in this PR setting the the value against the new default? If the default is `false` now it stands to reason that unless a tests is explicitly testing socket conservation that it should use the new default of `false`.

Thanks for bringing this up, Jake. After looking closer at the first test, `ClusterDistributionManagerDUnitTest`, it doesn't need to explicitly set conserve-sockets to true, just add an additional ignored exception due to a slightly different code path resulting in (expected) logging about `ack-severe-alert-threshold` being exceeded. I've modified the test to reflect this.

The second test, `CloseConnectionTest` is explicitly testing the behaviour of shared connections, so it requires that conserve-sockets=true. I think it was just an oversight to not set the value and instead rely on the default.

geode-core/src/distributedTest/java/org/apache/geode/internal/tcp/CloseConnectionTest.java
```diff
@@ -40,14 +41,15 @@
...
+  private static final long serialVersionUID = 3692493564204797623L;
   private VM vm0;
   private VM vm1;
 
   @Rule
   public DistributedRule distributedRule = new DistributedRule();
 
   @Rule
-  public CacheRule cacheRule = new CacheRule();
+  public CacheRule cacheRule = new CacheRule.Builder().addConfig(CONSERVE_SOCKETS, "true").build();
```
pivotal-jbarrett, line 52: Perhaps if this test is explicitly testing the conserver socket `true` scenario the name of the test class or some java docs should indicate that.

pivotal-jbarrett (approved):
Reviewed for content and looks good. Would like to see a comment about the 1 test that is explicitly setting conserve sockets `true`. I have not review for correctness or test failures.

davebarnes97 (approved):
+1 for all the doc changes (mods to .erb source files).

geode-core/src/distributedTest/java/org/apache/geode/internal/tcp/CloseConnectionTest.java
DonalEvans: The test class only contains one test case, which is `sharedSenderShouldRecoverFromClosedSocket()`. It's not immediately obvious that this implies `conserve-sockets=true` so adding some javadocs would be a good idea, I think.

onichols-pivotal (approved):

PR #5839, opened 2020-12-10
GEODE-7861: Improve error reporting in GMSJoinLeave.join()

kamilla1201:
When a member is unable to contact a join the system due to a failure to contact a locator, we could be a lot more specific, which would help users debug issues in their environment. We currently print out

> Unable to join the distributed system.  Operation either timed out, was stopped or Locator does not exist.

This change makes the exception message more specific. Instead of printing out the same message for 3 different problems, this change separates it out into 3 different messages. In case of a failure to contact a locator, the exception will include the underlying exception from the last locator we tried to contact as a cause. Also, the exception message will include the list of locators we tried to contact.

### For all changes:

### Note:

### DIFF
--- a/geode-membership/src/integrationTest/java/org/apache/geode/distributed/internal/membership/gms/membership/GMSJoinLeaveJUnitTest.java
+++ b/geode-membership/src/integrationTest/java/org/apache/geode/distributed/internal/membership/gms/membership/GMSJoinLeaveJUnitTest.java
@@ -22,6 +22,7 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
+import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.anyBoolean;
 import static org.mockito.ArgumentMatchers.anyInt;
 import static org.mockito.ArgumentMatchers.eq;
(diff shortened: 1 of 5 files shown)


geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/membership/GMSJoinLeave.java
```diff
@@ -383,40 +383,44 @@ public boolean join() throws MemberStartupException {
...
+        // unable to contact any of the locators
+        if (state.hasContactedAJoinedLocator) {
+          throw new MemberStartupException("Unable to join the distributed system in "
+              + (System.currentTimeMillis() - startTime) + "ms");
+        }
 
-      return this.isJoined;
+        if (state.locatorsContacted == 0) {
+          throw new MembershipConfigurationException(
+              "Unable to join the distributed system. Could not contact any of the locators: "
```
echobravopapa, line 414: as I'm understanding this change, this section is the critical change to improve upon error reporting... I don't see unit testing for the improvement; there is a new test that looks for the above Exception string. Only it seems that a test could validate the expanded error information has been sent as well.

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/membership/GMSJoinLeave.java
```diff
@@ -428,6 +432,24 @@ public boolean join() throws MemberStartupException {
     }
   }
 
+  boolean hasCoordinatorJoinedCluster(int viewId, long retrySleep)
```
echobravopapa, line 435: looks like this extraction was test driven

echobravopapa (changes requested):
This is looking good, the error logging now includes the list of Locators.
two things I'd like to see addressed
- the feature request asked to include "underlying exception from the last locator we tried to contact as a cause..."  
- please add a bit more detail to your commit messages to give a preview of the change to reviewers

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/membership/GMSJoinLeave.java
bschuchardt: How about naming this something indicating that a sleep is going to be performed, like "pauseIfThereIsNoCoordinator"?

bschuchardt (changes requested):

geode-membership/src/integrationTest/java/org/apache/geode/distributed/internal/membership/gms/membership/GMSJoinLeaveJUnitTest.java
```diff
@@ -115,11 +117,18 @@ public void initMocks(boolean enableNetworkPartition) throws Exception {
 
   public void initMocks(boolean enableNetworkPartition, boolean useTestGMSJoinLeave)
       throws Exception {
+    String locator = "localhost[12345]";
+    initMocks(enableNetworkPartition, useTestGMSJoinLeave, locator, locator);
+  }
+
+  public void initMocks(boolean enableNetworkPartition, boolean useTestGMSJoinLeave,
+      String locators, String startLoactor)
```
bschuchardt, line 125: typo:  "startLocator"

bschuchardt (approved):

echobravopapa (approved):

PR #5840, opened 2020-12-11
GEODE-8782: Add getPrincipal method to FunctionContext interface

jdeppe-pivotal:
- Add the ability to retrieve the Principal from the FunctionContext
  when a SecurityManager is enabled.

Authored-by: Jens Deppe <jdeppe@vmware.com>

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/execute/FunctionRemoteContext.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/execute/FunctionRemoteContext.java
@@ -31,8 +31,6 @@ import org.apache.geode.internal.serialization.StaticSerialization;
 
 /**
  * FunctionContext for remote/target nodes
- *
- *
  */
 public class FunctionRemoteContext implements DataSerializable {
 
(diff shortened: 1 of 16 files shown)


jinmeiliao:
I see you are binding the subject to the threads running the function and also adding the the parameter `principal` to the constructor of the `FunctoinContext`, I wonder if we only need to do one of the two? If subject is already in the thread context, you don't need to pass it in the constructor, you can just directly get it from the thread local by calling `getCache().getSecurityService().getPrinicipal()`

geode-core/src/main/java/org/apache/geode/internal/cache/execute/FunctionRemoteContext.java
```diff
@@ -84,6 +85,10 @@ public void fromData(DataInput in) throws IOException, ClassNotFoundException {
       this.bucketArray = BucketSetHelper.fromSet(bucketSet);
     }
     this.isReExecute = DataSerializer.readBoolean(in);
+
+    if (StaticSerialization.getVersionForDataStream(in).isNotOlderThan(KnownVersion.GEODE_1_14_0)) {
```
upthewaterspout, line 89: Instead of this, I think you could make this class implement VersionedDataSerializable which might handle writing the principal differently.

upthewaterspout (approved):

geode-core/src/main/java/org/apache/geode/internal/cache/execute/FunctionRemoteContext.java
jdeppe-pivotal: I tried that but had some upgrade tests that became flaky. I may have done something wrong, but using `VersionedDataSerializable` seemed to not work for this.

jdeppe-pivotal:
@jinmeiliao Yes, good idea. I've added the principal to `FunctionContextImpl` and removed the code that rebound the `Subject` on function threads.

@upthewaterspout I've also removed the default implementation for `FunctionContext.getPrincipal()` since FunctionContext (and impls) are never serialized anywhere and it is also never implemented by user code.

jinmeiliao (approved):

jinmeiliao:
After you are done with this, can you file another jira ticket to redo the `QueryCommand`? Currently we are sending the principal using `DataCommandRequest`, now that the principal is in the function context, we don't need to do that anymore. Or if you can lump it into this PR, that would be great too!

PR #5843, opened 2020-12-12
Feature/geode 8768

mivanac:

### For all changes:

### Note:


kohlmu-pivotal:
@mivanac Is this a "work in progress" PR? If so, please change its status to "Draft".
Otherwise, @nabarunnag @agingade would is be possible to review these changes?

mivanac:
This PR is ready to be reviewed.

mkevo (approved):

pivotal-amurmann:
Adding folks who according to CODEOWNERS should review this.
@jinmeiliao @jdeppe-pivotal @kirklund 
@boglesby @gesterzhou 
@bschuchardt @Bill 

Wow, lots of owners for one PR!

bschuchardt (changes requested):
The changes have left DistributionLocatorId in a bad state.  Its class javadoc needs to be updated to include hostNameForClients and memberName, and the equality-check methods need to be documented.  The name "additionalEqualityCheck", in particular, seems meaningless to me and I had to read its code and look at its uses in order to figure out what it does.  A javadoc will help but I think a more meaningful name is needed for this new method.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/misc/WanLocatorDiscoveryDUnitTest.java
```diff
@@ -0,0 +1,132 @@
...
+    ports = AvailablePortHelper.getRandomAvailableTCPPorts(5);
+    int site1Port =
+        setupWanSite1();
+    setupWanSite2(site1Port);
+  }
+
+  private int setupWanSite1() throws IOException {
+    Properties locator_ln_props = new Properties();
+    // create a cluster
+    locator_ln_props.setProperty(MCAST_PORT, "0");
```
jinmeiliao, line 70: You don't need all these when you use ClusterStartupRule. The rule sets those properties for you already so that you can avoid all these boiler code.

The rule also allocates the available ports for you so that you don't need to pre-allocate them.

use something like:

locator = cluster.startLocatorVM(index, l->l.withPropety(....).withConnectionToLocator(port0, port1, port2));

jinmeiliao (changes requested):

bschuchardt (approved):
Thank you Mario

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/misc/WanLocatorDiscoveryDUnitTest.java
```diff
@@ -65,25 +65,29 @@ private void setupWanSites() throws IOException {
...
-    Properties locator_ln_props = new Properties();
-    // create a cluster
-    locator_ln_props.setProperty(MCAST_PORT, "0");
-    locator_ln_props.setProperty(DISTRIBUTED_SYSTEM_ID, "1");
-
-    locator_ln1 = cluster.startLocatorVM(0, ports[0], locator_ln_props);
-    locator_ln2 = cluster.startLocatorVM(1, ports[1], locator_ln_props, locator_ln1.getPort());
-    return locator_ln1.getPort();
+    locator_ln1 = cluster.startLocatorVM(0, ports[0], VersionManager.CURRENT_VERSION,
+        i -> i.withProperty(MCAST_PORT, "0").withProperty(DISTRIBUTED_SYSTEM_ID, "1"));
```
jinmeiliao, line 69: you can get rid of `withProperty(MCAST_PORT, "0")`, it's already set by the rule

jinmeiliao (approved):

bschuchardt (approved):

geode-core/src/main/java/org/apache/geode/internal/admin/remote/DistributionLocatorId.java
```diff
@@ -104,6 +122,26 @@ public DistributionLocatorId(InetAddress host, int port, String bindAddress, SSL
...
+  /**
+   * Constructs a DistributionLocatorId with a String of the form: hostname[port] or
+   * hostname:bindaddress[port] or hostname@bindaddress[port]
+   * and membername
+   * <p>
+   * The :bindaddress portion is optional. hostname[port] is the more common form.
+   * <p>
+   * Example: merry.gemstone.com[7056]<br>
+   * Example w/ bind address: merry.gemstone.com:81.240.0.1[7056], or
+   * merry.gemstone.com@fdf0:76cf:a0ed:9449::16[7056]
```
gesterzhou, line 137: is this "::" a typo here?

geode-core/src/main/java/org/apache/geode/internal/admin/remote/DistributionLocatorId.java
```diff
@@ -104,6 +122,26 @@ public DistributionLocatorId(InetAddress host, int port, String bindAddress, SSL
    * two.
    */
   public DistributionLocatorId(String marshalled) {
+    this(marshalled, null);
+  }
+
+  /**
+   * Constructs a DistributionLocatorId with a String of the form: hostname[port] or
+   * hostname:bindaddress[port] or hostname@bindaddress[port]
+   * and membername
```
gesterzhou, line 131: it's better to put an example of membername here.

geode-core/src/main/java/org/apache/geode/internal/admin/remote/DistributionLocatorId.java
```diff
@@ -317,6 +363,43 @@ public boolean equals(Object other) {
...
+    if (this.membername != null && that.membername != null) {
+      if (this.membername.equals(that.membername))
+        return true;
+
+      return false;
+    }
+
+    if (!StringUtils.equals(this.hostnameForClients, that.hostnameForClients))
+      return false;
+    if (this.host != that.host && !(this.host != null && this.host.equals(that.host)))
```
gesterzhou, line 375: I think it should be "||", not "&&".

gesterzhou (changes requested):

geode-core/src/main/java/org/apache/geode/internal/admin/remote/DistributionLocatorId.java
gesterzhou: Bruce suggested to use:

if (!Objects.equals(this.host, that.host) {
  return false;
}

gesterzhou (approved):

ladyVader:
It seems that DistributionLocatorId.equals() can return true if one of the two compared objects doesn't have the memberName and the other does ... but DistributionLocatorId.hashCode() will be different ... if we invoke contains() on a Set of DistributionLocatorIds ... contains() will return false in that case.

I also found the logging a bit misleading once toString() didn't include the memberName.

mivanac:
Equals() and hashCode(), are modified to support cases, when sites with old implementation are discovered with new locators. Also this is the reason why membername is not included in toString().

mivanac:
If this is creating problems in your tests, I will need to modify logic.

bschuchardt:
@mivanac I think the best thing to do at this point is revert the merge and write some tests to fix the problems with

1. membername being an empty string.  In this case all other fields are ignored.
2. hashcode returning a value that places IDs having/not-having membernames in different hashmap buckets

I noticed #1 when reviewing code when people were talking about #2.  I wrote a test for #1.  A test for #2 is harder to write.

DistributionLocatorIdJUnitTest.java
```java
@Test
public void testEqualsWithMemberId() throws UnknownHostException {
  InetAddress address = InetAddress.getLocalHost();
  DistributionLocatorId dLI1 = new DistributionLocatorId(address, 40404, "127.0.0.1", null);
  DistributionLocatorId dLI2 = dLI1;
  @SuppressWarnings("RedundantStringConstructorCall")
  DistributionLocatorId dLI3 =
      new DistributionLocatorId(address, 40404, new String("127.0.0.1"), null);
  @SuppressWarnings("RedundantStringConstructorCall")
  DistributionLocatorId dLI4 = new DistributionLocatorId(InetAddress.getByName("localhost"),
      40405, new String("127.0.0.1"), null);
  dLI1.setMembername("");
  dLI3.setMembername("");
  dLI4.setMembername("");
  assertTrue(dLI1.equals(dLI2));
  assertTrue(dLI1.equals(dLI3));
  assertFalse(dLI1.equals(dLI4));
}
```

PR #5851, opened 2020-12-14
GEODE-8790: Evicted entries sometimes remain in region map

luissson:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/map/RegionMapDestroy.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/map/RegionMapDestroy.java
@@ -366,11 +366,17 @@ public class RegionMapDestroy {
           retry = true;
           doContinue = true;
           return;
-        } else if (!isEviction) {
+        } else {
           try {
-            handleEntryNotFound();
+            if (!isEviction) {
+              handleEntryNotFound();
+            }
           } finally {
-            removeEntryOrLeaveTombstone();
+            if (isEviction && newRegionEntry.isTombstone()) {
+              return;
+            } else {
+              removeEntryOrLeaveTombstone();
+            }
           }
         }
       } // synchronized(newRegionEntry)
--- a/geode-core/src/test/java/org/apache/geode/internal/cache/AbstractRegionMapTest.java
+++ b/geode-core/src/test/java/org/apache/geode/internal/cache/AbstractRegionMapTest.java
@@ -634,11 +634,7 @@ public class AbstractRegionMapTest {
         anyBoolean(), anyBoolean());
     verify(arm._getOwner(), never()).basicDestroyPart3(any(), any(), anyBoolean(), anyBoolean(),
         anyBoolean(), any());
-    // This seems to be a bug. We should not leave an entry in the map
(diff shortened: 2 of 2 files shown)


geode-core/src/main/java/org/apache/geode/internal/cache/map/RegionMapDestroy.java
```diff
@@ -372,6 +372,8 @@ private void retryRemoveWithTombstone() {
           } finally {
             removeEntryOrLeaveTombstone();
           }
+        } else if (regionEntry == null && !newRegionEntry.isTombstone()) {
```
dschneider-pivotal, line 375: I think it would be better to change it so the only the we don't do if we are evicting is handleEntryNotFound(). When evicting we do not want an exception if the entry is already gone.
So I would change line 369 to just be an "else" and then put a "if (!isEviction)" on handleEntryNotFound (even better might be to change handleEntryNotFound to test isEviction and return since this is the only place it is called).
Then you can get rid of your new code on 375 and 376.
I do think it is probably correct to call removeEntryOrLeaveTombstone for an eviction just to make sure it follows the normal rules of scheduling a tombstone when concurrency checks are enabled. This method is already a noop if evicting without concurrency checks.

dschneider-pivotal (changes requested):

geode-core/src/main/java/org/apache/geode/internal/cache/map/RegionMapDestroy.java
```diff
@@ -528,6 +528,9 @@ private void handleEntryNotFound() {
     boolean throwException = false;
     EntryNotFoundException entryNotFoundException = null;
 
+    if(isEviction)
```
dschneider-pivotal, line 531: This is just a code style thing but add curlys around the return statement.

dschneider-pivotal (changes requested):
Looks good to me. Just add those missing curly brackets.

geode-core/src/main/java/org/apache/geode/internal/cache/map/RegionMapDestroy.java
luissson: done!

geode-core/src/main/java/org/apache/geode/internal/cache/map/RegionMapDestroy.java
luissson: I moved a bit fast here with the first solution, as it appears we fail to pass the unit test evictDestroyOfExistingTombstoneWithConcurrencyChecksReturnsFalse. Suggesting that we don't properly handle the case where we're in an eviction and already set the region entry as tombstone. If we break up the check on isEviction and newRegionEntry.isTombstone we cause more unit tests to fail.

With the current issue, we're trying to catch the specific case where we have concurrency checks enabled, we're in an eviction, and we've created a new region entry marked as tombstone. I've restructured a some to avoid a clause with a lone call to removeEntryOrLeaveTombstone that addresses this issue and passes unit tests.

dschneider-pivotal (approved):

PR #5854, opened 2020-12-15
GEODE-6622: WAN rolling upgrade test bind exception

demery-pivotal:
Before: WANRollingUpgradeSecondaryEventsNotReprocessedAfterCurrentSiteMemberFailover allocated ports for two locators by making two separate calls to AvailablePort. AvailablePort randomly samples ports with replacement, and so two calls can return the same port if the port is not put into use between the calls.

Now: The test allocates both ports at the same time, and does so via AvailablePortHelper, which samples ports round robin rather than random with replacement.


mhansonp (approved):

mhansonp (approved):
Thanks for cleaning these up!

demery-pivotal:
The one failing distributed test (MemberMXBeanDistributedTest) is unrelated to this PR and is known to fail intermittently:
https://issues.apache.org/jira/browse/GEODE-8191

PR #5858, opened 2020-12-16
GEODE-8795: Lucene queries should utilize post-processing if enabled

jdeppe-pivotal:
- This now relies on the principal being available in the
  FunctionContext.
- PDX objects are passed directly to the post-processor thus avoiding
  any possible serialization issues.

### For all changes:

### Note:

### DIFF
--- a/geode-lucene/src/main/java/org/apache/geode/cache/lucene/internal/results/LuceneGetPageFunction.java
+++ b/geode-lucene/src/main/java/org/apache/geode/cache/lucene/internal/results/LuceneGetPageFunction.java
@@ -28,11 +28,15 @@ import org.apache.geode.cache.execute.FunctionContext;
 import org.apache.geode.cache.execute.RegionFunctionContext;
 import org.apache.geode.cache.partition.PartitionRegionHelper;
 import org.apache.geode.internal.cache.EntrySnapshot;
+import org.apache.geode.internal.cache.InternalCache;
 import org.apache.geode.internal.cache.PrimaryBucketException;
 import org.apache.geode.internal.cache.Token;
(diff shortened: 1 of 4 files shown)


jdeppe-pivotal:
Almost all the work is still from @nabarunnag. I had to make a different change to bring the principal into the `FunctionContext`.

One difference is that the original fix was not passing a `PDXInstance` (if PDX was in use) to the post-processor but was instead deserializing to the domain object first. If we just pass whatever is retrieved (PDX or otherwise) the post processor can then make a decision how to handle it. Otherwise the user needs to ensure that their domain classes are available on the server.

geode-lucene/src/main/java/org/apache/geode/cache/lucene/internal/results/LuceneGetPageFunction.java
```diff
@@ -65,15 +70,19 @@ public void execute(FunctionContext context) {
...
     if (entry == null) {
       return null;
     }
 
-    final Object value = entry.getRegionEntry().getValue(null);
+    Object value = entry.getRegionEntry().getValue(null);
     if (value == null || Token.isInvalidOrRemoved(value)) {
       return null;
+    } else if (securityService.needPostProcess()) {
+      value = securityService.postProcess(principal, region.getFullPath(), key, entry.getValue(),
```
upthewaterspout, line 84: I'm a little concerned that `value` here might be a CachedDeserializable in some cases. I think this whole function and PageEntry class are designed to try to pass the value back to the user without deserializing it if it can. 

However I can't think of a case that you didn't test already, so I guess it's doing the right thing.

upthewaterspout (approved):

PR #5862, opened 2020-12-17
GEODE-8799: Increase defaults for OperationExecutors.MAX_THREADS and ClusterOperationExecutors.MAX_PR_THREADS

DonalEvans:
- OperationExecutors.MAX_THREADS new default is 300
- ClusterOperationExecutors.MAX_PR_THREADS new default is
Math.max(Runtime.getRuntime().availableProcessors() * 32, 200))
- Removed unit test case for setting system property, since it cannot be
cleared afterwards without unloading the class

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/distributed/internal/OperationExecutors.java
+++ b/geode-core/src/main/java/org/apache/geode/distributed/internal/OperationExecutors.java
@@ -26,7 +26,7 @@ import org.apache.geode.distributed.internal.membership.InternalDistributedMembe
  */
 public interface OperationExecutors {
   int MAX_THREADS =
-      Integer.getInteger("DistributionManager.MAX_THREADS", 300);
+      Integer.getInteger("DistributionManager.MAX_THREADS", 1000);
 
   int MAX_FE_THREADS = Integer.getInteger("DistributionManager.MAX_FE_THREADS",
       Math.max(Runtime.getRuntime().availableProcessors() * 16, 100));
--- a/geode-core/src/main/java/org/apache/geode/distributed/internal/OperationExecutors.java
+++ b/geode-core/src/main/java/org/apache/geode/distributed/internal/OperationExecutors.java
@@ -26,7 +26,7 @@ import org.apache.geode.distributed.internal.membership.InternalDistributedMembe
  */
 public interface OperationExecutors {
   int MAX_THREADS =
-      Integer.getInteger("DistributionManager.MAX_THREADS", 100);
+      Integer.getInteger("DistributionManager.MAX_THREADS", 300);
 
   int MAX_FE_THREADS = Integer.getInteger("DistributionManager.MAX_FE_THREADS",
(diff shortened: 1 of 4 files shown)


geode-core/src/main/java/org/apache/geode/internal/cache/properties.html
```diff
@@ -383,7 +383,7 @@ <h1>Properties, in Alphabetical Order</h1>
 <dd>
 <em>Public:</em> false
 <p>
-<em>Integer</em> (default is 1)
+<em>Integer</em> (default is 200)
```
jhutchison, line 386: not sure if I'm missing something, but it looks like default will be 200, if the machine has less than 6 cores, other wise the default is num processors * 32. Not sure how significant this info is to our users?

jhutchison (commented):
+1

geode-core/src/main/java/org/apache/geode/internal/cache/properties.html
DonalEvans: Thanks for pointing this out, I'll update the description.

geode-core/src/main/java/org/apache/geode/distributed/internal/OperationExecutors.java
```diff
@@ -26,7 +26,7 @@
  */
 public interface OperationExecutors {
   int MAX_THREADS =
-      Integer.getInteger("DistributionManager.MAX_THREADS", 100);
+      Integer.getInteger("DistributionManager.MAX_THREADS", 300);
```
agingade, line 29: There is another ticket to address this issue: GEODE-8357
This change has an impact on overall system requirement and performance; it will be nice to send a proposal to developer community. This helps to get feedback from engineers who had encountered issue with this setting and suggested the recommendation to the users (support teams); there was conversation about changing this to big number like 1000. 
It will be good to come up with a right number based on previous experience in setting this number.

We also need to think about setting the system properties automatically; some kind of auto-tuning service in the product that will adopt/changes the value based on the resource availability and the system load.

nabarunnag: @agingade we created this PR after receiving feedback from customers and support. This is one of the requested features from the field and these are the numbers we have received from support.
We are also aligning these numbers with MAX_FE_THREADS, which was changed to cpu*32 to tune the internal tests.

This can be a temporary band-aid till we start on GEODE-8357, which has more extensive changes and we can involve the community feedback.

DonalEvans: Having talked to some other people with more hands-on experience with hitting the limit currently set by `MAX_THREADS`, a value of 1000 was decided to be more appropriate. The PR has been updated to reflect this.

boglesby (approved):

pivotal-eshu (approved):
+1

PR #5872, opened 2020-12-23
GEODE-8802: clarify mcast-port default in docs

karensmolermiller:

### DIFF
--- a/geode-docs/configuring/running/firewalls_ports.html.md.erb
+++ b/geode-docs/configuring/running/firewalls_ports.html.md.erb
@@ -166,7 +166,8 @@ This table contains properties potentially involved in firewall behavior, with a
 <tr>
 <td><p>Multicast</p></td>
 <td><code class="ph codeph">mcast-port</code></td>
-<td>10334</td>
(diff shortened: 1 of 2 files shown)


davebarnes97 (approved):

upthewaterspout (commented):
I think the default is actually 0, period. @Bill or @bschuchardt  might be able to clarify.

bschuchardt (changes requested):
The default setting is zero, regardless of whether locators are given or not.

geode-docs/configuring/running/firewalls_ports.html.md.erb
```diff
@@ -166,7 +166,8 @@ This table contains properties potentially involved in firewall behavior, with a
 <tr>
 <td><p>Multicast</p></td>
 <td><code class="ph codeph">mcast-port</code></td>
-<td>10334</td>
+<td>0 if the <code class="ph codeph">locators</code> property is set,
+and 10334 otherwise</td>
```
upthewaterspout, line 170: I think this part is still incorrect - it talks about a 10334 default.

bschuchardt (approved):

PR #5879, opened 2021-01-06
GEODE-8800: Slower receiver is not allowed with cluster TLS/SSL

gesterzhou:
Co-authored-by: Xiaojian Zhou <zhouxh@vmware.com>

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/integrationTest/java/org/apache/geode/internal/SSLConfigJUnitTest.java
+++ b/geode-core/src/integrationTest/java/org/apache/geode/internal/SSLConfigJUnitTest.java
@@ -56,9 +56,9 @@ import static org.apache.geode.distributed.ConfigurationProperties.SSL_ENABLED_C
 import static org.apache.geode.distributed.ConfigurationProperties.SSL_ENDPOINT_IDENTIFICATION_ENABLED;
 import static org.apache.geode.internal.security.SecurableCommunicationChannel.ALL;
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.catchThrowable;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.fail;
 
 import java.util.Map.Entry;
 import java.util.Properties;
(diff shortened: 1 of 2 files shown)


dschneider-pivotal (changes requested):
The change you made looks good but I think you need to make an additional change. If you look at the method Connection.readHandshakeForSender you will see that it is possible for a member that has not configured async-distribution-timeout to set it to true for a particular sender it creates to another member. So, just to be complete, I'd like you to also do a check in that method and if the sender will be using SSL to fail the handshake if it requires async.

geode-core/src/main/java/org/apache/geode/distributed/internal/DistributionConfigImpl.java
```diff
@@ -1030,6 +1032,20 @@ public DistributionConfigImpl(Properties nonDefault, boolean ignoreGemFirePropsF
     modifiable = false;
   }
 
+  /*
+   * When ssl-enabled-components specified CLUSTER or ALL, the slower receiver is not allowed.
```
dschneider-pivotal, line 1036: I noticed here and in your test method names you use "slower". Change that to "slow". It is a "slow receiver" not a "slower receiver".

dschneider-pivotal (changes requested):

geode-core/src/integrationTest/java/org/apache/geode/internal/SSLConfigJUnitTest.java
```diff
@@ -153,6 +156,66 @@ public static void initializeSSLMaps() {
   @After
   public void tearDownTest() {}
 
+  @Test
+  public void slowerReceiverShouldThrowExceptionWhenEnabledClusterTLS() throws Exception {
+    Properties props = new Properties();
+    props.setProperty(SSL_ENABLED_COMPONENTS, "cluster");
+    props.setProperty(ASYNC_DISTRIBUTION_TIMEOUT, "1");
+
+    try {
```
kirklund, line 165: I recommend using AssertJ catchThrowable for these expected exceptions. It would end up looking like:
```

Throwable thrown = catchThrowable(() -> new DistributionConfigImpl(props));
assertThat(thrown)
  .isInstanceOf(IllegalArgumentException.class)
  .hasMessageContaining("async-distribution-timeout greater than 0 is not allowed with cluster TLS/SSL.");
```

kirklund (approved):
I'm going to approve the PR, but please consider using catchThrowable and assertThat for expected exceptions.

jinmeiliao:
As far as I know, we rarely do multi-property cross validation at startup, i.e, we don't check if property A is set then property B  can't be set. Initial validation is usually based on one property only (i.e integer is between a number and string is only a certain values etc.).  Multi-property conflict is usually handled by ignoring one property (if A is set, then B has no effect). I wonder if we should take this route instead.

geode-core/src/main/java/org/apache/geode/distributed/internal/DistributionConfigImpl.java
gesterzhou: fixed.

gesterzhou:
We discussed the alternative to ignore the async-distribution-timeout when SSL is enabled. Finally we decided to throw exception instead of silently ignore it.

geode-core/src/integrationTest/java/org/apache/geode/internal/SSLConfigJUnitTest.java
gesterzhou: Fixed.

dschneider-pivotal:
My concern was that we can have a cluster in which all the members use SSL but we might only have some of them configure slow receiver. But if that happens then any of the members that configured slow receiver will fail to start (with this change) so I think this current PR is enough without adding an additional check in Connection.readHandshakeForSender

dschneider-pivotal (approved):

PR #5891, opened 2021-01-11
GEODE-8447: add seconds to the localized pattern

jinmeiliao:


mhansonp (approved):

geode-core/src/test/java/org/apache/geode/management/internal/json/QueryResultFormatterTest.java
```diff
@@ -119,16 +120,20 @@ public void testPrimitives() throws Exception {
...
+    Date date = new Date(time);
+    SimpleDateFormat format = DateFormatter.createLocalizedDateFormat();
+    String expectedString = format.format(date);
     QueryResultFormatter javaDateResult =
         new QueryResultFormatter(100).add(RESULT, date);
     checkResult(javaDateResult,
         "{\"result\":[[\"java.util.Date\",\"" + expectedString + "\"]]}");
 
-    java.sql.Date sqlDate = new java.sql.Date(0);
+    java.sql.Date sqlDate = new java.sql.Date(time);
```
kirklund, line 136: This is probably another old bug, but I'm fairly certain that `java.sql.Date` should be replaced by `java.util.Date`. The sql version should only be used with JDBC.

kirklund (changes requested):

geode-core/src/test/java/org/apache/geode/management/internal/json/QueryResultFormatterTest.java
jinmeiliao: This test is only to make sure our QueryResultFormatter can format `java.sql.Date` object. It's used to format customer region data, they could use any date time object.

jchen21 (approved):

geode-core/src/test/java/org/apache/geode/management/internal/json/QueryResultFormatterTest.java
jchen21: The query result can contain either `java.util.Date` or `java.sql.Date`. `java.util.Date` is tested in line 133.

kirklund (approved):

PR #5905, opened 2021-01-15
GEODE-8837: Establishes GFE_81 as the oldest supported client.

pivotal-jbarrett:

### For all changes:

### Note:


geode-core/src/distributedTest/java/org/apache/geode/internal/BackwardCompatibilitySerializationDUnitTest.java
```diff
@@ -59,11 +56,11 @@
...
   public static boolean toDataCalled = false;
-  public static boolean toDataPre66Called = false;
-  public static boolean toDataPre70called = false;
+  public static boolean toDataPre11Called = false;
+  public static boolean toDataPre15called = false;
   public static boolean fromDataCalled = false;
-  public static boolean fromDataPre66Called = false;
-  public static boolean fromDataPre70Called = false;
+  public static boolean fromDataPre11Called = false;
+  public static boolean fromDataPre15Called = false;
```
Bill, line 63: I see what was `66` is now `11` and what was `70` is now `15`. What is the significance of that mapping? How did you pick those replacements?

I get that you didn't want this test testing deprecated versions any more. But why test `1_1_0`and `1_5_0` and not say `1_13_0` and `1_14_0`? Is the choice arbitrary?

geode-core/src/integrationTest/java/org/apache/geode/cache/client/ClientCacheFactoryJUnitTest.java
```diff
@@ -367,13 +366,13 @@ public void testOldClientIDDeserialization() throws Exception {
...
-    assertThat(newMemberID.getVersion()).isEqualTo(KnownVersion.GFE_82);
-    assertThat(newID.getClientVersion()).isEqualTo(KnownVersion.GFE_82);
+    assertThat(newMemberID.getVersion()).isEqualTo(KnownVersion.GFE_81);
+    assertThat(newID.getClientVersion()).isEqualTo(KnownVersion.GFE_81);
 
     assertThat(newMemberID.getUuidLeastSignificantBits()).isEqualTo(0);
     assertThat(newMemberID.getUuidMostSignificantBits()).isEqualTo(0);
 
-    gmsID.setUUID(new UUID(1234L, 5678L));
+    ((MemberIdentifier) memberID).setUUID(new UUID(1234L, 5678L));
```
Bill, line 375: I believe this cast is unnecessary. Same for next two casts in this method.

geode-core/src/main/java/org/apache/geode/cache/client/internal/ClientSideHandshakeImpl.java
```diff
@@ -233,7 +230,7 @@ public ServerQueueStatus handshakeWithServer(Connection conn, ServerLocation loc
 
       member = readServerMember(dis);
 
-      serverQStatus = new ServerQueueStatus(endpointType, queueSize, member);
+      ServerQueueStatus serverQStatus = new ServerQueueStatus(endpointType, queueSize, member);
```
Bill, line 233: make `final`

geode-core/src/main/java/org/apache/geode/cache/client/internal/ClientSideHandshakeImpl.java
```diff
@@ -301,18 +298,18 @@ private InternalDistributedMember readServerMember(DataInputStream p_dis) throws
   public ServerQueueStatus handshakeWithSubscriptionFeed(Socket sock, boolean isPrimary)
       throws IOException, AuthenticationRequiredException, AuthenticationFailedException,
       ServerRefusedConnectionException, ClassNotFoundException {
-    ServerQueueStatus serverQueueStatus = null;
+    final ServerQueueStatus serverQueueStatus;
```
Bill, line 301: why not just return the value on line 354 and eliminate this variable entirely?

geode-core/src/main/java/org/apache/geode/cache/client/internal/ClientSideHandshakeImpl.java
```diff
@@ -336,20 +333,17 @@ public ServerQueueStatus handshakeWithSubscriptionFeed(Socket sock, boolean isPr
...
       }
-      HashMap instantiatorMap = DataSerializer.readHashMap(dis);
-      for (Iterator itr = instantiatorMap.entrySet().iterator(); itr.hasNext();) {
-        Map.Entry instantiator = (Map.Entry) itr.next();
-        Integer id = (Integer) instantiator.getKey();
-        ArrayList instantiatorArguments = (ArrayList) instantiator.getValue();
-        InternalInstantiator.register((String) instantiatorArguments.get(0),
-            (String) instantiatorArguments.get(1), id, false);
+      final Map<Integer, List<String>> instantiatorMap = DataSerializer.readHashMap(dis);
+      for (final Map.Entry<Integer, List<String>> entry : instantiatorMap.entrySet()) {
```
Bill, line 337: Good cleanup here! Could go a step further:

```
      instantiatorMap.entrySet().stream().forEach(
          entry -> InternalInstantiator.register(entry.getValue().get(0), entry.getValue().get(1),
                  entry.getKey(), false));
```

geode-core/src/main/java/org/apache/geode/cache/query/internal/cq/CqServiceProvider.java
```diff
@@ -59,7 +60,7 @@ public static CqService create(InternalCache cache) {
       return new MissingCqService();
     }
 
-    return factory.create(cache);
+    return factory.create(cache, CommandInitializer.getDefaultInstance());
```
Bill, line 63: is there a way to do this without calling the deprecated `getDefaultInstance`?

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/CommandInitializer.java
```diff
@@ -111,37 +89,76 @@
...
+   * @return legacy singleton instance. Instance is not immutable.
    */
-  public static void registerCommand(int messageType,
+  @Deprecated
+  public static CommandInitializer getDefaultInstance() {
+    return instance;
+  }
+
+  final Map<KnownVersion, Map<Integer, Command>> unmodifiableRegisteredCommands;
+  final LinkedHashMap<KnownVersion, ConcurrentMap<Integer, Command>> modifiableRegisteredCommands;
```
Bill, line 111: Why must we retain a reference to the modifiable commands? It looks like it is to support `CqServiceFactoryImpl` calling `register()`. Why do we keep the two separate maps?

Bill (commented):
I've asked a few questions and made a few suggestions and a request.

Hardest part of this PR is the big picture. It's hard for me to tell how to validate whether it was right to e.g. remove `Destroy.java` and `Destroy65.java`, and leave `Destroy70.java`. A map of the big picture showing where the code moved would be helpful. On the other hand, perhaps our tests are sufficient?

geode-core/src/distributedTest/java/org/apache/geode/internal/BackwardCompatibilitySerializationDUnitTest.java
pivotal-jbarrett: Yup, totally arbitrary really.

geode-core/src/main/java/org/apache/geode/cache/query/internal/cq/CqServiceProvider.java
pivotal-jbarrett: Not without a much deeper refactoring. We will need a way to connect two very distant components and that is just out of scope for what I want to get done now. So it's no worse than it was. It should discourage continued abuse. It is now testable. So I see a net positive here.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/CommandInitializer.java
pivotal-jbarrett: Ah... to maintain the contract that the maps returned by the `get(Version)` method are immutable views. The previous implementation made the outer map immutable but not the inner map so the caller could technically mutate the map causing who knows what issues. Now the only way to mutate is through `register`, which mutates the backing maps but keeps the immutable wrappers intact.

pivotal-jbarrett: You can think of the unmodifiable version of the map as just a map of unmodifiable wrappers. It doesn't contain any data itself. I am not committed to it though if you don't like it.

geode-serialization/src/test/java/org/apache/geode/internal/serialization/KnownVersionJUnitTest.java
```diff
@@ -44,18 +39,21 @@ public void testKnownVersionClass() throws Exception {
...
-    assertTrue(later.compareTo(later) == 0);
+    assertEquals(later, later);
+    // noinspection EqualsWithItself for testing
+    assertEquals(0, later.compareTo(later));
     assertTrue(earlier.compareTo(later) < 0);
 
     assertTrue(later.compareTo(Versioning.getVersion(earlier.ordinal())) > 0);
-    assertTrue(later.compareTo(Versioning.getVersion(later.ordinal())) == 0);
+    assertEquals(0, later.compareTo(Versioning.getVersion(later.ordinal())));
     assertTrue(earlier.compareTo(Versioning.getVersion(later.ordinal())) < 0);
```
DonalEvans, line 57: Instead of `assertTrue()`, might it be better to use `assertThat().isGreaterThan()` and `assertThat().isLessThan()` for these checks?

geode-core/src/distributedTest/java/org/apache/geode/internal/BackwardCompatibilitySerializationDUnitTest.java
Bill: ok.

geode-core/src/main/java/org/apache/geode/cache/query/internal/cq/CqServiceProvider.java
Bill: okey dokey

DonalEvans (commented):
Some suggestions and questions, but overall looks good.

geode-serialization/src/main/java/org/apache/geode/internal/serialization/KnownVersion.java
```diff
@@ -367,61 +377,48 @@ public static String unsupportedVersionMessage(final short ordinal) {
...
+    return Arrays.stream(VALUES).filter(x -> x != null && x != TEST_VERSION)
         .collect(Collectors.toList());
   }
 
   /**
    * package-protected for use by Versioning factory
    */
-  static KnownVersion getKnownVersionOrDefault(final short ordinal,
-      final KnownVersion defaultKnownVersion) {
+  static KnownVersion getKnownVersionOrDefault(final short ordinal) {
```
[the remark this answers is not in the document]
pivotal-jbarrett: It turns out it is only used in one place too. I think I am just going to remove it.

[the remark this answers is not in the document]
pivotal-jbarrett: I take that back, I won't remove it since it encapsulates some nastiness. But I will rename it.

geode-core/src/distributedTest/java/org/apache/geode/internal/BackwardCompatibilitySerializationDUnitTest.java
```diff
@@ -160,7 +149,7 @@ public void testFromDataFromLowerVersionToHigher() throws Exception {
   public void testAllMessages() throws Exception {
     // list of msgs not created using reflection
     // taken from DSFIDFactory.create()
-    ArrayList<Integer> constdsfids = new ArrayList<Integer>();
+    ArrayList<Integer> constdsfids = new ArrayList<>();
```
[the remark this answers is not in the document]
pivotal-jbarrett: Do you think this test is supposed to be testing all those types as well? I think that might be out of scope for this PR. My gut tells me this test was testing the idea of backwards compatibility and not the specifics of messages.

geode-core/src/integrationTest/java/org/apache/geode/cache/client/ClientCacheFactoryJUnitTest.java
pivotal-jbarrett: Very odd. I wonder what I did for it to add those casts. Fixing.

geode-core/src/main/java/org/apache/geode/cache/client/internal/ClientSideHandshakeImpl.java
pivotal-jbarrett: Could be said for so many places. What made you pick this one place I missed. ;)

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/CommandInitializer.java
pivotal-jbarrett: The unmodifiable maps are just wrappers and not copies so changes made to the modifiable map are visible to unmodifiable map.

geode-core/src/main/java/org/apache/geode/cache/client/internal/ClientSideHandshakeImpl.java
```diff
@@ -125,23 +123,23 @@ public static void setVersionForTesting(short ver) {
...
   private void setOverrides() {
-    this.clientConflation = determineClientConflation();
+    clientConflation = determineClientConflation();
 
     // As of May 2009 ( GFE 6.0 ):
     // Note that this.clientVersion is used by server side for accepting
     // handshakes.
     // Client side handshake code uses this.currentClientVersion which can be
     // set via tests.
     if (currentClientVersion.isNotOlderThan(KnownVersion.GFE_603)) {
```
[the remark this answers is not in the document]
pivotal-jbarrett: Removal of older version checks are largely out of scope for this PR. As per the proposal email, the goal is to set the ground work for the much larger effort of removal.

geode-core/src/test/java/org/apache/geode/internal/cache/tier/sockets/BaseCommandJUnitTest.java
```diff
@@ -27,18 +27,13 @@
...
 
 @RunWith(JUnitParamsRunner.class)
 public class BaseCommandJUnitTest {
 
   public BaseCommand[] getCommands() {
-    return new BaseCommand[] {(BaseCommand) ExecuteRegionFunction61.getCommand(),
-        (BaseCommand) ExecuteRegionFunction65
-            .getCommand(),
-        (BaseCommand) ExecuteRegionFunction66.getCommand()};
+    return new BaseCommand[] {(BaseCommand) ExecuteRegionFunction66.getCommand()};
```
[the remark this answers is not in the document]
pivotal-jbarrett: Keen eye!

geode-core/src/test/java/org/apache/geode/internal/cache/tier/sockets/CommandInitializerTest.java
```diff
@@ -0,0 +1,227 @@
...
+            "Please add a command set for " + version + " of Geode to CommandInitializer",
+            CommandInitializer.getDefaultInstance().get(version));
+      }
+    }
+  }
+
+  @Test
+  public void initializeGeode18Commands() {
+    @SuppressWarnings("unchecked")
+    final Map<Integer, Command> commands = mock(Map.class);
```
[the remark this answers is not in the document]
pivotal-jbarrett: Honestly I could delete this test. There is an integration test that does pretty much what you describe. The rational for this test was to assert that I didn't have any duplicates when I refactored each of the versions. Checking the map wouldn't weed out two puts on the same key but this does.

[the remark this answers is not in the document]
pivotal-jbarrett: I think I will keep because it is faster feedback on versioning commands than waiting for an integration test.

geode-cq/src/test/java/org/apache/geode/cache/query/cq/internal/CqServiceFactoryImplTest.java
```diff
@@ -0,0 +1,75 @@
...
+import org.apache.geode.distributed.DistributedSystem;
+import org.apache.geode.internal.cache.InternalCache;
+import org.apache.geode.internal.cache.tier.MessageType;
+import org.apache.geode.internal.cache.tier.sockets.CommandRegistry;
+import org.apache.geode.internal.serialization.KnownVersion;
+
+public class CqServiceFactoryImplTest {
+
+  @Test
+  public void initialize() {
```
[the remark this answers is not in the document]
pivotal-jbarrett: Yes, it really doesn't make sense anymore since initialize isn't the thing even being tested anymore. Thanks!

geode-serialization/src/test/java/org/apache/geode/internal/serialization/KnownVersionJUnitTest.java
pivotal-jbarrett: Static analyzer didn't pick that one up... Thanks!

DonalEvans (approved):

geode-serialization/src/main/java/org/apache/geode/internal/serialization/KnownVersion.java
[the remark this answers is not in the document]
Bill: I'm very glad you didn't remove it. Indeed it is a key part of the abstraction barrier.

geode-core/src/distributedTest/java/org/apache/geode/internal/BackwardCompatibilitySerializationDUnitTest.java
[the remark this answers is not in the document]
Bill: My comment was based on a quick read of the comment in the test method. That sent me over to `DSFIDFactory.create` where I found more cases than were accounted for in the test. Made me think those were added after the test and that we forgot to add them to the test.

geode-core/src/main/java/org/apache/geode/cache/client/internal/ClientSideHandshakeImpl.java
Bill: I have to make at least one `final` comment on every PR. Also this one just jumped out at me.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/CommandInitializer.java
Bill: Thanks @pivotal-jbarrett.

geode-serialization/src/main/java/org/apache/geode/internal/serialization/KnownVersion.java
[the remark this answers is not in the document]
pivotal-jbarrett: I am not in favor of adding parameters that are unused especially when the use is currently in a single place and likely not going to expand beyond that place, given it is package private. The caller never set anything other than `null` meaning that if there was no known value it wanted `null` back. I changed the method to do what the single use caller wanted and documented the behavior.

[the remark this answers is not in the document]
Bill: ok so you're gonna rename it?

geode-core/src/distributedTest/java/org/apache/geode/internal/BackwardCompatibilitySerializationDUnitTest.java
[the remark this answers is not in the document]
Bill: I'm willing to call this out of scope.

geode-core/src/main/java/org/apache/geode/cache/client/internal/ClientSideHandshakeImpl.java
Bill: just a suggestion

geode-serialization/src/main/java/org/apache/geode/internal/serialization/KnownVersion.java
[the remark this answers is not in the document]
Bill: I see you renamed it to `getKnownVersion`. I would have preferred it if you had taken my suggestion: `getKnownVersionOrNull` to make it clear to the caller that a null could come back. While we don't do that much in Geode, I did try hard to eliminate nulls everywhere I could in the versioning code when I did the rework last year. This change introduces a null in a null-less bastion. 

I'm going to accept this change anyway in the interest of harmony. Oh well.

PR #5915, opened 2021-01-15
GEODE-8840:implement redis RENAMENX command

jhutchison:

### For all changes:

### Note:


geode-redis/src/main/java/org/apache/geode/redis/internal/executor/key/RenameNxExecutor.java
```diff
@@ -0,0 +1,32 @@
...
+    List<ByteArrayWrapper> commandElems = command.getProcessedCommandWrappers();
+    ByteArrayWrapper key = command.getKey();
+    ByteArrayWrapper newKey = commandElems.get(2);
+    RedisKeyCommands redisKeyCommands = getRedisKeyCommands(context);
+
+    if (redisKeyCommands.exists(newKey)) {
+      return RedisResponse.integer(0);
+    }
+
+    if (!redisKeyCommands.rename(key, newKey)) {
```
jdeppe-pivotal, line 26: I think this will need a bit more work since there is a race condition here. Consider the case when two separate requests (on different servers) perform the check above and each request makes it through the check...

Before digging into that, I'd suggest writing a concurrent test to highlight the problem here.

jhutchison: will do, thx.

jhutchison:
this is no longer an immediate concern.  will address issue in future

PR #5925, opened 2021-01-20
GEODE-8671: Two threads calling get and retrieve the same PdxInstance, resulting in corruption

jchen21:

### For all changes:

### Note:


geode-core/src/distributedTest/java/org/apache/geode/internal/cache/RegionConcurrentOperationDUnitTest.java
```diff
@@ -131,6 +140,49 @@ public void getOnPreLoadedRegionFromMultipleThreadsReturnSameObject() throws Exc
...
+  @Test
+  public void getOnPartitionedRegionFromMultipleThreadsReturnsDifferentPdxInstances()
+      throws Exception {
+    String regionName = getClass().getSimpleName();
+    CacheFactory cacheFactory = new CacheFactory();
+    cacheFactory.setPdxReadSerialized(true);
+    cacheRule.createCache(cacheFactory);
+    InternalCache cache = cacheRule.getCache();
+    cache.setCopyOnRead(true);
+    Region region = cache.createRegionFactory(PARTITION)
```
DonalEvans, line 152: The compiler warning here (and elsewhere in the file) can be avoided by using `Region<Object, Object>`.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/RegionConcurrentOperationDUnitTest.java
```diff
@@ -131,6 +140,49 @@ public void getOnPreLoadedRegionFromMultipleThreadsReturnSameObject() throws Exc
...
+            return value;
+          }
+        }
+      };
+
+      // In this test, two threads are doing gets. One thread puts the value
+      // We expect that the threads will *always* get different PdxInstance values
+      Future get1 = executorServiceRule.submit(getValue);
+      Future get2 = executorServiceRule.submit(getValue);
+      Future put = executorServiceRule.submit(() -> region.put(key, new TestValue()));
```
DonalEvans, line 172: Compiler warnings here (and elsewhere in the file) can be avoided by using `Future<Object>`.

DonalEvans (approved):
Looks good, just a few small code clean-up suggestions.

upthewaterspout (approved):
Looks good. I'm curious to see whether this has any impact on performance in other areas since we changed the general CopyHelper class. But I think that is actually the right thing to do, since PdxInstances are not safe to share between threads.

upthewaterspout:
Minor nitpick - can you please provide a description of what you changed and why in the final commit message. I don't see one in this PR or the individual commits, but maybe I missed something. Thanks!

dschneider-pivotal (changes requested):
I think the real bug here is that some methods on PdxInstance are not thread safe. They should all be thread safe. Once we make them thread safe then we should return the same PdxInstance even if copy-on-read is true just like we do for other immutable objects.
If we can identify the PdxInstance methods that are not thread safe, then we can discuss how to make them so. The original ticket mentioned toString and serialization as not being thread safe. We need a test that demonstrates what is not thread safe.

kirklund (changes requested):
I recently created a new `DistributedBlackboard` Rule for dunit tests. You can then use it as an API:
```
@Rule
```
I updated several dunit tests to use it including:
```
geode-core/src/distributedTest/java/org/apache/geode/internal/cache/partitioned/fixed/FixedPartitioningDUnitTest.java
```

dschneider-pivotal (approved):

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/RegionConcurrentOperationDUnitTest.java
kirklund: If the Future is not going to have a result then go with `Future<Void>`.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/RegionConcurrentOperationDUnitTest.java
```diff
@@ -61,9 +63,14 @@ private static DUnitBlackboard getBlackboard() {
   @Rule
   public ExecutorServiceRule executorServiceRule = new ExecutorServiceRule();
 
+  @Rule
+  public DistributedBlackboard blackboard = new DistributedBlackboard();
+
   @After
   public void tearDown() {
-    blackboard.initBlackboard();
+    if (blackboard != null) {
```
kirklund, line 71: You can delete this `tearDown`. The `DistributedBlackboard` Rule handles this for you. (If it doesn't, that's a bug that I need to know about.)

kirklund (approved):
Looks like you have a few small changes to make but I'll give my approval now. Please consider adding unit test coverage for the changes in LocalRegion. There is an existing `LocalRegionTest` that uses Mockito. I also suggest you remove the `tearDown` since it's doing the same thing that the `DistributedBlackboard` Rule does after every test method.

kirklund:
> Looks like you have a few small changes to make but I'll give my approval now. Please consider adding unit test coverage for the changes in LocalRegion. There is an existing `LocalRegionTest` that uses Mockito. I also suggest you remove the `tearDown` since it's doing the same thing that the `DistributedBlackboard` Rule does after every test method.

By the way, I verified that you don't need to invoke `DUnitBlackboard.initBlackboard()` in the `tearDown`. The Rule completely dereferences the instance of `DUnitBlackboard` and so it becomes available to GC. The Rule handles the initialization and cleanup of `DUnitBlackboard`.

PR #5927, opened 2021-01-20
GEODE-8846: Fix for transaction commit failure

jvarenina:
Issue: Transaction commit fail when transactionID integer overflows to
negative value, because server interprets negative value as
non-transactional message.

Fix: Client is impacted to always overflow transactionID to zero instead
of to Integer.MIN_VALUE(-2147483648).

Additionally, this commit removes unused methods from TXManagerImpl.java:
- _incrementTXUniqueIDForReplay
- incrementTXUniqueIDForReplay

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClientServerTransactionDistributedTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClientServerTransactionDistributedTest.java
@@ -1,0 +1,152 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.apache.geode.internal.cache;
+
+import static org.apache.geode.test.dunit.VM.getHostName;
+import static org.apache.geode.test.dunit.VM.getVM;
+import static org.junit.Assert.assertEquals;
+
+import java.io.Serializable;
+
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+
+import org.apache.geode.cache.PartitionAttributes;
+import org.apache.geode.cache.PartitionAttributesFactory;
+import org.apache.geode.cache.Region;
+import org.apache.geode.cache.RegionShortcut;
+import org.apache.geode.cache.client.ClientRegionFactory;
+import org.apache.geode.cache.client.ClientRegionShortcut;
+import org.apache.geode.cache.client.PoolFactory;
+import org.apache.geode.cache.client.PoolManager;
+import org.apache.geode.cache.client.internal.PoolImpl;
+import org.apache.geode.cache.server.CacheServer;
+import org.apache.geode.distributed.internal.ServerLocation;
+import org.apache.geode.logging.internal.log4j.api.LogService;
+import org.apache.geode.test.dunit.VM;
+import org.apache.geode.test.dunit.rules.CacheRule;
+import org.apache.geode.test.dunit.rules.ClientCacheRule;
+import org.apache.geode.test.dunit.rules.DistributedRule;
+import org.apache.geode.test.junit.rules.serializable.SerializableTestName;
+
+@SuppressWarnings("serial")
+public class ClientServerTransactionDistributedTest implements Serializable {
+
+  private String hostName;
+  private String uniqueName;
+  private String regionName;
+  private VM server1;
+  private int port1;
+
+  @Rule
+  public DistributedRule distributedRule = new DistributedRule();
+
+  @Rule
+  public CacheRule cacheRule = new CacheRule();
+
+  @Rule
+  public ClientCacheRule clientCacheRule = new ClientCacheRule();
+
+  @Rule
+  public SerializableTestName testName = new SerializableTestName();
+
+  @Before
+  public void setUp() {
+    server1 = getVM(0);
+    hostName = getHostName();
+    uniqueName = getClass().getSimpleName() + "_" + testName.getMethodName();
+    regionName = uniqueName + "_region";
+  }
+
+  @Test
+  public void clientTransactionIDAboveIntegerMaxValue() {
+    port1 = server1.invoke(() -> createServerRegion());
+
+    // Test that transaction ID overflow to zero
+    TXManagerImpl.INITIAL_UNIQUE_ID_VALUE = Integer.MAX_VALUE;
+    createClientRegion(true, port1);
+
+    TXManagerImpl txManager =
+        (TXManagerImpl) clientCacheRule.getClientCache().getCacheTransactionManager();
+
+    txManager.begin();
+    TXStateProxyImpl txStateProxy = (TXStateProxyImpl) txManager.getTXState();
+    int transactionID = ((TXId) txStateProxy.getTransactionId()).getUniqId();
+
+    int numOfOperations = 5;
+    putData(numOfOperations);
+    txManager.commit();
+
+    server1.invoke(() -> verifyTransactionResult(numOfOperations));
+    assertEquals(0, transactionID);
+  }
+
+  private void putData(int numberOfEntries) {
+    Region region = clientCacheRule.getClientCache().getRegion(regionName);
+    for (int key = 0; key < numberOfEntries; key++) {
+      String value = getValue(key);
+      region.put(key, value);
+    }
+  }
+
+  private void verifyTransactionResult(int numberOfEntries) {
+    Region region = cacheRule.getCache().getRegion(regionName);
+    for (int i = 0; i < numberOfEntries; i++) {
+      LogService.getLogger().info("region get key {} value {} ", i, region.get(i));
+    }
+    for (int i = 0; i < numberOfEntries; i++) {
+      assertEquals("value" + i, region.get(i));
+    }
+  }
+
+  private int createServerRegion() throws Exception {
+    PartitionAttributesFactory factory = new PartitionAttributesFactory();
+    PartitionAttributes partitionAttributes = factory.create();
+    cacheRule.getOrCreateCache().createRegionFactory(RegionShortcut.PARTITION)
+        .setPartitionAttributes(partitionAttributes).create(regionName);
+    CacheServer server = cacheRule.getCache().addCacheServer();
+    server.setPort(0);
+    server.start();
+    return server.getPort();
+  }
+
+  private void createClientRegion(boolean connectToFirstPort, int... ports) {
+    clientCacheRule.createClientCache();
+
+    PoolImpl pool = getPool(ports);
+    ClientRegionFactory crf =
+        clientCacheRule.getClientCache().createClientRegionFactory(ClientRegionShortcut.LOCAL);
+    crf.setPoolName(pool.getName());
+    crf.create(regionName);
+
+    if (ports.length > 1 && connectToFirstPort) {
+      // first connection to the first port in the list
+      pool.acquireConnection(new ServerLocation(hostName, ports[0]));
+    }
+  }
+
+  private PoolImpl getPool(int... ports) {
+    PoolFactory factory = PoolManager.createFactory();
+    for (int port : ports) {
+      factory.addServer(hostName, port);
+    }
+    return (PoolImpl) factory.create(uniqueName);
+  }
+
+  private String getValue(int key) {
+    return "value" + key;
+  }
+}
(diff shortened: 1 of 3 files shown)


jvarenina:
Hi reviewers,

I'm not sure in this solution anymore. Now, I think that it would be better to change uniqID from AtomicInteger to AtomicLong. The Integer.MAX_VALUE is just to small, and because of that, uniqID would overflow to often with this solution. That could possibly cause conflict between new transactions and old transactions (that are suspended or hang) with same uniqID.

It would require much bigger impact on the code to use AtomicLong. Also, It would be necessary to preserve backward compatibility with the old client that would still use integer for the uniqID.

What do you think? Do you maybe see any possible issues with changing AtomicInteger to AtomicLong?

mkevo (approved):

geode-core/src/main/java/org/apache/geode/internal/cache/TXManagerImpl.java
```diff
@@ -353,7 +356,8 @@ public void begin() {
             "Current thread has paused its transaction so it can not start a new transaction");
       }
     }
-    TXId id = new TXId(this.distributionMgrId, this.uniqId.incrementAndGet());
+    TXId id = new TXId(this.distributionMgrId,
+        this.uniqId.updateAndGet(i -> i == Integer.MAX_VALUE ? 0 : i + 1));
```
pivotal-eshu, line 360: Please note 0 was not a valid TXId before -- due to the uniqId.incrementAndGet() method call. It would be better to wrap to 1 instead of 0.

pivotal-eshu (commented):
Please note that only when TXId overflows to -1, server treats it as a non-transactional. Sever would treat other negative txIDs as transactional though. The fix is still needed.

pivotal-eshu:
> Hi reviewers,
> 
> I'm not sure in this solution anymore. Now, I think that it would be better to change uniqID from AtomicInteger to AtomicLong. The Integer.MAX_VALUE is just to small, and because of that, uniqID would overflow to often with this solution. That could possibly cause conflict between new transactions and old transactions (that are suspended or hang) with same uniqID.
> 
> It would require much bigger impact on the code to use AtomicLong. Also, It would be necessary to preserve backward compatibility with the old client that would still use integer for the uniqID.
> 
> What do you think? Do you maybe see any possible issues with changing AtomicInteger to AtomicLong?

The integer is enough for GEODE use cases. Geode transaction is designed to be a small, short lived one, as it takes resources to perform a transaction (use TXState). I do not think a collision is likely scenario for geode user applications. The transaction should be committed if overflow ever occurs, otherwise it would be a user error. And geode does have default 3 minutes (if inactive) for client to work on a transaction before expiring the transaction.
this.transactionTimeToLive = Integer
    .getInteger(GeodeGlossary.GEMFIRE_PREFIX + "cacheServer.transactionTimeToLive", 180);
There is no need to change to AtomicLong.

geode-core/src/main/java/org/apache/geode/internal/cache/TXManagerImpl.java
```diff
@@ -375,7 +379,8 @@ public void begin() {
    */
   public TXStateProxy beginJTA() {
     checkClosed();
-    TXId id = new TXId(this.distributionMgrId, this.uniqId.incrementAndGet());
+    TXId id = new TXId(this.distributionMgrId,
+        this.uniqId.updateAndGet(i -> i == Integer.MAX_VALUE ? 0 : i + 1));
```
pivotal-eshu, line 383: It might be better to have a method getNewTXUniqueId() to generate the next TXId?

pivotal-eshu (changes requested):

geode-core/src/main/java/org/apache/geode/internal/cache/TXManagerImpl.java
jvarenina: Thanks for your comments. You are right, I have updated the code to wrap to 1.

geode-core/src/main/java/org/apache/geode/internal/cache/TXManagerImpl.java
jvarenina: I have created new method called getNewTXId(), since it returns TXId object. Hope this is fine? I can change it to getNewTXUniqueID() if needed.

jvarenina:
> > Hi reviewers,
> > I'm not sure in this solution anymore. Now, I think that it would be better to change uniqID from AtomicInteger to AtomicLong. The Integer.MAX_VALUE is just to small, and because of that, uniqID would overflow to often with this solution. That could possibly cause conflict between new transactions and old transactions (that are suspended or hang) with same uniqID.
> > It would require much bigger impact on the code to use AtomicLong. Also, It would be necessary to preserve backward compatibility with the old client that would still use integer for the uniqID.
> > What do you think? Do you maybe see any possible issues with changing AtomicInteger to AtomicLong?
> 
> The integer is enough for GEODE use cases. Geode transaction is designed to be a small, short lived one, as it takes resources to perform a transaction (use TXState). I do not think a collision is likely scenario for geode user applications. The transaction should be committed if overflow ever occurs, otherwise it would be a user error. And geode does have default 3 minutes (if inactive) for client to work on a transaction before expiring the transaction.
> this.transactionTimeToLive = Integer
> .getInteger(GeodeGlossary.GEMFIRE_PREFIX + "cacheServer.transactionTimeToLive", 180);
> There is no need to change to AtomicLong.

Your are completely right. Thank you on this clarification.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClientServerTransactionDistributedTest.java
```diff
@@ -0,0 +1,152 @@
...
+    TXManagerImpl txManager =
+        (TXManagerImpl) clientCacheRule.getClientCache().getCacheTransactionManager();
+
+    txManager.begin();
+    TXStateProxyImpl txStateProxy = (TXStateProxyImpl) txManager.getTXState();
+    int transactionID = ((TXId) txStateProxy.getTransactionId()).getUniqId();
+
+    int numOfOperations = 5;
+    putData(numOfOperations);
+    txManager.commit();
```
pivotal-eshu, line 91: To verify the transactional behavior, I think rollback will be a better testing case. To verify it is an indeed transaction. If operations are non transactional, rollback will have no effect -- data are still put into the cache.

pivotal-eshu (approved):
Looks good. Better to have a test case that can verify the fix -- it is still transactional after the overflow.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClientServerTransactionDistributedTest.java
```diff
@@ -0,0 +1,152 @@
...
+    PartitionAttributes partitionAttributes = factory.create();
+    cacheRule.getOrCreateCache().createRegionFactory(RegionShortcut.PARTITION)
+        .setPartitionAttributes(partitionAttributes).create(regionName);
+    CacheServer server = cacheRule.getCache().addCacheServer();
+    server.setPort(0);
+    server.start();
+    return server.getPort();
+  }
+
+  private void createClientRegion(boolean connectToFirstPort, int... ports) {
```
DonalEvans, line 126: This method is only used once, so `connectToFirstPort` is always true and `ports.length` is always equal to 1. Could it be simplified to remove unused code?

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClientServerTransactionDistributedTest.java
```diff
@@ -0,0 +1,152 @@
...
+    clientCacheRule.createClientCache();
+
+    PoolImpl pool = getPool(ports);
+    ClientRegionFactory crf =
```
DonalEvans, line 130: The compiler warning on this line can be fixed by using `ClientRegionFactory<Object, Object>`.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClientServerTransactionDistributedTest.java
```diff
@@ -0,0 +1,152 @@
...
+    PoolFactory factory = PoolManager.createFactory();
+    for (int port : ports) {
+      factory.addServer(hostName, port);
+    }
+    return (PoolImpl) factory.create(uniqueName);
+  }
+
+  private String getValue(int key) {
+    return "value" + key;
+  }
```
DonalEvans, line 151: This method can probably be inlined.

DonalEvans (changes requested):
A few code clean-up requests. Also, I echo what Eric said about testing that a transaction rollback is possible.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClientServerTransactionDistributedTest.java
```diff
@@ -0,0 +1,152 @@
...
+import org.apache.geode.test.junit.rules.serializable.SerializableTestName;
+
+@SuppressWarnings("serial")
+public class ClientServerTransactionDistributedTest implements Serializable {
+
+  private String hostName;
+  private String uniqueName;
+  private String regionName;
+  private VM server1;
+  private int port1;
```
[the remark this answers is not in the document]
jvarenina: Hi @DonalEvans , first thanks for your comments. Related to this comment, I haven't made the conversion, because I have created new TC for rollback where this variable is used. Hope this is ok?

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClientServerTransactionDistributedTest.java
jvarenina: You are right, I have created additional test for the rollback case.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClientServerTransactionDistributedTest.java
```diff
@@ -0,0 +1,152 @@
...
+  public void setUp() {
+    server1 = getVM(0);
+    hostName = getHostName();
+    uniqueName = getClass().getSimpleName() + "_" + testName.getMethodName();
+    regionName = uniqueName + "_region";
+  }
+
+  @Test
+  public void clientTransactionIDAboveIntegerMaxValue() {
+    port1 = server1.invoke(() -> createServerRegion());
```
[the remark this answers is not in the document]
jvarenina: done

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClientServerTransactionDistributedTest.java
```diff
@@ -0,0 +1,152 @@
...
+    int numOfOperations = 5;
+    putData(numOfOperations);
+    txManager.commit();
+
+    server1.invoke(() -> verifyTransactionResult(numOfOperations));
+    assertEquals(1, transactionID);
+  }
+
+  private void putData(int numberOfEntries) {
+    Region region = clientCacheRule.getClientCache().getRegion(regionName);
```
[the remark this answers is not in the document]
jvarenina: done

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClientServerTransactionDistributedTest.java
```diff
@@ -0,0 +1,152 @@
...
+    for (int key = 0; key < numberOfEntries; key++) {
+      String value = getValue(key);
+      region.put(key, value);
+    }
+  }
+
+  private void verifyTransactionResult(int numberOfEntries) {
+    Region region = cacheRule.getCache().getRegion(regionName);
```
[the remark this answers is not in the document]
jvarenina: done

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClientServerTransactionDistributedTest.java
```diff
@@ -0,0 +1,152 @@
...
+      LogService.getLogger().info("region get key {} value {} ", i, region.get(i));
+    }
+    for (int i = 0; i < numberOfEntries; i++) {
+      assertEquals("value" + i, region.get(i));
+    }
+  }
+
+  private int createServerRegion() throws Exception {
+    PartitionAttributesFactory factory = new PartitionAttributesFactory();
+    PartitionAttributes partitionAttributes = factory.create();
```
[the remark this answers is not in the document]
jvarenina: done

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClientServerTransactionDistributedTest.java
jvarenina: done

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClientServerTransactionDistributedTest.java
jvarenina: done

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClientServerTransactionDistributedTest.java
jvarenina: done

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClientServerTransactionDistributedTest.java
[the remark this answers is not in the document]
DonalEvans: Yeah, this is fine, thanks for addressing the feedback!

DonalEvans (approved):

PR #5934, opened 2021-01-20
GEODE-8824 fix some rule overlap and unintended singleton codeowners

onichols-pivotal:


upthewaterspout:
Since it looks like you are restricting the list of owners for CODEOWNERS and .asf.yml in this PR, should this have a dev list discussion?

onichols-pivotal:
> Since it looks like you are restricting the list of owners for CODEOWNERS and .asf.yml in this PR, should this have a dev list discussion?

That seems reasonable.  The question was previously asked on the dev list "who owns the CODEOWNERS file"  but never answered...

CODEOWNERS is currently ownerless since the syntax Robert tried didn't work.  I'd like to be tagged on any modification just to review that changes are syntactically sound, minimum of 2 owners is maintained, formatting is consistent, and all owners are committers.  I have no desire to interfere with any actual changes in ownership.  But I would also be fine with keeping CODEOWNERS ownerless if that is perceived as the only democratic solution.

upthewaterspout (approved):

PR #5935, opened 2021-01-20
GEODE-8772: Make tests assign ports

demery-pivotal:
Some tests in ManagementRequestLoggingDistributedTest and
RestAPIsAndInterOpsDUnitTest were using the default JMX port (1099).
Some tests in TombstoneDUnitTest were using the default cache server
port (40404). When these tests run in parallel outside of docker, they
can fail to bind these ports if another test has already bound them.

These tests now assign ports via AvailablePortHelper.

Authored-by: Dale Emery <demery@vmware.com>

### DIFF
--- a/geode-assembly/src/distributedTest/java/org/apache/geode/rest/internal/web/controllers/RestAPIsAndInterOpsDUnitTest.java
+++ b/geode-assembly/src/distributedTest/java/org/apache/geode/rest/internal/web/controllers/RestAPIsAndInterOpsDUnitTest.java
@@ -23,6 +23,8 @@ import static org.apache.geode.distributed.ConfigurationProperties.JMX_MANAGER_S
 import static org.apache.geode.distributed.ConfigurationProperties.LOCATORS;
 import static org.apache.geode.distributed.ConfigurationProperties.MCAST_PORT;
 import static org.apache.geode.distributed.ConfigurationProperties.START_DEV_REST_API;
+import static org.apache.geode.internal.AvailablePortHelper.getRandomAvailableTCPPort;
+import static org.apache.geode.internal.AvailablePortHelper.getRandomAvailableTCPPorts;
 import static org.assertj.core.api.Assertions.assertThat;
 
 import java.io.BufferedReader;
@@ -63,7 +65,6 @@ import org.apache.geode.cache.client.ClientCacheFactory;
 import org.apache.geode.cache.client.ClientRegionFactory;
 import org.apache.geode.cache.client.ClientRegionShortcut;
 import org.apache.geode.cache.server.CacheServer;
-import org.apache.geode.internal.AvailablePortHelper;
 import org.apache.geode.internal.cache.InternalCache;
 import org.apache.geode.internal.cache.xmlcache.RegionAttributesCreation;
 import org.apache.geode.pdx.PdxInstance;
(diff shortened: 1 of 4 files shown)


geode-assembly/src/distributedTest/java/org/apache/geode/rest/internal/web/controllers/RestAPIsAndInterOpsDUnitTest.java
```diff
@@ -168,15 +169,18 @@ private CacheServer createRegionAndStartCacheServer(String[] regions, Cache cach
...
+    int httpPort = ports[1];
+
     Properties props = new Properties();
     props.setProperty(MCAST_PORT, String.valueOf(0));
     props.setProperty(LOCATORS, locators);
 
     props.setProperty(JMX_MANAGER, "true");
     props.setProperty(JMX_MANAGER_START, "true");
-    props.setProperty(JMX_MANAGER_PORT, "0");
+    props.setProperty(JMX_MANAGER_PORT, String.valueOf(jmxManagerPort));
```
dschneider-pivotal, line 182: In this case you changed from "0" to using getRandomAvailableTCPPorts. I would have thought "0" is even better because using getRandomAvailableTCPPorts has a small race in which the port it hands back could end up being used by another before this thread can bind to it. I thought "0" was better because we are telling the operating system to give us an available port at the time we bind to it so we get rid of the race window. It can be hard to do in some tests when they need to know ahead of time what the actual port is. Do you know using "0" causes test issues?

dschneider-pivotal (approved):
These changes look good. The only one I had a question about was the one that used to use "0"

geode-assembly/src/distributedTest/java/org/apache/geode/rest/internal/web/controllers/RestAPIsAndInterOpsDUnitTest.java
demery-pivotal: That edit might be a mistake. My intention in this PR was to fix only uses of default ports.

Note: For this property (according to the docs) 0 means don't allow remote client connections. I'll have to look closer at what happens when 0 is specified here.

Other note: Using ephemeral ports can reduce the frequency of the race, but it doesn't eliminate it, because Geode's membership port range overlaps the ephemeral port range. So here's a troubling sequence:

1. Component A "reserves" a port in the membership port range.
2. Component B (perhaps running in a separate test, a separate member, or even a non-Geode process) binds an ephemeral port, and gets the "reserved" port.
3. Component A tries to bind the port it "reserved" and fails.

I'm not sure how often this troubling sequence happens, but it does happen. Requesting a non-membership port (via AvailablePortHelper) helps by assigning ports outside the ephemeral port range.

demery-pivotal: I think this change is unnecessary and I've reverted it. Thanks for catching that.

geode-assembly/src/distributedTest/java/org/apache/geode/rest/internal/web/controllers/RestAPIsAndInterOpsDUnitTest.java
```diff
@@ -168,6 +168,8 @@ private CacheServer createRegionAndStartCacheServer(String[] regions, Cache cach
   }
 
   private int startManager(final String locators, final String[] regions) throws IOException {
+    int httpPort = getRandomAvailableTCPPort();
```
mhansonp, line 171: this seems like it could be final...

mhansonp (commented):
Other than a few int's that could be final int's it looks good to me.

PR #5939, opened 2021-01-21
Geode-8853: DUNIT tests for HSTRLEN command

ringles:
Add DUNIT test for HSTRLEN Redis command, promote to "supported".

### DIFF
--- a/geode-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/hash/HstrlenDUnitTest.java
+++ b/geode-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/hash/HstrlenDUnitTest.java
@@ -1,0 +1,96 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package org.apache.geode.redis.internal.executor.hash;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.Random;
+
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Test;
+import redis.clients.jedis.Jedis;
+
+import org.apache.geode.redis.ConcurrentLoopingThreads;
+import org.apache.geode.test.awaitility.GeodeAwaitility;
+import org.apache.geode.test.dunit.rules.MemberVM;
+import org.apache.geode.test.dunit.rules.RedisClusterStartupRule;
+
+public class HstrlenDUnitTest {
+
+  @ClassRule
+  public static RedisClusterStartupRule clusterStartUp = new RedisClusterStartupRule(4);
+
+  private static final String LOCAL_HOST = "127.0.0.1";
+  private static final int JEDIS_TIMEOUT =
+      Math.toIntExact(GeodeAwaitility.getTimeout().toMillis());
+  private static Jedis jedis1;
+  private static Jedis jedis2;
+  private static Jedis jedis3;
+
+  @BeforeClass
+  public static void classSetup() {
+    MemberVM locator = clusterStartUp.startLocatorVM(0);
+    clusterStartUp.startRedisVM(1, locator.getPort());
+    clusterStartUp.startRedisVM(2, locator.getPort());
+
+    int redisServerPort1 = clusterStartUp.getRedisPort(1);
+    int redisServerPort2 = clusterStartUp.getRedisPort(2);
+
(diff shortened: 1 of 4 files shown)


sabbey37 (approved):
It seems good!  I always worry there are scenarios I'm forgetting about.

geode-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/hash/HstrlenDUnitTest.java
```diff
@@ -0,0 +1,95 @@
...
+          int newLength = rand.nextInt(9) + 1;
+          String newVal = makeStringOfRepeatedDigits(newLength);
+          jedis1.hset(key, field, newVal);
+        },
+        (i) -> assertThat(jedis2.hstrlen(key, field)).isBetween(1L, 9L),
+        (i) -> assertThat(jedis3.hstrlen(key, field)).isBetween(1L, 9L))
+            .run();
+
+    String value = jedis1.hget(key, field);
+    int expectedLength = Integer.parseInt(Character.toString(value.charAt(0)));
```
jhutchison, line 84: this line seemed pretty confusing to me when I looked at it.  Maybe there is some way to express that the value is comprised of the digit which is also supposed to be it's length?   Or maybe just a comment (in general I'm not a fan, but sometimes....)

geode-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/hash/HstrlenDUnitTest.java
```diff
@@ -0,0 +1,95 @@
...
+    jedis3 = new Jedis(LOCAL_HOST, redisServerPort1, JEDIS_TIMEOUT);
+  }
+
+  @Before
+  public void testSetup() {
+    jedis1.flushAll();
+  }
+
+  @Test
+  public void hstrlenWorks_whileAlsoUpdatingHash() {
```
jhutchison, line 65: this might be too finicky, but I was also confused about the intent of the test when I first read it-  maybe something like HSTrenLen will not fail throw error or corrupt data...  I guess I wasn't sure what aspect of it "working" that the test was intended to check when I read I initially read it

geode-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/hash/HstrlenDUnitTest.java
ringles: I'll pull the "Character.toString(value.charAt(0))" out into a variable with a descriptive name. That should do it...

geode-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/hash/HstrlenDUnitTest.java
ringles: Yes, a clearer name is probably called for here. Thanks!

geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/hash/AbstractHashesIntegrationTest.java
```diff
@@ -199,6 +199,16 @@ public void testHStrLen_failsForNonHashes() {
...
+  @Test
+  public void testHStrlen_givenWrongNumberOfArguments() {
+    assertThatThrownBy(() -> jedis.sendCommand(Protocol.Command.HSTRLEN))
+        .hasMessageContaining("wrong number of arguments");
+    assertThatThrownBy(() -> jedis.sendCommand(Protocol.Command.HSTRLEN, "1"))
+        .hasMessageContaining("wrong number of arguments");
+    assertThatThrownBy(() -> jedis.sendCommand(Protocol.Command.HSTRLEN, "1", "2", "3"))
+        .hasMessageContaining("wrong number of arguments");
+  }
+
```
sabbey37, line 211: Thanks for adding these! I made this comment in the `HSETNX` PR as well:
I know for most of the other commands (del, getset, exists, etc.) we check the error message more closely, like:
```
.hasMessageContaining("ERR wrong number of arguments for 'get' command");
```
I'm not sure if it's necessary, but maybe we could do the same here to be consistent? I realize there are a few areas where we're still just checking `wrong number of arguments`. We could eventually change those in a different PR.

sabbey37 (changes requested):

sabbey37 (approved):

geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/hash/AbstractHashesIntegrationTest.java
sabbey37: Did we want to change this? Or create a different PR for it? or not change it?

sabbey37 (changes requested):

geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/hash/AbstractHashesIntegrationTest.java
sabbey37: Did we want to change this? Or create a different PR for it? Or not change it? Or something else?

nonbinaryprogrammer: I second this. It would be pretty awkward to return an error message for the wrong command

nonbinaryprogrammer (approved):

geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/hash/AbstractHashesIntegrationTest.java
nonbinaryprogrammer: resolved by f65f4ce

PR #5940, opened 2021-01-21
GEODE-8858: unit/integration tests for HSETNX

ringles:
Add missing unit/integration tests for HSETNX. Confirm "WRONGTYPE" error when operating on non-hash keys.

### DIFF
--- a/geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/hash/AbstractHashesIntegrationTest.java
+++ b/geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/hash/AbstractHashesIntegrationTest.java
@@ -387,6 +387,18 @@ public abstract class AbstractHashesIntegrationTest implements RedisPortSupplier
     jedis.del(set_key);
   }
 
+  @Test
+  public void hsetnx_shouldThrowError_givenWrongNumberOfArguments() {
+    assertThatThrownBy(() -> jedis.sendCommand(Protocol.Command.HSETNX))
+        .hasMessageContaining("wrong number of arguments");
+    assertThatThrownBy(() -> jedis.sendCommand(Protocol.Command.HSETNX, "1"))
+        .hasMessageContaining("wrong number of arguments");
+    assertThatThrownBy(() -> jedis.sendCommand(Protocol.Command.HSETNX, "1", "2"))
+        .hasMessageContaining("wrong number of arguments");
+    assertThatThrownBy(() -> jedis.sendCommand(Protocol.Command.HSETNX, "1", "2", "3", "4"))
+        .hasMessageContaining("wrong number of arguments");
+  }
+
   /**
    * Test the HVALS command
    */
--- a/geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/hash/AbstractHashesIntegrationTest.java
+++ b/geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/hash/AbstractHashesIntegrationTest.java
@@ -376,11 +376,12 @@ public abstract class AbstractHashesIntegrationTest implements RedisPortSupplier
     jedis.sadd(set_key, field);
 
     assertThatThrownBy(
-        () -> jedis.hsetnx(string_key, field, "something else")).isInstanceOf(JedisDataException.class)
-        .hasMessageContaining("WRONGTYPE");
+        () -> jedis.hsetnx(string_key, field, "something else"))
+            .isInstanceOf(JedisDataException.class)
+            .hasMessageContaining("WRONGTYPE");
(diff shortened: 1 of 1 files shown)


geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/hash/AbstractHashesIntegrationTest.java
```diff
@@ -365,6 +365,28 @@ public void testHSetNXExecutor() {
...
+    String field = "field";
+    String value = "value";
+
+    jedis.set(string_key, value);
+    jedis.sadd(set_key, field);
+
+    assertThatThrownBy(
+        () -> jedis.hsetnx(string_key, field, "something else"))
+            .isInstanceOf(JedisDataException.class)
+            .hasMessageContaining("WRONGTYPE");
```
sabbey37, line 381: It would be nice if we checked the full error message here.  We store the rest of it as a constant in `RedisConstants`, so it could be something like:
```
.hasMessageContaining("WRONGTYPE " + ERROR_WRONG_TYPE);
```

if you want to be even more exact, it could be:
```
.hasMessage("(error) WRONGTYPE " + ERROR_WRONG_TYPE);
```

I know it might seem like overkill, but I remember us having problems with other error messages not matching Redis's exactly.  There are other tests where we just check that the message contains `WRONGTYPE`, maybe we could change those as well.

sabbey37 (changes requested):
It would also be good to add tests for too few and too many arguments (for example if I do:
```
SADD orange red yellow
HSETNX orange red
```
We'd want to make sure we get the wrong number of arguments error vs. the WRONGTYPE error. Maybe it is overkill, but I think it would be useful to make sure our behavior is the same as Redis, and in case someone ever changed our implementation or refactored, there would be tests verifying this behavior is the same.

geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/hash/AbstractHashesIntegrationTest.java
ringles: Good call on the argument count check, I did that for the HSTRLEN PR that's currently pending, too!

Think I might add a story for the WRONGTYPE messages, since that will touch a bunch of otherwise-unrelated commands.

geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/hash/AbstractHashesIntegrationTest.java
```diff
@@ -365,6 +365,40 @@ public void testHSetNXExecutor() {
...
+  public void hsetnx_shouldThrowError_givenWrongNumberOfArguments() {
+    assertThatThrownBy(() -> jedis.sendCommand(Protocol.Command.HSETNX))
+        .hasMessageContaining("wrong number of arguments");
+    assertThatThrownBy(() -> jedis.sendCommand(Protocol.Command.HSETNX, "1"))
+        .hasMessageContaining("wrong number of arguments");
+    assertThatThrownBy(() -> jedis.sendCommand(Protocol.Command.HSETNX, "1", "2"))
+        .hasMessageContaining("wrong number of arguments");
+    assertThatThrownBy(() -> jedis.sendCommand(Protocol.Command.HSETNX, "1", "2", "3", "4"))
+        .hasMessageContaining("wrong number of arguments");
+  }
```
sabbey37, line 400: This is great!  Thanks for adding these.  I know for most of the other commands (del, getset, exists, etc.) we check the error message more closely, like:

```
.hasMessageContaining("ERR wrong number of arguments for 'get' command");
```
I'm not sure if it's necessary, but maybe we could do the same here to be consistent?  I realize there are a few areas where we're still just checking `wrong number of arguments`.  We could eventually change those in a different PR.

sabbey37 (changes requested):

geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/hash/AbstractHashesIntegrationTest.java
jdeppe-pivotal: I've already made these changes in another PR.

jdeppe-pivotal (approved):

sabbey37 (approved):

geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/hash/AbstractHashesIntegrationTest.java
sabbey37: @jdeppe-pivotal says he'll fix it in another PR.

PR #5948, opened 2021-01-24
GEODE-8872: added client impacts

mivanac:
Description of the use case:
When deploying geode, for locator we define hostname-for-clients. This is external address used for WAN, and external clients.
If we also deploy some clients in internal network (for example in the same kubernetes cluster as geode), then for those clients to connect with locator, communication will go from internal network to external, then from external to internal. This increases latency, and we should communicate over the shortest path.
(Description added by mhansonp)


bschuchardt:
Mario, this is marked as a Draft PR.  Do you want early reviews before taking it out of Draft status?

mivanac:
Comments are appreciated in any PR status.

geode-core/src/main/java/org/apache/geode/distributed/internal/ServerLocator.java
```diff
@@ -378,24 +379,40 @@ public void endResponse(Object request, long startTime) {
...
+    String host;
+    if (requestInternal) {
+      host = p.getInternalHost();
+    } else {
+      host = p.getHost();
+    }
+    return new ServerLocation(host, p.getPort());
+  }
+
   protected static ServerLocation buildServerLocation(GridProfile p) {
```
DonalEvans, line 416: Could uses of this method be replaced with calls to the new, two-argument method instead, to reduce redundancy?

kirklund (commented):
The difference between `host` and `internalHost` gets really muddy. I have a decent understanding of what your intent is... The user can specify connecting to the Locator from two different networks (one network that is external to the cluster and one within the clustering network, maybe).

You're designing an API that matches the implementation very closely. Would any user ever want 3 different networks to choose from? What if the Locator machine has 3 NICs and we want internal, external, other? What if we want to allow some clients to connect to internal, others to external, or block clients entirely from one of them?

The developers who added "HostNameForClients" concept originally meant for that to be the address that clients connect to, whether that be internal or external to the cluster network. Now this adds in the concept of "InternalHost". The messiness is most obvious in the `ServerLocator` -- there's the old field `hostNameForClients` and the new field 'hostName` which is also known as `internalHost` in other classes such as `ControllerProfile`. When you go from the concept of a single host address for clients connecting to the locator to having support for two or multiple addresses, that's when you need to have a feel for the direction of the API itself and make sure we don't lock it into something that will be specific to one user or difficult to alter in the future due to API backwards compatibility.

You're adding in something that is basically what the original developers intended for `hostNameForClients` to provide. So, if you have an `internalHost` name in a deployment, what does `hostNameForClients` mean? And if we're going to support multiple networking interfaces in `ServerLocator` why limit it to two instead of multiple? And what about the similar configuration props in the system like `jmx-manager-hostname-for-clients` and `jmx-manager-bind-address`.

Now I'm not suggesting over-designing something to support multiple locator address spaces for ServerLocator but it feels like "InternalHost" is confused with "HostNameForClients" which then becomes more obvious when looking at the implementation of that `fillInProfile` method. But any new APIs require a lot of thought about future direction as well as consistency with the rest of the APIs and configuration. I'm not claiming everything is currently consistent but we need to continually try to move it in that direction, and the similar config props also worry me.

bschuchardt (commented):
I'm not sure why we need this.  The JIRA doesn't have much information and there's no description in this PR.
Do we really want to expose internal host information to clients?  The hostNameForClients setting tells clients how they should access servers and it seems to me that we should stick with that and not expose server-side information like this.

mivanac:
When deploying geode, for locator we define hostname-for-clients. This is extern address used for WAN, and extern clients.
If we also deploy some clients in internal network (for example in the same kubernetes cluster as geode), then for those clients to connect with locator, communication will go from internal network to external, then from external to internal. This increases latency, and we should communicate over the shortest path.

dschneider-pivotal (approved):
The only file I own in this PR is ConnectionPoolFactoryJUnitTest.java and the changes to it LGTM.

For the bigger picture I have the same concerns as Kirk and Bruce and would like to see those addressed before this PR is merged in.

geode-core/src/main/java/org/apache/geode/cache/client/PoolFactory.java
```diff
@@ -220,6 +220,14 @@
    * @since GemFire 6.5
    */
   boolean DEFAULT_MULTIUSER_AUTHENTICATION = false;
+  /**
+   * The default value for whether to request locator internal address.
```
DonalEvans, line 224: Slightly more detail would be good here, explaining when exactly the locator internal address would be requested if this was set to true.

geode-core/src/main/java/org/apache/geode/internal/cache/GridAdvisor.java
```diff
@@ -257,6 +257,9 @@ public Set adviseProfileRemove() {
 
     private String host;
 
+    /** valid only for ControllerProfile */
+    private String internalHost;
```
DonalEvans, line 261: It doesn't seem correct to have a field in a parent class that only belongs to one of its child classes. Can this be moved into ControllerProfile?

geode-core/src/main/java/org/apache/geode/internal/cache/GridAdvisor.java
```diff
@@ -432,6 +447,17 @@ public boolean equals(Object obj) {
         if (this.gp.getPort() == other.gp.getPort()) {
           final String thisHost = this.gp.getHost();
           final String otherHost = other.gp.getHost();
+          final String thisInternalHost = this.gp.getInternalHost();
```
DonalEvans, line 450: If internalHost is being included in `equals()` calculations, it should probably also be included in hashCode()`.

geode-dunit/src/main/java/org/apache/geode/cache/client/internal/LocatorTestBase.java
```diff
@@ -141,6 +141,34 @@ protected int startLocator(final String hostName, final String otherLocators) th
     return locator.getPort();
   }
 
+  protected int startLocator(final String hostName, final String otherLocators,
```
DonalEvans, line 144: This method is never used.

DonalEvans (changes requested):
I share Bruce and Kirk's concerns about the overall design/motivation of these changes. Also, there are no tests associated with the new code. Unit tests should be written to exercise the added/modified code, DUnit tests should be written to confirm that the functionality is as intended, and where appropriate, backwards compatibility tests should be added to ensure that no isses have been introduced by these changes.

geode-core/src/main/java/org/apache/geode/cache/client/Pool.java
```diff
@@ -207,6 +207,13 @@ default boolean getThreadLocalConnections() {
    */
   boolean getMultiuserAuthentication();
 
+  /**
+   * Returns true ir request locator internal address is enabled on this pool.
+   *
+   * @see PoolFactory#setRequestLocatorInternalAddress(boolean)
+   * @since Geode 1.14
+   */
+  boolean getRequestLocatorInternalAddress();
```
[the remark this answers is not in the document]
mivanac: Thanks for comment. I will update.

mivanac:
Updated code. Working on additional test.

geode-core/src/test/java/org/apache/geode/cache/client/internal/PoolImplTest.java
```diff
@@ -137,6 +158,8 @@ private PoolImpl getPool(int retryAttemptsAttribute) {
     doReturn(1).when(poolAttributes).getMaxConnections();
     doReturn((long) 10e8).when(poolAttributes).getPingInterval();
     doReturn(retryAttemptsAttribute).when(poolAttributes).getRetryAttempts();
+    doReturn(requestLocatorInternalAddressEnabled).when(poolAttributes)
```
DonalEvans, line 161: The comment above this change states that these mocks are required for pool validation and setup to complete and that "The values of these attributes have no importance to the assertions of the test itself." In the case of this line, that's not true though, as the value of `requestLocatorInternalAddressEnabled` returned is explicitly what's being tested in the added test cases, and the pool can still be succesfully created with this line commented out. It might be better to change the signature of the `getPool()` method to take a `PoolAttributes` argument which can have the required additional mocking added in the individual test cases that require it, and leave `getPool()` to only do the mocking that's necessary to create a pool. I believe the same is true of the `retryAttemptsAttribute` argument, so that could also be removed from `getPool()` and the necessary mocking done only in the test methods that require it.

DonalEvans (changes requested):

DonalEvans (commented):
Changes look good from a code quality point of view, but I'm hesitant to approve the PR until Kirk and Bruce feel that their questions/concerns have been answered, as they know this stuff better than me.

mivanac:
If we deploy clients in internal network (for example in the same kubernetes cluster as geode), then for those clients to connect with locator, communication will go from internal network to external, then from external to internal. This increases latency, and we should communicate over the shortest path. This is the reason that for some clients we need internal address exposed.

mhansonp (approved):

mhansonp:
Hi, I would really appreciate it if you made the effort to clear our the warnings that are in all of the tests you touch. That will leave us cleaner code and a more clear understanding of our real problems (e.g. deprecated code). It is not required for this PR, but it should be something we try to do. Thanks!

mivanac:
I have explained use case for this PR. If community is not interested in such change, I will close this PR.

DonalEvans:
Tagging @kirklund for a re-review or to continue the discussion, since his concerns may have been addressed at this point.

Bill (changes requested):
After reading the ticket and @kirklund's review, I agree with @kirklund and that's why I'm requesting changes.

FWIW, I didn't think about it for a really long time, but it seems to me that if we want clients in separate networks to access servers efficiently (directly) perhaps we should think about a new stand-alone server-location implementation that can sit in a particular network and act as a gateway.

In that case we wouldn't specify anything like hostname-for-clients on a cache server—rather, we'd configure the particular gateway. And if we had such a (new) component (`ClientGatewayLauncher` anybody??) a user could choose to deploy none of them or one of them or even more.

There are challenges with my first idea. So here's a second: we keep our current topology but instead of adding a second hostname-for-clients configuration parameter to the cache server (as the PR does), we add the ability to configure some pattern-based mapping configuration in the server-location service. I'm thinking of a configuration that can provide server location mapping based on the IP address of the client.

agingade:
@mivanac 
>> I have explained use case for this PR. If community is not interested in such change, I will close this PR.
Its not community is interested or not; it is the nature of the problem that could raise many questions/concerns and its effectiveness that could bring different views and could take time to resolve (for good). If you can see many community members are involved here trying to get to an optimal conclusion.
Looking at the problem and the solution proposed; this problem seems to be a good candidate to raise an RFC, where discussion could have been carried out to understand the pros and cons of the issue and solutions.
From the problem description it seems you are trying to address performance issue with client connection. It would have been nice to see, what kind of performance impact we are looking at (e.g. with prototype)
Because of the sensitive nature of the issue (use/exposure of a n/w info) it will be nice to talked about any security issues that this could bring about.
As I mentioned above; some of the areas that could have been discussed in RFC are:
- Benefits (with some sample perf number)
- Security concerns/issues with this
- Case where an external client tries to fetch the internal info (outside of k8 env - what happens in this case) 
- How other client connection could be impacted or can use similar in future (Java, C++, JMX, Gfsh, pulse connections)
- Requirement for Backward compatibility,  rolling upgrade tests (I am not seeing any here)

mivanac:
Hi,
thanks for all comments.

From received comments, I see more concerns regarding exposing internal address and security, so I will close current PR,
and new proposal describe in RFC,

Regards,
Mario

mhansonp:
> Hi,
> thanks for all comments.
> 
> From received comments, I see more concerns regarding exposing internal address and security, so I will close current PR,
> and new proposal describe in RFC,
> 
> Regards,
> Mario

Thanks Mario!

mivanac:
RFC created for this improvement https://cwiki.apache.org/confluence/display/GEODE/Get+locators+internal+host+addresses+for+clients+in+same+subnet

PR #5949, opened 2021-01-25
GEODE-8857: Redis SMOVE command should not do type checking in the executor

jdeppe-pivotal:
- This fixes an issue with hits/misses stats.

### For all changes:

### Note:

### DIFF
--- a/geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/set/AbstractSMoveIntegrationTest.java
+++ b/geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/set/AbstractSMoveIntegrationTest.java
@@ -17,13 +17,11 @@ package org.apache.geode.redis.internal.executor.set;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
+import java.util.ArrayList;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Set;
-import java.util.concurrent.Callable;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.Future;
+import java.util.concurrent.atomic.AtomicLong;
 
 import org.junit.After;
 import org.junit.Before;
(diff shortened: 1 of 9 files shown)


geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/set/AbstractSMoveIntegrationTest.java
```diff
@@ -80,6 +81,17 @@ public void givenMoreThanFourArguments_returnsWrongNumberOfArgumentsError() {
...
+  @Test
+  public void testSmove_returnsWrongType_whenWrongTypeIsUsed() {
+    jedis.set("not-a-set", "value");
+    assertThatThrownBy(() -> jedis.smove("not-a-set", "some-set", "foo"))
+        .hasMessage("WRONGTYPE " + RedisConstants.ERROR_WRONG_TYPE);
+
+    jedis.hset("not-another-set", "field", "value");
+    assertThatThrownBy(() -> jedis.smove("not-another-set", "some-set", "foo"))
+        .hasMessage("WRONGTYPE " + RedisConstants.ERROR_WRONG_TYPE);
+  }
```
sabbey37, line 93: Looks like the Redis tests are failing when a destination key of the wrong type is specified (vs. the source key).  Could we write tests for that as well?

sabbey37 (changes requested):

geode-redis/src/main/java/org/apache/geode/redis/internal/RedisCommandType.java
```diff
@@ -181,6 +181,7 @@
    ***************************************/
   // do not call these directly, only to be used in other commands
   INTERNALPTTL(null, INTERNAL, new ExactParameterRequirements(2)),
+  INTERNALTYPE(null, INTERNAL, new ExactParameterRequirements(2)),
   INTERNALSMEMBERS(null, INTERNAL, new ExactParameterRequirements(3)),
```
sabbey37, line 185: Is it weird that a user would be able to call these from the CLI and would get different error messages if they did (like "wrong number of arguments" instead of "unknown command" error?  Not sure why they would and maybe it doesn't really matter.  I guess I'm just wondering if there's a way of preventing that or some other way of doing this.

geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/set/AbstractSMoveIntegrationTest.java
jdeppe-pivotal: Yes, I've already fixed that and added a test :)

sabbey37 (approved):

geode-redis/src/main/java/org/apache/geode/redis/internal/RedisCommandType.java
jdeppe-pivotal: I have a story to address this. Also, would like to hear your thoughts on how better to address the problem we're trying to fix with these 'internal' commands.

PR #5950, opened 2021-01-25
GEODE-8878: PR clear should also send a lock message to the secondary members.

jinmeiliao:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/distributedTest/java/org/apache/geode/cache/query/partitioned/PRClearCreateIndexDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/cache/query/partitioned/PRClearCreateIndexDUnitTest.java
@@ -1,0 +1,264 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package org.apache.geode.cache.query.partitioned;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.io.Serializable;
+import java.util.stream.IntStream;
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+
+import org.apache.geode.cache.Region;
+import org.apache.geode.cache.client.ClientCache;
+import org.apache.geode.cache.client.ClientRegionShortcut;
+import org.apache.geode.cache.query.QueryService;
+import org.apache.geode.distributed.internal.ClusterDistributionManager;
+import org.apache.geode.distributed.internal.DistributionMessage;
+import org.apache.geode.distributed.internal.DistributionMessageObserver;
+import org.apache.geode.internal.cache.DistributedClearOperation;
+import org.apache.geode.internal.cache.DistributedClearOperation.ClearRegionMessage;
+import org.apache.geode.internal.cache.PartitionedRegionClearMessage;
+import org.apache.geode.test.dunit.AsyncInvocation;
+import org.apache.geode.test.dunit.rules.ClientVM;
+import org.apache.geode.test.dunit.rules.ClusterStartupRule;
+import org.apache.geode.test.dunit.rules.MemberVM;
+
+public class PRClearCreateIndexDUnitTest implements Serializable {
+  @Rule
+  public ClusterStartupRule cluster = new ClusterStartupRule(4, true);
+
+  private MemberVM primary, secondary;
+  private ClientVM client;
+
+  @Before
+  public void before() throws Exception {
+    int locatorPort = ClusterStartupRule.getDUnitLocatorPort();
+    primary = cluster.startServerVM(0, locatorPort);
+    secondary = cluster.startServerVM(1, locatorPort);
+
+    // create region on server1 first, making sure server1 has the primary bucket
+    primary.invoke(() -> {
+      DistributionMessageObserver.setInstance(new MessageObserver());
+      Region<Object, Object> region =
+          ClusterStartupRule.memberStarter.createPartitionRegion("regionA",
+              f -> f.setTotalNumBuckets(1).setRedundantCopies(2));
+      IntStream.range(0, 100).forEach(i -> region.put(i, "value" + i));
+    });
+
+    // server2 has the secondary bucket
+    secondary.invoke(() -> {
+      DistributionMessageObserver.setInstance(new MessageObserver());
+      ClusterStartupRule.memberStarter.createPartitionRegion("regionA",
+          f -> f.setTotalNumBuckets(1).setRedundantCopies(2));
+    });
+  }
+
+  @After
+  public void after() throws Exception {
+    primary.invoke(() -> {
+      DistributionMessageObserver.setInstance(null);
+    });
+    secondary.invoke(() -> {
+      DistributionMessageObserver.setInstance(null);
+    });
+  }
+
+  // All tests create index on secondary members. These tests are making sure we are requesting
+  // locks for clear on secondary members as well. If we create index on the primary, the clear
+  // and createIndex will run sequentially so there would be no error. But if we create index on
+  // the secondary member and if the secondary member will not
+  // request a lock for clear operation, it will result in an EntryDestroyedException when create
+  // index is happening.
+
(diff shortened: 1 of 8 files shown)


geode-core/src/distributedTest/java/org/apache/geode/cache/query/partitioned/PRClearCreateIndexDUnitTest.java
```diff
@@ -0,0 +1,264 @@
...
+   * 3. then since it already locked the current member, won't send a OP_LOCK_FOR_CLEAR message
+   * to secondaries, only OP_CLEAR will be sent
+   */
+  public void clearFromInterestedClientConnectingToPrimaryMember() throws Exception {
+    int port = primary.getPort();
+    client = cluster.startClientVM(2, c -> c.withServerConnection(port).withPoolSubscription(true));
+    AsyncInvocation createIndex = secondary.invokeAsync(PRClearCreateIndexDUnitTest::createIndex);
+
+    AsyncInvocation clear = client.invokeAsync(() -> {
+      Thread.sleep(200);
```
kirklund, line 167: Why are the Thread sleeps needed in this test?

jinmeiliao: to reduce the possibility that the clear would start before the create index. We want clear operation falls inside one of the "create index" operations.

geode-core/src/main/java/org/apache/geode/internal/cache/DistributedClearOperation.java
```diff
@@ -211,7 +215,7 @@ protected boolean operateOnRegion(CacheEvent event, ClusterDistributionManager d
       switch (this.clearOp) {
         case OP_CLEAR:
           region.clearRegionLocally((RegionEventImpl) event, false, this.rvv);
```
gesterzhou, line 217: why you did not change this line too?

jinmeiliao: It's a change suggested by IntelliJ, I don't know why it didn't suggest this line. I will change it

geode-core/src/distributedTest/java/org/apache/geode/cache/query/partitioned/PRClearCreateIndexDUnitTest.java
```diff
@@ -0,0 +1,264 @@
...
+    int locatorPort = ClusterStartupRule.getDUnitLocatorPort();
+    primary = cluster.startServerVM(0, locatorPort);
+    secondary = cluster.startServerVM(1, locatorPort);
+
+    // create region on server1 first, making sure server1 has the primary bucket
+    primary.invoke(() -> {
+      DistributionMessageObserver.setInstance(new MessageObserver());
+      Region<Object, Object> region =
+          ClusterStartupRule.memberStarter.createPartitionRegion("regionA",
+              f -> f.setTotalNumBuckets(1).setRedundantCopies(2));
```
gesterzhou, line 61: redundantCopy should be 1 here

geode-core/src/distributedTest/java/org/apache/geode/cache/query/partitioned/PRClearCreateIndexDUnitTest.java
```diff
@@ -0,0 +1,264 @@
...
+   * 4. primary will send a OP_CLEAR message back to the secondary to clear
+   */
+  @Test
+  public void clearFromInterestedClientConnectingToSecondaryMember() throws Exception {
+    int port = secondary.getPort();
+    client = cluster.startClientVM(2, c -> c.withServerConnection(port).withPoolSubscription(true));
+    AsyncInvocation createIndex = secondary.invokeAsync(PRClearCreateIndexDUnitTest::createIndex);
+
+    AsyncInvocation clear = client.invokeAsync(() -> {
+      Thread.sleep(200);
```
gesterzhou, line 138: will that cause race condition in test?

geode-core/src/distributedTest/java/org/apache/geode/cache/query/partitioned/PRClearCreateIndexDUnitTest.java
jinmeiliao: is it wrong to use 2? I want to have secondary members to also store data.

geode-core/src/distributedTest/java/org/apache/geode/cache/query/partitioned/PRClearCreateIndexDUnitTest.java
jinmeiliao: Yes, that's the purpose. We want to cause race condition in test, so that we can reproduce the exception before the fix.

geode-core/src/distributedTest/java/org/apache/geode/cache/query/partitioned/PRClearCreateIndexDUnitTest.java
gesterzhou: 1 means 1 secondary. You only have 2 servers

geode-core/src/distributedTest/java/org/apache/geode/cache/query/partitioned/PRClearCreateIndexDUnitTest.java
dschneider-pivotal: But it does not look like the sleep will ensure that the clear happens during the createIndex. It looks like it could happen before, during, or after. One of these three might fail but we can't be sure if we have run the test will all three.
Once it does happen it might fail and then we have a flaky test.
Would it be possible to add some internal observer to the product that it will call at the right moment during createIndex? Then you could have that observer count down a latch that the clear is waiting on.

jinmeiliao: Yes, it's not to ensure the clear will happen during the createIndex, it's just to increase the possibility. The test's was originally created to reproduce the issue fairly reliably, and it has served that purpose. Me and Kirk tried to find a way to make it a definite failure, but we found out it's not gonna be easy. So now the test is actually not relying on the race condition to happen (even though it will happen in most cases). It's actually asserting on the fact that the member received the right messages.

geode-core/src/distributedTest/java/org/apache/geode/cache/query/partitioned/PRClearCreateIndexDUnitTest.java
jinmeiliao: ok, I will make the change

geode-core/src/main/java/org/apache/geode/internal/cache/DistributedClearOperation.java
jinmeiliao: Oh, this cast is necessary, the event is of type `CacheEvent` when passed in.

dschneider-pivotal (approved):

geode-core/src/distributedTest/java/org/apache/geode/cache/query/partitioned/PRClearCreateIndexDUnitTest.java
kirklund: My review would be to remove the sleep and figure out a different way to test it even if it's at the unit test level instead of in a distributed or integration test.

kirklund (changes requested):
As tough as it is, we really need to avoid that Thread.sleep. That's just another new flaky test waiting to run at the wrong time on a slow machine.

geode-core/src/distributedTest/java/org/apache/geode/cache/query/partitioned/PRClearCreateIndexDUnitTest.java
jinmeiliao: I can remove the sleep. The test, after serving its original purpose of reproducing the bug, isn't relying on the race to happen anymore. The tests are making sure each member getting the right combination of messages.

gesterzhou (approved):

jinmeiliao:
> As tough as it is, we really need to avoid that Thread.sleep. That's just another new flaky test waiting to run at the wrong time on a slow machine.

@kirklund I've removed the sleep. The test isn't relying on the race to happen. The tests are making sure each member getting the right combination of messages. Please re-review.

geode-core/src/distributedTest/java/org/apache/geode/cache/query/partitioned/PRClearCreateIndexDUnitTest.java
```diff
@@ -0,0 +1,265 @@
...
+    assertThat(observer.isLock_secondary())
+        .describedAs("OP_LOCK_FOR_CLEAR received: %s", observer.isLock_secondary())
+        .isEqualTo(lockSecondary);
+    assertThat(observer.isClear_secondary())
+        .describedAs("OP_CLEAR received: %s", observer.isClear_secondary())
+        .isEqualTo(clearSecondary);
+  }
+
+  private static class MessageObserver extends DistributionMessageObserver {
+    private boolean lock_secondary = false;
```
kirklund, line 219: The thread invoking the `public boolean isXxxxx` methods will generally be the JUnit Main thread. The thread invoking `beforeProcessMessage` will be an internal Geode thread, so you should either make these booleans `volatile` or change the class to use `AtomicBooleans`.

It'll probably pass GREEN most of the time as is but this could end up being a source of flakiness.

kirklund (approved):
Approved, but please make the concurrency changes I marked in one inner-class.

PR #5953, opened 2021-01-26
GEODE-8811: Fix new value in events for failing client

kirklund:


geode-core/src/main/java/org/apache/geode/internal/cache/InternalRegion.java
```diff
@@ -449,6 +450,8 @@ default void unlockWhenRegionIsInitializing() {
 
   CachePerfStats getRegionPerfStats();
 
+  void handleInterestEvent(InterestRegistrationEvent event);
```
jchen21, line 453: Do you want to make it `public`?

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientRegistrationEventQueueManager.java
```diff
@@ -37,15 +39,25 @@
...
+      final ClientUpdateMessageImpl clientMessage,
       final Conflatable conflatable,
       final Set<ClientProxyMembershipID> originalFilterClientIDs,
       final CacheClientNotifier cacheClientNotifier) {
-    if (registeringProxyEventQueues.isEmpty())
+    if (registeringProxyEventQueues.isEmpty()) {
       return;
+    }
+
+    if (originalFilterClientIDs.isEmpty()) {
```
jchen21, line 55: How about combine all the conditions into a single `if` statement, instead of two?

dschneider-pivotal (approved):

geode-core/src/main/java/org/apache/geode/internal/cache/InternalRegion.java
kirklund: Yes, I think it's ok to make this public. Users can already use Reflection to invoke non-public methods on LocalRegion anyway. My overall method is this:

1. switch over to using interfaces for all dependencies and interactions between classes
2. use dependency injection to make unit testing easy
3. review the interfaces and break up larger ones to better represent the actual dependencies/interactions between packages or components

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientRegistrationEventQueueManager.java
kirklund: Fixed!

jchen21 (approved):

PR #5957, opened 2021-01-26
GEODE-8875: Removing packages element from log4j2 configuration file

upthewaterspout:
This element causes log4j to scan all geode jars on startup. This slows down
the startup time of all geode processes, including gfsh.

It is not necessary because log4j has an annotation processor to generate a
config file that helps it find plugs that it needs.


jdeppe-pivotal (approved):
Awesome! I see a 6x speedup in test starts now.

karensmolermiller (approved):
The docs change looks good to me. I haven't looked at the code/test changes.

bschuchardt (approved):
If the premise is correct I don't have a problem with the changes to the membership test or log4j configuration.

rhoughton-pivot (approved):

kirklund (approved):

mhansonp (approved):

PR #5947, opened 2021-01-22
 GEODE-8870: Removes GFE_5* and GFE_6*

pivotal-jbarrett:
Thank you for submitting a contribution to Apache Geode.

In order to streamline the review of the contribution we ask you
to ensure the following steps have been taken:

### For all changes:
- [ ] Is there a JIRA ticket associated with this PR? Is it referenced in the commit message?

- [ ] Has your PR been rebased against the latest commit within the target branch (typically `develop`)?

- [ ] Is your initial contribution a single, squashed commit?

- [ ] Does `gradlew build` run cleanly?

- [ ] Have you written or updated unit tests to verify your changes?

- [ ] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)?

### Note:
Please ensure that once the PR is submitted, check Concourse for build issues and
submit an update to your PR as soon as possible. If you need help, please send an
email to dev@geode.apache.org.

PR #5967, opened 2021-01-28
GEODE-8882: Enable JetBrains NotNull Annotation

Bill:
[GEODE-8882](https://issues.apache.org/jira/browse/GEODE-8882)

This new annotation support is documented in the Geode Wiki: https://cwiki.apache.org/confluence/display/GEODE/@NotNull+Annotation

Oh and woopsie! We've already found our first bug via the `@NotNull` annotation:

image

Under "more actions" I pick:

image

Where I find this little gem passing a `null` for `clientVersion`:

image

Which will blow up with the recently-merged-to-`develop` changes from PR #5947 (which removed the null-check in the method of interest.)

Incidentally that call location is `ClientRegistrationMetadata.getAndValidateClientVersion()` where we were unable to ascertain the client's version! In that case our latest code on `develop` will generate an NPE on the server.

### DIFF
--- a/boms/geode-all-bom/src/test/resources/expected-pom.xml
+++ b/boms/geode-all-bom/src/test/resources/expected-pom.xml
@@ -878,6 +878,12 @@
         <scope>compile</scope>
       </dependency>
       <dependency>
+        <groupId>org.jetbrains</groupId>
+        <artifactId>annotations</artifactId>
+        <version>20.1.0</version>
+        <scope>compile</scope>
+      </dependency>
+      <dependency>
         <groupId>org.apache.geode</groupId>
         <artifactId>geode-common</artifactId>
         <version>${version}</version>
--- a/buildSrc/src/main/groovy/org/apache/geode/gradle/plugins/DependencyConstraints.groovy
+++ b/buildSrc/src/main/groovy/org/apache/geode/gradle/plugins/DependencyConstraints.groovy
@@ -272,5 +272,10 @@ class DependencyConstraints implements Plugin<Project> {
     dependencySet(group: 'org.springframework.session', version: '2.4.1') {
       entry('spring-session-data-redis')
     }
+
+    dependencySet(group: 'org.jetbrains', version: '20.1.0') {
+      entry('annotations')
+    }
+
   }
 }
--- a/geode-common/build.gradle
+++ b/geode-common/build.gradle
@@ -26,8 +26,6 @@ dependencies {
   implementation(platform(project(':boms:geode-all-bom')))
   implementation('com.fasterxml.jackson.core:jackson-databind')
 
-  api('org.jetbrains:annotations')
-
   // test
   testImplementation('junit:junit')
   testImplementation('org.assertj:assertj-core')
--- a/geode-common/build.gradle
+++ b/geode-common/build.gradle
@@ -26,6 +26,7 @@ dependencies {
   implementation(platform(project(':boms:geode-all-bom')))
   implementation('com.fasterxml.jackson.core:jackson-databind')
 
+  api('org.jetbrains:annotations')
 
   // test
   testImplementation('junit:junit')
--- a/geode-core/build.gradle
+++ b/geode-core/build.gradle
@@ -202,6 +202,8 @@ dependencies {
     // find bugs leaks in from spring, needed to remove warnings.
     compileOnly('com.google.code.findbugs:jsr305')
 
+    compileOnly('org.jetbrains:annotations')
+
     //Jgroups is a core component of our membership system.
     implementation('org.jgroups:jgroups')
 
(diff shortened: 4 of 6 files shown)


pivotal-jbarrett:
Awesome! Not only did it find a bug but that there was no test coverage for that case. Can you fix both issue in this PR?

geode-common/build.gradle
```diff
@@ -26,6 +26,7 @@ dependencies {
   implementation(platform(project(':boms:geode-all-bom')))
   implementation('com.fasterxml.jackson.core:jackson-databind')
 
+  api('org.jetbrains:annotations')
```
onichols-pivotal, line 29: https://github.com/JetBrains/java-annotations suggests `compileOnly` rather than `api` ... is the intent to ship jetbrains annotations .jar with Geode, or use it only at compile time?  I could see arguments either way (or maybe for technical reasons we don't actually have a choice?), just want to be sure we get the right scope because scopes are tricky.  @rhoughton-pivot any thoughts?

pivotal-jbarrett: I concur, it should be compileOnly.

echobravopapa (commented):
I'll have a second look when this is resolved, thanks @Bill

geode-common/build.gradle
Bill: I want to introduce a compile-time dependency on `annotation`, in any subproject that depends on `geode-common`, because I want to save my colleagues from having to add an explicit dependency on `annotation` to the `build.gradle` for each subproject, as they desire to add a `@NotNull` annotation in some new place.

I tried changing this to `compileOnly('org.jetbrains:annotations')` but now the compiler/IDE says  `java: package org.jetbrains.annotations does not exist` where I'm trying to import it over in `SocketMessageWriter.java` in `geode-core`.

buildSrc/src/main/groovy/org/apache/geode/gradle/plugins/DependencyConstraints.groovy
```diff
@@ -272,5 +272,10 @@ class DependencyConstraints implements Plugin<Project> {
     dependencySet(group: 'org.springframework.session', version: '2.4.1') {
       entry('spring-session-data-redis')
     }
+
```
rhoughton-pivot, line 275: Are we expecting to have more modules from `org.jetbrains` in the product? If not, then the syntax above for individual modules is a little cleaner.

geode-common/build.gradle
rhoughton-pivot: @kohlmu-pivotal wanted the same thing: to have a java project or .gradle file, whose only goal was to make it easy for other Geode subprojects to get dependencies. Maybe he found a solution?

buildSrc/src/main/groovy/org/apache/geode/gradle/plugins/DependencyConstraints.groovy
Bill: @pivotal-jbarrett  asked explicitly for this syntax. I don't have an opinion on it

Bill:
I gave up on trying to make the compile-time dependency on JetBrains annotations global within Geode and simply added the compile-time dependency directly to `geode-core`. Others that wish to use the `@NotNull` annotation in other Geode subprojects can follow the example.

geode-common/build.gradle
Bill: I'm calling this resolved. I gave up trying to make the compile-time dependency global since it seems kinda harder than it's worth.

buildSrc/src/main/groovy/org/apache/geode/gradle/plugins/DependencyConstraints.groovy
pivotal-jbarrett: Flip a coin...

boms/geode-all-bom/src/test/resources/expected-pom.xml
```diff
@@ -877,6 +877,12 @@
         <version>2.4.1</version>
         <scope>compile</scope>
       </dependency>
+      <dependency>
+        <groupId>org.jetbrains</groupId>
+        <artifactId>annotations</artifactId>
+        <version>20.1.0</version>
+        <scope>compile</scope>
```
pivotal-jbarrett, line 884: The seems off. I would expect the default scope for this type to be `compileOnly`, though I don't think it really matters much in a bom.

pivotal-jbarrett (approved):

boms/geode-all-bom/src/test/resources/expected-pom.xml
Bill: is there any such thing as a `compileOnly` scope for a Maven project?

pivotal-jbarrett: sort of... compile scoped with optional true.

pivotal-jbarrett (approved):

PR #5969, opened 2021-01-28
GEODE-8873: Do not log stack trace for Exception in MessageDispatcher

DonalEvans:
- For benign exceptions encountered after stopping the MessageDispatcher, do not log the entire Exception stack trace at info level

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/MessageDispatcher.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/MessageDispatcher.java
@@ -19,6 +19,7 @@ import static org.apache.geode.util.internal.UncheckedUtils.uncheckedCast;
 import java.io.IOException;
 import java.net.Socket;
 import java.nio.ByteBuffer;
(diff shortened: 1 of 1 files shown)


davebarnes97 (approved):

bschuchardt (approved):
This fixes the problem but I'm wondering if the code would be cleaner if you did just one null check on _list_ like

> if (list == null) {
>   list = Collections.emptyList();
> }
> 
Then you could get rid of the rest of the null checks.

echobravopapa (approved):
Change looks good, appreciate the clean up along the way.

DonalEvans:
> This fixes the problem but I'm wondering if the code would be cleaner if you did just one null check on _list_ like
> 
> > if (list == null) {
> > list = Collections.emptyList();
> > }
> 
> Then you could get rid of the rest of the null checks.

I took a slightly different approach to the one suggested, since if `peek()` throws an exception, it would still have been possible for `list` to be null in the last two places it was used. Instantiating `list` at the time of declaration and then using `addAll()` prevents it from ever being null.

PR #5978, opened 2021-01-29
GEODE-8894 allow individual deltas to trigger bucket size recalculation

ringles:
The Redis subsystem uses Deltas heavily, but by default deltas do not trigger an update to the size of their buckets. This leads to incorrect memory usage accounting over the long term, especially with the use of Redis commands like "APPEND". (See GEODE-8859.)

It is possible to set the system property "DELTAS_RECALCULATE_SIZE", but this is a global value that would affect the processing of all deltas, including non-Redis operations.

Instead, this update adds a new default method to the Delta interface, that can be overridden by individual Delta implementations (such as Redis). This triggers the same behavior as DELTAS_RECALCULATE_SIZE, but on a per-delta basis. Thus, other Geode operations will not force bucket size recalculations unless the global property is set, but Redis statistics will be correct (once they are updated to override this new method).

### DIFF
--- a/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
@@ -166,7 +166,7 @@ public class DeltaForceSizingFlagDUnitTest {
     return vm0.invoke("getSizeFromPRStats", () -> {
       Cache cache = ClusterStartupRule.getCache();
       assertThat(cache).isNotNull();
-      LocalRegion region = (LocalRegion) cache.getRegion(TEST_REGION_NAME);
+      InternalRegion region = (InternalRegion) cache.getRegion(TEST_REGION_NAME);
       if (region instanceof PartitionedRegion) {
         long total = 0;
         PartitionedRegion pr = (PartitionedRegion) region;
@@ -186,7 +186,7 @@ public class DeltaForceSizingFlagDUnitTest {
 
       Cache cache = ClusterStartupRule.getCache();
       assertThat(cache).isNotNull();
-      LocalRegion region = (LocalRegion) cache.getRegion(TEST_REGION_NAME);
+      InternalRegion region = (InternalRegion) cache.getRegion(TEST_REGION_NAME);
       return region.getEvictionCounter();
     });
   }
@@ -195,7 +195,7 @@ public class DeltaForceSizingFlagDUnitTest {
     return vm0.invoke("getObjectSizerInvocations", () -> {
       Cache cache = ClusterStartupRule.getCache();
       assertThat(cache).isNotNull();
-      LocalRegion region = (LocalRegion) cache.getRegion(TEST_REGION_NAME);
+      InternalRegion region = (InternalRegion) cache.getRegion(TEST_REGION_NAME);
       return getObjectSizerInvocations(region);
     });
   }
@@ -204,12 +204,12 @@ public class DeltaForceSizingFlagDUnitTest {
     vm0.invoke("Put data", () -> {
       Cache cache = ClusterStartupRule.getCache();
       assertThat(cache).isNotNull();
-      LocalRegion region = (LocalRegion) cache.getRegion(TEST_REGION_NAME);
+      InternalRegion region = (InternalRegion) cache.getRegion(TEST_REGION_NAME);
       region.put(DeltaForceSizingFlagDUnitTest.DELTA_KEY, value);
     });
   }
 
-  protected static int getObjectSizerInvocations(LocalRegion region) {
+  protected static int getObjectSizerInvocations(InternalRegion region) {
     TestObjectSizer sizer = (TestObjectSizer) region.getEvictionAttributes().getObjectSizer();
     int result = sizer.invocations.get();
     logger.info("objectSizerInvocations=" + result);
@@ -240,7 +240,7 @@ public class DeltaForceSizingFlagDUnitTest {
     vm.invoke("assertValueType", () -> {
       Cache cache = ClusterStartupRule.getCache();
       assertThat(cache).isNotNull();
-      LocalRegion region = (LocalRegion) cache.getRegion(TEST_REGION_NAME);
+      InternalRegion region = (InternalRegion) cache.getRegion(TEST_REGION_NAME);
       Object value = region.getValueInVM(DeltaForceSizingFlagDUnitTest.DELTA_KEY);
       switch (expectedType) {
         case RAW_VALUE:
--- a/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
@@ -1,0 +1,1070 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.apache.geode.internal.cache;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+import java.io.DataInput;
+import java.io.DataOutput;
+import java.io.File;
+import java.io.IOException;
+import java.util.Random;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import org.junit.Test;
+
+import org.apache.geode.DataSerializable;
+import org.apache.geode.DataSerializer;
+import org.apache.geode.cache.AttributesFactory;
+import org.apache.geode.cache.Cache;
+import org.apache.geode.cache.DataPolicy;
+import org.apache.geode.cache.EntryEvent;
+import org.apache.geode.cache.EvictionAction;
+import org.apache.geode.cache.EvictionAttributes;
+import org.apache.geode.cache.InterestPolicy;
+import org.apache.geode.cache.PartitionAttributes;
+import org.apache.geode.cache.PartitionAttributesFactory;
+import org.apache.geode.cache.Region;
+import org.apache.geode.cache.Scope;
+import org.apache.geode.cache.SubscriptionAttributes;
+import org.apache.geode.cache.control.ResourceManager;
+import org.apache.geode.cache.partition.PartitionRegionHelper;
+import org.apache.geode.cache.util.CacheListenerAdapter;
+import org.apache.geode.cache.util.ObjectSizer;
+import org.apache.geode.distributed.DistributedMember;
+import org.apache.geode.distributed.internal.InternalDistributedSystem;
+import org.apache.geode.test.dunit.Assert;
+import org.apache.geode.test.dunit.Host;
+import org.apache.geode.test.dunit.SerializableCallable;
+import org.apache.geode.test.dunit.SerializableRunnable;
+import org.apache.geode.test.dunit.VM;
+import org.apache.geode.test.dunit.cache.internal.JUnit4CacheTestCase;
+
+/**
+ * A test of the when we will use the object sizer to determine the actual size of objects wrapped
+ * in CacheDeserializables.
+ * <p>
+ * <p>
+ * <p>
+ * TODO - I was intending to add tests that have an index and object sizer, but it appears we don't
+ * support indexes on regions with overflow to disk.
+ */
+
+public class DeltaForceSizingFlagDUnitTest extends JUnit4CacheTestCase {
+
+  public DeltaForceSizingFlagDUnitTest() {
+    super();
+  }
+
+  @Test
+  public void testRRMemLRU() {
+    doRRMemLRUTest();
+  }
+
+  @Test
+  public void testRRMemLRUDeltaAndFlag() {
+    doRRMemLRUDeltaTest(true);
+  }
+
+  @Test
+  public void testRRMemLRUDelta() {
+    doRRMemLRUDeltaTest(false);
+  }
+
+  @Test
+  public void testRRListener() {
+    Host host = Host.getHost(0);
+    VM vm0 = host.getVM(0);
+    VM vm1 = host.getVM(1);
+
+    createRR(vm0);
+    createRR(vm1);
+
+    addListener(vm0);
+    addListener(vm1);
+
+    doListenerTestRR(vm0, vm1);
+  }
+
+  @Test
+  public void testPRMemLRU() {
+    doPRMemLRUTest();
+  }
+
+  @Test
+  public void testPRMemLRUAndFlagDeltaPutOnPrimary() {
+    doPRDeltaTestLRU(false, false, true, false);
+  }
+
+  @Test
+  public void testPRMemLRUDeltaPutOnPrimary() {
+    doPRDeltaTestLRU(false, false, true, false);
+  }
+
+  @Test
+  public void testPRMemLRUAndFlagDeltaPutOnSecondary() {
+    doPRDeltaTestLRU(false, false, false, true);
+  }
+
+  @Test
+  public void testPRMemLRUDeltaPutOnSecondary() {
+    doPRDeltaTestLRU(false, false, false, true);
+  }
+
+  @Test
+  public void testPRNoLRUDelta() {
+    doPRNoLRUDeltaTest(false);
+  }
+
+  @Test
+  public void testPRNoLRUAndFlagDelta() {
+    doPRNoLRUDeltaTest(true);
+  }
+
+  @Test
+  public void testPRListener() {
+    Host host = Host.getHost(0);
+    VM vm0 = host.getVM(0);
+    VM vm1 = host.getVM(1);
+
+    createPR(vm0, true);
+    createPR(vm1, true);
+
+    addListener(vm0);
+    addListener(vm1);
+
+    doListenerTestPR(vm0, vm1);
+  }
+
+  @Test
+  public void testPRHeapLRU() {
+    Host host = Host.getHost(0);
+    VM vm0 = host.getVM(0);
+    VM vm1 = host.getVM(1);
+
+    createPRHeapLRU(vm0);
+    createPRHeapLRU(vm1);
+
+    put(vm0, new TestKey("a"), new TestObject(100, 1000));
+
+    assertValueType(vm0, new TestKey("a"), ValueType.CD_SERIALIZED);
+    assertValueType(vm1, new TestKey("a"), ValueType.CD_SERIALIZED);
+
+    assertEquals(1, getObjectSizerInvocations(vm0));
+    long origSize0 = getSizeFromPRStats(vm0);
+    assertTrue("Size was " + origSize0, 1000 > origSize0);
+    assertEquals(1, getObjectSizerInvocations(vm1));
+    long origSize1 = getSizeFromPRStats(vm1);
+    assertTrue("Size was " + origSize1, 1000 > origSize1);
+
+    get(vm0, new TestKey("a"), new TestObject(100, 1000));
+
+    assertValueType(vm0, new TestKey("a"), ValueType.CD_DESERIALIZED);
+    assertValueType(vm1, new TestKey("a"), ValueType.CD_SERIALIZED);
+    // assertTrue(1000 < getSizeFromPRStats(vm0));
+    assertEquals(3, getObjectSizerInvocations(vm0));
+    // assertTrue(1000 > getSizeFromPRStats(vm1));
+    assertEquals(1, getObjectSizerInvocations(vm1));
+
+    // Test what happens when we reach the heap threshold??
+  }
+
+  @Test
+  public void testRRHeapLRU() {
+    Host host = Host.getHost(0);
+    VM vm0 = host.getVM(0);
+    VM vm1 = host.getVM(1);
+
+    createRRHeapLRU(vm0);
+    createRRHeapLRU(vm1);
+
+    put(vm0, "a", new TestObject(100, 1000));
+
+    assertValueType(vm0, "a", ValueType.RAW_VALUE);
+    assertValueType(vm1, "a", ValueType.CD_SERIALIZED);
+
+    assertEquals(1, getObjectSizerInvocations(vm0));
+    assertEquals(0, getObjectSizerInvocations(vm1));
+
+    get(vm1, "a", new TestObject(100, 1000));
+
+    assertValueType(vm0, "a", ValueType.RAW_VALUE);
+    assertValueType(vm1, "a", ValueType.CD_DESERIALIZED);
+    assertEquals(1, getObjectSizerInvocations(vm0));
+    assertEquals(1, getObjectSizerInvocations(vm1));
+
+    // Test what happens when we reach the heap threshold??
+  }
+
+  @Test
+  public void testPRHeapLRUDeltaWithFlagPutOnPrimary() {
+    doPRDeltaTestLRU(false, true, true, false);
+  }
+
+  @Test
+  public void testPRHeapLRUDeltaPutOnPrimary() {
+    doPRDeltaTestLRU(false, true, true, false);
+  }
+
+  @Test
+  public void testPRHeapLRUDeltaWithFlagPutOnSecondary() {
+    doPRDeltaTestLRU(false, true, false, true);
+  }
+
+  @Test
+  public void testPRHeapLRUDeltaPutOnSecondary() {
+    doPRDeltaTestLRU(false, true, false, true);
+  }
+
+  // test to cover bug41916
+  @Test
+  public void testLargeDelta() {
+    Host host = Host.getHost(0);
+    VM vm0 = host.getVM(0);
+    VM vm1 = host.getVM(1);
+
+    createPR(vm0, false);
+    createPR(vm1, false);
+    // make a string bigger than socket-buffer-size which defaults to 32k
+    int BIG_DELTA_SIZE = 32 * 1024 * 2;
+    StringBuilder sb = new StringBuilder(BIG_DELTA_SIZE);
+    for (int i = 0; i < BIG_DELTA_SIZE; i++) {
+      sb.append('7');
+    }
+    TestDelta delta1 = new TestDelta(true, sb.toString());
+
+    assignPRBuckets(vm0);
+    boolean vm0isPrimary = prHostsBucketForKey(vm0, 0);
+    if (!vm0isPrimary) {
+      assertEquals(true, prHostsBucketForKey(vm1, 0));
+    }
+    VM primaryVm;
+    VM secondaryVm;
+    if (vm0isPrimary) {
+      primaryVm = vm0;
+      secondaryVm = vm1;
+    } else {
+      primaryVm = vm1;
+      secondaryVm = vm0;
+    }
+
+    put(secondaryVm, 0, delta1);
+  }
+
+  void doPRDeltaTestLRU(boolean shouldSizeChange, boolean heapLRU, boolean putOnPrimary,
+      boolean wasDelta) {
+    Host host = Host.getHost(0);
+    VM vm0 = host.getVM(0);
+    VM vm1 = host.getVM(1);
+
+    if (heapLRU) {
+      createPRHeapLRU(vm0);
+      createPRHeapLRU(vm1);
+    } else {// memLRU
+      createPR(vm0, true);
+      createPR(vm1, true);
+    }
+    assignPRBuckets(vm0);
+    boolean vm0isPrimary = prHostsBucketForKey(vm0, 0);
+    if (!vm0isPrimary) {
+      assertEquals(true, prHostsBucketForKey(vm1, 0));
+    }
+    VM primaryVm;
+    VM secondaryVm;
+    if (vm0isPrimary) {
+      primaryVm = vm0;
+      secondaryVm = vm1;
+    } else {
+      primaryVm = vm1;
+      secondaryVm = vm0;
+    }
+
+    TestDelta delta1 = new TestDelta(false, "12345", shouldSizeChange);
+    if (putOnPrimary) {
+      put(primaryVm, 0, delta1);
+    } else {
+      put(secondaryVm, 0, delta1);
+    }
+    // if the put is done on the primary then it will be CD_DESERIALIZED on the primary
+    // otherwise it will be CD_SERIALIZED on the primary.
+    if (putOnPrimary) {
+      assertValueType(primaryVm, 0, ValueType.CD_DESERIALIZED);
+      assertEquals(1, getObjectSizerInvocations(primaryVm));
+    } else {
+      assertValueType(primaryVm, 0, ValueType.CD_SERIALIZED);
+      assertEquals(0, getObjectSizerInvocations(primaryVm));
+    }
+    // It will always be CD_SERIALIZED on the secondary.
+    assertValueType(secondaryVm, 0, ValueType.CD_SERIALIZED);
+    assertEquals(0, getObjectSizerInvocations(secondaryVm));
+
+    long origEvictionSize0 = getSizeFromEvictionStats(primaryVm);
+    long origEvictionSize1 = getSizeFromEvictionStats(secondaryVm);
+    long origPRSize0 = getSizeFromPRStats(primaryVm);
+    long origPRSize1 = getSizeFromPRStats(secondaryVm);
+    delta1.info = "1234567890";
+    delta1.hasDelta = true;
+    // Update the delta
+    if (putOnPrimary) {
+      put(primaryVm, 0, delta1);
+    } else {
+      put(secondaryVm, 0, delta1);
+    }
+
+    assertValueType(primaryVm, 0, ValueType.CD_DESERIALIZED);
+    assertValueType(secondaryVm, 0, ValueType.CD_DESERIALIZED);
+
+    if (shouldSizeChange) {
+      assertEquals(2, getObjectSizerInvocations(primaryVm));
+      // once when we deserialize the value in the cache
+      // and once when we size the new value from applying the delta
+      assertEquals(2, getObjectSizerInvocations(secondaryVm));
+    } else if (wasDelta) {
+      assertEquals(0, getObjectSizerInvocations(primaryVm));
+      // 1 sizer invoke since the first value needs to be deserialized
+      assertEquals(0, getObjectSizerInvocations(secondaryVm));
+    } else {
+      assertEquals(1, getObjectSizerInvocations(primaryVm));
+      // 1 sizer invoke since the first value needs to be deserialized
+      assertEquals(0, getObjectSizerInvocations(secondaryVm));
+    }
+
+    long finalEvictionSize0 = getSizeFromEvictionStats(primaryVm);
+    long finalEvictionSize1 = getSizeFromEvictionStats(secondaryVm);
+    long finalPRSize0 = getSizeFromPRStats(primaryVm);
+    long finalPRSize1 = getSizeFromPRStats(secondaryVm);
+    if (shouldSizeChange) {
+      // I'm not sure what the change in size should be, because we went
+      // from serialized to deserialized
+      assertTrue(finalEvictionSize0 - origEvictionSize0 != 0);
+      assertTrue(finalPRSize0 - origPRSize0 != 0);
+      assertTrue(finalEvictionSize1 - origEvictionSize1 != 0);
+      assertTrue(finalPRSize1 - origPRSize1 != 0);
+    } else {
+      assertEquals(0, finalEvictionSize1 - origEvictionSize1);
+      assertEquals(0, finalPRSize0 - origPRSize0);
+      assertEquals(0, finalPRSize1 - origPRSize1);
+    }
+  }
+
+  private void addListener(VM vm) {
+    vm.invoke(new SerializableRunnable("Add listener") {
+      @Override
+      public void run() {
+        Cache cache = getCache();
+        Region region = cache.getRegion("region");
+        try {
+          region.getAttributesMutator().addCacheListener(new TestCacheListener());
+        } catch (Exception e) {
+          Assert.fail("couldn't create index", e);
+        }
+      }
+    });
+  }
+
+
+  private void doListenerTestRR(VM vm0, VM vm1) {
+    assertEquals(0, getObjectSizerInvocations(vm0));
+    assertEquals(0, getObjectSizerInvocations(vm1));
+    put(vm0, "a", new TestObject(100, 100000));
+    assertEquals(1, getObjectSizerInvocations(vm0));
+    assertEquals(1, getObjectSizerInvocations(vm1));
+
+    long origEvictionSize0 = getSizeFromEvictionStats(vm0);
+    long origEvictionSize1 = getSizeFromEvictionStats(vm1);
+
+    assertValueType(vm0, "a", ValueType.RAW_VALUE);
+    assertValueType(vm1, "a", ValueType.CD_DESERIALIZED);
+    assertTrue(origEvictionSize0 >= 100000);
+    assertTrue(origEvictionSize1 >= 100000);
+
+    put(vm0, "a", new TestObject(200, 200000));
+    assertEquals(2, getObjectSizerInvocations(vm0));
+    assertEquals(2, getObjectSizerInvocations(vm1));
+
+    long finalEvictionSize0 = getSizeFromEvictionStats(vm0);
+    long finalEvictionSize1 = getSizeFromEvictionStats(vm1);
+
+    assertValueType(vm0, "a", ValueType.RAW_VALUE);
+    assertValueType(vm1, "a", ValueType.CD_DESERIALIZED);
+    assertEquals(100000, finalEvictionSize0 - origEvictionSize0);
+    assertEquals(100000, finalEvictionSize1 - origEvictionSize1);
+  }
+
+  private void doListenerTestPR(VM vm0, VM vm1) {
+    assertEquals(0, getObjectSizerInvocations(vm0));
+    assertEquals(0, getObjectSizerInvocations(vm1));
+    put(vm0, "a", new TestObject(100, 100000));
+    assertEquals(1, getObjectSizerInvocations(vm0));
+    assertEquals(1, getObjectSizerInvocations(vm1));
+
+    long origEvictionSize0 = getSizeFromEvictionStats(vm0);
+    long origEvictionSize1 = getSizeFromEvictionStats(vm1);
+    long origPRSize0 = getSizeFromPRStats(vm1);
+    long origPRSize1 = getSizeFromPRStats(vm1);
+
+    assertValueType(vm0, "a", ValueType.CD_DESERIALIZED);
+    assertValueType(vm1, "a", ValueType.CD_DESERIALIZED);
+    assertTrue(origEvictionSize1 >= 100000);
+    assertTrue(origEvictionSize0 >= 100000);
+    assertTrue(origPRSize0 <= 500);
+    assertTrue(origPRSize1 <= 500);
+
+    put(vm0, "a", new TestObject(200, 200000));
+    assertEquals(2, getObjectSizerInvocations(vm0));
+    assertEquals(2, getObjectSizerInvocations(vm1));
+
+    long finalEvictionSize0 = getSizeFromEvictionStats(vm0);
+    long finalEvictionSize1 = getSizeFromEvictionStats(vm1);
+    long finalPRSize0 = getSizeFromPRStats(vm0);
+    long finalPRSize1 = getSizeFromPRStats(vm1);
+
+    assertValueType(vm0, "a", ValueType.CD_DESERIALIZED);
+    assertValueType(vm1, "a", ValueType.CD_DESERIALIZED);
+    assertEquals(100000, finalEvictionSize0 - origEvictionSize0);
+    assertEquals(100000, finalEvictionSize1 - origEvictionSize1);
+    assertEquals(100, finalPRSize0 - origPRSize0);
+    assertEquals(100, finalPRSize1 - origPRSize1);
+  }
+
+
+  private void doRRMemLRUTest() {
+    Host host = Host.getHost(0);
+    VM vm0 = host.getVM(0);
+    VM vm1 = host.getVM(1);
+
+    createRR(vm0);
+    createRR(vm1);
+    put(vm0, "a", new TestObject(100, 100000));
+    long origEvictionSize0 = getSizeFromEvictionStats(vm0);
+    long origEvictionSize1 = getSizeFromEvictionStats(vm1);
+    put(vm0, "a", new TestObject(200, 200000));
+
+    assertValueType(vm0, "a", ValueType.RAW_VALUE);
+    assertValueType(vm1, "a", ValueType.CD_SERIALIZED);
+    assertEquals(2, getObjectSizerInvocations(vm0));
+
+    long finalEvictionSize0 = getSizeFromEvictionStats(vm0);
+    long finalEvictionSize1 = getSizeFromEvictionStats(vm1);
+    assertEquals(100000, finalEvictionSize0 - origEvictionSize0);
+    assertEquals(100, finalEvictionSize1 - origEvictionSize1);
+
+    assertEquals(0, getObjectSizerInvocations(vm1));
+
+    // Do a get to make sure we deserialize the object and calculate
+    // the size adjustment
+    Object v = new TestObject(200, 200000);
+    get(vm1, "a", v);
+    int vSize = CachedDeserializableFactory.calcSerializedMemSize(v);
+
+    assertValueType(vm1, "a", ValueType.CD_DESERIALIZED);
+    long evictionSizeAfterGet = getSizeFromEvictionStats(vm1);
+    assertEquals(1, getObjectSizerInvocations(vm1));
+    assertEquals(200000 + CachedDeserializableFactory.overhead() - vSize,
+        evictionSizeAfterGet - finalEvictionSize1);
+
+    // Do a put that will trigger an eviction if it is deserialized
+    put(vm0, "b", new TestObject(100, 1000000));
+
+    assertEquals(1, getEvictions(vm0));
+    assertEquals(0, getEvictions(vm1));
+
+    // Do a get to make sure we deserialize the object and calculate
+    // the size adjustment
+    get(vm1, "b", new TestObject(100, 1000000));
+    assertEquals(1, getEvictions(vm1));
+  }
+
+  private void doPRMemLRUTest() {
+    Host host = Host.getHost(0);
+    VM vm0 = host.getVM(0);
+    VM vm1 = host.getVM(1);
+
+    createPR(vm0, true);
+    createPR(vm1, true);
+    put(vm0, 0, new TestObject(100, 100000));
+    assertValueType(vm0, 0, ValueType.CD_SERIALIZED);
+    assertValueType(vm1, 0, ValueType.CD_SERIALIZED);
+    long origEvictionSize0 = getSizeFromEvictionStats(vm0);
+    long origEvictionSize1 = getSizeFromEvictionStats(vm1);
+    long origPRSize0 = getSizeFromPRStats(vm0);
+    long origPRSize1 = getSizeFromPRStats(vm1);
+    put(vm0, 0, new TestObject(200, 200000));
+
+    assertEquals(0, getObjectSizerInvocations(vm0));
+
+    long finalEvictionSize0 = getSizeFromEvictionStats(vm0);
+    long finalPRSize0 = getSizeFromPRStats(vm0);
+    long finalEvictionSize1 = getSizeFromEvictionStats(vm1);
+    long finalPRSize1 = getSizeFromPRStats(vm1);
+    assertEquals(100, finalEvictionSize0 - origEvictionSize0);
+    assertEquals(100, finalEvictionSize1 - origEvictionSize1);
+    assertEquals(100, finalPRSize0 - origPRSize0);
+    assertEquals(100, finalPRSize1 - origPRSize1);
+
+    assertEquals(0, getObjectSizerInvocations(vm1));
+
+    // Do a get to see if we deserialize the object and calculate
+    // the size adjustment
+    Object v = new TestObject(200, 200000);
+    get(vm0, 0, v);
+    int vSize = CachedDeserializableFactory.calcSerializedMemSize(v);
+    assertValueType(vm0, 0, ValueType.CD_DESERIALIZED);
+    assertValueType(vm1, 0, ValueType.CD_SERIALIZED);
+    long evictionSizeAfterGet = getSizeFromEvictionStats(vm0);
+    long prSizeAfterGet = getSizeFromPRStats(vm0);
+    assertEquals(1, getObjectSizerInvocations(vm0));
+    assertEquals(0, getObjectSizerInvocations(vm1));
+    assertEquals(200000 + CachedDeserializableFactory.overhead() - vSize,
+        evictionSizeAfterGet - finalEvictionSize0);
+    assertEquals(0, prSizeAfterGet - finalPRSize0);
+
+    // Do a put that will trigger an eviction if it is deserialized
+    // It should not be deserialized.
+    put(vm0, 113, new TestObject(100, 1024 * 1024));
+    assertValueType(vm0, 113, ValueType.CD_SERIALIZED);
+    assertValueType(vm1, 113, ValueType.CD_SERIALIZED);
+    long evictionSizeAfterPutVm1 = getSizeFromEvictionStats(vm1);
+
+    assertEquals(0, getEvictions(vm0));
+    assertEquals(0, getEvictions(vm1));
+
+    // Do a get to make sure we deserialize the object and calculate
+    // the size adjustment which should force an eviction
+    get(vm1, 113, new TestObject(100, 1024 * 1024));
+    long evictionSizeAfterGetVm1 = getSizeFromEvictionStats(vm1);
+    assertValueType(vm0, 113, ValueType.CD_SERIALIZED);
+    assertValueType(vm1, 113, ValueType.EVICTED);
+    assertEquals(1, getObjectSizerInvocations(vm0)); // from the get of key 0 on vm0
+    assertEquals(0, getEvictions(vm0));
+    assertEquals(1, getObjectSizerInvocations(vm1));
+    assertEquals(2, getEvictions(vm1));
+  }
+
+  private void doRRMemLRUDeltaTest(boolean shouldSizeChange) {
+    Host host = Host.getHost(0);
+    VM vm0 = host.getVM(0);
+    VM vm1 = host.getVM(1);
+
+    createRR(vm0);
+    createRR(vm1);
+    TestDelta delta1 = new TestDelta(false, "12345", shouldSizeChange);
+    put(vm0, "a", delta1);
+
+    assertValueType(vm0, "a", ValueType.RAW_VALUE);
+    assertValueType(vm1, "a", ValueType.CD_SERIALIZED);
+    assertEquals(1, getObjectSizerInvocations(vm0));
+    assertEquals(0, getObjectSizerInvocations(vm1));
+
+    long origEvictionSize0 = getSizeFromEvictionStats(vm0);
+    long origEvictionSize1 = getSizeFromEvictionStats(vm1);
+    delta1.info = "1234567890";
+    delta1.hasDelta = true;
+    // Update the delta
+    put(vm0, "a", delta1);
+
+    assertValueType(vm0, "a", ValueType.RAW_VALUE);
+    assertValueType(vm1, "a", ValueType.CD_DESERIALIZED);
+
+    assertEquals(2, getObjectSizerInvocations(vm0));
+
+    long finalEvictionSize0 = getSizeFromEvictionStats(vm0);
+    long finalEvictionSize1 = getSizeFromEvictionStats(vm1);
+    assertEquals(5, finalEvictionSize0 - origEvictionSize0);
+    if (shouldSizeChange) {
+      assertEquals(1, getObjectSizerInvocations(vm1));
+      // I'm not sure what the change in size should be, because we went
+      // from serialized to deserialized
+      assertTrue(finalEvictionSize1 - origEvictionSize1 != 0);
+    } else {
+      // we invoke the sizer once when we deserialize the original to apply the delta to it
+      assertEquals(0, getObjectSizerInvocations(vm1));
+      assertEquals(0, finalEvictionSize1 - origEvictionSize1);
+    }
+  }
+
+  private void doPRNoLRUDeltaTest(boolean shouldSizeChange) {
+    Host host = Host.getHost(0);
+    VM vm0 = host.getVM(0);
+    VM vm1 = host.getVM(1);
+
+    createPR(vm0, false);
+    createPR(vm1, false);
+
+    TestDelta delta1 = new TestDelta(false, "12345", shouldSizeChange);
+    put(vm0, "a", delta1);
+    long origPRSize0 = getSizeFromPRStats(vm0);
+    long origPRSize1 = getSizeFromPRStats(vm1);
+
+    // Update the delta
+    delta1.info = "1234567890";
+    delta1.hasDelta = true;
+    put(vm0, "a", delta1);
+    long finalPRSize0 = getSizeFromPRStats(vm0);
+    long finalPRSize1 = getSizeFromPRStats(vm1);
+
+    if (shouldSizeChange) {
+      // I'm not sure what the change in size should be, because we went
+      // from serialized to deserialized
+      logger.info("finalPRSize0:" + finalPRSize0 + " origPRSize0: " + origPRSize0);
+      logger.info("finalPRSize1:" + finalPRSize1 + " origPRSize1: " + origPRSize1);
+      assertTrue(finalPRSize0 - origPRSize0 != 0);
+      assertTrue(finalPRSize1 - origPRSize1 != 0);
+    } else {
+      assertEquals(0, finalPRSize0 - origPRSize0);
+      assertEquals(0, finalPRSize1 - origPRSize1);
+    }
+  }
+
+  private long getSizeFromPRStats(VM vm0) {
+    return (Long) vm0.invoke(new SerializableCallable() {
+
+      @Override
+      public Object call() {
+        Cache cache = getCache();
+        LocalRegion region = (LocalRegion) cache.getRegion("region");
+        if (region instanceof PartitionedRegion) {
+          long total = 0;
+          PartitionedRegion pr = ((PartitionedRegion) region);
+          for (int i = 0; i < pr.getPartitionAttributes().getTotalNumBuckets(); i++) {
+            total += pr.getDataStore().getBucketSize(i);
+          }
+          return total;
+        } else {
+          return 0L;
+        }
+      }
+    });
+  }
+
+
+  private long getSizeFromEvictionStats(VM vm0) {
+    return (Long) vm0.invoke(new SerializableCallable() {
+
+      @Override
+      public Object call() {
+        Cache cache = getCache();
+        LocalRegion region = (LocalRegion) cache.getRegion("region");
+        return getSizeFromEvictionStats(region);
+      }
+    });
+  }
+
+  private long getEvictions(VM vm0) {
+    return (Long) vm0.invoke(new SerializableCallable() {
+
+      @Override
+      public Object call() {
+        Cache cache = getCache();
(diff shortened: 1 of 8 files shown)


geode-core/src/distributedTest/java/org/apache/geode/internal/cache/TestDelta.java
```diff
@@ -62,10 +79,18 @@ public boolean hasDelta() {
     return hasDelta;
   }
 
+  @Override
+  public boolean getForceRecalculateSize() {
+    // new Exception("RINGLES - getForceRecalculateSize Called: " +
+    // forceRecalculateSize).printStackTrace();
```
DonalEvans, line 85: This comment should be removed.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/TestDelta.java
```diff
@@ -62,10 +79,18 @@ public boolean hasDelta() {
...
+    return forceRecalculateSize;
+  }
+
   @Override
   public synchronized void toDelta(DataOutput out) throws IOException {
     // new Exception("DAN - To Delta Called").printStackTrace();
```
DonalEvans, line 91: This comment should also be removed.

geode-core/src/main/java/org/apache/geode/Delta.java
```diff
@@ -23,27 +23,25 @@
...
+   * this object must be transmitted in its entirety.
    */
   boolean hasDelta();
 
   /**
    * This method is invoked on an application object at the delta sender, if GemFire determines the
    * presence of a delta by calling {@link Delta#hasDelta()} on the object. The delta is written to
    * the {@link DataOutput} object provided by GemFire.
-   *
+   * <p>
```
DonalEvans, line 43: Is this a necessary addition?

geode-core/src/main/java/org/apache/geode/internal/cache/EntryEventImpl.java
```diff
@@ -165,6 +165,9 @@
    */
   private byte[] deltaBytes = null;
 
+  /* Is this special about recalculating size? */
```
DonalEvans, line 168: Can this comment be removed?

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,1070 @@
...
+import org.apache.geode.test.dunit.cache.internal.JUnit4CacheTestCase;
+
+/**
+ * A test of the when we will use the object sizer to determine the actual size of objects wrapped
+ * in CacheDeserializables.
+ * <p>
+ * <p>
+ * <p>
+ * TODO - I was intending to add tests that have an index and object sizer, but it appears we don't
+ * support indexes on regions with overflow to disk.
```
DonalEvans, line 63: This comment should be cleaned up a bit. The first sentence has some grammatical errors and the TODO should be removed.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,1070 @@
...
+ */
+
+public class DeltaForceSizingFlagDUnitTest extends JUnit4CacheTestCase {
```
DonalEvans, line 66: JUnit4CacheTestCase is deprecated, so this test class should not extend it.

DonalEvans (changes requested):
Some minor cleanup of comments, but the main change required is that the test class added is extending a deprecated class and should be redone to instead use DistributedRule or ClusterStartupRule.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
ringles: Fair enough; this was actually left over from the original code taken from SizingFlagDUnitTest.java.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/TestDelta.java
ringles: If so, I'll go ahead and remove the "DAN" ones, too.

geode-core/src/main/java/org/apache/geode/internal/cache/EntryEventImpl.java
ringles: Well, it probably shouldn't be removed, since all the other fields have comments as well. However, it can be improved.

geode-core/src/main/java/org/apache/geode/Delta.java
ringles: No, it'll be yanked, thanks!

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
ringles: Updated to use CacheTestCase.

DonalEvans: CacheTestCase is an abstract class that extends JUnit4CacheTestCase and does not change any of the behaviour in it and should really also be deprecated. Per the comment in the `@deprecated` tag in JUnit4CacheTestCase, "Please use {@link DistributedRule} and Geode User APIs or {@link ClusterStartupRule} instead."

ringles: The code's been updated to use ClusterStartupRule instead. A little more cleanup of comments and cruft from the original inspiration (SizingFlagDUnitTest) has been done as well.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,995 @@
...
+      LocalRegion region = (LocalRegion) cache.getRegion("region");
+      return getSizeFromEvictionStats(region);
+    });
+  }
+
+  private long getEvictions(VM vm0) {
+    return (Long) vm0.invoke("getEvictions", () -> {
+      Cache cache = ClusterStartupRule.getCache();
+      LocalRegion region = (LocalRegion) cache.getRegion("region");
+      return getEvictions(region);
```
DonalEvans, line 644: `getEvictions()` is only used here, so can be inlined.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,995 @@
...
+    });
+  }
+
+  private int getObjectSizerInvocations(VM vm0) {
+    return (Integer) vm0.invoke("getObjectSizerInvocations", () -> {
+      Cache cache = ClusterStartupRule.getCache();
+      LocalRegion region = (LocalRegion) cache.getRegion("region");
+      return getObjectSizerInvocations(region);
```
DonalEvans, line 652: `getObjectSizerInvocations` is only used here, so can be inlined.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,995 @@
...
+    memberVM.invoke("assignPRBuckets", () -> {
+      Cache cache = ClusterStartupRule.getCache();
+      PartitionRegionHelper.assignBucketsToPartitions(cache.getRegion("region"));
+    });
+  }
+
+  private boolean prHostsBucketForKey(MemberVM memberVM, final Object key) {
+    Boolean result = (Boolean) memberVM.invoke("prHostsBucketForKey", () -> {
+      Cache cache = ClusterStartupRule.getCache();
+      DistributedMember myId = cache.getDistributedSystem().getDistributedMember();
```
DonalEvans, line 666: The compiler warning here can be fixed by asserting that `cache` is not null before calling methods on it.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,995 @@
...
+      Region region = cache.getRegion("region");
```
DonalEvans, line 667: Compiler warnings here can be fixed by using `Region<Object, Object>`.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,995 @@
...
+
+  private void assignPRBuckets(MemberVM memberVM) {
+    memberVM.invoke("assignPRBuckets", () -> {
+      Cache cache = ClusterStartupRule.getCache();
+      PartitionRegionHelper.assignBucketsToPartitions(cache.getRegion("region"));
+    });
+  }
+
+  private boolean prHostsBucketForKey(MemberVM memberVM, final Object key) {
+    Boolean result = (Boolean) memberVM.invoke("prHostsBucketForKey", () -> {
```
DonalEvans, line 664: The cast to Boolean is redundant and can be removed.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,995 @@
...
+      DistributedMember myId = cache.getDistributedSystem().getDistributedMember();
+      Region region = cache.getRegion("region");
+      DistributedMember hostMember = PartitionRegionHelper.getPrimaryMemberForKey(region, key);
+      if (hostMember == null) {
+        throw new IllegalStateException("bucket for key " + key + " is not hosted!");
+      }
+      boolean res = Boolean.valueOf(myId.equals(hostMember));
+      return res;
+    });
+    return result.booleanValue();
```
DonalEvans, line 675: The call to `Boolean.valueOf()` is unnecessary, and the first return statement can be changed to simply `return myId.equals(hostMember);`. Also, the call of `result. booleanValue()` is unnecessary, meaning that the method can instead just return the result of the `invoke()` call rather than assigning a variable.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,995 @@
...
+  }
+
+  private void put(VM vm0, final Object key, final Object value) {
+    vm0.invoke("Put data", () -> {
+      Cache cache = ClusterStartupRule.getCache();
+      LocalRegion region = (LocalRegion) cache.getRegion("region");
```
DonalEvans, line 681: Compiler warning can be fixed here by asserting that `cache` is not null before calling methods on it.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,995 @@
...
+      region.put(key, value);
+    });
+  }
+
+  private void get(VM vm0, final Object key, final Object value) {
+    vm0.invoke("Get data", () -> {
+      Cache cache = ClusterStartupRule.getCache();
+      LocalRegion region = (LocalRegion) cache.getRegion("region");
```
DonalEvans, line 689: Compiler warning can be fixed here by asserting that `cache` is not null before calling methods on it.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,995 @@
...
+  }
+
+  private static long getEvictions(LocalRegion region) {
+    return region.getTotalEvictions();
+  }
+
+  private void createRR(MemberVM memberVM) {
+    memberVM.invoke("Create replicateRegion", () -> {
+      Cache cache = ClusterStartupRule.getCache();
+      AttributesFactory<Integer, TestDelta> attr = new AttributesFactory<Integer, TestDelta>();
```
DonalEvans, line 713: `AttributesFactory` is deprecated and should not be used. `Cache.createRegionFactory()` should be used instead.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,995 @@
...
+      attr.setDiskSynchronous(true);
+      attr.setDataPolicy(DataPolicy.REPLICATE);
+      attr.setScope(Scope.DISTRIBUTED_ACK);
+      attr.setEvictionAttributes(EvictionAttributes.createLRUMemoryAttributes(1,
+          new TestObjectSizer(), EvictionAction.OVERFLOW_TO_DISK));
+      attr.setDiskDirs(getMyDiskDirs());
```
DonalEvans, line 719: `setDiskDirs()` is deprecated and should not be used. Instead, a disk store should be created using `DiskStoreFactory` and passed to the `RegionFactory` using `setDiskStoreName()` with the name of the disk store.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,995 @@
...
+      Region region = cache.createRegion("region", attr.create());
+    });
+  }
+
+  private void assertValueType(VM vm, final Object key, final ValueType expectedType) {
+    vm.invoke("assertValueType", () -> {
+      Cache cache = ClusterStartupRule.getCache();
+      LocalRegion region = (LocalRegion) cache.getRegion("region");
```
DonalEvans, line 727: Compiler warning here can be fixed by asserting that `cache` is not null before calling methods on it.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,995 @@
...
+      Object value = region.getValueInVM(key);
+      switch (expectedType) {
+        case RAW_VALUE:
+          assertTrue("Value was " + value + " type " + value.getClass(),
+              !(value instanceof CachedDeserializable));
```
DonalEvans, line 732: This assert can be simplified to `assertFalse()` and removing the negation from the condition being asserted on.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,995 @@
...
+          assertTrue("Value was " + value + " type " + value.getClass(),
+              value instanceof CachedDeserializable);
+          assertTrue("Value not serialized",
+              ((CachedDeserializable) value).getValue() instanceof byte[]);
+          break;
+        case CD_DESERIALIZED:
+          assertTrue("Value was " + value + " type " + value.getClass(),
+              value instanceof CachedDeserializable);
+          assertTrue("Value was serialized",
+              !(((CachedDeserializable) value).getValue() instanceof byte[]));
```
DonalEvans, line 744: This assert can be simplified to `assertFalse()` and removing the negation from the condition being asserted on.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,995 @@
...
+          break;
+        case EVICTED:
+          assertEquals(null, value);
```
DonalEvans, line 747: This should be `assertNull(value);`

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,995 @@
...
+      }
+    });
+  }
+
+  private static File[] getMyDiskDirs() {
+    long random = new Random().nextLong();
+    File file = new File(Long.toString(random));
+    file.mkdirs();
+    return new File[] {file};
+  }
```
DonalEvans, line 758: This method might be better if it was replaced by something using the TemporaryFolder Rule, to ensure that everything is cleaned up properly after the test.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,995 @@
...
+
+  private void createPR(MemberVM memberVM, final boolean enableLRU) {
+    memberVM.invoke("Create partitioned region", () -> {
+      Cache cache = ClusterStartupRule.getCache();
+      AttributesFactory<Integer, TestDelta> attr = new AttributesFactory<Integer, TestDelta>();
```
DonalEvans, line 763: `AttributesFactory` is deprecated and should not be used. `Cache.createRegionFactory()` should be used instead.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,995 @@
...
+      attr.setDiskSynchronous(true);
+      PartitionAttributesFactory<Integer, TestDelta> paf =
+          new PartitionAttributesFactory<Integer, TestDelta>();
+      paf.setRedundantCopies(1);
+      if (enableLRU) {
+        paf.setLocalMaxMemory(1); // memlru limit is 1 megabyte
+        attr.setEvictionAttributes(EvictionAttributes
+            .createLRUMemoryAttributes(new TestObjectSizer(), EvictionAction.OVERFLOW_TO_DISK));
+        attr.setDiskDirs(getMyDiskDirs());
```
DonalEvans, line 772: `setDiskDirs()` is deprecated and should not be used. Instead, a disk store should be created using `DiskStoreFactory` and passed to the `RegionFactory` using `setDiskStoreName()` with the name of the disk store.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,995 @@
...
+      attr.setSubscriptionAttributes(new SubscriptionAttributes(InterestPolicy.ALL));
+      Region<Integer, TestDelta> region = cache.createRegion("region", attr.create());
+    });
+
+  }
+
+  private void createRRHeapLRU(MemberVM memberVM) {
+    memberVM.invoke("Create rr-heap-lru", () -> {
+      Cache cache = ClusterStartupRule.getCache();
+      ResourceManager manager = cache.getResourceManager();
```
DonalEvans, line 786: Compiler warning can be fixed here by asserting that `cache` is not null before calling methods on it.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,995 @@
...
+      manager.setCriticalHeapPercentage(95);
+      manager.setEvictionHeapPercentage(90);
+      AttributesFactory<Integer, TestDelta> attr = new AttributesFactory<Integer, TestDelta>();
```
DonalEvans, line 789: `AttributesFactory` is deprecated and should not be used. `Cache.createRegionFactory()` should be used instead.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,995 @@
...
+      attr.setEvictionAttributes(EvictionAttributes.createLRUHeapAttributes(new TestObjectSizer(),
+          EvictionAction.OVERFLOW_TO_DISK));
+      attr.setDiskDirs(getMyDiskDirs());
+      attr.setDataPolicy(DataPolicy.REPLICATE);
+      attr.setScope(Scope.DISTRIBUTED_ACK);
+      attr.setDiskDirs(getMyDiskDirs());
```
DonalEvans, line 795: `setDiskDirs()` is deprecated and should not be used. Instead, a disk store should be created using `DiskStoreFactory` and passed to the `RegionFactory` using `setDiskStoreName()` with the name of the disk store. Also, the method is called twice here, which seems redundant.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,995 @@
...
+  }
+
+  private void createPRHeapLRU(MemberVM memberVM) {
+    memberVM.invoke("Create pr-heap-lru", () -> {
+      Cache cache = ClusterStartupRule.getCache();
+      ResourceManager manager = cache.getResourceManager();
+      manager.setCriticalHeapPercentage(95);
+      manager.setEvictionHeapPercentage(90);
+
+      AttributesFactory<Integer, TestDelta> attr = new AttributesFactory<Integer, TestDelta>();
```
DonalEvans, line 808: `AttributesFactory` is deprecated and should not be used. `Cache.createRegionFactory()` should be used instead.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,995 @@
...
+      attr.setDiskDirs(getMyDiskDirs());
+      Region region = cache.createRegion("region", attr.create());
+    });
+
+  }
+
+  private void createPRHeapLRU(MemberVM memberVM) {
+    memberVM.invoke("Create pr-heap-lru", () -> {
+      Cache cache = ClusterStartupRule.getCache();
+      ResourceManager manager = cache.getResourceManager();
```
DonalEvans, line 804: Compiler warning can be fixed here by asserting that `cache` is not null before calling methods on it.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,995 @@
...
+      attr.setPartitionAttributes(prAttr);
+      attr.setDataPolicy(DataPolicy.PARTITION);
+      attr.setSubscriptionAttributes(new SubscriptionAttributes(InterestPolicy.ALL));
+      Region<Integer, TestDelta> region = cache.createRegion("region", attr.create());
+    });
+
+  }
+
+  private static class TestObjectSizer implements ObjectSizer {
+    private AtomicInteger invocations = new AtomicInteger();
```
DonalEvans, line 824: This can be made final.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,995 @@
...
+
+    @Override
+    public int sizeof(Object o) {
+      if (InternalDistributedSystem.getLogger().fineEnabled()) {
+        InternalDistributedSystem.getLogger()
+            .fine("TestObjectSizer invoked"/* , new Exception("stack trace") */);
+      }
```
DonalEvans, line 831: This block contains calls to deprecated methods and should be removed. If someone needs to debug the test in the future, they can always add their own logging at that time.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,995 @@
...
+        return ((TestDelta) o).info.length();
+      }
+      if (o instanceof Integer) {
+        return 0;
+      }
+      if (o instanceof TestKey) {
+        invocations.incrementAndGet();
+        return ((TestKey) o).value.length();
+      }
+      throw new RuntimeException("Unpected type to be sized " + o.getClass() + ", object=" + o);
```
DonalEvans, line 847: Typo here, should be "Unexpected".

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,995 @@
...
+    }
+  }
+
+  private static class TestKey implements DataSerializable {
```
DonalEvans, line 851: Compiler warning here can be fixed by either adding a `serialVersionUID` (the IDE should be able to generate one automatically for you) or adding a `@SuppressWarnings("serial")` annotation above the class.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,995 @@
...
+    String value;
+
+    public TestKey() {
```
DonalEvans, line 854: The compiler warning here can be fixed by adding a `@SuppressWarnings("unused")` annotation above the constructor.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,995 @@
...
+      }
+      TestKey other = (TestKey) obj;
+      if (value == null) {
+        if (other.value != null) {
+          return false;
+        }
+      } else if (!value.equals(other.value)) {
+        return false;
+      }
+      return true;
```
DonalEvans, line 899: This bit of code can be simplified to:
```
if (value == null) {
  return other.value == null;
} else {
  return value.equals(other.value);
}
```

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,995 @@
...
+    }
+
+  }
+
+  private static class TestObject implements DataSerializable {
```
DonalEvans, line 904: See earlier comment about adding a `serialVersionUID` or suppressing the warning.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,995 @@
...
+    public int sizeForSizer;
+    public int sizeForSerialization;
+
+    public TestObject() {
```
DonalEvans, line 908: See earlier comment about suppressing the "unused" warning here.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,995 @@
...
+        return false;
+      }
+      TestObject other = (TestObject) obj;
+      if (sizeForSerialization != other.sizeForSerialization) {
+        return false;
+      }
+      if (sizeForSizer != other.sizeForSizer) {
+        return false;
+      }
+      return true;
```
DonalEvans, line 962: This can be simplified to `return sizeForSizer == other.sizeForSizer;`.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,995 @@
...
+    }
+
+    @Override
+    public String toString() {
+      return "TestObject [sizeForSerialization=" + sizeForSerialization + ", sizeForSizer="
+          + sizeForSizer + "]";
+    }
+  }
+
+  public static class TestCacheListener extends CacheListenerAdapter {
```
DonalEvans, line 972: Compiler warnings can be fixed by using `TestCacheListener<K, V> extends CacheListenerAdapter<K, V>` here.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,995 @@
...
+
+    @Override
+    public void afterCreate(EntryEvent event) {
```
DonalEvans, line 975: Complier warning here can be fixed by using `EntryEvent<K, V> event`.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,995 @@
...
+      // Make sure we deserialize the new value
+      event.getRegion().getCache().getLogger().fine("invoked afterCreate with " + event);
+      event.getRegion().getCache().getLogger().info(String.format("%s",
+          "value is " + event.getNewValue()));
+    }
+
+    @Override
+    public void afterUpdate(EntryEvent event) {
```
DonalEvans, line 983: Complier warning here can be fixed by using `EntryEvent<K, V> event`.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,995 @@
...
+    }
+  }
+
+  public static class TestCacheListener extends CacheListenerAdapter {
+
+    @Override
+    public void afterCreate(EntryEvent event) {
+      // Make sure we deserialize the new value
+      event.getRegion().getCache().getLogger().fine("invoked afterCreate with " + event);
+      event.getRegion().getCache().getLogger().info(String.format("%s",
```
DonalEvans, line 978: `Region.getCache()` is deprecated and should not be used. `(Cache) Region.getRegionService()` should be used instead to get the cache.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,995 @@
...
+          "value is " + event.getNewValue()));
+    }
+
+    @Override
+    public void afterUpdate(EntryEvent event) {
+      // Make sure we deserialize the new value
+      event.getRegion().getCache().getLogger().fine("invoked afterUpdate with ");
+      event.getRegion().getCache().getLogger().info(String.format("%s",
```
DonalEvans, line 986: `Region.getCache()` is deprecated and should not be used. `(Cache) Region.getRegionService()` should be used instead to get the cache.

DonalEvans (changes requested):
Lots of code clean-up and fixes, I hope it's not overwhelming.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
ringles: That can't be called from a static context. Is there a non-deprecated alternative?

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/TestDelta.java
```diff
@@ -33,12 +33,28 @@
...
+    this.info = info;
+    this.forceRecalculateSize = forceRecalculateSize;
+  }
+
+  @Override
+  public String toString() {
+    return "TestDelta{" +
+        "info='" + info + '\'' +
+        "forceRecalculateSize='" + forceRecalculateSize + '\'' +
+        '}';
```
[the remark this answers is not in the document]
ringles: IntelliJ automatically reformats it that way when you type it in. It takes some awkward juggling to avoid this "help".

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
jdeppe-pivotal: Any reason you couldn't just declare a `private static final Logger logger = LogService.getLogger();` in the class?
Otherwise the call would be `((Cache)event.getRegion().getRegionService()).getLogger...`.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,995 @@
...
+    createPR(server2, true);
+
+    addListener(server1);
+    addListener(server2);
+
+    doListenerTestPR(server1.getVM(), server2.getVM());
+  }
+
+  @Test
+  public void testPRHeapLRU() {
```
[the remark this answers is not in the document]
ringles: Some of this is in order to ensure that the existence of forceRecalculateSize flag doesn't screw up other sizing calculations by accident. But since it's being tested in SizingFlagDUnitTest, we can lose it here.

geode-core/src/main/java/org/apache/geode/Delta.java
```diff
@@ -53,7 +51,13 @@
    * This method throws an {@link InvalidDeltaException} when the delta in the {@link DataInput}
    * cannot be applied to the object. GemFire automatically handles an {@link InvalidDeltaException}
    * by reattempting the update by sending the full application object.
-   *
    */
   void fromDelta(DataInput in) throws IOException, InvalidDeltaException;
+
+  /**
+   * Allows Delta implementations to ensure bucket sizes are recalculated after delta is applied
```
dschneider-pivotal, line 58: Since this method is part of the external API improve these javadocs.
1. Add a @since 1.14
2. Explain what geode does for delta size by default (it is calculated once when the entry is first created and any changes in actual size will not be noticed by geode. This can cause the statistics that report memory use to not be accurate), why this default was chosen (to optimize the use case in which the actual size of the object implementing Delta does not change), how to override this default behavior (implement this method in your class that implements Delta and have it return true), and explain why you would want to do that (if your Delta causes the actual size of the instance implementing Delta to change). It would also be good to point out that changing this default behavior to recalculate may hurt performance.

geode-core/src/main/java/org/apache/geode/internal/cache/EntryEventImpl.java
```diff
@@ -165,6 +165,9 @@
    */
   private byte[] deltaBytes = null;
 
+  /** If true, causes deltas to trigger recalculation of bucket size **/
+  private boolean forceRecalculateSize = false;
```
dschneider-pivotal, line 169: I'd like to see this new variable removed and instead, in the two places we use it, call delta.getForceRecalculateSize().
You can easily do this on line 1722 because it had just done (on line 1718) a test that v is an instanceof Delta. So add the next line: Delta delta = (Delta)v;

On line 1865 the delta instance is the "value" local variable. You can see this on line 1841 when this code casts "value" to Delta. So I'd recommend you add a local variable at 1841 like so: Delta delta = (Delta)value;
and then use that on 1865 to decide if we should recalculate.
Having the extra state on EntryEventImpl adds complexity and in the future it could get out of sync with Delta.getForceRecalculateSize(). By just calling this new method I think the code will be better.

dschneider-pivotal (changes requested):

geode-core/src/main/java/org/apache/geode/Delta.java
ringles: Missed the "@SInCE 1.14" the first time, adding now.

geode-core/src/main/java/org/apache/geode/internal/cache/EntryEventImpl.java
ringles: Sorted out; looks like when I attempted this once I grabbed the wrong delta value at one point. Anyway, it works now without the member variable.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,995 @@
...
+import org.apache.geode.test.dunit.Assert;
+import org.apache.geode.test.dunit.VM;
+import org.apache.geode.test.dunit.rules.ClusterStartupRule;
+import org.apache.geode.test.dunit.rules.MemberVM;
+
+/**
+ * Tests the use of the per-delta "forceRecalculateSize" flag.
+ */
+
+public class DeltaForceSizingFlagDUnitTest {
```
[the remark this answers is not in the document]
ringles: The parameterization makes it hard to run individual tests, and with the number of tests significantly reduced, the benefit of this would be marginal. If we have to add more tests, though, I'll keep it in mind.

geode-core/src/main/java/org/apache/geode/Delta.java
```diff
@@ -55,7 +55,20 @@
   void fromDelta(DataInput in) throws IOException, InvalidDeltaException;
 
   /**
-   * Allows Delta implementations to ensure bucket sizes are recalculated after delta is applied
+   * By default, buckets do not recalculate their size when deltas are applied. This optimizes for
```
dschneider-pivotal, line 58: Instead of "buckets" I would focus this on deltas. So say something like: "By default, the size of an entry value that implements the Delta interface is not recalculated when the delta changes.

geode-core/src/main/java/org/apache/geode/Delta.java
```diff
@@ -55,7 +55,20 @@
...
+   * the case where the sie of an entry does not change. However, if the size does increase or
```
dschneider-pivotal, line 59: typo: "sie" -> "size"

geode-core/src/main/java/org/apache/geode/Delta.java
```diff
@@ -55,7 +55,20 @@
...
+   * decrease, this default behavior can result in the memory usage statistics becoming inaccurate.
+   *
+   * There is a global Geode property, DELTAS_RECALCULATE_SIZE, which can be used to cause all
```
dschneider-pivotal, line 62: I think it might be confusing the calls this a "Geode property" because it might get confused with the gemfire properties. This is actually a java system property that Geode reads. And the actual property is "gemfire.DELTAS_RECALCULATE_SIZE".

geode-core/src/main/java/org/apache/geode/Delta.java
```diff
@@ -55,7 +55,20 @@
...
+   * deltas to trigger bucket size recalculation when deltas are applied. By default, this is set
+   * to 'false' because of potential performance impacts when every delta triggers a recalculation.
+   *
+   * This method allows bucket size recalculation on a per-delta basis. For example, geode-redis
```
dschneider-pivotal, line 66: Once again I would not refer to bucket size in this context. Changing "bucket" to "delta" here makes sense. You could add a section that says what difference having an accurate delta size makes. You could mention that geode has statistics that track memory usage and that geode attempts to keep a partitioned region's memory usage balanced across the cluster using the size.

geode-core/src/main/java/org/apache/geode/Delta.java
```diff
@@ -55,7 +55,20 @@
...
+   * uses deltas heavily, and many commands (e.g. APPEND) cause the size of the entry to change.
+   * By overriding this method to return 'true', Redis memory usage statistics can be kept accurate
```
dschneider-pivotal, line 68: I recommend that you remove the redis example stuff from this section.

geode-core/src/main/java/org/apache/geode/internal/cache/EntryEventImpl.java
```diff
@@ -1719,7 +1716,7 @@ private void setNewValueInRegion(final InternalRegion owner, final RegionEntry r
       int vSize;
       Object ov = basicGetOldValue();
       if (ov instanceof CachedDeserializable && !(GemFireCacheImpl.DELTAS_RECALCULATE_SIZE
-          || this.forceRecalculateSize)) {
+          || ((org.apache.geode.Delta) v).getForceRecalculateSize())) {
```
dschneider-pivotal, line 1719: Would it be possible to import Delta so we these instanceof and casts get a bit shorter?

dschneider-pivotal (changes requested):
Thanks for getting rid of the EntryEventImpl field.
The javadocs just need a few tweaks.

dschneider-pivotal (approved):

geode-core/src/main/java/org/apache/geode/internal/cache/EntryEventImpl.java
ringles: Done!

ringles: Done!

geode-core/src/main/java/org/apache/geode/Delta.java
ringles: New version of comment up.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,995 @@
...
+    }
+  }
+
+  private long getSizeFromPRStats(VM vm0) {
+    return (Long) vm0.invoke("getSizeFromPRStats", () -> {
+      Cache cache = ClusterStartupRule.getCache();
+      LocalRegion region = (LocalRegion) cache.getRegion("region");
+      if (region instanceof PartitionedRegion) {
+        long total = 0;
+        PartitionedRegion pr = ((PartitionedRegion) region);
```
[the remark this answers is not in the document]
ringles: Done.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,995 @@
...
+        for (int i = 0; i < pr.getPartitionAttributes().getTotalNumBuckets(); i++) {
```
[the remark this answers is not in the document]
ringles: Done.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,995 @@
...
+      }
+    });
+  }
+
+  private long getSizeFromEvictionStats(VM vm0) {
+    return (Long) vm0.invoke("getSizeFromEvictionStats", () -> {
+
+      Cache cache = ClusterStartupRule.getCache();
+      LocalRegion region = (LocalRegion) cache.getRegion("region");
+      return getSizeFromEvictionStats(region);
```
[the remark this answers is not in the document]
ringles: I'm actually in favor of breaking down methods into smaller units with descriptive names. I think this is actually a more readable version, especially since under the hood the compiler will inline it anyway.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
ringles: See above comment.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
ringles: Ditto above.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
ringles: Done.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
ringles: Done.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
ringles: Done.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
ringles: Strikes me as a little confusing; it might have some benefit to compilation, except JIT compilers should optimize this anyway. Given that, it's just a little more noise for the reader to process. Am I missing something?

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
ringles: Done.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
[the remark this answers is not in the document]
DonalEvans: In this case, the `getSizeFromEvictionStats(region)` method is so small that this line be replaced by simply `return region.getEvictionCounter();`. Adding a method to the test class just for this one call seems to be adding unnecessary complexity.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
DonalEvans: Here also, this line would become just `return ((TestObjectSizer) region.getEvictionAttributes().getObjectSizer()).invocations.get();` which seems pretty clear, especially given the context of being called from a method named "getObjectSizerInvocations".

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
DonalEvans: No, it's just something the static analyzer flagged, and I've read that liberally using final is good practice e.g. https://softwareengineering.stackexchange.com/a/98703

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
ringles: Done.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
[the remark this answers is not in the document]
ringles: True, though an "EvictionCounter" is not obviously a "size" as such, so the name adds a touch of clarity of intent for the Geode newbie. Still, for one line I'm not too worried and I've inlined it.

geode-core/src/main/java/org/apache/geode/internal/cache/LocalRegion.java
```diff
@@ -1737,7 +1737,6 @@ private void extractDeltaIntoEvent(Object value, EntryEventImpl event) {
               throw new DeltaSerializationException("Caught exception while sending delta", e);
             }
             event.setDeltaBytes(hdos.toByteArray());
-            getCachePerfStats().endDeltaPrepared(start);
```
DonalEvans, line 1740: With the removal of this line, the `start` variable on line 1731 can now be removed as it's no longer used.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,460 @@
...
+      }
+      if (o instanceof TestKey) {
+        invocations.incrementAndGet();
+        return ((TestKey) o).value.length();
+      }
+      throw new RuntimeException("Unexpected type to be sized " + o.getClass() + ", object=" + o);
+    }
+  }
+
+  private static class TestKey implements DataSerializable {
```
DonalEvans, line 331: This class is no longer instantiated anywhere in the test, so can be removed.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,460 @@
...
+      }
+    }
+
+  }
+
+  private static class TestObject implements DataSerializable {
+    public int sizeForSizer;
+    public int sizeForSerialization;
+
+    public TestObject(int sizeForSerialization, int sizeForSizer) {
```
DonalEvans, line 381: This class is no longer instantiated anywhere in the test, so can be removed.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,460 @@
...
+          + sizeForSizer + "]";
+    }
+  }
+
+  public static class TestCacheListener<K, V> extends CacheListenerAdapter<K, V> {
+
+    @Override
+    public void afterCreate(EntryEvent<K, V> event) {
+      // Make sure we deserialize the new value
+      logger.debug("invoked afterCreate with " + event);
```
DonalEvans, line 442: Since this is a test, the only time this logging should ever really need to be looked at is if the test fails. It would make life easier in that scenario if this was logged at info level so that as much information is available for anyone looking at a failure without having to rerun the test with finer logging enabled.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,460 @@
...
+      logger.info(String.format("%s",
```
DonalEvans, line 443: The use of `String.format()` here is unnecessary.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,460 @@
...
+          "value is " + event.getNewValue()));
+    }
+
+    @Override
+    public void afterUpdate(EntryEvent<K, V> event) {
+      // Make sure we deserialize the new value
+      logger.debug("invoked afterUpdate with ");
+      logger.info(String.format("%s",
+          "value is " + event.getNewValue()));
```
DonalEvans, line 452: See above comments in the `afterCreate()` method.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,460 @@
...
+      regionFactory.create(TEST_REGION_NAME);
+    });
+  }
+
+  private static class TestObjectSizer implements ObjectSizer {
+    private final AtomicInteger invocations = new AtomicInteger();
+
+    @Override
+    public int sizeof(Object o) {
+      logger.debug("TestObjectSizer invoked"/* , new Exception("stack trace") */);
```
DonalEvans, line 311: The commented out code here should be removed, possibly along with the rest of the log message. It doesn't seem to add much in terms of helping a developer debug a possible future failure. If the logging is left in, it should be changed to info level.

DonalEvans (changes requested):
A bit more unused code to be removed, and some changes to test logging to make it more useful/visible in the event of a test failure.

geode-core/src/main/java/org/apache/geode/internal/cache/EntryEventImpl.java
```diff
@@ -1713,10 +1713,11 @@ private void setNewValueInRegion(final InternalRegion owner, final RegionEntry r
     // This is a horrible hack, but we need to get the size of the object
     // When we store an entry. This code is only used when we do a put
     // in the primary.
-    if (v instanceof org.apache.geode.Delta && getRegion().isUsedForPartitionedRegionBucket()) {
+    if (v instanceof Delta && getRegion().isUsedForPartitionedRegionBucket()) {
       int vSize;
       Object ov = basicGetOldValue();
-      if (ov instanceof CachedDeserializable && !GemFireCacheImpl.DELTAS_RECALCULATE_SIZE) {
+      if (ov instanceof CachedDeserializable && !(GemFireCacheImpl.DELTAS_RECALCULATE_SIZE
```
kirklund, line 1719: Please add unit test coverage to EntryEventImplTest for any changes to EntryEventImpl.

You don't need to add tests tests for importing org.apache.geode.Delta, but anything that involves touching the logic or behavior like this block should have unit test coverage in addition to integration/distributed tests.

geode-core/src/main/java/org/apache/geode/internal/cache/EntryEventImpl.java
```diff
@@ -1858,7 +1859,8 @@ private void processDeltaBytes(Object oldValueInVM) {
       if (wasCD) {
         CachedDeserializable old = (CachedDeserializable) oldValueInVM;
         int valueSize;
-        if (GemFireCacheImpl.DELTAS_RECALCULATE_SIZE) {
+        if (GemFireCacheImpl.DELTAS_RECALCULATE_SIZE
```
kirklund, line 1862: Same here.

kirklund (changes requested):

geode-core/src/main/java/org/apache/geode/internal/cache/EntryEventImpl.java
kirklund: If you're looking for ideas on how to get started, one basic technique from Working Effectively with Legacy Code is to extract your conditional as a method (static or non-static) and provide unit tests for that new method:
```
  @VisibleForTesting
  static boolean shouldRecalculateSize(Delta deltaValue) {
    return GemFireCacheImpl.DELTAS_RECALCULATE_SIZE || deltaValue.getForceRecalculateSize();
  }
```
And use it inline where you need to change the behavior (note this could use a LOT more cleanup but this is a start so you can start with unit test coverage:
```
      if (ov instanceof CachedDeserializable && !shouldRecalculateSize((Delta)v)) {
```
And now some unit test coverage:
```
  @Test
  public void shouldRecalculateSize_returnsTrue_ifDELTAS_RECALCULATE_SIZEisTrue() {
    GemFireCacheImpl.DELTAS_RECALCULATE_SIZE = true;
    Delta deltaValue = mock(Delta.class);
    when(deltaValue.getForceRecalculateSize())
        .thenReturn(false);

    boolean value = EntryEventImpl.shouldRecalculateSize(deltaValue);

    assertThat(value).isTrue();
  }

  @Test
  public void shouldRecalculateSize_returnsFalse_ifDELTAS_RECALCULATE_SIZEisFalse() {
    GemFireCacheImpl.DELTAS_RECALCULATE_SIZE = false;
    Delta deltaValue = mock(Delta.class);
    when(deltaValue.getForceRecalculateSize())
        .thenReturn(false);

    boolean value = EntryEventImpl.shouldRecalculateSize(deltaValue);

    assertThat(value).isFalse();
  }

  @Test
  public void shouldRecalculateSize_returnsTrue_ifGetForceRecalculateSizeIsTrue() {
    GemFireCacheImpl.DELTAS_RECALCULATE_SIZE = false;
    Delta deltaValue = mock(Delta.class);
    when(deltaValue.getForceRecalculateSize())
        .thenReturn(true);

    boolean value = EntryEventImpl.shouldRecalculateSize(deltaValue);

    assertThat(value).isTrue();
  }

  @Test
  public void shouldRecalculateSize_returnsFalse_ifGetForceRecalculateSizeIsFalse() {
    GemFireCacheImpl.DELTAS_RECALCULATE_SIZE = false;
    Delta deltaValue = mock(Delta.class);
    when(deltaValue.getForceRecalculateSize())
        .thenReturn(false);

    boolean value = EntryEventImpl.shouldRecalculateSize(deltaValue);

    assertThat(value).isFalse();
  }

  @Test
  public void shouldRecalculateSize_returnsTrue_ifGetForceRecalculateSizeIsFalse_andDELTAS_RECALCULATE_SIZEisTrue() {
    GemFireCacheImpl.DELTAS_RECALCULATE_SIZE = true;
    Delta deltaValue = mock(Delta.class);
    when(deltaValue.getForceRecalculateSize())
        .thenReturn(false);

    boolean value = EntryEventImpl.shouldRecalculateSize(deltaValue);

    assertThat(value).isTrue();
  }

  @After
  public void tearDown() {
    GemFireCacheImpl.DELTAS_RECALCULATE_SIZE = false;
  }
```
This way you can add test coverage, improve code readability, and make the types a bit more clear. You can also extract a non-static method or do further refactoring but you'll end up needing to do a lot more setUp in the unit tests.

These new tests should also make it obvious that GemFireCacheImpl.DELTAS_RECALCULATE_SIZE should probably be moved and changed and tested. But that's another PR.

kirklund (changes requested):

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
DonalEvans: This change was marked as resolved, but the unused internal class is still present.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
DonalEvans: This change was marked as resolved, but the unused internal class is still present.

geode-core/src/main/java/org/apache/geode/internal/cache/EntryEventImpl.java
ringles: Thanks for the very clear and thorough comment! I've taken your advice by breaking out the method and unit-testing it; I did rearrange and rename the tests slightly, though. I think it makes the intent a bit clearer but let me know if it can be improved.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
ringles: Lost in the shuffle of four dozen comments, my apologies. They should be cleaned up now.

DonalEvans (approved):

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
ringles: Sadly, it's a bit difficult to call TemporaryFolder.newFolder() from a static context. :(

geode-core/src/main/java/org/apache/geode/Delta.java
```diff
@@ -53,7 +51,27 @@
    * This method throws an {@link InvalidDeltaException} when the delta in the {@link DataInput}
    * cannot be applied to the object. GemFire automatically handles an {@link InvalidDeltaException}
    * by reattempting the update by sending the full application object.
-   *
    */
   void fromDelta(DataInput in) throws IOException, InvalidDeltaException;
+
+  /**
```
kirklund, line 57: The JDK classes use a single unclosed `<p>` at the start of each paragraph in javadocs:
```
/**
   * By default, entry sizes are not recalculated when deltas are applied. This optimizes for the
   * case where the size of an entry does not change. However, if an entry size does increase or
   * decrease, this default behavior can result in the memory usage statistics becoming inaccurate.
   * These are used to monitor the health of Geode instances, and for balancing memory usage across
   * partitioned regions.
   *
   * <p>
   * There is a system property, gemfire.DELTAS_RECALCULATE_SIZE, which can be used to cause all
   * deltas to trigger entry size recalculation when deltas are applied. By default, this is set
   * to 'false' because of potential performance impacts when every delta triggers a recalculation.
   *
   * <p>
   * To allow entry size recalculation on a per-delta basis, classes that extend the Delta interface
   * should override this method to return 'true'. This may impact performance of specific delta
   * types, but will not globally affect the performance of other Geode delta operations.
   *
   * @since 1.14
   */
 ```

Comment (kirklund) [reply geode-core/src/main/java/org/apache/geode/Delta.java:57]: Just to be sure... do we really want to document `gemfire.DELTAS_RECALCULATE_SIZE` in the javadocs of a User API class? Is it new? I think new system properties should be documented as `geode.PROPERTY` instead of `gemfire.`.

**Review by kirklund — CHANGES_REQUESTED** [review]
Looks like we currently have the `gemfire.DELTAS_RECALCULATE_SIZE` system property as a way to set GemFireCacheImpl.DELTAS_RECALCULATE_SIZE.

Some additional considerations to think about:

We really need to move away from referencing implementation classes like GemFireCacheImpl, even when it's for a constant. We added some annotations such as `@MutableForTesting` and `@MakeNotStatic`. 

I think the team should probably discuss GemFireCacheImpl.DELTAS_RECALCULATE_SIZE and how we can get rid of that mutable public static. You might find that it would be better to add a User API instead of adding a User documented system property for a mutable public static that we want to eventually get rid of.

If it were me, I would instead add `setDeltasRecalculateSize` (or something similar) to `CacheFactory` as the new User documented way to specify true or false. You would also add it to `InternalCacheBuilder`. Then we could replace `GemFireCacheImpl.DELTAS_RECALCULATE_SIZE` with a private final boolean that is instead passed in via the constructors of GemFireCacheImpl.

File: geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java [hunk 67]
```diff
@@ -0,0 +1,319 @@
...
+    } else {
+      assertEquals(0, finalPRSize0 - origPRSize0);
+      assertEquals(0, finalPRSize1 - origPRSize1);
+    }
+  }
+
+  private long getSizeFromPRStats(VM vm0) {
+    return vm0.invoke("getSizeFromPRStats", () -> {
+      Cache cache = ClusterStartupRule.getCache();
+      assertNotNull(cache);
```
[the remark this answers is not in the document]
DonalEvans: Jumping in here since I was the one who requested the inclusion of the `assertNotNull(cache)` calls in the first place. Without some way of confirming that the cache is not null before calling methods on it, the compiler complains that `cache` may be null. Is there a way to prevent the compiler warnings related to `ClusterStartupRule.getCache()` possibly returning null without explicitly checking the value returned?

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,319 @@
...
+              (((CachedDeserializable) value).getValue() instanceof byte[]));
+          break;
+        case EVICTED:
+          assertNull(value);
+          break;
+      }
+    });
+  }
+
+  private static File[] getMyDiskDirs() {
```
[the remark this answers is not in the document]
ringles: Removing "static" from getMyDiskDirs() compiles, but causes the tests testRRMemLRUDelta() and testRRMemLRUDeltaAndFlag() to fail with:

Caused by: java.io.NotSerializableException: org.apache.geode.internal.cache.DeltaForceSizingFlagDUnitTest
	at java.io.ObjectOutputStream.writeObject0(ObjectOutputStream.java:1184)
	at java.io.ObjectOutputStream.writeArray(ObjectOutputStream.java:1378)

Alternatively, SerializableTemporaryFolder can't be used from a static context unless it's annotated with ClassRule. Unfortunately, in that case the same tests fail with:

Caused by: java.lang.IllegalStateException: the temporary folder has not yet been created
	at org.junit.rules.TemporaryFolder.getRoot(TemporaryFolder.java:299)
	at org.junit.rules.TemporaryFolder.newFolder(TemporaryFolder.java:196)
	at org.junit.rules.TemporaryFolder.newFolder(TemporaryFolder.java:177)
	at org.apache.geode.internal.cache.DeltaForceSizingFlagDUnitTest.getMyDiskDirs(DeltaForceSizingFlagDUnitTest.java:281)
	at org.apache.geode.internal.cache.DeltaForceSizingFlagDUnitTest.lambda$createRR$f182e747$1(DeltaForceSizingFlagDUnitTest.java:241)

geode-core/src/main/java/org/apache/geode/Delta.java
ringles: gemfire.DELTAS_RECALCULATE_SIZE is not new; it's been around since at least 2017 (git-annotate shows "4/24/17 Lund") and its definition was tweaked "12/11/19 Lund". (Though that was a large commit touching a lot of files.)

It probably makes sense to document it somewhere else, too, but it seems relevant here...

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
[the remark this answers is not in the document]
ringles: For the moment, they've been tweaked to use the AssertJ form.

geode-core/src/main/java/org/apache/geode/Delta.java
ringles: (The <p> tag has been addressed.)

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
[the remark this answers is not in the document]
ringles: Again, it may be time to move substantial restructuring to a separate Jira.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,319 @@
...
+
+  protected static int getObjectSizerInvocations(LocalRegion region) {
+    TestObjectSizer sizer = (TestObjectSizer) region.getEvictionAttributes().getObjectSizer();
+    int result = sizer.invocations.get();
+    logger.info("objectSizerInvocations=" + result);
+    return result;
+  }
+
+  private void createRR(MemberVM memberVM) {
+    memberVM.invoke("Create replicateRegion", () -> {
```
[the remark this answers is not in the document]
ringles: Is it possible to make that restructuring part of a separate Jira? This PR has been hanging around for a while and gone through substantial revision already. The suggestions have been good, and the code's a lot clearer now - but we may be reaching a point of diminishing returns. I'm worried that this PR might drift out of sync with develop, and become much harder to merge.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/DeltaForceSizingFlagDUnitTest.java
```diff
@@ -0,0 +1,319 @@
...
+    assertTrue(file.mkdirs());
+    return new File[] {file};
+  }
+
+  private void createPR(MemberVM memberVM) {
+    memberVM.invoke("Create partitioned region", () -> {
+      Cache cache = ClusterStartupRule.getCache();
+      assertNotNull(cache);
+
+      RegionFactory<Integer, TestDelta> regionFactory = cache.createRegionFactory();
```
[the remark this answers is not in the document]
ringles: Did a quick rearrangement, I think it flows better and it's low-hanging fruit for sure.

kirklund (approved):
Approved after further discussion with @ringles. The plan is to break up any additional changes for feedback into multiple PRs.

PR #5987, opened 2021-02-01
GEODE-8817: server hangs in cache.close()

bschuchardt:
Long ago we learned that socket operations can hang if there is a true network failure.  The shutdown of peer-to-peer communication sockets was altered to be performed asynchronously in an executor so that, if the Socket.close() operation was blocked, shutdown of the rest of the cache could continue.

This PR extends that logic to client/server and WAN communications using the SocketCloser class that is used for peer-to-peer communications.

Since the blocking problem is in the Java platform itself and only occurs if there is a real network problem it is not possible to write a unit test for the hang observed in this ticket, but I've enhanced SocketCloser to have both before & after Runnables and have added tests for that.

I've also tried out @Bill's recent enablement of "@NotNull" in SocketCloser.  Check it out.

### For all changes:

### Note:


geode-core/src/integrationTest/java/org/apache/geode/internal/net/SocketCloserIntegrationTest.java
```diff
@@ -113,26 +113,34 @@ public void testAsync() {
...
+  public void testOpenSocketCloser() {
+    final AtomicBoolean beforeSocketCloseRunnableWasCalled = new AtomicBoolean();
+    final AtomicBoolean afterSocketCloseRunnableWasCalled = new AtomicBoolean();
 
-    Socket s = createClosableSocket();
-    s.close();
-    this.socketCloser.asyncClose(s, "A", () -> runnableCalled.set(true));
-    await().until(() -> !runnableCalled.get());
+    final Socket closableSocket = createClosableSocket();
+    this.socketCloser.asyncClose(closableSocket, "A",
```
Bill, line 121: no `this` needed

geode-core/src/integrationTest/java/org/apache/geode/internal/net/SocketCloserIntegrationTest.java
```diff
@@ -113,26 +113,34 @@ public void testAsync() {
...
-    final AtomicBoolean runnableCalled = new AtomicBoolean();
+    final AtomicBoolean beforeSocketCloseRunnableWasCalled = new AtomicBoolean();
+    final AtomicBoolean afterSocketCloseRunnableWasCalled = new AtomicBoolean();
 
     final Socket closableSocket = createClosableSocket();
     this.socketCloser.close();
-    this.socketCloser.asyncClose(closableSocket, "A", () -> runnableCalled.set(true));
-    await()
-        .until(() -> runnableCalled.get() && closableSocket.isClosed());
+    this.socketCloser.asyncClose(closableSocket, "A",
```
Bill, line 139: no `this` needed

geode-core/src/main/java/org/apache/geode/distributed/internal/LonerDistributionManager.java
```diff
@@ -1054,14 +1055,12 @@ public boolean awaitTermination(long timeout, TimeUnit unit) throws InterruptedE
...
     public <T> Future<T> submit(final Callable<T> task) {
-      Exception ex = null;
-      T result = null;
       try {
-        result = task.call();
+        T result = task.call();
+        return new CompletedFuture<T>(result);
       } catch (Exception e) {
-        ex = e;
+        return new CompletedFuture<T>(e);
```
Bill, line 1062: cleaner!

geode-core/src/main/java/org/apache/geode/internal/util/CompletedFuture.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.concurrent.TimeoutException;
+
+/**
+ * CompletedFuture is a Future that reports that it has finished.
+ * It's intended to be used as the initial value of a Future variable
+ * instead of using null.
+ *
+ * @param <T>
+ */
+public class CompletedFuture<T> implements Future<T> {
```
Bill, line 29: Rather than our own class, why not use `CompletableFuture.completedFuture(value)`?

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/CacheClientProxy.java
```diff
@@ -954,7 +954,9 @@ private boolean closeSocket() {
     String remoteHostAddress = this._remoteHostAddress;
     if (this._socketClosed.compareAndSet(false, true) && remoteHostAddress != null) {
       // Only one thread is expected to close the socket
-      this._cacheClientNotifier.getSocketCloser().asyncClose(this._socket, remoteHostAddress, null);
+      this._cacheClientNotifier.getSocketCloser().asyncClose(this._socket, remoteHostAddress,
+          () -> {
+          });
```
Bill, line 959: no `this`

Bill (commented):
Made a few requests here. I know it's a bit early. Please consider.

geode-core/src/integrationTest/java/org/apache/geode/internal/net/SocketCloserIntegrationTest.java
bschuchardt: done

geode-core/src/integrationTest/java/org/apache/geode/internal/net/SocketCloserIntegrationTest.java
bschuchardt: done

geode-core/src/main/java/org/apache/geode/internal/util/CompletedFuture.java
bschuchardt: done - thanks for that

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/CacheClientProxy.java
bschuchardt: done

kirklund (approved):

agingade (approved):
The area of code is more advanced to me. Based on my understanding and code walk through it looks fine to me. But any other expert comments are much appreciated.

geode-core/src/main/java/org/apache/geode/internal/net/SocketCloser.java
```diff
@@ -163,60 +164,72 @@ private Future asyncExecute(String address, Runnable runnableToExecute) {
...
    * closed then a normal synchronous close is done.
    *
    * @param socket the socket to close
    * @param address identifies who the socket is connected to
-   * @param extra an optional Runnable with stuff to execute before the socket is closed
+   * @param runBeforeClose a Runnable with stuff to execute before the socket is closed
    */
-  public void asyncClose(final Socket socket, final String address, final Runnable extra) {
-    if (socket == null || socket.isClosed()) {
+  @NotNull
```
Bill, line 176: I don't think it makes sense to add the `@NotNull` annotation to a `void` method, since the method isn't returning anything at all.

geode-core/src/main/java/org/apache/geode/internal/net/SocketCloser.java
```diff
@@ -163,60 +164,72 @@ private Future asyncExecute(String address, Runnable runnableToExecute) {
...
+   * Depending on how the SocketCloser is configured (see ASYNC_CLOSE_WAIT_MILLISECONDS)
+   * this method may block for a certain amount of time. If it is called after the SocketCloser is
+   * closed then a normal synchronous close is done.
+   *
+   * @param socket the socket to close
+   * @param address identifies who the socket is connected to
+   * @param runBeforeClose a Runnable with stuff to execute before the socket is closed
+   * @param runAfterClose a Runnable with stuff to execute after the socket is closed
+   */
+  @NotNull
```
Bill, line 193: ditto

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ServerConnection.java
```diff
@@ -1412,11 +1412,19 @@ public boolean cleanup() {
       getAcceptor().decClientServerConnectionCount();
     }
 
-    try {
-      theSocket.close();
-    } catch (Exception ignored) {
-    }
+    if (!theSocket.isClosed()) {
+      final String closerName =
+          communicationMode.isWAN() ? "WANSocketCloser" : "CacheServerSocketCloser";
```
Bill, line 1417: `closerName` is being passed as the `address` parameter of `asyncClose()`. The doc on that method says:

```
   * @param address identifies who the socket is connected to
```

Other calls to `asyncClose()` send an address. If there is a reason we don't want to send an address here, would you mind adding a comment to explain it?

geode-core/src/main/java/org/apache/geode/distributed/internal/LonerDistributionManager.java
```diff
@@ -1054,14 +1055,14 @@ public boolean awaitTermination(long timeout, TimeUnit unit) throws InterruptedE
 
     @Override
     public <T> Future<T> submit(final Callable<T> task) {
-      Exception ex = null;
-      T result = null;
+      CompletableFuture future = new CompletableFuture<>();
```
Bill, line 1058: This new code looks like an improvement. But the method is on `DummyExecutor`: question: is this class used? Because I don't see any static usages of the class.

Can we remove `DummyExecutor` entirely?

Bill (changes requested):
I saw a few things.

geode-core/src/main/java/org/apache/geode/distributed/internal/LonerDistributionManager.java
bschuchardt: yes!  done!

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ServerConnection.java
bschuchardt: comments added.  I don't want cache shutdown to use a different executor for each remote address.  There could be hundreds of them.

geode-core/src/main/java/org/apache/geode/internal/net/SocketCloser.java
bschuchardt: done

geode-core/src/main/java/org/apache/geode/internal/net/SocketCloser.java
bschuchardt: done

Bill (approved):
Approve!

geode-core/src/main/java/org/apache/geode/internal/net/SocketCloser.java
```diff
@@ -163,60 +164,73 @@ private Future asyncExecute(String address, Runnable runnableToExecute) {
...
-   * @param address identifies who the socket is connected to
-   * @param extra an optional Runnable with stuff to execute before the socket is closed
+   * @param address key used to determine which executor to use. Usually the address of a peer or
+   *        client
+   * @param runBeforeClose a Runnable with stuff to execute before the socket is closed
+   * @param runAfterClose a Runnable with stuff to execute after the socket is closed
    */
-  public void asyncClose(final Socket socket, final String address, final Runnable extra) {
-    if (socket == null || socket.isClosed()) {
+  public void asyncClose(@NotNull final Socket socket, @NotNull final String address,
```
echobravopapa, line 194: NotNull 🚀

geode-core/src/main/java/org/apache/geode/internal/net/SocketCloser.java
```diff
@@ -163,60 +164,73 @@ private Future asyncExecute(String address, Runnable runnableToExecute) {
...
       } finally {
         closedLock.unlock();
       }
-      if (submittedTask != null) {
+      if (!doItInline) {
         waitForFutureTaskWithTimeout(submittedTask);
+        return;
       }
     } catch (RejectedExecutionException | OutOfMemoryError ignore) {
       // If we can't start a thread to close the socket just do it inline.
```
echobravopapa, line 228: maybe delete these comments now ...

echobravopapa (approved):
some possibly dead comment cleanup, else its good stuf

PR #5990, opened 2021-02-02
GEODE-8893: adding a sleep in test script for the server to start com…

jinmeiliao:
…pletely.


onichols-pivotal (approved):
of course some awaitility condition would be even better, but this seems like an improvement for the majority of cases

PR #5992, opened 2021-02-02
GEODE-8906: capture additional crash artifacts (fix typo)

onichols-pivotal:


DonalEvans (approved):

dickcav (approved):

jdeppe-pivotal (commented):
Ah - I was looking for this today. If you're open to it can we capture any logs found in `${HOME}/.gradle`. Recently we've seen some Stress test failing with Gradle Workers falling over and I suspect memory issues. I think the logs would help diagnose.

PR #6003, opened 2021-02-03
GEODE-8374: Remove no-op of unsupported system property.

echobravopapa:

### For all changes:

### Note:


bschuchardt (approved):
less code is good code

amitchrs8:
HI Dave,
          I had raised this jira and the intention of this jira was to remove the upper bound of this range of timeout which is set to 12437ms and make it User configuratble. I mean user should be able to configure it using member-timeout or  if you want to provide any other config. The issue is in big network , some components might take more time(more than 12437ms) to reconnect to cluster during membership change. Since this  is set to a small value 12437ms so those members get thrown out of cluster. I looked into the code change and doc change, and I see the upper limit is still there. This won't solve the problem for which this jira was raised.
       Can you please take a relook into this issue.
Thanks
Amit

PR #6007, opened 2021-02-04
GEODE-7864: Clean up InternalResourceManager

DonalEvans:
- Remove duplicate fine-level logging from deliverEventFromRemote()
 - Clean up raw use of parameterized classes
 - Clean up other static analyzer warnings

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/control/InternalResourceManager.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/control/InternalResourceManager.java
@@ -224,9 +224,9 @@ public class InternalResourceManager implements ResourceManager {
   public void deliverEventFromRemote(final ResourceEvent event) {
     assert !event.isLocal();
 
-    if (cache.getLogger().fineEnabled()) {
-      cache.getLogger()
-          .fine("New remote event to deliver for member " + event.getMember() + ": event=" + event);
+    if (logger.isTraceEnabled()) {
+      logger.trace(
+          "New remote event to deliver for member " + event.getMember() + ": event=" + event);
     }
 
     runWithNotifyExecutor(() -> deliverLocalEvent(event));
@@ -267,8 +267,7 @@ public class InternalResourceManager implements ResourceManager {
       notifyExecutor.execute(runnable);
     } catch (RejectedExecutionException ignore) {
       if (!closed) {
-        cache.getLogger()
-            .warning("No memory events will be delivered because of RejectedExecutionException");
+        logger.warn("No memory events will be delivered because of RejectedExecutionException");
       }
     }
   }
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/control/InternalResourceManager.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/control/InternalResourceManager.java
@@ -62,8 +62,6 @@ import org.apache.geode.util.internal.GeodeGlossary;
 
 /**
  * Implementation of ResourceManager with additional internal-only methods.
- * <p>
- * TODO: cleanup raw and typed collections
  */
 public class InternalResourceManager implements ResourceManager {
   private static final Logger logger = LogService.getLogger();
@@ -83,7 +81,7 @@ public class InternalResourceManager implements ResourceManager {
     }
   }
 
-  private Map<ResourceType, Set<ResourceListener>> listeners = new HashMap<>();
+  private final Map<ResourceType, Set<ResourceListener<?>>> listeners = new HashMap<>();
 
   private final ScheduledExecutorService scheduledExecutor;
   private final ExecutorService notifyExecutor;
@@ -104,7 +102,7 @@ public class InternalResourceManager implements ResourceManager {
 
   private final ResourceManagerStats stats;
   private final ResourceAdvisor resourceAdvisor;
-  private boolean closed = true;
+  private boolean closed;
 
   private final Map<ResourceType, ResourceMonitor> resourceMonitors;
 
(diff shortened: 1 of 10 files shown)


geode-core/src/main/java/org/apache/geode/internal/cache/control/InternalResourceManager.java
```diff
@@ -125,99 +123,99 @@ public static InternalResourceManager createResourceManager(final InternalCache
...
     for (ResourceType resourceType : new ResourceType[] {ResourceType.HEAP_MEMORY,
         ResourceType.OFFHEAP_MEMORY}) {
-      Set<ResourceListener> emptySet = new CopyOnWriteArraySet<ResourceListener>();
-      this.listeners.put(resourceType, emptySet);
+      Set<ResourceListener<?>> emptySet = new CopyOnWriteArraySet<>();
+      listeners.put(resourceType, emptySet);
     }
 
-    this.closed = false;
+    closed = false;
```
dschneider-pivotal, line 164: You got rid of "closed" being initialized to true so you should probably get rid of this line also. I think the old logic was: "start it at true and then at the end of the constructor set it to false (we must have had some code that allowed "this" to escape the constructor). We then set it to true in closed. But now its initial value is false so it does not make sense to reset it to false at the end of the constructor. We do pass "this" into the tempMonitors but those are just owned by this object so I don't see any real escape of "this".

DonalEvans: Good catch, thanks. It seems that the only time we check the value of `closed` is to determine whether or not to log a warning after catching an exception in `runWithNotifyExecutor()`, so even in the event that the wrong value was used, the impact would be negligible.

geode-core/src/main/java/org/apache/geode/internal/cache/control/InternalResourceManager.java
```diff
@@ -228,41 +226,29 @@ public void removeResourceListener(final ResourceType resourceType,
   public void deliverEventFromRemote(final ResourceEvent event) {
     assert !event.isLocal();
```
aimiebell, line 225: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Use of identifier that is a keyword in later versions of Java (NM_FUTURE_KEYWORD_USED_AS_IDENTIFIER)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-use-of-identifier-that-is-a-keyword-in-later-versions-of-java-nm-future-keyword-used-as-identifier).
The identifier is a word that is reserved as a keyword in later versions of Java, and your code will need to be changed in order to compile it in later versions of Java.

geode-core/src/main/java/org/apache/geode/internal/cache/control/InternalResourceManager.java
```diff
@@ -125,99 +123,97 @@ public static InternalResourceManager createResourceManager(final InternalCache
...
     // Initialize the listener sets so that it only needs to be done once
     for (ResourceType resourceType : new ResourceType[] {ResourceType.HEAP_MEMORY,
         ResourceType.OFFHEAP_MEMORY}) {
-      Set<ResourceListener> emptySet = new CopyOnWriteArraySet<ResourceListener>();
-      this.listeners.put(resourceType, emptySet);
+      Set<ResourceListener<?>> emptySet = new CopyOnWriteArraySet<>();
+      listeners.put(resourceType, emptySet);
     }
-
-    this.closed = false;
```
mhansonp, line 166: Why did this go away? It seems necessary off the cuff..

mhansonp (changes requested):

geode-core/src/main/java/org/apache/geode/internal/cache/control/InternalResourceManager.java
DonalEvans: See the comment from Darrel above. We were previously initializing `closed` to true, then setting it to false at the end of the constructor. Now we simply initialize it to false and don't have to re-set it to false in the constructor.

geode-core/src/main/java/org/apache/geode/internal/cache/control/InternalResourceManager.java
```diff
@@ -228,41 +224,29 @@ public void removeResourceListener(final ResourceType resourceType,
   public void deliverEventFromRemote(final ResourceEvent event) {
     assert !event.isLocal();
 
-    if (this.cache.getLogger().fineEnabled()) {
-      this.cache.getLogger()
+    if (cache.getLogger().fineEnabled()) {
```
kirklund, line 227: I would go ahead and replace all usages of the old proprietary LogWriter API (cache.getLogger() returns LogWriter) with Log4J2 Logger.

kirklund (approved):
Approved, but please consider updating from LogWriter to Logger whenever you change a class, especially for cleanup.

geode-core/src/main/java/org/apache/geode/internal/cache/control/InternalResourceManager.java
DonalEvans: Will do!

geode-core/src/main/java/org/apache/geode/internal/cache/control/InternalResourceManager.java
mhansonp: ok

mhansonp (approved):

nabarunnag (approved):
LGTM

PR #6008, opened 2021-02-05
GEODE-8918: fix replication in multisite systems

mkevo:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/AbstractRegion.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/AbstractRegion.java
@@ -741,8 +741,8 @@ public abstract class AbstractRegion implements InternalRegion, AttributesMutato
         // This is for all regions except pdx Region
         if (!isPdxTypesRegion) {
           // Make sure we are distributing to only those senders whose id
-          // is available on this region
-          if (allGatewaySenderIds.contains(sender.getId())) {
+          // is available on this region and whose state is running
+          if (allGatewaySenderIds.contains(sender.getId()) && sender.isRunning()) {
             allRemoteDSIds.add(sender.getRemoteDSId());
           }
         } else { // this else is for PDX region
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/AbstractRegion.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/AbstractRegion.java
@@ -742,7 +742,7 @@ public abstract class AbstractRegion implements InternalRegion, AttributesMutato
         if (!isPdxTypesRegion) {
           // Make sure we are distributing to only those senders whose id
           // is available on this region and whose state is running
-          if (allGatewaySenderIds.contains(sender.getId()) && sender.isRunning()) {
+          if (hasRunningGatewaySender(allGatewaySenders, sender)) {
             allRemoteDSIds.add(sender.getRemoteDSId());
           }
         } else { // this else is for PDX region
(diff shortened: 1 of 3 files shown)


DonalEvans (approved):

geode-core/src/main/java/org/apache/geode/internal/cache/AbstractRegion.java
```diff
@@ -741,8 +741,8 @@ public boolean isPdxTypesRegion() {
         // This is for all regions except pdx Region
         if (!isPdxTypesRegion) {
           // Make sure we are distributing to only those senders whose id
-          // is available on this region
-          if (allGatewaySenderIds.contains(sender.getId())) {
+          // is available on this region and whose state is running
+          if (allGatewaySenderIds.contains(sender.getId()) && sender.isRunning()) {
```
kirklund, line 745: Please consider writing a unit test for any code that only has integration/distributed test coverage. The following example is using a basic technique from Working Effectively with Legacy Code.

For example, you could create a new method called:
```
  @VisibleForTesting
  static boolean hasRunningGatewaySender(Set<GatewaySender> senders, GatewaySender sender) {
    return senders.contains(sender) && sender.isRunning();
  }
```
And then you can add unit tests for this new method to `AbstractRegionTest`:
```
import java.util.Collections;
import org.apache.geode.cache.wan.GatewaySender;

  @Test
  public void hasRunningGatewaySender_returnsFalse_ifSendersIsEmpty() {
    GatewaySender sender = mock(GatewaySender.class);

    boolean value = AbstractRegion.hasRunningGatewaySender(Collections.emptySet(), sender);

    assertThat(value).isFalse();
  }
```
You can also write it as a non-static method, but the unit tests will require a bit more setUp.

kirklund (changes requested):

geode-core/src/main/java/org/apache/geode/internal/cache/AbstractRegion.java
mkevo: Done!

geode-core/src/main/java/org/apache/geode/internal/cache/AbstractRegion.java
```diff
@@ -741,8 +741,8 @@ public boolean isPdxTypesRegion() {
...
             allRemoteDSIds.add(sender.getRemoteDSId());
           }
         } else { // this else is for PDX region
```
nabarunnag, line 748: Do we also need a check if sender is running for PDX region

geode-core/src/main/java/org/apache/geode/internal/cache/AbstractRegion.java
nabarunnag, line 745: can we use the new method hasRunningGatewaySender

nabarunnag (dismissed):

kirklund (approved):
Thank you for adding new unit and dunit tests!

geode-core/src/main/java/org/apache/geode/internal/cache/AbstractRegion.java
mkevo: I tested it and it works without checking if sender is running. All entries will be delivered to all remote sites for pdx regions.

geode-core/src/main/java/org/apache/geode/internal/cache/AbstractRegion.java
mkevo: Done!

mkevo:
@nabarunnag , can you re-review it please?

mkevo:
@gesterzhou, @boglesby, @nabarunnag can you please review this PR?

gesterzhou (approved):

PR #6009, opened 2021-02-05
GEODE-8917: Revise gfsh export stack-traces docs

karensmolermiller:


davebarnes97 (commented):
The change in syntax implies a change in semantics that is not explained.
If the `file` param is no longer required, what happens when it is not specified?

boglesby (approved):

davebarnes97 (approved):

PR #6011, opened 2021-02-06
GEODE-8924: Add VM restart tests for Redis and Spring sessions

jdeppe-pivotal:
- Use Spring repositories to perform fetch and save operations.

### For all changes:

### Note:

### DIFF
--- a/geode-redis/src/distributedTest/java/org/apache/geode/redis/session/SessionsAndCrashesDUnitTest.java
+++ b/geode-redis/src/distributedTest/java/org/apache/geode/redis/session/SessionsAndCrashesDUnitTest.java
@@ -27,6 +27,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 
+import org.assertj.core.api.SoftAssertions;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.BeforeClass;
@@ -100,10 +101,11 @@ public class SessionsAndCrashesDUnitTest {
   @Before
   @SuppressWarnings("unchecked")
   public void setup() {
+    // Focus our connections on server2 and server3 since these are the ones going to be
+    // restarted.
     String[] args = new String[] {
-        "" + redisPorts[2],
         "" + redisPorts[1],
-        "" + redisPorts[0]};
+        "" + redisPorts[2]};
 
     springContext = SpringApplication.run(RedisSpringTestApplication.class, args);
     sessionRepository = springContext.getBean(SessionRepository.class);
@@ -124,8 +126,8 @@ public class SessionsAndCrashesDUnitTest {
     AtomicBoolean running = new AtomicBoolean(true);
     AtomicReference<String> phase = new AtomicReference<>("STARTUP");
 
-    Future<Integer> future1 = executor.submit(() -> sessionUpdater(0, running, phase));
-    Future<Integer> future2 = executor.submit(() -> sessionUpdater(1, running, phase));
+    Future<Integer> future1 = executor.submit(() -> sessionUpdater(1, running, phase));
+    Future<Integer> future2 = executor.submit(() -> sessionUpdater(2, running, phase));
 
     GeodeAwaitility.await().during(1, TimeUnit.SECONDS).until(() -> true);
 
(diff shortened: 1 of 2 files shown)


geode-redis/src/distributedTest/java/org/apache/geode/redis/session/SessionsAndCrashesDUnitTest.java
```diff
@@ -0,0 +1,236 @@
...
+    phase.set("CRASH 4 REBALANCING");
+    rebalanceAllRegions(server3);
+
+    phase.set("FINISHING");
+    GeodeAwaitility.await().during(10, TimeUnit.SECONDS).until(() -> true);
+
+    running.set(false);
+    future1.get();
+    future2.get();
+  }
```
nonbinaryprogrammer, line 162: Could you validate that the values on the session are correct, or at least not garbage? 
This could theoretically pass if the session updater threads were all stalled for the ten seconds the test is running for. We should confirm that all the attributes we expect attr-#-[0->NUM_SESSIONS] exist.
Even better would be to validate that the count at the end of the value is greater than or equal to the modCount at the end of the attribute name. This would indicate that we made it through more than one iteration and no data was lost in the rebalancing.

geode-redis/src/distributedTest/java/org/apache/geode/redis/session/SessionsAndCrashesDUnitTest.java
```diff
@@ -0,0 +1,236 @@
...
+    Throwable latestException = null;
+
+    for (int i = 0; i < 10; i++) {
+      try {
+        sessionRepository.save(session);
+        return;
+      } catch (RedisSystemException rex) {
+        latestException = rex;
+        try {
+          Thread.sleep(2000);
```
nonbinaryprogrammer, line 210: why is this needed and can it be less than 2 seconds? If not then you might need to make the GeodeAwaitility.during() (line 157) at least 20 seconds, because it's possible that we could spend that long trying to save the session and end up not going through all 10 attempts.

geode-redis/src/distributedTest/java/org/apache/geode/redis/session/SessionsAndCrashesDUnitTest.java
```diff
@@ -0,0 +1,236 @@
...
+    future1.get();
+    future2.get();
+  }
+
+  private Integer sessionUpdater(int index, AtomicBoolean running, AtomicReference<String> phase) {
+    int count = 0;
+    while (running.get()) {
+      int modCount = count % NUM_SESSIONS;
+      String sessionId = sessionIds.get(modCount);
+      Session session = findSession(sessionId);
```
nonbinaryprogrammer, line 169: between this and the call to saveSession(), the sessionUpdater method could take over 40 seconds. the wait in the test for the finish phase only waits 10 seconds, so a call to sessionUpdater could potentially be cut short.

nonbinaryprogrammer (changes requested):

geode-redis/src/distributedTest/java/org/apache/geode/redis/session/SessionsAndCrashesDUnitTest.java
```diff
@@ -0,0 +1,236 @@
...
+
+  @ClassRule
+  public static ExecutorServiceRule executor = new ExecutorServiceRule();
+
+  private static final int JEDIS_TIMEOUT =
+      Math.toIntExact(GeodeAwaitility.getTimeout().toMillis());
+  private static final int NUM_SESSIONS = 100;
+  private static final List<String> sessionIds = new ArrayList<>(NUM_SESSIONS);
+  private static MemberVM locator;
+  private static MemberVM server1;
```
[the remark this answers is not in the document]
jdeppe-pivotal: I want to have different servers to stop/start. `server1` is the lead member and thus has a much higher quorum weight. if it were to shut down then the remaining members would shut down since they would lose quorum. So instead, leave the lead member alone and use the other 2.

geode-redis/src/distributedTest/java/org/apache/geode/redis/session/SessionsAndCrashesDUnitTest.java
```diff
@@ -0,0 +1,236 @@
...
+        .withConnectionToLocator(locatorPort));
+  }
+
+  @Before
+  @SuppressWarnings("unchecked")
+  public void setup() {
+    String[] args = new String[] {
+        "" + redisPorts[2],
+        "" + redisPorts[1],
+        "" + redisPorts[0]};
```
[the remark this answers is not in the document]
jdeppe-pivotal: I've removed the connection to server1 so that the connections are only established to the 2 servers that are to be shut down and restarted.

geode-redis/src/distributedTest/java/org/apache/geode/redis/session/SessionsAndCrashesDUnitTest.java
jdeppe-pivotal: Added some validation.

geode-redis/src/distributedTest/java/org/apache/geode/redis/session/SessionsAndCrashesDUnitTest.java
jdeppe-pivotal: Your comment prompted me to see if we can get away with just a single retry and it looks like the test is quite happy doing that.

geode-redis/src/distributedTest/java/org/apache/geode/redis/session/SessionsAndCrashesDUnitTest.java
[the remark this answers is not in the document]
nonbinaryprogrammer: 👍

nonbinaryprogrammer (approved):
looks good, thanks for addressing those comments!

PR #6013, opened 2021-02-09
GEODE-8768_1: locator discovery improvement

mivanac:

### For all changes:

### Note:


geode-core/src/main/java/org/apache/geode/internal/admin/remote/DistributionLocatorId.java
```diff
@@ -305,7 +354,11 @@ public String toString() {
    * Indicates whether some other object is "equal to" this one.
    *
    * @param other the reference object with which to compare.
-   * @return true if this object is the same as the obj argument; false otherwise.
+   *
+   *        If member name is defined in both objects, and both objects have same member name,
```
bschuchardt, line 358: this javadoc change no longer applies since the equals() method no longer checks member names

geode-wan/src/main/java/org/apache/geode/cache/client/internal/locator/wan/LocatorHelper.java
```diff
@@ -46,13 +46,42 @@ public static boolean addLocator(int distributedSystemId, DistributionLocatorId
     Set<DistributionLocatorId> existingValue =
         allLocatorsInfo.putIfAbsent(distributedSystemId, locatorsSet);
     if (existingValue != null) {
-      if (!existingValue.contains(locator)) {
+      if (locator.getMemberName() != null) {
```
bschuchardt, line 49: Mario, I think you still need need to check for an empty string and treat that the same as a null.  Maybe it would be easier to have the default value be an empty string in DistributionLocatorId.  Then you could check for an empty string and not need a null check.

geode-wan/src/main/java/org/apache/geode/cache/client/internal/locator/wan/LocatorHelper.java
```diff
@@ -46,13 +46,42 @@ public static boolean addLocator(int distributedSystemId, DistributionLocatorId
...
+        DistributionLocatorId tempLocator = null;
```
bschuchardt, line 50: This could use a comment about why the code is looking for an old ID with the same member name and removing it.  A better variable name than tempLocator would help, too.

geode-wan/src/main/java/org/apache/geode/cache/client/internal/locator/wan/LocatorHelper.java
```diff
@@ -46,13 +46,42 @@ public static boolean addLocator(int distributedSystemId, DistributionLocatorId
...
+        for (DistributionLocatorId locElement : existingValue) {
```
bschuchardt, line 51: This code is pretty much duplicated below.  It might read better if you extracted it to a method like getLocatorWithSameMemberName(locatorId, locatorIdCollection)

geode-wan/src/main/java/org/apache/geode/cache/client/internal/locator/wan/LocatorHelper.java
```diff
@@ -100,6 +129,17 @@ public static boolean addExchangedLocators(Map<Integer, Set<DistributionLocatorI
           if (!localLocators.equals(entry.getValue())) {
             entry.getValue().removeAll(localLocators);
             for (DistributionLocatorId locator : entry.getValue()) {
+              if (locator.getMemberName() != null && !localLocators.isEmpty()) {
+                boolean locatorExist = false;
+                for (DistributionLocatorId locId : localLocators) {
```
bschuchardt, line 134: duplicate code - also this could use stream()->findFirst() instead of iterating

geode-wan/src/main/java/org/apache/geode/cache/client/internal/locator/wan/WanLocatorDiscovererImpl.java
```diff
@@ -65,11 +66,19 @@ private void exchangeLocalLocators(int port, DistributionConfigImpl config,
       LocatorMembershipListener locatorListener, final String hostnameForClients) {
     String localLocator = config.getStartLocator();
     DistributionLocatorId locatorId = null;
+
+    InternalDistributedSystem system = InternalDistributedSystem.getConnectedInstance();
```
bschuchardt, line 70: same comment - we want to stop using this static method to find an InternalDistributedSystem, and you have a DistributionConfig parameter that provides the memberName that you need.

geode-wan/src/main/java/org/apache/geode/cache/client/internal/locator/wan/LocatorMembershipListenerImpl.java
```diff
@@ -108,11 +109,18 @@ public void locatorJoined(final int distributedSystemId, final DistributionLocat
       final DistributionLocatorId sourceLocator) {
     // DistributionLocatorId for local locator.
     DistributionLocatorId localLocatorId;
+
+    InternalDistributedSystem system = InternalDistributedSystem.getConnectedInstance();
```
bschuchardt, line 113: we're trying to get rid of the use of static methods like this.  The DistributionConfig argument passed to this method should have the memberName in it.  You should use that instead.

geode-wan/src/main/java/org/apache/geode/cache/client/internal/locator/wan/WanLocatorDiscovererImpl.java
```diff
@@ -113,10 +122,17 @@ private RemoteLocatorJoinRequest buildRemoteDSJoinRequest(int port, Distribution
       final String hostnameForClients) {
     String localLocator = config.getStartLocator();
     DistributionLocatorId locatorId = null;
+
+    InternalDistributedSystem system = InternalDistributedSystem.getConnectedInstance();
```
bschuchardt, line 126: same comment about this method and the config parameter

bschuchardt (changes requested):
This approach should work.  I have some comments on the new code.

geode-wan/src/main/java/org/apache/geode/cache/client/internal/locator/wan/LocatorHelper.java
mivanac: Hi, I added this part to cover empty string

  public String getMemberName() {
    if (this.membername != null && this.membername == "")
      return null;

    return this.membername;
  }

, but can also add to have empty string instead of null.

geode-wan/src/main/java/org/apache/geode/cache/client/internal/locator/wan/WanLocatorDiscovererImpl.java
mivanac: Thanks for comments. Updated.

geode-wan/src/main/java/org/apache/geode/cache/client/internal/locator/wan/LocatorHelper.java
```diff
@@ -100,6 +129,14 @@ public static boolean addExchangedLocators(Map<Integer, Set<DistributionLocatorI
           if (!localLocators.equals(entry.getValue())) {
             entry.getValue().removeAll(localLocators);
             for (DistributionLocatorId locator : entry.getValue()) {
+              if (!locator.getMemberName().equals(DistributionConfig.DEFAULT_NAME)
```
bschuchardt, line 132: would you please add a comment here as to why it's okay to ignore the other map's locator ID when the membername matches one that's already in the local map?

bschuchardt (changes requested):
Asking for a few more changes.
Also, Lynn's test is still failing with these changes but I haven't been able to figure out why.

geode-wan/src/main/java/org/apache/geode/cache/client/internal/locator/wan/LocatorHelper.java
```diff
@@ -46,13 +47,42 @@ public static boolean addLocator(int distributedSystemId, DistributionLocatorId
...
+            Set<String> alllocators = allServerLocatorsInfo.get(distributedSystemId);
+            alllocators.remove(existingLocator.toString());
+            existingValue.add(locator);
+            addServerLocator(distributedSystemId, locatorListener, locator);
+            locatorListener.locatorJoined(distributedSystemId, locator, sourceLocator);
+            return true;
+          }
+          return false;
+        }
         existingValue.add(locator);
```
bschuchardt, line 72: This code looks wrong to me.  The old code would not invoke addLocator if "existingValue.contains(locator)" returned false.  You still need this check if the check "if (existingLocator != null)" returns false.

if (existingLocator != null) {
  your new code
} else if (!existingValue.contains(locator)) {
  existingValue.add(locator);
  addServerLocator(...);
  locatorListener.locatorJoined(...);
}

The way you've got it coded a new locator with the same host:port but a different memberName will be added as a new locator even if existingValue contains its ID (because equals ignores the member name).  This will cause new threads to be spawned to communicate with that locator, whereas the old code wouldn't do that.

bschuchardt (changes requested):

kirklund (changes requested):
Some of the classes you've altered currently only have Distributed Test coverage. Please create Unit Test for each class. You don't have to provide complete unit test coverage for the class, but try to at least cover the code you've changed within a unit test (using Mockito, etc). Some of the classes might end up needing better constructor injection or other changes just to make it (unit) testable. Feel free to make constructors or methods package-private with annotation `VisibleForTesting` so that a test within the same package can poke around as needed.

geode-wan/src/main/java/org/apache/geode/cache/client/internal/locator/wan/LocatorHelper.java
bschuchardt: This is something that Lynn's test does.  It appends the PID of a locator to its name, so when her test restarts a locator it has the same host:port but a different name.  I didn't think that was bad on first review but it probably accounts for the high number of threads created and subsequent OutOfMemoryError when new threads couldn't be created.

mivanac: Main logic is that member name is main parameter, and if it does not exist, then we apply existing logic.
This is the way we can determine that existing locator has change host[port].

bschuchardt: Yes, but this has changed the behavior for when the host:port has _not_ changed but the member name _has_ changed.  I think that's what's breaking Lynn's test.

geode-wan/src/main/java/org/apache/geode/cache/client/internal/locator/wan/LocatorHelper.java
```diff
@@ -46,13 +47,56 @@ public static boolean addLocator(int distributedSystemId, DistributionLocatorId
...
+            addServerLocator(distributedSystemId, locatorListener, locator);
+            locatorListener.locatorJoined(distributedSystemId, locator, sourceLocator);
+            return true;
+          }
+          return false;
+        } else if (!existingValue.contains(locator)) {
+          existingValue.add(locator);
+          addServerLocator(distributedSystemId, locatorListener, locator);
+          locatorListener.locatorJoined(distributedSystemId, locator, sourceLocator);
+        } else {
```
bschuchardt, line 75: This block needs a _return true_

geode-wan/src/main/java/org/apache/geode/cache/client/internal/locator/wan/LocatorHelper.java
bschuchardt: The new changes still seem to have the same problem.  If I add a new locator with the same host:port but a **different** member _name_ it will still spawn a thread to invoke locatorJoined in the _else_ clause after "} else if (!existingValue.contains(locator)) {".
Your PR's intent is to scrub an old ID for a new locator that has the **same** member _name_ regardless of its host:port.  If that check fails, meaning there is no ID with the same member _name_, we should then check to see if there is already an ID with the host:port.  If a match is found we should not invoke locatorJoined on the listener. 
 Every invocation of that method spawns a thread and that's causing problems.
Adding some unit tests for these required behaviors would help ensure that things work correctly.

mivanac: I think that test case is not realistic. Is it realistic that locator at restart will change its name. But if this is the case, and also if we are now storing member name, then in that case we should store new member name. I can change this, so test case passes, but this will be inconsistent behavior.

bschuchardt: hmm, I see what you mean Mario.  The membername field is new, but to be consistent we should invoke the listener in case it cares about that new field.  Hold off on making that change while I look into Lynn's failures further.  Thanks for your patience.

bschuchardt (approved):
Looks good Mario

geode-core/src/test/java/org/apache/geode/internal/admin/remote/DistributionLocatorIdJUnitTest.java
```diff
@@ -46,4 +48,25 @@ public void testEquals() throws UnknownHostException {
...
 
+  @Test
+  public void testEquals_and_DetailCompare() throws UnknownHostException {
+    DistributionLocatorId dLI1 = new DistributionLocatorId(40404, "127.0.0.1", null);
+    DistributionLocatorId dLI2 =
+        new DistributionLocatorId(40404, "127.0.0.1", "127.0.1.0", "member2");;
+    DistributionLocatorId dLI3 = new DistributionLocatorId(40404, "127.0.0.1", null, "member3");
+    DistributionLocatorId dLI4 = new DistributionLocatorId(dLI3.marshal());
+
+    assertTrue(dLI1.equals(dLI2));
```
kirklund, line 59: Can you please update the test to use AssertJ instead of JUnit Asserts? If these assertions ever fail, they will provide much better details for fixing the code if they're changed to:
```
assertThat(dLI1).isEqualTo(dLI2);
```

geode-core/src/test/java/org/apache/geode/internal/admin/remote/DistributionLocatorIdJUnitTest.java
```diff
@@ -46,4 +48,25 @@ public void testEquals() throws UnknownHostException {
 
   }
 
+  @Test
+  public void testEquals_and_DetailCompare() throws UnknownHostException {
+    DistributionLocatorId dLI1 = new DistributionLocatorId(40404, "127.0.0.1", null);
```
kirklund, line 53: Let's change these variable names to something more description than an abbreviation:
```
DistributionLocatorId distributionLocatorId1 = new DistributionLocatorId(40404, "127.0.0.1", null);
```

geode-wan/src/test/java/org/apache/geode/cache/client/internal/locator/wan/WanLocatorDiscovererTest.java
```diff
@@ -0,0 +1,95 @@
...
+    when(config.getName()).thenReturn("loc1");
+    when(config.getBindAddress()).thenReturn("127.0.0.1");
+    when(config.getDistributedSystemId()).thenReturn(1);
+    when(config.getLocators()).thenReturn(locators);
+    when(config.getRemoteLocators()).thenReturn("");
+
+    WanLocatorDiscovererImpl test_wan = new WanLocatorDiscovererImpl();
+    test_wan.discover(40401, config, locatorMembershipListener, null);
+
+    assertThat(allLocatorsInfo.get(1).size()).isEqualTo(4);
```
kirklund, line 89: These size assertions provide much better failure messages if you change them to:
```
assertThat(allLocatorsInfo.get(1)).hasSize(4);
```
Then if the size is wrong, it'll actually list out all the elements which helps debug what went wrong.

geode-wan/src/test/java/org/apache/geode/cache/client/internal/locator/wan/LocatorHelperTest.java
```diff
@@ -0,0 +1,184 @@
...
+    Set<DistributionLocatorId> responseSet = new HashSet<>();
+    responseSet.add(loc1);
+    responseSet.add(loc2);
+    responseSet.add(loc3);
+    responseSet.add(loc4);
+    responseLocatorsInfo.put(1, responseSet);
+    when(locatorMembershipListener.getAllLocatorsInfo()).thenReturn(emptyLocatorList);
+    when(locatorMembershipListener.getAllServerLocatorsInfo()).thenReturn(emptyServerLocatorList);
+
+    assertTrue(
```
kirklund, line 173: Let's use keep the assertions more consistent and use AssertJ here too:
```
assertThat(LocatorHelper.addExchangedLocators(responseLocatorsInfo, locatorMembershipListener)).isTrue();
```

geode-core/src/test/java/org/apache/geode/internal/admin/remote/DistributionLocatorIdJUnitTest.java
```diff
@@ -46,4 +48,25 @@ public void testEquals() throws UnknownHostException {
...
+    DistributionLocatorId dLI2 =
+        new DistributionLocatorId(40404, "127.0.0.1", "127.0.1.0", "member2");;
```
kirklund, line 55: Please delete the extra semi-colon at the end of this line.

kirklund (changes requested):
Thank you for adding new and updating both unit tests and system tests! I've requested a few changes in line, including one with an example of changing the size assertion to a different syntax that will provide better failure message.

Overall, you should try to always use `hasSize()` for the size assertions:
```
assertThat(listener.getAllLocatorsInfo()).hasSize(2);
```
...instead of:
```
assertThat(listener.getAllLocatorsInfo().size()).isEqualTo(2);
```

geode-core/src/test/java/org/apache/geode/internal/admin/remote/DistributionLocatorIdJUnitTest.java
```diff
@@ -46,4 +48,25 @@ public void testEquals() throws UnknownHostException {
...
+    assertTrue(dLI1.equals(dLI3));
+    assertTrue(dLI1.equals(dLI4));
+
+    assertTrue(dLI1.getMemberName().equals(DistributionConfig.DEFAULT_NAME));
+    assertTrue(dLI2.getMemberName().equals("member2"));
+    assertTrue(dLI3.getMemberName().equals("member3"));
+    assertTrue(dLI4.getMemberName().equals(DistributionConfig.DEFAULT_NAME));
+
+    assertTrue(dLI1.detailCompare(dLI3));
+    assertFalse(dLI2.detailCompare(dLI4));
```
kirklund, line 69: Please use:
```
assertThat(dLI2.detailCompare(dLI4)).isFalse();
```
The failure message isn't any better but it's best if we consistently use AssertJ for all of the assertions.

geode-core/src/test/java/org/apache/geode/internal/admin/remote/DistributionLocatorIdJUnitTest.java
```diff
@@ -46,4 +48,25 @@ public void testEquals() throws UnknownHostException {
...
+    DistributionLocatorId dLI2 =
+        new DistributionLocatorId(40404, "127.0.0.1", "127.0.1.0", "member2");;
+    DistributionLocatorId dLI3 = new DistributionLocatorId(40404, "127.0.0.1", null, "member3");
+    DistributionLocatorId dLI4 = new DistributionLocatorId(dLI3.marshal());
+
+    assertTrue(dLI1.equals(dLI2));
+    assertTrue(dLI1.equals(dLI3));
+    assertTrue(dLI1.equals(dLI4));
+
+    assertTrue(dLI1.getMemberName().equals(DistributionConfig.DEFAULT_NAME));
```
kirklund, line 63: AssertJ flavor would be:
```
assertThat(dLI1.getMemberName()).isEqualTo(DistributionConfig.DEFAULT_NAME);
```
This one will provide a better failure message.

geode-wan/src/test/java/org/apache/geode/cache/client/internal/locator/wan/LocatorHelperTest.java
```diff
@@ -0,0 +1,184 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.apache.geode.cache.client.internal.locator.wan;
+
+import static org.apache.geode.internal.Assert.assertTrue;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.Assert.assertFalse;
```
kirklund, line 19: There's a great IntelliJ plugin that will convert JUnit Assertions to AssertJ called **Assertions2Assert2**. It's not perfect but it can help a lot.

geode-core/src/test/java/org/apache/geode/internal/admin/remote/DistributionLocatorIdJUnitTest.java
demery-pivotal: You can improve the failure message by adding an `as()` that describes the meaning of the boolean. Something like:
```
assertThat(dLI2.detailCompare(dLI4))
    .as("Distribution locator IDs 2 and 4 are equal")
    .isFalse();
```

kirklund (approved):

mivanac:
Hi @jdeppe-pivotal, @jinmeiliao , @Bill , @gesterzhou and @boglesby can you review this PR.

jinmeiliao (approved):

mivanac:
Hi @jdeppe-pivotal, @Bill , @gesterzhou and @boglesby,
just a reminder, could you review this PR.

mivanac:
HI @gesterzhou and @boglesby,
can you review this PR. It is only waiting on your review.

gesterzhou (approved):

PR #6019, opened 2021-02-09
GEODE-8933: update INFO to return maxmemory field

ringles:
Updates the Geode Redis INFO command to return the new field "maxmemory". (Also fixes an issue with the mem_fragmentation_ratio field, which should always be greater than zero.)

### DIFF
--- a/geode-redis/src/main/java/org/apache/geode/redis/internal/executor/server/InfoExecutor.java
+++ b/geode-redis/src/main/java/org/apache/geode/redis/internal/executor/server/InfoExecutor.java
@@ -28,6 +28,8 @@ import org.apache.geode.redis.internal.statistics.RedisStats;
 
 public class InfoExecutor extends AbstractExecutor {
 
+  private static final Long ONE_MEGABYTE = 1024 * 1024 * 1024L;
+
   private DecimalFormat decimalFormat = new DecimalFormat("0.00");
 
   @Override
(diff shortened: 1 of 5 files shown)


geode-redis/src/main/java/org/apache/geode/redis/internal/executor/server/InfoExecutor.java
```diff
@@ -28,6 +28,8 @@
 
 public class InfoExecutor extends AbstractExecutor {
 
+  private static final Long ONE_MEGABYTE = 1024 * 1024 * 1024L;
```
sabbey37, line 31: I think 1 MB is `1024L x 1024L`, or 1048576 bytes... not sure if I'm missing something here (assuming the intention is to convert MB to bytes)?

sabbey37 (changes requested):

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/server/InfoExecutor.java
ringles: Good point. Until we can return real values from Geode it's not critical but we'll fix that ASAP!

sabbey37 (approved):

sabbey37 (commented):
I think the addition of the field and tests are fine.  I can't think of a better way to test it.  Would definitely be good for @jdeppe-pivotal  or @dschneider-pivotal  to review and make sure `getLocalMaxMemory()` is the right thing to return there since they know more about memory in Geode.

jdeppe-pivotal (approved):

PR #6021, opened 2021-02-10
GEODE-8906: capture gradle artifacts too

onichols-pivotal:
* fix previous PR #5992 to pull in the gradle logs _before_ the rsync


jdeppe-pivotal (approved):
Thanks for fixing this!

dickcav (approved):

PR #6022, opened 2021-02-10
GEODE-8934: add incrby to supported commands

nonbinaryprogrammer:

### For all changes:

### Note:


geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/string/AbstractStringIntegrationTest.java
```diff
@@ -26,141 +28,190 @@
...
+          jedis1.incrBy(key, increment);
+        },
+        (i) -> {
+          int increment = ThreadLocalRandom.current().nextInt(-50, 50);
+          expectedValue.addAndGet(increment);
+          jedis2.incrBy(key, increment);
+        }).run();
+
+    assertThat(Integer.parseInt(jedis1.get(key))).isEqualTo(expectedValue.get());
   }
```
sabbey37, line 216: There's actually an `AbstractIncrByIntegrationTest` file. Some of these tests could be merged into that file. Others are duplicates. A test that isn't in either file: testing the max and min 64 bit signed integer value (9,223,372,036,854,775,807 and -9223372036854775808, respectively).  We should test both trying to increment a key that is set above/below these values (so setting a key to`9223372036854775808` then trying to incrby 1 (which results in `ERR value is not an integer or out of range`)) as well as incrementing a key that is right at the max/min value   (so setting a key to `9223372036854775807` then trying to incrby 1 (which results in `ERR increment or decrement would overflow`))

sabbey37 (changes requested):

geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/string/AbstractStringIntegrationTest.java
nonbinaryprogrammer: I just found that file as well. good idea to condense them into one. And I like that test so I'll definitely implement that

geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/string/AbstractIncrByIntegrationTest.java
```diff
@@ -18,79 +18,124 @@
...
+
+  @Test
+  public void testIncrBy_IncrementingMaxValueThrowsError() {
+    String key = "key";
+    Long increment = Long.MAX_VALUE / 2;
+
+    jedis1.set(key, String.valueOf(Long.MAX_VALUE));
+    assertThatThrownBy(() -> jedis1.incrBy(key, increment))
+        .hasMessageContaining("ERR increment or decrement would overflow");
+  }
```
sabbey37, line 117: Thanks for adding this test and merging the files!  We get a different error when incrementing a key that is set above the max long value values (so setting a key to 9223372036854775808 then trying to incrby 1 (which results in `ERR value is not an integer or out of range`). Could we add a test for that as well?

sabbey37 (changes requested):

geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/string/AbstractStringIntegrationTest.java
```diff
@@ -33,125 +33,128 @@
...
+
   private static final int REDIS_CLIENT_TIMEOUT =
       Math.toIntExact(GeodeAwaitility.getTimeout().toMillis());
 
   @Before
   public void setUp() {
-    jedis = new Jedis("localhost", getPort(), REDIS_CLIENT_TIMEOUT);
+    jedis1 = new Jedis("localhost", getPort(), REDIS_CLIENT_TIMEOUT);
+    jedis2 = new Jedis("localhost", getPort(), REDIS_CLIENT_TIMEOUT);
   }
```
sabbey37, line 46: I should've mentioned this previously, but you could revert the changes in this file, and we don't need `jedis2` anymore.  You could always do that in a different PR though.

sabbey37 (approved):

PR #6025, opened 2021-02-10
GEODE-8936: add INCR to list of supported commands

nonbinaryprogrammer:

### For all changes:

### Note:


geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/string/AbstractStringIntegrationTest.java
```diff
@@ -128,6 +130,62 @@ public void testStrlen_withFloatData() {
...
+
+    jedis.set(key, "0");
+
+    new ConcurrentLoopingThreads(NUM_ITERATIONS,
+        (i) -> jedis.incr(key),
+        (i) -> jedis2.incr(key)).run();
+
+    assertThat(Integer.parseInt(jedis.get(key))).isEqualTo(expectedValue);
+  }
+
```
sabbey37, line 188: There's actually an `AbstractIncrIntegrationTest` file.  Some of these could be merged into that file. Others are duplicates.  A couple tests that aren't in either file: performing `INCR` on a key that doesn't exist, and we test the max 64 bit signed integer value, but it would be interesting to do an incr with the min value as well (min value is -9,223,372,036,854,775,808, so maybe doing an INCR on -9,223,372,036,854,775,809 to see if we'll return the minimum value, and doing an INCR on -9,223,372,036,854,775,810 to make sure we still get the OVERFLOW error when an INCR is performed).

sabbey37 (changes requested):
Something else to think about is that we changed `INCRBYFLOAT` to parse a BigDecimal rather than a long.  I'm not sure if we want to change all INCR/DECR related commands to parse BigIntegers or BigDecimals?  (If we did, then we wouldn't have to worry about the max/min 64 bit signed integer value tests... it would be different from Redis, but we are already different due to the change in `INCRBYFLOAT`)

**Update:** Jens told me BigDecimals were used in `INCRBYFLOAT` so that floating point would be more accurate... and keeping long for other INCR/DECR ops will be faster performance-wise.  So unless we think there's a strong use case (maybe page view count?) or selling point for being able to INCR/DECR larger values than Redis, maybe that change doesn't make sense.

nonbinaryprogrammer:
> Something else to think about is that we changed `INCRBYFLOAT` to parse a BigDecimal rather than a long. I'm not sure if we want to change all INCR/DECR related commands to parse BigIntegers or BigDecimals? (If we did, then we wouldn't have to worry about the max/min 64 bit signed integer value tests... it would be different from Redis, but we are already different due to the change in `INCRBYFLOAT`)
> 
> **Update:** Jens told me BigDecimals were used in `INCRBYFLOAT` so that floating point would be more accurate... and keeping long for other INCR/DECR ops will be faster performance-wise. So unless we think there's a strong use case (maybe page view count?) or selling point for being able to INCR/DECR larger values than Redis, maybe that change doesn't make sense.

I think that in this case we should favor performance over consistency with another command, especially since that command isn't consistent with native redis. thank you for looking into that though

geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/string/AbstractStringIntegrationTest.java
nonbinaryprogrammer: Good ideas, I'll do that :)

geode-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/string/AbstractStringIntegrationTest.java
```diff
@@ -35,6 +35,7 @@
   private Jedis jedis;
   private static final int REDIS_CLIENT_TIMEOUT =
       Math.toIntExact(GeodeAwaitility.getTimeout().toMillis());
+  private static final int NUM_ITERATIONS = 1000;
```
sabbey37, line 38: I don't think this variable is used in this class any longer.

sabbey37 (changes requested):
Similary to INCRBY, could we also add a test that increments a key that is set above the max long value value (so setting a key to 9223372036854775808 then trying to `INCR` that key (which results in `ERR value is not an integer or out of range`))?

sabbey37 (approved):

PR #6036, opened 2021-02-16
GEODE-8856: Persist gateway-sender startup-action

jvarenina:
New startup-action parameter with values "stop", "pause" and "start" is now persisted
during the runtime when following commands are issued:

- pause gateway-sender   --> startup-action="pause"
- stop gateway-sender      --> startup-action="stop"
- start gateway-sender      --> startup-action="start"
- resume gateway-sender --> startup-action="start"

The startup-action parameter is persisted within cluster configuration when above gfsh
commands are executed successfully on at least one of the servers. Parameter is
not updated and persisted when commands are executed per member as
cluster configuration is not persisted in that case.

New startup-action parameter will now inter-work with manual-start in a following way:

- If manual-start="true" and startup-action parameter is missing, then gateway sender
   will require manual start (same as before).
- if manual-start is not set (or "false") and startup-action parameter is missing, then
   gateway sender will be started automatically (same as before).
- If parameter startup-action is available in cluster configuration at startup,
   then gateway-sender will try to reach that state regardless of manual-start
   parameter value.

The manual-start is also improved in order to fully comply to above requirement to
start gateway sender in stopped state.

Current problem with manual-start parameter:

Currently, when manual-start is configured to be "true", then colocated persistent
parallel gateway sender queue region and buckets are not recovered after server
is restarted. Because of that the main persistent region that is colocated with
gateway sender queue region cannot reach online status.

Solution to manual-start problem implemented in this commit:

When manual-start parameter is "true" or gateway sender startup-action is "stop", then
persistent parallel gateway-sender queues will now be recovered (if needed) from
persistent storage during startup of the server. Queues will be recovered by using the
existing mechanism that is used when gateway sender is automatically recovered
(manual-start==false) after server is restarted. In this case parallel gateway sender
queue persistent region and buckets are recovered (if needed) right after the main
persistent region and buckets are recovered.

Additionally in above case, parallel gateway sender will now reach the same state
that it has when first started and then stopped by using gfsh command. In that state
parallel gateway sender queue buckets remain on the servers, but dispatcher threads are
stopped and non of the events are stored in queues.


onichols-pivotal (approved):

kirklund (changes requested):
Please keep watch for opportunities to create (or update) unit tests, especially if you're altering logic behavior in a class with no unit tests. High-end system tests like the DUnit tests are great but they're only 1/3 of the testing that can and should be done. 

At a minimum, I would hope to see new unit test coverage for changes you're making to CacheXmlParser, PauseGatewaySenderCommand, ResumeGatewaySenderCommand, StartGatewaySenderCommand, StopGatewaySenderCommand. It's not just about the test coverage at a unit test level with mocking, but the positive design impact this has on forcing developers to use constructor dependency injection, good abstraction with interfaces, and following good object oriented design principles (https://en.wikipedia.org/wiki/SOLID).

Your contribution looks great otherwise and fits in well with the existing WAN code. Unfortunately, that code and its tests is not a good example to follow especially for unit testing and SOLID principles, especially as we try to refactor the codebase to improve code quality.

To be clear, I'm asking for at least some updated or new unit tests for any changes more interesting than adding a field with getter/setter. Even then, you could test to see if the class properly retains the correct value of the new field when serialized and deserialized.

jvarenina:
@kirklund thanks for the comments! I will add new unit tests to this PR.

rhoughton-pivot (approved):

mhansonp:
Restarted concourse-ci/DistributedTestOpenJDK11 run.

geode-core/src/test/java/org/apache/geode/internal/cache/xmlcache/CacheXmlParserTest.java
```diff
@@ -0,0 +1,175 @@
...
+@RunWith(PowerMockRunner.class)
+@PrepareForTest(WANServiceProvider.class)
+public class CacheXmlParserTest {
+
+  @Mock
+  private GatewaySenderFactory gatewaySenderFactory;
+
+  @Before
+  public void setUp() {
+    PowerMockito.mockStatic(WANServiceProvider.class);
```
kirklund, line 58: The Geode community decided on the dev-list to get rid of and avoid `PowerMockito`. It causes too many problems including corrupting the unit testing JVM for later unit tests. It also discourages us from refactoring classes to be better structured and easier to unit test.

The main Jira ticket is [https://issues.apache.org/jira/browse/GEODE-6143](https://issues.apache.org/jira/browse/GEODE-6143), but there are additional sub-tickets you can search Jira for.

kirklund (changes requested):
Thanks for adding more tests! Unfortunately, the Geode community decided to disallow PowerMockito and rip it out. We are in the process of refactoring code that has unit tests using PowerMockito and changing to tests to use regular Mockito.

karensmolermiller (approved):
The updated doc file looks fine to me.

geode-core/src/test/java/org/apache/geode/internal/cache/xmlcache/CacheXmlParserTest.java
jvarenina: Thanks for the comment. I have removed PowerMock from the unit tests.

geode-core/src/main/java/org/apache/geode/internal/cache/wan/InternalGatewaySender.java
```diff
@@ -46,7 +46,7 @@
 
   void destroy(boolean initiator);
 
-  void setStartEventProcessorInPausedState();
+  void setStartEventProcessorInPausedState(boolean isPaused);
```
gesterzhou, line 49: so far this fix is a refactoring of current code to save the "paused" state into cache config. In that regard, this fix is good. 

However, since we now have many states now and all these different states will be saved and persisted into cache config, we should not always only persist "paused" state and the method name should also be generic.

jvarenina: Hi @gesterzhou ,

Thank you very much for the comment!

The state parameter is persisted within cluster configuration on locator at the reception of gfsh commands that update the state of gateway-sender (create, pause, stop and start gateway-sender commands). When member restarts and joins the cluster, then it will receive cluster configuration containing the latest value of state parameter, and will start gateway-sender accordingly. So this means that only locators will have the latest gateway-sender state that is configured with gfsh commands, and servers will have the state which was used at startup of server and gateway-sender creation.

The flag `setStartEventProcessorInPausedState` was actually used by` async-event-queue `code for following cases:

1. list async-event-queues command  -> the flag is used to determine `Created with paused event processing` parameter value in printout:
```

 Member  |   ID   | Batch Size | Persistent | Disk Store | Max Memory |                         Listener                         | Created with paused event processing | Currently Paused
-------- | ------ | ---------- | ---------- | ---------- | ---------- | -------------------------------------------------------- | ------------------------------------ | ----------------
server-1 | queue1 | 100        | false      | null       | 100        | org.apache.geode.internal.cache.wan.MyAsyncEventListener | false                                | false
server-2 | queue2 | 100        | false      | null       | 100        | org.apache.geode.internal.cache.wan.MyAsyncEventListener | false                                | false
```

2. When gateway-sender (used by async-event-queue) is started using interface GatewaySender.start() then this flag is checked, and if set, then gateway-sender is started in paused state.

In this PR, I have reused the "setStartEventProcessorInPausedState()" just to start gateway-sender in paused state as described in step 2. I have also reset the flag right after gateway-sender is started in paused state, so that this flag doesn't interfere with future start commands. This case is explained in commit message on following link https://github.com/apache/geode/pull/6036/commits/29afa4476c379a53d79f6919ad92b6314601040b

Not sure that I fully understood your request. Do you maybe propose that we change this flag `setStartEventProcessorInPausedState(bolean)` to e.g. `setCreateStateGatewaySender(GatewaySenderState state)` that will reflect the state that was used at gateway-sender creation?

I would be really grateful if you could share some more details.

geode-core/src/main/java/org/apache/geode/internal/cache/xmlcache/CacheXmlGenerator.java
```diff
@@ -1373,6 +1373,11 @@ private void generateGatewaySender(GatewaySender sender) throws SAXException {
     if (generateDefaults() || sender.getAlertThreshold() != GatewaySender.DEFAULT_ALERT_THRESHOLD)
       atts.addAttribute("", "", ALERT_THRESHOLD, "", String.valueOf(sender.getAlertThreshold()));
 
+    // gateway-sender state
+    if (sender.getState() != null) {
+      atts.addAttribute("", "", STATE, "", sender.getState().getState());
+    }
+
```
boglesby, line 1380: You probably need a unit test for this change. See CacheXmlGeode10DUnitTest.

geode-core/src/main/java/org/apache/geode/internal/cache/xmlcache/CacheXmlParser.java
```diff
@@ -642,6 +660,16 @@ private void startGatewaySender(Attributes atts) {
...
     }
 
+    // Gateway-sender state
+    String state = atts.getValue(STATE);
+    if (Objects.equals(state, GatewaySenderState.RUNNING.getState()) ||
+        Objects.equals(state, GatewaySenderState.STOPPED.getState()) ||
+        Objects.equals(state, GatewaySenderState.PAUSED.getState())) {
+      gatewaySenderFactory.setState(GatewaySenderState.valueOf(state.toUpperCase()));
+    } else {
+      gatewaySenderFactory.setState(null);
```
boglesby, line 670: You probably should throw an exception here if an invalid state is used

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/functions/GatewaySenderFunctionArgs.java
```diff
@@ -47,6 +47,7 @@
   private final List<String> gatewayEventFilters;
   private final List<String> gatewayTransportFilters;
   private final Boolean enforceThreadsConnectSameReceiver;
+  private final String state;
```
boglesby, line 50: The GatewaySenderFunctionArgs defines a state field. That object is only used by the CreateGatewaySenderCommand. The CreateGatewaySenderCommand doesn't have any way to set the state when a sender is created. Should it?

jvarenina: Thank you for the comment!

Currently, `CreateGatewaySenderCommand` can only manipulate with the initial state of gateway-sender by using deprecated `--manual-start` parameter (default value is false). This parameter does not update the new `state` parameter.

As you suggested, we could also add `--state` parameter in `create gateway-sender`. We could maybe then completely remove the depreciated `--manual-start` parameter, since `--state` will do the same thing (and even more considering "paused" state) as `--manual-start` parameter. As it is now, these two parameters have functional conflict which is currently resolved as described in RFC under "Solution" chapter: https://cwiki.apache.org/confluence/display/GEODE/Persist+gateway-sender+state+within+Cluster+Configuration?moved=true

What do you think? Do you see benefit in having --state parameter in create gateway-sender command?

geode-core/src/main/java/org/apache/geode/internal/cache/xmlcache/CacheXmlParser.java
jvarenina: Yes, you are right! I will throw `InternalGemFireException`, since same is done for other parameters (e.g. orderPolicy).

jvarenina: Yes, you are right. I will throw `InternalGemFireException`, since same is done for other parameters (e.g. orderPolicy).

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/functions/GatewaySenderFunctionArgs.java
boglesby: One of the problems with manual-start is that if it is set to true, it doesn't recover the sender's persistence files.

The same is also true of state=stopped.

This causes data regions to not be recovered in the parallel case since the data and queue regions are co-located.

A warning like this is displayed:
```
[warn 2021/04/01 09:32:47.877 PDT server-ln-1 <ColocationLogger for Trade> tid=0x3c] Persistent data recovery for region /Trade is prevented by offline colocated region
	/ny_PARALLEL_GATEWAY_SENDER_QUEUE
```
A test to cause this is:

1. Start a server with a persistent region and sender
2. Shutdown the server
3. Set manual-start=true (either export / cluster configuration or modify xml)
4. Restart the server

If we allow a persistent, parallel sender to be created with state=stopped or even save the stopped state when the sender shuts down, we'll have the same problem. I'm not sure we should allow either of those cases.

jvarenina: Thank you for pointing this out. I have reproduced the problem by following your steps. 

It seem that this PR is missing some cases with persistent (and also redundant) geateway-sender queues and region. I will create those cases and investigate this particular issue a little bit more, in order to see what can be done here.

jinmeiliao (approved):

davebarnes97 (approved):

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/functions/GatewaySenderFunctionArgs.java
jvarenina: Hi @boglesby ,

I have covered the scenario when persistent parallel gateway sender is started in "stopped" state in following commit: https://github.com/apache/geode/pull/6036/commits/6abcb7aa7a027187afe64037e2a78e011f126d29

I only followed the similar scenario when gateway-sender is started, and then stopped using gfsh commands. In that case all relevant threads that handle events are stopped (no data is enqueued), but collocated queue and data regions continue coexist and work. So in order to cover this problematic case, I only bring gateway-sender to the same state as it would be after it is started and then stopped by using gfsh commands (start and stop gateway-sender).

You can find more information in below TC's that verify the solution. I tried to describe every TC in comment block just before the each case, so it would be easier to review. Hope this will help.

- ParallelGatewaySenderOperationClusterConfigDUnitTest.java
- ParallelGatewaySenderOperationClusterConfigDUnitTest.java

I would be really grateful if you could share your opinion on this solution. Also, I just noticed that I have introduced the changes you proposed in you first two comments. I will introduce them in next commit.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
```diff
@@ -1189,11 +1189,11 @@ public void postCreateRegion() {
       for (GatewaySender sender : cache.getAllGatewaySenders()) {
         if (sender.isParallel() && allGatewaySenderIds.contains(sender.getId())) {
           /*
-           * get the ParallelGatewaySender to create the colocated partitioned region for this
+           * get the ParallelGatewaySender to create the collocated partitioned region for this
```
DonalEvans, line 1192: This should be "colocated" with one L.

jvarenina: OK, I will change that. Thanks for the feedback.

gesterzhou (approved):

geode-wan/src/main/java/org/apache/geode/cache/wan/internal/parallel/ParallelGatewaySenderImpl.java
```diff
@@ -56,6 +62,43 @@ public void start() {
...
+    this.getLifeCycleLock().writeLock().lock();
+    try {
+      if (eventProcessor != null) {
+        // Already recovered in stopped state
+        return;
+      }
+
+      Set<Region> targetRs = new HashSet<Region>();
+      for (InternalRegion pr : this.getCache().getApplicationRegions()) {
+        if (((LocalRegion) pr).getAllGatewaySenderIds().contains(this.getId())) {
```
albertogpz, line 76: The cast to `LocalRegion` is not necessary.

albertogpz (changes requested):
As a general comment, there is a lot of repeated code in the pause, resume, start and stop commands. I wonder if common functions could be used to reuse.
The same goes for the DUnit tests. Maybe you can have one file with generic test cases that can then be converted to specific ones for pause, resume, start and stop with parameters.
That way the code repetition would be much lower.

geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
```diff
@@ -305,8 +307,9 @@ public int compare(Region o1, Region o2) {
     queueEmptyLock = new StoppableReentrantLock(sender.getCancelCriterion());
     queueEmptyCondition = queueEmptyLock.newCondition();
 
-    // initialize the conflation thread pool if conflation is enabled
-    if (sender.isBatchConflationEnabled()) {
+    removalThread = new BatchRemovalThread(this.sender.getCache(), this);
```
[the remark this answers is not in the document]
jvarenina: 1. Yes, because I don't want to start batch removal processe in stopped state. Before this impact is introduced the removal thread processor was initialized and started as follows:

```
  /** Start the background batch removal thread. */
  public void start() {
    // at present, this won't be accessed by multiple threads,
    // still, it is safer approach to synchronize it
    synchronized (ParallelGatewaySenderQueue.class) {
      if (removalThread == null) {
        removalThread = new BatchRemovalThread(this.sender.getCache(), this);
        removalThread.start();
      }
    }
  }
```

Now I want to only initialize the BatchRemovalThread because some of its functions (check function sendQueueRemovalMesssageForDroppedEvent) are used while handling events in stopped state. There is no need to start batch removal processor thread in stopped state.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/parallel/ParallelWANPropagationLoopBackDUnitTest.java
```diff
@@ -410,9 +410,9 @@ public void unstartedSenderShouldNotAddReceivedEventsIntoTmpDropped() {
     vm3.invoke(() -> WANTestBase.validateRegionSize(getTestMethodName() + "_PR", 100));
     vm5.invoke(() -> WANTestBase.validateRegionSize(getTestMethodName() + "_PR", 100));
 
-    // verify tmpDroppedEvents should be 100 at site-ln, because the sender is not started yet
-    vm3.invoke(() -> WANTestBase.verifyTmpDroppedEventSize("ln", 100));
-    vm5.invoke(() -> WANTestBase.verifyTmpDroppedEventSize("ln", 100));
+    // verify tmpDroppedEvents should be 0 at site-ln, because the sender is not started yet
+    vm3.invoke(() -> WANTestBase.verifyTmpDroppedEventSize("ln", 0));
+    vm5.invoke(() -> WANTestBase.verifyTmpDroppedEventSize("ln", 0));
```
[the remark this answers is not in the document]
jvarenina: The thing is that I have changed design based behavior of manual-start parameter. In design base, when this parameter is set to "true" then server will not recover **persistent** parallel gateway-sender queue region at startup. Data region in this case would remain offline until collocated queue region is also recovered (queue regions will be recovered after gateway-sender is started with gfsh command). With this new solution, colocated queue region will be recovered in stopped state right after data region is recovered. This way data region can immediately reach online status regardless of gateway-sender status. I have decided to change this design base behavior because manual-start parameter is deprecated and not recommended to use in production. Also there are lots of test cases that are using` manual-start=true`, and this way I could verify that recovering in stopped state is working on many design base test cases. This impact related to manual-start is really small, and I can easily remove it. Should I return design base behavior in this case?

So to answer why it is now expected to have 0 temporary queued events? The reason is because gateway-sender will now reach stopped state at startup (queue buckets will be recovered). This is the same state as when running gateway sender is stopped with gfsh command. In that case events aren't stored in temporary dropped queue but instead registerEventDroppedInPrimaryQueue function is called:

```
private void recordDroppedEvent(EntryEventImpl event) {
  if (!this.isPrimary()) {
    return;
  }
  if (this.eventProcessor != null) {
    this.eventProcessor.registerEventDroppedInPrimaryQueue(event);
  } else {
    tmpDroppedEvents.add(event);
      logger.debug("added to tmpDroppedEvents event: {}", event);
    }
  }
    logger.debug("Returning without putting into the gateway sender queue:" + event);
  }
}
```

You have introduced above behavior in GEODE-8491. Please notice when gateway-sender is recovered in stopped state then eventProcessor is different then null.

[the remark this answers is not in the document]
albertogpz: I understand. Given that tmpDroppedEvents does not seem to be used in this case, maybe the comments in the test case should be updated and it would be great to have another test case in which tmpDroppedEvents is used.

geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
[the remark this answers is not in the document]
albertogpz: ok

albertogpz (commented):
Thanks for the effort. With the reuse, the lines of code have been reduced significantly.

geode-core/src/main/java/org/apache/geode/cache/wan/GatewaySender.java
```diff
@@ -258,6 +271,8 @@
    */
   boolean isRunning();
 
+
+
```
albertogpz, line 275: Remove these extra lines.

geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ConcurrentParallelGatewaySenderEventProcessor.java
```diff
@@ -212,6 +217,7 @@ public void run() {
         }
       }
     }
+
```
albertogpz, line 220: Remove this line, please.

geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderEventProcessor.java
```diff
@@ -78,13 +88,18 @@ protected void initializeMessageQueue(String id, boolean cleanQueues) {
...
-    queue.start();
+    if (!shouldOnlyRecoverQueues) {
+      queue.start();
+    }
     this.queue = queue;
 
-    if (queue.localSize() > 0) {
-      queue.notifyEventProcessorIfRequired();
+    if (!shouldOnlyRecoverQueues) {
+      if (((ParallelGatewaySenderQueue) queue).localSize() > 0) {
```
albertogpz, line 100: The two casts to `ParallelGatewaySenderQueue` are redundant.

geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySender.java
```diff
@@ -1230,6 +1239,22 @@ public void enqueueTempEvents() {
     }
   }
 
+  /**
+   * During sender is recovered in stopped state, if there are any cache operations while
+   * queue and event processor is being created then these events should be stored in
+   * tmpDroppedEvents temporary queue. Once event processor is created then queue will be
+   * drained and ParallelQueueRemovalMessage will be sent.
+   */
+  public void enqueueTempDroppedEvents() {
```
albertogpz, line 1248: This method could be better called. `processTempDroppedEvents()`

geode-core/src/main/java/org/apache/geode/internal/cache/ColocationHelper.java
```diff
@@ -181,7 +181,6 @@ public static boolean checkMembersColocation(PartitionedRegion partitionedRegion
         return false;
       }
     }
-
```
albertogpz, line 184: Why was this line removed?

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/parallel/ParallelWANPersistenceEnabledGatewaySenderDUnitTest.java
```diff
@@ -158,10 +158,18 @@ public void testPartitionedRegionWithPersistentGatewaySender() {
...
+    AsyncInvocation inv2 = vm3.invokeAsync(() -> WANTestBase
+        .createPersistentPartitionedRegion(getTestMethodName(), null, 1, 100, isOffHeap()));
+    try {
+      inv1.join();
+      inv2.join();
+
+    } catch (InterruptedException e) {
+      e.printStackTrace();
+      fail();
+    }
```
albertogpz, line 172: What is the reason for changing this to be done in parallel?

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/parallel/ParallelWANPersistenceEnabledGatewaySenderDUnitTest.java
```diff
@@ -1892,9 +1908,10 @@ public void testpersistentWanGateway_restartSender_expectAllEventsReceived_scena
     LogWriterUtils.getLogWriter().info("All the senders are now running...");
 
     // ----------------------------------------------------------------------------------------------------
-
-    vm2.invoke(() -> WANTestBase.validateRegionSize(getTestMethodName(), 3000));
-    vm3.invoke(() -> WANTestBase.validateRegionSize(getTestMethodName(), 3000));
+    vm7.invoke(() -> WANTestBase.doPuts(getTestMethodName(), 500));
+    vm4.invoke(() -> WANTestBase.validateRegionSize(getTestMethodName(), 3000));
+    vm2.invoke(() -> WANTestBase.validateRegionSize(getTestMethodName(), 500));
+    vm3.invoke(() -> WANTestBase.validateRegionSize(getTestMethodName(), 500));
```
albertogpz, line 1914: I do not understand why this case should behave differently now. I would expect that the legacy behavior is kept.
Why is it?

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/parallel/ParallelWANPropagationLoopBackDUnitTest.java
```diff
@@ -365,9 +365,9 @@ public void testParallelPropagationLoopBack3SitesNtoNTopologyPutFromOneDS() {
    * Site-NY: dsid=1: senderId="ln": vm3, vm6
    * NY site's sender's manual-start=true
    *
-   * Make sure the events are sent from LN to NY and will not be added into tmpDroppedEvents
-   * while normal events put from NY site can still be added to tmpDroppedEvents
-   * Start the sender, make sure the events in tmpDroppedEvents are sent to LN finally
+   * Verify that events aren't added to tmpDroppedEvents after gateway-sender is recovered in
+   * stopped state with manual-start=true, and that none of the events are sent to remote site
+   * after sender is started.
```
albertogpz, line 370: I do not see in this case that the sender is recovered in stopped state. The only thing I see is that it is started.
This test case might need some rethinking.

geode-wan/src/main/java/org/apache/geode/cache/wan/internal/parallel/ParallelGatewaySenderImpl.java
```diff
@@ -56,6 +62,44 @@ public void start() {
...
+  @Override
+  public void recoverInStoppedState() {
+    this.getLifeCycleLock().writeLock().lock();
+    try {
+      if (eventProcessor != null) {
+        // Already recovered in stopped state
+        return;
+      }
+
+      Set<Region> targetRs = new HashSet<Region>();
```
albertogpz, line 74: The `Region` type in the `HashSet` is not needed.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/wancommand/ClusterConfigStartStopPauseAndResumeGatewaySenderOperationDUnitTest.java
```diff
@@ -0,0 +1,716 @@
...
+    clientSite1.invoke(() -> checkDataAvailable(keysQueued));
+    clientSite1.invoke(() -> checkDataNotAvailable(keysNotQueued));
+  }
+
+  /**
+   * This test case verifies that the server during recovery of gateway-sender in stopped state
+   * stores events in tmpDroppedEvents queue, and handles them after sender has been recovered.
+   */
+  @Test
+  public void testSubscriptionQueueWanTrafficWhileServerIsRestarted() throws Exception {
```
albertogpz, line 433: Could we have a more meaningful name for this test?
I do not see the relation between the description and the name of the test case.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/wancommand/ClusterConfigStartStopPauseAndResumeGatewaySenderOperationDUnitTest.java
```diff
@@ -0,0 +1,716 @@
...
+    // Check that data is sent over the gateway-sender
+    server1Site2.invoke(() -> checkQueueSize("ln", 0));
+    clientSite1.invoke(() -> checkDataAvailable(keysQueued));
+    clientSite1.invoke(() -> checkDataNotAvailable(keysNotQueued));
+  }
+
+  /**
+   * This test case verifies that the server during recovery of gateway-sender in stopped state
+   * stores events in tmpDroppedEvents queue, and handles them after sender has been recovered.
+   */
```
albertogpz, line 431: Would it be possible to check that tmpDroppedEvents is not zero at some point of the test and then, at the end, it is zero?
Otherwise, I do not think that the test case does what it should.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/wancommand/ClusterConfigStartStopPauseAndResumeGatewaySenderOperationDUnitTest.java
```diff
@@ -0,0 +1,716 @@
...
+  }
+
+  public static void localQueueSize(String senderId) {
+    GatewaySender sender = getGatewaySender(senderId);
+    int totalSize = 0;
+    Set<RegionQueue> queues = ((AbstractGatewaySender) sender).getQueues();
+    if (queues != null) {
+      for (RegionQueue q : queues) {
+        ConcurrentParallelGatewaySenderQueue prQ = (ConcurrentParallelGatewaySenderQueue) q;
+        prQ.localSize(true);
```
albertogpz, line 494: This line should contain:
`totalSize += prQ.localSize(true);`

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/wancommand/ClusterConfigStartStopPauseAndResumeGatewaySenderOperationDUnitTest.java
```diff
@@ -0,0 +1,716 @@
...
+    await().untilAsserted(() -> assertEquals("Expected tmpDroppedEvents size: " + size
+        + " but actual size: " + ags.getTmpDroppedEventSize(), size, ags.getTmpDroppedEventSize()));
+  }
+
+  public static void testLocalQueueIsEmpty(String senderId) {
+    await()
+        .untilAsserted(() -> localQueueSize(senderId));
+  }
+
+  public static void localQueueSize(String senderId) {
```
albertogpz, line 487: I would change the name of this method to something like:
`assertThatLocalQueueSizeIsZero`

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/parallel/ParallelWANPersistenceEnabledGatewaySenderDUnitTest.java
kirklund: If you're going to use `AsyncInvocation` please add a type such as `Void`, use `await()`, give a useful name to the variable, remove the try-catch, and add any exceptions such as `InterruptedException` to the `throws` clause of the test:
```
public void testPartitionedRegionWithPersistentGatewaySender() throws InterruptedException {
  ...
  AsyncInvocation<Void> createPersistentPartitionedRegionAsync = vm2.invokeAsync(() -> WANTestBase
        .createPersistentPartitionedRegion(getTestMethodName(), null, 1, 100, isOffHeap()));
  ...
  createPersistentPartitionedRegionAsync.await();
}
```
As @albertogpz pointed out, there's really not enough value to doing to this for creation of a region.

I usually only do the above when it's really valuable to the test itself such as two member performing region operations. When I do that I name the `AsyncInvocation` something like this:
```
AsyncInvocation<Void> performPuts = ...
```

jvarenina:
The flaky test `AutoConnectionSourceDUnitTest.testClientDynamicallyDropsStoppedLocator `that is failing in `concourse-ci/distributed-test-openjdk8`  is not connected to this PR.

mhansonp (approved):

kirklund:
Next to each reviewer in the upper right is a circle of two arrows to "Re-request review". If you click that it pops back into our queue of PRs to review. I definitely recommend using that!

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
```diff
@@ -1193,10 +1195,15 @@ public void postCreateRegion() {
...
            * region.
            */
+          InternalGatewaySender senderImpl = (InternalGatewaySender) sender;
           if (sender.isRunning()) {
-            AbstractGatewaySender senderImpl = (AbstractGatewaySender) sender;
             ((ConcurrentParallelGatewaySenderQueue) senderImpl.getQueues()
                 .toArray(new RegionQueue[1])[0]).addShadowPartitionedRegionForUserPR(this);
+          } else if (senderImpl.getEventProcessor() == null) {
+            if (senderImpl.getStartupAction() == GatewaySenderStartupAction.STOP
+                || sender.isManualStart()) {
```
kirklund, line 1204: The conditionals of this else-if block are kind of convoluted. I recommend extracting it to a single method with a name that describes it well and returns a single boolean so that you can rewrite it as:
```
} else if (isManualStartupRequired()) {
  senderImpl.recoverInStoppedState();
}
```
I'm not sure if `isManualStartupRequired()` is a good name... I'm just guessing here.

geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderEventProcessor.java
```diff
@@ -78,13 +88,18 @@ protected void initializeMessageQueue(String id, boolean cleanQueues) {
...
+      queue.start();
+    }
     this.queue = queue;
 
-    if (queue.localSize() > 0) {
-      queue.notifyEventProcessorIfRequired();
+    if (!recoverQueuesOnly) {
+      if (queue.localSize() > 0) {
+        queue.notifyEventProcessorIfRequired();
+      }
```
kirklund, line 102: Try to avoid nesting especially when it's unnecessary.
```
if (!recoverQueuesOnly && queue.localSize() > 0) {
  queue.notifyEventProcessorIfRequired();
}
```
Or you could extract that conditional to a small private method with a descriptive name to make it easier to understand. It's hard to figure out what `!recoverQueuesOnly && queue.localSize() > 0` really means unless I study the rest of the code in this class.

geode-core/src/main/java/org/apache/geode/internal/cache/xmlcache/CacheXmlParser.java
```diff
@@ -642,6 +661,24 @@ private void startGatewaySender(Attributes atts) {
...
+    String id = atts.getValue(ID);
+    // Gateway-sender startup action
+    InternalGatewaySenderFactory internalGatewaySenderFactory =
+        (InternalGatewaySenderFactory) gatewaySenderFactory;
+    String startupAction = atts.getValue(STARTUP_ACTION);
+    if (startupAction == null) {
+      internalGatewaySenderFactory.setStartupAction(GatewaySenderStartupAction.NONE);
+    } else if (Objects.equals(startupAction, GatewaySenderStartupAction.START.getAction()) ||
+        Objects.equals(startupAction, GatewaySenderStartupAction.STOP.getAction()) ||
+        Objects.equals(startupAction, GatewaySenderStartupAction.PAUSE.getAction())) {
```
kirklund, line 673: You should probably just the `name()` which is built-in to Java enums:
```
} else if (GatewaySenderStartupAction.START.name().equals(startupAction)) ||
```

geode-core/src/test/java/org/apache/geode/internal/cache/xmlcache/CacheXmlParserTest.java
```diff
@@ -0,0 +1,196 @@
...
+import static org.apache.geode.internal.cache.xmlcache.CacheXml.IS_SERVER;
+import static org.apache.geode.internal.cache.xmlcache.CacheXml.LOCK_LEASE;
+import static org.apache.geode.internal.cache.xmlcache.CacheXml.LOCK_TIMEOUT;
+import static org.apache.geode.internal.cache.xmlcache.CacheXml.MESSAGE_SYNC_INTERVAL;
+import static org.apache.geode.internal.cache.xmlcache.CacheXml.REMOTE_DISTRIBUTED_SYSTEM_ID;
+import static org.apache.geode.internal.cache.xmlcache.CacheXml.SEARCH_TIMEOUT;
+import static org.apache.geode.internal.cache.xmlcache.CacheXml.STARTUP_ACTION;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.Assert.assertThrows;
+import static org.junit.Assert.assertTrue;
```
kirklund, line 30: Please use `org.assertj.core.api.Assertions.assertThat` instead of the JUnit Assertions.

geode-core/src/test/java/org/apache/geode/internal/cache/xmlcache/CacheXmlParserTest.java
```diff
@@ -0,0 +1,196 @@
...
+    when(cacheCreation.createGatewaySenderFactory()).thenReturn(gatewaySenderFactory);
+
+    CacheXmlParser parser = new CacheXmlParser(cacheCreation);
+    Exception exception = assertThrows(InternalGemFireException.class,
+        () -> parser.startElement("http://geode.apache.org/schema/cache", "gateway-sender",
+            "gateway-sender",
+            attrs));
+    assertTrue(exception.getMessage()
+        .contains(
+            "An invalid startup-action value (pausede) was configured for gateway sender sender1"));
```
kirklund, line 193: Here's the way you do this with AssertJ:
```
```
```
Throwable thrown = catchThrowable(() -> 
  parser.startElement("http://geode.apache.org/schema/cache", "gateway-sender", "gateway-sender", attrs));

assertThat(thrown)
  .isInstanceOf(InternalGemFireException.class)
  .hasMessage("An invalid startup-action value (pausede) was configured for gateway sender sender1");
```

geode-core/src/test/java/org/apache/geode/internal/cache/xmlcache/CacheXmlParserTest.java
```diff
@@ -0,0 +1,196 @@
...
+  private CacheCreation cacheCreation;
+
+  @Before
+  public void setUp() {
+    cacheCreation = mock(CacheCreation.class);
+  }
+
+  @Test
+  public void testStartCacheParametersSet() throws SAXException {
+    AttributesImpl attrs = new AttributesImpl();
```
kirklund, line 59: Please avoid abbreviations as much as possible and just spell out `attributes`.

Also, if the APIs that use `attrs` support the interface `Attributes` then please always declare everything as the interfaces instead of the impl.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/parallel/ParallelWANPersistenceEnabledGatewaySenderDUnitTest.java
```diff
@@ -972,15 +968,24 @@ public void testPersistentPRWithPersistentGatewaySender_Restart_DoOps() {
...
+
+    try {
+      inv1.join();
+      inv2.join();
+      inv3.join();
+      inv4.join();
+    } catch (InterruptedException e) {
+      e.printStackTrace();
+      fail();
+    }
```
kirklund, line 988: Please don't catch Exceptions in tests unless you need to assert its type or message. You just need to make sure the test method has `throws InterruptedException` (which it does) and then delete the try-catch blocks.

Please make sure that you define the `AsyncInvocation` types as `AsyncInvocation<Void>` unless they have a return value. Use descriptive names for these variables and use `await()` instead of `join()`. The javadocs on `AsyncInvocation` have some good guidelines.

```
AsyncInvocation<Void> createRegionInVm4 = vm4.invokeAsync(() -> WANTestBase
        .createPersistentPartitionedRegion(getTestMethodName(), "ln", 1, 100, isOffHeap()));

createRegionInVm4.await();
```

However, in this case you should NOT use `invokeAsync` and `AsyncInvocations` unless this test has a very specific need to create the regions asynchronously. It's more common to use this API for spawning remote threads that are doing region operations asynchronously. So it's probably better to just synchronously create these regions:
```
import static org.apache.geode.internal.cache.wan.WANTestBase.createPersistentPartitionedRegion;

vm4.invoke(() -> createPersistentPartitionedRegion(getTestMethodName(), "ln", 1, 100, isOffHeap()));
vm5.invoke(() -> createPersistentPartitionedRegion(getTestMethodName(), "ln", 1, 100, isOffHeap()));
```

geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySender.java
```diff
@@ -185,6 +187,19 @@
...
+   * Activate this hook by setting system property <code>ENABLE_TEST_HOOK_TEMP_DROPPED_EVENTS</code>
+   */
+  private volatile ConcurrentLinkedQueue<EntryEventImpl> testHookTempDroppedEvents;
+
+  /**
+   * Only used for testing purpose. This property enables test hook which collects all
+   * wan replication events that are dropped by parallel gateway senders.
+   */
+  private static final boolean ENABLE_TEST_HOOK_TEMP_DROPPED_EVENTS =
+      Boolean.getBoolean("enable-test-hook-temp-dropped-events");
```
kirklund, line 202: Please try to think of some ways to test dropped events without having to resort to a test hook. Test hooks in the product code should always be the very last resort. It would be better to refactor the class, wrap it with a decorator, introduce a factory that can wrap all instances with an optional decorator, etc.

In general, I recommend very detailed white-box testing at the unit test level. The system test level (includes dunit/distributed tests) should be lean more towards higher level black-block testing.

kirklund (changes requested):

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/parallel/ParallelWANPersistenceEnabledGatewaySenderDUnitTest.java
jvarenina: Hi @kirklund ,

In this particular case I have to use asynchronous creation of region, otherwise test fails with following log:

`[vm4] [info 2021/11/09 09:53:11.278 CET  <PersistentBucketRecoverer for region ln_PARALLEL_GATEWAY_SENDER_QUEUE> tid=0xbe] Region /testPersistentPRWithPersistentGatewaySender_Restart_DoOps (and any colocated sub-regions) has potentially stale data.  Buckets [1, 4, 5, 9, 10, 12, 13, 14, 16, 20, 24, 25, 26, 27, 30, 31, 33, 34, 36, 38, 41, 43, 44, 45, 46, 49, 50, 54, 55, 56, 60, 61, 62, 63, 65, 68, 69, 71, 73, 76, 79, 81, 83, 85, 88, 92, 94, 95, 97, 98] are waiting for another offline member to recover the latest data. My persistent id is:`

This happens because now gateway sender queue region is recovered right after main region is created. I will update the test as you suggested to use `AsyncInvocation<Void>` and `await()`.

You have more information related to this in following comment: https://github.com/apache/geode/pull/6036#discussion_r642882417

Hope this is OK to you?

onichols-pivotal:

jvarenina:
Hi @onichols-pivotal ,

Sorry, I missed to apply some of the comments. I will fully focus on that now and apply them as soon as possible. I would like to merge this PR as it's solution is complete (comments are just related to tests), and I also think that this is good feature to have in Geode, especially if you run Geode in automated environments like Kuberentes. I know it is hard to review this PR as lots of things have changed through time, but I open to any suggestions from you side to make it more easier for you to review it. Sorry for any inconvenience.

PR #6037, opened 2021-02-16
GEODE-7245: Remove most uses of latestViewReadLock in GMSMembership

kamilla1201:
This change replaces most uses of `latestViewReadLock` with a volatile read of the latestView.
It also makes access to `shutdownMembers` synchronized. Currently, modification of `shutdownMembers` is done under a sync but access isn't synchronized.

GEODE-7245 is an improvement ticket, so it shouldn't require writing new tests.

### For all changes:

### Note:

### DIFF
--- a/geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSMembership.java
+++ b/geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSMembership.java
@@ -275,7 +275,7 @@ public class GMSMembership<ID extends MemberIdentifier> implements Membership<ID
    * otherwise becomes pretty aggressive when a member is shutting down.
    */
   private final Set<ID> shutdownMembers =
-      Collections.synchronizedSet(new BoundedLinkedHashMap<ID, Object>().keySet());
+      Collections.synchronizedSet(Collections.newSetFromMap(new BoundedLinkedHashMap<>()));
 
   /**
    * per bug 39552, keep a list of members that have been shunned and for which a message is
@@ -1198,6 +1198,7 @@ public class GMSMembership<ID extends MemberIdentifier> implements Membership<ID
       logger.debug("Membership: recording shutdown status of {}", id);
     }
     this.shutdownMembers.add(id);
+    services.getHealthMonitor().memberShutdown(id, reason);
     services.getJoinLeave().memberShutdown(id, reason);
   }
 
--- a/geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSMembership.java
+++ b/geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSMembership.java
@@ -1488,7 +1488,7 @@ public class GMSMembership<ID extends MemberIdentifier> implements Membership<ID
    * <p>
    * Like isShunned, this method holds the view lock while executing
    *
-   * Concurrency: protected by {@link #latestViewReadLock} ReentrantReadLock
+   * Concurrency: protected by {@link #latestViewReadLock}
    *
    * @param m the member in question
    * @return true if the given member is a surprise member
--- a/geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSMembership.java
+++ b/geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSMembership.java
@@ -223,7 +223,7 @@ public class GMSMembership<ID extends MemberIdentifier> implements Membership<ID
   /**
    * This is the latest view (ordered list of IDs) that has been installed
    *
-   * All accesses to this object are protected via {@link #latestViewLock}
+   * Writing to this object is protected via {@link #latestViewLock}
    */
   private volatile MembershipView<ID> latestView = new MembershipView<>();
 
@@ -263,7 +263,7 @@ public class GMSMembership<ID extends MemberIdentifier> implements Membership<ID
    *
    * Members are removed after {@link #SHUNNED_SUNSET} seconds have passed.
    *
-   * Accesses to this list needs to be under the read or write lock of {@link #latestViewLock}
+   * Writing to this list needs to be under the write lock of {@link #latestViewLock}
    *
    * @see System#currentTimeMillis()
    */
@@ -273,6 +273,9 @@ public class GMSMembership<ID extends MemberIdentifier> implements Membership<ID
   /**
    * Members that have sent a shutdown message. This is used to suppress suspect processing that
    * otherwise becomes pretty aggressive when a member is shutting down.
+   *
+   * Accesses to this map should be synchronized on the map to avoid concurrent
+   * modification exceptions
    */
   private final Map<ID, Object> shutdownMembers = new BoundedLinkedHashMap<>();
 
@@ -281,7 +284,7 @@ public class GMSMembership<ID extends MemberIdentifier> implements Membership<ID
    * printed. Contents of this list are cleared at the same time they are removed from
    * {@link #shunnedMembers}.
    *
-   * Accesses to this list needs to be under the read or write lock of {@link #latestViewLock}
+   * Writing to this list needs to be under the write lock of {@link #latestViewLock}
    */
   private final HashSet<ID> shunnedAndWarnedMembers = new HashSet<>();
   /**
@@ -295,7 +298,7 @@ public class GMSMembership<ID extends MemberIdentifier> implements Membership<ID
    * {@link #surpriseMemberTimeout} milliseconds have passed, a view containing the member has not
    * arrived, the member is removed from membership and member-left notification is performed.
    * <p>
-   * > Accesses to this list needs to be under the read or write lock of {@link #latestViewLock}
+   * > Writing to this list needs to be under the write lock of {@link #latestViewLock}
    *
    * @see System#currentTimeMillis()
    */
--- a/geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSMembership.java
+++ b/geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSMembership.java
@@ -273,11 +273,9 @@ public class GMSMembership<ID extends MemberIdentifier> implements Membership<ID
   /**
    * Members that have sent a shutdown message. This is used to suppress suspect processing that
    * otherwise becomes pretty aggressive when a member is shutting down.
-   *
-   * Accesses to this map should be synchronized on the map to avoid concurrent
-   * modification exceptions
    */
-  private final Map<ID, Object> shutdownMembers = new BoundedLinkedHashMap<>();
+  private final Set<ID> shutdownMembers =
+      Collections.synchronizedSet(new BoundedLinkedHashMap<ID, Object>().keySet());
 
   /**
    * per bug 39552, keep a list of members that have been shunned and for which a message is
@@ -658,12 +656,10 @@ public class GMSMembership<ID extends MemberIdentifier> implements Membership<ID
       return; // Explicit deletion, no upcall.
     }
 
-    synchronized (shutdownMembers) {
-      if (!shutdownMembers.containsKey(dm)) {
-        // if we've received a shutdown message then DistributionManager will already have
-        // notified listeners
-        listener.memberDeparted(dm, crashed, reason);
-      }
+    if (!shutdownMembers.contains(dm)) {
+      // if we've received a shutdown message then DistributionManager will already have
+      // notified listeners
+      listener.memberDeparted(dm, crashed, reason);
     }
   }
 
@@ -1201,18 +1197,14 @@ public class GMSMembership<ID extends MemberIdentifier> implements Membership<ID
     if (logger.isDebugEnabled()) {
       logger.debug("Membership: recording shutdown status of {}", id);
     }
-    synchronized (this.shutdownMembers) {
-      this.shutdownMembers.put(id, id);
-      services.getHealthMonitor()
-          .memberShutdown(id, reason);
-      services.getJoinLeave().memberShutdown(id, reason);
-    }
+    this.shutdownMembers.add(id);
+    services.getJoinLeave().memberShutdown(id, reason);
   }
 
   @Override
   public Set<ID> getMembersNotShuttingDown() {
     synchronized (shutdownMembers) {
-      return latestView.getMembers().stream().filter(id -> !shutdownMembers.containsKey(id))
+      return latestView.getMembers().stream().filter(id -> !shutdownMembers.contains(id))
           .collect(
               Collectors.toSet());
     }
@@ -1302,11 +1294,7 @@ public class GMSMembership<ID extends MemberIdentifier> implements Membership<ID
 
   @Override
   public void suspectMember(ID mbr, String reason) {
-    final boolean isMemberShuttingDown;
-    synchronized (shutdownMembers) {
-      isMemberShuttingDown = this.shutdownMembers.containsKey(mbr);
-    }
-    if (!this.shutdownInProgress && !isMemberShuttingDown) {
+    if (!this.shutdownInProgress && !shutdownMembers.contains(mbr)) {
       verifyMember(mbr, reason);
     }
   }
@@ -1487,9 +1475,15 @@ public class GMSMembership<ID extends MemberIdentifier> implements Membership<ID
   }
 
   private boolean isShunnedOrNew(final ID m) {
-    return shunnedMembers.containsKey(m) || isNew(m);
+    latestViewReadLock.lock();
+    try {
+      return shunnedMembers.containsKey(m) || isNew(m);
+    } finally { // synchronized
+      latestViewReadLock.unlock();
+    }
   }
 
+  // must be invoked under view read or write lock
   private boolean isNew(final ID m) {
     return !latestView.contains(m) && !surpriseMembers.containsKey(m);
   }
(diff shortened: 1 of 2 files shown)


echobravopapa (approved):
Nice work, thanks for the updated comments as well.

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSMembership.java
```diff
@@ -273,6 +273,9 @@ boolean isDistributionMessage() {
   /**
    * Members that have sent a shutdown message. This is used to suppress suspect processing that
    * otherwise becomes pretty aggressive when a member is shutting down.
+   *
+   * Accesses to this map should be synchronized on the map to avoid concurrent
+   * modification exceptions
```
Bill, line 278: Since we only ever `shutdownMembers.put(id, id)` this can be a `Set<ID>`. It needn't be a map.

Furthermore, if we made it a concurrent map-based set, we could eliminate all the synchronization below.

This declaration could become:

```
private final Set<ID> shutdownMembers = ConcurrentHashMap.newKeySet();
```

Then code below like this:

```
    synchronized (shutdownMembers) {
      if (!shutdownMembers.containsKey(dm)) {
        // if we've received a shutdown message then DistributionManager will already have
        // notified listeners
        listener.memberDeparted(dm, crashed, reason);
      }
    }

```

changes to this:

```
      if (!shutdownMembers.containsKey(dm)) {
        // if we've received a shutdown message then DistributionManager will already have
        // notified listeners
        listener.memberDeparted(dm, crashed, reason);
      }
```

and code below like this:

```
    synchronized (shutdownMembers) {
      return latestView.getMembers().stream().filter(id -> !shutdownMembers.containsKey(id))
          .collect(
              Collectors.toSet());
    }
```

changes to code like this:

```
    final Set<ID> membersNotShuttingDown = new HashSet<ID>(latestView.getMembers());
    final Set<ID> shutdownMembersStable = this.shutdownMembers;
    membersNotShuttingDown.removeIf(shutdownMembersStable::contains);
    return membersNotShuttingDown;
```

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSMembership.java
```diff
@@ -677,10 +658,12 @@ private void removeWithViewLock(ID dm, boolean crashed, String reason) {
...
 
-    if (!shutdownMembers.containsKey(dm)) {
-      // if we've received a shutdown message then DistributionManager will already have
-      // notified listeners
-      listener.memberDeparted(dm, crashed, reason);
+    synchronized (shutdownMembers) {
+      if (!shutdownMembers.containsKey(dm)) {
+        // if we've received a shutdown message then DistributionManager will already have
+        // notified listeners
+        listener.memberDeparted(dm, crashed, reason);
```
bschuchardt, line 665: Is it necessary to hold the lock on shutdownMembers while invoking a listener?  The point of this check is to avoid invoking memberDeparted if a shutdown message from **dm** has already been processed.  In that case we've already invoked memberDeparted on the listener.

Bill (changes requested):
I have one change request below.

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSMembership.java
```diff
@@ -273,6 +273,9 @@ boolean isDistributionMessage() {
   /**
    * Members that have sent a shutdown message. This is used to suppress suspect processing that
    * otherwise becomes pretty aggressive when a member is shutting down.
+   *
+   * Accesses to this map should be synchronized on the map to avoid concurrent
+   * modification exceptions
    */
   private final Map<ID, Object> shutdownMembers = new BoundedLinkedHashMap<>();
```
bschuchardt, line 280: I'm wondering if this could be made a synchronized collection with Collections.synchronizedMap().  Then we could get rid of the external synchronization.

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSMembership.java
```diff
@@ -1524,15 +1487,9 @@ public boolean isShunned(ID m) {
...
-      return shunnedMembers.containsKey(m) || isNew(m);
-    } finally { // synchronized
-      latestViewReadLock.unlock();
-    }
+    return shunnedMembers.containsKey(m) || isNew(m);
   }
 
-  // must be invoked under view read or write lock
   private boolean isNew(final ID m) {
     return !latestView.contains(m) && !surpriseMembers.containsKey(m);
```
bschuchardt, line 1494: I think the read-lock is still needed here.  While the "shunned" state of the member is disconnected from the other two checks, surpriseMembers is modified while forming a new membership view and we want these two checks to be atomic.

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSMembership.java
Bill: @bschuchardt points out that this won't work by itself since we'd lose the bounding behavior of the `BoundedLinkedHashMap`. It's ensuring that the collection never holds more than 1000 items. And it's the 1000 most-recent ones.

bschuchardt (commented):
Nice work - I like all of the code clean-up and simplification.

I'm a little worried about the distributed lock service elder selection.  I think it's probably broken in its current state but this PR might make it worse.  ClusterElderManager gets the current view's members and then asks which of them are surprise members.  The view could change while this is happening and the surpriseMembers collection could be modified.  I'm not sure why that's being done, but it does seem like that code wants consistency between surpriseMembers and the current view while it does that check.  Maybe that code needs to be executed under the view lock?

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSMembership.java
Bill: We could keep the synchronization and improve things a tiny bit by declaring the field this way:

```
  private final Set<ID> shutdownMembers = new BoundedLinkedHashMap<ID,ID>().keySet();
```

That'd let us `shutdownMembers.put(id)` and use `shutdownMembers.contains(id)` instead of `shutdownMembers.containsKey(id)`.

bschuchardt (approved):

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSMembership.java
```diff
@@ -1231,23 +1197,16 @@ public void shutdownMessageReceived(ID id, String reason) {
...
+    this.shutdownMembers.add(id);
+    services.getJoinLeave().memberShutdown(id, reason);
   }
 
   @Override
   public Set<ID> getMembersNotShuttingDown() {
-    latestViewReadLock.lock();
-    try {
-      return latestView.getMembers().stream().filter(id -> !shutdownMembers.containsKey(id))
+    synchronized (shutdownMembers) {
```
pivotal-jbarrett, line 1207: Only a single thread can read this set of members now. Previously all threads could read concurrently.

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSMembership.java
```diff
@@ -1231,23 +1197,16 @@ public void shutdownMessageReceived(ID id, String reason) {
...
       logger.debug("Membership: recording shutdown status of {}", id);
     }
-    synchronized (this.shutdownMembers) {
-      this.shutdownMembers.put(id, id);
-      services.getHealthMonitor()
-          .memberShutdown(id, reason);
-      services.getJoinLeave().memberShutdown(id, reason);
-    }
+    this.shutdownMembers.add(id);
+    services.getJoinLeave().memberShutdown(id, reason);
```
pivotal-jbarrett, line 1202: All threads can now concurrently invoke `memberShutdown`, is this desirable?

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSMembership.java
```diff
@@ -1360,13 +1319,8 @@ public boolean verifyMember(ID mbr, String reason) {
...
   @Override
   public ID[] getAllMembers(final ID[] arrayType) {
-    latestViewReadLock.lock();
-    try {
-      List<ID> keySet = latestView.getMembers();
-      return keySet.toArray(arrayType);
-    } finally {
-      latestViewReadLock.unlock();
-    }
+    final List<ID> keySet = latestView.getMembers();
```
pivotal-jbarrett, line 1323: Is the latest view and/or members list immutable?

pivotal-jbarrett (changes requested):
We seem to now have a confusing mix of read/write lock, synchronized and atomic/volatile operations.  I would feel more comfortable with this change if we could reduce the different types of concurrency protections were are using here.

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSMembership.java
Bill: `latestView` isa `MembershipView` and yes, this PR makes `MembershipView` immutable (see `MembershipView.java`)

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSMembership.java
Bill: Previously reading threads waited for the `latestViewReadLock`.

But also there was previously a latent bug here, to wit, puts to the non-concurrent collection `shutdownMembers` were protected (via synchronization on the collection) whereas the read access to that collection by the `filter()` loop here was not protected via synchronization.

I think synchronization here is a valid fix for that problem.

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSMembership.java
Bill: My bigger concern in this version of this method is the lack of synchronization around `shutdownMembers.add(id)` since that collection is not thread-safe.

@pivotal-jbarrett I believe @kamilla1201 has a new commit coming to fix CI…

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSMembership.java
kamilla1201: This PR also makes `members` list unmodifiable.

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSMembership.java
kamilla1201: I pushed a commit a few hours ago that makes `shutdownMembers` a `synchronizedSet`. So `shutdownMembers.add(id)` call should be thread-safe here. I'm not sure whether the concurrent invocation of memberShutdown can cause any issues.

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSMembership.java
kamilla1201: Yes, as Bill said, `shutdownMembers` was a non-concurrent collection, and the read access to it wasn't synched. Although `shutdownMembers` is a `synchronizedSet` now, synchronization is still needed around `stream()` operation.

pivotal-jbarrett: You are saying that currently the shutdownMembers collection can be mutated without acquiring the write lock? If that is the case wouldn't the more appropriate fix be to acquire the write lock before mutation rather than adding a serializing synchronization block here?

pivotal-jbarrett: Or give `shutdownMembers` the same immutability treatment that `lastView` has.

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSMembership.java
pivotal-jbarrett: The other concern here is now updates to `shutdownMembers` and the call to `memberShutdown` can be concurrently interleaved.

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSMembership.java
pivotal-jbarrett: So if the view is immutable, what is the r/w lock protecting anymore?

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSMembership.java
pivotal-jbarrett: Doing so also allows you to make the "not shutting down" members list cacheable rather than creating a new list every time the method is called.

Bill (changes requested):
Many of the comments talking about `latestViewLock` are wrong since that lock is never used directly any more. Only its read and write locks are used. Please review all occurrences of `latestViewLock` in comments and be sure they are each talking (specifically) about the pertinent read or write lock.

image

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSMembership.java
bschuchardt: Since this is a potentially long linked list I don't think copy-on-write is a good pattern to use for shutdownMembers.
This method is invoked on every message-send, so maybe using a synchronized collection isn't a great idea and we should use the read/write lock instead.

pivotal-jbarrett: If the list isn't updated frequently even a long list with copy on write would be better less intrusive than frequently grabbing a read lock. If this method is called for every p2p message we certainly can't have it synchronizing, it should do the least amount of blocking as possible.

pivotal-jbarrett: Wait a second, we are creating a new set of members not shutting down for every p2p message sent?

bschuchardt: @pivotal-jbarrett yes, we've always created a new collection like this for the reply processors.  This code now seems to create two (if you look at its primary use in ClusterDistributionManager) so that could be made more efficient.

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSMembership.java
bschuchardt: @pivotal-jbarrett shutdownMembers and memberShutdown are not required to be in sync.  The memberShutdown invocation informs GMSJoinLeave so it can record a LeaveRequest in case this is the membership coordinator or that it should become one.
The latestViewLock is for protecting the data used in processView() from being altered while a new view is being prepared.  The shutdownMembers field isn't used in processView() and so doesn't need to be protected by this lock.

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSMembership.java
bschuchardt: The lock is protecting new view formation in processView().  This data needs to be stable while the new view is created.

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSMembership.java
pivotal-jbarrett: If the `shutdownMembers` list isn't changing frequently, like every ms, then repeatedly recreating the same set is just adding pressure to the GC and CPU time all while under a synchronized lock for no real gain. I assume this `shutdownMembers` list is updated with the member id of the member that has said it is or has gone down, which isn't a frequent event in production. Seems like we should be optimizing this method to be returning a rather static list of members known not to be shutting down even if that means the infrequent setting of a shutting down member takes a bit more time to handle, by copying lists and invalidating or updating the not shutdown servers list. 

At the very least the addition of this synchronized block is going to have a noticeable impact on the p2p messaging at scale since all messages will be serialized through this lock. If we deem refactoring out the list creation is out of scope for this PR then using a read/write lock here should be in scope as a first step.

kamilla1201: @pivotal-jbarrett We'll consider creating another ticket to optimize the method, as you said. For now, I replaced the synchronized blocks with read/write locks.

pivotal-jbarrett: Awesome! Thanks!

pivotal-jbarrett (approved):

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSMembership.java
```diff
@@ -1545,7 +1488,7 @@ private boolean isNew(final ID m) {
    * <p>
    * Like isShunned, this method holds the view lock while executing
    *
-   * Concurrency: protected by {@link #latestViewLock} ReentrantReadWriteLock
+   * Concurrency: protected by {@link #latestViewReadLock} ReentrantReadWriteLock
```
Bill, line 1491: Typo: it's a re-entrant read lock, not a re-entrant read-write lock

Bill (changes requested):
One last documentation tweak then I approve.

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSMembership.java
Bill: I misspoke and the comment is still wrong. This would be better:

```java
* Concurrency: protected by {@link #latestViewReadLock}
```

no such thing as a "ReentrantReadWriteLock". Sorry for the confusion!

Bill (approved):
Approved!

PR #6038, opened 2021-02-16
GEODE-8947: Use waiting thread pool only in limited senario.

pivotal-eshu:
* Only when processing transactional message and when conserve-sockets set to true case, a
    separate thread in waiting pool will be used to process the message.
  * This is to addresses performance issue if there won't be deadlock.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/tx/RemoteOperationMessage.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/tx/RemoteOperationMessage.java
@@ -14,7 +14,6 @@
  */
 package org.apache.geode.internal.cache.tx;
 
-import static org.apache.geode.distributed.ConfigurationProperties.CONSERVE_SOCKETS;
 
 import java.io.DataInput;
 import java.io.DataOutput;
@@ -180,9 +179,8 @@ public abstract class RemoteOperationMessage extends DistributionMessage
       return;
     }
 
-    String conserveSockets = getConserveSocketsSetting(dm);
-    if (conserveSockets != null && conserveSockets.equals("false")) {
-      // reply inline for CONSERVE_SOCKETS == false case.
+    if (dm.getSystem().threadOwnsResources()) {
+      // reply inline if thread owns socket.
       doRemoteOperation(dm, cache);
       return;
     }
@@ -193,10 +191,6 @@ public abstract class RemoteOperationMessage extends DistributionMessage
       // reply inline for non-transactional case.
       doRemoteOperation(dm, cache);
     }
-  }
-
-  String getConserveSocketsSetting(ClusterDistributionManager dm) {
-    return dm.getSystem().getProperties().getProperty(CONSERVE_SOCKETS);
   }
 
   boolean isTransactional() {
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/tx/RemoteOperationMessage.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/tx/RemoteOperationMessage.java
@@ -14,6 +14,8 @@
  */
 package org.apache.geode.internal.cache.tx;
 
+import static org.apache.geode.distributed.ConfigurationProperties.CONSERVE_SOCKETS;
+
 import java.io.DataInput;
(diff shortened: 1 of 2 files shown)


geode-core/src/main/java/org/apache/geode/internal/cache/tx/RemoteOperationMessage.java
```diff
@@ -177,7 +179,28 @@ public void process(final ClusterDistributionManager dm) {
...
+    if (isTransactional()) {
+      dm.getExecutors().getWaitingThreadPool().execute(() -> doRemoteOperation(dm, cache));
+    } else {
+      // reply inline for non-transactional case.
+      doRemoteOperation(dm, cache);
+    }
+  }
+
+  String getConserveSocketsSetting(ClusterDistributionManager dm) {
+    return dm.getSystem().getProperties().getProperty(CONSERVE_SOCKETS);
```
DonalEvans, line 199: I think it would be better to get the value of `conserve-sockets` from `DistributionConfigImpl` rather than check if a system property has been set, since a user may be using the default value. Since the default value of `conserve-sockets` was changed in 1.14, this would lead to a difference in behaviour when checking if the system property is present or not. If that's not possible, then a check should be included to see what the default value of `conserve-sockets` is (`DistributionConfig.DEFAULT_CONSERVE_SOCKETS`) if the system property is not set so that the correct action can be taken.

geode-core/src/test/java/org/apache/geode/internal/cache/tx/RemoteOperationMessageTest.java
```diff
@@ -302,6 +303,52 @@ public void processWithNullPointerExceptionFromOperationOnRegionWithSystemFailur
...
+  }
+
+  @Test
+  public void isTransactionalReturnsFalseIfTXUniqueIdIsNOTX() {
+    assertThat(msg.getTXUniqId()).isEqualTo(TXManagerImpl.NOTX);
+    assertThat(msg.isTransactional()).isFalse();
+  }
+
+  @Test
+  public void isTransactionalReturnsTrueIfCannotParticipateInTransaction() {
```
DonalEvans, line 337: This test name should be "isTransactionalReturnsFalseIfCannotParticipateInTransaction"

geode-core/src/test/java/org/apache/geode/internal/cache/tx/RemoteOperationMessageTest.java
```diff
@@ -302,6 +303,52 @@ public void processWithNullPointerExceptionFromOperationOnRegionWithSystemFailur
...
+    doReturn(1).when(msg).getTXUniqId();
+    doReturn(false).when(msg).canParticipateInTransaction();
+
+    assertThat(msg.isTransactional()).isFalse();
+  }
+
+  @Test
+  public void isTransactionalReturnsFalseIfHasTXUniqueIdAndCanParticipateInTransaction() {
```
DonalEvans, line 345: This test name should be "isTransactionalReturnsTrueIfHasTXUniqueIdAndCanParticipateInTransaction"

DonalEvans (changes requested):

geode-core/src/main/java/org/apache/geode/internal/cache/tx/RemoteOperationMessage.java
bschuchardt: I agree with Donal, though this isn't a System Property but a DistributionConfig converted into properties.  It would be better to get the config and query it.  Creating a Properties object isn't all that cheap.

bschuchardt: I think what you want to know here is whether the message is going to be sent on a thread-owned Socket or a shared Socket.  To determine that you need to use InternalDistributedSystem.threadOwnsResources(), not the conserve-sockets setting.

bschuchardt (changes requested):

bschuchardt (approved):
looks good

DonalEvans (approved):

gesterzhou (approved):

PR #6042, opened 2021-02-20
GEODE-8958: When timestamps get corrupted.

mhansonp:
The system would wait, now it does expires it and moves on.

### DIFF
--- a/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/versions/TombstoneDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/versions/TombstoneDUnitTest.java
@@ -171,7 +171,7 @@ public class TombstoneDUnitTest implements Serializable {
 
   @Test
   public void testWhenAnOutOfRangeTimeStampIsSeenWeExpireItInNonReplicateTombstoneSweeper() {
-    VM server1 = VM.getVM(-1);
+    VM server1 = VM.getVM(0);
     VM server2 = VM.getVM(1);
     final int FAR_INTO_THE_FUTURE = 1000000; // 1 million millis into the future
     final int count = 2000;
--- a/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/versions/TombstoneDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/versions/TombstoneDUnitTest.java
@@ -147,11 +147,10 @@ public class TombstoneDUnitTest implements Serializable {
 
   @Test
   public void testRewriteBadOldestTombstoneTimeReplicateForTimestamp() {
-    VM server1 = VM.getVM(-1);
+    VM server1 = VM.getVM(0);
     VM server2 = VM.getVM(1);
 
     final int count = 10;
-
     server1.invoke(() -> {
       createCacheAndRegion(RegionShortcut.REPLICATE_PERSISTENT);
       for (int i = 0; i < count; i++) {
@@ -166,18 +165,13 @@ public class TombstoneDUnitTest implements Serializable {
           ((InternalCache) cache).getTombstoneService().getSweeper((LocalRegion) region);
 
       RegionEntry regionEntry = ((LocalRegion) region).getRegionEntry("K0");
-      VersionTag<?> versionTag = regionEntry.getVersionStamp()
-          .asVersionTag();
-      versionTag.setVersionTimeStamp(System.currentTimeMillis() + 100000);
+      VersionTag<?> versionTag = regionEntry.getVersionStamp().asVersionTag();
+      versionTag.setVersionTimeStamp(System.currentTimeMillis() + 1000000);
+
       TombstoneService.Tombstone modifiedTombstone =
           new TombstoneService.Tombstone(regionEntry, (LocalRegion) region,
               versionTag);
       tombstoneSweeper.tombstones.add(modifiedTombstone);
-      if (tombstoneSweeper.getOldestTombstoneTime() > 0) {
-        System.out.println("We have a problem");
-      } else {
-        System.out.println("It works.");
-      }
       tombstoneSweeper.checkOldestUnexpired(System.currentTimeMillis());
       // Send tombstone gc message to vm1.
       assertThat(tombstoneSweeper.getOldestTombstoneTime()).isEqualTo(0);
--- a/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/versions/TombstoneDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/versions/TombstoneDUnitTest.java
@@ -124,8 +124,9 @@ public class TombstoneDUnitTest implements Serializable {
     VM server1 = VM.getVM(0);
     VM server2 = VM.getVM(1);
     final int FAR_INTO_THE_FUTURE = 1000000; // 1 million millis into the future
-
     final int count = 10;
+
+    // Create a cache and load some boiler plate entries
     server1.invoke(() -> {
       createCacheAndRegion(RegionShortcut.REPLICATE_PERSISTENT);
       for (int i = 0; i < count; i++) {
@@ -136,19 +137,31 @@ public class TombstoneDUnitTest implements Serializable {
     server2.invoke(() -> createCacheAndRegion(RegionShortcut.REPLICATE));
 
     server1.invoke(() -> {
+
+      // Now that we have a cache and a region specifically with data, we can start the real work
       TombstoneService.TombstoneSweeper tombstoneSweeper =
           ((InternalCache) cache).getTombstoneService().getSweeper((LocalRegion) region);
 
+      // Get one of the entries
       RegionEntry regionEntry = ((LocalRegion) region).getRegionEntry("K0");
+
+      /*
+       * Create a version tag with a timestamp far off in the future...
+       * It should be in the near past, but we are testing that a future tombstone will be cleared
+       */
       VersionTag<?> versionTag = regionEntry.getVersionStamp().asVersionTag();
       versionTag.setVersionTimeStamp(System.currentTimeMillis() + FAR_INTO_THE_FUTURE);
 
+      // Create the forged tombstone with the versionTag from the future
       TombstoneService.Tombstone modifiedTombstone =
           new TombstoneService.Tombstone(regionEntry, (LocalRegion) region,
               versionTag);
+
+      // Add it to the list of tombstones so that when checkOldestUnexpired is called it will see it
       tombstoneSweeper.tombstones.add(modifiedTombstone);
       tombstoneSweeper.checkOldestUnexpired(System.currentTimeMillis());
-      // Send tombstone gc message to vm1.
+
+      // Validate that the tombstone was cleared.
       assertThat(tombstoneSweeper.getOldestTombstoneTime()).isEqualTo(0);
     });
   }
--- a/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/versions/TombstoneDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/versions/TombstoneDUnitTest.java
@@ -61,6 +61,7 @@ public class TombstoneDUnitTest implements Serializable {
   private static Cache cache;
   private static Region<String, String> region;
   final String REGION_NAME = "TestRegion";
+  final int EXPIRY_TIME=600000;
 
   @Rule
   public DistributedRule distributedRule = new DistributedRule();
@@ -118,37 +119,13 @@ public class TombstoneDUnitTest implements Serializable {
     });
   }
 
-  @Test
-  public void testGetOldestTombstoneTimeReplicate() {
-    VM server1 = VM.getVM(0);
-    VM server2 = VM.getVM(1);
-
-    server1.invoke(() -> {
-      createCacheAndRegion(REPLICATE_PERSISTENT);
-      region.put("K1", "V1");
-      region.put("K2", "V2");
-    });
-
-    server2.invoke(() -> createCacheAndRegion(REPLICATE));
-
-    server1.invoke(() -> {
-      // Send tombstone gc message to vm1.
-      region.destroy("K1");
-
-      TombstoneService.TombstoneSweeper tombstoneSweeper =
-          ((InternalCache) cache).getTombstoneService().getSweeper((LocalRegion) region);
-
-      assertThat(tombstoneSweeper.getOldestTombstoneTime()).isGreaterThan(0)
-          .isLessThan(((InternalCache) cache).cacheTimeMillis());
-      performGC(1);
-      assertThat(tombstoneSweeper.getOldestTombstoneTime()).isEqualTo(0);
-    });
-  }
 
   @Test
-  public void testRewriteBadOldestTombstoneTimeReplicateForTimestamp() {
+  public void testWhenAnOutOfRangeTimeStampIsSeenWeExpireItInReplicateTombstoneSweeper() {
     VM server1 = VM.getVM(0);
     VM server2 = VM.getVM(1);
+    final int FAR_INTO_THE_FUTURE = 1000000; // 1 million millis into the future
+
 
     final int count = 10;
     server1.invoke(() -> {
@@ -166,7 +143,7 @@ public class TombstoneDUnitTest implements Serializable {
 
       RegionEntry regionEntry = ((LocalRegion) region).getRegionEntry("K0");
       VersionTag<?> versionTag = regionEntry.getVersionStamp().asVersionTag();
-      versionTag.setVersionTimeStamp(System.currentTimeMillis() + 1000000);
+      versionTag.setVersionTimeStamp(System.currentTimeMillis() + FAR_INTO_THE_FUTURE);
 
       TombstoneService.Tombstone modifiedTombstone =
           new TombstoneService.Tombstone(regionEntry, (LocalRegion) region,
@@ -178,9 +155,8 @@ public class TombstoneDUnitTest implements Serializable {
     });
   }
 
-
   @Test
-  public void testGetOldestTombstoneTimeReplicateForTimestamp() {
+  public void testGetOldestTombstoneTimeForReplicateTombstoneSweeper() {
     VM server1 = VM.getVM(0);
     VM server2 = VM.getVM(1);
     final int count = 10;
@@ -199,7 +175,7 @@ public class TombstoneDUnitTest implements Serializable {
       // Send tombstone gc message to vm1.
       for (int i = 0; i < count; i++) {
         region.destroy("K" + i);
-        assertThat(tombstoneSweeper.getOldestTombstoneTime() + 30000 - System.currentTimeMillis())
+        assertThat(tombstoneSweeper.getOldestTombstoneTime() + EXPIRY_TIME - System.currentTimeMillis())
             .isGreaterThan(0);
         performGC(1);
       }
@@ -209,7 +185,7 @@ public class TombstoneDUnitTest implements Serializable {
   }
 
   @Test
-  public void testGetOldestTombstoneTimeNonReplicate() {
+  public void testGetOldestTombstoneTimeForNonReplicateTombstoneSweeper() {
     VM client = VM.getVM(0);
     VM server = VM.getVM(1);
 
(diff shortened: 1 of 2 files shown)


geode-core/src/distributedTest/java/org/apache/geode/internal/cache/versions/TombstoneDUnitTest.java
```diff
@@ -144,6 +145,75 @@ public void testGetOldestTombstoneTimeReplicate() {
     });
   }
 
+  @Test
+  public void testRewriteBadOldestTombstoneTimeReplicateForTimestamp() {
+    VM server1 = VM.getVM(-1);
```
nabarunnag, line 150: I think the debug parameter needs to be changed to 0

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/versions/TombstoneDUnitTest.java
```diff
@@ -144,6 +145,75 @@ public void testGetOldestTombstoneTimeReplicate() {
...
+      RegionEntry regionEntry = ((LocalRegion) region).getRegionEntry("K0");
+      VersionTag<?> versionTag = regionEntry.getVersionStamp()
+          .asVersionTag();
+      versionTag.setVersionTimeStamp(System.currentTimeMillis() + 100000);
+      TombstoneService.Tombstone modifiedTombstone =
+          new TombstoneService.Tombstone(regionEntry, (LocalRegion) region,
+              versionTag);
+      tombstoneSweeper.tombstones.add(modifiedTombstone);
+      if (tombstoneSweeper.getOldestTombstoneTime() > 0) {
+        System.out.println("We have a problem");
```
nabarunnag, line 177: assert checks instead of System outs

geode-core/src/main/java/org/apache/geode/internal/cache/TombstoneService.java
```diff
@@ -749,7 +751,8 @@ protected boolean hasExpired(long msTillHeadTombstoneExpires) {
         testHook_forceExpirationCount--;
         return true;
       }
-      return msTillHeadTombstoneExpires <= 0;
+      // In case
+      return msTillHeadTombstoneExpires <= 0 || msTillHeadTombstoneExpires > EXPIRY_TIME;
```
nabarunnag, line 755: I maybe missing the math but long msTillHeadTombstoneExpires = oldest.getVersionTimeStamp() + EXPIRY_TIME - now;
the only time msTillHeadTombstoneExpires is greater than EXPIRY_TIME will be when oldest.timeStamp - now is a postive number. Can this occur?

nabarunnag (changes requested):
Had a few change requests and questions for my clarity

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/versions/TombstoneDUnitTest.java
mhansonp: You are correct. Thanks!

geode-core/src/main/java/org/apache/geode/internal/cache/TombstoneService.java
mhansonp: That is the problem we are seeing. The customer is having timestamps far into the future. That is why we are putting this change in... So once we subtract now from oldest.getVersionTimeStamp() the number is far greater than EXPIRY_TIME. This is intended to catch that condition...

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/versions/TombstoneDUnitTest.java
mhansonp: Actually these should not be there. I pulled them... Too much task switching sorry.

kirklund (commented):
The TombstoneDUnitTest looks good. Can we update the TombstoneServiceTest unit test to cover any of these changes? It would be nice to have some unit tests that target `return msTillHeadTombstoneExpires <= 0 || msTillHeadTombstoneExpires > EXPIRY_TIME;`.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/versions/TombstoneDUnitTest.java
```diff
@@ -144,6 +145,69 @@ public void testGetOldestTombstoneTimeReplicate() {
     });
   }
 
+  @Test
+  public void testRewriteBadOldestTombstoneTimeReplicateForTimestamp() {
```
DonalEvans, line 149: This test name is a little unclear. Would it be possible to rename the test to make it clear what behaviour is being tested and what the expected outcome is?

geode-core/src/main/java/org/apache/geode/internal/cache/TombstoneService.java
```diff
@@ -373,7 +373,8 @@ public Object getBlockGCLock() {
     return this.replicatedTombstoneSweeper.getBlockGCLock();
   }
 
-  protected static class Tombstone extends CompactVersionHolder {
+  @VisibleForTesting
+  public static class Tombstone extends CompactVersionHolder {
```
DonalEvans, line 377: To prevent having to make this class and its fields/methods public, would it be possible to move TombstoneDUnitTest to the same package as this one? Or to move this class to the same package as TeombstoneDUnitTest?

geode-core/src/main/java/org/apache/geode/internal/cache/TombstoneService.java
DonalEvans: It seems like this fix is just masking the real issue, which is that timestamps shouldn't be far in the future in the first place. Do we have any idea why/how that might be happening?

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/versions/TombstoneDUnitTest.java
```diff
@@ -144,6 +145,69 @@ public void testGetOldestTombstoneTimeReplicate() {
...
+      tombstoneSweeper.tombstones.add(modifiedTombstone);
+      tombstoneSweeper.checkOldestUnexpired(System.currentTimeMillis());
+      // Send tombstone gc message to vm1.
+      assertThat(tombstoneSweeper.getOldestTombstoneTime()).isEqualTo(0);
+    });
+  }
+
+
+  @Test
+  public void testGetOldestTombstoneTimeReplicateForTimestamp() {
```
DonalEvans, line 183: This test name is also a little unclear in what the test is actually doing. Would it be possible to change it to something that conveys what the behaviour being tested is and what the expected outcome is?

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/versions/TombstoneDUnitTest.java
```diff
@@ -144,6 +145,69 @@ public void testGetOldestTombstoneTimeReplicate() {
...
+
+    server2.invoke(() -> createCacheAndRegion(RegionShortcut.REPLICATE));
+
+    server1.invoke(() -> {
+      TombstoneService.TombstoneSweeper tombstoneSweeper =
+          ((InternalCache) cache).getTombstoneService().getSweeper((LocalRegion) region);
+      // Send tombstone gc message to vm1.
+      for (int i = 0; i < count; i++) {
+        region.destroy("K" + i);
+        assertThat(tombstoneSweeper.getOldestTombstoneTime() + 30000 - System.currentTimeMillis())
```
DonalEvans, line 202: Where is this value of 30000 coming from? Would it be possible to replace it with a reference to a constant rather than hard-coding it?

DonalEvans (changes requested):
I wonder if it's possible to convert the test cases added from DUnit to Unit or Integration tests, since they seem to be testing the behaviour of just one class, rather than the system as a whole. It's fine if that's not feasible.

dschneider-pivotal (approved):

geode-core/src/main/java/org/apache/geode/internal/cache/TombstoneService.java
gesterzhou: For such a workaround, we should put detail comments here

gesterzhou (approved):
Add comments for your workaround. Others look good.

geode-core/src/main/java/org/apache/geode/internal/cache/TombstoneService.java
mhansonp: Yup, I lost the comments some how, so I am putting them in now. I agree this abates the symptom. I don't prefer this approach, but at the same time, this will help people with their problem that should not be happening. We were suspicious this had to do with clock time adjustments, but we are still theorizing. This will keep their system from locking up though for a condition we can safely handle in this part of the system with sanity checking. Otherwise intervention is required when this problem arises. Also to be blunt, the current approach while fast is lacking in fault tolerance.

geode-core/src/main/java/org/apache/geode/internal/cache/TombstoneService.java
mhansonp: They are only public for test. Other than that, I am not sure let me look at it.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/versions/TombstoneDUnitTest.java
```diff
@@ -61,6 +61,7 @@
   private static Cache cache;
   private static Region<String, String> region;
   final String REGION_NAME = "TestRegion";
+  final int EXPIRY_TIME=600000;
```
DonalEvans, line 64: I think it would be better to directly reference the default tombstone timeout using `REPLICATE_TOMBSTONE_TIMEOUT_DEFAULT` since if it changes, then this value will no longer be correct.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/versions/TombstoneDUnitTest.java
mhansonp: Fixed in the latest version.

mhansonp: It is a defensive approach to a problem. I agree that we should be looking at the issue beyond this change, but the ramifications of having a bad timestamp make using geode very challenging. I would also not like to see the core bug closed, but I think it is safe to make the system run better. The way this code is written (should be rewritten) is that it looks at the first entry in the queue only. So if a timestamp is in the future it can be a long wait to clear tombstones. It doesn't matter if others should be cleared in the queue only the first...

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/versions/TombstoneDUnitTest.java
mhansonp: It is just supposed to be a tombstone time to live beyond the deletion. I have made a constant and named it to represent what is in the code.

geode-core/src/main/java/org/apache/geode/internal/cache/TombstoneService.java
```diff
@@ -749,7 +751,8 @@ protected boolean hasExpired(long msTillHeadTombstoneExpires) {
         testHook_forceExpirationCount--;
         return true;
       }
-      return msTillHeadTombstoneExpires <= 0;
+      // In case
```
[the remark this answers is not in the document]
mhansonp: added a more complete comment, somehow, the old one got deleted.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/versions/TombstoneDUnitTest.java
mhansonp: Done.

DonalEvans (approved):

kirklund (approved):

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/versions/TombstoneDUnitTest.java
```diff
@@ -117,35 +118,74 @@ public void testTombstoneGcMessagesBetweenPersistentAndNonPersistentRegion() {
...
+      RegionEntry regionEntry = ((LocalRegion) region).getRegionEntry("K0");
+      VersionTag<?> versionTag = regionEntry.getVersionStamp().asVersionTag();
+      versionTag.setVersionTimeStamp(System.currentTimeMillis() + FAR_INTO_THE_FUTURE);
+
+      TombstoneService.Tombstone modifiedTombstone =
+          new TombstoneService.Tombstone(regionEntry, (LocalRegion) region,
+              versionTag);
+      tombstoneSweeper.tombstones.add(modifiedTombstone);
+      tombstoneSweeper.checkOldestUnexpired(System.currentTimeMillis());
       // Send tombstone gc message to vm1.
```
jchen21, line 151: This comment does not apply to the modified code. It is better to update the comment accordingly.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/versions/TombstoneDUnitTest.java
```diff
@@ -117,35 +118,74 @@ public void testTombstoneGcMessagesBetweenPersistentAndNonPersistentRegion() {
...
+
+    server2.invoke(() -> createCacheAndRegion(RegionShortcut.REPLICATE));
+
+    server1.invoke(() -> {
       TombstoneService.TombstoneSweeper tombstoneSweeper =
           ((InternalCache) cache).getTombstoneService().getSweeper((LocalRegion) region);
+      // Send tombstone gc message to vm1.
+      for (int i = 0; i < count; i++) {
+        region.destroy("K" + i);
+        assertThat(
```
jchen21, line 176: I am not sure I understand the purpose of this assert. Even without the fix, the assert will succeed.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/versions/TombstoneDUnitTest.java
```diff
@@ -117,35 +118,74 @@ public void testTombstoneGcMessagesBetweenPersistentAndNonPersistentRegion() {
...
-      region.put("K1", "V1");
-      region.put("K2", "V2");
+      createCacheAndRegion(RegionShortcut.REPLICATE_PERSISTENT);
+      for (int i = 0; i < count; i++) {
+        region.put("K" + i, "V" + i);
+      }
     });
 
-    server2.invoke(() -> createCacheAndRegion(REPLICATE));
+    server2.invoke(() -> createCacheAndRegion(RegionShortcut.REPLICATE));
```
jchen21, line 136: Can you explain a bit why the region shortcut on server2 is not the same as that of server1 `REPLICATE_PERSISTENT` ? I understand this is not your code change though.

geode-core/src/main/java/org/apache/geode/internal/cache/TombstoneService.java
```diff
@@ -1078,8 +1087,8 @@ private void checkOldestUnexpired(long now) {
           if (logger.isTraceEnabled(LogMarker.TOMBSTONE_VERBOSE)) {
             logger.trace(LogMarker.TOMBSTONE_VERBOSE, "oldest unexpired tombstone is {}", oldest);
           }
-          long msTillHeadTombstoneExpires = oldest.getVersionTimeStamp() + EXPIRY_TIME - now;
-          if (hasExpired(msTillHeadTombstoneExpires)) {
+          long msUntilHeadTombstoneExpires = oldest.getVersionTimeStamp() + EXPIRY_TIME - now;
```
jchen21, line 1090: Minor pick: if you rename the variable to `msUntilHeadTombstoneExpires`, it is better to rename the variable at other places, e.g. at line 759 as well, in order to be consistent. It is better for code maintenance.

jchen21 (changes requested):

geode-core/src/main/java/org/apache/geode/internal/cache/TombstoneService.java
mhansonp: Done

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/versions/TombstoneDUnitTest.java
mhansonp: We are getting the remaining time until the tombstone should be expired. We are expecting it to be greater than 0.  Hence the assert. Then we force a GC and we expect it to be 0. Effectively this test is validating that performing a gc clears out the tombstones.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/versions/TombstoneDUnitTest.java
mhansonp: I have improved the comments.

geode-core/src/main/java/org/apache/geode/internal/cache/TombstoneService.java
```diff
@@ -423,8 +425,8 @@ protected void updateStatistics() {
     }
 
     @Override
-    protected boolean hasExpired(long msTillHeadTombstoneExpires) {
-      return msTillHeadTombstoneExpires <= 0;
+    protected boolean hasExpired(long msUntilTombstoneExpires) {
+      return msUntilTombstoneExpires <= 0;
```
agingade, line 429: We may have to add the large clock jump verification here too...I am assuming clients are also doing Tombstone GC similarly....Nice to add a test on the client side to verify it.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/versions/TombstoneDUnitTest.java
jchen21: First, usually immediately after `Region.destroy()`, the difference between `tombstoneSweeper.getOldestTombstoneTime()` and `System.currentTimeMillis()` is only a few milliseconds. So asserting that 
```
tombstoneSweeper.getOldestTombstoneTime()
                + TombstoneService.REPLICATE_TOMBSTONE_TIMEOUT_DEFAULT - System.currentTimeMillis())
                    .isGreaterThan(0)
```
is like asserting `REPLICATE_TOMBSTONE_TIMEOUT_DEFAULT` is greater than 0, i.e. asserting 600000 > 0. There is nothing wrong with that. However, a better assertion would be the one you deleted:
```
      assertThat(tombstoneSweeper.getOldestTombstoneTime()).isGreaterThan(0)
          .isLessThan(((InternalCache) cache).cacheTimeMillis());
```
Second, if we expect that GC clears out tombstones, I would recommend moving the following assertion
```
assertThat(tombstoneSweeper.getOldestTombstoneTime()).isEqualTo(0)
```
 into the `for` loop, instead of after the `for` loop.
Please correct me if I am wrong.

mhansonp: This would be wrong cacheTimeMillis has no guarantees about the next milliseconds value.  So dealing with values that small could cause a problem. Further,  we want to validate that we have an unexpired tombstone. What was there is doesn't account for the extended validity period that is the EXPIRY_TIMEOUT.

geode-core/src/main/java/org/apache/geode/internal/cache/TombstoneService.java
mhansonp: I have added that change.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/versions/TombstoneDUnitTest.java
mhansonp: fixed that.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/versions/TombstoneDUnitTest.java
jchen21: For DUnit test, `cacheTimeMillis()` is returns the same value as `System.currentTimeMillis()`.  To validate an unexpired tombstone, besides asserting getOldestTombstoneTime()  + REPLICATE_TOMBSTONE_TIMEOUT_DEFAULT - now > 0, there is also another condition msUntilTombstoneExpires < EXPIRY_TIME, as suggested in your code change in `hasExpired()`. Anyway, what I have suggested are for improvements.

jchen21 (approved):

nabarunnag (approved):

nabarunnag (approved):
Approving the PR based on 

- Changes requested by me were implemented.
- SMEs have validated and accepted that there are valid conditions in which tombstone TS can be set in the future

PR #6044, opened 2021-02-22
GEODE-8962: add an option to escape dollar($) character in query

mkevo:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/cache/query/internal/CompiledLike.java
+++ b/geode-core/src/main/java/org/apache/geode/cache/query/internal/CompiledLike.java
@@ -314,30 +314,6 @@ public class CompiledLike extends CompiledComparison {
           sb.append(ch);
           break;
         case '_': // replace with .
-        case '$': // replace with [$]
-          if (prevMetaChar) {
-            sb.append('\\');
-            sb.append('E');
-            prevMetaChar = false;
-          }
-
-          int numConsecutiveBackSlash = 0;
-          for (int j = i - 1; j > -1; --j) {
-            if (pattern.charAt(j) == '\\') {
-              ++numConsecutiveBackSlash;
-            } else {
-              break;
-            }
-          }
-
-          // Check if the $ has a valid escape. Backtrack to check for \.
-          // If the number of \ on back track is odd, then $ is escaped.
-          if ((numConsecutiveBackSlash % 2) != 0) {
-            sb.append("[$]");
-          } else {
-            sb.append(ch);
-          }
-          break;
         case '%': // replace with .*
           if (prevMetaChar) {
             sb.append('\\');
@@ -347,7 +323,7 @@ public class CompiledLike extends CompiledComparison {
 
           // Check if the % has a valid escape. Backtrack to check for \.
           // If the number of \ on back track is odd, then % is escaped.
-          numConsecutiveBackSlash = 0;
+          int numConsecutiveBackSlash = 0;
           for (int j = i - 1; j > -1; --j) {
             if (pattern.charAt(j) == '\\') {
               ++numConsecutiveBackSlash;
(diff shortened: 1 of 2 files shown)


geode-dunit/src/main/java/org/apache/geode/management/internal/cli/commands/QueryCommandDUnitTestBase.java
```diff
@@ -141,6 +141,19 @@ public void testSimpleQuery() {
...
+  public void testSimpleQueryWithEscapingCharacter() {
+    server1.invoke(() -> prepareDataForRegionWithSpecialCharacters(DATA_PAR_REGION_NAME_PATH));
+    String query = "query --query=\"select * from " + DATA_PAR_REGION_NAME_PATH
+        + " e where e LIKE 'value\\$'\"";
+    String query1 = "query --query=\"select * from " + DATA_PAR_REGION_NAME_PATH
+        + " e where e LIKE 'value\\%'\"";
+    CommandResult commandResult = gfsh.executeCommand(query);
+    CommandResult commandResult1 = gfsh.executeCommand(query1);
+    validateSelectResult(commandResult, true, 1, null);
+    validateSelectResult(commandResult1, true, 1, null);
```
nabarunnag, line 154:
```suggestion
    gfsh.executeAndAssertThat(query).statusIsSuccess().containsOutput("value$");
    gfsh.executeAndAssertThat(query1).statusIsSuccess().containsOutput("value%");
```

nabarunnag (changes requested):
Request to replace the deprecated calls to the rule.

DonalEvans (approved):

geode-core/src/main/java/org/apache/geode/cache/query/internal/CompiledLike.java
```diff
@@ -349,6 +349,30 @@ private String getRegexPattern(String pattern) {
...
+          numConsecutiveBackSlash = 0;
+          for (int j = i - 1; j > -1; --j) {
+            if (pattern.charAt(j) == '\\') {
+              ++numConsecutiveBackSlash;
+            } else {
+              break;
+            }
+          }
+
+          // Check if the $ has a valid escape. Backtrack to check for \.
```
agingade, line 368: Why "$" needs to be escaped being a literal in like predicate value. The special characters that are used in LIKE or "_", "%" and "\" these chars needs to be escaped with escape char "\". 
I think we should be able to compare the "$" literal within the like predicate without escaping it. Please let me know if I am missing anything.

mkevo: When it use the compare operation it checks string, but when using LIKE operation it check character by character, and when comes to "$" character it present it like endline character and cannot got right results. Now with escaping it will check if string has "$" character and compare it like any other character.

geode-dunit/src/main/java/org/apache/geode/management/internal/cli/commands/QueryCommandDUnitTestBase.java
```diff
@@ -200,6 +212,14 @@ private static void prepareDataForRegion(String regionPath) {
     }
   }
 
+  private static void prepareDataForRegionWithSpecialCharacters(String regionPath) {
+    InternalCache cache = ClusterStartupRule.getCache();
+    Region dataRegion = cache.getRegion(regionPath);
```
kirklund, line 217: Please try to use non-raw types in places like this. There's an IntelliJ inspection you can enable that'll highlight lines like this:
```
Region<Integer, String> dataRegion = cache.getRegion(regionPath);

dataRegion.put(1, "value$");
dataRegion.put(2, "value%");
```

kirklund: Don't worry about the other lines in this old class that use raw types.

kirklund (changes requested):

DonalEvans (approved):

mivanac (approved):

mhansonp (approved):
The test looks very reasonable to me. I am not versed in query, so that part needs  approval.

kirklund (approved):

mkevo:
Hi @nabarunnag, can you please re-review this?

PR #6049, opened 2021-02-23
GEODE-8097: expects at least one callback

jinmeiliao:

### DIFF
--- a/geode-management/src/test/java/org/apache/geode/management/internal/ClientClusterManagementServiceTest.java
+++ b/geode-management/src/test/java/org/apache/geode/management/internal/ClientClusterManagementServiceTest.java
@@ -186,7 +186,8 @@ public class ClientClusterManagementServiceTest {
         service.getFuture(opType, opId);
 
     await().untilAsserted(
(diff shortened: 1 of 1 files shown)


onichols-pivotal:
is this for the test failures where we see like 295 calls instead of the expected 1, and a stacktrace that looks like it's happening via recursion?  I can't help wondering if we should prevent the recursive call rather than change the test to say it's ok?

kirklund (approved):
Looks good. Thanks!

jinmeiliao:
> is this for the test failures where we see like 295 calls instead of the expected 1, and a stacktrace that looks like it's happening via recursion? I can't help wondering if we should prevent the recursive call rather than change the test to say it's ok?

It's not due to recursion. The test calls for a `getFuture` which repeatedly issues the `get` call until the async operation finishes. On a slow machine, there could be more than one `get` call issued.

onichols-pivotal (approved):

jchen21 (approved):

dschneider-pivotal (approved):

jdeppe-pivotal (approved):

PR #6051, opened 2021-02-23
GEODE-8862: Uses another thread to process DistributedCacheOperation

pivotal-eshu:
* Do not process DistributedCacheOperation in-line if scope is DISTRIBUTED_NO_ACK.
  * This is to solve a potential deadlock. The p2p reader thread could be blocked
    on synchronized lock of an entry, and could not handle the DLock GRANT message
    which is needed by another thread holding the synchronized lock.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/DistributedCacheOperation.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/DistributedCacheOperation.java
@@ -1106,15 +1106,8 @@ public abstract class DistributedCacheOperation {
         }
 
         final LocalRegion lclRgn = getLocalRegionForProcessing(dm);
-        if (lclRgn == null) {
-          this.closed = true;
-          if (logger.isDebugEnabled()) {
-            logger.debug("{} region not found, nothing to do", this);
-          }
-          return;
-        }
         sendReply = false;
-        if (lclRgn.getScope().isDistributedNoAck()) {
+        if (lclRgn != null && lclRgn.getScope().isDistributedNoAck()) {
           dm.getExecutors().getWaitingThreadPool().execute(() -> basicProcess(dm, lclRgn));
           return;
         }
@@ -1164,10 +1157,17 @@ public abstract class DistributedCacheOperation {
       }
 
       InitializationLevel oldLevel = ANY_INIT;
-      if (lclRgn.getScope().isDistributedNoAck()) {
+      if (lclRgn != null && lclRgn.getScope().isDistributedNoAck()) {
         oldLevel = LocalRegion.setThreadInitLevelRequirement(BEFORE_INITIAL_IMAGE);
       }
       try {
+        if (lclRgn == null) {
+          this.closed = true;
+          if (logger.isDebugEnabled()) {
+            logger.debug("{} region not found, nothing to do", this);
+          }
+          return;
+        }
         // Could this cause a deadlock, because this can block a P2P reader
         // thread which might be needed to read the create region reply??
         // DAN - I don't think this does anything because process called
@@ -1257,7 +1257,7 @@ public abstract class DistributedCacheOperation {
         SystemFailure.checkFailure();
         thr = t;
       } finally {
-        if (lclRgn.getScope().isDistributedNoAck()) {
+        if (lclRgn != null && lclRgn.getScope().isDistributedNoAck()) {
           LocalRegion.setThreadInitLevelRequirement(oldLevel);
         }
         checkVersionIsRecorded(this.versionTag, lclRgn);
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/DistributedCacheOperation.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/DistributedCacheOperation.java
@@ -1106,7 +1106,18 @@ public abstract class DistributedCacheOperation {
         }
 
         final LocalRegion lclRgn = getLocalRegionForProcessing(dm);
+        if (lclRgn == null) {
+          this.closed = true;
+          if (logger.isDebugEnabled()) {
+            logger.debug("{} region not found, nothing to do", this);
+          }
+          return;
+        }
         sendReply = false;
+        if (lclRgn.getScope().isDistributedNoAck()) {
+          dm.getExecutors().getWaitingThreadPool().execute(() -> basicProcess(dm, lclRgn));
+          return;
+        }
         basicProcess(dm, lclRgn);
       } catch (CancelException ignore) {
         this.closed = true;
@@ -1151,16 +1162,12 @@ public abstract class DistributedCacheOperation {
       if (logger.isTraceEnabled()) {
         logger.trace("DistributedCacheOperation.basicProcess: {}", this);
       }
+
+      InitializationLevel oldLevel = ANY_INIT;
+      if (lclRgn.getScope().isDistributedNoAck()) {
+        oldLevel = LocalRegion.setThreadInitLevelRequirement(BEFORE_INITIAL_IMAGE);
+      }
       try {
-        // LocalRegion lclRgn = getRegionFromPath(dm.getSystem(),
-        // this.regionPath);
-        if (lclRgn == null) {
-          this.closed = true;
-          if (logger.isDebugEnabled()) {
-            logger.debug("{} region not found, nothing to do", this);
-          }
-          return;
-        }
         // Could this cause a deadlock, because this can block a P2P reader
         // thread which might be needed to read the create region reply??
         // DAN - I don't think this does anything because process called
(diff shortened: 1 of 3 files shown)


geode-core/src/test/java/org/apache/geode/internal/cache/DistributedCacheOperationTest.java
```diff
@@ -17,22 +17,68 @@
...
+  private OperationExecutors executors;
+  private final int processorId = 1;
+
+  @Before
+  public void setup() {
+    message = spy(new TestCacheOperationMessage());
+    sender = mock(InternalDistributedMember.class);
+    versionTag = mock(VersionTag.class);
+
+    GemFireCacheImpl cache = Fakes.cache();
```
kirklund, line 67: I'd really like to see us stop using `Fakes.cache()` in new unit tests so we can get rid of it. Can you please see if you can remove this and mock everything `DistributedCacheOperation` needs directly? Also, please try to use `InternalCache` and `InternalRegion` instead of `GemFireCacheImpl` and `LocalRegion`. Let me know if you'd like some help or want me to pair with you.

kirklund (changes requested):

kirklund:
The changes look great. The only change I'd really like to see is to avoid the use of `Fakes.cache()`.

geode-core/src/test/java/org/apache/geode/internal/cache/DistributedCacheOperationTest.java
pivotal-eshu: Removed the Fakes.
As the product code method signature is using LocalRegion, using InternalRegion may need  some casting. Will leave it as is, unless we need to change the product code.

kirklund (approved):

geode-core/src/test/java/org/apache/geode/internal/cache/DistributedCacheOperationTest.java
```diff
@@ -72,11 +114,112 @@ public void endOperationIsInvokedOnDistributionError() {
...
+
+  @Test
+  public void processSendsReplyIfAdminDM() {
+    when(dm.getDMType()).thenReturn(ClusterDistributionManager.ADMIN_ONLY_DM_TYPE);
+
+    message.process(dm);
+
+    verify(message, never()).basicProcess(dm, region);
+    verify(message).sendReply(
+        eq(sender),
```
DonalEvans, line 132: It's not necessary to wrap the arguments here with `eq()`, so this can be simplified.

geode-core/src/test/java/org/apache/geode/internal/cache/DistributedCacheOperationTest.java
```diff
@@ -72,11 +114,112 @@ public void endOperationIsInvokedOnDistributionError() {
...
+  @Test
+  public void processSendsReplyIfLocalRegionIsNull() {
+    doReturn(null).when(message).getLocalRegionForProcessing(dm);
+
+    message.process(dm);
+
+    assertThat(message.closed).isTrue();
+    verify(message, never()).basicProcess(dm, region);
+    verify(message).sendReply(
+        eq(sender),
```
DonalEvans, line 147: It's not necessary to wrap the arguments here with `eq()`, so this can be simplified.

geode-core/src/test/java/org/apache/geode/internal/cache/DistributedCacheOperationTest.java
```diff
@@ -72,11 +114,112 @@ public void endOperationIsInvokedOnDistributionError() {
...
+  }
+
+  @Test
+  public void processSendsReplyIfGotCacheClosedException() {
+    CacheClosedException cacheClosedException = new CacheClosedException();
+    doThrow(cacheClosedException).when(message).getLocalRegionForProcessing(dm);
+
+    message.process(dm);
+
+    assertThat(message.closed).isTrue();
```
DonalEvans, line 160: This would be simpler as `assertTrue()`.

geode-core/src/test/java/org/apache/geode/internal/cache/DistributedCacheOperationTest.java
```diff
@@ -72,11 +114,112 @@ public void endOperationIsInvokedOnDistributionError() {
...
+    verify(message, never()).basicProcess(dm, region);
+    verify(message).sendReply(
+        eq(sender),
```
DonalEvans, line 163: It's not necessary to use `eq()` here.

geode-core/src/test/java/org/apache/geode/internal/cache/DistributedCacheOperationTest.java
```diff
@@ -72,11 +114,112 @@ public void endOperationIsInvokedOnDistributionError() {
...
+  }
+
+  @Test
+  public void processDoesNotSendReplyIfDistributedNoAck() {
+    when(scope.isDistributedNoAck()).thenReturn(true);
+
+    message.process(dm);
+
+    verify(message, never()).sendReply(
+        eq(sender),
```
DonalEvans, line 212: It's not necessary to use `eq()` for these arguments.

DonalEvans (changes requested):

DonalEvans (commented):
Just a few small code clean up requests. Also, it's outside the changes in this PR, but the compiler warning on line 222 in DistributedCacheOperationTest can be fixed by changing it to `TestOperation(CacheEvent<?, ?> event) {`

It also seems like these changes are causing some DUnit tests to fail, so I'll hold off on approving until that's been resolved.

geode-core/src/test/java/org/apache/geode/internal/cache/DistributedCacheOperationTest.java
pivotal-eshu: I believe there was a requirement that all new tests should use assertj methods.

DonalEvans (approved):

geode-core/src/main/java/org/apache/geode/internal/cache/DistributedCacheOperation.java
```diff
@@ -1107,6 +1107,10 @@ protected void process(final ClusterDistributionManager dm) {
 
         final LocalRegion lclRgn = getLocalRegionForProcessing(dm);
         sendReply = false;
+        if (lclRgn != null && lclRgn.getScope().isDistributedNoAck()) {
+          dm.getExecutors().getWaitingThreadPool().execute(() -> basicProcess(dm, lclRgn));
```
bschuchardt, line 1111: this previously caught CancelException and set this.closed=true.  That's no longer being done for distributed-no-ack regions.  Why is that okay?

bschuchardt (changes requested):

geode-core/src/main/java/org/apache/geode/internal/cache/DistributedCacheOperation.java
jchen21: I have the same concern. When a thread runs, if it throws an exception that is not caught, then that exception will just silently kill the thread. This can make it very hard to diagnose what is happening. The current thread has a few `catch` blocks and a `finally` block (line 1115-1147). If the newly spawned thread does not have the `catch` and `finally` blocks, I would like to understand why.

Another question is why use `getWaitingThreadPool()`  and not the other pool, e.g. `getThreadPool()`?

pivotal-eshu: In basicProcess call (using waiting thread pool for distributed-no-ack regions), CancelException is also caught and handled.
catch (RegionDestroyedException ignore) {
        this.closed = true;
          logger.debug("{} Region destroyed: nothing to do", this);
        }
      } catch (CancelException ignore) {
        this.closed = true;
          logger.debug("{} Cancelled: nothing to do", this);
        }
      }

jchen21: Good point. For `basicProcess()`, if debug is not enabled, some exceptions are caught, but not logged, although some flags are set, when the exceptions are caught. Will the new thread fail silently with such exceptions, if debug is not enabled?
And even for `process()`, logging the exceptions also depends on whether debug is enabled. I am not sure whether this is by design. If debug is not enabled, without the log message, it is hard to analyze the failures.

pivotal-eshu: I believe it is intentional. We do not want to pollute the user logging.

bschuchardt (approved):

mhansonp (approved):

geode-core/src/main/java/org/apache/geode/internal/cache/DistributedCacheOperation.java
gesterzhou: But for this corner case, I think an info level log is necessary.

pivotal-eshu:
As this fix causes some behavior change -- will try out another fix.

PR #6052, opened 2021-02-24
GEODE-8971: Add grace period when stopping gateway sender with group-…

albertogpz:
…transaction-events enabled

When the gateway sender is stopped there is a high probability that batches with
incomplete transactions are sent even if group-transaction-events is enabled.

The reason is that once the stop command reaches the gateway sender, it immediately
stops queueing events, and this could happen in the middle of receiving events for the
same transaction. If this is the case, some events for the transaction may have
reached the queue right before the stop command was received and the rest of events
for that transaction would not make it to the queue (they would be dropped) because
they arrived right after the stop command was received at the gateway sender.

One way to solve this issue is to allow for a grace period when the gateway sender
is stopped during which only events that belong to transactions already
in the queue are queued. Any other event received by the gateway sender
during this grace period would be dropped. After this grace period
the gateway sender will be effectively stopped.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/cache/wan/GatewaySender.java
+++ b/geode-core/src/main/java/org/apache/geode/cache/wan/GatewaySender.java
@@ -174,7 +174,32 @@ public interface GatewaySender {
    */
   int GET_TRANSACTION_EVENTS_FROM_QUEUE_RETRIES =
       Integer.getInteger(GeodeGlossary.GEMFIRE_PREFIX + "get-transaction-events-from-queue-retries",
-          10);
+          2);
+  /**
+   * Milliseconds to wait before retrying to get events for a transaction from the
+   * gateway sender queue when group-transaction-events is true.
+   */
+  int GET_TRANSACTION_EVENTS_FROM_QUEUE_WAIT_TIME_MS =
+      Integer.getInteger(
+          GeodeGlossary.GEMFIRE_PREFIX + "get-transaction-events-from-queue-wait-time-ms",
+          1);
+
+  /**
+   * When group-transaction-events is set to true and the gateway sender is stopped,
+   * there is a possibility that the stopping occurs such that for a transaction,
+   * not all events belonging to it reach the queue. The reason would be that
+   * some reach the queue right before the sender is stopped and the rest do not make
+   * it to the queue because the sender is just stopped.
+   * In order to prevent that the queue contains incomplete transactions
+   * due to the above circumstance, this parameter allows for a grace period
+   * of the number of milliseconds set in it before the gateway sender is
+   * actually stopped, in which only events to complete transactions are put in the queue.
+   * Other events received in this period would be dropped.
+   */
+  int TIME_TO_COMPLETE_TRANSACTIONS_BEFORE_STOP_MS =
+      Integer.getInteger(
+          GeodeGlossary.GEMFIRE_PREFIX + "time-to-complete-transactions-before-stop-ms",
+          1000);
 
   /**
    * The order policy. This enum is applicable only when concurrency-level is > 1.
--- a/geode-core/src/main/java/org/apache/geode/cache/wan/GatewaySender.java
+++ b/geode-core/src/main/java/org/apache/geode/cache/wan/GatewaySender.java
@@ -185,16 +185,13 @@ public interface GatewaySender {
           1);
 
   /**
-   * When group-transaction-events is set to true and the gateway sender is stopped,
-   * there is a possibility that the stopping occurs such that for a transaction,
-   * not all events belonging to it reach the queue. The reason would be that
-   * some reach the queue right before the sender is stopped and the rest do not make
-   * it to the queue because the sender is just stopped.
-   * In order to prevent that the queue contains incomplete transactions
-   * due to the above circumstance, this parameter allows for a grace period
-   * of the number of milliseconds set in it before the gateway sender is
-   * actually stopped, in which only events to complete transactions are put in the queue.
-   * Other events received in this period would be dropped.
+   * When group-transaction-events is true and the gateway sender is stopped,
+   * addition to the queue of a group of transaction events might be interrupted.
+   * To ensure that the queue does not contain incomplete transactions, this parameter
+   * allows for a grace period, specified in milliseconds, before the gateway sender is actually
+   * stopped, allowing complete transaction event groups to be queued. Any event received
+   * during the grace period that is not part of a transaction event group in the queue
+   * is dropped.
    */
   int TIME_TO_COMPLETE_TRANSACTIONS_BEFORE_STOP_MS =
       Integer.getInteger(
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/BucketRegionQueue.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/BucketRegionQueue.java
@@ -454,8 +454,8 @@ public class BucketRegionQueue extends AbstractBucketRegionQueue {
    * If a matching object also fulfills the endPredicate then the method
    * stops looking for more matching objects.
    */
-  public List<Object> getElementsMatching(Predicate<GatewaySenderEventImpl> matchingPredicate,
-      Predicate<GatewaySenderEventImpl> endPredicate) {
+  public List<Object> getElementsMatching(Predicate<InternalGatewayQueueEvent> matchingPredicate,
+      Predicate<InternalGatewayQueueEvent> endPredicate) {
     getInitializationLock().readLock().lock();
     try {
       if (this.getPartitionedRegion().isDestroyed()) {
@@ -466,10 +466,10 @@ public class BucketRegionQueue extends AbstractBucketRegionQueue {
       while (it.hasNext()) {
         Object key = it.next();
         Object object = optimalGet(key);
-        if (matchingPredicate.test((GatewaySenderEventImpl) object)) {
+        if (matchingPredicate.test((InternalGatewayQueueEvent) object)) {
           elementsMatching.add(object);
           this.eventSeqNumDeque.remove(key);
-          if (endPredicate.test((GatewaySenderEventImpl) object)) {
+          if (endPredicate.test((InternalGatewayQueueEvent) object)) {
             break;
           }
         }
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/BucketRegionQueue.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/BucketRegionQueue.java
@@ -37,6 +37,7 @@ import org.apache.geode.cache.EntryNotFoundException;
 import org.apache.geode.cache.RegionAttributes;
 import org.apache.geode.cache.RegionDestroyedException;
 import org.apache.geode.cache.TimeoutException;
+import org.apache.geode.cache.wan.GatewayQueueEvent;
 import org.apache.geode.internal.cache.execute.BucketMovedException;
 import org.apache.geode.internal.cache.persistence.query.mock.ByteComparator;
 import org.apache.geode.internal.cache.versions.RegionVersionVector;
@@ -478,16 +479,15 @@ public class BucketRegionQueue extends AbstractBucketRegionQueue {
     }
   }
 
-  public boolean isThereEventsMatching(Predicate matchingPredicate) {
+  public boolean isThereEventsMatching(Predicate<GatewayQueueEvent<?, ?>> matchingPredicate) {
     getInitializationLock().readLock().lock();
     try {
       if (this.getPartitionedRegion().isDestroyed()) {
         throw new BucketRegionQueueUnavailableException();
       }
-      Iterator<Object> it = this.eventSeqNumDeque.iterator();
-      while (it.hasNext()) {
-        Object object = optimalGet(it.next());
-        if (matchingPredicate.test(object)) {
+      for (Object o : eventSeqNumDeque) {
+        Object object = optimalGet(o);
+        if (matchingPredicate.test((GatewayQueueEvent<?, ?>) object)) {
           return true;
         }
       }
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/BucketRegionQueue.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/BucketRegionQueue.java
@@ -478,6 +478,25 @@ public class BucketRegionQueue extends AbstractBucketRegionQueue {
     }
   }
 
+  public boolean isThereEventsMatching(Predicate matchingPredicate) {
+    getInitializationLock().readLock().lock();
+    try {
+      if (this.getPartitionedRegion().isDestroyed()) {
+        throw new BucketRegionQueueUnavailableException();
+      }
+      Iterator<Object> it = this.eventSeqNumDeque.iterator();
+      while (it.hasNext()) {
+        Object object = optimalGet(it.next());
+        if (matchingPredicate.test(object)) {
+          return true;
+        }
+      }
+      return false;
+    } finally {
+      getInitializationLock().readLock().unlock();
+    }
+  }
+
   @Override
   protected void addToEventQueue(Object key, boolean didPut, EntryEventImpl event) {
     if (didPut) {
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/BucketRegionQueue.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/BucketRegionQueue.java
@@ -37,13 +37,13 @@ import org.apache.geode.cache.EntryNotFoundException;
 import org.apache.geode.cache.RegionAttributes;
 import org.apache.geode.cache.RegionDestroyedException;
 import org.apache.geode.cache.TimeoutException;
-import org.apache.geode.cache.wan.GatewayQueueEvent;
 import org.apache.geode.internal.cache.execute.BucketMovedException;
 import org.apache.geode.internal.cache.persistence.query.mock.ByteComparator;
 import org.apache.geode.internal.cache.versions.RegionVersionVector;
 import org.apache.geode.internal.cache.versions.VersionSource;
 import org.apache.geode.internal.cache.wan.AbstractGatewaySenderEventProcessor;
 import org.apache.geode.internal.cache.wan.GatewaySenderEventImpl;
+import org.apache.geode.internal.cache.wan.InternalGatewayQueueEvent;
 import org.apache.geode.internal.cache.wan.parallel.BucketRegionQueueUnavailableException;
 import org.apache.geode.internal.cache.wan.parallel.ConcurrentParallelGatewaySenderQueue;
 import org.apache.geode.internal.concurrent.Atomics;
@@ -454,21 +454,22 @@ public class BucketRegionQueue extends AbstractBucketRegionQueue {
    * If a matching object also fulfills the endPredicate then the method
    * stops looking for more matching objects.
    */
-  public List<Object> getElementsMatching(Predicate matchingPredicate, Predicate endPredicate) {
+  public List<Object> getElementsMatching(Predicate<GatewaySenderEventImpl> matchingPredicate,
+      Predicate<GatewaySenderEventImpl> endPredicate) {
     getInitializationLock().readLock().lock();
     try {
       if (this.getPartitionedRegion().isDestroyed()) {
         throw new BucketRegionQueueUnavailableException();
       }
-      List<Object> elementsMatching = new ArrayList<Object>();
+      List<Object> elementsMatching = new ArrayList();
       Iterator<Object> it = this.eventSeqNumDeque.iterator();
       while (it.hasNext()) {
         Object key = it.next();
         Object object = optimalGet(key);
-        if (matchingPredicate.test(object)) {
+        if (matchingPredicate.test((GatewaySenderEventImpl) object)) {
           elementsMatching.add(object);
           this.eventSeqNumDeque.remove(key);
-          if (endPredicate.test(object)) {
+          if (endPredicate.test((GatewaySenderEventImpl) object)) {
             break;
           }
         }
@@ -479,7 +480,7 @@ public class BucketRegionQueue extends AbstractBucketRegionQueue {
     }
   }
 
-  public boolean isThereEventsMatching(Predicate<GatewayQueueEvent<?, ?>> matchingPredicate) {
+  public boolean isThereEventsMatching(Predicate<InternalGatewayQueueEvent> matchingPredicate) {
     getInitializationLock().readLock().lock();
     try {
       if (this.getPartitionedRegion().isDestroyed()) {
@@ -487,7 +488,7 @@ public class BucketRegionQueue extends AbstractBucketRegionQueue {
       }
       for (Object o : eventSeqNumDeque) {
         Object object = optimalGet(o);
-        if (matchingPredicate.test((GatewayQueueEvent<?, ?>) object)) {
+        if (matchingPredicate.test((InternalGatewayQueueEvent) object)) {
           return true;
         }
       }
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/BucketRegionQueue.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/BucketRegionQueue.java
@@ -461,15 +461,18 @@ public class BucketRegionQueue extends AbstractBucketRegionQueue {
       if (this.getPartitionedRegion().isDestroyed()) {
         throw new BucketRegionQueueUnavailableException();
       }
-      List<Object> elementsMatching = new ArrayList();
+      List<Object> elementsMatching = new ArrayList<>();
       Iterator<Object> it = this.eventSeqNumDeque.iterator();
       while (it.hasNext()) {
         Object key = it.next();
-        Object object = optimalGet(key);
-        if (matchingPredicate.test((InternalGatewayQueueEvent) object)) {
-          elementsMatching.add(object);
+        Object event = optimalGet(key);
+        if (!(event instanceof InternalGatewayQueueEvent)) {
+          continue;
+        }
+        if (matchingPredicate.test((InternalGatewayQueueEvent) event)) {
+          elementsMatching.add(event);
           this.eventSeqNumDeque.remove(key);
-          if (endPredicate.test((InternalGatewayQueueEvent) object)) {
+          if (endPredicate.test((InternalGatewayQueueEvent) event)) {
             break;
           }
         }
@@ -480,15 +483,18 @@ public class BucketRegionQueue extends AbstractBucketRegionQueue {
     }
   }
 
-  public boolean isThereEventsMatching(Predicate<InternalGatewayQueueEvent> matchingPredicate) {
+  public boolean hasEventsMatching(Predicate<InternalGatewayQueueEvent> matchingPredicate) {
     getInitializationLock().readLock().lock();
     try {
       if (this.getPartitionedRegion().isDestroyed()) {
         throw new BucketRegionQueueUnavailableException();
       }
       for (Object o : eventSeqNumDeque) {
-        Object object = optimalGet(o);
-        if (matchingPredicate.test((InternalGatewayQueueEvent) object)) {
+        Object event = optimalGet(o);
+        if (!(event instanceof InternalGatewayQueueEvent)) {
+          continue;
+        }
+        if (matchingPredicate.test((InternalGatewayQueueEvent) event)) {
           return true;
         }
       }
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/EntryEventImpl.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/EntryEventImpl.java
@@ -2231,6 +2231,9 @@ public class EntryEventImpl implements InternalEntryEvent, InternalCacheEvent,
     if (this.tailKey != -1) {
       buf.append(";tailKey=" + tailKey);
     }
+    if (this.isTransactional()) {
+      buf.append(";transactionId=" + getTransactionId());
+    }
     buf.append("]");
     return buf.toString();
   }
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/EntryEventImpl.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/EntryEventImpl.java
@@ -2229,10 +2229,10 @@ public class EntryEventImpl implements InternalEntryEvent, InternalCacheEvent,
       buf.append(";inhibitDistribution");
     }
     if (this.tailKey != -1) {
-      buf.append(";tailKey=" + tailKey);
+      buf.append(";tailKey=").append(tailKey);
     }
     if (this.isTransactional()) {
-      buf.append(";transactionId=" + getTransactionId());
+      buf.append(";transactionId=").append(getTransactionId());
     }
     buf.append("]");
     return buf.toString();
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySender.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySender.java
@@ -1108,7 +1108,7 @@ public abstract class AbstractGatewaySender implements InternalGatewaySender, Di
           // Get substitution value to enqueue if necessary
           Object substituteValue = getSubstituteValue(clonedEvent, operation);
 
-          Predicate<GatewayQueueEvent<?, ?>> hasSameTransactionIdPredicate = null;
+          Predicate<InternalGatewayQueueEvent> hasSameTransactionIdPredicate = null;
           // In case the sender is about to be stopped, the event will only
           // be queued if there is any event in the queue with the same
           // transactionId as the one of this event
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySender.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySender.java
@@ -1108,7 +1108,7 @@ public abstract class AbstractGatewaySender implements InternalGatewaySender, Di
           // Get substitution value to enqueue if necessary
           Object substituteValue = getSubstituteValue(clonedEvent, operation);
 
-          Predicate hasSameTransactionIdPredicate = null;
+          Predicate<GatewayQueueEvent<?, ?>> hasSameTransactionIdPredicate = null;
           // In case the sender is about to be stopped, the event will only
           // be queued if there is any event in the queue with the same
           // transactionId as the one of this event
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySender.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySender.java
@@ -690,7 +690,7 @@ public abstract class AbstractGatewaySender implements InternalGatewaySender, Di
   }
 
   protected void preStop() {
-    if (!mustGroupTransactionEvents() || isStopping) {
+    if (!mustGroupTransactionEvents()) {
       return;
     }
     isStopping = true;
@@ -1115,9 +1115,9 @@ public abstract class AbstractGatewaySender implements InternalGatewaySender, Di
           if (isStopping && mustGroupTransactionEvents()
               && clonedEvent.getTransactionId() != null) {
             hasSameTransactionIdPredicate =
-                x -> x instanceof GatewaySenderEventImpl && clonedEvent.getTransactionId() != null
+                x -> x != null && clonedEvent.getTransactionId() != null
                     && clonedEvent.getTransactionId()
-                        .equals(((GatewaySenderEventImpl) x).getTransactionId());
+                        .equals((x).getTransactionId());
           }
           if (!ev.enqueueEvent(operation, clonedEvent, substituteValue, isLastEventInTransaction,
               hasSameTransactionIdPredicate)) {
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySender.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySender.java
@@ -25,6 +25,7 @@ import java.util.Set;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.function.Predicate;
 
 import org.apache.logging.log4j.Logger;
 
@@ -125,6 +126,8 @@ public abstract class AbstractGatewaySender implements InternalGatewaySender, Di
   protected boolean isParallel;
 
   protected boolean groupTransactionEvents;
+
+  protected volatile boolean isStopping = false;
 
   protected boolean isForInternalUse;
 
@@ -686,6 +689,22 @@ public abstract class AbstractGatewaySender implements InternalGatewaySender, Di
     return enqueue;
   }
 
+  protected void preStop() {
+    if (!mustGroupTransactionEvents() || isStopping) {
+      return;
+    }
+    isStopping = true;
+    try {
+      Thread.sleep(TIME_TO_COMPLETE_TRANSACTIONS_BEFORE_STOP_MS);
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
+    }
+  }
+
+  protected void postStop() {
+    isStopping = false;
+  }
+
   protected void stopProcessing() {
     // Stop the dispatcher
     AbstractGatewaySenderEventProcessor ev = this.eventProcessor;
@@ -1041,14 +1060,8 @@ public abstract class AbstractGatewaySender implements InternalGatewaySender, Di
         clonedEvent.setCallbackArgument(geCallbackArg);
       }
 
-      // If this gateway is not running, return
       if (!isRunning()) {
-        if (this.isPrimary()) {
-          recordDroppedEvent(clonedEvent);
-        }
-        if (isDebugEnabled) {
-          logger.debug("Returning back without putting into the gateway sender queue:" + event);
-        }
+        recordDroppedEvent(clonedEvent);
         return;
       }
 
@@ -1075,12 +1088,7 @@ public abstract class AbstractGatewaySender implements InternalGatewaySender, Di
         // If this gateway is not running, return
         // The sender may have stopped, after we have checked the status in the beginning.
         if (!isRunning()) {
-          if (isDebugEnabled) {
-            logger.debug("Returning back without putting into the gateway sender queue:" + event);
-          }
-          if (this.isPrimary()) {
-            recordDroppedEvent(clonedEvent);
-          }
+          recordDroppedEvent(clonedEvent);
           return;
         }
 
@@ -1100,7 +1108,21 @@ public abstract class AbstractGatewaySender implements InternalGatewaySender, Di
           // Get substitution value to enqueue if necessary
           Object substituteValue = getSubstituteValue(clonedEvent, operation);
 
-          ev.enqueueEvent(operation, clonedEvent, substituteValue, isLastEventInTransaction);
+          Predicate hasSameTransactionIdPredicate = null;
+          // In case the sender is about to be stopped, the event will only
+          // be queued if there is any event in the queue with the same
+          // transactionId as the one of this event
+          if (isStopping && mustGroupTransactionEvents()
+              && clonedEvent.getTransactionId() != null) {
+            hasSameTransactionIdPredicate =
+                x -> x instanceof GatewaySenderEventImpl && clonedEvent.getTransactionId() != null
+                    && clonedEvent.getTransactionId()
+                        .equals(((GatewaySenderEventImpl) x).getTransactionId());
+          }
+          if (!ev.enqueueEvent(operation, clonedEvent, substituteValue, isLastEventInTransaction,
+              hasSameTransactionIdPredicate)) {
+            recordDroppedEvent(event);
+          }
         } catch (CancelException e) {
           logger.debug("caught cancel exception", e);
           throw e;
@@ -1126,6 +1148,9 @@ public abstract class AbstractGatewaySender implements InternalGatewaySender, Di
   }
 
   private void recordDroppedEvent(EntryEventImpl event) {
+    if (!this.isPrimary()) {
+      return;
+    }
     if (this.eventProcessor != null) {
       this.eventProcessor.registerEventDroppedInPrimaryQueue(event);
     } else {
@@ -1133,6 +1158,9 @@ public abstract class AbstractGatewaySender implements InternalGatewaySender, Di
       if (logger.isDebugEnabled()) {
         logger.debug("added to tmpDroppedEvents event: {}", event);
       }
+    }
+    if (logger.isDebugEnabled()) {
+      logger.debug("Returning without putting into the gateway sender queue:" + event);
     }
   }
 
@@ -1530,7 +1558,7 @@ public abstract class AbstractGatewaySender implements InternalGatewaySender, Di
       try {
         logger.info("{}: Enqueueing synchronization event: {}",
             this, event);
-        this.eventProcessor.enqueueEvent(event);
+        this.eventProcessor.enqueueEvent(event, null);
         this.statistics.incSynchronizationEventsEnqueued();
       } catch (Throwable t) {
         logger.warn(String.format(
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySender.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySender.java
@@ -690,14 +690,13 @@ public abstract class AbstractGatewaySender implements InternalGatewaySender, Di
   }
 
   protected void preStop() {
-    if (!mustGroupTransactionEvents()) {
-      return;
-    }
     isStopping = true;
-    try {
-      Thread.sleep(TIME_TO_COMPLETE_TRANSACTIONS_BEFORE_STOP_MS);
-    } catch (InterruptedException e) {
-      Thread.currentThread().interrupt();
+    if (mustGroupTransactionEvents()) {
+      try {
+        Thread.sleep(TIME_TO_COMPLETE_TRANSACTIONS_BEFORE_STOP_MS);
+      } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
+      }
     }
   }
 
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySenderEventProcessor.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySenderEventProcessor.java
@@ -26,6 +26,7 @@ import java.util.Objects;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Predicate;
 
 import org.apache.logging.log4j.Logger;
 
@@ -173,14 +174,25 @@ public abstract class AbstractGatewaySenderEventProcessor extends LoggingThread
 
   protected abstract void initializeMessageQueue(String id, boolean cleanQueues);
 
-  public void enqueueEvent(EnumListenerEvent operation, EntryEvent event,
+  public boolean enqueueEvent(EnumListenerEvent operation, EntryEvent event,
       Object substituteValue) throws IOException, CacheException {
-    enqueueEvent(operation, event, substituteValue, false);
+    return enqueueEvent(operation, event, substituteValue, false, null);
   }
 
-  public abstract void enqueueEvent(EnumListenerEvent operation, EntryEvent event,
-      Object substituteValue, boolean isLastEventInTransaction) throws IOException, CacheException;
-
+  /**
+   *
+   * @param operation The operation
+   * @param event The event to be put in the queue
+   * @param substituteValue The substitute value
+   * @param isLastEventInTransaction True if this event is the last one in the
+   *        transaction it belongs to
+   * @param condition If not null, the event will be enqueued only if at least
+   *        one element in the queue matches the predicate
+   * @return False only if the condition is not null and no element in the queue matches it
+   */
+  public abstract boolean enqueueEvent(EnumListenerEvent operation, EntryEvent event,
+      Object substituteValue, boolean isLastEventInTransaction,
+      Predicate condition) throws IOException, CacheException;
 
   protected abstract void rebalance();
 
@@ -194,6 +206,9 @@ public abstract class AbstractGatewaySenderEventProcessor extends LoggingThread
       this.failureLogInterval.clear();
     } else {
       this.isStopped = isStopped;
+    }
+    if (isStopped) {
+      sender.postStop();
     }
   }
 
@@ -1386,7 +1401,8 @@ public abstract class AbstractGatewaySenderEventProcessor extends LoggingThread
     }
   }
 
-  protected abstract void enqueueEvent(GatewayQueueEvent event);
+  protected abstract boolean enqueueEvent(GatewayQueueEvent event,
+      Predicate condition);
 
   protected class SenderStopperCallable implements Callable<Boolean> {
     private final AbstractGatewaySenderEventProcessor p;
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySenderEventProcessor.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySenderEventProcessor.java
@@ -201,14 +201,9 @@ public abstract class AbstractGatewaySenderEventProcessor extends LoggingThread
   }
 
   protected void setIsStopped(boolean isStopped) {
+    this.isStopped = isStopped;
     if (isStopped) {
-      this.isStopped = true;
       this.failureLogInterval.clear();
-    } else {
-      this.isStopped = isStopped;
-    }
-    if (isStopped) {
-      sender.postStop();
     }
   }
 
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySenderEventProcessor.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySenderEventProcessor.java
@@ -192,7 +192,7 @@ public abstract class AbstractGatewaySenderEventProcessor extends LoggingThread
    */
   public abstract boolean enqueueEvent(EnumListenerEvent operation, EntryEvent event,
       Object substituteValue, boolean isLastEventInTransaction,
-      Predicate condition) throws IOException, CacheException;
+      Predicate<GatewayQueueEvent<?, ?>> condition) throws IOException, CacheException;
 
   protected abstract void rebalance();
 
@@ -1402,7 +1402,7 @@ public abstract class AbstractGatewaySenderEventProcessor extends LoggingThread
   }
 
   protected abstract boolean enqueueEvent(GatewayQueueEvent event,
-      Predicate condition);
+      Predicate<GatewayQueueEvent<?, ?>> condition);
 
   protected class SenderStopperCallable implements Callable<Boolean> {
     private final AbstractGatewaySenderEventProcessor p;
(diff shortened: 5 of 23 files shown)


geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
```diff
@@ -1797,6 +1816,22 @@ public static String getSenderId(String regionName) {
     return regionName.substring(1, queueStringStart);
   }
 
+  public boolean isThereEventsMatching(Object object, Predicate condition) {
```
kirklund, line 1819: If you know that `object` is always a `GatewaySenderEventImpl` then it would be better to use that in the method signature:
```
public boolean isThereEventsMatching(GatewaySenderEventImpl value, Predicate<Object> condition) {
```
...rather than casting it on the 1st line of the method.

kirklund (changes requested):
I've requested a few changes in-line and a couple general comments that aren't requests for changes.

A good rule is to always avoid raw types (`Predicate condition`) by providing a type (`Predicate<Object> condition`).

I really want to see us moving away from passing around impl classes like `GatewaySenderEventImpl` if an interface like `GatewayQueueEvent` defines the methods you need to call. If you need to call an impl method that's NOT on the interface, then use the impl for now. We'll probably eventually create a new internal interface like `InternalGatewayQueueEvent` if we find that we need to pass around the impl to call non-interface-defined methods. This will help us avoid concrete dependencies which then helps produce higher quality code and unit tests.

Looks like the existing `throw` statements for `BucketRegionQueueUnavailableException` all have no message. It's always better to provide a message, but it's ok as is since it's at least consistent.

albertogpz:
> I've requested a few changes in-line and a couple general comments that aren't requests for changes.
> 
> A good rule is to always avoid raw types (`Predicate condition`) by providing a type (`Predicate<Object> condition`).
> 
> I really want to see us moving away from passing around impl classes like `GatewaySenderEventImpl` if an interface like `GatewayQueueEvent` defines the methods you need to call. If you need to call an impl method that's NOT on the interface, then use the impl for now. We'll probably eventually create a new internal interface like `InternalGatewayQueueEvent` if we find that we need to pass around the impl to call non-interface-defined methods. This will help us avoid concrete dependencies which then helps produce higher quality code and unit tests.
> 
> Looks like the existing `throw` statements for `BucketRegionQueueUnavailableException` all have no message. It's always better to provide a message, but it's ok as is since it's at least consistent with the other lines that throw it.



> I've requested a few changes in-line and a couple general comments that aren't requests for changes.
> 
> A good rule is to always avoid raw types (`Predicate condition`) by providing a type (`Predicate<Object> condition`).
> 
> I really want to see us moving away from passing around impl classes like `GatewaySenderEventImpl` if an interface like `GatewayQueueEvent` defines the methods you need to call. If you need to call an impl method that's NOT on the interface, then use the impl for now. We'll probably eventually create a new internal interface like `InternalGatewayQueueEvent` if we find that we need to pass around the impl to call non-interface-defined methods. This will help us avoid concrete dependencies which then helps produce higher quality code and unit tests.
> 
> Looks like the existing `throw` statements for `BucketRegionQueueUnavailableException` all have no message. It's always better to provide a message, but it's ok as is since it's at least consistent with the other lines that throw it.

I have added the isLastEventInTransaction() and getTransactionId() to the GatewayQueueEvent interface. Do you see any drawback to it?

kirklund (changes requested):
Unless we have a real need to expose `TransactionId getTransactionId()` and `boolean isLastEventInTransaction()` to the User (it's in a non-internal package), then I would **_not_** add them to the `GatewayQueueEvent`.

I would create a new `InternalGatewayQueueEvent` interface that extends `GatewayQueueEvent` and add the two tx related methods to that internal interface instead. Then you could define the predicate as `Predicate<GatewayQueueEvent>`.

Adding `InternalGatewayQueueEvent` would probably require some follow-up PRs that update the WAN code to use it internally. This is part of long-term refactoring to improve the codebaser with interfaces and is similar to the many other internal interfaces that we've been adding including `InternalRegion` and `InternalCache`. Note that we tend to not parameterize these internal interfaces.

albertogpz:
> Unless we have a real need to expose `TransactionId getTransactionId()` and `boolean isLastEventInTransaction()` to the User (it's in a non-internal package), then I would **_not_** add them to the `GatewayQueueEvent`.
> 
> I would create a new `InternalGatewayQueueEvent` interface that extends `GatewayQueueEvent` and add the two tx related methods to that internal interface instead. Then you could define the predicate as `Predicate<InternalGatewayQueueEvent>`.
> 
> Adding `InternalGatewayQueueEvent` would probably require some follow-up PRs that update the WAN code to use it internally. This would them be part of longer term refactoring to improve the codebase with interfaces and is similar to the many other internal interfaces that we've been adding including `InternalRegion` and `InternalCache`. Note that we tend to not parameterize these internal interfaces.

Thanks for the comments. I have done as suggested.

geode-core/src/main/java/org/apache/geode/internal/cache/BucketRegionQueue.java
```diff
@@ -453,21 +454,22 @@ public Object peek() {
    * If a matching object also fulfills the endPredicate then the method
    * stops looking for more matching objects.
    */
-  public List<Object> getElementsMatching(Predicate matchingPredicate, Predicate endPredicate) {
+  public List<Object> getElementsMatching(Predicate<GatewaySenderEventImpl> matchingPredicate,
```
kirklund, line 457: You might want to change from `GatewaySenderEventImpl` to your new `InternalGatewaySenderEvent` everywhere you've added new code. I don't see anything that would obviously break.

kirklund (commented):
I'll just add a quick in-line comment about code that I'm in the CODEOWNERS list for.

The WAN code changes you have seem ok to me but will need more review by others. I will say the `Thread.sleep` calls make me nervous. The ones in DUnit tests tend to result in flaky test that fail intermittently, and I'm not sure why you added a sleep to the preStop in product code. Others might bring those up, so I'm just giving you a heads up to think about those sleeps.

kirklund (approved):
[deleted] I'll add my approval after others review the WAN changes more closely. The sleeps are kind of scary. There's probably a different way to fix the reported problem.

geode-core/src/main/java/org/apache/geode/cache/wan/GatewaySender.java
```diff
@@ -174,7 +174,32 @@
...
+   * When group-transaction-events is set to true and the gateway sender is stopped,
+   * there is a possibility that the stopping occurs such that for a transaction,
+   * not all events belonging to it reach the queue. The reason would be that
+   * some reach the queue right before the sender is stopped and the rest do not make
+   * it to the queue because the sender is just stopped.
+   * In order to prevent that the queue contains incomplete transactions
+   * due to the above circumstance, this parameter allows for a grace period
+   * of the number of milliseconds set in it before the gateway sender is
+   * actually stopped, in which only events to complete transactions are put in the queue.
+   * Other events received in this period would be dropped.
```
davebarnes97, line 197: Suggested rewrite for economy of language:
When group-transaction-events is true and the gateway sender is stopped,
addition to the queue of a group of transaction events might be interrupted.
To ensure that the queue does not contain incomplete transactions, this parameter
allows for a grace period, specified in milliseconds, before the gateway sender is actually
stopped, allowing complete transaction event groups to be queued. Any event received
during the grace period that are not part of a transaction event group are dropped.

albertogpz: Thanks Dave. Very good suggestion. I made a slight change in the last sentence of your proposal but other than that, I took it as is.

davebarnes97 (approved):

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/parallel/ParallelWANStatsDUnitTest.java
```diff
@@ -472,6 +473,322 @@ public void testPRParallelPropagationWithGroupTransactionEventsSendsBatchesWithC
...
+        (ArrayList<Integer>) vm4.invoke(() -> WANTestBase.getSenderStats("ln", 0));
+    v5List =
+        (ArrayList<Integer>) vm5.invoke(() -> WANTestBase.getSenderStats("ln", 0));
+
+    assertEquals(0, v4List.get(0) + v5List.get(0));
+
+    // Wait for events to replicate: when batches received does not change
+    // we can assume that replication has finished.
+    batchesReceived = (vm2.invoke(() -> getReceiverStats())).get(0);
+    while (true) {
```
DonalEvans, line 605: See earlier comments about not using `Thread.sleep` in tests.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/parallel/ParallelWANStatsDUnitTest.java
```diff
@@ -472,6 +473,322 @@ public void testPRParallelPropagationWithGroupTransactionEventsSendsBatchesWithC
...
+        break;
+      }
+    }
+
+    System.out.println("batchesReceived after restart: " + batchesReceived);
+
+    // Wait for events to replicate: when region size does not change
+    // we can assume that replication has finished.
+    orderRegionSize = vm2.invoke(() -> getRegionSize(orderRegionName));
+    while (true) {
```
DonalEvans, line 619: See earlier comments about not using `Thread.sleep` in tests.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/parallel/ParallelWANStatsDUnitTest.java
```diff
@@ -472,6 +473,322 @@ public void testPRParallelPropagationWithGroupTransactionEventsSendsBatchesWithC
...
+      int newRegionSize = vm2.invoke(() -> getRegionSize(orderRegionName));
+      if (orderRegionSize == newRegionSize) {
+        break;
+      }
+    }
+
+    // Wait for events to replicate: when region size does not change
+    // we can assume that replication has finished.
+    shipmentRegionSize = vm2.invoke(() -> getRegionSize(shipmentRegionName));
+    while (true) {
```
DonalEvans, line 630: See earlier comments about not using `Thread.sleep` in tests.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/parallel/ParallelWANStatsDUnitTest.java
```diff
@@ -472,6 +473,322 @@ public void testPRParallelPropagationWithGroupTransactionEventsSendsBatchesWithC
...
+      int oldBatchesReceived = batchesReceived;
+      Thread.sleep(1000);
+      batchesReceived = (vm2.invoke(() -> getReceiverStats())).get(0);
+      if (batchesReceived == oldBatchesReceived) {
+        break;
+      }
+    }
+
+    System.out.println("batchesReceived after restart: " + batchesReceived);
```
DonalEvans, line 614: Another print statement.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/parallel/ParallelWANStatsDUnitTest.java
```diff
@@ -472,6 +473,322 @@ public void testPRParallelPropagationWithGroupTransactionEventsSendsBatchesWithC
...
+    shipmentRegionSize = vm2.invoke(() -> getRegionSize(shipmentRegionName));
+    while (true) {
+      Thread.sleep(1000);
+      int newRegionSize = vm2.invoke(() -> getRegionSize(shipmentRegionName));
+      if (shipmentRegionSize == newRegionSize) {
+        break;
+      }
+    }
+
+    System.out.println("orderRegionSize after restart: " + shipmentRegionSize);
```
DonalEvans, line 638: More print statements.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/parallel/ParallelWANStatsDUnitTest.java
```diff
@@ -472,6 +473,322 @@ public void testPRParallelPropagationWithGroupTransactionEventsSendsBatchesWithC
...
+
+    System.out.println("v4List.get(0): " + v4List.get(0));
+    System.out.println("v5List.get(0): " + v5List.get(0));
+
+    // batches with incomplete transactions must be 0
+    assertEquals(0, (int) v4List.get(13));
+    assertEquals(0, (int) v5List.get(13));
+
+    // Only complete transactions (1 order + 10 shipments) must be replicated
+    assertEquals(10, shipmentRegionSize / orderRegionSize);
```
DonalEvans, line 649: I think these could be replaced with one assert, as described above.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/parallel/ParallelWANStatsDUnitTest.java
```diff
@@ -472,6 +473,322 @@ public void testPRParallelPropagationWithGroupTransactionEventsSendsBatchesWithC
...
+    assertEquals(0, shipmentRegionSize % orderRegionSize);
+  }
+
+  @Test
+  public void testPRParallelPropagationWithGroupTransactionEventsDoesNotSendBatchesWithIncompleteTransactionsIfGatewaySenderIsStoppedWhileReceivingTrafficAndLaterStartedReceiverStopped()
```
DonalEvans, line 654: As there is a lot of duplicated code between this test and the one above it, all comments applying to that test case likely also apply here.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialWANStatsDUnitTest.java
```diff
@@ -348,7 +349,276 @@ public void testReplicatedSerialPropagationWithBatchRedistWithGroupTransactionEv
   }
 
   @Test
-  public void testReplicatedSerialPropagationWithMultipleDispatchers() throws Exception {
+  public void testReplicatedSerialPropagationWithGroupTransactionEventsDoesNotSendBatchesWithIncompleteTransactionsIfGatewaySenderIsStoppedWhileReceivingTrafficAndLaterStarted()
```
DonalEvans, line 352: There is a lot of duplicated code between this test and the other added to this class. It should be pulled out into methods to avoid duplication.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialWANStatsDUnitTest.java
```diff
@@ -348,7 +349,276 @@ public void testReplicatedSerialPropagationWithBatchRedistWithGroupTransactionEv
...
+
+    vm2.invoke(() -> WANTestBase.createReplicatedRegion(regionName, null, isOffHeap()));
+
+    vm4.invoke(() -> WANTestBase.createReplicatedRegion(regionName, "ln", isOffHeap()));
+    vm5.invoke(() -> WANTestBase.createReplicatedRegion(regionName, "ln", isOffHeap()));
+
+    startSenderInVMs("ln", vm4, vm5);
+
+    final Map<Object, Object> keyValues = new LinkedHashMap<>();
+    int entries = 2200;
```
DonalEvans, line 380: Is there a reason for this specific value being used for `entries`?

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialWANStatsDUnitTest.java
```diff
@@ -348,7 +349,276 @@ public void testReplicatedSerialPropagationWithBatchRedistWithGroupTransactionEv
...
+    for (int i = 0; i < entries; i++) {
+      keyValues.put(i, i + "_Value");
+    }
+
+    int eventsPerTransaction = 11;
```
DonalEvans, line 385: I assume that this value is related to `batchSize`, so it would be good to make that explicit rather than just having it a seeminly arbitrary number.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialWANStatsDUnitTest.java
```diff
@@ -348,7 +349,276 @@ public void testReplicatedSerialPropagationWithBatchRedistWithGroupTransactionEv
...
+    System.out.println("Starting puts");
```
DonalEvans, line 386: Another print statement that should be removed. In this case, a description can be added as the first argument to the `invokeAsync()` call below to help with debugging.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialWANStatsDUnitTest.java
```diff
@@ -348,7 +349,276 @@ public void testReplicatedSerialPropagationWithBatchRedistWithGroupTransactionEv
...
+    AsyncInvocation<Void> inv1 =
+        vm4.invokeAsync(
+            () -> WANTestBase.doPutsInsideTransactions(regionName, keyValues,
+                eventsPerTransaction));
+
+    // wait for batches to be distributed and then stop the sender
+    System.out.println("Waiting for some batches to be distributed");
```
DonalEvans, line 393: Another print statement that can be moved to being the first argument of the `invoke()` call.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialWANStatsDUnitTest.java
```diff
@@ -348,7 +349,276 @@ public void testReplicatedSerialPropagationWithBatchRedistWithGroupTransactionEv
...
+    vm4.invoke(() -> await()
+        .until(() -> WANTestBase.getSenderStats("ln", -1).get(4) > 0));
+    System.out
+        .println("Some batches distributed: " + vm4.invoke(() -> getSenderStats("ln", -1).get(4)));
+
+    addIgnoredException("Exception occurred in CacheListener");
```
DonalEvans, line 399: It would be helpful to add a comment explaining why these exceptions can be ignored here.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialWANStatsDUnitTest.java
```diff
@@ -348,7 +349,276 @@ public void testReplicatedSerialPropagationWithBatchRedistWithGroupTransactionEv
...
+    addIgnoredException(RejectedExecutionException.class);
+
+    System.out.println("Stopping sender");
```
DonalEvans, line 402: More print statements.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialWANStatsDUnitTest.java
```diff
@@ -348,7 +349,276 @@ public void testReplicatedSerialPropagationWithBatchRedistWithGroupTransactionEv
...
+    inv1.await();
+    System.out.println("Puts completed");
+
+    vm4.invoke(() -> WANTestBase.validateRegionSize(regionName, entries));
+    vm5.invoke(() -> WANTestBase.validateRegionSize(regionName, entries));
+
+    // Wait for events to replicate: when batches received does not change
+    // we can assume that replication has finished.
+    int batchesReceived = (vm2.invoke(() -> getReceiverStats())).get(0);
+    while (true) {
```
DonalEvans, line 415: See comments about not using `Thread.sleep` in tests.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialWANStatsDUnitTest.java
```diff
@@ -348,7 +349,276 @@ public void testReplicatedSerialPropagationWithBatchRedistWithGroupTransactionEv
...
+      int oldBatchesReceived = batchesReceived;
+      Thread.sleep(1000);
+      batchesReceived = (vm2.invoke(() -> getReceiverStats())).get(0);
+      if (batchesReceived == oldBatchesReceived) {
+        break;
+      }
+    }
+
+    ArrayList<Integer> v4List =
```
DonalEvans, line 424: These should be declared as `List<Integer>`, which removes the need for casting them here and elsewhere.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialWANStatsDUnitTest.java
```diff
@@ -348,7 +349,276 @@ public void testReplicatedSerialPropagationWithBatchRedistWithGroupTransactionEv
...
+        (ArrayList<Integer>) vm4.invoke(() -> WANTestBase.getSenderStats("ln", -1));
+    ArrayList<Integer> v5List =
+        (ArrayList<Integer>) vm5.invoke(() -> WANTestBase.getSenderStats("ln", -1));
+
+    System.out.println("v4List.get(0):" + v4List.get(0));
```
DonalEvans, line 429: More print statements.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialWANStatsDUnitTest.java
```diff
@@ -348,7 +349,276 @@ public void testReplicatedSerialPropagationWithBatchRedistWithGroupTransactionEv
...
+    System.out.println("v5List.get(0):" + v5List.get(0));
+
+    // Wait for events to replicate: when region size does not change
+    // we can assume that replication has finished.
+    int regionSize = vm2.invoke(() -> getRegionSize(regionName));
+    while (true) {
```
DonalEvans, line 435: See comments about not using `Thread.sleep` in tests.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialWANStatsDUnitTest.java
```diff
@@ -348,7 +349,276 @@ public void testReplicatedSerialPropagationWithBatchRedistWithGroupTransactionEv
...
+      int oldRegionSize = regionSize;
+      Thread.sleep(1000);
+      regionSize = vm2.invoke(() -> getRegionSize(regionName));
+      if (regionSize == oldRegionSize) {
+        break;
+      }
+    }
+
+    // Only complete transactions (11 entries each) must be replicated
+    assertEquals(0, regionSize % eventsPerTransaction);
```
DonalEvans, line 445: This assertion is not strictly checking what the comment above it is describing. It's possible that the region size could be a multiple of `eventsPerTransaction` but still contain incomplete transactions. Given that we check that the incomplete transactions stat is 0 immediately after, this assertion can probably be removed.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialWANStatsDUnitTest.java
```diff
@@ -348,7 +349,276 @@ public void testReplicatedSerialPropagationWithBatchRedistWithGroupTransactionEv
...
+
+    // batches with incomplete transactions must be 0
+    assertEquals(0, (int) v4List.get(13));
+    assertEquals(0, (int) v5List.get(13));
+
+    System.out.println("regionSize: " + regionSize);
```
DonalEvans, line 451: More print statements.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialWANStatsDUnitTest.java
```diff
@@ -348,7 +349,276 @@ public void testReplicatedSerialPropagationWithBatchRedistWithGroupTransactionEv
...
+        (ArrayList<Integer>) vm4.invoke(() -> WANTestBase.getSenderStats("ln", 0));
+    v5List =
+        (ArrayList<Integer>) vm5.invoke(() -> WANTestBase.getSenderStats("ln", 0));
+
+    assertEquals(0, v4List.get(0) + v5List.get(0));
+
+    // Wait for events to replicate: when batches received does not change
+    // we can assume that replication has finished.
+    batchesReceived = (vm2.invoke(() -> getReceiverStats())).get(0);
+    while (true) {
```
DonalEvans, line 466: See comments about not using `Thread.sleep` in tests.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialWANStatsDUnitTest.java
```diff
@@ -348,7 +349,276 @@ public void testReplicatedSerialPropagationWithBatchRedistWithGroupTransactionEv
...
+        break;
+      }
+    }
+
+    System.out.println("batchesReceived: " + batchesReceived);
+
+    // Wait for events to replicate: when region size does not change
+    // we can assume that replication has finished.
+    regionSize = vm2.invoke(() -> getRegionSize(regionName));
+    while (true) {
```
DonalEvans, line 480: See comments about not using `Thread.sleep` in tests.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialWANStatsDUnitTest.java
```diff
@@ -348,7 +349,276 @@ public void testReplicatedSerialPropagationWithBatchRedistWithGroupTransactionEv
...
+      int oldBatchesReceived = batchesReceived;
+      Thread.sleep(1000);
+      batchesReceived = (vm2.invoke(() -> getReceiverStats())).get(0);
+      if (batchesReceived == oldBatchesReceived) {
+        break;
+      }
+    }
+
+    System.out.println("batchesReceived: " + batchesReceived);
```
DonalEvans, line 475: Another print statement.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialWANStatsDUnitTest.java
```diff
@@ -348,7 +349,276 @@ public void testReplicatedSerialPropagationWithBatchRedistWithGroupTransactionEv
...
+    while (true) {
+      int oldRegionSize = regionSize;
+      Thread.sleep(1000);
+      regionSize = vm2.invoke(() -> getRegionSize(regionName));
+      if (regionSize == oldRegionSize) {
+        break;
+      }
+    }
+
+    System.out.println("regionSize: " + regionSize);
```
DonalEvans, line 489: More print statements.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialWANStatsDUnitTest.java
```diff
@@ -348,7 +349,276 @@ public void testReplicatedSerialPropagationWithBatchRedistWithGroupTransactionEv
...
+
+    System.out.println("v4List.get(0):" + v4List.get(0));
+    System.out.println("v5List.get(0):" + v5List.get(0));
+
+    // batches with incomplete transactions must be 0
+    assertEquals(0, (int) v4List.get(13));
+    assertEquals(0, (int) v5List.get(13));
+
+    // Only complete transactions (11 events) must be replicated
+    assertEquals(0, regionSize % eventsPerTransaction);
```
DonalEvans, line 499: See above comment about this assertion not necessarily checking what is being described in the comment.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialWANStatsDUnitTest.java
```diff
@@ -348,7 +349,276 @@ public void testReplicatedSerialPropagationWithBatchRedistWithGroupTransactionEv
...
+  }
+
+  @Test
+  public void testReplicatedSerialPropagationWithGroupTransactionEventsDoesNotSendBatchesWithIncompleteTransactionsIfGatewaySenderIsStoppedWhileReceivingTrafficAndLaterStartedReceiverStopped()
```
DonalEvans, line 503: Given that there is a large amount of duplicat code between this test case and the one above it, all comments that apply to the first test case also likely apply to this one.

DonalEvans (changes requested):
Quite a lot of changes here to clean up the code. I'm also concerned that there's not enough test coverage for the changes in terms of possible failure conditions, such as if two threads call `stop()` on the same gateway sender within quick succession. More unit tests should also be added for the new/modified behaviour.

geode-core/src/main/java/org/apache/geode/cache/wan/GatewaySender.java
```diff
@@ -174,7 +174,32 @@
    */
   int GET_TRANSACTION_EVENTS_FROM_QUEUE_RETRIES =
       Integer.getInteger(GeodeGlossary.GEMFIRE_PREFIX + "get-transaction-events-from-queue-retries",
-          10);
+          2);
```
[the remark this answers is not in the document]
albertogpz: Prior to this change, retries were done immediately if it was detected that not all events for a transaction were in the batch. In this change, I have added a configurable sleep before retrying with a default value of 1ms because I saw that retries were sometimes done so fast that it was not enough to get the expected events in the queue and it was more reasonable to have less retries and have some waiting before retrying.

geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
```diff
@@ -702,44 +704,15 @@ public boolean put(Object object) throws InterruptedException, CacheException {
     boolean isDREvent = isDREvent(sender.getCache(), value);
 
     String regionPath = value.getRegionPath();
```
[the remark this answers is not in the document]
albertogpz: Good catch.

geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ConcurrentParallelGatewaySenderEventProcessor.java
```diff
@@ -132,22 +134,24 @@ public int eventQueueSize() {
...
   }
 
   @Override
-  public void enqueueEvent(EnumListenerEvent operation, EntryEvent event, Object substituteValue,
-      boolean isLastEventInTransaction) throws IOException, CacheException {
+  public boolean enqueueEvent(EnumListenerEvent operation, EntryEvent event, Object substituteValue,
+      boolean isLastEventInTransaction, Predicate<InternalGatewayQueueEvent> condition)
+      throws IOException, CacheException {
     Region region = event.getRegion();
     // int bucketId = PartitionedRegionHelper.getHashKey((EntryOperation)event);
```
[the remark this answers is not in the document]
albertogpz: ok

geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
```diff
@@ -702,44 +704,15 @@ public boolean put(Object object) throws InterruptedException, CacheException {
...
-    if (!this.userRegionNameToShadowPRMap.containsKey(regionPath)) {
-      if (isDebugEnabled) {
-        logger.debug("The userRegionNameToshadowPRMap is {}", userRegionNameToShadowPRMap);
-      }
-      logger.warn(
-          "GatewaySender: Not queuing the event {}, as the region for which this event originated is not yet configured in the GatewaySender",
-          value);
-      // does not put into queue
+    regionPath = getRegionPathForEventAndType(value, isDREvent, regionPath);
+    if (regionPath == null)
```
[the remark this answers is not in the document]
albertogpz: ok

geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
```diff
@@ -702,44 +704,15 @@ public boolean put(Object object) throws InterruptedException, CacheException {
...
-              value);
-        }
-        // does not put into queue
-        return false;
-      }
-    } else {
-      key = value.getEventId();
-    }
+    Object key = getKeyForEventAndType(value, isDREvent);
+    if (key == null)
```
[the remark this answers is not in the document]
albertogpz: ok

geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
```diff
@@ -870,6 +843,49 @@ public boolean put(Object object) throws InterruptedException, CacheException {
     return putDone;
   }
 
+  private String getRegionPathForEventAndType(GatewaySenderEventImpl event, boolean isDREvent,
+      String regionPath) {
+    boolean isDebugEnabled = logger.isDebugEnabled();
+    if (!isDREvent) {
+      Region region = sender.getCache().getRegion(regionPath, true);
```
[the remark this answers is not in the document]
albertogpz: ok

geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
```diff
@@ -1797,6 +1817,22 @@ public static String getSenderId(String regionName) {
     return regionName.substring(1, queueStringStart);
   }
 
+  public boolean isThereEventsMatching(GatewaySenderEventImpl event,
```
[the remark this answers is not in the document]
albertogpz: Agree

geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
```diff
@@ -1797,6 +1817,22 @@ public static String getSenderId(String regionName) {
...
+      Predicate<InternalGatewayQueueEvent> condition) {
+    boolean isDREvent = isDREvent(sender.getCache(), event);
+
+    String regionPath = event.getRegionPath();
```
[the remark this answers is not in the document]
albertogpz: ok

geode-core/src/main/java/org/apache/geode/internal/cache/wan/serial/ConcurrentSerialGatewaySenderEventProcessor.java
```diff
@@ -418,10 +424,13 @@ protected void registerEventDroppedInPrimaryQueue(EntryEventImpl droppedEvent) {
   }
 
   @Override
-  protected void enqueueEvent(GatewayQueueEvent event) {
+  protected boolean enqueueEvent(GatewayQueueEvent event,
+      Predicate<InternalGatewayQueueEvent> condition) {
     for (SerialGatewaySenderEventProcessor serialProcessor : this.processors) {
-      serialProcessor.enqueueEvent(event);
+      // TODO revisit handling of "condition" when the following enqueueEvent() method is supported:
```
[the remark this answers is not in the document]
albertogpz: ok

geode-core/src/main/java/org/apache/geode/internal/cache/wan/serial/SerialGatewaySenderQueue.java
```diff
@@ -847,7 +855,14 @@ private EventsAndLastKey peekEventsWithTransactionId(TransactionId transactionId
...
+  /**
+   * This method returns a list of objects that fulfill the matchingPredicate
+   * If a matching object also fulfills the endPredicate then the method
+   * stops looking for more matching objects.
+   */
+  List<KeyAndEventPair> getElementsMatching(Predicate<InternalGatewayQueueEvent> condition,
+      Predicate<InternalGatewayQueueEvent> stopCondition,
+      long lastKey) {
     Object object;
     List elementsMatching = new ArrayList<>();
```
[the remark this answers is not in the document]
albertogpz: ok

geode-core/src/main/java/org/apache/geode/internal/cache/wan/serial/SerialGatewaySenderQueue.java
```diff
@@ -862,19 +877,35 @@ EventsAndLastKey getElementsMatching(Predicate condition, Predicate stopConditio
         continue;
       }
 
-      if (condition.test(object)) {
-        elementsMatching.add(object);
-        peekedIds.add(currentKey);
-        extraPeekedIds.add(currentKey);
-        lastKey = currentKey;
+      if (condition.test((InternalGatewayQueueEvent) object)) {
```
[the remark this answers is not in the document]
albertogpz: ok

geode-core/src/main/java/org/apache/geode/internal/cache/wan/serial/SerialGatewaySenderQueue.java
```diff
@@ -847,7 +855,14 @@ private EventsAndLastKey peekEventsWithTransactionId(TransactionId transactionId
     }
   }
 
-  EventsAndLastKey getElementsMatching(Predicate condition, Predicate stopCondition, long lastKey) {
+  /**
+   * This method returns a list of objects that fulfill the matchingPredicate
+   * If a matching object also fulfills the endPredicate then the method
+   * stops looking for more matching objects.
+   */
+  List<KeyAndEventPair> getElementsMatching(Predicate<InternalGatewayQueueEvent> condition,
```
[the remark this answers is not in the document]
albertogpz: Good catch.

geode-core/src/main/java/org/apache/geode/internal/cache/wan/serial/SerialGatewaySenderQueue.java
```diff
@@ -862,19 +877,35 @@ EventsAndLastKey getElementsMatching(Predicate condition, Predicate stopConditio
...
           break;
         }
       }
     }
 
-    return new EventsAndLastKey(elementsMatching, lastKey);
+    return elementsMatching;
+  }
+
+  public boolean isThereEventsMatching(Predicate<InternalGatewayQueueEvent> condition) {
```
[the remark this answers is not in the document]
albertogpz: ok

geode-core/src/main/java/org/apache/geode/internal/cache/wan/serial/SerialGatewaySenderQueue.java
```diff
@@ -862,19 +877,35 @@ EventsAndLastKey getElementsMatching(Predicate condition, Predicate stopConditio
...
+    InternalGatewayQueueEvent object;
```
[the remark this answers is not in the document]
albertogpz: ok

geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySender.java
```diff
@@ -1100,7 +1108,21 @@ public void distribute(EnumListenerEvent operation, EntryEventImpl event,
...
+          Predicate<InternalGatewayQueueEvent> hasSameTransactionIdPredicate = null;
+          // In case the sender is about to be stopped, the event will only
+          // be queued if there is any event in the queue with the same
+          // transactionId as the one of this event
+          if (isStopping && mustGroupTransactionEvents()
+              && clonedEvent.getTransactionId() != null) {
+            hasSameTransactionIdPredicate =
+                x -> x instanceof GatewaySenderEventImpl && clonedEvent.getTransactionId() != null
+                    && clonedEvent.getTransactionId()
+                        .equals(((GatewaySenderEventImpl) x).getTransactionId());
```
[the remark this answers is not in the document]
albertogpz: ok

geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySender.java
```diff
@@ -1100,7 +1108,21 @@ public void distribute(EnumListenerEvent operation, EntryEventImpl event,
...
 
-          ev.enqueueEvent(operation, clonedEvent, substituteValue, isLastEventInTransaction);
+          Predicate<InternalGatewayQueueEvent> hasSameTransactionIdPredicate = null;
+          // In case the sender is about to be stopped, the event will only
+          // be queued if there is any event in the queue with the same
+          // transactionId as the one of this event
+          if (isStopping && mustGroupTransactionEvents()
+              && clonedEvent.getTransactionId() != null) {
+            hasSameTransactionIdPredicate =
+                x -> x instanceof GatewaySenderEventImpl && clonedEvent.getTransactionId() != null
```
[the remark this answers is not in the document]
albertogpz: ok

geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySenderEventProcessor.java
```diff
@@ -195,6 +207,9 @@ protected void setIsStopped(boolean isStopped) {
     } else {
       this.isStopped = isStopped;
     }
+    if (isStopped) {
```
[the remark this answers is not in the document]
albertogpz: Yep

geode-core/src/main/java/org/apache/geode/internal/cache/BucketRegionQueue.java
```diff
@@ -453,21 +454,22 @@ public Object peek() {
...
-  public List<Object> getElementsMatching(Predicate matchingPredicate, Predicate endPredicate) {
+  public List<Object> getElementsMatching(Predicate<InternalGatewayQueueEvent> matchingPredicate,
+      Predicate<InternalGatewayQueueEvent> endPredicate) {
     getInitializationLock().readLock().lock();
     try {
       if (this.getPartitionedRegion().isDestroyed()) {
         throw new BucketRegionQueueUnavailableException();
       }
-      List<Object> elementsMatching = new ArrayList<Object>();
+      List<Object> elementsMatching = new ArrayList();
```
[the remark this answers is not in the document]
albertogpz: ok

geode-core/src/main/java/org/apache/geode/internal/cache/BucketRegionQueue.java
```diff
@@ -453,21 +454,22 @@ public Object peek() {
...
       Iterator<Object> it = this.eventSeqNumDeque.iterator();
       while (it.hasNext()) {
         Object key = it.next();
         Object object = optimalGet(key);
-        if (matchingPredicate.test(object)) {
+        if (matchingPredicate.test((InternalGatewayQueueEvent) object)) {
```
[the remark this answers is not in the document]
albertogpz: ok

geode-core/src/main/java/org/apache/geode/internal/cache/BucketRegionQueue.java
```diff
@@ -478,6 +480,24 @@ public Object peek() {
     }
   }
 
+  public boolean isThereEventsMatching(Predicate<InternalGatewayQueueEvent> matchingPredicate) {
```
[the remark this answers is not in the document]
albertogpz: ok

geode-core/src/main/java/org/apache/geode/internal/cache/BucketRegionQueue.java
```diff
@@ -478,6 +480,24 @@ public Object peek() {
...
+    getInitializationLock().readLock().lock();
+    try {
+      if (this.getPartitionedRegion().isDestroyed()) {
+        throw new BucketRegionQueueUnavailableException();
+      }
+      for (Object o : eventSeqNumDeque) {
+        Object object = optimalGet(o);
+        if (matchingPredicate.test((InternalGatewayQueueEvent) object)) {
```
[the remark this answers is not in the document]
albertogpz: ok

geode-core/src/main/java/org/apache/geode/internal/cache/EntryEventImpl.java
```diff
@@ -2231,6 +2231,9 @@ public String toString() {
     if (this.tailKey != -1) {
       buf.append(";tailKey=" + tailKey);
     }
+    if (this.isTransactional()) {
+      buf.append(";transactionId=" + getTransactionId());
```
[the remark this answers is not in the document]
albertogpz: ok

geode-core/src/test/java/org/apache/geode/internal/cache/wan/serial/SerialGatewaySenderQueueJUnitTest.java
```diff
@@ -137,8 +137,8 @@ public void peekGetsExtraEventsWhenMustGroupTransactionEventsAndNotAllEventsForT
         .doAnswer(invocation -> null)
         .when(queue).peekAhead();
 
-    doAnswer(invocation -> new SerialGatewaySenderQueue.EventsAndLastKey(
-        Arrays.asList(new Object[] {event4}), 2L))
+    doAnswer(invocation -> Arrays
```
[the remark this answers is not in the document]
albertogpz: ok

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/WANTestBase.java
```diff
@@ -1320,6 +1320,23 @@ public static void checkGatewayReceiverStats(int processBatches, int eventsRecei
     assertEquals(creates, gatewayReceiverStats.getCreateRequest());
   }
 
+  public static List<Integer> getReceiverStats() {
```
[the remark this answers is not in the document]
albertogpz: I agree but the problem is that `GatewayReceiverStats` must be Serializable and it is not. That's why I implemented something similar to `gatewaySenderStats()`.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/parallel/ParallelWANStatsDUnitTest.java
```diff
@@ -472,6 +473,322 @@ public void testPRParallelPropagationWithGroupTransactionEventsSendsBatchesWithC
...
+
+    createSenderCustomerOrderShipmentPRs(vm4);
+    createSenderCustomerOrderShipmentPRs(vm5);
+
+    startSenderInVMs("ln", vm4, vm5);
+
+    int customers = 4;
+    int transactionsPerCustomer = 100;
+    int shipmentsPerTransaction = 10;
+    final LinkedHashMap<Object, Object> keyValuesInTransactions = new LinkedHashMap<>();
```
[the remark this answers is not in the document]
albertogpz: I agree

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/parallel/ParallelWANStatsDUnitTest.java
```diff
@@ -472,6 +473,322 @@ public void testPRParallelPropagationWithGroupTransactionEventsSendsBatchesWithC
...
+    createReceiverCustomerOrderShipmentPR(vm2);
+
+    createSenderCustomerOrderShipmentPRs(vm4);
+    createSenderCustomerOrderShipmentPRs(vm5);
+
+    startSenderInVMs("ln", vm4, vm5);
+
+    int customers = 4;
+    int transactionsPerCustomer = 100;
+    int shipmentsPerTransaction = 10;
```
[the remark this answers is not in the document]
albertogpz: ok

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/parallel/ParallelWANStatsDUnitTest.java
```diff
@@ -472,6 +473,322 @@ public void testPRParallelPropagationWithGroupTransactionEventsSendsBatchesWithC
...
+    AsyncInvocation<Void> inv1 =
+        vm4.invokeAsync(
+            () -> WANTestBase.doOrderAndShipmentPutsInsideTransactions(keyValuesInTransactions,
+                eventsPerTransaction));
+
+    // wait for some batches to be distributed and then stop the sender
+    vm4.invoke(() -> await()
+        .until(() -> WANTestBase.getSenderStats("ln", -1).get(4) > 0));
+
+    System.out.println("Stopping sender");
```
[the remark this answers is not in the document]
albertogpz: ok

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/parallel/ParallelWANStatsDUnitTest.java
```diff
@@ -472,6 +473,322 @@ public void testPRParallelPropagationWithGroupTransactionEventsSendsBatchesWithC
...
+    inv1.await();
+    int entries =
+        transactionsPerCustomer * customers;
+
+    vm4.invoke(() -> WANTestBase.validateRegionSize(orderRegionName, entries));
+    vm5.invoke(() -> WANTestBase.validateRegionSize(orderRegionName, entries));
+
+    // Wait for events to replicate: when batches received does not change
+    // we can assume that replication has finished.
+    int batchesReceived = (vm2.invoke(() -> getReceiverStats())).get(0);
```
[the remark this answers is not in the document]
albertogpz: ok

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/parallel/ParallelWANStatsDUnitTest.java
```diff
@@ -472,6 +473,322 @@ public void testPRParallelPropagationWithGroupTransactionEventsSendsBatchesWithC
...
+
+    System.out.println("v4List.get(0): " + v4List.get(0));
+    System.out.println("v5List.get(0): " + v5List.get(0));
+
+    // batches with incomplete transactions must be 0
+    assertEquals(0, (int) v4List.get(13));
+    assertEquals(0, (int) v5List.get(13));
+
+    // Only complete transactions (1 order + 10 shipments) must be replicated
+    assertEquals(10, shipmentRegionSize / orderRegionSize);
```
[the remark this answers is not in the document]
albertogpz: Indeed

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/parallel/ParallelWANStatsDUnitTest.java
```diff
@@ -472,6 +473,322 @@ public void testPRParallelPropagationWithGroupTransactionEventsSendsBatchesWithC
...
+    int entries =
+        transactionsPerCustomer * customers;
+
+    vm4.invoke(() -> WANTestBase.validateRegionSize(orderRegionName, entries));
+    vm5.invoke(() -> WANTestBase.validateRegionSize(orderRegionName, entries));
+
+    // Wait for events to replicate: when batches received does not change
+    // we can assume that replication has finished.
+    int batchesReceived = (vm2.invoke(() -> getReceiverStats())).get(0);
+    while (true) {
```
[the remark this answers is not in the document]
albertogpz: I have tried by checking that the number of batches sent by the sender is equal to the number of batches received by the receiver.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/WANTestBase.java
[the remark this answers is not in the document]
DonalEvans: I see, that's unfortunate but unavoidable then.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialWANStatsDUnitTest.java
albertogpz: ok

albertogpz:
> Quite a lot of changes here to clean up the code. I'm also concerned that there's not enough test coverage for the changes in terms of possible failure conditions, such as if two threads call `stop()` on the same gateway sender within quick succession. More unit tests should also be added for the new/modified behaviour.

Thanks a lot for your thorough review. Impressive, as usual.

I have added a couple of tests in `SerialGatewaySenderImplTest` and `ParallelGatewalSenderImplTest`to verify that two threads calling `stop()` on the same gateway sender within quick succession does not provoke any harm and the `preStop` and `postStop` methods do their job.

geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySender.java
```diff
@@ -686,6 +689,22 @@ public boolean beforeEnqueue(GatewayQueueEvent gatewayEvent) {
     return enqueue;
   }
 
+  protected void preStop() {
+    if (!mustGroupTransactionEvents() || isStopping) {
```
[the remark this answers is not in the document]
albertogpz: You are right. I have removed the `isStopping` variable from the condition.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/parallel/ParallelWANStatsDUnitTest.java
```diff
@@ -472,6 +473,200 @@ public void testPRParallelPropagationWithGroupTransactionEventsSendsBatchesWithC
...
+    // wait for some batches to be distributed and then stop the sender
+    vm4.invoke(() -> await()
+        .until(() -> WANTestBase.getSenderStats("ln", -1).get(4) > 0));
+
+    stopSenderInVMsAsync("ln", vm4, vm5);
+
+    // Wait for customer transactions to finish
+    inv1.await();
+    int orderEntries = transactionsPerCustomer * customers;
+    int shipmentEntries = orderEntries * 10;
```
DonalEvans, line 514: I think this could be `int shipmentEntries = orderEntries * shipmentsPerTransaction;` for additional clarity.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/parallel/ParallelWANStatsDUnitTest.java
```diff
@@ -472,6 +473,200 @@ public void testPRParallelPropagationWithGroupTransactionEventsSendsBatchesWithC
...
+    int batchesDistributed = v4List.get(4) + v5List.get(4);
+    checkOnlyCompleteTransactionsAreReplicated(shipmentsPerTransaction, batchesDistributed);
+  }
+
+  private void checkOnlyCompleteTransactionsAreReplicated(int shipmentsPerTransaction,
+      int batchesDistributed) {
+    // Only complete transactions (1 order + 10 shipments) must be replicated
+    int orderRegionSize = vm2.invoke(() -> getRegionSize(orderRegionName));
+    int shipmentRegionSize = vm2.invoke(() -> getRegionSize(shipmentRegionName));
+    assertEquals(shipmentRegionSize, 10 * orderRegionSize);
```
DonalEvans, line 647: I think that this could be `assertEquals(shipmentRegionSize, shipmentsPerTransaction * orderRegionSize);` for added clarity.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/parallel/ParallelWANStatsDUnitTest.java
```diff
@@ -472,6 +473,200 @@ public void testPRParallelPropagationWithGroupTransactionEventsSendsBatchesWithC
...
+    vm2.invoke(() -> WANTestBase.validateRegionSize(orderRegionName, batchesDistributed));
+    vm2.invoke(() -> WANTestBase.validateRegionSize(shipmentRegionName,
+        batchesDistributed * shipmentsPerTransaction));
+  }
+
+  private void waitForBatchesToBeAppliedInTheReceiver(int shipmentsPerTransaction) {
+    int batchesSentTotal = vm4.invoke(() -> WANTestBase.getSenderStats("ln", -1)).get(4) +
+        vm5.invoke(() -> WANTestBase.getSenderStats("ln", -1)).get(4);
+
+    // Wait for all batches to be received by the sender
```
DonalEvans, line 658: I think this comment is supposed to be "received by the receiver"

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialWANStatsDUnitTest.java
DonalEvans: In that case, it might be good to make the connection to transaction size explicit and make it `int entries = eventsPerTransaction * 200;`

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/parallel/ParallelWANStatsDUnitTest.java
albertogpz: Totally agree

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/parallel/ParallelWANStatsDUnitTest.java
albertogpz: Agree

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialWANStatsDUnitTest.java
```diff
@@ -348,7 +349,201 @@ public void testReplicatedSerialPropagationWithBatchRedistWithGroupTransactionEv
...
+    return vm4.invokeAsync(
+        () -> WANTestBase.doPutsInsideTransactions(regionName, keyValues,
+            eventsPerTransaction));
+  }
+
+  private void waitForBatchesToBeAppliedInTheReceiver(String regionName, int eventsPerTransaction) {
+    int batchesSentTotal = vm4.invoke(() -> WANTestBase.getSenderStats("ln", -1)).get(4) +
+        vm5.invoke(() -> WANTestBase.getSenderStats("ln", -1)).get(4);
+
+    // Wait for all batches to be received by the sender
```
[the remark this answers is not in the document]
DonalEvans: It looks like this got missed in the last set of changed, but other than that, everything looks good.

DonalEvans (approved):

geode-core/src/main/java/org/apache/geode/cache/wan/GatewaySender.java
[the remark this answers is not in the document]
gesterzhou: But 2 is still too less. How about 3~5?

geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySender.java
```diff
@@ -126,6 +127,8 @@
 
   protected boolean groupTransactionEvents;
 
+  protected volatile boolean isStopping = false;
```
gesterzhou, line 130: can we have a better name, such as isStoppingForTX?

geode-core/src/main/java/org/apache/geode/cache/wan/GatewaySender.java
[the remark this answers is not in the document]
albertogpz: > But 2 is still too less. How about 3~5?

Given that prior to this change we had 10 retries and no timeout, and these retries were done really fast (in less than 1 ms) I did not want to add a much higher waiting time because it does not seem necessary if the events will eventually reach the queue so I thought it would be better to have some small waiting time (2ms) and have a smaller number of retries. These numbers seem reasonable and have not caused flakyneess in the test cases.

If we put a higher waiting time and more retries as you are suggesting and, for some reason, the configuration of the gateway senders or the traffic received contains transactions but they cannot be completed (even with infinite time), then you are going to be delaying replication for a long time: each batch for waiting time*number of retries which could be a lot.

That's why I prefer to leave these values low and, if in some case it is necessary, they can be changed with the system properties.

geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySender.java
albertogpz: I would prefer to leave it as it is. The variable is just indicating that we are in the process of stopping the gateway sender.
I have changed the `AbstractGatewaySender.preStop()` method to make this more clear.

gesterzhou (approved):

PR #6053, opened 2021-02-24
GEODE-8964: Entry created via TX putIfAbsent with null value can write incorrect data to cache

luissson:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/TXCommitMessage.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/TXCommitMessage.java
@@ -754,8 +754,13 @@ public class TXCommitMessage extends PooledDistributionMessage
           ee.getRegion().invokeTXCallbacks(EnumListenerEvent.AFTER_CREATE, ee, true,
               isLastTransactionEvent);
         } else {
-          ee.getRegion().invokeTXCallbacks(EnumListenerEvent.AFTER_UPDATE, ee, true,
-              isLastTransactionEvent);
+          if (ee.getNewValue() == null) { // GEODE-8964, fixes GII and TX create conflict that
+            ee.getRegion(). // produces an Update with null value
+                invokeTXCallbacks(EnumListenerEvent.AFTER_CREATE, ee, true, isLastTransactionEvent);
+          } else {
+            ee.getRegion().invokeTXCallbacks(EnumListenerEvent.AFTER_UPDATE, ee, true,
+                isLastTransactionEvent);
+          }
         }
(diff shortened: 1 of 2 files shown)


geode-core/src/main/java/org/apache/geode/internal/cache/TXCommitMessage.java
```diff
@@ -754,8 +754,14 @@ void firePendingCallbacks(List<EntryEventImpl> callbacks) {
           ee.getRegion().invokeTXCallbacks(EnumListenerEvent.AFTER_CREATE, ee, true,
               isLastTransactionEvent);
         } else {
-          ee.getRegion().invokeTXCallbacks(EnumListenerEvent.AFTER_UPDATE, ee, true,
-              isLastTransactionEvent);
+          if (ee.getNewValue() == null) { // GEODE-8964, fixes GII and TX create conflict that
```
DonalEvans, line 757: Small nitpick, but could the comment here be put on its own line?

geode-core/src/test/java/org/apache/geode/internal/cache/TXCommitMessageTest.java
```diff
@@ -80,4 +81,48 @@ public void firePendingCallbacksSetsChangeAppliedToCacheInEventLocalFilterInfo()
...
+
+    when(updateEvent.getLocalFilterInfo()).thenReturn(null);
+    when(updateEvent.getNewValue()).thenReturn(null);
+    when(updateEvent.getRegion()).thenReturn(region);
+
+    doReturn(lastTxEvent).when(message).getLastTransactionEvent(callbacks);
+
+    message.firePendingCallbacks(callbacks);
+
+    verify(region).invokeTXCallbacks(EnumListenerEvent.AFTER_CREATE, updateEvent, true, false);
```
DonalEvans, line 103: The two test cases here can be condensed to one if this line is changed to `verify(region, only()).invokeTXCallbacks(EnumListenerEvent.AFTER_CREATE, updateEvent, true, false);` which will validate both that `invokeTXCallbacks()` is called with the correct arguments and that it is *only* called with those arguments, removing the need for a separate test to confirm that it's not called with the `AFTER_UPDATE` argument.

DonalEvans (changes requested):
Just a little code simplification/cleanup.

DonalEvans (approved):

luissson:
only() was on the :nose:, thanks for the feedback!

kirklund (approved):
Looks great. Thank you!

PR #6054, opened 2021-02-25
GEODE-8907: CI Failure: ClientServerTransactionCCEDUnitTest.testTxRemoveAll

luissson:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClientServerTransactionDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClientServerTransactionDUnitTest.java
@@ -911,6 +911,7 @@ public class ClientServerTransactionDUnitTest extends RemoteTransactionDUnitTest
       @Override
       public Object call() throws Exception {
         TXManagerImpl mgr = getGemfireCache().getTxManager();
+        mgr.resume(txId);
         TXStateProxy tx = mgr.pauseTransaction();
         assertNotNull(tx);
         mgr.unpauseTransaction(tx);
@@ -1934,7 +1935,7 @@ public class ClientServerTransactionDUnitTest extends RemoteTransactionDUnitTest
           pr.put(custId, cust);
           r.put(i, "value" + i);
         }
-        return mgr.getTransactionId();
+        return mgr.suspend();
       }
     });
 
@@ -1965,6 +1966,7 @@ public class ClientServerTransactionDUnitTest extends RemoteTransactionDUnitTest
       @Override
       public Object call() throws Exception {
         TXManagerImpl mgr = getGemfireCache().getTxManager();
+        mgr.resume(txid);
         mgr.commit();
         return null;
       }
@@ -2030,7 +2032,7 @@ public class ClientServerTransactionDUnitTest extends RemoteTransactionDUnitTest
     final int port3 = serverOnDatastore ? createRegionsAndStartServer(datastore, false)
         : createRegionOnServer(datastore, false, false);
 
-    /* final TXId txid = (TXId) */client.invoke(new SerializableCallable() {
+    final TXId txid = (TXId) client.invoke(new SerializableCallable() {
       @Override
       public Object call() throws Exception {
         System.setProperty(GeodeGlossary.GEMFIRE_PREFIX + "bridge.disableShufflingOfEndpoints",
@@ -2061,7 +2063,7 @@ public class ClientServerTransactionDUnitTest extends RemoteTransactionDUnitTest
           pr.put(custId, cust);
           r.put(i, "value" + i);
         }
-        return mgr.getTransactionId();
+        return mgr.suspend();
       }
     });
 
@@ -2079,7 +2081,8 @@ public class ClientServerTransactionDUnitTest extends RemoteTransactionDUnitTest
     client.invoke(new SerializableCallable() {
       @Override
       public Object call() throws Exception {
-        /* TXManagerImpl mgr = */ getGemfireCache().getTxManager();
+        TXManagerImpl mgr = getGemfireCache().getTxManager();
+        mgr.resume(txid);
         Region<Integer, String> r = getGemfireCache().getRegion(D_REFERENCE);
         Region<CustId, Customer> pr = getGemfireCache().getRegion(CUSTOMER);
         for (int i = 5; i < 10; i++) {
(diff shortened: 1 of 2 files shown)


DonalEvans (approved):
The changes look good, but I feel like this would be an excellent opportunity to clean up some of the warnings in the test files while you're there. Even just doing a find-replace for `getGemfireCache()` to `getCache()` and replacing all the `SerializableCallable` with lambdas would deal with a pretty big chunk of them. Maybe that would be better kept to a separate PR though.

luissson:
Good point, I did consider that but am in the process of rewriting RemoteTransactionDUnitTest which will address those concerns and more. I wanted to treat this PR as a quick fix to stop these sorts errors from popping up in CI while the rework is being done.

kirklund (approved):

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClientServerTransactionDUnitTest.java
```diff
@@ -2108,6 +2110,8 @@ public Object call() throws Exception {
         TXManagerImpl mgr = getGemfireCache().getTxManager();
         Region<Integer, String> r = getGemfireCache().getRegion(D_REFERENCE);
         Region<CustId, Customer> pr = getGemfireCache().getRegion(CUSTOMER);
+        TXStateProxy tx = mgr.internalSuspend();
```
pivotal-eshu, line 2113: This will not address the issue we saw in the pipeline.
The issue is the thread can not internalSuspend a transaction, if it does not in a txState. Which we saw in the test failure on the second call.

The correct fix is change the first callable return mgr.getTransactionId() to mgr.suspend().
    final TXId txid = (TXId) client.invoke(new SerializableCallable() {
      @Override
      public Object call() throws Exception {
        ClientCacheFactory ccf = new ClientCacheFactory();
        ccf.addPoolServer("localhost"/* getServerHostName(Host.getHost(0)) */, port1);
        ccf.addPoolServer("localhost", port2);
        ccf.setPoolLoadConditioningInterval(1);
        ccf.setPoolSubscriptionEnabled(false);
        ccf.set(LOG_LEVEL, getDUnitLogLevel());
        ClientCache cCache = getClientCache(ccf);
        ClientRegionFactory<CustId, Customer> custrf =
            cCache.createClientRegionFactory(ClientRegionShortcut.PROXY);
        ClientRegionFactory<Integer, String> refrf =
            cCache.createClientRegionFactory(ClientRegionShortcut.PROXY);
        Region<Integer, String> r = refrf.create(D_REFERENCE);
        Region<CustId, Customer> pr = custrf.create(CUSTOMER);
        // Region<Integer, String> order = refrf.create(ORDER);

        TXManagerImpl mgr = getGemfireCache().getTxManager();
        mgr.begin();
        int i = 0;
        for (int j = 0; j < 10; j++) {
          CustId custId = new CustId(i);
          Customer cust = new Customer("name" + i, "address" + i);
          getGemfireCache().getLogger().info("SWAP:putting:" + custId);
          pr.put(custId, cust);
          r.put(i, "value" + i);
        }
        return mgr.suspend();
      }
    });

And here you can resume transaction by calling mgr.resume(txId);

pivotal-eshu (changes requested):
The issues are two callable that RMI call are not invoked by the same thread.
To avoid that, you need to have the first thread in first call to suspend the transaction and then the second thread can resume it.
If first callable returns txStateProxy, you can use internalSuspend/internalResume. For returning txId, you need to change to have suspend/resume instead.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClientServerTransactionDUnitTest.java
```diff
@@ -2076,10 +2078,11 @@ public Object call() throws Exception {
       }
     });
 
-    client.invoke(new SerializableCallable() {
+    TXId txid2 = (TXId) client.invoke(new SerializableCallable() {
```
pivotal-eshu, line 2081: There is no need to have a new txId here, unless you want to assert this is the same as txId. There is only one transaction. Probably just ignore the return value here. And later resume the original txId instead of txid2.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/RemoteTransactionDUnitTest.java
```diff
@@ -3010,50 +3024,82 @@ public Object call() throws Exception {
...
-        mgr.commit();
-        assertEquals(expectedCustomer, custRegion.get(expectedCustId));
-        custRegion.destroy(expectedCustId);
-        return null;
-      }
-    };
-
     if (owner.equals(ds1)) {
-      datastore1.invoke(testFnOnDs);
+      TXId dsTxId = (TXId) datastore1.invoke(
```
pivotal-eshu, line 3028: Should not duplicate the code here, should reuse the existing callables.

To pass in different txIds in the callable, you can see how DoOpsInTX is implemented.

pivotal-eshu (changes requested):
There are a few test methods not fixed yet, like the following in the ClientServerTransactionDUnitTest:
testFailoverAfterCommitDistribution
testFailoverAfterCommitDistributionInvokesListenerInClientOnlyOnce
And testEventTracker (which has a different way to address the issue, but missed similar issue in another client call) You modify the test to all use suspend/resume call.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClientServerTransactionDUnitTest.java
```diff
@@ -3749,14 +3756,15 @@ public Object call() throws Exception {
...
-        return null;
+        return mgr.suspend();
       }
     });
 
     server.invoke(new SerializableCallable() {
       @Override
       public Object call() throws Exception {
         TXManagerImpl mgr = getGemfireCache().getTxManager();
+        // mgr.resume(txid2);
```
pivotal-eshu, line 3767: You may want to clean up this commented out line.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClientServerTransactionDUnitTest.java
```diff
@@ -3782,13 +3790,15 @@ public Object call() throws Exception {
       @Override
       public Object call() throws Exception {
         TXManagerImpl mgr = getGemfireCache().getTxManager();
+        mgr.resume(txid);
```
pivotal-eshu, line 3793: After you resume the transaction, the following code trying to resume can be safely removed.

pivotal-eshu (approved):
+1 Looks good.

PR #6057, opened 2021-02-26
add 1.12.1 to old versions on develop

onichols-pivotal:


rhoughton-pivot (approved):
Thanks for keeping our upgrade tests honest.

PR #6059, opened 2021-02-27
GEODE-8886 fix

mhansonp:
Getting rid of a write and a read in the wrong place...

### DIFF
--- a/geode-core/src/integrationTest/resources/org/apache/geode/codeAnalysis/sanctionedDataSerializables.txt
+++ b/geode-core/src/integrationTest/resources/org/apache/geode/codeAnalysis/sanctionedDataSerializables.txt
@@ -1918,7 +1918,6 @@ toData,17
 toDataPre_GEODE_1_14_0_0,271
 toDataPre_GFE_8_0_0_0,236
 
-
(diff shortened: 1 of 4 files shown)


geode-core/src/integrationTest/resources/org/apache/geode/codeAnalysis/sanctionedDataSerializables.txt
```diff
@@ -1912,12 +1912,13 @@ toData,254
...
 fromData,17
-fromDataPre_GEODE_1_14_0_0,293
+fromDataPre_GEODE_1_14_0_0,283
 fromDataPre_GFE_8_0_0_0,188
 toData,17
-toDataPre_GEODE_1_14_0_0,281
+toDataPre_GEODE_1_14_0_0,271
 toDataPre_GFE_8_0_0_0,236
 
+
```
mkevo, line 1921: There is no need to introduce a new blank line. Everything else is good.

mkevo (approved):

boglesby (approved):

gesterzhou (approved):
I think it's good. The locator can use current version to simulate rollingupgrade. But as long as it works, using old version locator is also fine.

PR #6063, opened 2021-02-27
GEODE-6150_1: Solution for Client/Server NIO SSL

mivanac:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/AcceptorImpl.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/AcceptorImpl.java
@@ -57,6 +57,8 @@ import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Supplier;
 
+import javax.net.ssl.SSLEngine;
+
 import io.micrometer.core.instrument.MeterRegistry;
 import org.apache.logging.log4j.Logger;
 
@@ -72,10 +74,12 @@ import org.apache.geode.cache.server.CacheServer;
 import org.apache.geode.cache.wan.GatewayTransportFilter;
 import org.apache.geode.distributed.DistributedMember;
 import org.apache.geode.distributed.internal.DistributionConfig;
+import org.apache.geode.distributed.internal.DistributionImpl;
 import org.apache.geode.distributed.internal.DistributionManager;
 import org.apache.geode.distributed.internal.InternalDistributedSystem;
 import org.apache.geode.distributed.internal.LonerDistributionManager;
 import org.apache.geode.distributed.internal.ReplyProcessor21;
+import org.apache.geode.internal.ByteBufferOutputStream;
 import org.apache.geode.internal.HeapDataOutputStream;
 import org.apache.geode.internal.SystemTimer;
 import org.apache.geode.internal.cache.BucketAdvisor;
@@ -93,6 +97,9 @@ import org.apache.geode.internal.cache.wan.GatewayReceiverStats;
 import org.apache.geode.internal.inet.LocalHostUtil;
 import org.apache.geode.internal.logging.CoreLoggingExecutors;
 import org.apache.geode.internal.monitoring.ThreadsMonitoring;
+import org.apache.geode.internal.net.BufferPool;
+import org.apache.geode.internal.net.ByteBufferSharing;
+import org.apache.geode.internal.net.NioSslEngine;
 import org.apache.geode.internal.net.SocketCloser;
 import org.apache.geode.internal.net.SocketCreator;
 import org.apache.geode.internal.security.SecurityService;
@@ -352,6 +359,8 @@ public class AcceptorImpl implements Acceptor, Runnable {
 
   private final SocketCloser socketCloser = new SocketCloser();
 
+  private final BufferPool bufferPool;
+
   /**
    * Constructs an AcceptorImpl for use within a CacheServer.
    *
@@ -499,18 +508,25 @@ public class AcceptorImpl implements Acceptor, Runnable {
       LinkedBlockingQueue<ByteBuffer> tmp_commQ = null;
       Set<ServerConnection> tmp_hs = null;
       SystemTimer tmp_timer = null;
+      BufferPool tmp_buffpool = null;
       if (isSelector()) {
         tmp_s = Selector.open(); // no longer catch ex to fix bug 36907
         tmp_q = new LinkedBlockingQueue<>();
         tmp_commQ = new LinkedBlockingQueue<>();
         tmp_hs = new HashSet<>(512);
         tmp_timer = new SystemTimer(internalCache.getDistributedSystem());
+        DistributionImpl distribution = (DistributionImpl) internalCache
+            .getInternalDistributedSystem().getDM().getDistribution();
+        if (distribution != null) {
+          tmp_buffpool = distribution.getDirectChannel().getConduit().getBufferPool();
+        }
       }
       selector = tmp_s;
       selectorQueue = tmp_q;
       commBufferQueue = tmp_commQ;
       selectorRegistrations = tmp_hs;
       hsTimer = tmp_timer;
+      bufferPool = tmp_buffpool;
       this.tcpNoDelay = tcpNoDelay;
     }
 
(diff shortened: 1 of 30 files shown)


geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/AcceptorImpl.java
```diff
@@ -1550,12 +1574,64 @@ public void refuseHandshake(OutputStream out, String message, byte exception) th
...
-  private boolean handOffQueueInitialization(Socket socket, CommunicationMode communicationMode) {
+  @Override
+  public void refuseHandshake(OutputStream out, String message, byte exception,
+      NioSslEngine sslEngine, Socket socket) throws IOException {
+    if (sslEngine == null) {
+      refuseHandshake(out, message, exception);
+      return;
+    }
+    try (ByteBufferOutputStream bbos =
+        new ByteBufferOutputStream(sslEngine.getEngine().getSession().getPacketBufferSize())) {
```
kirklund, line 1585: Here's another example of breaking the [Law of Demeter](https://en.wikipedia.org/wiki/Law_of_Demeter) and exposing internal knowledge that the acceptor shouldn't have to know about. It's better to embrace dependency injection and just pass in `int packetBufferSize` or even `javax.net.ssl.SSLSession sslSession` as a parameter.

One direction I think we really need to steer the code base towards is [dependency injection](https://en.wikipedia.org/wiki/Dependency_inversion_principle). The idea is to pass in all dependencies, either to the constructor or as parameters to a method call, instead of reaching out to fetch them.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/AcceptorImpl.java
```diff
@@ -1550,12 +1574,64 @@ public void refuseHandshake(OutputStream out, String message, byte exception) th
...
+      // Write refused reply
+      dos.writeByte(exception);
+
+      // write dummy endpointType
+      dos.writeByte(0);
+      // write dummy queueSize
+      dos.writeInt(0);
+
+      // Write the server's member
+      DistributedMember member = InternalDistributedSystem.getAnyInstance().getDistributedMember();
```
kirklund, line 1598: Here's another call to reach out for a dependency and it's even worse because it has to go to a singleton which we've been working to get rid of. This one might be an example of one that should be passed into the constructor since it never changes during the life of this acceptor instance:
```
private final DistributedMember member;

AcceptorImpl(..., DistributedMember member) {
  ...
  this.member = member;
}
```

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ServerConnection.java
```diff
@@ -254,6 +254,8 @@ public static void emptyCommBufferPool() {
   @MutableForTesting
   private static boolean TEST_VERSION_AFTER_HANDSHAKE_FLAG;
 
+  private NioSslEngine sslEngine;
```
kirklund, line 257: We need to strive to make fields like this `final`. There are several places in the new code where you close the `NioSslEngine` and set the field back to null. There are also lots of null checks. This makes the code thread unsafe because there's no form of synchronization or use of volatile. Non-final fields tend to lower performance and make the code a lot busier with null checks.

Why do you really need this to be mutable? And what is the state change for that changes this from non-null to null? Why does the `ServerConnection` remain valid and usable after closing and nulling out the `NioSslEngine`?

There are two possible approaches I would take:

1) I would try very hard to make `NioSslEngine sslEngine` a final field. 

I might even change it to be an interface so that I can either pass in a `NioSslEngine` or a dummy implementation of the interface to the constructor. A dummy implementation does nothing or returns values from methods that causes the calling code to not use nio ssl. 

2) If I really need the field to be mutable AND the class instance is exposed more than one thread, then I would use an AtomicReference to make sure it's thread-safe:
```
private final AtomicReference<NioSslEngine> sslEngine = new AtomicReference<>();
```
Using an interface with a dummy implementation instead of null can also be used with AtomicReferences.

kirklund (changes requested):

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/AcceptorImpl.java
```diff
@@ -499,18 +508,29 @@
...
       Set<ServerConnection> tmp_hs = null;
       SystemTimer tmp_timer = null;
+      BufferPool tmp_buffpool = null;
       if (isSelector()) {
         tmp_s = Selector.open(); // no longer catch ex to fix bug 36907
         tmp_q = new LinkedBlockingQueue<>();
         tmp_commQ = new LinkedBlockingQueue<>();
         tmp_hs = new HashSet<>(512);
         tmp_timer = new SystemTimer(internalCache.getDistributedSystem());
+        DistributionImpl distribution = (DistributionImpl) internalCache
```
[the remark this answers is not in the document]
mivanac: Thanks for comments. I changed this to create own pool for client-server communication.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/AcceptorImpl.java
mivanac: Thanks for comments. Updated.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/AcceptorImpl.java
```diff
@@ -1550,12 +1568,63 @@ public void refuseHandshake(OutputStream out, String message, byte exception) th
...
+      // Write dummy delta-propagation property value. This will never be read at
+      // receiver because the exception byte above will cause the receiver code
+      // throw an exception before the below byte could be read.
+      dos.writeBoolean(Boolean.TRUE);
+
+      bbos.flush();
+      ByteBuffer buffer = bbos.getContentBuffer();
+      try (final ByteBufferSharing outputSharing = sslEngine.wrap(buffer)) {
+        final ByteBuffer wrappedBuffer = outputSharing.getBuffer();
+        if (socket != null) {
```
bschuchardt, line 1612: Is this null check really needed?  If there's a sslEngine there ought to be a socket as well.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/Message.java
```diff
@@ -1160,10 +1220,119 @@ public void receive() throws IOException {
...
+        }
+
+        // now we need to read partLen - alreadyReadBytes off the wire
+        int off = alreadyReadBytes;
+        int remaining = partLen - off;
+
+        if (remaining > 0) {
+
+          this.sslEngine.ensureWrappedCapacity(remaining, cb,
+              BufferPool.BufferType.TRACKED_RECEIVER);
```
bschuchardt, line 1295: This overloads the DMStats statistic.  The descriptions of the buffer stats should be updated to reflect that cache server use of buffers is now included.

bschuchardt (changes requested):
Impressive amount of work.  I'm wondering why you chose to duplicate so much code and have null checks for NioSslEngine everywhere instead of using the NioFilter interface and NioPlainEngine for non-TLS communications.  That was set up specifically to avoid multiple code paths in the code using the engines.

I ran a lot of tests against this PR and didn't find any functional problems with it (aside from the improper DMStats updates).

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/AcceptorImpl.java
```diff
@@ -1115,7 +1126,7 @@ private void runSelectorLoop() {
             keysIterator.remove();
             final ServerConnection sc = (ServerConnection) key.attachment();
             try {
-              if (key.isValid() && key.isReadable()) {
+              if (key.isValid() && (key.isReadable() || key.isWritable())) {
```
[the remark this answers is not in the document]
mivanac: Without this, test are failing, and no connection can not be established.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/AcceptorImpl.java
mivanac: I will remove this.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/Message.java
mivanac: I am not sure how to report this statistics. Do we report bytes that are read, or unwrapped?

mivanac:
Thanks for comments. Regarding NioFilter, I did not look into it, because it seemed too complicated at that moment. Now when all other impacts are clear, I will try to incorporate it in solution.

DonalEvans (approved):
Approving the changes in `DeltaPropagationDUnitTest` which is the only file I'm a code owner for in this PR.

mivanac:
Non systemized solution. New PR open with solution.

PR #6075, opened 2021-03-02
GEODE-8943: Avoid filter registration during Transaction commit

agingade:

### For all changes:

### Note:


geode-cq/src/distributedTest/java/org/apache/geode/cache/query/cq/dunit/PartitionedRegionTxDUnitTest.java
```diff
@@ -0,0 +1,277 @@
...
+import org.apache.geode.test.dunit.DUnitBlackboard;
+import org.apache.geode.test.dunit.rules.ClientVM;
+import org.apache.geode.test.dunit.rules.ClusterStartupRule;
+import org.apache.geode.test.dunit.rules.MemberVM;
+import org.apache.geode.test.junit.categories.ClientSubscriptionTest;
+
+@Category(ClientSubscriptionTest.class)
+@RunWith(Parameterized.class)
+public class PartitionedRegionTxDUnitTest implements Serializable {
+  private static volatile DUnitBlackboard blackboard;
```
kirklund, line 67: Switch to using the DistributedBlackboard rule:
```
@Rule
```
You don't need to init or cleanup this blackboard. You just use it same as you use DUnitBlackboard.

geode-cq/src/distributedTest/java/org/apache/geode/cache/query/cq/dunit/PartitionedRegionTxDUnitTest.java
```diff
@@ -0,0 +1,277 @@
...
+    });
+
+    server2.invoke(() -> {
+      InternalCache cache = ClusterStartupRule.getCache();
+      cache.createRegionFactory(RegionShortcut.PARTITION).setPartitionAttributes(
+          new PartitionAttributesFactory().setRedundantCopies(1).setTotalNumBuckets(1).create())
+          .create(REGION_NAME);
+    });
+
+    AsyncInvocation serverAsync = server1.invokeAsync(() -> {
```
kirklund, line 117: With typing this becomes `AsyncInvocation<Void> serverAsync`

geode-core/src/test/java/org/apache/geode/internal/cache/FilterProfileTest.java
```diff
@@ -1,258 +0,0 @@
...
-import java.util.Collections;
-
-import org.junit.Before;
-import org.junit.Test;
-
-import org.apache.geode.cache.CacheEvent;
-import org.apache.geode.cache.Operation;
-import org.apache.geode.cache.query.internal.cq.CqService;
-
-public class FilterProfileTest {
```
kirklund, line 35: Do you really need to delete the `FilterProfileTest`?

pivotal-eshu:
There is a behavior change in transaction with register interest. (From my understanding this behavior exists from start of the tx implementation).
Now register interest will block transaction commit during the period of register interest -- esp. on the node performing the register interest from client. 
The performance hit shifts from calculating filterRoutingInfo again on remote node to no transactions allowed to commit when register interest is under way.
Also I worry about potential performance hit and even deadlock as the remote node processing register interest message can be blocked for a while (trying to get the write lock) while there are concurrent transactions.

agingade:
> There is a behavior change in transaction with register interest. (From my understanding this behavior exists from start of the tx implementation).
> Now register interest will block transaction commit during the period of register interest -- esp. on the node performing the register interest from client.
> The performance hit shifts from calculating filterRoutingInfo again on remote node to no transactions allowed to commit when register interest is under way.
> Also I worry about potential performance hit and even deadlock as the remote node processing register interest message can be blocked for a while (trying to get the write lock) while there are concurrent transactions.

Not sure what you meant by behavior change; can you be more explicit on this.
By design for partitioned regions, the events are processed on the primary bucket (where cache is updated) for all the peer-nodes and server hosting the subscription queues in questions. And replicated regions its processed locally in notifyBridgeClientes()

Yes. With the new change register interest is blocked while transaction in progress (and other way); this is one of the option that address the issue we are seeing. We could try out if there are any other better option to address this.

There could be impact when the registration and transaction in progress; but considering the use-case, where registration is meta operation (only happens initially when client comes up); it will have minimal impact. Also we have nice mentioned that overall transaction ops are slower compare to non-tx ops.

Yes, we need add/run tests to make sure there is no dead-lock issue within the node and between the nodes (distributed message).

geode-core/src/main/java/org/apache/geode/internal/cache/TXState.java
```diff
@@ -591,13 +593,14 @@ private void lockFilterRegistrationOnTxRegions() {
...
+        filterLockAcquiredRegions.add(region);
       }
     }
   }
 
   private void unlockFilterRegistrationOnTxRegions() {
-    for (InternalRegion region : getRegions()) {
-      if (lockableRegionForFilterRegistration(region)) {
+    for (InternalRegion region : filterLockAcquiredRegions) {
+      if (region != null) {
```
agingade, line 603: Will this region be destroyed; do we need check for destroyed regions...

geode-core/src/main/java/org/apache/geode/internal/cache/TxCallbackEventFactoryImpl.java
```diff
@@ -127,13 +126,21 @@ public EntryEventImpl createCallbackEvent(final InternalRegion internalRegion,
           }
         }
       } else if (computeFilterInfo) { // not a bucket
-        if (TxCallbackEventFactoryImpl.logger.isTraceEnabled()) {
-          TxCallbackEventFactoryImpl.logger.trace("createCBEvent computing routing for non-bucket");
-        }
-        FilterProfile fp = internalRegion.getFilterProfile();
-        if (fp != null) {
-          retVal.setLocalFilterInfo(fp.getLocalFilterRouting(retVal));
+        if (!isOriginRemote) {
```
agingade, line 129: Can both condition be replaced with AND condition. And move the comments from bottom to top...

geode-core/src/main/java/org/apache/geode/internal/cache/TXState.java
```diff
@@ -522,15 +522,15 @@ public void commit() throws CommitConflictException {
           // Take filter registration lock
           lockFilterRegistrationOnTxRegions();
           try {
+            attachFilterProfileInformation(entries);
+
```
agingade, line 526: This will work with locking changes...But still feel that, the attach should be called after applying to cache...
I believe the "applyChanges" is creating the CacheEvent/EntryEvent by not setting the old value; in that case can we change that code to set the old value....If the old value is not available in apply, can we get the old value as we are doing in attach and pass it to apply?

geode-core/src/main/java/org/apache/geode/internal/cache/TXState.java
pivotal-eshu: Even if region is destroyed, we still need to unlock the readLock it holds to avoid deadlock.

PR #6079, opened 2021-03-02
GEODE-8979: CI Failure: SSLSocketHostNameVerificationIntegrationTest

bschuchardt:
This test was closing a client socket before ensuring that a thread it
had created was finished.  If the socket is closed quickly enough it
could cause that thread to get an IOException and cause the test to
fail.

The fix is to ensure that the thread is finished before closing the
client socket in the tearDown method.  There are many other edits
to remove the use of "this.".

@kamilla1201 

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/integrationTest/java/org/apache/geode/internal/net/SSLSocketHostNameVerificationIntegrationTest.java
+++ b/geode-core/src/integrationTest/java/org/apache/geode/internal/net/SSLSocketHostNameVerificationIntegrationTest.java
@@ -55,6 +55,7 @@ import org.apache.geode.cache.ssl.CertificateMaterial;
 import org.apache.geode.distributed.internal.DMStats;
 import org.apache.geode.distributed.internal.DistributionConfig;
 import org.apache.geode.distributed.internal.DistributionConfigImpl;
+import org.apache.geode.test.awaitility.GeodeAwaitility;
 import org.apache.geode.test.dunit.IgnoredException;
 import org.apache.geode.test.junit.categories.MembershipTest;
 import org.apache.geode.test.junit.runners.CategoryWithParameterizedRunnerFactory;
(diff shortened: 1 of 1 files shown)


geode-core/src/integrationTest/java/org/apache/geode/internal/net/SSLSocketHostNameVerificationIntegrationTest.java
```diff
@@ -211,18 +211,17 @@ private Thread startServerNIO(final ServerSocket serverSocket, int timeoutMillis
       } catch (Throwable throwable) {
         serverException = throwable;
       } finally {
-        if (engine != null && socket != null) {
+        if (engine != null) {
```
echobravopapa, line 214: null check is no longer needed?

echobravopapa (approved):
I like the change, just curious about removing the null check...

geode-core/src/integrationTest/java/org/apache/geode/internal/net/SSLSocketHostNameVerificationIntegrationTest.java
bschuchardt: Intellij informed me that if (engine != null) it isn't possible that (socket == null).  I agreed & removed that check.

kamilla1201 (approved):

geode-core/src/integrationTest/java/org/apache/geode/internal/net/SSLSocketHostNameVerificationIntegrationTest.java
echobravopapa: what a helpful tool!

Bill (approved):
Fixes the problem and cleans up some syntax.

Good one Bruce!

PR #6083, opened 2021-03-03
GEODE-8998: fix NPE caused by thread-monitor-enabled=false

dschneider-pivotal:
The P2PReaderExecutorGroup is now called directly since the p2p reader always needs
a non-null instance. When thread monitoring is disabled, the p2p reader will still call
suspend and resume even though the P2PReaderExecutorGroup is not being monitored.
But these calls are very cheap so it is better to call them instead of doing a bunch of
null checks.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/monitoring/ThreadsMonitoringImplDummy.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/monitoring/ThreadsMonitoringImplDummy.java
@@ -32,17 +32,17 @@ public class ThreadsMonitoringImplDummy implements ThreadsMonitoring {
   @Override
   public void endMonitor() {}
 
-  private static class DummyAbstractExecutor extends AbstractExecutor {
-    private static final DummyAbstractExecutor SINGLETON = new DummyAbstractExecutor();
+  private static class DummyExecutor extends AbstractExecutor {
+    private static final DummyExecutor SINGLETON = new DummyExecutor();
 
-    private DummyAbstractExecutor() {
-      super("dummyExecutorGroup", 0L);
+    private DummyExecutor() {
+      super("DummyExecutor", 0L);
     }
   }
 
   @Override
   public AbstractExecutor createAbstractExecutor(Mode mode) {
-    return DummyAbstractExecutor.SINGLETON;
+    return DummyExecutor.SINGLETON;
   }
 
   @Override
--- a/geode-core/src/main/java/org/apache/geode/internal/monitoring/ThreadsMonitoringImplDummy.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/monitoring/ThreadsMonitoringImplDummy.java
@@ -17,6 +17,7 @@ package org.apache.geode.internal.monitoring;
 
 import java.util.concurrent.ConcurrentMap;
 
+import org.apache.geode.annotations.Immutable;
 import org.apache.geode.internal.monitoring.executor.AbstractExecutor;
 
 public class ThreadsMonitoringImplDummy implements ThreadsMonitoring {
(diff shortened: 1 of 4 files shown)


boglesby (approved):

bschuchardt (approved):
looks okay to me.  You could remove the switch-statement case for p2p readers from ThreadsMonitoring.createAbstractExecutor() since it's now dead code.

kirklund (approved):

Bill (changes requested):
This PR seems to defeat the original purpose of `ThreadsMonitoring.createAbstractExecutor(Mode)` which was ostensibly to map the enum values (of `Mode`) defined in `org.apache.geode.internal.monitoring` to implementation classes defined in `org.apache.geode.internal.monitoring.executor`. The point, it seems, was to provide some encapsulation (of the monitoring package.)

There are currently two calls to the factory method outside of tests:

1. `ThreadsMonitoringImpl.startMonitor(Mode)` in `org.apache.geode.internal.monitoring` which in turn is called in 6 places outside the monitoring package
2. `Connection.readMessages()` in `org.apache.geode.internal.tcp` (the call changed in this PR)

It seems like this PR breaks the encapsulation currently provided by the factory method.

Wouldn't it be better to modify `ThreadsMonitoringImplDummy.createAbstractExecutor(Mode)` to return a kind of `AbstractExecutor` that minimally functions but doesn't hurt anything?

That change could be localized to monitoring and could keep the established contract with callers outside monitoring. The added benefit is it eliminates a place where we are currently returning `null` which in turn could eliminate other null checks.

kirklund (approved):

boglesby (approved):

geode-core/src/main/java/org/apache/geode/internal/monitoring/ThreadsMonitoringImplDummy.java
```diff
@@ -32,9 +32,17 @@ public boolean startMonitor(Mode mode) {
   @Override
   public void endMonitor() {}
 
+  private static class DummyAbstractExecutor extends AbstractExecutor {
+    private static final DummyAbstractExecutor SINGLETON = new DummyAbstractExecutor();
+
+    private DummyAbstractExecutor() {
+      super("dummyExecutorGroup", 0L);
```
Bill, line 39: From looking at other group names they are typically capitalized like class names and don't end in the word `Group`. So I recommend:

`"DummyExecutor"`

geode-core/src/main/java/org/apache/geode/internal/monitoring/ThreadsMonitoringImplDummy.java
```diff
@@ -32,9 +32,17 @@ public boolean startMonitor(Mode mode) {
   @Override
   public void endMonitor() {}
 
+  private static class DummyAbstractExecutor extends AbstractExecutor {
```
Bill, line 35: I think this new class is not meant to be subclassed so I recommend removing `Abstract` from the name.

Bill (changes requested):
Minor changes requested.

geode-core/src/main/java/org/apache/geode/internal/monitoring/ThreadsMonitoringImplDummy.java
dschneider-pivotal: done

geode-core/src/main/java/org/apache/geode/internal/monitoring/ThreadsMonitoringImplDummy.java
dschneider-pivotal: done

Bill (approved):
Perfect!

PR #6085, opened 2021-03-03
GEODE-8965  Support Redis-style OOM error message

ringles:
This allows behavior simulating the Redis "noevict" out-of-memory policy. Operations that exceed the Geode configuration value of "critical-heap-percentage" will return a Redis-style "OOM" error message, which clients can respond to.

Because we want to allow things like deleting and expiring Redis keys, this add an "AllowExecutionInLowMemory" interface that allows functions to attempt to execute even in low memory conditions. This way, operations that free memory will be allowed, but operations that attempt to allocate new memory will still fail with LowMemoryException.

### DIFF
--- a/geode-redis/src/distributedTest/java/org/apache/geode/redis/OutOfMemoryDUnitTest.java
+++ b/geode-redis/src/distributedTest/java/org/apache/geode/redis/OutOfMemoryDUnitTest.java
@@ -15,12 +15,15 @@
 
 package org.apache.geode.redis;
 
+import static org.apache.geode.test.awaitility.GeodeAwaitility.await;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatNoException;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.assertj.core.api.AssertionsForClassTypes.catchThrowable;
 
 import java.util.Arrays;
 import java.util.Properties;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import org.junit.AfterClass;
 import org.junit.Before;
@@ -46,13 +49,13 @@ public class OutOfMemoryDUnitTest {
   public ExecutorServiceRule executor = new ExecutorServiceRule();
 
   private static final String expectedEx = "Member: .*? above .*? critical threshold";
-  public static final String FILLER_KEY = "fillerKey-";
-  public static final String PRESSURE_KEY = "pressureKey-";
+  private static final String FILLER_KEY = "fillerKey-";
+  private static final String PRESSURE_KEY = "pressureKey-";
   private static final String LOCAL_HOST = "127.0.0.1";
-  public static final int KEY_TTL_SECONDS = 10;
+  private static final int KEY_TTL_SECONDS = 10;
   private static final int MAX_ITERATION_COUNT = 4000;
-  public static final int LARGE_VALUE_SIZE = 128 * 1024;
-  public static final int PRESSURE_VALUE_SIZE = 4 * 1024;
+  private static final int LARGE_VALUE_SIZE = 128 * 1024;
+  private static final int PRESSURE_VALUE_SIZE = 4 * 1024;
   private static final int JEDIS_TIMEOUT =
       Math.toIntExact(GeodeAwaitility.getTimeout().toMillis());
   private static Jedis jedis1;
@@ -65,15 +68,11 @@ public class OutOfMemoryDUnitTest {
 
   @BeforeClass
   public static void classSetup() {
-    Properties serverProperties;
-    MemberVM locator;
-    int redisServerPort1;
-    int redisServerPort2;
-
     IgnoredException.addIgnoredException(expectedEx);
-    serverProperties = new Properties();
 
-    locator = clusterStartUp.startLocatorVM(0);
+    MemberVM locator = clusterStartUp.startLocatorVM(0);
+
+    Properties serverProperties = new Properties();
     server1 = clusterStartUp.startRedisVM(1, serverProperties, locator.getPort());
     server2 = clusterStartUp.startRedisVM(2, serverProperties, locator.getPort());
 
@@ -82,12 +81,11 @@ public class OutOfMemoryDUnitTest {
     server2.getVM().invoke(() -> RedisClusterStartupRule.getCache().getResourceManager()
         .setCriticalHeapPercentage(5.0F));
 
-    redisServerPort1 = clusterStartUp.getRedisPort(1);
-    redisServerPort2 = clusterStartUp.getRedisPort(2);
+    int redisServerPort1 = clusterStartUp.getRedisPort(1);
+    int redisServerPort2 = clusterStartUp.getRedisPort(2);
 
     jedis1 = new Jedis(LOCAL_HOST, redisServerPort1, JEDIS_TIMEOUT);
     jedis2 = new Jedis(LOCAL_HOST, redisServerPort2, JEDIS_TIMEOUT);
-
   }
 
   @Before
@@ -139,7 +137,9 @@ public class OutOfMemoryDUnitTest {
 
     fillMemory(jedis2, true);
 
-    GeodeAwaitility.await().until(() -> jedis2.ttl(FILLER_KEY + 1) == -2);
+    await().untilAsserted(() -> {
+      assertThat(jedis2.ttl(FILLER_KEY + 1)).isEqualTo(-2);
+    });
   }
 
   // TODO: test that write operations become allowed after memory has dropped
@@ -159,17 +159,20 @@ public class OutOfMemoryDUnitTest {
   }
 
   private void addMultipleKeys(Jedis jedis, String valueString, boolean withExpiration) {
-    int i = 0;
-    while (i < MAX_ITERATION_COUNT) {
-      try {
-        setRedisKeyAndValue(jedis, withExpiration, valueString, i);
-      } catch (JedisException je) {
-        assertThat(je).hasMessageContaining("OOM command not allowed");
-        break;
+    // count is final because it is never reassigned
+    AtomicInteger count = new AtomicInteger();
+
+    Throwable thrown = catchThrowable(() -> {
+      for (count.set(0); count.get() < MAX_ITERATION_COUNT; count.incrementAndGet()) {
+        setRedisKeyAndValue(jedis, withExpiration, valueString, count.get());
       }
-      i++;
-    }
-    assertThat(i).isLessThan(MAX_ITERATION_COUNT);
+    });
+
+    assertThat(thrown)
+        .isInstanceOf(Exception.class)
+        .hasMessageContaining("OOM command not allowed");
+
+    assertThat(count.get()).isLessThan(MAX_ITERATION_COUNT);
   }
 
   private void setRedisKeyAndValue(Jedis jedis, boolean withExpiration, String valueString,
--- a/geode-redis/src/distributedTest/java/org/apache/geode/redis/OutOfMemoryDUnitTest.java
+++ b/geode-redis/src/distributedTest/java/org/apache/geode/redis/OutOfMemoryDUnitTest.java
@@ -47,11 +47,12 @@ public class OutOfMemoryDUnitTest {
 
   private static final String expectedEx = "Member: .*? above .*? critical threshold";
   public static final String FILLER_KEY = "fillerKey-";
+  public static final String PRESSURE_KEY = "pressureKey-";
   private static final String LOCAL_HOST = "127.0.0.1";
   public static final int KEY_TTL_SECONDS = 10;
   private static final int MAX_ITERATION_COUNT = 4000;
   public static final int LARGE_VALUE_SIZE = 128 * 1024;
-  public static final int SMALL_VALUE_SIZE = 16 * 1024;
+  public static final int PRESSURE_VALUE_SIZE = 4 * 1024;
   private static final int JEDIS_TIMEOUT =
       Math.toIntExact(GeodeAwaitility.getTimeout().toMillis());
   private static Jedis jedis1;
@@ -59,6 +60,8 @@ public class OutOfMemoryDUnitTest {
 
   private static MemberVM server1;
   private static MemberVM server2;
+
+  private static Thread memoryPressureThread;
 
   @BeforeClass
   public static void classSetup() {
@@ -84,6 +87,7 @@ public class OutOfMemoryDUnitTest {
 
     jedis1 = new Jedis(LOCAL_HOST, redisServerPort1, JEDIS_TIMEOUT);
     jedis2 = new Jedis(LOCAL_HOST, redisServerPort2, JEDIS_TIMEOUT);
+
   }
 
   @Before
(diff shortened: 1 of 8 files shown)


geode-redis/src/main/java/org/apache/geode/redis/internal/netty/Coder.java
```diff
@@ -278,6 +281,16 @@ public static ByteBuf getErrorResponse(ByteBufAllocator alloc, String error) {
     return response;
   }
 
+  public static ByteBuf getOOMResponse(ByteBufAllocator alloc, String error) {
+    byte[] errorAr = stringToBytes(error);
+    ByteBuf response = alloc.buffer(errorAr.length + 25);
```
DonalEvans, line 286: The reason that a value of 25 specifically is added to the buffer size here is not clear to me. Other methods in this class, (`getCustomErrorResponse()` for instance) also add 25 to the buffer size, but write fewer bytes to the buffer in total. Is there some way of programmatically determining what the ideal buffer size should be given the expected contents? Failing that, just adding an explanatory comment would go a long way to making the code easier to maintain.

geode-redis/src/distributedTest/java/org/apache/geode/redis/OutOfMemoryDUnitTest.java
```diff
@@ -0,0 +1,175 @@
...
+import redis.clients.jedis.Jedis;
+import redis.clients.jedis.exceptions.JedisException;
+
+import org.apache.geode.test.awaitility.GeodeAwaitility;
+import org.apache.geode.test.dunit.IgnoredException;
+import org.apache.geode.test.dunit.rules.MemberVM;
+import org.apache.geode.test.dunit.rules.RedisClusterStartupRule;
+import org.apache.geode.test.junit.rules.ExecutorServiceRule;
+
+@SuppressWarnings("unchecked")
```
DonalEvans, line 41: This suppression is not needed.

geode-redis/src/distributedTest/java/org/apache/geode/redis/OutOfMemoryDUnitTest.java
```diff
@@ -0,0 +1,175 @@
...
+  private static final String LOCAL_HOST = "127.0.0.1";
+  public static final int KEY_TTL_SECONDS = 10;
+  private static final int MAX_ITERATION_COUNT = 4000;
+  public static final int VALUE_SIZE = 128 * 1024;
+  private static final int JEDIS_TIMEOUT =
+      Math.toIntExact(GeodeAwaitility.getTimeout().toMillis());
+  private static Jedis jedis1;
+  private static Jedis jedis2;
+
+  private static Properties locatorProperties;
```
DonalEvans, line 61: `locatorProperties`, `serverProperties`, `locator`, `redisServerPort1` and `redisServerPort2` can all be converted to local variables, as they're only used in the `classSetup()` method.

geode-redis/src/distributedTest/java/org/apache/geode/redis/OutOfMemoryDUnitTest.java
```diff
@@ -0,0 +1,175 @@
...
+    IgnoredException.addIgnoredException(expectedEx);
+    locatorProperties = new Properties();
+    serverProperties = new Properties();
+    locatorProperties.setProperty(MAX_WAIT_TIME_RECONNECT, "15000");
+
+    locator = clusterStartUp.startLocatorVM(0, locatorProperties);
+    server1 = clusterStartUp.startRedisVM(1, serverProperties, locator.getPort());
+    server2 = clusterStartUp.startRedisVM(2, serverProperties, locator.getPort());
+
+    server1.getVM().invoke(() -> {
```
DonalEvans, line 83: This invocation and the one below can be changed into one-line lambdas, without the need for curly brackets, which neatens things up a little.

geode-redis/src/distributedTest/java/org/apache/geode/redis/OutOfMemoryDUnitTest.java
```diff
@@ -0,0 +1,175 @@
...
+          jedis.setex(FILLER_KEY + i, KEY_TTL_SECONDS, valueString);
+        } else {
+          jedis.set(FILLER_KEY + i, valueString);
+        }
+      } catch (JedisException je) {
+        assertThat(je).hasMessageContaining("OOM command not allowed");
+        break;
+      }
+      i++;
+    }
```
sabbey37, line 164: It looks like the `shouldReturnOOMError_forWriteOperations_whenThresholdReached` is failing in CI because the OOM error never gets thrown.  I'm wondering if we're not filling the memory entirely.  Maybe you could do a while(hasMemoryAvailable){} loop (`hasMemoryAvailable` being a boolean variable that's initially set to `true`) then set `hasMemoryAvailable` to `false` when the `JedisException` is caught to ensure the memory is full beforehand.

sabbey37 (changes requested):

geode-redis/src/distributedTest/java/org/apache/geode/redis/OutOfMemoryDUnitTest.java
```diff
@@ -0,0 +1,175 @@
...
+    fillMemory(jedis1, MAX_ITERATION_COUNT, true);
+
+    GeodeAwaitility.await().until(() -> jedis2.ttl(FILLER_KEY + 1) == -2);
+  }
+
+  // TODO: test that write operations become allowed after memory has dropped
+  // below critical levels. Difficult to do right now because of vagaries of the
+  // Java garbage collector.
+
+  private int fillMemory(Jedis jedis, int maxIterations, boolean withExpiration) {
```
DonalEvans, line 150: The return value of this method is never used, so the method signature could be changed to return `void`. Also, unless there is a situation where `maxIterations` will not be `MAX_ITERATION_COUNT` then this argument could be removed and the constant used directly.

geode-redis/src/distributedTest/java/org/apache/geode/redis/OutOfMemoryDUnitTest.java
```diff
@@ -0,0 +1,175 @@
...
+        break;
+      }
+      i++;
+    }
+    assertThat(i).isLessThan(maxIterations);
+    return i;
+  }
+
+  // TODO: use this when write testing figured out
+  private void deleteKeysToClearMemory(Jedis jedis, int keysToDelete) {
```
DonalEvans, line 170: Just personal preference, but especially for methods this small, if they're not needed now but may be in future, I feel like they should be added with the code that actually uses them.

DonalEvans (approved):

DonalEvans:
Forgot to add my comment when applying my review: Just some small code clean-up suggestions, and a request for some clarification in the `Coder` class.

geode-redis/src/main/java/org/apache/geode/redis/internal/netty/Coder.java
jdeppe-pivotal: This is inherited from the original author(s) and we don't know what this is for. We've also found excessive (and unnecessary) allocations in this class which will shortly be corrected. That change will also remove these weird constants.

geode-redis/src/distributedTest/java/org/apache/geode/redis/OutOfMemoryDUnitTest.java
```diff
@@ -0,0 +1,181 @@
...
+
+  private static int redisServerPort1;
+  private static int redisServerPort2;
+
+  @BeforeClass
+  public static void classSetup() {
+    IgnoredException.addIgnoredException(expectedEx);
+    locatorProperties = new Properties();
+    serverProperties = new Properties();
+    locatorProperties.setProperty(MAX_WAIT_TIME_RECONNECT, "15000");
```
jdeppe-pivotal, line 77: Since each test is already adding ignored exceptions, this can probably be removed here.
`serverProperties` is empty and can be removed.
The locator doesn't need the reconnect setting, so also doesn't need any properties.

geode-redis/src/distributedTest/java/org/apache/geode/redis/OutOfMemoryDUnitTest.java
ringles: It's more complicated than that. The latest iteration does a GC before the 'fill' operation, and adds progressively smaller keys to cram the memory as full as possible. It still fails every so often, apparently because we don't have fine enough control over other usages of the JVM memory.

geode-redis/src/distributedTest/java/org/apache/geode/redis/OutOfMemoryDUnitTest.java
ringles: Spiked

geode-redis/src/distributedTest/java/org/apache/geode/redis/OutOfMemoryDUnitTest.java
ringles: Done!

geode-redis/src/distributedTest/java/org/apache/geode/redis/OutOfMemoryDUnitTest.java
ringles: Yeah, that was iffy, but it's yanked now.

geode-redis/src/distributedTest/java/org/apache/geode/redis/OutOfMemoryDUnitTest.java
ringles: The properties are removed, but we do actually need the IgnoredException here or we get issues in the Class instantiation.

geode-redis/src/distributedTest/java/org/apache/geode/redis/OutOfMemoryDUnitTest.java
DonalEvans: Might a possible solution be to spin off a separate thread that continuously adds keys to maintain memory pressure, so that even if something does free up some memory after we first hit the OOM error in `fillMemory()`, we're able to get back to a low memory situation?

geode-redis/src/distributedTest/java/org/apache/geode/redis/OutOfMemoryDUnitTest.java
```diff
@@ -0,0 +1,188 @@
...
+
+    while (valueSize > 1) {
+      valueString = makeLongStringValue(LARGE_VALUE_SIZE);
+      addMultipleKeys(jedis, valueString, withExpiration);
+      valueSize /= 2;
+    }
+  }
+
+  private void addMultipleKeys(Jedis jedis, String valueString, boolean withExpiration) {
+    int i = 0;
```
kirklund, line 152: For-loop with `catchThrowable` is another option here. `catchThrowable` is using a lambda and lambdas cannot use a non-final int so... I used an `AtomicInteger` which is final but the value it contains is mutable.
```

// count is final because it is never reassigned
AtomicInteger count = new AtomicInteger();

Throwable thrown = catchThrowable(() -> {
  for (count.set(0); count.get() < MAX_ITERATION_COUNT; count.incrementAndGet()) {
    setRedisKeyAndValue(jedis, withExpiration, valueString, count.get());
  }
});

assertThat(thrown)
    .isInstanceOf(Exception.class)
    .hasMessageContaining("OOM command not allowed");

assertThat(count.get()).isLessThan(MAX_ITERATION_COUNT);
```

geode-redis/src/distributedTest/java/org/apache/geode/redis/OutOfMemoryDUnitTest.java
```diff
@@ -0,0 +1,188 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+import redis.clients.jedis.Jedis;
+import redis.clients.jedis.exceptions.JedisException;
+
+import org.apache.geode.test.awaitility.GeodeAwaitility;
+import org.apache.geode.test.dunit.IgnoredException;
+import org.apache.geode.test.dunit.rules.MemberVM;
+import org.apache.geode.test.dunit.rules.RedisClusterStartupRule;
+import org.apache.geode.test.junit.rules.ExecutorServiceRule;
```
kirklund, line 38: Just in case you find it useful in the future...

There is also a `org.apache.geode.test.dunit.rules.DistributedExecutorServiceRule` which can be used from within any DUnit VM.

kirklund (changes requested):
Any JVM that throws a VirtualMachineError, such as OutOfMemoryError, is effectively broken and should not be used for anything after that (even another test). So I recommend changing from static "class" scaffolding to per-test-method scaffolding so that every DUnit VM is replaced with fresh JVMs for each test. Also make sure you never add a future test that causes the main JUnit JVM to run OOM since that JVM must remain around for every test-method in the test class.

Testing OOM and using GC APIs is notorious for resulting in flakiness so be aware of that and keep an eye on it!

geode-redis/src/distributedTest/java/org/apache/geode/redis/OutOfMemoryDUnitTest.java
```diff
@@ -0,0 +1,188 @@
...
+public class OutOfMemoryDUnitTest {
+
+  @ClassRule
+  public static RedisClusterStartupRule clusterStartUp = new RedisClusterStartupRule(4);
+
+  @Rule
+  public ExecutorServiceRule executor = new ExecutorServiceRule();
+
+  private static final String expectedEx = "Member: .*? above .*? critical threshold";
+  public static final String FILLER_KEY = "fillerKey-";
```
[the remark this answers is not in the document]
ringles: Done!

(BTW, the intent is to test Geode's low memory handling, not the JVM. We're setting the CriticalHeapPercentage to force LowMemoryExceptions quickly. At that level, there should be no chance of actually filling up the JVM's heap.)

geode-redis/src/distributedTest/java/org/apache/geode/redis/OutOfMemoryDUnitTest.java
```diff
@@ -0,0 +1,188 @@
...
+      Math.toIntExact(GeodeAwaitility.getTimeout().toMillis());
+  private static Jedis jedis1;
+  private static Jedis jedis2;
+
+  private static MemberVM server1;
+  private static MemberVM server2;
+
+  @BeforeClass
+  public static void classSetup() {
+    Properties serverProperties;
```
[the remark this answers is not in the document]
ringles: Done, for the rest of the items in classSetup().

geode-redis/src/distributedTest/java/org/apache/geode/redis/OutOfMemoryDUnitTest.java
```diff
@@ -0,0 +1,188 @@
...
+  }
+
+  @Test
+  public void shouldAllowExpiration_afterThresholdReached() {
+    IgnoredException.addIgnoredException(expectedEx);
+    IgnoredException.addIgnoredException("LowMemoryException");
+
+    fillMemory(jedis1, true);
+
+    GeodeAwaitility.await().until(() -> jedis2.ttl(FILLER_KEY + 1) == -2);
```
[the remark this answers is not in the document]
ringles: That is clearer, thanks!

geode-redis/src/distributedTest/java/org/apache/geode/redis/OutOfMemoryDUnitTest.java
ringles: 👍

geode-redis/src/distributedTest/java/org/apache/geode/redis/OutOfMemoryDUnitTest.java
ringles: I added such a thread - maybe not entirely elegantly yet - and altered fillMemory() in a couple ways. First, it does multiple forceGC() calls, to help ensure that we're filling the memory with our Redis data. Secondly, it loops multiple times with progressively smaller chunks of data, to help ensure every nook and cranny is filled.

Haven't seen it fail YET, but I have to admit I'm still worried about flakiness.

geode-redis/src/distributedTest/java/org/apache/geode/redis/OutOfMemoryDUnitTest.java
ringles: I like it!

geode-redis/src/distributedTest/java/org/apache/geode/redis/OutOfMemoryDUnitTest.java
```diff
@@ -0,0 +1,225 @@
...
+  public void shouldAllowExpiration_afterThresholdReached() {
+    IgnoredException.addIgnoredException(expectedEx);
+    IgnoredException.addIgnoredException("LowMemoryException");
+
+    fillMemory(jedis2, true);
+
+    await().untilAsserted(() -> {
+      assertThat(jedis2.ttl(FILLER_KEY + 1)).isEqualTo(-2);
+    });
+  }
```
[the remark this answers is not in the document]
ringles: Technically for those tests, we mostly care that we get to a state we can execute such commands after the first OOM. But it certainly can't hurt, so I've add the memory-pressure thread to the other tests, too.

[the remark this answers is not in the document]
sabbey37: Thanks, Ray!

sabbey37 (approved):
Looks good, Ray!

PR #6093, opened 2021-03-04
GEODE-9003: Remove the member from replicates as GII candidate if it'…

gesterzhou:
…s not part of the same distributed system

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/persistence/PersistentRecoveryOrderDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/persistence/PersistentRecoveryOrderDUnitTest.java
@@ -877,7 +877,8 @@ public class PersistentRecoveryOrderDUnitTest extends CacheTestCase {
           createReplicateRegion(regionName, getDiskDirs(getVMId()));
         });
         assertThat(thrown).isInstanceOf(ConflictingPersistentDataException.class);
-        assertThat(thrown.getMessage()).contains("is split-brained from all other members.");
+        assertThat(thrown.getMessage())
+            .contains("was not part of the same distributed system as the local data");
       }
     });
   }
--- a/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/persistence/PersistentRecoveryOrderDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/persistence/PersistentRecoveryOrderDUnitTest.java
@@ -881,6 +881,32 @@ public class PersistentRecoveryOrderDUnitTest extends CacheTestCase {
     });
   }
 
+  @Test
+  public void testRecoverableSplitBrain() {
+    vm2.invoke(() -> {
+      createReplicateRegion(regionName, getDiskDirs(getVMId()));
+    });
+    vm0.invoke(() -> {
+      createReplicateRegion(regionName, getDiskDirs(getVMId()));
+      putEntry("A", "B");
+      getCache().getRegion(regionName).close();
+    });
+
+    vm1.invoke(() -> {
+      createReplicateRegion(regionName, getDiskDirs(getVMId()));
+      updateEntry("A", "C");
+      getCache().getRegion(regionName).close();
+    });
+
+    // VM0 doesn't know that VM1 ever existed so it will start up.
+    vm0.invoke(() -> createReplicateRegion(regionName, getDiskDirs(getVMId())));
+
+    vm1.invoke(() -> {
+      createReplicateRegion(regionName, getDiskDirs(getVMId()));
+      validateEntry("A", "C");
+    });
+  }
+
   /**
    * Test to make sure that if if a member crashes while a GII is in progress, we wait for the
    * member to come back for starting.
--- a/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/persistence/PersistentRecoveryOrderDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/persistence/PersistentRecoveryOrderDUnitTest.java
@@ -877,6 +877,7 @@ public class PersistentRecoveryOrderDUnitTest extends CacheTestCase {
           createReplicateRegion(regionName, getDiskDirs(getVMId()));
         });
         assertThat(thrown).isInstanceOf(ConflictingPersistentDataException.class);
+        assertThat(thrown.getMessage()).contains("is split-brained from all other members.");
       }
     });
   }
@@ -894,12 +895,16 @@ public class PersistentRecoveryOrderDUnitTest extends CacheTestCase {
 
     vm1.invoke(() -> {
       createReplicateRegion(regionName, getDiskDirs(getVMId()));
+      validateEntry("A", "B");
       updateEntry("A", "C");
       getCache().getRegion(regionName).close();
     });
 
     // VM0 doesn't know that VM1 ever existed so it will start up.
-    vm0.invoke(() -> createReplicateRegion(regionName, getDiskDirs(getVMId())));
+    vm0.invoke(() -> {
+      createReplicateRegion(regionName, getDiskDirs(getVMId()));
+      validateEntry("A", "C");
+    });
 
     vm1.invoke(() -> {
       createReplicateRegion(regionName, getDiskDirs(getVMId()));
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/persistence/PersistenceAdvisorImpl.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/persistence/PersistenceAdvisorImpl.java
@@ -514,8 +514,10 @@ public class PersistenceAdvisorImpl implements InternalPersistenceAdvisor {
     persistenceAdvisorObserver.observe(regionPath);
 
     boolean equal = false;
-    for (Map.Entry<InternalDistributedMember, PersistentMemberState> entry : remoteStates
-        .getStateOnPeers().entrySet()) {
+    Iterator<Map.Entry<InternalDistributedMember, PersistentMemberState>> iterator =
+        remoteStates.getStateOnPeers().entrySet().iterator();
+    while (iterator.hasNext()) {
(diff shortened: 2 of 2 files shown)


geode-core/src/main/java/org/apache/geode/internal/cache/persistence/PersistenceAdvisorImpl.java
```diff
@@ -533,7 +535,8 @@ public boolean checkMyStateOnMembers(Set<InternalDistributedMember> replicates)
         String message = String.format(
             "Region %s remote member %s with persistent data %s was not part of the same distributed system as the local data from %s",
             regionPath, member, remoteId, myId);
-        throw new ConflictingPersistentDataException(message);
+        logger.warn(message);
+        iterator.remove();
```
jchen21, line 539: Is synchronization neededhere? The `stateOnPeers` maybe shared with other threads.

geode-core/src/main/java/org/apache/geode/internal/cache/persistence/PersistenceAdvisorImpl.java
```diff
@@ -533,7 +535,8 @@ public boolean checkMyStateOnMembers(Set<InternalDistributedMember> replicates)
         String message = String.format(
             "Region %s remote member %s with persistent data %s was not part of the same distributed system as the local data from %s",
             regionPath, member, remoteId, myId);
-        throw new ConflictingPersistentDataException(message);
+        logger.warn(message);
```
jchen21, line 538: How about `logger.error(message)`?

jchen21 (commented):
You can create the pull request in `Draft` mode.

geode-core/src/main/java/org/apache/geode/internal/cache/persistence/PersistenceAdvisorImpl.java
gesterzhou: it won't. This data structure is in reply message. Only the message processor will use it

geode-core/src/main/java/org/apache/geode/internal/cache/persistence/PersistenceAdvisorImpl.java
gesterzhou: it's not an error. It's race condition cause of the member to be temporarily out of sync. Geode can auto fix it. We just don't need this out-of-sync member to be a GII provider candidate.

dschneider-pivotal: A warning message will cause customers and support to think something is wrong. Is the user suppose to take same action based on this message? If you are correct that this is something geode will deal with automatically then at the most this should be an "info" message or "debug".

geode-core/src/main/java/org/apache/geode/internal/cache/persistence/PersistenceAdvisorImpl.java
dschneider-pivotal, line 539: I don't understand why you are going to the trouble of doing this remove.
This remove will modify the HashMap returned by remoteStates.getStateOnPeers(). This is the field "stateOnPeers" in PersistentStateQueryResults. But this instance of PersistentStateQueryResults is created in this method on line 512:     PersistentStateQueryResults remoteStates = getMyStateOnMembers(replicates);
So it seems to me like no other thread will have this same instance. fetchPersistentStateQueryResults always sends a message and then this instance comes back as the result. So I can't find any other code that would ever see this entry that you are removing. So it seems to me like this entire PR could just be not to throw the exception. If you get rid of the remove then you don't need the iterator change. Am I missing something?

geode-core/src/main/java/org/apache/geode/internal/cache/persistence/PersistenceAdvisorImpl.java
gesterzhou: OK, I will change to info

geode-core/src/main/java/org/apache/geode/internal/cache/persistence/PersistenceAdvisorImpl.java
gesterzhou: I made a mistake. I want to remove the member from replicates. I will fix it

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/persistence/PersistentRecoveryOrderDUnitTest.java
```diff
@@ -881,6 +881,32 @@ public void testSplitBrain() {
...
+    });
+
+    vm1.invoke(() -> {
+      createReplicateRegion(regionName, getDiskDirs(getVMId()));
+      updateEntry("A", "C");
+      getCache().getRegion(regionName).close();
+    });
+
+    // VM0 doesn't know that VM1 ever existed so it will start up.
+    vm0.invoke(() -> createReplicateRegion(regionName, getDiskDirs(getVMId())));
```
jchen21, line 902: Can we also do `validateEntry("A", "C");` on `vm0`?

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/persistence/PersistentRecoveryOrderDUnitTest.java
```diff
@@ -881,6 +881,32 @@ public void testSplitBrain() {
     });
   }
 
+  @Test
+  public void testRecoverableSplitBrain() {
```
jchen21, line 885: I am afraid this test does not cover the code change you have in `PersistenceAdvisorImpl`.  I don't see the log message of line 537 in `PersistenceAdvisorImpl` in the test run. Nor did I see `ConflictingPersistentDataException` in the test run. In the test, `vm2` is always online. Perhaps this is not a scenario of split brain that you expect. Maybe I am missing something.

jchen21 (commented):
I am trying to understand the split brain scenario that you want to cover. However, when I look at the JIRA https://issues.apache.org/jira/browse/GEODE-9003. There is no description. I don't see problem or solution description in the pull request either. It would be better to put some description of the issue in the JIRA or this pull request. That would help the others understand the issue and your fix. And it helps the code maintenance as well.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/persistence/PersistentRecoveryOrderDUnitTest.java
jchen21: This test runs successfully even without your code change in `PersistenceAdvisorImpl`. So either the test has to be modified or the fix on the product has to be updated.

gesterzhou: I fixed to enhance the test cases.

geode-core/src/main/java/org/apache/geode/internal/cache/persistence/PersistenceAdvisorImpl.java
```diff
@@ -552,6 +555,10 @@ public boolean checkMyStateOnMembers(Set<InternalDistributedMember> replicates)
         }
       }
     }
+    if (replicates.isEmpty() && splitBrainHappened) {
+      throw new ConflictingPersistentDataException(String
+          .format("Region %s from %s is split-brained from all other members.", regionPath, myId));
```
dschneider-pivotal, line 560: I think "is split-brained" needs to be improved. Base this message on the info one you have above. How about: "Region %s from %s was not part of the same distributed system as the existing online members"

dschneider-pivotal (changes requested):

geode-core/src/main/java/org/apache/geode/internal/cache/persistence/PersistenceAdvisorImpl.java
gesterzhou: fixed.

geode-core/src/main/java/org/apache/geode/internal/cache/persistence/PersistenceAdvisorImpl.java
```diff
@@ -552,6 +555,10 @@ public boolean checkMyStateOnMembers(Set<InternalDistributedMember> replicates)
         }
       }
     }
+    if (replicates.isEmpty() && splitBrainHappened) {
```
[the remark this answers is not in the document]
gesterzhou: fixed

dschneider-pivotal (approved):

geode-core/src/main/java/org/apache/geode/internal/cache/persistence/PersistenceAdvisorImpl.java
```diff
@@ -533,7 +533,12 @@ public boolean checkMyStateOnMembers(Set<InternalDistributedMember> replicates)
         String message = String.format(
             "Region %s remote member %s with persistent data %s was not part of the same distributed system as the local data from %s",
             regionPath, member, remoteId, myId);
-        throw new ConflictingPersistentDataException(message);
+        replicates.remove(member);
```
upthewaterspout, line 536: We are modifying `replicates`, which is a set that was passed into this method. I'm not quite sure what is going to happen to later logic if we have removed some of the replicates from the set that gets passed in. So I think it would be better to make a copy of replicates at the top of the method, to make sure we are not breaking other logic.

Otherwise, changes look good!

gesterzhou: The replicates are supposed to be adjusted here. We should not create a copy.
the whole method is to remove some of the entries from the replicates. It will remove the EQUAL ones, and offline ones, then I add the logic to remove the ones who don't know me. That means I will not GII from all these removed replicates as candidates. 
If the method removed all the replicates by other cases (such as EQUAL or offline), then I will recover from my own disk. 
As long as there's one entry in replicates, will request GII from it. 
If the last replicate is removed by this use case, should throw exception.

PR #6095, opened 2021-03-05
GEODE-8953: Re-introduce transaction details regarding non-transactional changes

davebarnes97:
Restore some cautionary text that was removed during an earlier re-write of the transaction section.

### DIFF
--- a/geode-docs/developing/transactions/design_considerations.html.md.erb
+++ b/geode-docs/developing/transactions/design_considerations.html.md.erb
@@ -33,6 +33,7 @@ This section discusses how transactions interact with other
 -  **[Mixing Transactions with Queries and Indexes](#transactions-queries)**
 -  **[Mixing Transactions with Eviction](#transactions-eviction)**
 -  **[Mixing Transactions with Expiration](#transactions-expiration)**
+-  **[Mixing Transactions with Non-transactional Operations](#transactions-nontransactions)**
 -  **[Changing the Handling of Dirty Reads](#transactions-dirty-reads)**
 
 ## <a id="colocate-PRs" class="no-quick-link"></a>Colocate Partitioned Regions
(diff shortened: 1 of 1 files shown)


albertogpz (approved):
Looks good to me. Thanks for addressing this.

geode-docs/developing/transactions/design_considerations.html.md.erb
```diff
@@ -117,6 +118,21 @@ immediately after the commit.
 A transaction disables expiration on any region entries affected
 by the transaction.
 
+## <a id="transactions-nontransactions" class="no-quick-link"></a>Mixing Transactions with Nontransactional Operations
```
karensmolermiller, line 121: Add hyphen to "Nontransactional" to become Non-transactional.

geode-docs/developing/transactions/design_considerations.html.md.erb
```diff
@@ -117,6 +118,21 @@ immediately after the commit.
...
+
+For best performance, non-transactional operations do not acquire the exclusive locks used to check
+for conflicts in a transaction. A transaction operating on the same data as a non-transactional actor
+is unable to detect the conflict caused by a non-transactional operation.
+
+A user application should adopt the policy of always using transactions or no transactions at all
```
karensmolermiller, line 127: The phrase "no transactions at all" feels awkward to me. Maybe better (?):
> A user application should adopt the policy of always using transactions or the policy of not using transactions at all for certain regions or sets of entries, so transactional entries will not be modified by operations
outside of the transaction.

karensmolermiller (changes requested):

geode-docs/developing/transactions/design_considerations.html.md.erb
davebarnes97: Good catch, thanks.

geode-docs/developing/transactions/design_considerations.html.md.erb
davebarnes97: I've replaced this paragraph with:
If using transactions, an application should adopt the policy of designating certain regions or sets of entries
exclusively for transactional puts, updates, and deletions, so transactional entries will not be modified by non-transactional.
operations.

karensmolermiller (approved):

PR #6096, opened 2021-03-05
GEODE-9004: Fix issues with queries targeting a Map field

albertogpz:
Several changes have been made to solve issues
found when running queries targeting a Map field
with and without indexes in the Map field.

- For queries without an index in the Map field
the behavior has been changed so that
for queries like "positions['SUN'] = null" instead
of returning those entries fulfiling
that condition and also those entries
where the Map does not contain the 'SUN' key
or the Map is null, it now returns
only entries where the Map contains the
'SUN' key and the value is equal to null.

- As a consequence of the above, queries
without an index in the Map field like
"positions['SUN'] != null" will now return
any entry that does not contain
the positions field Map with the 'SUN' key
and with value equal to null.

- For queries with an index in the Map field,
either specifying several keys or '*'
the behavior has been modified (fixed) so that
"!=" conditions targeting the indexed
Map fields return the same entries returned
when the index is not present. Prior to this
change, only entries containing the Map
were returned but if the Map was not
present in the entry then that entry was
not returned.

- Indexes for a field Map
specifying just one key, e.g.
positions['SUN'] have been changed
internally from Range indexes
to Map indexes in order to align
the behavior with Map indexes
(those that specified more than
one key or '*' for a Map field)
so that the space taken by this type
of index is equivalent to the
one taken by one specifying
more than one key for the Map.
This has also solved some wrong
behavior of these indexes that
sometimes did not return the expected
number of entries.
The drawback of this change is that
"!=" queries for the map field
will not make use of the index.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/integrationTest/java/org/apache/geode/cache/query/functional/INOperatorJUnitTest.java
+++ b/geode-core/src/integrationTest/java/org/apache/geode/cache/query/functional/INOperatorJUnitTest.java
@@ -225,11 +225,11 @@ public class INOperatorJUnitTest {
 
     q = CacheUtils.getQueryService().newQuery(" UNDEFINED IN SET(UNDEFINED)");
     result = q.execute();
-    assertThat(result).isEqualTo(QueryService.UNDEFINED);
+    assertThat(result).isEqualTo(TRUE);
 
     q = CacheUtils.getQueryService().newQuery(" UNDEFINED IN SET(UNDEFINED,UNDEFINED)");
     result = q.execute();
-    assertThat(result).isEqualTo(QueryService.UNDEFINED);
+    assertThat(result).isEqualTo(TRUE);
   }
 
   @Test
--- a/geode-core/src/integrationTest/java/org/apache/geode/cache/query/functional/IndexOperatorJUnitTest.java
+++ b/geode-core/src/integrationTest/java/org/apache/geode/cache/query/functional/IndexOperatorJUnitTest.java
@@ -145,7 +145,7 @@ public class IndexOperatorJUnitTest {
     map.put("0", new Integer(11));
     map.put("1", new Integer(12));
     Object result = runQuery(map, null);
-    if (result != null)
+    if (result != QueryService.UNDEFINED)
       fail();
   }
 
@@ -171,7 +171,7 @@ public class IndexOperatorJUnitTest {
     map.put("0", new Integer(11));
     map.put("1", new Integer(12));
     Object result = runQuery(map, QueryService.UNDEFINED);
-    if (result != null)
+    if (result != QueryService.UNDEFINED)
       fail();
   }
 
--- a/geode-core/src/integrationTest/java/org/apache/geode/cache/query/functional/IndexOperatorJUnitTest.java
+++ b/geode-core/src/integrationTest/java/org/apache/geode/cache/query/functional/IndexOperatorJUnitTest.java
@@ -19,6 +19,7 @@
  */
 package org.apache.geode.cache.query.functional;
 
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.Assert.fail;
 
 import java.util.ArrayList;
@@ -145,8 +146,7 @@ public class IndexOperatorJUnitTest {
     map.put("0", new Integer(11));
     map.put("1", new Integer(12));
     Object result = runQuery(map, null);
-    if (result != QueryService.UNDEFINED)
-      fail();
+    assertThat(result).isEqualTo(QueryService.UNDEFINED);
   }
 
   @Test
@@ -171,8 +171,7 @@ public class IndexOperatorJUnitTest {
     map.put("0", new Integer(11));
     map.put("1", new Integer(12));
     Object result = runQuery(map, QueryService.UNDEFINED);
-    if (result != QueryService.UNDEFINED)
-      fail();
+    assertThat(result).isEqualTo(QueryService.UNDEFINED);
   }
 
   @Test
--- a/geode-core/src/integrationTest/java/org/apache/geode/cache/query/internal/index/IndexStatisticsJUnitTest.java
+++ b/geode-core/src/integrationTest/java/org/apache/geode/cache/query/internal/index/IndexStatisticsJUnitTest.java
@@ -370,7 +370,7 @@ public class IndexStatisticsJUnitTest {
 
     assertEquals(0, keyIndexStats.getReadLockCount());
 
-    assertEquals(100, keyIndexStats.getTotalUses());
+    assertEquals(0, keyIndexStats.getTotalUses());
 
     for (int i = 0; i < 50; i++) {
       region.invalidate(Integer.toString(i));
--- a/geode-core/src/integrationTest/java/org/apache/geode/cache/query/internal/index/IndexUseJUnitTest.java
+++ b/geode-core/src/integrationTest/java/org/apache/geode/cache/query/internal/index/IndexUseJUnitTest.java
@@ -15,6 +15,7 @@
 package org.apache.geode.cache.query.internal.index;
 
 import static org.apache.geode.cache.Region.SEPARATOR;
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
@@ -353,7 +354,7 @@ public class IndexUseJUnitTest {
 
     Index i2 =
         qs.createIndex("Index2", IndexType.FUNCTIONAL, "objs.liist[0]", SEPARATOR + "testRgn objs");
-    assertTrue(i2 instanceof CompactMapRangeIndex);
+    assertThat(i2).isInstanceOf(CompactMapRangeIndex.class);
     CacheUtils.getLogger().info("Executing query: " + query);
     QueryObserverImpl observer = new QueryObserverImpl();
     QueryObserverHolder.setInstance(observer);
@@ -1196,7 +1197,7 @@ public class IndexUseJUnitTest {
     Index i5 = qs.createIndex("Index5", IndexType.FUNCTIONAL, "itr1.testFields['complex']",
         SEPARATOR + "testRgn itr1");
 
-    assertTrue(i1 instanceof CompactMapRangeIndex);
+    assertThat(i1).isInstanceOf(CompactMapRangeIndex.class);
 
     // Execute Queries with Indexes
     for (int i = 0; i < queries.length; i++) {
--- a/geode-core/src/integrationTest/java/org/apache/geode/cache/query/internal/index/IndexUseJUnitTest.java
+++ b/geode-core/src/integrationTest/java/org/apache/geode/cache/query/internal/index/IndexUseJUnitTest.java
@@ -342,7 +342,7 @@ public class IndexUseJUnitTest {
 
     evaluateMapTypeIndexUsage("objs.maap['key2']", SEPARATOR + "testRgn objs", queries,
         queriesIndexNotUsed,
-        CompactRangeIndex.class);
+        CompactMapRangeIndex.class);
 
     String query = "SELECT DISTINCT * FROM " + SEPARATOR + "testRgn itr1  WHERE itr1.liist[0] >= 2";
     SelectResults withoutIndex, withIndex;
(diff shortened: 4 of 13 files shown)


geode-core/src/integrationTest/java/org/apache/geode/cache/query/functional/INOperatorJUnitTest.java
```diff
@@ -225,11 +225,11 @@ public void testUNDEFINED() throws Exception {
 
     q = CacheUtils.getQueryService().newQuery(" UNDEFINED IN SET(UNDEFINED)");
     result = q.execute();
-    assertThat(result).isEqualTo(QueryService.UNDEFINED);
+    assertThat(result).isEqualTo(TRUE);
```
kirklund, line 228: Another syntax to consider: `assertThat(result).isTrue();`

geode-core/src/integrationTest/java/org/apache/geode/cache/query/functional/IndexOperatorJUnitTest.java
```diff
@@ -145,7 +145,7 @@ public void testWithNULL() throws Exception {
     map.put("0", new Integer(11));
     map.put("1", new Integer(12));
     Object result = runQuery(map, null);
-    if (result != null)
+    if (result != QueryService.UNDEFINED)
```
kirklund, line 148: I realize that you're editing code that isn't great but `fail();` is a terrible anti-pattern in testing.

Try this:
```
assertThat(result).isNotEqualTo(QueryService.UNDEFINED);
```

geode-core/src/integrationTest/java/org/apache/geode/cache/query/functional/IndexOperatorJUnitTest.java
```diff
@@ -171,7 +171,7 @@ public void testWithUNDEFINED() throws Exception {
     map.put("0", new Integer(11));
     map.put("1", new Integer(12));
     Object result = runQuery(map, QueryService.UNDEFINED);
-    if (result != null)
+    if (result != QueryService.UNDEFINED)
```
kirklund, line 174: `assertThat(result).isNotEqualTo(QueryService.UNDEFINED);`

geode-core/src/integrationTest/java/org/apache/geode/cache/query/internal/index/IndexUseJUnitTest.java
```diff
@@ -353,7 +353,7 @@ public void testCompactMapIndexUsageWithIndexOnSingleKey() throws Exception {
 
     Index i2 =
         qs.createIndex("Index2", IndexType.FUNCTIONAL, "objs.liist[0]", SEPARATOR + "testRgn objs");
-    assertTrue(i2 instanceof CompactRangeIndex);
+    assertTrue(i2 instanceof CompactMapRangeIndex);
```
kirklund, line 356: I highly recommend updating any test that you touch to use AssertJ instead of JUnit Assert. You can even install an IntelliJ plugin that'll automate most of the work for you.

JUnit assertions like this will produce a terrible failure message:
```
ava.lang.AssertionError
	at org.junit.Assert.fail(Assert.java:87)
	at org.junit.Assert.assertTrue(Assert.java:42)
	at org.junit.Assert.assertTrue(Assert.java:53)
	at org.apache.geode.IndexStatisticsJUnitTest.testCompactMapIndexUsageWithIndexOnSingleKey(IndexStatisticsJUnitTest.java:356)
```
Changing to AssertJ:
```
assertThat(i2).isInstanceOf(CompactMapRangeIndex.class);
```
The failure message will then be:
```
java.lang.AssertionError: 
Expecting:
  java.lang.Object@735b478
to be an instance of:
  java.util.Collection
but was instance of:
  java.lang.Object

	at org.apache.geode.IndexStatisticsJUnitTest.testCompactMapIndexUsageWithIndexOnSingleKey(IndexStatisticsJUnitTest.java:356)
```

geode-core/src/integrationTest/java/org/apache/geode/cache/query/internal/index/IndexUseJUnitTest.java
```diff
@@ -1196,7 +1196,7 @@ public void testCompactMapIndexUsageManyKeysWithVariousValueTypes() throws Excep
     Index i5 = qs.createIndex("Index5", IndexType.FUNCTIONAL, "itr1.testFields['complex']",
         SEPARATOR + "testRgn itr1");
 
-    assertTrue(i1 instanceof CompactRangeIndex);
+    assertTrue(i1 instanceof CompactMapRangeIndex);
```
kirklund, line 1199: AssertJ will work much better here.

geode-core/src/main/java/org/apache/geode/cache/query/internal/ResultsCollectionPdxDeserializerWrapper.java
```diff
@@ -218,4 +218,12 @@ public void setElementType(ObjectType elementType) {
...
   }
 
+  public String toString() {
+    String out = "size: " + size() + "\n";
+    Iterator iter = iterator();
+    while (iter.hasNext()) {
+      out += iter.next() + "\n";
+    }
+    return out;
+  }
```
kirklund, line 228: All instances of `\n` that you find should be changed to `System.lineSeparator()`.

geode-junit/src/main/java/org/apache/geode/cache/query/data/Portfolio.java
```diff
@@ -184,12 +184,14 @@ public int hashCode() {
...
-    while (iter.hasNext()) {
-      Map.Entry entry = (Map.Entry) iter.next();
-      out += entry.getKey() + ":" + entry.getValue() + ", ";
+    if (positions != null) {
+      Iterator iter = positions.entrySet().iterator();
+      while (iter.hasNext()) {
+        Map.Entry entry = (Map.Entry) iter.next();
+        out += entry.getKey() + ":" + entry.getValue() + ", ";
+      }
+      out += "\n P1:" + position1 + ", P2:" + position2;
```
kirklund, line 193: Please change all instances of `\n` to `System.lineSeparator()`.

geode-core/src/integrationTest/java/org/apache/geode/cache/query/internal/index/MapRangeIndexMaintenanceJUnitTest.java
```diff
@@ -341,6 +341,155 @@ public void testNullMapValuesInIndexOnLocalRegionForCompactMap() throws Exceptio
...
+        .execute();
+    System.out.println("Query: " + query + ", result: " + result);
+    assertEquals(6, result.size());
+
+    query = "select * from " + SEPARATOR + "portfolio p";
+    result = (SelectResults) qs
+        .newQuery(query)
+        .execute();
+    System.out.println("Query: " + query + ", result: " + result);
+    assertEquals(7, result.size());
```
kirklund, line 490: AssertJ will print out the entire result and why the assertion failed if you change to:
```
assertThat(result).hasSize(7);
```

kirklund (changes requested):
Overall, switching from JUnit Assert to AssertJ will provide much better failure messages that will really help anyone dealing with test failures.

`\n` should always be replaced with `System.lineSeparator()`.

geode-core/src/integrationTest/java/org/apache/geode/cache/query/internal/index/MapRangeIndexMaintenanceJUnitTest.java
```diff
@@ -341,6 +341,155 @@ public void testNullMapValuesInIndexOnLocalRegionForCompactMap() throws Exceptio
...
+    long mapIndexKeys =
+        ((CompactMapRangeIndex) keyIndex1).internalIndexStats.getNumberOfMapIndexKeys();
+    long values =
+        ((CompactMapRangeIndex) keyIndex1).internalIndexStats.getNumberOfValues();
+    assertEquals(5, keys);
+    assertEquals(4, mapIndexKeys);
+    assertEquals(5, values);
+  }
+
+  public void testQueriesForValueInMapField(Region region, QueryService qs) throws Exception {
```
DonalEvans, line 413: The compiler warning here can be cleaned up by using `Region<Object, Object>` here.

geode-core/src/integrationTest/java/org/apache/geode/cache/query/internal/index/MapRangeIndexMaintenanceJUnitTest.java
```diff
@@ -341,6 +341,155 @@ public void testNullMapValuesInIndexOnLocalRegionForCompactMap() throws Exceptio
...
+    // Empty map
+    Portfolio p = new Portfolio(1, 1);
+    p.positions = new HashMap();
```
DonalEvans, line 416: Compiler warnings can be cleaned up by using `HashMap<>()` throughout this method.

geode-core/src/integrationTest/java/org/apache/geode/cache/query/internal/index/MapRangeIndexMaintenanceJUnitTest.java
```diff
@@ -341,6 +341,155 @@ public void testNullMapValuesInIndexOnLocalRegionForCompactMap() throws Exceptio
...
+    region.put(6, p6);
+
+    // One more with null map
+    Portfolio p7 = new Portfolio(7, 7);
+    p7.positions = null;
+    region.put(7, p7);
+
+    String query;
+    query = "select * from " + SEPARATOR + "portfolio p where p.positions['SUN'] = null";
+    SelectResults result = (SelectResults) qs
```
DonalEvans, line 458: Compiler warnings here and elsewhere in this method can be fixed by using `SelectResults<Object>` and `UncheckedUtils.uncheckedCast(qs.newQuery(query).execute());`

geode-core/src/integrationTest/java/org/apache/geode/cache/query/internal/index/MapRangeIndexMaintenanceJUnitTest.java
```diff
@@ -341,6 +341,155 @@ public void testNullMapValuesInIndexOnLocalRegionForCompactMap() throws Exceptio
...
+        .newQuery(query)
+        .execute();
+    System.out.println("Query: " + query + ", result: " + result);
```
DonalEvans, line 461: If output is strictly necessary in this method, it should be via a logger, not `System.out`. Instead, however, it might be better to assert on the query result to confirm that it matches what is expected beyond just having the correct size, so that any failure will be self explanatory and not require this debug logging.

DonalEvans (changes requested):
Some code clean up requests.

albertogpz:
> Overall, switching from JUnit Assert to AssertJ will provide much better failure messages that will really help anyone dealing with test failures.
> 
> `\n` should always be replaced with `System.lineSeparator()`.

Thanks for your comments, Kirk. I thought of changing completely the files I edited to AssertJ but then I realized it would be better just to change to AssertJ my changes and leave the cleaning to another PR so that the core of this PR is not hidden by the other changes.

DonalEvans:
The StressNewTest failures you're seeing in pre-checkin might be fixed if you rebase your branch on develop. They look very similar to the failures that were being caused by GEODE-8905, which has since been reverted.

geode-core/src/integrationTest/java/org/apache/geode/cache/query/internal/index/MapRangeIndexMaintenanceJUnitTest.java
DonalEvans: It looks like this use of the raw class instead of the parameterized one got missed in the last set of changes.

geode-core/src/integrationTest/java/org/apache/geode/cache/query/functional/INOperatorJUnitTest.java
mhansonp: +1 for consistency.

geode-core/src/integrationTest/java/org/apache/geode/cache/query/internal/index/IndexStatisticsJUnitTest.java
```diff
@@ -370,7 +370,7 @@ public void testStatsForCompactMapRangeIndex() throws Exception {
 
     assertEquals(0, keyIndexStats.getReadLockCount());
 
-    assertEquals(100, keyIndexStats.getTotalUses());
```
mhansonp, line 373: Why did this change with no other code changes in the test method?

albertogpz:
Opened a new pull request with some changes to the original solution but taking into account all the comments already received.

PR #6100, opened 2021-03-06
GEODE-9000: Added logic to avoid NPE during processing Network Partition

echobravopapa:

### For all changes:

### Note:

### DIFF
--- a/geode-membership/src/integrationTest/java/org/apache/geode/distributed/internal/membership/gms/membership/GMSJoinLeaveJUnitTest.java
+++ b/geode-membership/src/integrationTest/java/org/apache/geode/distributed/internal/membership/gms/membership/GMSJoinLeaveJUnitTest.java
@@ -16,6 +16,7 @@ package org.apache.geode.distributed.internal.membership.gms.membership;
 
 import static org.apache.geode.test.awaitility.GeodeAwaitility.await;
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatNoException;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 import static org.hamcrest.core.IsEqual.equalTo;
 import static org.junit.Assert.assertEquals;
@@ -859,7 +860,7 @@ public class GMSJoinLeaveJUnitTest {
 
     // this creates a NPE that must be handled appropriately
     gmsJoinLeave.currentView = null;
-    gmsJoinLeave.processNetworkPartitionMessage(message);
+    assertThatNoException().isThrownBy(() -> gmsJoinLeave.processNetworkPartitionMessage(message));
 
   }
 
--- a/geode-membership/src/integrationTest/java/org/apache/geode/distributed/internal/membership/gms/membership/GMSJoinLeaveJUnitTest.java
+++ b/geode-membership/src/integrationTest/java/org/apache/geode/distributed/internal/membership/gms/membership/GMSJoinLeaveJUnitTest.java
@@ -849,6 +849,21 @@ public class GMSJoinLeaveJUnitTest {
   }
 
   @Test
+  public void testNetworkPartitionDetectedWithNullView() throws Exception {
+    initMocks(true);
+    prepareAndInstallView(mockMembers[0], createMemberList(mockMembers[0], gmsJoinLeaveMemberId));
+
+    NetworkPartitionMessage message = new NetworkPartitionMessage();
+    message.setSender(gmsJoinLeaveMemberId);
+
+
+    // this creates a NPE that must be handled appropriately
+    gmsJoinLeave.currentView = null;
+    gmsJoinLeave.processNetworkPartitionMessage(message);
+
+  }
+
+  @Test
   public void testNetworkPartitionDetected() throws Exception {
     initMocks(true);
     prepareAndInstallView(mockMembers[0], createMemberList(mockMembers[0], gmsJoinLeaveMemberId));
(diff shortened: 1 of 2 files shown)


geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/membership/GMSJoinLeave.java
```diff
@@ -1456,12 +1456,18 @@ void processNetworkPartitionMessage(NetworkPartitionMessage<ID> msg) {
...
+        String str = "Membership coordinator " + msg.getSender()
+            + " has declared that a network partition has occurred";
+        forceDisconnect(str);
+      } else {
+        logger.warn("Ignoring the network partition message from a non-member: " + msg.getSender());
+      }
     } else {
-      logger.warn("Ignoring the network partition message from a non-member: " + msg.getSender());
+      logger.info(
+          "Ignoring, likely this message was intended for the previous Membership service... ");
```
jujoramos, line 1470: Not sure whether this should be logged at `info` level, even if users see it, they won't know what to do about it. If the `else` branch it's harmless in itself, I'd say we should just use `debug` level instead.

jujoramos (commented):
Can we add some tests to make sure the fix works as expected and, also, to prevent somebody else from re-introducing it in the future by accident?.
Otherwise, looks good to me.

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/membership/GMSJoinLeave.java
echobravopapa: wilco

jujoramos (changes requested):
Can we add some tests to make sure the fix works as expected and to prevent somebody else from re-introducing it in the future by accident?.

echobravopapa:
@jujoramos I added a unit test

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/membership/GMSJoinLeave.java
```diff
@@ -1456,12 +1456,18 @@ void processNetworkPartitionMessage(NetworkPartitionMessage<ID> msg) {
       return;
     }
     ID sender = msg.getSender();
-    if (getView().getMembers().contains(sender)) {
-      String str = "Membership coordinator " + msg.getSender()
-          + " has declared that a network partition has occurred";
-      forceDisconnect(str);
+
+    if (getView() != null && isJoined) {
+      if (getView().getMembers().contains(sender)) {
```
Bill, line 1461: It appears to me that the only way `currentView` can be `null` is before the first call to `installView(GMSMembershipView<ID>)`. It is TBD if that method is ever called with a `null` reference. If we could ensure that it was never called with a `null` reference then the null-check in the current PR might be ok, since that would mean `currentView` goes from `null` to non-null and then never goes to `null` again.

I have done some analysis on `installView(GMSMembershipView<ID>)` by adding the `@NotNull` annotation to `Service.installView(@NotNull GMSMembershipView<ID> v)` and then looking for violations found by IntelliJ inspections. Of the 6 usages outside tests, IntelliJ's inspections doesn't flag any of them as violations.

Since I'm so new to using this inspection, I don't quite trust it (or my interpretation of it.) I wonder if it might be a good idea to make the code here a little more defensive by capturing the `currentView` here, and then operating on that. Like:

```
    final GMSMembershipView<ID> view = getView();
    if (view != null && isJoined) {
      if (view.getMembers().contains(sender)) {
```

That'd ensure that any changes to `currentView` between the two calls to `getView()` don't bite us.

Bill (commented):
I'd like your take on my comment below.

geode-membership/src/integrationTest/java/org/apache/geode/distributed/internal/membership/gms/membership/GMSJoinLeaveJUnitTest.java
```diff
@@ -848,6 +848,21 @@ private InstallViewMessage getInstallViewMessage(GMSMembershipView view, Object
...
+    initMocks(true);
+    prepareAndInstallView(mockMembers[0], createMemberList(mockMembers[0], gmsJoinLeaveMemberId));
+
+    NetworkPartitionMessage message = new NetworkPartitionMessage();
+    message.setSender(gmsJoinLeaveMemberId);
+
+
+    // this creates a NPE that must be handled appropriately
+    gmsJoinLeave.currentView = null;
+    gmsJoinLeave.processNetworkPartitionMessage(message);
```
jujoramos, line 862: Nitpick: maybe use `assertThatCode(() -> gmsJoinLeave.processNetworkPartitionMessage(message)).doesNotThrowAnyException();` to make the test more explicit instead?.

jujoramos (approved):

bschuchardt (commented):
I suppose that one thing we could do to make GMSJoinLeave better is to get rid of the null checks entirely and have currentView's initial value be an empty view with an invalid (-1) view ID.

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/membership/GMSJoinLeave.java
```diff
@@ -1456,12 +1456,18 @@ void processNetworkPartitionMessage(NetworkPartitionMessage<ID> msg) {
...
+        String str = "Membership coordinator " + msg.getSender()
+            + " has declared that a network partition has occurred";
+        forceDisconnect(str);
+      } else {
+        logger.warn("Ignoring the network partition message from a non-member: " + msg.getSender());
+      }
     } else {
-      logger.warn("Ignoring the network partition message from a non-member: " + msg.getSender());
+      logger.debug(
+          "Ignoring, likely this message was intended for the previous Membership service... ");
```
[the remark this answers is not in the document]
bschuchardt: I'd like to see more detail about what's being ignored.  If I saw this message in a log I wouldn't know what it means.  How about logging the message itself?  That would show what it is and who sent it.

PR #6104, opened 2021-03-09
GEODE-9016: Fix the NPE for PutAll with CQ LOCAL_DESTROY message type

jchen21:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/DistributedPutAllOperation.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/DistributedPutAllOperation.java
@@ -819,7 +819,7 @@ public class DistributedPutAllOperation extends AbstractUpdateOperation {
   }
 
   @Override
-  protected void removeDestroyTokensFromCqResultKeys(FilterRoutingInfo filterRouting) {
+  void removeDestroyTokensFromCqResultKeys(FilterRoutingInfo filterRouting) {
     for (InternalDistributedMember m : filterRouting.getMembers()) {
       FilterInfo filterInfo = filterRouting.getFilterInfo(m);
       if (filterInfo.getCQs() == null) {
@@ -845,13 +845,10 @@ public class DistributedPutAllOperation extends AbstractUpdateOperation {
           for (int i = 0; i < this.putAllData.length; i++) {
             @Unretained
             EntryEventImpl entryEvent = getEventForPosition(i);
-            if (entryEvent != null) {
-              if (cq != null && cq.getFilterID() != null && cq.getFilterID().equals(cqID)
-                  && (e.getValue().equals(MessageType.LOCAL_DESTROY))
-                  && entryEvent.getKey() != null) {
-                cq.removeFromCqResultKeys(entryEvent.getKey(), true);
-              }
-
+            if (entryEvent != null && entryEvent.getKey() != null && cq != null
+                && cq.getFilterID() != null && cq.getFilterID().equals(cqID)
+                && (e.getValue().equals(MessageType.LOCAL_DESTROY))) {
+              cq.removeFromCqResultKeys(entryEvent.getKey(), true);
             }
           }
         }
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/DistributedPutAllOperation.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/DistributedPutAllOperation.java
@@ -47,7 +47,6 @@ import org.apache.geode.distributed.internal.membership.InternalDistributedMembe
 import org.apache.geode.internal.InternalDataSerializer;
 import org.apache.geode.internal.cache.FilterRoutingInfo.FilterInfo;
 import org.apache.geode.internal.cache.ha.ThreadIdentifier;
-import org.apache.geode.internal.cache.partitioned.Bucket;
 import org.apache.geode.internal.cache.partitioned.PutAllPRMessage;
 import org.apache.geode.internal.cache.tier.MessageType;
 import org.apache.geode.internal.cache.tier.sockets.ClientProxyMembershipID;
@@ -819,38 +818,18 @@ public class DistributedPutAllOperation extends AbstractUpdateOperation {
   }
 
   @Override
-  void removeDestroyTokensFromCqResultKeys(FilterRoutingInfo filterRouting) {
-    for (InternalDistributedMember m : filterRouting.getMembers()) {
-      FilterInfo filterInfo = filterRouting.getFilterInfo(m);
-      if (filterInfo.getCQs() == null) {
-        continue;
-      }
-
-      CacheDistributionAdvisor.CacheProfile cf =
-          (CacheDistributionAdvisor.CacheProfile) ((Bucket) getRegion()).getPartitionedRegion()
-              .getCacheDistributionAdvisor().getProfile(m);
-
-      if (cf == null || cf.filterProfile == null || cf.filterProfile.isLocalProfile()
-          || cf.filterProfile.getCqMap().isEmpty()) {
-        continue;
-      }
-
-      for (Object value : cf.filterProfile.getCqMap().values()) {
-        ServerCQ cq = (ServerCQ) value;
-
-        for (Map.Entry<Long, Integer> e : filterInfo.getCQs().entrySet()) {
-          Long cqID = e.getKey();
-          // For the CQs satisfying the event with destroy CQEvent, remove
-          // the entry from CQ cache.
-          for (int i = 0; i < this.putAllData.length; i++) {
-            @Unretained
-            EntryEventImpl entryEvent = getEventForPosition(i);
-            if (entryEvent != null && entryEvent.getKey() != null && cq != null
-                && cq.getFilterID() != null && cq.getFilterID().equals(cqID)
-                && (e.getValue().equals(MessageType.LOCAL_DESTROY))) {
-              cq.removeFromCqResultKeys(entryEvent.getKey(), true);
-            }
-          }
+  void doRemoveDestroyTokensFromCqResultKeys(FilterInfo filterInfo, ServerCQ cq) {
+    for (Map.Entry<Long, Integer> e : filterInfo.getCQs().entrySet()) {
+      Long cqID = e.getKey();
+      // For the CQs satisfying the event with destroy CQEvent, remove
+      // the entry from CQ cache.
+      for (int i = 0; i < this.putAllData.length; i++) {
+        @Unretained
+        EntryEventImpl entryEvent = getEventForPosition(i);
+        if (entryEvent != null && entryEvent.getKey() != null && cq != null
+            && cq.getFilterID() != null && cq.getFilterID().equals(cqID)
+            && e.getValue() != null && e.getValue().equals(MessageType.LOCAL_DESTROY)) {
+          cq.removeFromCqResultKeys(entryEvent.getKey(), true);
         }
       }
     }
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/DistributedPutAllOperation.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/DistributedPutAllOperation.java
@@ -25,6 +25,7 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 
 import it.unimi.dsi.fastutil.objects.Object2IntOpenHashMap;
@@ -39,13 +40,16 @@ import org.apache.geode.cache.RegionAttributes;
 import org.apache.geode.cache.Scope;
 import org.apache.geode.cache.persistence.PersistentReplicatesOfflineException;
 import org.apache.geode.cache.query.internal.cq.CqService;
+import org.apache.geode.cache.query.internal.cq.ServerCQ;
 import org.apache.geode.distributed.DistributedMember;
 import org.apache.geode.distributed.internal.DirectReplyProcessor;
 import org.apache.geode.distributed.internal.membership.InternalDistributedMember;
 import org.apache.geode.internal.InternalDataSerializer;
 import org.apache.geode.internal.cache.FilterRoutingInfo.FilterInfo;
 import org.apache.geode.internal.cache.ha.ThreadIdentifier;
+import org.apache.geode.internal.cache.partitioned.Bucket;
 import org.apache.geode.internal.cache.partitioned.PutAllPRMessage;
+import org.apache.geode.internal.cache.tier.MessageType;
 import org.apache.geode.internal.cache.tier.sockets.ClientProxyMembershipID;
 import org.apache.geode.internal.cache.tier.sockets.VersionedObjectList;
 import org.apache.geode.internal.cache.tx.RemotePutAllMessage;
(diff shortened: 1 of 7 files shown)


geode-core/src/main/java/org/apache/geode/internal/cache/DistributedPutAllOperation.java
```diff
@@ -814,6 +818,46 @@ protected FilterRoutingInfo getRecipientFilterRouting(Set cacheOpRecipients) {
...
+      CacheDistributionAdvisor.CacheProfile cf =
+          (CacheDistributionAdvisor.CacheProfile) ((Bucket) getRegion()).getPartitionedRegion()
+              .getCacheDistributionAdvisor().getProfile(m);
+
+      if (cf == null || cf.filterProfile == null || cf.filterProfile.isLocalProfile()
+          || cf.filterProfile.getCqMap().isEmpty()) {
+        continue;
+      }
+
+      for (Object value : cf.filterProfile.getCqMap().values()) {
```
kirklund, line 838: In the future, you might want to try extracting some of those inner-loops or inner-if-blocks to their own methods. How you name those methods can then really improve understanding. It's also too easy to get lost in those deeply nested structures like this.

geode-cq/src/distributedTest/java/org/apache/geode/cache/query/cq/dunit/PartitionedRegionCqQueryDUnitTest.java
```diff
@@ -120,6 +123,60 @@
...
+      }
+      region.putAll(buffer);
+    });
+
+    client.invoke(() -> {
+      QueryService cqService = getCache().getQueryService();
+      CqQuery cqQuery = cqService.getCq(cqName);
+      if (cqQuery == null) {
+        fail("Failed to get CQ " + cqName);
+      }
```
kirklund, line 157: Please use assertions instead of if-fail patterns:
```
assertThat(cqQuery)
    .withFailMessage("Failed to get CQ " + cqName)
    .isNotNull();
```

geode-cq/src/distributedTest/java/org/apache/geode/cache/query/cq/dunit/PartitionedRegionCqQueryDUnitTest.java
```diff
@@ -120,6 +123,60 @@
...
+      try {
+        cqQuery.executeWithInitialResults();
+      } catch (Exception ex) {
+        fail("Failed to execute  CQ " + cqName, ex);
+      }
```
kirklund, line 162: You don't need to catch Exception and invoke fail. It's better to let `VM.invoke(...)` throw the Exception wrapped in an RMIException. You already have `throws Exception` on the test method so JUnit will handle it and display the full stack perfectly:
```
cqQuery.executeWithInitialResults();
```

geode-cq/src/distributedTest/java/org/apache/geode/cache/query/cq/dunit/PartitionedRegionCqQueryDUnitTest.java
```diff
@@ -120,6 +123,60 @@
...
+      for (int i = 1; i < numObjects; i++) {
+        Portfolio p = new Portfolio(-1 * i);
+        buffer.put("" + i, p);
+      }
+      region.putAll(buffer);
+    });
+
+    cqHelper.closeClient(client);
+    cqHelper.closeServer(server2);
+    cqHelper.closeServer(server1);
```
kirklund, line 177: It's better (and more industry standard for writing JUnit tests) if you can figure out a way to move cleanup code at the end of the test method to a `tearDown` method annotated with `@After`. For example, if you make `cqHelper.closeClient` return without error if client is null or not running, then you could hoist these local variables to fields and the test would look more like this:
```
private VM client;
private VM server1;
private VM server2;

@Before 
public void setUp() {
 server1 = getVM(0);
 server2 = getVM(1);
 client = getVM(2);
}

@After
public void tearDown() {
  cqHelper.closeClient(client);
  cqHelper.closeServer(server2);
  cqHelper.closeServer(server1);
}
```
And `cqHelper` close methods could have some early-outs like this:
```
public void closeClient(VM vm) {
  if (vm == null) {
    return;
  }
  // etc
}
```

geode-cq/src/distributedTest/java/org/apache/geode/cache/query/cq/dunit/PartitionedRegionCqQueryDUnitTest.java
```diff
@@ -120,6 +123,60 @@
...
+    final String host = Host.getHost(0).getHostName();
+    final int port = server2.invoke(() -> PartitionedRegionCqQueryDUnitTest.getCacheServerPort());
+    createClient(client, port, host);
+    createCQ(client, cqName, cqs[0]);
+
+    int numObjects = 1000;
+
+    server1.invoke(() -> {
+      Region region = getCache().getRegion(SEPARATOR + "root" + SEPARATOR + regions[0]);
+      Map buffer = new HashMap();
```
kirklund, line 144: You should avoid raw types in both product and test code. For example: `Region<String, Portfolio>` or even `Region<String, Object>`. Same with Map.

geode-cq/src/distributedTest/java/org/apache/geode/cache/query/cq/dunit/PartitionedRegionCqQueryDUnitTest.java
```diff
@@ -120,6 +123,60 @@
...
+  @Test
+  public void testPutAllWithCQLocalDestroy() throws Exception {
+    VM server1 = getVM(0);
+    VM server2 = getVM(1);
+    VM client = getVM(2);
+
+    final String cqName = "testPutAllWithCQLocalDestroy_0";
+    createServer(server1);
+    createServer(server2);
+    final String host = Host.getHost(0).getHostName();
```
kirklund, line 135: Avoid `Host` and just use:
```
String hostName = VM.getHostName();
```

geode-cq/src/distributedTest/java/org/apache/geode/cache/query/cq/dunit/PartitionedRegionCqQueryDUnitTest.java
```diff
@@ -120,6 +123,60 @@
...
+    final int port = server2.invoke(() -> PartitionedRegionCqQueryDUnitTest.getCacheServerPort());
```
kirklund, line 136: Remove the name of the test anytime you see this, so you just have:
```
int port = server2.invoke(() -> getCacheServerPort());
```

kirklund (changes requested):
I marked some areas in the dunit test that could be improved. My only other recommendation is to please figure out a way to add test(s) to the existing (but mostly empty) DistributedPutAllOperationTest unit test using mocks for at least the changes involving `removeDestroyTokensFromCqResultKeys`.

geode-cq/src/distributedTest/java/org/apache/geode/cache/query/cq/dunit/PartitionedRegionCqQueryDUnitTest.java
kirklund: PS: You don't really need `SEPARATOR + "root" + SEPARATOR`. This will also work:
```
getCache().getRegion(regions[0]);
```

geode-core/src/main/java/org/apache/geode/internal/cache/DistributedPutAllOperation.java
```diff
@@ -814,6 +818,46 @@ protected FilterRoutingInfo getRecipientFilterRouting(Set cacheOpRecipients) {
     return consolidated;
   }
 
+  @Override
+  protected void removeDestroyTokensFromCqResultKeys(FilterRoutingInfo filterRouting) {
```
agingade, line 822: The same may need to be done in RemoveAll.

geode-cq/src/distributedTest/java/org/apache/geode/cache/query/cq/dunit/PartitionedRegionCqQueryDUnitTest.java
```diff
@@ -120,6 +123,60 @@
 
   private static int bridgeServerPort;
 
+  @Test
+  public void testPutAllWithCQLocalDestroy() throws Exception {
```
agingade, line 127: Test needs to be added for removeAll.

geode-cq/src/distributedTest/java/org/apache/geode/cache/query/cq/dunit/PartitionedRegionCqQueryDUnitTest.java
jchen21: Actually, for this test, we still need `SEPARATOR + "root" + SEPARATOR`, otherwise `getRegion()` can't find the region.

geode-cq/src/distributedTest/java/org/apache/geode/cache/query/cq/dunit/PartitionedRegionCqQueryDUnitTest.java
jchen21: I would love the clean the test code as you suggested. This test also uses a lot of already deprecated test classes and APIs. It's better to refactor it. However, introducing @Before and @After would affect all 14 test methods in the file, it would be better to have a separate JIRA to track the refactoring. I have created https://issues.apache.org/jira/browse/GEODE-9020 to track it.

kirklund: Feel free to fit into the test in the best way you can. Another alternative is to extract the new test out to a new test class (now in a followup PR). You could also modernize the whole test but that would be best done as a followup PR.

kirklund (approved):

geode-core/src/test/java/org/apache/geode/internal/cache/DistributedPutAllOperationTest.java
```diff
@@ -33,4 +47,39 @@ public void shouldBeMockable() throws Exception {
 
     assertThat(mockDistributedPutAllOperation.getBaseEvent()).isSameAs(mockEntryEventImpl);
   }
+
+  @Test
+  public void testDoRemoveDestroyTokensFromCqResultKeys() {
```
agingade, line 52: A small refactoring request...to make it readable, can this be broken down (with new lines) into:
setup
action
validation/verification

geode-cq/src/distributedTest/java/org/apache/geode/cache/query/cq/dunit/PartitionedRegionCqQueryDUnitTest.java
```diff
@@ -120,6 +125,109 @@
...
+
+    final String cqName = "testPutAllWithCQLocalDestroy_0";
+    createServer(server1);
+    createServer(server2);
+    final String host = VM.getHostName();
+    final int port = server2.invoke(() -> getCacheServerPort());
+    createClient(client, port, host);
+    createCQ(client, cqName, cqs[0]);
+
+    int numObjects = 1000;
```
agingade, line 142: Does this need to be that big...If its not needed can we keep it small, say 5 or 10. Makes test to run faster.

agingade (approved):
Approving with the comments taken care.

geode-cq/src/distributedTest/java/org/apache/geode/cache/query/cq/dunit/PartitionedRegionCqQueryDUnitTest.java
jchen21: Yes. It does need a larger number. Small number sometimes does not consistently reproduce the bug.

DonalEvans (approved):

geode-core/src/main/java/org/apache/geode/internal/cache/DistributedPutAllOperation.java
```diff
@@ -814,6 +817,23 @@ protected FilterRoutingInfo getRecipientFilterRouting(Set cacheOpRecipients) {
...
+        @Unretained
+        EntryEventImpl entryEvent = getEventForPosition(i);
+        if (entryEvent != null && entryEvent.getKey() != null && cq != null
+            && cq.getFilterID() != null && cq.getFilterID().equals(cqID)
+            && e.getValue() != null && e.getValue().equals(MessageType.LOCAL_DESTROY)) {
+          cq.removeFromCqResultKeys(entryEvent.getKey(), true);
+        }
+      }
+    }
+  }
```
nabarunnag, line 836:
```suggestion
  void doRemoveDestroyTokensFromCqResultKeys(FilterInfo filterInfo, ServerCQ cq, int length) {
    for (Map.Entry<Long, Integer> e : filterInfo.getCQs().entrySet()) {
      Long cqID = e.getKey();
      // For the CQs satisfying the event with destroy CQEvent, remove
      // the entry from CQ cache.
      for (int i = 0; i < length; i++) {
        @Unretained
        EntryEventImpl entryEvent = getEventForPosition(i);
        if (entryEvent != null && entryEvent.getKey() != null && cq != null
            && cq.getFilterID() != null && cq.getFilterID().equals(cqID)
            && e.getValue() != null && e.getValue().equals(MessageType.LOCAL_DESTROY)) {
          cq.removeFromCqResultKeys(entryEvent.getKey(), true);
        }
      }
    }
  }
```

nabarunnag: Maybe this method can be put in a common class and reused in DistributedRemoveAllOperation and DistributedPutAllOperations

geode-cq/src/distributedTest/java/org/apache/geode/cache/query/cq/dunit/PartitionedRegionCqQueryDUnitTest.java
```diff
@@ -120,6 +125,111 @@
...
+          getCache().getRegion(SEPARATOR + "root" + SEPARATOR + regions[0]);
+      // PutAll with entries that do not satisfy CQ. This is to generate LOCAL_DESTROY CQ event
+      Map<String, Object> buffer = new HashMap();
+      for (int i = 1; i < numObjects; i++) {
+        Portfolio p = new Portfolio(-1 * i);
+        buffer.put("" + i, p);
+      }
+      region.putAll(buffer);
+    });
+
```
nabarunnag, line 175: Should there be an assert at the end of the test to verify the behavior ?

geode-cq/src/distributedTest/java/org/apache/geode/cache/query/cq/dunit/PartitionedRegionCqQueryDUnitTest.java
```diff
@@ -120,6 +125,111 @@
...
+      Region<String, Object> region =
+          getCache().getRegion(SEPARATOR + "root" + SEPARATOR + regions[0]);
+      Set<String> keys = new HashSet<>();
+      for (int i = 1; i < numObjects; i++) {
+        keys.add("" + i);
+      }
+      // This is to generate LOCAL_DESTROY CQ event
+      region.removeAll(keys);
+    });
+
```
nabarunnag, line 227: Should there be an assert at the end of the test to verify the behavior ?

nabarunnag (changes requested):

geode-cq/src/distributedTest/java/org/apache/geode/cache/query/cq/dunit/PartitionedRegionCqQueryDUnitTest.java
agingade: Can you please help me understand why consistency an issue here; in that case how sure we are that 1000 entries will recreate the issue all the time.

geode-core/src/main/java/org/apache/geode/internal/cache/DistributedPutAllOperation.java
jchen21: I don't think it is necessary to pass the length of `putAllData` as a parameter. `putAllData` is a field of `DistributedPutAllOperation`. We can easily get its length without getting it from a parameter. Maybe I am missing something you proposed.

For code refactoring with a common class of `DistributedRemoveAllOperation` and `DistributedPutAllOperation`, if necessary, it would be better to do it in another pull request. This pull request focuses on fixing the bug.

nabarunnag (approved):
LGTM

PR #6106, opened 2021-03-09
GEODE-9017: Reload key and trust store upon change

aaronlindsey:
Reload the KeyManager and TrustManager in the SSLContext when the corresponding key store files are modified on disk.

See [the RFC](https://cwiki.apache.org/confluence/display/GEODE/Make+key+and+trust+stores+reload+automatically+upon+change) for more background on this change.

One thing I am not sure about is whether this should be the default behavior. I currently have the file watching key and trust managers enabled when the user does not set the "use default SSL context" flag (which means they are enabled under Geode's default settings). However, I could see how changing the default behavior might be surprising to some users. I could instead add a flag which enables the "automatic reload" behavior and make it disabled by default if the reviewers think that would be better.


aaronlindsey:
Hmm `./gradlew build` works but the concourse build job is failing. I will take a look...

geode-assembly/src/acceptanceTest/java/org/apache/geode/ssl/CertificateRotationTest.java
```diff
@@ -0,0 +1,427 @@
...
+
+    for (String name : memberNames) {
+      await().untilAsserted(() -> assertThat(logsForMember(name))
+          .as("The cluster's trust manager has been updated twice")
+          .haveExactly(2, linesMatching(updatedTrustManager)));
+    }
+
+    await().untilAsserted(() -> assertThat(logsForClient())
+        .as("The client's trust manager has been updated twice")
+        .haveExactly(2, linesMatching(updatedTrustManager)));
```
kirklund, line 221: Fancy :)

geode-core/src/main/java/org/apache/geode/internal/net/filewatch/FileWatchingX509ExtendedKeyManager.java
```diff
@@ -0,0 +1,137 @@
...
+    return keyManager.get().getServerAliases(s, principals);
+  }
+
+  private void loadKeyManager() {
+    KeyManager[] keyManagers;
+    try {
+      keyManagers = keyManagerSupplier.get();
+    } catch (Throwable t) {
+      throw new RuntimeException("Unable to load KeyManager", t);
+    }
```
kirklund, line 122: Usually we try to avoid catching VirtualMachineErrors and its subclasses. For example if OutOfMemoryError is thrown and caught here, there's actually no guarantee it will be thrown again/elsewhere even if the JVM is OOM.

It's usually more appropriate to use InternalGemFireException or GemFireSecurityException (both are RuntimeExceptions):
```
throw new InternalGemFireException("Unable to load KeyManager", t);
```

kirklund (approved):
Looks good overall. I recommend changing the try-catch-throw-RuntimeException block as described.

There's a `LogFileAssert` that I wrote (it's implemented with several files). You could try extending that to be more generic or more useful for your testing. Not requesting it just mentioning it.

geode-core/src/main/java/org/apache/geode/internal/net/filewatch/FileWatchingX509ExtendedKeyManager.java
aaronlindsey: Thanks! I will change this.

aaronlindsey:
Looking at the PR checks, there is a timing issue with my integration tests I will need to fix. Also, PMD flagged the static ConcurrentHashMaps I was using to prevent multiple file watches from being created for the same path. So I need to figure out a different way to do this without static mutable state.

onichols-pivotal:
> PMD flagged the static ConcurrentHashMaps I was using to prevent multiple file watches from being created for the same path. So I need to figure out a different way to do this without static mutable state.

PMD is not particularly smart...don't change perfectly good code _just_ to play games with PMD.  I believe you can annotate it with @Immutable to suppress the PMD message.

bschuchardt (approved):
looks good

aaronlindsey:
The latest commit makes a subtle change to `SSLUtil.createAndConfigureSSLContext`. It removes some code duplication by adding a single implementation for loading the key/trust managers instead of one implementation in `SocketCreator` and another implementation in `SSLUtil`. The SocketCreator implementation is chosen because it handles passwords provided by env var, encrypted passwords, and having the key/trust store file located in the user's home directory. However, this means that now SSLUtil will behave the same way as SocketCreator (whereas previously SSLUtil did not handle env var passwords, encrypted passwords, or home directory key/trust store file).

I'm requesting a re-review from codeowners of the relevant files since it was a non-trivial change.

PR #6117, opened 2021-03-11
GEODE-9023: Add sharding support to redis region

jdeppe-pivotal:
This ability allows for the future mapping of cluster slots to geode
buckets which is preemptive to enabling redis clustering support.

- Bucket count defaults to 128 and must be a a power of 2 and a factor
  of 16384 (the number of redis slots).
- Supports using hashtags in the key.

### For all changes:

### Note:


jdeppe-pivotal:
I have another PR, based on this one, which introduces the `RedisKey` object. That PR will refactor some of the routingId calculation.

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/cluster/CRC16.java
```diff
@@ -0,0 +1,65 @@
...
+package org.apache.geode.redis.internal.executor.cluster;
+
+public class CRC16 {
+
+  // CCITT/SDLC/HDLC x^16 + x^12 + x^5 + 1 (CRC-16-CCITT)
+  private static final int CCITT_POLY = 0x8408;
+  private static final short[] crcTable = new short[256];
+
+  // Create the table up front
+  static {
```
nabarunnag, line 25: This table will never change, will it be ok just to have a hardcoded table ?

nabarunnag (approved):
LGTM

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/cluster/CRC16.java
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  /**
+   * Calculate CRC with most significant byte first. Assume all inputs are valid.
+   *
+   * @param data the byte array to use
+   * @param start starting index into the byte array
+   * @param end ending index (exclusive) into the byte array
+   */
+  public static int calculate(byte[] data, int start, int end) {
```
nabarunnag, line 48: Still thinking, but will it be ok to keep the value as short instead of int ? To keep the algo similar to that at https://redis.io/topics/cluster-spec Appendix A

geode-redis/src/distributedTest/java/org/apache/geode/redis/internal/cluster/RedisPartitionResolverDUnitTest.java
```diff
@@ -0,0 +1,119 @@
...
+  private static final String LOCAL_HOST = "127.0.0.1";
+  private static final int JEDIS_TIMEOUT =
+      Math.toIntExact(GeodeAwaitility.getTimeout().toMillis());
+  private static Jedis jedis1;
+
+  private static MemberVM server1;
+  private static MemberVM server2;
+  private static MemberVM server3;
+
+  private static int redisServerPort1;
```
DonalEvans, line 56: This can be converted to a local variable in `classSetup()` since it's not used outside of that method.

geode-redis/src/distributedTest/java/org/apache/geode/redis/internal/cluster/RedisPartitionResolverDUnitTest.java
```diff
@@ -0,0 +1,119 @@
...
+  private static int redisServerPort2;
+  private static int redisServerPort3;
```
DonalEvans, line 58: These two fields are never used, and can be safely removed, along with the calls to `cluster.getRedisPort()` initializing them.

geode-redis/src/main/java/org/apache/geode/redis/internal/data/ByteArrayWrapper.java
```diff
@@ -107,6 +115,39 @@ public int compareTo(ByteArrayWrapper other) {
...
+  public synchronized Object getRoutingId() {
+    if (routingId == null && value != null) {
+      int startHashtag = Integer.MAX_VALUE;
+      int endHashtag = 0;
+
+      for (int i = 0; i < value.length; i++) {
+        if (value[i] == '{' && startHashtag == Integer.MAX_VALUE) {
+          startHashtag = i;
+        }
+        if (value[i] == '}') {
```
DonalEvans, line 133: This could be an `else if` since it's not possible for `value[i]` to be equal to both '{' and '}'.

DonalEvans (changes requested):
Just a couple of clean-up/simplifiaction requests

DonalEvans:
StressNewTest failures can probably be fixed by rebasing your branch on develop.

geode-redis/src/main/java/org/apache/geode/redis/internal/RegionProvider.java
```diff
@@ -14,19 +14,24 @@
...
 public class RegionProvider {
   /**
    * The name of the region that holds data stored in redis.
    */
-  private static final String REDIS_DATA_REGION = "__REDIS_DATA";
-  private static final String REDIS_CONFIG_REGION = "__REDIS_CONFIG";
+  public static final String REDIS_DATA_REGION = "__REDIS_DATA";
+  public static final String REDIS_CONFIG_REGION = "__REDIS_CONFIG";
+  public static final int REDIS_REGION_BUCKETS = Integer.getInteger("redis.region.buckets", 128);
+  public static final int REDIS_SLOTS = Integer.getInteger("redis.slots", 16384);
```
sabbey37, line 33: I'm probably missing something obvious, but I was wondering why we try to get `redis.slots` via System.getProperty... will we allow it to be configured at some point? The Redis cluster docs seem to say there are always `16384` slots.

geode-redis/src/main/java/org/apache/geode/redis/internal/RegionProvider.java
```diff
@@ -14,19 +14,24 @@
...
 
 public class RegionProvider {
   /**
    * The name of the region that holds data stored in redis.
    */
-  private static final String REDIS_DATA_REGION = "__REDIS_DATA";
-  private static final String REDIS_CONFIG_REGION = "__REDIS_CONFIG";
+  public static final String REDIS_DATA_REGION = "__REDIS_DATA";
+  public static final String REDIS_CONFIG_REGION = "__REDIS_CONFIG";
+  public static final int REDIS_REGION_BUCKETS = Integer.getInteger("redis.region.buckets", 128);
```
sabbey37, line 32: Do we need to have some check to ensure that if `redis.region.buckets` is set via a system property that it is a power of 2 and a factor of 16384? (maybe that's coming in the next PR?)

sabbey37 (approved):
Just from reading over the Redis cluster docs, this seems reasonable.

geode-redis/src/main/java/org/apache/geode/redis/internal/data/ByteArrayWrapper.java
```diff
@@ -107,6 +115,39 @@ public int compareTo(ByteArrayWrapper other) {
...
+          break;
+        }
+      }
+
+      if (endHashtag - startHashtag <= 1) {
+        startHashtag = -1;
+        endHashtag = value.length;
+      }
+
+      routingId = (CRC16.calculate(value, startHashtag + 1, endHashtag) % REDIS_SLOTS)
```
pivotal-jbarrett, line 144: I think we are ok right now for bucket size 128 but anything larger and the `Integer` cache won't kick in here. We should consider a local cache of `Integer` routing ids and ignore any built in `Integer` cache.

pivotal-jbarrett (changes requested):

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/cluster/CRC16.java
jdeppe-pivotal: Since it's just statically generated, I think it's OK to leave as is.

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/cluster/CRC16.java
jdeppe-pivotal: Will do.

geode-redis/src/main/java/org/apache/geode/redis/internal/data/ByteArrayWrapper.java
```diff
@@ -107,6 +115,39 @@ public int compareTo(ByteArrayWrapper other) {
     return arrayCmp(value, other.value);
   }
 
+  /**
+   * Used by the {@link RedisPartitionResolver} to map slots to buckets. Supports using hashtags
+   * in the same way that redis does.
+   *
+   * @see <a href="https://redis.io/topics/cluster-spec">Redis Cluster Spec</a>
+   */
+  public synchronized Object getRoutingId() {
```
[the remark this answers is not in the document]
jdeppe-pivotal: Currently the `ByteArrayWrapper` is used for both keys and values. I have another PR that will introduce a `RedisKey` which will do exactly what you're suggesting.

geode-redis/src/main/java/org/apache/geode/redis/internal/RegionProvider.java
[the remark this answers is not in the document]
jdeppe-pivotal: done

geode-redis/src/main/java/org/apache/geode/redis/internal/RegionProvider.java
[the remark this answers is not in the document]
jdeppe-pivotal: Made it fixed

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
```diff
@@ -8149,8 +8149,8 @@ public void dumpSelfEntryFromAllPartitionedRegions() {
    * A test method to get the list of all the bucket ids for the partitioned region in the data
    * Store.
    */
-  public List getLocalBucketsListTestOnly() {
-    List localBucketList = null;
+  public List<Integer> getLocalBucketsListTestOnly() {
```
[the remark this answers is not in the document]
jdeppe-pivotal: Unfortunately no - it's used in quite a few other places.

geode-redis/src/main/java/org/apache/geode/redis/internal/RegionProvider.java
jdeppe-pivotal: Yes and done.

geode-redis/src/main/java/org/apache/geode/redis/internal/RegionProvider.java
jdeppe-pivotal: Made constant.

geode-redis/src/main/java/org/apache/geode/redis/internal/data/ByteArrayWrapper.java
jdeppe-pivotal: Thanks!

geode-redis/src/distributedTest/java/org/apache/geode/redis/internal/cluster/RedisPartitionResolverDUnitTest.java
jdeppe-pivotal: Done

geode-redis/src/distributedTest/java/org/apache/geode/redis/internal/cluster/RedisPartitionResolverDUnitTest.java
jdeppe-pivotal: Just removed and inlined the call.

DonalEvans (approved):

geode-redis/src/main/java/org/apache/geode/redis/internal/RegionProvider.java
```diff
@@ -51,4 +65,23 @@ public RegionProvider(InternalCache cache) {
...
     return configRegion;
   }
+
+  /**
+   * Validates that the value passed in is a power of 2 and that it is not greater than
+   * {@link #REDIS_SLOTS}
+   *
+   * @throws ManagementException if there is a problem with the value
+   */
+  protected static void validateBuckets(int buckets) {
```
dschneider-pivotal, line 75: the name "validateBuckets" is a little misleading. I thought it was going to validate the actual buckets. Perhaps "validateBucketCount" would be better.

geode-redis/src/main/java/org/apache/geode/redis/internal/RegionProvider.java
```diff
@@ -51,4 +65,23 @@ public RegionProvider(InternalCache cache) {
...
+    if (buckets <= 0 || ((buckets & (buckets - 1)) != 0)) {
+      throw new ManagementException(
+          "Could not start Redis Server - redis region buckets must be a power of 2. Configured value is invalid: "
```
dschneider-pivotal, line 78: Put the system property name "redis.region.buckets" in a constant and use it in the exception strings instead of "redis region buckets" to help communicate that they set the system property to an illegal value. Some readers of this exception may not be able to tie it back to the system property otherwise.

geode-redis/src/test/java/org/apache/geode/redis/internal/executor/cluster/CRC16JUnitTest.java
```diff
@@ -0,0 +1,35 @@
...
+ * the License.
+ */
+
+package org.apache.geode.redis.internal.executor.cluster;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.Test;
+
+public class CRC16JUnitTest {
```
upthewaterspout, line 22: Is this enough for us to feel confident in our implementation? Should we compare our results to Jedis or Lettuce's CRC16 implementation?

geode-redis/src/main/java/org/apache/geode/redis/internal/data/ByteArrayWrapper.java
```diff
@@ -39,6 +45,8 @@
    */
   protected byte[] value;
 
+  private transient Object routingId;
```
upthewaterspout, line 48: Do we actually call the PartitionResolver more than once on a value? I wonder if we ought to just compute the routingID in the partition resolver and not bother caching the result. This is taking up extra memory.

geode-redis/src/main/java/org/apache/geode/redis/internal/RegionProvider.java
```diff
@@ -51,4 +67,24 @@ public RegionProvider(InternalCache cache) {
...
+   * Validates that the value passed in is a power of 2 and that it is not greater than
+   * {@link #REDIS_SLOTS}
+   *
+   * @throws ManagementException if there is a problem with the value
+   */
+  protected static void validateBucketCount(int buckets) {
+    if (buckets <= 0 || ((buckets & (buckets - 1)) != 0)) {
+      throw new ManagementException(
+          String.format(
+              "Could not start Redis Server - System property '%s' must be a power of 2. Configured value is invalid: %d",
```
upthewaterspout, line 81: Technically, I don't know if we really need to restrict users to powers of two. I think your math will work fine with any number of buckets.

upthewaterspout (commented):
Overall I like the idea of getting this in to 1.14 and develop so we can potentially migrate 1.14 users to a clustered redis.

geode-redis/src/test/java/org/apache/geode/redis/internal/executor/cluster/CRC16JUnitTest.java
pivotal-jbarrett: I imagine there is a sample test out there for this specific CRC16 algorithm.

geode-redis/src/main/java/org/apache/geode/redis/internal/data/ByteArrayWrapper.java
pivotal-jbarrett: I would hope we only call it once for each incoming key operation. So yes, calculating in the resolver would work but then we leak the internals of `ByteArrayWrapper` to another class right? What if `ByteArrayWrapper.hashCode()` was the CRC16 value, then we would be good.

geode-redis/src/main/java/org/apache/geode/redis/internal/RegionProvider.java
pivotal-jbarrett: If the slots count, fixed at 16k, is not evenly divisible by the bucket count then you end up with less slots in some buckets. If the data is evenly distributed across the buckets, as described in the Redis docs by use of CRC16, then data would be imbalanced too.

geode-redis/src/main/java/org/apache/geode/redis/internal/RegionProvider.java
```diff
@@ -14,28 +14,44 @@
...
    * The name of the region that holds data stored in redis.
    */
-  private static final String REDIS_DATA_REGION = "__REDIS_DATA";
-  private static final String REDIS_CONFIG_REGION = "__REDIS_CONFIG";
+  public static final String REDIS_DATA_REGION = "__REDIS_DATA";
+  public static final String REDIS_CONFIG_REGION = "__REDIS_CONFIG";
+  public static final String REDIS_REGION_BUCKETS_PARAM = "redis.region.buckets";
+  public static final int REDIS_REGION_BUCKETS =
+      Integer.getInteger(REDIS_REGION_BUCKETS_PARAM, 128);
+  public static final int REDIS_SLOTS = Integer.getInteger("redis.slots", 16384);
```
pivotal-jbarrett, line 36: Comments said you made this fixed but it is still configurable.

pivotal-jbarrett (approved):
Minor cosmetic issues I think left so I am approving.

geode-redis/src/main/java/org/apache/geode/redis/internal/RegionProvider.java
jdeppe-pivotal: I'll remove the restriction - it's going to be undocumented in any case (at least for now).

geode-redis/src/main/java/org/apache/geode/redis/internal/data/ByteArrayWrapper.java
jdeppe-pivotal: Seems like it is called very often - certainly more than once / key

geode-redis/src/main/java/org/apache/geode/redis/internal/RegionProvider.java
jdeppe-pivotal: Thanks - fixed.

geode-redis/src/main/java/org/apache/geode/redis/internal/RegionProvider.java
jdeppe-pivotal: Done

geode-redis/src/test/java/org/apache/geode/redis/internal/executor/cluster/CRC16JUnitTest.java
jdeppe-pivotal: Testing against Jedis' implementation now.

geode-redis/src/main/java/org/apache/geode/redis/internal/RegionProvider.java
jdeppe-pivotal: Oops - thanks!

onichols-pivotal (approved):

PR #6140, opened 2021-03-15
GEODE-9035: Enable Redis TCL tests for all Hash commands

ringles:
This enables native Redis TCL tests for all of the Hash commands so that they can be run against Geode's compatibility-with-Redis module in the pipeline. This helps ensure that Geode's implementations of the commands match native Redis behavior.


nonbinaryprogrammer (approved):
looks good!

sabbey37 (commented):
It doesn't look like we actually re-generated the patch file (the insertions/deletions were not updated at the top of the file, date is the same, etc.), but the tests seem to pass in the pipeline and when I apply the patch locally, the changes are accepted without any issues... so maybe that's ok?

jdeppe-pivotal (approved):
Looks good.

PR #6146, opened 2021-03-16
GEODE-9044: Introduce RedisKey as key object for RedisData entries

jdeppe-pivotal:
- RedisKey extends ByteArrayWrapper and adds a routingId integer value.
  This is used by the RedisPartitionResolver to map entries to buckets.

### For all changes:

### Note:


jdeppe-pivotal:
I tried to break this up into one commit which has the 'actual' change and the other is switching usage of `ByteArrayWrapper` to `RedisKey` where appropriate.

nonbinaryprogrammer (approved):
looks good!

geode-redis/src/main/java/org/apache/geode/redis/internal/data/RedisKey.java
```diff
@@ -0,0 +1,94 @@
...
+      }
+    }
+
+    if (endHashtag - startHashtag <= 1) {
+      startHashtag = -1;
+      endHashtag = value.length;
+    }
+
+    // & (REDIS_SLOTS - 1) is equivalent to % REDIS_SLOTS but supposedly faster
+    routingId = (CRC16.calculate(value, startHashtag + 1, endHashtag) & (REDIS_SLOTS - 1))
```
pivotal-jbarrett, line 60: Why not keep and re-use the calculated CRC16 for `int hashCode()`?

pivotal-jbarrett, line 60: Calculating a `routingId` and serializing it not only does this key need to know something about the number of buckets and slots / buckets but if the buckets changed all the key objects have to change. If instead we keep just the CRC16 hash and let `PartitionResolver` calculate the slot/buck from the has we can decouple this.

pivotal-jbarrett (changes requested):

geode-redis/src/main/java/org/apache/geode/redis/internal/data/RedisKey.java
jdeppe-pivotal: Done

geode-redis/src/main/java/org/apache/geode/redis/internal/data/RedisKey.java
```diff
@@ -56,33 +52,31 @@ public RedisKey(byte[] value) {
...
-    DataSerializer.writeInteger(routingId, out);
+    out.writeInt(crc16);
     super.toData(out, context);
   }
 
   @Override
   public void fromData(DataInput in, DeserializationContext context)
       throws IOException, ClassNotFoundException {
-    routingId = DataSerializer.readInteger(in);
+    crc16 = in.readInt();
```
pivotal-jbarrett, line 72: Wondering out loud now if it is even worth serializing the CRC when we could just reconstitute it when deserializing.

pivotal-jbarrett (changes requested):

geode-redis/src/main/java/org/apache/geode/redis/internal/data/RedisKey.java
```diff
@@ -0,0 +1,88 @@
...
+  public KnownVersion[] getSerializationVersions() {
+    return null;
+  }
+
+  /**
+   * Used by the {@link RedisPartitionResolver} to map slots to buckets
+   */
+  @Override
+  public int hashCode() {
+    return crc16;
```
upthewaterspout, line 86: We don't want to use this as the hashCode! The crc16 is based on the hashtag, which is used to group keys into a single slot. A lot of keys are going to have the same hashtag, that's the point. So a lot of keys are going to have the same crc16 in this case.

The hashCode is used when adding the entry to the ConcurrentHashMap of the bucket. So if all of the keys have the same hashCode, our hashtable be a linked list.

Even the CRC16 of the whole key could be a bad idea, because we are creating more collisions by only having a 16 bit hashcode instead of a 32 bit one.

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/cluster/RedisPartitionResolver.java
```diff
@@ -15,16 +15,20 @@
...
+import org.apache.geode.redis.internal.data.RedisKey;
 
-public class RedisPartitionResolver implements PartitionResolver<ByteArrayWrapper, RedisData> {
+public class RedisPartitionResolver implements PartitionResolver<RedisKey, RedisData> {
 
   @Override
-  public Object getRoutingObject(EntryOperation<ByteArrayWrapper, RedisData> opDetails) {
-    return opDetails.getKey().getRoutingId();
+  public Object getRoutingObject(EntryOperation<RedisKey, RedisData> opDetails) {
+    // & (REDIS_SLOTS - 1) is equivalent to % REDIS_SLOTS but supposedly faster
```
upthewaterspout, line 30: You made my brain explode :) Is this why we are restricting the number of buckets to a power of 2? That seems not worth doing just to use & instead of / here.

upthewaterspout (changes requested):

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/cluster/RedisPartitionResolver.java
pivotal-jbarrett: No, it isn't the reason. The reason for powers of 2 is to maintain an even distribution of slots to buckets. If buckets is not a factor of slots then some buckets will hold more slots than others, leading to an imbalance of data.

geode-redis/src/main/java/org/apache/geode/redis/internal/data/RedisKey.java
pivotal-jbarrett: Good point. Probably worth adding that `crc16()` method back (though not with an `Integer` return type) for the `ParitionResolver` to use and fall back to the default `hashCode()` here.

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/cluster/RedisPartitionResolver.java
```diff
@@ -15,16 +15,20 @@
...
+    return opDetails.getKey().hashCode() & (REDIS_SLOTS - 1) / REDIS_SLOTS_PER_BUCKET;
```
upthewaterspout, line 31: Isn't the hashcode a 16 bit CRC? So doing `& (REDIS_SLOTS - 1)` is a noop? You are just taking the lower 16 bits of a 16 bit value? Maybe just doing this for safety's sake?

pivotal-jbarrett: Slots in 14 bits, this masks off the top 2 bits.

geode-redis/src/main/java/org/apache/geode/redis/internal/data/RedisKey.java
jdeppe-pivotal: Switched back

upthewaterspout (approved):
Looks good. With regard to caching the crc16 in the region entry - I agree it probably doesn't need to be serialized. I'm also not sure whether it would be more efficient to cache the hashCode or the CRC16 - I wish we didn't call the PartitionResolver more than once!

pivotal-jbarrett:
> Looks good. With regard to caching the crc16 in the region entry - I agree it probably doesn't need to be serialized. I'm also not sure whether it would be more efficient to cache the hashCode or the CRC16 - I wish we didn't call the PartitionResolver more than once!

At some point there is a tipping point were the serialization of 2 bytes is cheaper than the CPU time to recalculate the CRC remotely. Keeping in mind that for Geode that CPU is also trying to process primary updates too. Or would it even need to be calculated remotely if since it is only used for bucket placement, and that is already known when replicating.

Similar question can likely be said for the hash code's 4 bytes. Since to be placed in to correct map slot remotely the hash code must be known so is recalculation on the remote side more costly in CPU than the 4 bytes on the wire?

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/cluster/CRC16.java
```diff
@@ -59,6 +59,11 @@ public static int calculate(byte[] data, int start, int end) {
     return crc;
   }
 
+  public static int calculate(String data) {
+    byte[] bytes = data.getBytes();
```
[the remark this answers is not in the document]
jdeppe-pivotal: This is here just for test convenience.

geode-redis/src/main/java/org/apache/geode/redis/internal/data/RedisKey.java
```diff
@@ -56,33 +52,31 @@ public RedisKey(byte[] value) {
...
   @Override
   public int getDSFID() {
-    return DataSerializableFixedID.REDIS_BYTE_ARRAY_WRAPPER_KEY;
+    return DataSerializableFixedID.REDIS_KEY;
   }
 
   @Override
   public void toData(DataOutput out, SerializationContext context) throws IOException {
-    DataSerializer.writeInteger(routingId, out);
+    out.writeInt(crc16);
```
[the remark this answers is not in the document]
jdeppe-pivotal: Yup - done.

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/cluster/RedisPartitionResolver.java
```diff
@@ -24,7 +27,8 @@
 
   @Override
   public Object getRoutingObject(EntryOperation<RedisKey, RedisData> opDetails) {
-    return opDetails.getKey().getRoutingId();
+    // & (REDIS_SLOTS - 1) is equivalent to % REDIS_SLOTS but supposedly faster
+    return opDetails.getKey().getCrc16() & (REDIS_SLOTS - 1) / REDIS_SLOTS_PER_BUCKET;
```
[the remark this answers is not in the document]
jdeppe-pivotal: There already was a conversation about the bucket number. It was switched away from being a ^2.

geode-redis/src/main/java/org/apache/geode/redis/internal/data/RedisKey.java
jdeppe-pivotal: Done

geode-redis/src/main/java/org/apache/geode/redis/internal/executor/cluster/RedisPartitionResolver.java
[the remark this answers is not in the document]
pivotal-jbarrett: > It was switched away from being a ^2.
Just for clarification, it is still a power of 2 and the recommendation in the comment is a power of 2 by looking at the source. I don't recall any conversation where we agreed to not have it be a power of 2.

pivotal-jbarrett (approved):

geode-redis/src/test/java/org/apache/geode/redis/internal/data/RedisKeyJUnitTest.java
```diff
@@ -0,0 +1,97 @@
...
+    assertThat(key.getCrc16()).isEqualTo(CRC16.calculate("{}{user1000}"));
+
+    key = new RedisKey("foo{{user1000}}bar".getBytes());
+    assertThat(key.getCrc16()).isEqualTo(CRC16.calculate("{user1000"));
+  }
+
+  @Test
+  public void testSerialization_withPositiveSignedShortCRC16() throws Exception {
+    RedisKey keyOut = new RedisKey("012345".getBytes());
+    assertThat((short) keyOut.getCrc16()).isPositive();
```
Bill, line 74: It's not clear why this CRC value should be positive. Since we are trying to verify that the CRC16-XMODEM was correctly computed, this should test:

```
assertThat((short) keyOut.getCrc16()).isEqualTo((short)31712);
```

geode-redis/src/test/java/org/apache/geode/redis/internal/data/RedisKeyJUnitTest.java
```diff
@@ -0,0 +1,97 @@
...
+    ByteArrayDataInput in = new ByteArrayDataInput(out.toByteArray());
+
+    RedisKey keyIn = DataSerializer.readObject(in);
+    assertThat(keyIn).isEqualTo(keyOut);
+  }
+
+  @Test
+  public void testSerialization_withNegativeSignedShortCRC16() throws Exception {
+    RedisKey keyOut = new RedisKey("k2".getBytes());
+    assertThat((short) keyOut.getCrc16()).isNegative();
```
Bill, line 87: It's not clear why this CRC value should be negative. Since we are trying to verify that the CRC16-XMODEM was correctly computed, this should test:

```
    assertThat((short) keyOut.getCrc16()).isEqualTo((short)49601);
```

geode-redis/src/main/java/org/apache/geode/redis/internal/data/RedisKey.java
```diff
@@ -0,0 +1,88 @@
...
+  public RedisKey() {}
+
+  public RedisKey(byte[] value) {
+    super(value);
+
+    int startHashtag = Integer.MAX_VALUE;
+    int endHashtag = 0;
+
+    for (int i = 0; i < value.length; i++) {
+      if (value[i] == '{' && startHashtag == Integer.MAX_VALUE) {
```
Bill, line 42: Missing test case for this branch.

geode-redis/src/main/java/org/apache/geode/redis/internal/data/RedisKey.java
```diff
@@ -0,0 +1,88 @@
...
+        startHashtag = i;
+      } else if (value[i] == '}') {
```
Bill, line 44: Missing test case for this branch.

geode-redis/src/main/java/org/apache/geode/redis/internal/data/RedisKey.java
```diff
@@ -0,0 +1,88 @@
...
+        endHashtag = i;
+        break;
+      }
+    }
```
Bill, line 48: Missing test cases for mis-matched hash tag delimiters e.g.:

`{xyz`
`pdq}`
`}{`

geode-redis/src/main/java/org/apache/geode/redis/internal/data/RedisKey.java
```diff
@@ -0,0 +1,88 @@
...
+
+  public RedisKey() {}
+
+  public RedisKey(byte[] value) {
+    super(value);
+
+    int startHashtag = Integer.MAX_VALUE;
+    int endHashtag = 0;
+
+    for (int i = 0; i < value.length; i++) {
```
Bill, line 41: missing test case for empty array

Bill (changes requested):
Allocation of the new id value for `REDIS_KEY` looks good.

But I'm requesting changes to rectify the lack of sufficient testing of computation of CRC16 in the `RedisKey` constructor.

geode-redis/src/test/java/org/apache/geode/redis/internal/data/RedisKeyJUnitTest.java
jdeppe-pivotal: The CRC16 value is really an unsigned short. Since shorts in Java are signed, we're holding it in an int. However, when serializing on the the wire, it is sent as a short to save some bits. When deserializing we `& 0xffff` the short to get it back to an `int` and effectively an unsigned short value. I wanted to make sure that the `& 0xffff` bit worked and the semantics of that were codified in a test.

geode-redis/src/test/java/org/apache/geode/redis/internal/data/RedisKeyJUnitTest.java
jdeppe-pivotal: Ditto above.

geode-redis/src/main/java/org/apache/geode/redis/internal/data/RedisKey.java
jdeppe-pivotal: I'm not quite sure about your concern. There are 10 sample inputs in `testRoutingId_withHashtags`. Could you express a relevant test case in the form of an actual input example?

geode-redis/src/main/java/org/apache/geode/redis/internal/data/RedisKey.java
jdeppe-pivotal: Ditto

geode-redis/src/main/java/org/apache/geode/redis/internal/data/RedisKey.java
jdeppe-pivotal: Ditto

geode-redis/src/main/java/org/apache/geode/redis/internal/data/RedisKey.java
Bill: oh yeah ok, I just ran this with coverage in IJ and confirmed all branches except `getSerializationVersions()` are covered. I just didn't see `testRoutingId_withHashtags()` for some reason!

geode-redis/src/main/java/org/apache/geode/redis/internal/data/RedisKey.java
Bill: Looks like your test method draws directly from the spec in https://redis.io/topics/cluster-spec which is great!

geode-redis/src/test/java/org/apache/geode/redis/internal/data/RedisKeyJUnitTest.java
Bill: Oh I see @jdeppe-pivotal. I looked at that first assertion and stopped reading.

geode-redis/src/main/java/org/apache/geode/redis/internal/data/RedisKey.java
jdeppe-pivotal: Added

Bill (approved):
Approved

PR #6156, opened 2021-03-18
GEODE-6150_2: SSL selector impacts

mivanac:
Added modification to support selector mode for SSL connections.
Added impacts to use NioFilter in selector mode, for SSL and non-SSL connections.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/HeapDataOutputStream.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/HeapDataOutputStream.java
@@ -23,6 +23,8 @@ import java.nio.channels.SocketChannel;
 
 import org.apache.geode.DataSerializer;
 import org.apache.geode.internal.cache.BytesAndBitsForCompactor;
+import org.apache.geode.internal.net.ByteBufferSharing;
+import org.apache.geode.internal.net.NioFilter;
 import org.apache.geode.internal.serialization.KnownVersion;
 import org.apache.geode.internal.serialization.StaticSerialization;
 import org.apache.geode.internal.tcp.ByteBufferInputStream;
@@ -159,7 +161,8 @@ public class HeapDataOutputStream extends
     return result;
   }
 
-  public void sendTo(SocketChannel chan, ByteBuffer out) throws IOException {
+  public void sendTo(SocketChannel chan, ByteBuffer out, NioFilter ioFilter)
+      throws IOException {
     finishWriting();
     if (size() == 0) {
       return;
@@ -167,24 +170,23 @@ public class HeapDataOutputStream extends
     out.clear();
     if (this.chunks != null) {
       for (ByteBuffer bb : this.chunks) {
-        sendChunkTo(bb, chan, out);
+        sendChunkTo(bb, chan, out, ioFilter);
       }
     }
-    sendChunkTo(this.buffer, chan, out);
-    flushBuffer(chan, out);
+    sendChunkTo(this.buffer, chan, out, ioFilter);
+    flushBuffer(chan, out, ioFilter);
   }
 
   /**
    * sends the data from "in" by writing it to "sc" through "out" (out is used to chunk to data and
    * is probably a direct memory buffer).
    */
-  private void sendChunkTo(ByteBuffer in, SocketChannel sc, ByteBuffer out) throws IOException {
+  private void sendChunkTo(ByteBuffer in, SocketChannel sc, ByteBuffer out, NioFilter ioFilter)
+      throws IOException {
     int bytesSent = in.remaining();
     if (in.isDirect()) {
-      flushBuffer(sc, out);
-      while (in.remaining() > 0) {
-        sc.write(in);
-      }
+      flushBuffer(sc, out, ioFilter);
+      writeToSC(sc, in, ioFilter);
     } else {
       // copy in to out. If out fills flush it
       int OUT_MAX = out.remaining();
@@ -202,7 +204,7 @@ public class HeapDataOutputStream extends
           out.put(bytes, off, bytesThisTime);
           off += bytesThisTime;
           len -= bytesThisTime;
-          flushBuffer(sc, out);
+          flushBuffer(sc, out, ioFilter);
           OUT_MAX = out.remaining();
         }
         in.position(in.limit());
@@ -417,5 +419,29 @@ public class HeapDataOutputStream extends
     source.sendTo(this.buffer);
   }
 
+  private void flushBuffer(SocketChannel sc, ByteBuffer out, NioFilter ioFilter)
+      throws IOException {
+    if (out.position() == 0)
+      return;
+    out.flip();
+    writeToSC(sc, out, ioFilter);
+    out.clear();
+  }
+
+  private void writeToSC(SocketChannel sc, ByteBuffer out, NioFilter ioFilter)
+      throws IOException {
+    if (ioFilter == null) {
+      while (out.remaining() > 0) {
+        sc.write(out);
+      }
+    } else {
+      try (final ByteBufferSharing outputSharing = ioFilter.wrap(out)) {
+        final ByteBuffer wrappedBuffer = outputSharing.getBuffer();
+        while (wrappedBuffer.remaining() > 0) {
+          sc.write(wrappedBuffer);
+        }
+      }
+    }
+  }
 
 }
--- a/geode-core/src/main/java/org/apache/geode/internal/HeapDataOutputStream.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/HeapDataOutputStream.java
@@ -186,7 +186,7 @@ public class HeapDataOutputStream extends
     int bytesSent = in.remaining();
     if (in.isDirect()) {
       flushBuffer(sc, out, ioFilter);
-      writeToSC(sc, in, ioFilter);
+      writeToSocketChannel(sc, in, ioFilter);
     } else {
       // copy in to out. If out fills flush it
       int OUT_MAX = out.remaining();
@@ -424,11 +424,11 @@ public class HeapDataOutputStream extends
     if (out.position() == 0)
       return;
     out.flip();
-    writeToSC(sc, out, ioFilter);
+    writeToSocketChannel(sc, out, ioFilter);
     out.clear();
   }
 
-  private void writeToSC(SocketChannel sc, ByteBuffer out, NioFilter ioFilter)
+  private void writeToSocketChannel(SocketChannel sc, ByteBuffer out, NioFilter ioFilter)
       throws IOException {
     if (ioFilter == null) {
       while (out.remaining() > 0) {
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/AcceptorImpl.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/AcceptorImpl.java
@@ -57,6 +57,8 @@ import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Supplier;
 
+import javax.net.ssl.SSLEngine;
+
 import io.micrometer.core.instrument.MeterRegistry;
 import org.apache.logging.log4j.Logger;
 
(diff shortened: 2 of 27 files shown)


onichols-pivotal:
I'm not really clear what I'm looking at here.  Given zero description of the intended change beyond "SSL selector impacts" it's hard to tell, but many of the changes look more like a merge-gone-bad...

mivanac:
Main difference to previous PR, is modification of Message class, to use generic NioFilter, not just NioSslEngine.
I will add description.

mivanac:
Also you should read RFC https://cwiki.apache.org/confluence/display/GEODE/Server-Side+Thread+Pooling+for+Client-Server+TLS before reviewing this PR.

DonalEvans:
Looking at the changes included in this PR, it seems like something has gone wrong with rebasing/merging the git branch. It looks like at least 4 of the most recent commits to develop are somehow also included in this change set.

echobravopapa (changes requested):
@mivanac what's the plan for getting the RFC discussed? https://cwiki.apache.org/confluence/display/GEODE/Server-Side+Thread+Pooling+for+Client-Server+TLS

the RFC at present is a placeholder, awaiting your expansion of the your plan on all of this...

mivanac:
There was no comments on this RFC until now, only comments on previous PRs. So I don`t know what are we discussing, potential new alternatives? Current PR is next step in systematization of solution. I can write this info in RFC.

geode-core/src/main/java/org/apache/geode/internal/HeapDataOutputStream.java
```diff
@@ -159,32 +161,32 @@ public int sendTo(SocketChannel chan) throws IOException {
...
+  private void sendChunkTo(ByteBuffer in, SocketChannel sc, ByteBuffer out, NioFilter ioFilter)
+      throws IOException {
     int bytesSent = in.remaining();
     if (in.isDirect()) {
-      flushBuffer(sc, out);
-      while (in.remaining() > 0) {
-        sc.write(in);
-      }
+      flushBuffer(sc, out, ioFilter);
+      writeToSC(sc, in, ioFilter);
```
agingade, line 189: I am not expert in this area. But from the code changes it looks like we are calling writeToSC twice now (in flushBuffer)...Is this expected?

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/AcceptorImpl.java
```diff
@@ -967,7 +981,8 @@ private void checkForStuckKeys() {
             tmpSel.selectNow(); // clear canceled key
             sc.registerWithSelector2(selector);
           } else {
-            if (tmpsk.isValid() && tmpsk.isReadable()) {
+            if (tmpsk.isValid() && (tmpsk.isReadable() || tmpsk.isWritable())) {
```
dschneider-pivotal, line 984: Why did you add isWritable? I think the old code only used non-blocking on reads. Are you also now doing non-blocking writes?
Note this common expression (line 984 and 1129) could be extracted into a method

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/AcceptorImpl.java
```diff
@@ -1455,22 +1478,24 @@ private void handleNewClientConnection(final Socket socket) throws IOException {
             "Rejected connection from {} because current connection count of {} is greater than or equal to the configured max of {}",
             new Object[] {socket.getInetAddress(), curCnt,
                 maxConnections});
+
         try {
           refuseHandshake(socket.getOutputStream(),
               String.format("exceeded max-connections %s",
                   maxConnections),
-              REPLY_REFUSED);
+              REPLY_REFUSED, ioFilter, socket);
```
dschneider-pivotal, line 1486: since you are now passing "socket" into refuseHandshake, you no longer need to also pass in socket.getOutputStream()

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientRegistrationMetadata.java
```diff
@@ -47,51 +49,67 @@
...
-        dataOutputStream =
-            new VersionedDataOutputStream(unversionedDataOutputStream, clientVersion);
+    InputStream inputStream = null;
+    try {
+      if (ioFilter == null) {
+        inputStream = socket.getInputStream();
       } else {
-        dataInputStream = unversionedDataInputStream;
-        dataOutputStream = unversionedDataOutputStream;
+        inputStream = ioFilter.getInputStream(socket);
```
dschneider-pivotal, line 68: shouldn't the initialization of inputStream happen right before the try (instead of after) since all the finally block does is close the inputStream? Also if you then extracted these lines of code into a method you could then have: final InputStream inputStream = getInputStream(socket, ioFilter); try ...

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/Message.java
```diff
@@ -882,6 +1000,12 @@ void readPayloadFields(final int numParts, final int len) throws IOException {
     int readSecurePart = checkAndSetSecurityPart();
 
     int bytesRemaining = len;
+
+    if (this.ioFilter != null) {
```
dschneider-pivotal, line 1004: In some cases we have a filter even if not using SSL, NioPlainEngine. What is the difference between ioFilter being null and it being an instance of NioPlainEngine? Would it be possible to have it never be null? If so, it would be nice to get rid of all these null checks

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/AcceptorImpl.java
mivanac: Without this, test are failing, and no connection can not be established.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientRegistrationMetadata.java
mivanac: Updated

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/AcceptorImpl.java
mivanac: updated

geode-core/src/main/java/org/apache/geode/internal/HeapDataOutputStream.java
```diff
@@ -415,5 +417,29 @@ public void write(ByteBufferInputStream.ByteSource source) {
     source.sendTo(this.buffer);
   }
 
+  private void flushBuffer(SocketChannel sc, ByteBuffer out, NioFilter ioFilter)
```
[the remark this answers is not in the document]
mivanac: deleted old method

geode-core/src/main/java/org/apache/geode/internal/HeapDataOutputStream.java
```diff
@@ -415,5 +417,29 @@ public void write(ByteBufferInputStream.ByteSource source) {
...
+      throws IOException {
+    if (out.position() == 0)
+      return;
+    out.flip();
+    writeToSC(sc, out, ioFilter);
+    out.clear();
+  }
+
+  private void writeToSC(SocketChannel sc, ByteBuffer out, NioFilter ioFilter)
```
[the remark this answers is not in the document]
mivanac: updated

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/AcceptorImpl.java
```diff
@@ -1424,10 +1439,18 @@ private void handleNewClientConnection(final Socket socket) throws IOException {
...
+    final NioFilter ioFilter;
     try {
       if (isSelector()) {
-        communicationMode = getCommunicationModeForSelector(socket);
+        if (socketCreator.forCluster().useSSL()) {
+          ioFilter = createNIOSSLEngine(socket.getChannel());
+          communicationMode = getCommunicationModeForSSLSelector(socket, ioFilter);
+        } else {
+          ioFilter = new NioPlainEngine(bufferPool);
+          communicationMode = getCommunicationModeForSelector(socket);
```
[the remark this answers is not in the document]
mivanac: working on it

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/Message.java
mivanac: I will investigate, but I think there are cases when for socketchannel there are not defined iofilter.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/AcceptorImpl.java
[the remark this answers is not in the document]
mivanac: updated

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/Message.java
```diff
@@ -663,35 +672,53 @@ void flushBuffer() throws IOException {
     cb.clear();
   }
 
+  private void writeToSC(ByteBuffer cb) throws IOException {
```
dschneider-pivotal, line 675: rename writeToSocketChannel

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/Message.java
```diff
@@ -1161,10 +1288,146 @@ public void receive() throws IOException {
   public void receive(ServerConnection sc, int maxMessageLength, Semaphore dataLimiter,
       Semaphore msgLimiter) throws IOException {
     this.serverConnection = sc;
+    if (sc.getIOFilter() != null) {
```
dschneider-pivotal, line 1291: I see this pattern in many places. It seems like the null check could be moved into setIOFilter which would simplify all the callers.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ServerConnection.java
```diff
@@ -1307,7 +1312,11 @@ SelectableChannel getSelectableChannel() {
   }
 
   void registerWithSelector2(Selector s) throws ClosedChannelException {
-    getSelectableChannel().register(s, SelectionKey.OP_READ, this);
+    if ((getIOFilter() != null) && (getIOFilter() instanceof NioSslEngine)) {
```
dschneider-pivotal, line 1315: I think the "!= null" is not needed since "null instanceof X" always returns false.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ServerSideHandshakeImpl.java
```diff
@@ -49,6 +52,7 @@
   private final KnownVersion clientVersion;
 
   private final byte replyCode;
+  private ServerConnection connection;
```
dschneider-pivotal, line 55: make "connection" final

kirklund (changes requested):
Changes look good, but I really want to see some new unit tests and/or integration tests to cover the changes in the classes listed below. I usually focus more on writing unit rather than integration tests. 

Unit tests do not touch filesystem, sockets, or generally do not use real classes other than small data classes outside the class under test. Integration tests can use the filesystem, sockets, or even create DistributedSystem or Cache instances.

The first step of creating a new unit test for an existing class usually involves changing constructors so you can pass in all the dependencies. This can involve adding chained constructors that chain down to a single final constructor that accepts all dependencies which is then annotated with `@VisibleForTesting` so you can pass in mocks for the tests.

Writing tests for existing tests that have no tests can end up being a sizable portion of the work for pull requests like this. I generally just try to add test coverage for the behavior I've actually touched, changed, or added to rather than back-filling tests for the entire class.

Here's the classes that I believe need at least some new unit tests and/or integration test coverage for the changes you've made in the PR:
* AcceptorImpl (AcceptorImplTest and AcceptorImplIntegrationTest)
* CacheClientNotifier (CacheClientNotifierTest)
* ClientRegistrationMetadata (new ClientRegistrationMetadataTest)
* HeapDataOutputStream (HeapDataOutputStreamTest or HeapDataOutputStreamJUnitTest)
* Message (MessageJUnitTest, maybe a new integration test)
* MessageDispatcher (no tests exist)
* Part (no tests exist)
* ServerConnection (ServerConnectionTest and/or ServerConnectionIntegrationTest)
* ServerSideHandshakeImpl (no tests exist)
* SocketMessageWriter (SocketMessageWriterTest or maybe a new SocketMessageWriterIntegrationTest if using real sockets)

onichols-pivotal:

mivanac:
This PR is on hold.

PR #6159, opened 2021-03-19
GEODE-8946: User Guide - add Upgrade link to Getting Started page

davebarnes97:
Loose end from previous PR: Add an entry on the top-level Getting Started page to the Upgrade section


karensmolermiller (approved):
Excellent.  Thanks for finding and correcting this detail.

PR #6179, opened 2021-03-24
DRAFT: GEODE-6143: Remove PowerMock from offheap tests

kirklund:


rhoughton-pivot (approved):

mkevo (approved):

geode-core/src/main/java/org/apache/geode/internal/offheap/ReferenceCounterInstance.java
```diff
@@ -0,0 +1,115 @@
...
+      }
+      retryCount++;
+      if (retryCount > 1000) {
+        throw new IllegalStateException("tried to write " + (rawBits + 1) + " to @"
+            + Long.toHexString(memAddr) + " 1,000 times.");
+      }
+    } while (!AddressableMemoryManager.writeIntVolatile(
+        memAddr + OffHeapStoredObject.REF_COUNT_OFFSET, rawBits,
+        rawBits + 1));
+    // debugLog("use inced ref count " + (uc+1) + " @" + Long.toHexString(memAddr), true);
```
DonalEvans, line 52: This commented out code can be removed.

geode-core/src/main/java/org/apache/geode/internal/offheap/ReferenceCounterInstance.java
```diff
@@ -0,0 +1,115 @@
...
+    boolean returnToAllocator;
+    do {
+      returnToAllocator = false;
+      rawBits =
+          AddressableMemoryManager.readIntVolatile(memAddr + OffHeapStoredObject.REF_COUNT_OFFSET);
+      if ((rawBits & OffHeapStoredObject.MAGIC_MASK) != OffHeapStoredObject.MAGIC_NUMBER) {
+        String msg = "It looks like off heap memory @" + Long.toHexString(memAddr)
+            + " was already freed. rawBits=" + Integer.toHexString(rawBits) + " history="
+            + ReferenceCountHelper.getFreeRefCountInfo(memAddr);
+        // debugLog(msg, true);
```
DonalEvans, line 77: This commented out code can be removed.

geode-core/src/main/java/org/apache/geode/internal/offheap/ReferenceCounterInstance.java
```diff
@@ -0,0 +1,115 @@
...
+        throw new IllegalStateException(msg);
+      }
+      int curCount = rawBits & OffHeapStoredObject.REF_COUNT_MASK;
+      if ((curCount) == 0) {
+        // debugLog("too many frees @" + Long.toHexString(memAddr), true);
```
DonalEvans, line 82: This commented out code can be removed.

geode-core/src/main/java/org/apache/geode/internal/offheap/ReferenceCountHelper.java
```diff
@@ -81,72 +85,72 @@ public static Object createReferenceCountOwner() {
...
-    getInstance().skipRefCountTracking();
+    delegate().skipRefCountTracking();
   }
 
   /**
    * Returns true if currently tracking reference counts.
    */
-  public static boolean isRefCountTracking() {
-    return getInstance().isRefCountTracking();
+  static boolean isRefCountTracking() {
```
DonalEvans, line 94: This method (and several others in this class) are unused. Would it be appropriate to remove them?

geode-core/src/main/java/org/apache/geode/internal/offheap/ReferenceCountHelperImpl.java
```diff
@@ -278,28 +290,47 @@ void freeRefCountInfo(Long address) {
...
+  void getReferenceCountInfoTestHook(ConcurrentMap<Long, List<RefCountChangeInfo>> stacktraces,
+      long address) {
+    // nothing
+  }
+
+  /**
+   * This method is overridden during testing to allow simulation of a race to be the first to
+   * reference a given address
+   */
+  void refCountChangedTestHook(Long address, boolean decRefCount, int rc) {
```
DonalEvans, line 333: The `address` argument in this method is a `Long` but the `address` argument in `getReferenceCountInfoTestHook()` is a `long`. For consistency, would it be best for both to be the same type?

geode-core/src/test/java/org/apache/geode/internal/offheap/ReferenceCountHelperImplTest.java
```diff
@@ -12,976 +12,966 @@
...
-    ai = rchi.getReenterCount();
-    assertEquals(1, ai.get());
-    assertEquals(rchi.getReferenceCountOwner(), re);
+    RegionEntry regionEntry = mock(RegionEntry.class);
+    referenceCountHelperImpl.setReferenceCountOwner(regionEntry);
+    reenterCount = referenceCountHelperImpl.getReenterCount();
+    assertThat(reenterCount.get()).isOne();
+    assertThat(regionEntry).isEqualTo(referenceCountHelperImpl.getReferenceCountOwner());
 
     Long address = (long) 0x1000;
```
DonalEvans, line 199: Multiple compiler warnings in this class can be fixed by replacing lines like this with just `long address = 0x1000;`. Alternately, as the address used is always the same in every test, it could be extracted to a constant.

DonalEvans (changes requested):
Just some small clean-up suggestions.

geode-core/src/main/java/org/apache/geode/internal/offheap/ReferenceCountHelper.java
dschneider-pivotal: be careful with this. You could check with Eric but I think some of these methods are used when trying to debug offheap leaks (i.e. orphans)

dschneider-pivotal (approved):

geode-core/src/main/java/org/apache/geode/internal/offheap/ReferenceCountHelper.java
kirklund: I'm going to keep the unused methods in this class and restore the methods to public. We can clean this class up further next time.

geode-core/src/test/java/org/apache/geode/internal/offheap/ReferenceCountHelperImplTest.java
kirklund: It's too difficult to explain why I left it as `Long` so I'll just make the change. This change will cause the test to take slightly longer to run because it will change a `long` back to `Long` for use as a key in a map.

geode-core/src/main/java/org/apache/geode/internal/offheap/ReferenceCountHelperImpl.java
kirklund: I don't want to change this class further without someone spending time analyzing the change in performance for using Long vs long parameters. There is a small performance change if I change the to/from a wrapper instead of primitive.

DonalEvans (approved):

nabarunnag (approved):
The methods doesClearNeedtoCheckForOffHeap and doWithOffHeap extracted into their own classes and covers all existing functionality.

--Approved!

onichols-pivotal (approved):

gesterzhou (approved):

pivotal-eshu (approved):

PR #6182, opened 2021-03-24
GEODE-9062: Fix for DataSerializableJUnit in StressTest

alb3rtobr:


geode-core/src/test/java/org/apache/geode/internal/DataSerializableJUnitTest.java
```diff
@@ -2255,63 +2259,83 @@ public void testSupportedClasses() throws Exception {
...
-      assertTrue(Class_testSupportedClasses2.fromDataInvoked);
-      assertEquals(o, o2);
+      assertThat(Class_testSupportedClasses2.fromDataInvoked).isTrue();
+      assertThat(o).isEqualTo(o2);
     } finally {
       InternalDataSerializer.unregister(id);
       InternalDataSerializer.unregister(id2);
     }
+
+    Class_testSupportedClasses2.wasInvoked = false;
```
bschuchardt, line 2274: It would be nice to move this to an "@After" or a "@Before" and get rid of the duplicate code.

alb3rtobr: It is not exactly the same code. Each of the three failing tests is changing a different static class. `testSupportedClasses` is using `Class_testSupportedClasses2`, `testUDDS2` uses `Class_testSupportedClasses3` and `testUDDS4` uses `Class_testSupportedClasses4`.

alb3rtobr: I mean, I could move the code to an `@After` method, but it will not remove any code, I should do the same calls anyway.

bschuchardt (approved):
Thanks for clarifying that for me and for fixing this test

kirklund (approved):

PR #6193, opened 2021-03-25
GEODE-9063: Autocomplete with mandatory params first

alb3rtobr:


geode-gfsh/src/integrationTest/java/org/apache/geode/management/internal/cli/GfshParserConverterTest.java
```diff
@@ -103,7 +103,7 @@ public void testJsonKey() {
   public void testUnspecifiedValueToStringArray() {
     String command = "change loglevel --loglevel=finer --groups=group1,group2";
     ParseResult result = parser.parse(command);
-    String[] memberIdValue = (String[]) result.getArguments()[0];
```
mhansonp, line 106: Why did this switch to 1?

alb3rtobr: Due to the change in the order of parameters in `ChangeLogLevelCommand`, the memberId is returned in second position by the parser, so I had to modify the array index from 0 to 1.

jinmeiliao (approved):
LGTM

geode-gfsh/src/integrationTest/java/org/apache/geode/management/internal/cli/GfshParserAutoCompletionIntegrationTest.java
```diff
@@ -447,4 +447,124 @@ public void testCompletionCreateGatewaySenderWithDash() {
     assertThat(candidate.getCandidates()).hasSize(1);
     assertThat(candidate.getFirstCandidate()).isEqualTo(buffer + "id");
   }
+
+  @Test
+  public void forChangeLogLevelWithSpaceCompletionOffersMandatoryOptionsInAlphabeticalOrder() {
```
albertogpz, line 452: I would rename this test case and the rest to align them with the ones already testing something similar, like for example, testCompletionCreateGatewaySenderWithSpace.

In this case, the name would be: testCompletionChangeLogLevelWithSpace.
If you want to add also the expected outcome in the name like you already have in your original proposal, better.

This comments applies to all new test cases.

albertogpz (changes requested):

albertogpz (approved):
Nice changes. Thanks!

PR #6195, opened 2021-03-25
GEODE-9067: register-interest done during rolling upgrade can take longer than needed

mhansonp:
Basically this code was being run to individually get keys for "older versions" that could have been done in bulk except this code didn't have a version limit. Older versions of the correct vintage are not supported anymore so this code should be (have been) removed. Unfortunately older versions for this code mean any previous version, so it was choosing the single entry approach when it should have been using the bulk approach. 

This gets rid of the single approach.

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
@@ -4559,18 +4559,6 @@ public class PartitionedRegion extends LocalRegion
         }
       }
 
-      // Handle old nodes for Rolling Upgrade support
-      Set<Integer> failedSet = handleOldNodes(nodeToBuckets, values, servConn);
-      // Add failed buckets to nodeToBuckets map so that these will be tried on
-      // remote nodes.
-      if (!failedSet.isEmpty()) {
-        for (Integer bId : failedSet) {
-          failures.put(bId, bucketKeys.get(bId));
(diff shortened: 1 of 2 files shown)


geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
```diff
@@ -4602,33 +4589,6 @@ void updateNodeToBucketMap(
...
-        bucketKeys = (HashMap<Integer, HashSet>) bucketInfo;
-        buckets = bucketKeys.keySet();
-      }
-
-      fetchKeysAndValues(values, servConn, failures, member, bucketKeys, buckets);
-    }
-    return failures;
-  }
-
   void fetchKeysAndValues(VersionedObjectList values, ServerConnection servConn,
```
DonalEvans, line 4592: With the removal of `handleOldNodes()`, this method is only used in test code. It could be safely removed along with the test cases that use it. The same is also true of the `getFetchKeysResponse()` and `getValuesForKeys()` methods, which would also now only be used in the same two test cases. The `removeMembersWithSameOrNewerVersion()` method in the `DistributionManager` interface and subclasses is also unused following this change, although in the case of this method, it may be fine to keep it.

DonalEvans (changes requested):
Some additional dead code that can be removed.

DonalEvans (approved):

PR #6199, opened 2021-03-26
GEODE-8876: Statistics not collecting correct value for gets when tra…

jvarenina:
…nsaction is used

The issue:

Statistics for gets and misses in CachePerfStats is not incremented when get operation
is executed on server as a part of transaction to retrieve value from the locally hosted bucket.
The problem is that PartitionRegion#endGet method is overridden to do nothing, and therefore
statistics in not updated when that method is called for this particular case. It is wrongly stated there
in comment that the BucketRegion#endGet will be eventually called and statistics incremented.

The fix:

The solution is to actually increment the statistics in PartitionRegion#endGet method.

Additional changes are made to keep the design base behavior:
  - increment miss on server when netSearch and netLoad is successfully performed
  - increment miss on client when data is fetched from remote server (LocalRegion#findObjectInSystem)
  - increment miss on server when data is fetched from remote bucket (PartitionedRegion#getRemotely)

Design base behavior is changed for the case when data is retrieved by using local cache loader. 
In that case miss statistics will not be incremented.

### For all changes:

### Note:


onichols-pivotal:
This PR has been untouched for several months, can it be closed?

jvarenina:
> This PR has been untouched for several months, can it be closed?

Hi @onichols-pivotal ,

Thanks for the reminder. At that time I didn't know how to solve this bug without breaking other cases. I have pushed new commit now, lets see if it will work this time.

geode-core/src/distributedTest/java/org/apache/geode/management/CachePerfStatsTest.java
```diff
@@ -0,0 +1,298 @@
...
+    server1 = getVM(0);
+    server2 = getVM(1);
+    hostName = getHostName();
+    uniqueName = getClass().getSimpleName() + "_" + testName.getMethodName();
+    regionNamePartitioned = uniqueName + "_partition_region";
+    regionNameReplicated = uniqueName + "_replicated_region";
+  }
+
+  @Test
+  public void testGetsRatesPartitioned() {
```
DonalEvans, line 86: I'm not sure I understand what "rates" refers to in this class. Could this just be "getStatsAreIncrementedCorrectlyForPartitionedRegion" or something similar?

geode-core/src/distributedTest/java/org/apache/geode/management/CachePerfStatsTest.java
```diff
@@ -0,0 +1,298 @@
...
+import org.apache.geode.internal.cache.PartitionedRegion;
+import org.apache.geode.internal.cache.PartitionedRegionDataStore;
+import org.apache.geode.internal.cache.TXManagerImpl;
+import org.apache.geode.test.dunit.VM;
+import org.apache.geode.test.dunit.rules.CacheRule;
+import org.apache.geode.test.dunit.rules.ClientCacheRule;
+import org.apache.geode.test.dunit.rules.DistributedRule;
+import org.apache.geode.test.junit.rules.serializable.SerializableTestName;
+
+public class CachePerfStatsTest implements Serializable {
```
DonalEvans, line 51: Could this be named `CachePerfStatsDUnitTest` to differentiate it from the `CachePerfStatsTest` unit test class?

geode-core/src/distributedTest/java/org/apache/geode/management/CachePerfStatsTest.java
jvarenina: I will change it to `getStatsAreIncrementedCorrectlyForPartitionedRegion`. This "rates" left from old case/file that I deleted, and I'm also not sure what does it mean. Thanks for the review"

geode-core/src/distributedTest/java/org/apache/geode/management/CachePerfStatsTest.java
jvarenina: This makes sense, I will rename the file as you suggested.

DonalEvans (approved):

kirklund (approved):

mkevo (approved):

PR #6200, opened 2021-03-26
GEODE-9004: Fix issues in queries targeting a Map field

albertogpz:
Queries in which map fields are involved
in the condition, sometimes do not return the same entries
if map indexes are used, compared to when map indexes
are not used.

Differences were found when the condition on the
map field was of type '!=' and also when the condition
on the map field was comparing the value with null.

Example1:
Running a query with the following condition:
positions['SUN'] = null

in a region with the following entries:
entry1.positions=null
entry2.positions={'a'=>'b'}
entry3.positions={'SUN'=>null}

- will return entry1 and entry3 if there are no
  indexes defined.
- will return no entries if the following index is defined:
  positions['SUN','c']
- will return entry3 if the following index is defined:
  positions[*]

Example2:
Running a query with the following condition:
positions['SUN'] != '3'

in a region with the following entries:
entry1.positions=null
entry2.positions={'a'=>'b'}
entry3.positions={'SUN'=>'4'}
entry4.positions={'SUN'=>'3'}
entry5.positions={'SUN'=>null}

- will return all entries except for entry4 if:
  there are no indexes defined.
- will return entry3 if the following index is defined:
  positions['SUN','c']
- will return entry3 and entry5 if the following index is defined:
  positions[*]

In order to have the same results for these
queries no matter if indexes are used,
the following changes have been made:
- When running queries, the result of getting the
value from a Map that does not contain the key used
to access the value will be
considered UNDEFINED instead of null.
This will imply that queries where the condition is
positions['SUN'] = null will not return entries
that do not have the 'SUN' key in the positions Map.

- Queries with map indexes will not be able to use
the indexes if the query is of type '!=' against
any of the indexed map fields.
Example:
for index positions['SUN', 'ERIC'] or
positions[*]
the following query will not use indexes:
positions['SUN'] != '44'

These changes must be documented accordingly
in the corresponding indexes section.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/integrationTest/java/org/apache/geode/cache/query/internal/index/MapRangeIndexMaintenanceJUnitTest.java
+++ b/geode-core/src/integrationTest/java/org/apache/geode/cache/query/internal/index/MapRangeIndexMaintenanceJUnitTest.java
@@ -50,7 +50,7 @@ public class MapRangeIndexMaintenanceJUnitTest {
   private static final String INDEX_NAME = "keyIndex1";
 
   private static QueryService qs;
-  private static Region region;
+  private static Region<Object, Object> region;
   private static Index keyIndex1;
 
   @Before
@@ -353,7 +353,7 @@ public class MapRangeIndexMaintenanceJUnitTest {
   }
 
   @Test
-  public void testQueriesForValueInMapFieldWithMapIndexWithOneKey() throws Exception {
+  public void testQueriesForValueInMapFieldWithCompactMapIndexWithOneKey() throws Exception {
     region =
         CacheUtils.getCache().createRegionFactory(RegionShortcut.REPLICATE).create("portfolio");
     qs = CacheUtils.getQueryService();
@@ -384,7 +384,7 @@ public class MapRangeIndexMaintenanceJUnitTest {
   }
 
   @Test
-  public void testQueriesForValueInMapFieldWithMapIndexWithSeveralKeys() throws Exception {
+  public void testQueriesForValueInMapFieldWithCompactMapIndexWithSeveralKeys() throws Exception {
     region =
         CacheUtils.getCache().createRegionFactory(RegionShortcut.REPLICATE).create("portfolio");
     qs = CacheUtils.getQueryService();
@@ -421,7 +421,7 @@ public class MapRangeIndexMaintenanceJUnitTest {
   }
 
   @Test
-  public void testQueriesForValueInMapFieldWithMapIndexWithStar() throws Exception {
+  public void testQueriesForValueInMapFieldWithCompactMapIndexWithStar() throws Exception {
     region =
         CacheUtils.getCache().createRegionFactory(RegionShortcut.REPLICATE).create("portfolio");
     qs = CacheUtils.getQueryService();
@@ -442,7 +442,7 @@ public class MapRangeIndexMaintenanceJUnitTest {
     // positions map takes for each key (null not included)
     // for each entry in the region:
     // "something", null, "nothing", "more", "hey", "tip"
-    assertThat(keys).isEqualTo(6);
+    assertThat(keys).isEqualTo(7);
     // The number of mapIndexKeys must be equal to the number of different keys
     // that appear in entries of the region:
     // "IBM", "ERICSSON", "HP", "SUN", null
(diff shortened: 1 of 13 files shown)


DonalEvans:
The PR description says that "These changes must be documented accordingly in the corresponding indexes section." Is there a GEODE ticket for this docs work? If not, one should be created, or the docs changes included as part of this PR.

albertogpz:
> The PR description says that "These changes must be documented accordingly in the corresponding indexes section." Is there a GEODE ticket for this docs work? If not, one should be created, or the docs changes included as part of this PR.

I agree, Donal. My plan is to update the documentation as part of this PR if the solution is accepted. Given that the outcome of some queries without indexes change and that indexes will not be used with queries using "!=" with Map indexes, I want to have first the blessing from the community and only then update the documentation.

DonalEvans:
I'm not a query expert, so I don't understand what the reasoning is behind map indexes no longer being used for queries using `!=`. Would it be possible to get an explanation of the motivation behind this change?

albertogpz:
> I'm not a query expert, so I don't understand what the reasoning is behind map indexes no longer being used for queries using `!=`. Would it be possible to get an explanation of the motivation behind this change?

With the current implementation, map indexes are used (if defined) in queries that have a condition using "!=" against a value in the map field indexed. The problem is that these queries offer different outcome if the index is defined vs when the index is not defined.
In this PR I have made the queries offer the same output no matter if the index is defined or not but at the expense of not using the index if the condition is "!=" when the index is defined. I have not been able to come up with a solution without losing this "feature". That's the reason behind "map indexes no longer being used for queries using `!=`."

mhansonp (approved):

geode-core/src/integrationTest/java/org/apache/geode/cache/query/internal/index/MapRangeIndexMaintenanceJUnitTest.java
```diff
@@ -341,6 +344,167 @@ public void testNullMapValuesInIndexOnLocalRegionForCompactMap() throws Exceptio
...
+        ((CompactRangeIndex) keyIndex1).internalIndexStats.getNumberOfMapIndexKeys();
+    long values =
+        ((CompactRangeIndex) keyIndex1).internalIndexStats.getNumberOfValues();
+    long uses =
+        ((CompactRangeIndex) keyIndex1).internalIndexStats.getTotalUses();
+
+    assertThat(keys).isEqualTo(4);
+    assertThat(mapIndexKeys).isEqualTo(0);
+    assertThat(values).isEqualTo(7);
+    assertThat(uses).isEqualTo(4);
```
DonalEvans, line 376: It's not immediately obvious where these values are coming from. Woudl it be possible to add a comment explaining why they are what they are?

geode-core/src/integrationTest/java/org/apache/geode/cache/query/internal/index/MapRangeIndexMaintenanceJUnitTest.java
```diff
@@ -341,6 +344,167 @@ public void testNullMapValuesInIndexOnLocalRegionForCompactMap() throws Exceptio
...
+        ((CompactMapRangeIndex) keyIndex1).internalIndexStats.getNumberOfMapIndexKeys();
+    long values =
+        ((CompactMapRangeIndex) keyIndex1).internalIndexStats.getNumberOfValues();
+    long uses =
+        ((CompactMapRangeIndex) keyIndex1).internalIndexStats.getTotalUses();
+
+    assertThat(keys).isEqualTo(3);
+    assertThat(mapIndexKeys).isEqualTo(1);
+    assertThat(values).isEqualTo(3);
+    assertThat(uses).isEqualTo(2);
```
DonalEvans, line 401: See above comment about it being unclear where these values are coming from.

geode-core/src/integrationTest/java/org/apache/geode/cache/query/internal/index/MapRangeIndexMaintenanceJUnitTest.java
```diff
@@ -341,6 +344,167 @@ public void testNullMapValuesInIndexOnLocalRegionForCompactMap() throws Exceptio
...
+    long mapIndexKeys =
+        ((CompactMapRangeIndex) keyIndex1).internalIndexStats.getNumberOfMapIndexKeys();
+    long values =
+        ((CompactMapRangeIndex) keyIndex1).internalIndexStats.getNumberOfValues();
+    long uses =
+        ((CompactMapRangeIndex) keyIndex1).internalIndexStats.getTotalUses();
+    assertThat(keys).isEqualTo(5);
+    assertThat(mapIndexKeys).isEqualTo(4);
+    assertThat(values).isEqualTo(5);
+    assertThat(uses).isEqualTo(2);
```
DonalEvans, line 424: See above comment about it being unclear where these values are coming from.

geode-core/src/integrationTest/java/org/apache/geode/cache/query/internal/index/MapRangeIndexMaintenanceJUnitTest.java
```diff
@@ -341,6 +344,167 @@ public void testNullMapValuesInIndexOnLocalRegionForCompactMap() throws Exceptio
     assertEquals(1, result.size());
   }
 
+  @Test
+  public void testQueriesForValueInMapFieldWithoutIndex() throws Exception {
+    region =
```
DonalEvans, line 349: Multiple warnings in this file can be cleaned up by making `region` a `Region<Object, Object>` where it's declared.

DonalEvans (changes requested):
A few requests for code clean-up/clarifying comments. The content of the changes looks okay to me, but the conceptual changes need someone with a better understanding of the expected behaviour of queries and indexes.

kirklund (approved):

geode-core/src/integrationTest/java/org/apache/geode/cache/query/internal/index/MapRangeIndexMaintenanceJUnitTest.java
albertogpz: I have added comments about these and the other. Nevertheless, there is no documentation on them so I am not 100% sure that they are working correctly. Anyhow, it is not something changed by this PR.

geode-core/src/integrationTest/java/org/apache/geode/cache/query/internal/index/MapRangeIndexMaintenanceJUnitTest.java
```diff
@@ -341,6 +344,306 @@ public void testNullMapValuesInIndexOnLocalRegionForCompactMap() throws Exceptio
...
+    long values =
+        ((CompactMapRangeIndex) keyIndex1).internalIndexStats.getNumberOfValues();
+    long uses =
+        ((CompactMapRangeIndex) keyIndex1).internalIndexStats.getTotalUses();
+
+    // The number of keys must be equal to the number of different values the
+    // positions map takes for each key (null not included)
+    // for each entry in the region:
+    // "something", null, "nothing", "more", "hey", "tip"
+    assertThat(keys).isEqualTo(7);
```
DonalEvans, line 445: This value is now in disagreement with the comment above, as the list of values includes only 6 values. Either the explanation in the comment needs to be changed, or the value being returned here is incorrect and there is a bug somewhere. Also, it is a little confusing in terms of what "null not included" is referring to, since the list of values includes null.

geode-core/src/integrationTest/java/org/apache/geode/cache/query/internal/index/MapRangeIndexMaintenanceJUnitTest.java
```diff
@@ -341,6 +344,306 @@ public void testNullMapValuesInIndexOnLocalRegionForCompactMap() throws Exceptio
...
+    // The number of mapIndexKeys must be equal to the number of different keys
+    // that appear in entries of the region:
+    // "IBM", "ERICSSON", "HP", "SUN", null
+    assertThat(mapIndexKeys).isEqualTo(5);
+    // The number of values must be equal to the number of values the
+    // positions map takes for each key (null not included)
+    // for each entry in the region:
+    // "something", null, "nothing", "more", "empty", "hey", "more", "tip"
+    assertThat(values).isEqualTo(8);
```
DonalEvans, line 454: This comment is a little confusing in terms of what "null not included" is referring to, since the list of values includes null.

geode-core/src/integrationTest/java/org/apache/geode/cache/query/partitioned/PRIndexStatisticsJUnitTest.java
```diff
@@ -273,8 +273,8 @@ public void testStatsForCompactMapRangeIndex() throws Exception {
       query.execute();
     }
 
-    // Both RangeIndex should be used
-    assertEquals(100 /* Execution time */, keyIndexStats.getTotalUses());
+    // Index should not be used
```
DonalEvans, line 276: Could this comment (and others in this class) be expanded slightly to explain why the index should not be used here? I am assuming it's because the query uses `!=`, but it would be good to make it explicit.

DonalEvans (changes requested):

DonalEvans (approved):
Approved with the caveat that someone with deeper knowledge of queries and indexes should sign off on this, as it represents a potentially significant change in behaviour.

agingade:
@albertogpz 
Thanks for your contributions to make the query engine robust. Sorry for delay in responding...

Here is what I believe should be happening.
- The result for a query should be consistent in both using index or non-index case.

- The query engine returns UNDEFINED when it is unable to find the next level field. 
E.g: if address.city and address is null (This is documented). 
This is not same when you are looking for a non existing "key" in the map; UNDEFINED needs to be returned when positions is null and query is trying to access field from it. 
E.g.: positions['*']  should be returning UNDEFINED. 

- Query engine supports heterogenous objects stored in a region. 
E.g: Employee or Customer. 
Inline with supporting this, its designed/architected such that if a field is not found in the object it will be ignored. 
E.g query with employeeID is not going to return customer objects unless it has that field. 

- To be inline with the above design (query expectation), when a map field is not present available it should ignore that entry/object from adding to the result.
E.g. if positions['SUN'] if SUN key is not present query should ignore that object. 
This is also different from null check.
If there is a SUN key with null value it should be returned for queries looking for null value. And non null check will return if the key is there and its value is not null.
Try the query with non map field and the behavior should be same.

Please let me know if you have any questions on the expected behavior.

The overall behavior of the query on map should be in-line with non-map fields. 

The usage of index should not be avoided; unless the results are inconsistent with non-index query results. Instead of avoiding/blocking the use of index, it will be good to address the issues with indexed queries and make the behavior consistent.

albertogpz:
@agingade 
Thanks for your comments. Please, let me answer inline to your questions.

> @albertogpz
> Thanks for your contributions to make the query engine robust. Sorry for delay in responding...
> 
> Here is what I believe should be happening.
> 
>     * The result for a query should be consistent in both using index or non-index case.
> 
This is what I tried to fix with this PR and the test new cases now show that results are the same with and without indexes while prior to the PR, they returned different results.

>     * The query engine returns UNDEFINED when it is unable to find the next level field.
>       E.g: if address.city and address is null (This is documented).
>       This is not same when you are looking for a non existing "key" in the map; UNDEFINED needs to be returned when positions is null and query is trying to access field from it.
>       E.g.: positions['*']  should be returning UNDEFINED.
So, you mean if positions is not null but it only contains a mapping for the 'SUN" key, positions["ERIC"] should return null instead of UNDEFINED?

> 
>     * Query engine supports heterogenous objects stored in a region.
>       E.g: Employee or Customer.
>       Inline with supporting this, its designed/architected such that if a field is not found in the object it will be ignored.
>       E.g query with employeeID is not going to return customer objects unless it has that field.
> 
ok. That should not have been changed.

>     * To be inline with the above design (query expectation), when a map field is not present available it should ignore that entry/object from adding to the result.
>       E.g. if positions['SUN'] if SUN key is not present query should ignore that object.
>       This is also different from null check.
>       If there is a SUN key with null value it should be returned for queries looking for null value. And non null check will return if the key is there and its value is not null.
Currently if you have an entry for which positions has the following mappings: {"SUN" => null} and another entry for which positions has the following mapping: {"ERICSSON" => "3"} querying for positions["SUN"] = null will return both entries.
With my PR, only the first one would be returned. What should be the right behavior?


>       Try the query with non map field and the behavior should be same.
> 
> 
> Please let me know if you have any questions on the expected behavior.
> 
> The overall behavior of the query on map should be in-line with non-map fields.
> 
> The usage of index should not be avoided; unless the results are inconsistent with non-index query results. Instead of avoiding/blocking the use of index, it will be good to address the issues with indexed queries and make the behavior consistent.
> 
I tried to support the use of indexes in all queries with my PR but I found no way to do it with != queries using an index of type: positions[*] (index on all keys) or positions["SUN", "ERICSSON"].
For the second type of index, I have an alternative solution in a draft PR that will allow the use of the index although it is more costly in memory (see https://github.com/apache/geode/pull/6238)

> Can you confirm above requirements are met in this PR.

albertogpz:
> We had a small impromtu discussion with the query engine team and some research on how NoSQL treats null (or nil in some worlds)
> 
>     * in MongoDB we founnd that
>       `The { item : null } query matches documents that either contain the item field whose value is null or that do not contain the item field.`
> 
>     * in redis nil is returned if the field does not exist.
> 
> 
> Hence we think that the current behavior i.e. querying without indexes is the right result and we should try to make the indexes return the same result.

@nabarunnag Thanks for your comments. If @agingade agrees (I will check his answers after having provided mine to his comments) I will look into this possibility.

albertogpz:
> > We had a small impromtu discussion with the query engine team and some research on how NoSQL treats null (or nil in some worlds)
> > ```
> > * in MongoDB we founnd that
> >   `The { item : null } query matches documents that either contain the item field whose value is null or that do not contain the item field.`
> > 
> > * in redis nil is returned if the field does not exist.
> > ```
> > 
> > 
> > Hence we think that the current behavior i.e. querying without indexes is the right result and we should try to make the indexes return the same result.
> 
> @nabarunnag Thanks for your comments. If @agingade agrees (I will check his answers after having provided mine to his comments) I will look into this possibility.

@nabarunnag and @agingade : I have created a new pull request with a solution in which I keep the current behavior of the queries without indexes: https://github.com/apache/geode/pull/6279
Please, dismiss this PR and check the new one.
I have not been able to find a complete solution for the "allkeys" type of index (e.g. positions[*]) and with the solution, queries with this type of index and queries of type "!=" or comparing to null will not use the index even if configured. This is a limitation that could be tackled in a future PR.

PR #6219, opened 2021-03-30
a first whack at ByteBufferSharing to fix TLS handshake corruption

Bill:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/tcp/Connection.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/tcp/Connection.java
@@ -82,6 +82,7 @@ import org.apache.geode.internal.monitoring.ThreadsMonitoring;
 import org.apache.geode.internal.monitoring.executor.AbstractExecutor;
 import org.apache.geode.internal.net.BufferPool;
 import org.apache.geode.internal.net.ByteBufferSharing;
+import org.apache.geode.internal.net.ByteBufferSharingImpl;
 import org.apache.geode.internal.net.NioFilter;
 import org.apache.geode.internal.net.NioPlainEngine;
 import org.apache.geode.internal.net.SocketCreator;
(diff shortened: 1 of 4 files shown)


geode-core/src/main/java/org/apache/geode/internal/tcp/Connection.java
```diff
@@ -2643,17 +2633,21 @@ void writeFully(SocketChannel channel, ByteBuffer buffer, boolean forceAsync,
...
+      allocSize = owner.getConduit().tcpBufferSize;
     }
-    return inputBuffer;
+    inputSharing = new ByteBufferSharingImpl(
+        getBufferPool().acquireDirectReceiveBuffer(allocSize),
+        BufferPool.BufferType.TRACKED_RECEIVER,
+        getBufferPool());
+  }
+
+  private ByteBufferSharing shareInputBuffer() throws IOException {
```
bschuchardt, line 2649: since you're using inputSharing.close() everywhere you could similarly use inputSharing.close() instead of putting it in this method.  That might be easier to follow.

bschuchardt (commented):
I think this is a good solution

geode-core/src/main/java/org/apache/geode/internal/tcp/Connection.java
```diff
@@ -1502,25 +1503,15 @@ public void run() {
...
-  }
 
-  private void releaseInputBuffer() {
-    synchronized (inputBufferLock) {
-      ByteBuffer tmp = inputBuffer;
-      if (tmp != null) {
-        inputBuffer = null;
-        getBufferPool().releaseReceiveBuffer(tmp);
-      }
+      inputSharing.close();
```
bschuchardt, line 1514: I'm wondering if this method needs to claim its own reference to the buffer to balance this close()

geode-core/src/main/java/org/apache/geode/internal/tcp/Connection.java
Bill: yes!

Bill:
An alternative PR, using this same approach, has been merged to `develop` so I'm closing this one.

PR #6225, opened 2021-03-30
GEODE-9101: fix VisibleNode attribute in MemberMXBean

mkevo:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/distributed/internal/ClusterDistributionManager.java
+++ b/geode-core/src/main/java/org/apache/geode/distributed/internal/ClusterDistributionManager.java
@@ -350,13 +350,13 @@ public class ClusterDistributionManager implements DistributionManager {
               }
             }
           }
+        } else {
+          distributionManager.addNewMember(id); // add ourselves
         }
-        distributionManager.addNewMember(id); // add ourselves
       }
 
       // Send out a StartupMessage to the other members.
       StartupOperation op = new StartupOperation(distributionManager, transport);
-
       try {
         if (!distributionManager.sendStartupMessage(op)) {
           // Well we didn't hear back from anyone else. We assume that
@@ -1793,9 +1793,7 @@ public class ClusterDistributionManager implements DistributionManager {
    */
   private void handleManagerStartup(InternalDistributedMember theId) {
     // Note test is under membersLock
-    if (!this.getId().getName().equals(theId.getName())) {
-      stats.incNodes(1);
-    }
+    stats.incNodes(1);
     addMemberEvent(new MemberJoinedEvent(theId));
   }
 
--- a/geode-core/src/main/java/org/apache/geode/distributed/internal/ClusterDistributionManager.java
+++ b/geode-core/src/main/java/org/apache/geode/distributed/internal/ClusterDistributionManager.java
@@ -1734,7 +1734,7 @@ public class ClusterDistributionManager implements DistributionManager {
   public void removeUnfinishedStartup(InternalDistributedMember m, boolean departed) {
     synchronized (unfinishedStartupsLock) {
       if (logger.isDebugEnabled()) {
-        logger.debug("removeUnfinishedStartup for {} wtih {}", m, unfinishedStartups);
+        logger.debug("removeUnfinishedStartup for {} with {}", m, unfinishedStartups);
       }
       if (unfinishedStartups == null)
         return; // not yet done with startup
@@ -1793,7 +1793,7 @@ public class ClusterDistributionManager implements DistributionManager {
    */
   private void handleManagerStartup(InternalDistributedMember theId) {
     // Note test is under membersLock
-    if (theId.getVmKind() != ClusterDistributionManager.LOCATOR_DM_TYPE) {
+    if (!this.getId().getName().equals(theId.getName())) {
       stats.incNodes(1);
     }
     addMemberEvent(new MemberJoinedEvent(theId));
@@ -1881,9 +1881,8 @@ public class ClusterDistributionManager implements DistributionManager {
     removeHostedLocators(theId);
     redundancyZones.remove(theId);
 
-    if (theId.getVmKind() != ClusterDistributionManager.LOCATOR_DM_TYPE) {
-      stats.incNodes(-1);
-    }
+    stats.incNodes(-1);
+
     String msg;
     if (memberCrashed && !shouldInhibitMembershipWarnings()) {
       msg =
--- a/geode-core/src/distributedTest/java/org/apache/geode/management/CacheManagementDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/management/CacheManagementDUnitTest.java
@@ -302,6 +302,18 @@ public class CacheManagementDUnitTest implements Serializable {
     });
   }
 
+  @Test
+  public void testVisibleNodes() throws Exception {
+    this.managementTestRule.createMember(this.memberVMs[0]);
+    this.managementTestRule.createMember(this.memberVMs[1]);
+
+    this.memberVMs[1].invoke("visibleNodes", () -> {
+      ManagementService service = this.managementTestRule.getExistingManagementService();
+      MemberMXBean memberMXBean = service.getMemberMXBean();
+      assertThat(memberMXBean.getVisibleNodes()).isEqualTo(2);
+    });
+  }
+
   private void verifyQueryMBeans(final VM managerVM) {
     managerVM.invoke("validateQueryMBeans", () -> {
       SystemManagementService service = this.managementTestRule.getSystemManagementService();
--- a/geode-core/src/distributedTest/java/org/apache/geode/management/CacheManagementDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/management/CacheManagementDUnitTest.java
@@ -98,7 +98,7 @@ public class CacheManagementDUnitTest implements Serializable {
   }
 
   @Test
-  public void testGemFireConfigData() throws Exception {
+  public void testGemFireConfigData() {
     this.managementTestRule.createMembers();
     this.managementTestRule.createManagers();
 
@@ -152,7 +152,7 @@ public class CacheManagementDUnitTest implements Serializable {
    * Invoke remote operations on MemberMBean
    */
   @Test
-  public void testMemberMBeanOpsRemote() throws Exception {
+  public void testMemberMBeanOpsRemote() {
     this.managementTestRule.createMembers();
     this.managementTestRule.createManagers();
     this.managerVM.invoke(() -> invokeRemoteMemberMXBeanOps());
@@ -162,7 +162,7 @@ public class CacheManagementDUnitTest implements Serializable {
    * Creates and starts a managerVM. Multiple Managers
    */
   @Test
-  public void testManager() throws Exception {
+  public void testManager() {
     this.managementTestRule.createMember(this.memberVMs[0]);
     this.managementTestRule.createMember(this.memberVMs[1]);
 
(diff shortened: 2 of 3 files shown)


geode-core/src/main/java/org/apache/geode/distributed/internal/ClusterDistributionManager.java
```diff
@@ -1881,9 +1881,8 @@ public void handleManagerDeparture(InternalDistributedMember theId, boolean memb
     removeHostedLocators(theId);
     redundancyZones.remove(theId);
 
-    if (theId.getVmKind() != ClusterDistributionManager.LOCATOR_DM_TYPE) {
-      stats.incNodes(-1);
-    }
+    stats.incNodes(-1);
+
```
kirklund, line 1885: I think you're trying to change "nodes" to show the number of "servers" in the cluster but it's a different concept. "nodes" is a statistic in DistributionStats that reflects the number of peer-to-peer processes in the cluster, and we really need this statistic to remain as is.

I recommend adding one or more new attribute(s) to MemberMXBean or DistributedSystemMXBean which reflect the number of "non-locator members" or "servers". There are several related operations on DistributedSystemMXBean such listLocators(), listMembers(), listServers(). All three of these are "nodes".

listMembers() should show you all of the locators and servers (ie, any peer-to-peer process) in the cluster.

listLocators() should should you all of the locators only.

listServers() should show you all of the servers only.

There is a subtle difference between Attributes and Operations in JMX. You probably want an Attribute in this case so that it can be monitored with JMX APIs. But I'll leave further planning to you.

kirklund (changes requested):

geode-core/src/main/java/org/apache/geode/distributed/internal/ClusterDistributionManager.java
mkevo: I don't want to change "nodes" to show the number of "servers".
How it works by now?
If we have 2 locators and 2 servers the VisibleNodes attribute has the following values:
```
Locator 1: 2 (count Server 1 and Server 2)
Locator 2: 2 (count Server 1 and Server 2)
Server 1: 3 (count Server 2, and twice Server 1)
Server 2: 3 (count  Server 1, and twice Server 2)
```

With my changes it will count every member without itself:
```
Locator 1: 3 (count Locator 2, Server 1, and Server 2)
Locator 2: 3 (count Locator 1, Server 1, and Server 2)
Server 1: 3 (count Locator 1, Locator 2, and Server 2)
Server 2: 3 (count Locator 1, Locator 2, and Server 1)
```
What do you mean by the number of "peer-to-peer" processes in the cluster? 
Is that include both locators and servers?

jinmeiliao: From the original code, it looks like it doesn't think a locator should be included in the count, so, shouldn't the correct count should be:

```
Locator 1: 2 (count Server 1 and Server 2) 
Locator 2: 2 (count Server 1 and Server 2)
Server 1: 2 (count Server 1 and Server 2) 
Server 2: 2 (count Server 1 and Server 2)
```
instead?

mkevo: In the description of this attribute state the following:
`The current number of nodes in this distributed system visible to this member.`

What are nodes there, both locators and servers or just servers?
Also, should the current server be visible to itself?

If we go with @jinmeiliao proposal we should change the description to be more clear.
`The current number of servers in this distributed system visible to this member.`

kirklund: I understand the purpose of the PR better now. However, the PR is still trying to redefine the "nodes" statistic in DistributionStats which is not correct. You should instead add a new attribute to MemberMXBean that returns the count of OTHER members in the cluster.

Here is the definition of the "nodes" statistic:
```
    final String nodesDesc = "The current number of nodes in this distributed system.";
```
So, if you have 2 servers and 2 locators, "nodes" should be 4 within all 4 JVMs.

The earliest version of GemFire had locators which were not members (ie not nodes), but locators now are definitely members (ie nodes). The next type of locators in GemFire were members (ie nodes) but without caches. As of GemFire 7.0, all locators have a cache. As of Geode 1.0, all locators are members with caches. The methods to create that earliest type of locator still exists but are deprecated... so deprecated internal methods is now the only way to create the thin locators that are not peer-to-peer members of the cluster.

Changing the attribute of an MXBean is _different_ from changing the meaning of the statistic. The statistic is not supposed to ignore locators.

jinmeiliao (approved):

mkevo:
@kirklund , can you re-review it, please?

geode-core/src/main/java/org/apache/geode/distributed/internal/ClusterDistributionManager.java
```diff
@@ -350,13 +350,13 @@ static ClusterDistributionManager create(InternalDistributedSystem system,
               }
             }
           }
+        } else {
+          distributionManager.addNewMember(id); // add ourselves
```
kirklund, line 354: This block of code needs to addNewMember even when the name is "". Why does it need to move it into an else-block?

If you start a Geode process without using `start locator` or `start server`, you can do so without a `name` but it's still a new member.

geode-core/src/main/java/org/apache/geode/distributed/internal/ClusterDistributionManager.java
```diff
@@ -1793,9 +1793,7 @@ private String prettifyReason(String r) {
    */
   private void handleManagerStartup(InternalDistributedMember theId) {
     // Note test is under membersLock
-    if (theId.getVmKind() != ClusterDistributionManager.LOCATOR_DM_TYPE) {
-      stats.incNodes(1);
-    }
+    stats.incNodes(1);
```
kirklund, line 1796: I think @Bill @bschuchardt or @dschneider-pivotal should review this change instead of me.

geode-core/src/main/java/org/apache/geode/distributed/internal/ClusterDistributionManager.java
kirklund: I'm marking this as resolved because I think developers who have worked on geode-membership should review it instead of me.

geode-core/src/main/java/org/apache/geode/distributed/internal/ClusterDistributionManager.java
mkevo: In the method, _create_(ClusterDistributionManager.java) it adds itself twice(ClusterDistributionManager constructor and there again). So it increments this attribute for both _addNewMember()_ call.

echobravopapa:
@kirklund I noticed your comments on the JIRA ticket, do you still think this warrants dev list discussion and possibly even an RFC?   @mkevo did you see his comments on the ticket?

mkevo:
Hi @echobravopapa, I think that for adding a new attribute we need to start the discussion on the dev list or write an RFC. But for this case, we just changed the existing attribute to print the correct values. 
Some time ago, when this attribute was added, the locators weren't members/nodes, then it is changed and locators became members/nodes without a cache. From GemFire 7.0, all locators also have a cache and they are members/nodes and we should change this attribute to get the correct value for all visible nodes.

echobravopapa:
@mkevo would you mind pushing an empty commit so the PR pipeline will re-run, thanks.

kirklund (commented):
This PR is actually redefining what "node" means in DistributionStats in order to "correct" the value of "nodes" on the MemberMXBean. 

We should not change what "node" means in DistributionStats. The current value of that stat is correct for DistributionStats.

You should add a new attribute to MemberMXBean such as "members" or "servers" or "locators" or all three. But changing the underlying value of "node" is not the correct way to fix the problem in MemberMXBean. The authors of MemberMXBean should not have used "node" from the stats -- they should have used the membership layer to count the number of members, server, and/or locators.

I recommended a dev-list discussion because this change is contentious (I disagree with it) and if you add new attributes to MemberMXBean that is adding new User APIs for which we have always said needs RFC or dev-list discussion.

onichols-pivotal:
looks like this may not be moving forward, can it be closed?

mkevo:
> looks like this may not be moving forward, can it be closed?

I still waiting for an answer on the geode dev mail list. So this cannot be closed.

kirklund:
> > looks like this may not be moving forward, can it be closed?
> 
> I still waiting for an answer on the geode dev mail list. So this cannot be closed.

@mkevo since this is dragging out and you didn't get an answer on dev, I would write up a simple proposal on the wiki (similar to the other proposals) and start a review of that on the dev list.

mkevo:
> > > looks like this may not be moving forward, can it be closed?
> > 
> > 
> > I still waiting for an answer on the geode dev mail list. So this cannot be closed.
> 
> @mkevo since this is dragging out and you didn't get an answer on dev, I would write up a simple proposal on the wiki (similar to the other proposals) and start a review of that on the dev list.

Thanks @kirklund!

mkevo:
> > > looks like this may not be moving forward, can it be closed?
> > 
> > 
> > I still waiting for an answer on the geode dev mail list. So this cannot be closed.
> 
> @mkevo since this is dragging out and you didn't get an answer on dev, I would write up a simple proposal on the wiki (similar to the other proposals) and start a review of that on the dev list.

Hi @kirklund,
What is the status of this activity?

mkevo:
> > > > looks like this may not be moving forward, can it be closed?
> > > 
> > > 
> > > I still waiting for an answer on the geode dev mail list. So this cannot be closed.
> > 
> > 
> > @mkevo since this is dragging out and you didn't get an answer on dev, I would write up a simple proposal on the wiki (similar to the other proposals) and start a review of that on the dev list.
> 
> Hi @kirklund, What is the status of this activity?

Hi @kirklund, any news on this?

mivanac (approved):

geode-core/src/main/java/org/apache/geode/distributed/internal/ClusterDistributionManager.java
kirklund: Why is it ok to remove the if-condition from here?

geode-core/src/main/java/org/apache/geode/distributed/internal/ClusterDistributionManager.java
kirklund: We probably need some unit tests for ClusterDistributionManager which would be closer to the actual code than JMX.

I can't sign off on this change for now though because as far as I know the value of `nodes` is correct. I still think you should convince dev-list that the value of `nodes` is incorrect or you should leave the value alone.

geode-core/src/main/java/org/apache/geode/distributed/internal/ClusterDistributionManager.java
mkevo: The stats should count both locators and servers, as both of them are nodes. With if-condition locators wouldn't count.

geode-core/src/main/java/org/apache/geode/distributed/internal/ClusterDistributionManager.java
mkevo: Hi @kirklund ,
I send a mail to the dev list to discuss what is the right way to fix this. No response.
I wrote an RFC with adding a new attribute. The only one response, to fix this ticket instead of adding a new attribute.

kirklund (commented):
@mario Please see GEODE-10358 and the test I wrote in PR #7765 to understand why removing `if (theId.getVmKind() != ClusterDistributionManager.LOCATOR_DM_TYPE) {` will result in redefining the `DistributionStats` `nodes` stat which is incorrect.

I'm going to -1 this PR and the RFC. Thank you, Kirk

PR #6246, opened 2021-03-31
GEODE-9060: Remove the member from a copy of replicates as GII candid…

gesterzhou:
…ate if

it's not part of the same distributed system, but leave original replicates
unchanged.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/persistence/PersistentRecoveryOrderDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/persistence/PersistentRecoveryOrderDUnitTest.java
@@ -877,7 +877,39 @@ public class PersistentRecoveryOrderDUnitTest extends CacheTestCase {
           createReplicateRegion(regionName, getDiskDirs(getVMId()));
         });
         assertThat(thrown).isInstanceOf(ConflictingPersistentDataException.class);
+        assertThat(thrown.getMessage())
+            .contains("was not part of the same distributed system as the local data");
       }
+    });
+  }
+
+  @Test
+  public void testRecoverableSplitBrain() {
+    vm2.invoke(() -> {
+      createReplicateRegion(regionName, getDiskDirs(getVMId()));
+    });
+    vm0.invoke(() -> {
+      createReplicateRegion(regionName, getDiskDirs(getVMId()));
+      putEntry("A", "B");
+      getCache().getRegion(regionName).close();
+    });
+
+    vm1.invoke(() -> {
+      createReplicateRegion(regionName, getDiskDirs(getVMId()));
+      validateEntry("A", "B");
+      updateEntry("A", "C");
+      getCache().getRegion(regionName).close();
+    });
+
+    // VM0 doesn't know that VM1 ever existed so it will start up.
+    vm0.invoke(() -> {
+      createReplicateRegion(regionName, getDiskDirs(getVMId()));
+      validateEntry("A", "C");
+    });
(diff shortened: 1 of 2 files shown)


geode-core/src/distributedTest/java/org/apache/geode/internal/cache/persistence/PersistentRecoveryOrderDUnitTest.java
```diff
@@ -877,10 +877,42 @@ public void testSplitBrain() {
           createReplicateRegion(regionName, getDiskDirs(getVMId()));
         });
         assertThat(thrown).isInstanceOf(ConflictingPersistentDataException.class);
+        assertThat(thrown.getMessage())
+            .contains("was not part of the same distributed system as the local data");
```
kirklund, line 881: Typically you should just use `thrown` as the subject for the assertion rather than `thrown.getMessage()`:
```
assertThat(thrown).hasMessageContaining("was not part of the same distributed system as the local data");
```
That way you can chain together multiple details:
```
assertThat(thrown)
    .isInstanceOf(ConflictingPersistentDataException.class)
    .hasMessageContaining("was not part of the same distributed system as the local data");
```
...and you could even check `hasCauseInstanceOf` or other options.

geode-core/src/main/java/org/apache/geode/internal/cache/persistence/PersistenceAdvisorImpl.java
```diff
@@ -510,18 +510,19 @@ public PersistentMembershipView getMembershipView() {
   public boolean checkMyStateOnMembers(Set<InternalDistributedMember> replicates)
       throws ReplyException {
     PersistentStateQueryResults remoteStates = getMyStateOnMembers(replicates);
+    Set<InternalDistributedMember> copyOfReplicates = new HashSet<>();
+    copyOfReplicates.addAll(replicates);
```
kirklund, line 514: You can shrink this down to one statement:
```
Set<InternalDistributedMember> copyOfReplicates = new HashSet<>(replicates);
```

kirklund (approved):
Two suggestions and Approval.

geode-core/src/main/java/org/apache/geode/internal/cache/persistence/PersistenceAdvisorImpl.java
gesterzhou: all fixed

kirklund (approved):

geode-core/src/main/java/org/apache/geode/internal/cache/persistence/PersistenceAdvisorImpl.java
```diff
@@ -510,18 +510,18 @@ public PersistentMembershipView getMembershipView() {
   public boolean checkMyStateOnMembers(Set<InternalDistributedMember> replicates)
       throws ReplyException {
     PersistentStateQueryResults remoteStates = getMyStateOnMembers(replicates);
+    Set<InternalDistributedMember> copyOfReplicates = new HashSet<>(replicates);
```
dschneider-pivotal, line 513: Instead of always having this code create a copy every time it is called, start the copyofReplicates out as null. Then, in the exceptional case when we need to remove from it (line 540), you can ask if it is null and then make it a copy of  replicates.

dschneider-pivotal (changes requested):
This looks correct. I just am requesting a small change to make it perform better

dschneider-pivotal (approved):

PR #6273, opened 2021-04-06
GEODE-9125: fix GetMemberInformationFunction to return correct status…

jinmeiliao:
… when there is no default cache server


aaronlindsey (approved):
Makes sense and LGTM 👍

onichols-pivotal (approved):

kirklund (approved):

PR #6296, opened 2021-04-08
GEODE-9136: make RedisData implement Sizeable

nonbinaryprogrammer:
- make RedisData implement Sizeable
- reimplement delta force recalculate size
- add tests for hash add/remove/edit wo size change
- add tests for string add/append/remove/edit wo size change
- add tests for set add/remove/edit wo size change
- add tests to verify our math is right
- keep track of the data size as part of RedisHash, RedisString,
  and RedisSet

Signed-off-by: Ray Ingles <ringles@vmware.com>

### For all changes:

### Note:

### DIFF
--- a/geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/PartitionedRegionStatsUpdateTest.java
+++ b/geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/PartitionedRegionStatsUpdateTest.java
@@ -29,7 +29,6 @@ import org.junit.Ignore;
 import org.junit.Test;
 import redis.clients.jedis.Jedis;
 
-import org.apache.geode.internal.size.ReflectionObjectSizer;
 import org.apache.geode.test.awaitility.GeodeAwaitility;
 import org.apache.geode.test.dunit.rules.MemberVM;
 import org.apache.geode.test.dunit.rules.RedisClusterStartupRule;
@@ -52,9 +51,6 @@ public class PartitionedRegionStatsUpdateTest {
   public static final String HASH_KEY = "hash key";
   public static final String LONG_APPEND_VALUE = String.valueOf(Integer.MAX_VALUE);
   public static final String FIELD = "field";
-  private static final int SET_SIZE = 1000;
-
-  private final ReflectionObjectSizer ros = ReflectionObjectSizer.getInstance();
 
   @BeforeClass
   public static void classSetup() {
--- a/geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/PartitionedRegionStatsUpdateTest.java
+++ b/geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/PartitionedRegionStatsUpdateTest.java
@@ -337,7 +337,7 @@ public class PartitionedRegionStatsUpdateTest {
   }
 
   @Test
-  @Ignore
+  @Ignore("confirm that bucket size stats are being calculated correctly before enabling")
   public void should_showMembersAgreeUponUsedSetMemory_afterDeltaPropagation() {
     jedis1.sadd(SET_KEY, "other"); // two sadds are required to force
     jedis1.sadd(SET_KEY, "value"); // deserialization on both servers
--- a/geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/PartitionedRegionStatsUpdateTest.java
+++ b/geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/PartitionedRegionStatsUpdateTest.java
@@ -381,7 +381,7 @@ public class PartitionedRegionStatsUpdateTest {
     assertThat(finalDataStoreBytesInUse).isEqualTo(initialDataStoreBytesInUse);
   }
 
-  /******* confirm our math is right *******/
+  /******* check DatastoreBytesInUse using reflection *******/
 
   @Test
   public void string_bytesInUse_shouldReflectActualSizeOfDataInRegion() {
@@ -495,6 +495,8 @@ public class PartitionedRegionStatsUpdateTest {
 
     assertThat(actual).isCloseTo(expected, offset);
   }
+
+  /******* helper methods *******/
 
   private List<String> makeMemberList(int setSize, String baseString) {
     List<String> members = new ArrayList<>();
--- a/geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/PartitionedRegionStatsUpdateTest.java
+++ b/geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/PartitionedRegionStatsUpdateTest.java
@@ -103,16 +103,20 @@ public class PartitionedRegionStatsUpdateTest {
 
   @Test
   public void should_showDecreaseInDatastoreBytesInUse_givenStringValueDeleted() {
-    jedis1.set(STRING_KEY, "value");
-
     long initialDataStoreBytesInUse =
         clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server1);
+
+    jedis1.set(STRING_KEY, "value");
+
+    long intermediateDataStoreBytesInUse =
+        clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server1);
+    assertThat(intermediateDataStoreBytesInUse).isGreaterThan(initialDataStoreBytesInUse);
 
     jedis1.del(STRING_KEY);
 
     long finalDataStoreBytesInUse = clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server1);
 
-    assertThat(finalDataStoreBytesInUse).isLessThan(initialDataStoreBytesInUse);
+    assertThat(finalDataStoreBytesInUse).isEqualTo(initialDataStoreBytesInUse);
   }
 
   @Test
@@ -246,22 +250,6 @@ public class PartitionedRegionStatsUpdateTest {
   }
 
   @Test
-  public void should_ShowNoIncreaseInDatastoreBytesInUse_givenHashValueSizeDoesNotIncrease() {
-    jedis1.hset(HASH_KEY, FIELD, "value");
-
-    long initialDataStoreBytesInUse =
-        clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server1);
-
-    for (int i = 0; i < 1000; i++) {
-      jedis1.hsetnx(HASH_KEY, FIELD, LONG_APPEND_VALUE);
-    }
-
-    long finalDataStoreBytesInUse = clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server1);
-
-    assertThat(finalDataStoreBytesInUse).isEqualTo(initialDataStoreBytesInUse);
-  }
-
-  @Test
   public void should_showDecreaseInDatastoreBytesInUse_givenHashValueDeleted() {
     long initialDataStoreBytesInUse =
         clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server1);
@@ -331,7 +319,7 @@ public class PartitionedRegionStatsUpdateTest {
     assertThat(finalDataStoreBytesInUse).isEqualTo(initialDataStoreBytesInUse);
   }
 
-  // confirm that the other member agrees upon size
+  /******* confirm that the other member agrees upon size *******/
 
   @Test
   public void should_showMembersAgreeUponUsedHashMemory_afterDeltaPropagation() {
@@ -393,7 +381,7 @@ public class PartitionedRegionStatsUpdateTest {
     assertThat(finalDataStoreBytesInUse).isEqualTo(initialDataStoreBytesInUse);
   }
 
-  // confirm our math is right
+  /******* confirm our math is right *******/
 
   @Test
   public void string_bytesInUse_shouldReflectActualSizeOfDataInRegion() {
--- a/geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/PartitionedRegionStatsUpdateTest.java
+++ b/geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/PartitionedRegionStatsUpdateTest.java
@@ -32,6 +32,7 @@ import org.assertj.core.data.Offset;
 import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.ClassRule;
+import org.junit.Ignore;
 import org.junit.Test;
 import redis.clients.jedis.Jedis;
 
@@ -372,23 +373,24 @@ public class PartitionedRegionStatsUpdateTest {
   }
 
   @Test
+  @Ignore("confirm that bucket size stats are being calculated correctly before enabling")
   public void should_showMembersAgreeUponUsedStringMemory_afterDeltaPropagation() {
-    jedis1.set(STRING_KEY, "eulav"); // two sets are required to force
-    jedis2.set(STRING_KEY, "value"); // deserialization on both servers
+    jedis1.set(STRING_KEY, "12345"); // two sets are required to force
+    jedis1.set(STRING_KEY, "value"); // deserialization on both servers
     // otherwise primary/secondary can disagree on size, and which server is primary varies
 
     long initialDataStoreBytesInUse =
         clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server1);
 
     for (int i = 0; i < 10; i++) {
-      jedis1.set(STRING_KEY, "value");
+      jedis1.append(STRING_KEY, "a");
     }
 
     assertThat(jedis1.exists(STRING_KEY)).isTrue();
     assertThat(jedis2.exists(STRING_KEY)).isTrue();
 
-    assertThat(jedis1.get(STRING_KEY)).isEqualTo("value");
-    assertThat(jedis2.get(STRING_KEY)).isEqualTo("value");
+    assertThat(jedis1.get(STRING_KEY)).isEqualTo("valueaaaaaaaaaa");
+    assertThat(jedis2.get(STRING_KEY)).isEqualTo("valueaaaaaaaaaa");
 
     long server1FinalDataStoreBytesInUse =
         clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server1);
@@ -400,138 +402,5 @@ public class PartitionedRegionStatsUpdateTest {
 
     assertThat(server2FinalDataStoreBytesInUse)
         .isEqualTo(initialDataStoreBytesInUse);
-  }
-
-  /******* check DatastoreBytesInUse using reflection *******/
-
-  @Test
-  public void string_bytesInUse_shouldReflectActualSizeOfDataInRegion() {
-    String value = "value";
-
-    jedis1.set("key2", value);
-
-    Long regionOverhead = clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server1);
-
-    jedis1.set(STRING_KEY, value);
-
-    for (int i = 0; i < 100; i++) {
-      jedis1.append(STRING_KEY, LONG_APPEND_VALUE);
-      value += LONG_APPEND_VALUE;
-    }
-
-    Long actual = clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server1);
-    Long expected = ros.sizeof(value.getBytes()) + regionOverhead;
-    Offset<Long> offset = Offset.offset(Math.round(expected * 0.05));
-
-    assertThat(actual).isCloseTo(expected, offset);
-  }
-
-  @Test
-  @SuppressWarnings("unchecked")
-  public void set_bytesInUse_shouldReflectActualSizeOfDataInRegion() {
-    String baseValue = "string";
-
-    Set values = new HashSet<ByteArrayWrapper>();
-    values.add(new ByteArrayWrapper(baseValue.getBytes()));
-    jedis1.sadd(SET_KEY, baseValue);
-
-    for (int i = 0; i < 1000; i++) {
-      jedis1.sadd(SET_KEY, baseValue + i);
-      values.add(new ByteArrayWrapper((baseValue + i).getBytes()));
-    }
-
-    Long actual = clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server1);
-    int expected = ros.sizeof(values);
-    Offset<Long> offset = Offset.offset(Math.round(expected * 0.05));
-
-    assertThat(actual).isCloseTo(expected, offset);
-  }
-
-  @Test
-  @SuppressWarnings("unchecked")
-  public void hash_bytesInUse_shouldReflectActualSizeOfDataInRegion() {
-    String baseValue = "value";
-    String baseField = "field";
-
-    HashMap<ByteArrayWrapper, ByteArrayWrapper> values = new HashMap<>();
-    values.put(new ByteArrayWrapper(baseField.getBytes()),
-        new ByteArrayWrapper(baseValue.getBytes()));
-    jedis1.hset(HASH_KEY, baseField, baseValue);
-
-    for (int i = 0; i < 10_000; i++) {
-      jedis1.hset(HASH_KEY, baseField + i, baseValue + i);
-      values.put(new ByteArrayWrapper((baseField + i).getBytes()),
-          new ByteArrayWrapper((baseValue + i).getBytes()));
-    }
-
-    Long actual = clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server1);
-    int expected = ros.sizeof(values);
-    Offset<Long> offset = Offset.offset(Math.round(expected * 0.05));
-
-    assertThat(actual).isCloseTo(expected, offset);
-  }
-
-  @Test
-  public void should_getAccurateSize_whenUsingRedisSetConstructor() {
-    final String KEY = "key1";
-    List<String> members1 = makeMemberList(SET_SIZE, "member1-");
-
-    HashSet<ByteArrayWrapper> localHashSetCopy = new HashSet<>(
-        members1.stream().map(a -> new ByteArrayWrapper(a.getBytes()))
-            .collect(Collectors.toSet()));
-
-    jedis1.sadd(KEY, members1.toArray(new String[] {}));
-    jedis1.sadd(KEY, "forceDelta");
-
-    Long actual = clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server1);
-    int expected = ros.sizeof(localHashSetCopy);
-    Offset<Long> offset = Offset.offset(Math.round(expected * 0.05));
-
-    assertThat(actual).isCloseTo(expected, offset);
-  }
-
-  @Test
-  public void should_notCountDuplicates_whenUsingRedisSetConstructor() {
-    final String KEY = "key1";
-    final String MEMBER_BASE_STRING = "member1-";
-
-    List<String> distinctMembers = makeMemberList(SET_SIZE, MEMBER_BASE_STRING);
-    HashSet<ByteArrayWrapper> localHashSetCopy = new HashSet<>(
-        distinctMembers.stream().distinct().map(a -> new ByteArrayWrapper(a.getBytes()))
-            .collect(Collectors.toSet()));
-
-    jedis1.sadd(KEY, distinctMembers.toArray(new String[] {}));
-    assertThat(jedis1.scard(KEY)).isEqualTo(SET_SIZE);
-
-    localHashSetCopy.add(new ByteArrayWrapper(MEMBER_BASE_STRING.getBytes()));
-    assertThat(localHashSetCopy.size()).isEqualTo(SET_SIZE + 1);
-
-    List<String> duplicateMembers = makeListOfDuplicates(SET_SIZE, MEMBER_BASE_STRING);
-    jedis1.sadd(KEY, duplicateMembers.toArray(new String[] {}));
-    assertThat(jedis1.scard(KEY)).isEqualTo(SET_SIZE + 1);
-
-    Long actual = clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server1);
-    int expected = ros.sizeof(localHashSetCopy);
-    Offset<Long> offset = Offset.offset(Math.round(expected * 0.05));
-
-    assertThat(actual).isCloseTo(expected, offset);
-  }
-
-  /******* helper methods *******/
-
-  private List<String> makeMemberList(int setSize, String baseString) {
-    List<String> members = new ArrayList<>();
-    for (int i = 0; i < setSize; i++) {
-      members.add(baseString + i);
-    }
-    return members;
-  }
-
-  private List<String> makeListOfDuplicates(int setSize, String baseString) {
-    List<String> members = new ArrayList<>();
-    for (int i = 0; i < setSize; i++) {
-      members.add(baseString);
-    }
-    return members;
   }
 }
--- a/geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/PartitionedRegionStatsUpdateTest.java
+++ b/geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/PartitionedRegionStatsUpdateTest.java
@@ -337,7 +337,6 @@ public class PartitionedRegionStatsUpdateTest {
   }
 
   @Test
-  @Ignore("confirm that bucket size stats are being calculated correctly before enabling")
   public void should_showMembersAgreeUponUsedSetMemory_afterDeltaPropagation() {
     jedis1.sadd(SET_KEY, "other"); // two sadds are required to force
     jedis1.sadd(SET_KEY, "value"); // deserialization on both servers
@@ -363,24 +362,26 @@ public class PartitionedRegionStatsUpdateTest {
   }
 
   @Test
-  @Ignore("confirm that bucket size stats are being calculated correctly before enabling")
+  @Ignore("find a way to force deserialization on both members before enabling")
   public void should_showMembersAgreeUponUsedStringMemory_afterDeltaPropagation() {
+    String value = "value";
+
     jedis1.set(STRING_KEY, "12345"); // two sets are required to force
-    jedis1.set(STRING_KEY, "value"); // deserialization on both servers
+    jedis1.set(STRING_KEY, value); // deserialization on both servers
     // otherwise primary/secondary can disagree on size, and which server is primary varies
 
     long initialDataStoreBytesInUse =
         clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server1);
 
     for (int i = 0; i < 10; i++) {
-      jedis1.append(STRING_KEY, "a");
+      jedis1.set(STRING_KEY, value);
     }
 
     assertThat(jedis1.exists(STRING_KEY)).isTrue();
     assertThat(jedis2.exists(STRING_KEY)).isTrue();
 
-    assertThat(jedis1.get(STRING_KEY)).isEqualTo("valueaaaaaaaaaa");
-    assertThat(jedis2.get(STRING_KEY)).isEqualTo("valueaaaaaaaaaa");
+    assertThat(jedis1.get(STRING_KEY)).isEqualTo(value);
+    assertThat(jedis2.get(STRING_KEY)).isEqualTo(value);
 
     long server1FinalDataStoreBytesInUse =
         clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server1);
--- a/geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/PartitionedRegionStatsUpdateTest.java
+++ b/geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/PartitionedRegionStatsUpdateTest.java
@@ -337,8 +337,9 @@ public class PartitionedRegionStatsUpdateTest {
   }
 
   @Test
+  @Ignore
   public void should_showMembersAgreeUponUsedSetMemory_afterDeltaPropagation() {
-    jedis1.sadd(SET_KEY, "initialValue"); // two sadds are required to force
+    jedis1.sadd(SET_KEY, "other"); // two sadds are required to force
     jedis1.sadd(SET_KEY, "value"); // deserialization on both servers
     // otherwise primary/secondary can disagree on size, and which server is primary varies
 
--- a/geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/PartitionedRegionStatsUpdateTest.java
+++ b/geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/PartitionedRegionStatsUpdateTest.java
@@ -374,11 +374,7 @@ public class PartitionedRegionStatsUpdateTest {
   @Test
   public void should_showMembersAgreeUponUsedStringMemory_afterDeltaPropagation() {
     jedis1.set(STRING_KEY, "eulav"); // two sets are required to force
-    System.out.println("HERE server1:" + clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server1));
-    System.out.println("HERE server2: " + clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server2));
     jedis2.set(STRING_KEY, "value"); // deserialization on both servers
-    System.out.println("HERE server1:" + clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server1));
-    System.out.println("HERE server2: " + clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server2));
     // otherwise primary/secondary can disagree on size, and which server is primary varies
 
     long initialDataStoreBytesInUse =
--- a/geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/PartitionedRegionStatsUpdateTest.java
+++ b/geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/PartitionedRegionStatsUpdateTest.java
@@ -60,7 +60,7 @@ public class PartitionedRegionStatsUpdateTest {
   public static final String FIELD = "field";
   private static final int SET_SIZE = 1000;
 
-  private ReflectionObjectSizer ros = ReflectionObjectSizer.getInstance();
+  private final ReflectionObjectSizer ros = ReflectionObjectSizer.getInstance();
 
   @BeforeClass
   public static void classSetup() {
@@ -336,14 +336,19 @@ public class PartitionedRegionStatsUpdateTest {
 
     assertThat(jedis1.hgetAll(HASH_KEY).size()).isEqualTo(1);
 
-    long finalDataStoreBytesInUse = clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server2);
+    long server2FinalDataStoreBytesInUse =
+        clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server2);
+    long server1FinalDataStoreBytesInUse =
+        clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server1);
 
-    assertThat(finalDataStoreBytesInUse).isEqualTo(initialDataStoreBytesInUse);
+    assertThat(server1FinalDataStoreBytesInUse)
+        .isEqualTo(server2FinalDataStoreBytesInUse)
+        .isEqualTo(initialDataStoreBytesInUse);
   }
 
   @Test
   public void should_showMembersAgreeUponUsedSetMemory_afterDeltaPropagation() {
-    jedis1.sadd(SET_KEY, "value"); // two sadds are required to force
+    jedis1.sadd(SET_KEY, "initialValue"); // two sadds are required to force
     jedis1.sadd(SET_KEY, "value"); // deserialization on both servers
     // otherwise primary/secondary can disagree on size, and which server is primary varies
 
@@ -354,31 +359,51 @@ public class PartitionedRegionStatsUpdateTest {
       jedis1.sadd(SET_KEY, "value");
     }
 
-    assertThat(jedis1.scard(SET_KEY)).isEqualTo(1);
+    assertThat(jedis1.scard(SET_KEY)).isEqualTo(2);
 
-    long finalDataStoreBytesInUse = clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server2);
+    long server1FinalDataStoreBytesInUse =
+        clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server1);
+    long server2FinalDataStoreBytesInUse =
+        clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server2);
 
-    assertThat(finalDataStoreBytesInUse).isEqualTo(initialDataStoreBytesInUse);
+    assertThat(server1FinalDataStoreBytesInUse)
+        .isEqualTo(server2FinalDataStoreBytesInUse)
+        .isEqualTo(initialDataStoreBytesInUse);
   }
 
   @Test
   public void should_showMembersAgreeUponUsedStringMemory_afterDeltaPropagation() {
-    jedis1.set(STRING_KEY, "value"); // two sets are required to force
-    jedis1.set(STRING_KEY, "value"); // deserialization on both servers
+    jedis1.set(STRING_KEY, "eulav"); // two sets are required to force
+    System.out.println("HERE server1:" + clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server1));
+    System.out.println("HERE server2: " + clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server2));
+    jedis2.set(STRING_KEY, "value"); // deserialization on both servers
+    System.out.println("HERE server1:" + clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server1));
+    System.out.println("HERE server2: " + clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server2));
     // otherwise primary/secondary can disagree on size, and which server is primary varies
 
     long initialDataStoreBytesInUse =
-        clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server2);
+        clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server1);
 
     for (int i = 0; i < 10; i++) {
       jedis1.set(STRING_KEY, "value");
     }
 
     assertThat(jedis1.exists(STRING_KEY)).isTrue();
+    assertThat(jedis2.exists(STRING_KEY)).isTrue();
 
-    long finalDataStoreBytesInUse = clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server2);
+    assertThat(jedis1.get(STRING_KEY)).isEqualTo("value");
+    assertThat(jedis2.get(STRING_KEY)).isEqualTo("value");
 
-    assertThat(finalDataStoreBytesInUse).isEqualTo(initialDataStoreBytesInUse);
+    long server1FinalDataStoreBytesInUse =
+        clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server1);
+    long server2FinalDataStoreBytesInUse =
+        clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server2);
+
+    assertThat(server1FinalDataStoreBytesInUse)
+        .isEqualTo(initialDataStoreBytesInUse);
+
+    assertThat(server2FinalDataStoreBytesInUse)
+        .isEqualTo(initialDataStoreBytesInUse);
   }
 
   /******* check DatastoreBytesInUse using reflection *******/
--- a/geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/PartitionedRegionStatsUpdateTest.java
+++ b/geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/PartitionedRegionStatsUpdateTest.java
@@ -268,7 +268,7 @@ public class PartitionedRegionStatsUpdateTest {
   @Test
   public void should_showNoIncreaseInDatastoreBytesInUse_givenHSetDoesNotIncreaseHashSize() {
     jedis2.hset(HASH_KEY, FIELD, "initialvalue"); // two hsets are required to force
-    jedis2.hset(HASH_KEY, FIELD, "value");        // deserialization on both servers
+    jedis2.hset(HASH_KEY, FIELD, "value"); // deserialization on both servers
 
     long initialDataStoreBytesInUse =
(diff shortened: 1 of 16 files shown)


geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/PartitionedRegionStatsUpdateTest.java
```diff
@@ -0,0 +1,512 @@
...
+      jedis1.set(STRING_KEY, "value");
+    }
+
+    long finalDataStoreBytesInUse = clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server1);
+
+    assertThat(finalDataStoreBytesInUse).isEqualTo(initialDataStoreBytesInUse);
+  }
+
+  @Test
+  public void should_showIncreaseInDatastoreBytesInUse_givenSetValueSizeIncreases() {
```
sabbey37, line 165: Not a big deal, but this should be "Increase" instead of "Increases"

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/PartitionedRegionStatsUpdateTest.java
```diff
@@ -0,0 +1,512 @@
...
+    long finalDataStoreBytesInUse = clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server1);
+
+    assertThat(finalDataStoreBytesInUse).isEqualTo(initialDataStoreBytesInUse);
+  }
+
+  @Test
+  public void should_showDecreaseInDatastoreBytesInUse_givenSetValueDeleted() {
+    long initialDataStoreBytesInUse =
+        clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server1);
+
```
sabbey37, line 200: For this test as well as the `givenXDeleted` tests below, we aren't actually testing that the bytes decreased as the test name suggests.  We are testing that they return to their initial values.  The above test (`should_showDecreaseInDatastoreBytesInUse_givenStringValueDeleted`) does test that the bytes decrease.... we could potentially do both things (get the bytes before any operation, get the bytes after an operation, and get the bytes after a del, then verify the bytes decreased and then returned to their initial value)...

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/PartitionedRegionStatsUpdateTest.java
```diff
@@ -0,0 +1,512 @@
...
+      jedis1.hsetnx(HASH_KEY, FIELD, LONG_APPEND_VALUE);
+    }
+
+    long finalDataStoreBytesInUse = clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server1);
+
+    assertThat(finalDataStoreBytesInUse).isEqualTo(initialDataStoreBytesInUse);
+  }
+
+  @Test
+  public void should_showDecreaseInDatastoreBytesInUse_givenHashValueDeleted() {
```
sabbey37, line 260: Same comment as above (regarding not actually testing what the test name suggests).

sabbey37:
I would feel more comfortable if @nabarunnag, @upthewaterspout and/or @dschneider-pivotal reviewed the PR as well with their knowledge of Sizeable and Geode core in general.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisHash.java
```diff
@@ -58,6 +59,9 @@
   private ConcurrentHashMap<UUID, List<ByteArrayWrapper>> hScanSnapShots;
   private ConcurrentHashMap<UUID, Long> hScanSnapShotCreationTimes;
   private ScheduledExecutorService HSCANSnapshotExpirationExecutor = null;
+  private static final int PER_STRING_OVERHEAD = PER_OBJECT_OVERHEAD + 46;
+  private static final int PER_HASH_OVERHEAD = PER_OBJECT_OVERHEAD + 116;
```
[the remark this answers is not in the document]
ringles: These values were derived by empirically comparing them to the measurements from the ReflectionObjectSizer. The ReflectionObjectSizer gives an exact figure, but takes far too long (so we removed it from the final code). We confirmed that these values converge to the ROS values as the number of keys of various types increases.

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/PartitionedRegionStatsUpdateTest.java
```diff
@@ -0,0 +1,512 @@
...
+
+    long finalDataStoreBytesInUse = clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server1);
+
+    assertThat(finalDataStoreBytesInUse).isEqualTo(initialDataStoreBytesInUse);
+  }
+
+  @Test
+  public void should_showNoIncreaseInDatastoreBytesInUse_givenHSetDoesNotIncreaseHashSize() {
+    jedis2.hset(HASH_KEY, FIELD, "initialvalue"); // two hsets are required to force
+    jedis2.hset(HASH_KEY, FIELD, "value"); // deserialization on both servers
```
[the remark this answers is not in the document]
ringles: We're actually _not_ checking both servers specifically. But it's a roll of the dice which server is the primary and which is the secondary. If we don't force a delta, the secondary server stores the initial put in serialized form, which has a different size from the deserialized form. Which means that 50% of the runs will fail, because they happen to connect to the secondary rather than the primary.

It's kind of hard to sum that up in variable names and/or comments. But if it's unclear to you, we'll probably need to revisit and see if we can make it clearer. The tests below that talk about "members agree" are supposed to actually check both servers.

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/PartitionedRegionStatsUpdateTest.java
```diff
@@ -0,0 +1,512 @@
...
+
+    long finalDataStoreBytesInUse = clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server1);
+
+    assertThat(finalDataStoreBytesInUse).isEqualTo(initialDataStoreBytesInUse);
+  }
+
+  // confirm that the other member agrees upon size
+
+  @Test
+  public void should_showMembersAgreeUponUsedHashMemory_afterDeltaPropagation() {
```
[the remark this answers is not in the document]
ringles: We can probably clean up the name some. The idea is that the operations in the "_afterDeltaPropagation()" tests don't add memory, and thus shouldn't change the bytesInUse. Not sure checking after each step is needed.

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/PartitionedRegionStatsUpdateTest.java
ringles: Added an intermediate step to both tests to confirm the bytesInUse actually goes up when a set/hash is added.

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/PartitionedRegionStatsUpdateTest.java
```diff
@@ -0,0 +1,512 @@
...
+      jedis1.sadd(SET_KEY, "value" + i);
+    }
+
+    long finalDataStoreBytesInUse = clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server1);
+
+    assertThat(finalDataStoreBytesInUse).isGreaterThan(initialDataStoreBytesInUse);
+  }
+
+  @Test
+  public void should_showNoIncreaseInDatastoreBytesInUse_givenSetValueSizeDoesNotIncreases() {
```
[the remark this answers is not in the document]
ringles: Yup, fixing!

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisHash.java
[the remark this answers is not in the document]
sabbey37: Good to know. Thank you!

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/PartitionedRegionStatsUpdateTest.java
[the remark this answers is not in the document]
sabbey37: That definitely makes sense.  Looking through the other tests, I can see why we don't always need this deserialization trigger.  The confusing part to me is that we randomly decided to use `jedis2` and check `server2` in this test.  It is also very similar to this test: `should_ShowNoIncreaseInDatastoreBytesInUse_givenHashValueSizeDoesNotIncrease()`, but I understand the need for both.... not sure if there's a better way of naming them so people can more easily distinguish between them.

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/PartitionedRegionStatsUpdateTest.java
```diff
@@ -0,0 +1,526 @@
...
+        clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server1);
+
+    for (int i = 0; i < 1000; i++) {
+      jedis1.hsetnx(HASH_KEY, FIELD, LONG_APPEND_VALUE);
+    }
+
+    long finalDataStoreBytesInUse = clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server1);
+
+    assertThat(finalDataStoreBytesInUse).isEqualTo(initialDataStoreBytesInUse);
+  }
```
sabbey37, line 262: This test is the same as `should_showNoIncreaseInDatastoreBytesInUse_givenHSetNXDoesNotIncreaseHashSize`

(The only difference is that this one uses `LONG_APPEND_VALUE` instead of `value`, but that shouldn't matter since `hsetnx` does a `putifabsent` for the field)

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/PartitionedRegionStatsUpdateTest.java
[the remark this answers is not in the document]
sabbey37: I think there is value in checking at least once (maybe at the end?) that `server1` and `server2` have the same  bytes in use.... especially if we want to actually test that the `membersAgreeUponUsedHashMemory` .

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisHash.java
```diff
@@ -145,13 +148,15 @@ private void shutDownHscanSnapshotScheduledRemoval() {
   public synchronized void toData(DataOutput out, SerializationContext context) throws IOException {
     super.toData(out, context);
     DataSerializer.writeHashMap(hash, out);
+    DataSerializer.writeInteger(hashSize.get(), out);
```
upthewaterspout, line 151: I wonder if we actually want to serialize this field or just recompute it during deserialization?

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisHash.java
```diff
@@ -58,6 +59,9 @@
   private ConcurrentHashMap<UUID, List<ByteArrayWrapper>> hScanSnapShots;
   private ConcurrentHashMap<UUID, Long> hScanSnapShotCreationTimes;
   private ScheduledExecutorService HSCANSnapshotExpirationExecutor = null;
+  private static final int PER_STRING_OVERHEAD = PER_OBJECT_OVERHEAD + 46;
+  private static final int PER_HASH_OVERHEAD = PER_OBJECT_OVERHEAD + 116;
+  private AtomicInteger hashSize = new AtomicInteger(PER_HASH_OVERHEAD);
```
upthewaterspout, line 64: Does this need to be an `AtomicInteger` or will just an `int` do? I see all of our update methods are synchronized, and I don't think the HashMap is safe under concurrent modification.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSet.java
```diff
@@ -46,16 +47,23 @@
 import org.apache.geode.redis.internal.delta.RemsDeltaInfo;
 
 public class RedisSet extends AbstractRedisData {
-
   private HashSet<ByteArrayWrapper> members;
 
+  private static final int PER_MEMBER_OVERHEAD = PER_OBJECT_OVERHEAD + 70;
+  private static final int PER_SET_OVERHEAD = PER_OBJECT_OVERHEAD + 240;
+
+  private AtomicInteger setSize = new AtomicInteger(PER_SET_OVERHEAD);
```
upthewaterspout, line 55: Some atomic question here.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisHash.java
upthewaterspout, line 63: I wonder if we need unit tests of these constants, so we remember to update them if someone changes the code?

upthewaterspout (commented):
Overall, looks good, just a few questions inline.

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/PartitionedRegionStatsUpdateTest.java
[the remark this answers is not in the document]
sabbey37: That definitely makes sense. Looking through the other tests, I can see why we don't always need this deserialization trigger. The confusing part to me is that we randomly decided to use jedis2 and check server2 in this test. It is also very similar to this test: `should_ShowNoIncreaseInDatastoreBytesInUse_givenHashValueSizeDoesNotIncrease()`, but I understand the need for both.... not sure if there's a better way of naming them so people can more easily distinguish between them.

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/PartitionedRegionStatsUpdateTest.java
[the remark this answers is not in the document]
sabbey37: I think there is value in checking at least once (maybe at the end?) that `server1` and `server2` have the same bytes in use.... especially if we want to actually test that the `membersAgreeUponUsedHashMemory` .

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisHash.java
nonbinaryprogrammer: from discussion with @ringles: it seems like writing it over the wire would save us some time recomputing the value. as the data set gets bigger, that could take us a good chunk of time. Is there a disadvantage to serializing it that we are not aware of?

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisHash.java
nonbinaryprogrammer: done! see RedisHashTest, RedisStringTest, and RedisSetTest

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/PartitionedRegionStatsUpdateTest.java
```diff
@@ -0,0 +1,526 @@
...
+    long intermediateDataStoreBytesInUse =
+        clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server1);
+    assertThat(intermediateDataStoreBytesInUse).isGreaterThan(initialDataStoreBytesInUse);
+
+    jedis1.del(HASH_KEY);
+
+    long finalDataStoreBytesInUse = clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server1);
+
+    assertThat(finalDataStoreBytesInUse).isEqualTo(initialDataStoreBytesInUse);
+  }
```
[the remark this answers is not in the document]
nonbinaryprogrammer: done!

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/PartitionedRegionStatsUpdateTest.java
```diff
@@ -0,0 +1,514 @@
...
+
+  /******* confirm that the other member agrees upon size *******/
+
+  @Test
+  public void should_showMembersAgreeUponUsedHashMemory_afterDeltaPropagation() {
```
sabbey37, line 325: I commented on this previously, but I think there is value for all these `AgreeUponUsedHashMemory` tests in checking at least once (maybe at the end?) that `server1` and `server2` have the same bytes in use.... especially if we want to actually test that the `membersAgreeUponUsedHashMemory`.

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisHashTest.java
```diff
@@ -258,6 +258,13 @@ public void hscanSnaphots_shouldExpireAfterExpiryPeriod() {
     });
   }
 
+  @Test
+  public void hashSizeOverhead_shouldNotBeChanged_withoutForethoughtAndTesting() {
+    assertThat(RedisHash.PER_OBJECT_OVERHEAD).isEqualTo(8);
+    assertThat(RedisHash.getPerStringOverhead()).isEqualTo(RedisHash.PER_OBJECT_OVERHEAD + 46);
+    assertThat(RedisHash.getPerHashOverhead()).isEqualTo(RedisHash.PER_OBJECT_OVERHEAD + 116);
+  }
```
sabbey37, line 266: Could we derive the overhead via the `ReflectionObjectSizer` here and compare that to the constant instead of hard coding what we derived?  Right now, someone could alter the code in a way that increases the overhead without changing the constant, but these tests wouldn't catch it.

sabbey37 (changes requested):

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisHashTest.java
upthewaterspout: +1 to what Sarah said!

nonbinaryprogrammer: we went to go write tests for this and found that we'd already done it in PartitionedRegionStatsUpdateTest. it's the last set of tests, divided by `/**` comments

upthewaterspout: Are we looking at hash_bytesInUse_shouldReflectActualSizeOfDataInRegion and those tests? Maybe it would be better if those tests were actually computing the size of RedisHash, rather than HashMap<ByteArrayWrapper,ByteArrayWrapper>?

sabbey37: Could we do some more direct unit testing in the `RedisHashTest.java`, maybe something like:

```
assertThat(ros.sizeof(new RedisHash())).isEqualTo(RedisHash.getPerHashOverhead());
assertThat(ros.sizeof(new RedisHash(createListOfDataElements(1))))
        .isEqualTo(RedisHash.getPerHashOverhead() + (RedisHash.getPerStringOverhead() * 2));
```

...and something similar in `RedisStringTest.java` and `RedisSetTest.java`

...it should be noted that the above assertions currently fail though, so maybe that's not the correct way to test it.  How did we originally derive the `PER_HASH_OVERHEAD` and `PER_STRING_OVERHEAD`? @upthewaterspout what is the proper way to derive/test those constants?

ringles: This turns out to be hard. We can create a ReflectionObjectSizer in the constructor, and get the size of the object without any members added.

Then we'd have to examine at least one member being added, and attempt to derive the per-object overhead as we added it. Except that it turns out that an empty HashSet doesn't allocate some storage. When you add the first object, you add both the object and the internal storage. So we could then add a second object, and based on the length attempt to figure out what the 'internal storage overhead' is...

Anyway, it's not trivial. But we might have to do it because it looks like - no surprise and as feared - JVM v11 doesn't have the same overhead as v8.

ringles: I was able to get the Sets to be dynamic - however, the values didn't change for Java 8, Java 11, and Liberica v8. Hashes are more complicated - the numbers just don't seem to add up, trying to figure things out from inside the constructor.

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/PartitionedRegionStatsUpdateTest.java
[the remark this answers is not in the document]
ringles: It's been surprisingly hard to guarantee that deserialization happens for Strings and Sets. The tests end up being flaky. I've ignored them for now... if there's a way to reliably make sure both servers have deserialized versions, I'd love to hear it.

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/PartitionedRegionStatsUpdateTest.java
```diff
@@ -0,0 +1,514 @@
...
+  private static final int JEDIS_TIMEOUT = Math.toIntExact(GeodeAwaitility.getTimeout().toMillis());
+  private static final String LOCAL_HOST = "127.0.0.1";
+  public static final String STRING_KEY = "string key";
+  public static final String SET_KEY = "set key";
+  public static final String HASH_KEY = "hash key";
+  public static final String LONG_APPEND_VALUE = String.valueOf(Integer.MAX_VALUE);
+  public static final String FIELD = "field";
+  private static final int SET_SIZE = 1000;
+
+  private ReflectionObjectSizer ros = ReflectionObjectSizer.getInstance();
```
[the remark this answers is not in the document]
ringles: At this point the ReflectionObjectSizer isn't needed anymore.

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/PartitionedRegionStatsUpdateTest.java
ringles: Unfortunately, we're having real trouble guaranteeing that things are stored in deserialized form on the secondary...

nonbinaryprogrammer:
> This pull request **introduces 1 alert** when merging [c84b2ec](https://github.com/apache/geode/commit/c84b2ec76118313948c7707e1ac3cc3acd2ca661) into [2e2c8f5](https://github.com/apache/geode/commit/2e2c8f5ccbb11414e41ff03f4c7cb6e9ed38e77e) - [view on LGTM.com](https://lgtm.com/projects/g/apache/geode/rev/pr-d4708faa7cf62a838be81f11154582123517e29c)
> 
> **new alerts:**
> 
>     * 1 for Implicit narrowing conversion in compound assignment

this has been resolved.

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisSetTest.java
```diff
@@ -165,4 +169,65 @@ public void setExpirationTimestamp_stores_delta_that_is_stable() throws IOExcept
...
+    StringBuffer sb = new StringBuffer(stringSize);
+    sb.append(base);
+    for (int i = base.length(); i < stringSize; i++) {
+      int randy = random.nextInt(10);
+      sb.append(randy);
+    }
+    String javaString = sb.toString();
+    return javaString;
+  }
+
```
sabbey37, line 232: Do we need to add tests like the ones that are in `RedisHash` (e.g. sizeShouldDecreaseWhenValueIsRemoved, shouldReturnToSetOverhead_whenAllMembersAreRemoved, etc.)?

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisStringTest.java
```diff
@@ -351,4 +354,50 @@ public void setExpirationTimestamp_stores_delta_that_is_stable() throws IOExcept
     o2.fromDelta(in);
     assertThat(o2).isEqualTo(o1);
   }
+
+  @Test
+  public void overheadConstants_shouldNotChange_withoutForethoughtAndTesting() {
+    assertThat(RedisString.PER_OBJECT_OVERHEAD).isEqualTo(8);
+    assertThat(RedisString.getPerStringOverhead()).isEqualTo(RedisString.PER_STRING_OVERHEAD);
```
sabbey37, line 361: The `getPerStringOverhead` method just returns the same constant we are checking against.  Could we test this constant using the ROS size of `RedisString` like we did for `RedisHash`?  Also, could we change the name of the test to something like `constantRedisStringOverhead_shouldEqual_calculatedOverhead` ?

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisStringTest.java
```diff
@@ -351,4 +354,50 @@ public void setExpirationTimestamp_stores_delta_that_is_stable() throws IOExcept
...
+  @SuppressWarnings("unchecked")
+  @Test
+  public void should_calculateSize_closeToROSSize_ofIndividualInstances() {
+    String javaString;
+    for (int i = 0; i < 512; i += 8) {
+      javaString = makeStringOfSpecifiedSize(i);
+      RedisString string = new RedisString(new RedisKey(javaString.getBytes()));
+
+      int expected = reflectionObjectSizer.sizeof(string);
+      Long actual = Long.valueOf(string.getSizeInBytes());
```
sabbey37, line 373: Intellij complains of `unnecessary boxing` for this as well as the similar line in the test below it... looks like it can just be cast to a `long` vs. using `Long.valueOf()`.  Apparently this is preferred due to space/time performance.

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisStringTest.java
```diff
@@ -351,4 +354,50 @@ public void setExpirationTimestamp_stores_delta_that_is_stable() throws IOExcept
...
+      assertThat(actual).isCloseTo(expected, offset);
+    }
+  }
+
+  @SuppressWarnings("unchecked")
+  @Test
+  public void should_calculateSize_closeToROSSize_ofLargeStrings() {
+    String javaString = makeStringOfSpecifiedSize(10_000);
+
+    RedisKey byteArrayWrapper = new RedisKey(javaString.getBytes());
```
sabbey37, line 385: It seems strange that we make a new `RedisKey` here as well as the test above this vs. creating a new `ByteArrayWrapper`.

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisStringTest.java
```diff
@@ -351,4 +354,50 @@ public void setExpirationTimestamp_stores_delta_that_is_stable() throws IOExcept
...
+  }
+
+  private String makeStringOfSpecifiedSize(final int stringSize) {
+    StringBuffer sb = new StringBuffer(stringSize);
+    for (int i = 0; i < stringSize; i++) {
+      sb.append("a");
+    }
+    String javaString = sb.toString();
+    return javaString;
+  }
```
sabbey37, line 402: Do we need to add tests like the ones that are in `RedisHash` (e.g. sizeShouldChangeWhenValueIsChanged, shouldReturnToStringOverhead_whenValueIsRemoved, etc.)?

sabbey37 (changes requested):
We are inconsistent with our use of `this...` (this.hash, this.someVariable, etc.). Since the use of `this.` is redundant and it seems we are trying to remove it altogether in other parts of Geode code, could we remove `this.` from the code that was added?  I realize it is used inconsistently throughout our Redis code (some of which I probably added myself), but that could be addressed in a separate PR.

Also tagging @upthewaterspout to re-review since he reviewed it initially.

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/PartitionedRegionStatsUpdateTest.java
```diff
@@ -0,0 +1,396 @@
...
+        clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server1);
+
+    assertThat(server1FinalDataStoreBytesInUse)
+        .isEqualTo(server2FinalDataStoreBytesInUse)
+        .isEqualTo(initialDataStoreBytesInUse);
+  }
+
+  @Test
+  @Ignore("confirm that bucket size stats are being calculated correctly before enabling")
+  public void should_showMembersAgreeUponUsedSetMemory_afterDeltaPropagation() {
```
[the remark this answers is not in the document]
nonbinaryprogrammer: nope, we're still running up against deserialization issues. can't figure out how to force the data to be deserialized on both members, so they don't always agree on the bytes in use. these were abandoned here to be picked up in a future story (which I will go create now) because of the size of this PR and the difficulty of getting those tests running. The same is the case for the test that you commented on below.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisHash.java
```diff
@@ -161,16 +168,46 @@ public int getDSFID() {
...
+    }
+
+    return oldvalue;
   }
 
   private synchronized ByteArrayWrapper hashPutIfAbsent(ByteArrayWrapper field,
       ByteArrayWrapper value) {
-    return hash.putIfAbsent(field, value);
+    ByteArrayWrapper oldvalue = hash.putIfAbsent(field, value);
+
```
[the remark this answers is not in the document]
nonbinaryprogrammer: good catch

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisSetTest.java
nonbinaryprogrammer: yes, good call.

nonbinaryprogrammer:
sorry, had to force push to since I rebased with develop to fix the memory overhead test that was failing because I decreased the memory overhead for hashes. I will definitely be squashing all of this out, and writing a detailed commit message.

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/PartitionedRegionStatsUpdateTest.java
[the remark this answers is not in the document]
nonbinaryprogrammer: the jira ticket for these tests: https://issues.apache.org/jira/browse/GEODE-9243

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSet.java
```diff
@@ -35,27 +35,61 @@
...
+    temp_hashset.add(baw1);
+    int one_entry_hashset_size = reflectionObjectSizer.sizeof(temp_hashset);
+    temp_hashset.add(baw2);
+    int two_entries_hashset_size = reflectionObjectSizer.sizeof(temp_hashset);
+
+    perMemberOverhead = two_entries_hashset_size - one_entry_hashset_size + 5;
+    internalHashsetStorageOverhead =
+        two_entries_hashset_size - (2 * perMemberOverhead) - base_hashset_size;
+
+    myCalculatedSize = baseRedissetOverhead;
```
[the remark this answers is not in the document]
nonbinaryprogrammer: switched to using constants and added tests

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisHashTest.java
```diff
@@ -258,11 +261,380 @@ public void hscanSnaphots_shouldExpireAfterExpiryPeriod() {
...
 
+  /************* Hash Size *************/
+  /******* constants *******/
+  @Test
+  public void constantBaseRedisHashOverhead_shouldEqualCalculatedOverhead() {
+    ReflectionObjectSizer reflectionObjectSizer = ReflectionObjectSizer.getInstance();
+    int baseRedisHashOverhead = reflectionObjectSizer.sizeof(new RedisHash());
+
+    assertThat(baseRedisHashOverhead).isEqualTo(BASE_REDIS_HASH_OVERHEAD);
+  }
```
sabbey37, line 272: This test is the same as `should_calculateSize_equalToROSSize_ofEmptyIndividualInstance()`.  Do we need both? Also, it looks like we already have the class variable `reflectionObjectSizer` that we could use instead of getting another instance.

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisSetTest.java
```diff
@@ -165,4 +174,142 @@ public void setExpirationTimestamp_stores_delta_that_is_stable() throws IOExcept
...
+    int finalSize = set.getSizeInBytes();
+    int expectedSize = initialSize - value.length() - PER_MEMBER_OVERHEAD;
+
+    assertThat(finalSize).isEqualTo(expectedSize);
+  }
+
+  /******** constants *******/
+  @Test
+  public void baseOverheadConstant_shouldMatchCalculatedValue() {
+    ReflectionObjectSizer reflectionObjectSizer = ReflectionObjectSizer.getInstance();
```
sabbey37, line 233: We already have the class variable `reflectionObjectSizer` that we could use instead of getting another instance.

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisSetTest.java
```diff
@@ -165,4 +174,142 @@ public void setExpirationTimestamp_stores_delta_that_is_stable() throws IOExcept
...
+    int baseRedisSetOverhead = reflectionObjectSizer.sizeof(new RedisSet()) + 18;
```
sabbey37, line 234: I'm wondering how we came up with `18` here?  It is close to the size of an empty `ByteArrayWrapper` (which is 16)... not sure what the additional 2 would be... maybe padding?  Either way, a comment would be good explaining it.

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisSetTest.java
```diff
@@ -165,4 +174,142 @@ public void setExpirationTimestamp_stores_delta_that_is_stable() throws IOExcept
...
+
+    HashSet<ByteArrayWrapper> temp_hashset = new HashSet<>();
+    int base_hashset_size = reflectionObjectSizer.sizeof(temp_hashset);
```
sabbey37, line 237: This could be done in one line so you wouldn't need the `temp_hashset` variable.  Also, not sure why we switched to snake case for the variable name?  Could we change it to `baseHashSetSize` for consistency?

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisSetTest.java
```diff
@@ -165,4 +174,142 @@ public void setExpirationTimestamp_stores_delta_that_is_stable() throws IOExcept
...
+    baseRedisSetOverhead += base_hashset_size;
+
+    assertThat(baseRedisSetOverhead).isEqualTo(BASE_REDIS_SET_OVERHEAD);
+  }
+
+  @Test
+  public void perMemberOverheadConstant_shouldMatchCalculatedValue() {
+    ReflectionObjectSizer reflectionObjectSizer = ReflectionObjectSizer.getInstance();
```
sabbey37, line 245: We already have the class variable `reflectionObjectSizer` that we could use instead of getting another instance.

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisSetTest.java
```diff
@@ -165,4 +174,142 @@ public void setExpirationTimestamp_stores_delta_that_is_stable() throws IOExcept
...
+    HashSet<ByteArrayWrapper> temp_hashset = new HashSet<>();
+    ByteArrayWrapper member1 = new ByteArrayWrapper("a".getBytes());
+    ByteArrayWrapper member2 = new ByteArrayWrapper("b".getBytes());
+    temp_hashset.add(member1);
+    int one_entry_hashset_size = reflectionObjectSizer.sizeof(temp_hashset);
+
+    temp_hashset.add(member2);
+    int two_entries_hashset_size = reflectionObjectSizer.sizeof(temp_hashset);
+
+    int perMemberOverhead = two_entries_hashset_size - one_entry_hashset_size + 5;
```
sabbey37, line 256: Not sure why we switched to snake case for these variable names?  Could we change it to camel case for consistency?

A comment explaining the `+5` would be good.

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisSetTest.java
```diff
@@ -165,4 +174,142 @@ public void setExpirationTimestamp_stores_delta_that_is_stable() throws IOExcept
...
+
+    assertThat(perMemberOverhead).isEqualTo(PER_MEMBER_OVERHEAD);
+  }
+
+  @Test
+  public void internalHashsetStorageOverheadConstant_shouldMatchCalculatedValue() {
+    ReflectionObjectSizer reflectionObjectSizer = ReflectionObjectSizer.getInstance();
```
sabbey37, line 263: We already have the class variable `reflectionObjectSizer` that we could use instead of getting another instance.

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisSetTest.java
```diff
@@ -165,4 +174,142 @@ public void setExpirationTimestamp_stores_delta_that_is_stable() throws IOExcept
...
+
+    temp_hashset.add(baw1);
+    temp_hashset.add(baw2);
+
+    int two_entries_hashset_size = reflectionObjectSizer.sizeof(temp_hashset);
+
+    int internalHashsetStorageOverhead =
+        two_entries_hashset_size - (2 * PER_MEMBER_OVERHEAD) - base_hashset_size;
+
+    assertThat(internalHashsetStorageOverhead).isEqualTo(INTERNAL_HASH_SET_STORAGE_OVERHEAD);
```
sabbey37, line 279: Not sure why we switched to snake case for these variable names? Could we change it to camel case for consistency?

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisStringTest.java
```diff
@@ -351,4 +354,95 @@ public void setExpirationTimestamp_stores_delta_that_is_stable() throws IOExcept
     o2.fromDelta(in);
     assertThat(o2).isEqualTo(o1);
   }
+
+  /************* Size in Bytes Tests *************/
+  /******* constructors *******/
+  @SuppressWarnings("unchecked")
+  @Test
+  public void should_calculateSize_closeToROSSize_ofLargeStrings() {
```
sabbey37, line 362: This title could be `equalToROSSize` since we expect exact equality.

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisStringTest.java
```diff
@@ -351,4 +354,95 @@ public void setExpirationTimestamp_stores_delta_that_is_stable() throws IOExcept
...
+
+    int finalSize = string.getSizeInBytes();
+
+    assertThat(finalSize).isEqualTo(BASE_REDIS_STRING_OVERHEAD);
+  }
+
+  /******* constants *******/
+  @Test
+  public void overheadConstants_shouldMatchCalculatedValue() {
+    assertThat(RedisString.PER_OBJECT_OVERHEAD).isEqualTo(8); // see todo in RedisString
```
sabbey37, line 431: I'm a little confused about why we're checking this since we don't use the `PER_OBJECT_OVERHEAD` at all to calculate the size of `RedisString`.  Is the todo mentioned here this one: `TODO add delta support` ?

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisStringTest.java
```diff
@@ -351,4 +354,95 @@ public void setExpirationTimestamp_stores_delta_that_is_stable() throws IOExcept
...
+
+    int actual = string.getSizeInBytes();
+    int expected = reflectionObjectSizer.sizeof(string);
+
+    assertThat(actual).isEqualTo(expected);
+  }
+
+  @SuppressWarnings("unchecked")
+  @Test
+  public void should_calculateSize_closeToROSSize_ofStringOfVariousSizes() {
```
sabbey37, line 374: This title could be `equalToROSSize` since we expect exact equality.

sabbey37 (changes requested):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisString.java
```diff
@@ -711,6 +716,11 @@ protected void handleSetExpiration(SetOptions options) {
     }
   }
 
+  @VisibleForTesting
+  protected static int getBaseRedisStringOverhead() {
+    return BASE_REDIS_STRING_OVERHEAD;
+  }
+
```
sabbey37, line 723: This method is no longer used.

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/PartitionedRegionStatsUpdateTest.java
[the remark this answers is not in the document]
nonbinaryprogrammer: I'm pretty hesitant to do that since the old tracker tickets are still referenced and almost no one has access to them, so it's useless information. so I don't think we should be adding more ticket numbers to the codebase

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisHashTest.java
nonbinaryprogrammer: they test slightly different things but since they have a lot of overlapping code, I put both assertions in this test and deleted the other one

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisSetTest.java
nonbinaryprogrammer: I also have no idea why it is 18. I didn't do the original Set implementation. figuring it out will take a lot of testing so I'd rather not have to. but I'll check with Ray on if he remembers where that came from.

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisSetTest.java
nonbinaryprogrammer: yeah I thought I had changed everything from snake case but I must have missed at least one.

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisHashTest.java
sabbey37: ah ok, I see that now. If we want to keep both that makes sense.

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisSetTest.java
nonbinaryprogrammer: I'm also not sure about that one. I pulled this math over from the RedisSortedSet and didn't write that original code. I'll check with Ray when he's back from his covid vaccine.

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisStringTest.java
nonbinaryprogrammer: the TODO that I'm referring to is: `// TODO: for a 64bit jvm with small ops this is 12; for other 64bit jvms it is 16`. If you command-click on the `PER_OBJECT_OVERHEAD` constant, it's the TODO right above where that constant is defined in the Sizeable interface. We do need to keep the constant, at least for now, but if you feel that the test does not add value, I can remove that.

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisSetTest.java
nonbinaryprogrammer: through a lot more work on RedisSet, I was able to get rid of the +18. the mysterious +5 is still there, but we now know this has something to do with the way the ByteArrayWrapper resizes. hopefully we can get rid of that magic number once we finish the story on getting rid of ByteArrayWrappers.

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisSetTest.java
nonbinaryprogrammer: the magic 5 remains, after a lot more testing of RedisSet, and some changes to the implementation. it technically shouldn't be a constant, and instead is related to the size of the set, and of the bytearraywrappers within the set. it is possible for us to be more accurate here, but it requires a lot of cleverness. Since there's a story to remove ByteArrayWrappers, it seems like a waste of time to get the size exactly right all the time, when we can be within 5% without all that math.

nonbinaryprogrammer:
Draft of commit message (feel free to leave comments):
GEODE-9136: Make RedisData implement Sizeable

- use Sizeable in RedisString, RedisSet, and RedisHash
- add unit tests of bytes in use for all three classes
- update memory overhead tests to reflect decrease in overhead 
- calculate estimated size in unit tests and use constants in the classes

RedisData should implement Sizeable to increase the accuracy of the estimation of bytes in use. It is important that we are close to the right size in order for rebalancing to work properly, however the exact value is not important. These changes make the size for sets and hashes be within 5% of the size measured by using reflection. The accuracy changes as you add entries, and as entries get larger, because of the way that hashes, sets, and ByteArrayWrappers get resized. As we fill up the hash and set, we become more accurate in our estimation of the size. When full, the next entry added causes the structure to be resized, which means that our data now only accounts for a portion of the total size of the structure. We could have kept track of the resizes that happen in order to be even more accurate in our estimations, but doing so would add computational overhead for adding each key, and for each field at that key. It was decided that the benefits do not outweigh the cost.

RedisHash and RedisSet both use constants to account for the overheads of:
- an empty (RedisHash/RedisSet)
- an empty (HashMap/HashSet) within the (RedisHash/RedisSet)
- each member in the (hash/set)
RedisHash also has an overhead for the first value in the hash.
We decided to use constants in the RedisData classes in order to reduce the number of calculations we have to do every time we add a new key. The math used to derive those constants live in the RedisSetTest and RedisHashTest classes. The tests of the constants will fail if the internal implementation of the classes changes. If the overheads decrease, then the constants need to be updated to reflect that. If the overheads increase, think through the changes before adding to the overhead. Significant changes to RedisSet or RedisHash could potentially break the math, so be sure to check where the problem is, in case we need to change the calculations.
There is a magical +5 in RedisSetTest.perMemberOverheadConstant_shouldMatchCalculatedValue. We think it came from the resizing, and that 5 happens to work ok, even though it should be dynamically calculated. But as previously said, the benefits do not outweigh the cost.

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/data/PartitionedRegionStatsUpdateTest.java
```diff
@@ -0,0 +1,396 @@
...
+    long server1FinalDataStoreBytesInUse =
+        clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server1);
+    long server2FinalDataStoreBytesInUse =
+        clusterStartUpRule.getDataStoreBytesInUseForDataRegion(server2);
+
+    assertThat(server1FinalDataStoreBytesInUse)
+        .isEqualTo(initialDataStoreBytesInUse);
+
+    assertThat(server2FinalDataStoreBytesInUse)
+        .isEqualTo(initialDataStoreBytesInUse);
```
sabbey37, line 394: This will currently never be true.  We could change the for loop to just do just set the key to `value` repeatedly. then make sure the STRING_KEY is still equal to `value`. Then do the same assertion we did for sets and hashes:  
```
    assertThat(server1FinalDataStoreBytesInUse)
        .isEqualTo(server2FinalDataStoreBytesInUse)
        .isEqualTo(initialDataStoreBytesInUse);
```

Unfortunately, it still doesn't pass though.

sabbey37 (changes requested):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSet.java
```diff
@@ -243,24 +251,28 @@ private boolean membersRemove(ByteArrayWrapper memberToRemove) {
...
 
-  private synchronized boolean membersAddAll(AddsDeltaInfo addsDeltaInfo) {
+  private synchronized void membersAddAll(AddsDeltaInfo addsDeltaInfo) {
     ArrayList<ByteArrayWrapper> adds = addsDeltaInfo.getAdds();
     sizeInBytes += adds.stream().mapToInt(a -> a.length() + PER_MEMBER_OVERHEAD).sum();
-    return members.addAll(adds);
+    if (members.size() == 1) {
+      sizeInBytes += INTERNAL_HASH_SET_STORAGE_OVERHEAD;
+    }
+    members.addAll(adds);
```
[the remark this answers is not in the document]
nonbinaryprogrammer: you're right. the other place we did this adds the member before doing that check, hence the `==1`

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSet.java
```diff
@@ -46,16 +46,37 @@
...
+  // Note: the per member overhead is known to not be constant. it changes as more members are
+  // added, and/or as the members get longer
+  protected static final int BASE_REDIS_SET_OVERHEAD = 112;
+  protected static final int PER_MEMBER_OVERHEAD = 77;
+  protected static final int INTERNAL_HASH_SET_STORAGE_OVERHEAD = 86;
+
+  private int sizeInBytes;
+
   RedisSet(Collection<ByteArrayWrapper> members) {
+    sizeInBytes += BASE_REDIS_SET_OVERHEAD;
```
sabbey37, line 65: Should we move this to the empty `RedisSet` constructor like we did for `RedisHash`?

nonbinaryprogrammer:
> very minor, but if you end up making other changes: if
> HASH_MAP_VALUE_PAIR_OVERHEAD & SIZE_OF_OVERHEAD_OF_FIRST_PAIR are the same value, do we need both?

take a look at `RedisHashTest.constantFirstPairOverhead_shouldEqual_calculatedOverhead`; that's where the calculation for `SIZE_OF_OVERHEAD_OF_FIRST_PAIR` happens. The value is based off `HASH_MAP_VALUE_PAIR_OVERHEAD`, but is not necessarily equal, and has different meaning. so I do believe that the two are needed

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSet.java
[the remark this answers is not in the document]
sabbey37: Thinking about it more, we don't seem to need this check at all.  Similar to `membersRemoveAll`, because this method is only ever called from `applyDelta`, we actually don't need to check if the set is empty. The set will never be empty in this method.  When the key/value are initially created in the region, there is no delta applied/sent (from the perspective of the `RedisSet` class).

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSet.java
```diff
@@ -214,23 +239,43 @@ public int getDSFID() {
...
+    members.addAll(adds);
   }
 
-  private synchronized boolean membersRemoveAll(RemsDeltaInfo remsDeltaInfo) {
-    return members.removeAll(remsDeltaInfo.getRemoves());
+  private synchronized void membersRemoveAll(RemsDeltaInfo remsDeltaInfo) {
+    ArrayList<ByteArrayWrapper> removes = remsDeltaInfo.getRemoves();
+    if (members.size() == removes.size()) {
+      sizeInBytes = BASE_REDIS_SET_OVERHEAD;
+    } else {
```
sabbey37, line 273: We could remove this, but it doesn't hurt to leave it in.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSet.java
```diff
@@ -214,23 +239,43 @@ public int getDSFID() {
...
+      sizeInBytes -= removes.stream().mapToInt(a -> a.length() + PER_MEMBER_OVERHEAD).sum();
+    }
+    members.removeAll(removes);
   }
```
sabbey37, line 277: To test this properly, we could add a test similar to `should_showMembersAgreeUponUsedSetMemory_afterDeltaPropagation()`, but check the bytes in use after deleting some members from a set.

sabbey37 (approved):
It looks good!

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSet.java
DonalEvans: I've added a test that checks this

PR #6302, opened 2021-04-09
GEODE-9068: Follow-up to APIs compatible with Redis doc update 

davebarnes97:
Fixes one bad link in the user guide left-hand navigation panel.
Eliminates 537 warnings when generating the user guide.


nonbinaryprogrammer (approved):

jhutchison (commented):
looks good!

PR #6308, opened 2021-04-12
GEODE-9139 SSLException in starting up a Locator

bschuchardt:
These changes preserve the bind-address passed to a Locator by Gfsh and modify the way hostnames are handled when SNI endpoint validation is enabled.

LocatorLauncher: Preserve the bind-address string specified by the user for cluster communications

JGroupsMessenger: Use the bind-address string or the Inetaddress host-name when forming the local identifier so it will be available in the membership view for other members of the cluster.

ServiceConfig: Allow use of DNS when SNI endpoint validation is enabled.

### For all changes:

### Note:


mhansonp (approved):

geode-assembly/src/acceptanceTest/java/org/apache/geode/client/sni/DualServerSNIAcceptanceTest.java
```diff
@@ -97,6 +107,17 @@ public void after() {
     ensureCacheClosed();
   }
 
+  @AfterClass
+  public static void afterClass() throws Exception {
+    // try {
+    // String output =
+    // docker.get().exec(options("-T"), "locator-maeve",
```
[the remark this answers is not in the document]
bschuchardt: This commented-out code is extremely useful in debugging problems & I don't want to lose track of it again.

onichols-pivotal (approved):

Bill (commented):
I made a few minor comments. Rather than making more scatter-gun comments, I thought I'd share a more cohesive insight: `bindAddress` and `bindAddressString` would be easier to understand if they were one object (e.g. in `LocatorLauncher`).

The `HostAndPort` class, while not exactly what we want, since we don't have a port in this case, has the key behavior we want, to wit, to construct with either an IP address or a symbolic name, and to provide the ability to resolve a name (late) and to cache the result of that resolution.

At first I thought this suggestion might make for a larger diff. But now I'm thinking it's about the same. Everywhere we were passing `bindAddress` (an `InetAddress`) we could pass one of these new things, maybe a `Host` object. Just as `HostAndPort` has `getSocketInetAddress()`, maybe `Host` could have `getInetAddress()`?

jinmeiliao (approved):

bschuchardt:
@Bill, I couldn't use the HostAndPort class without making things crazy because bind-address is optional but the locator port is not.  I would have to modify HostAndPort to make the host address optional, which would change the meaning of that class too much.

Instead I created a new HostAddress class that holds an InetSocketAddress based either on an InetAddress or a hostName, combining the two parameters that were in the original PR.

rhoughton-pivot (approved):

geode-core/src/main/java/org/apache/geode/distributed/Locator.java
```diff
@@ -101,7 +105,7 @@
    */
   public static Locator startLocator(int port, File logFile) throws IOException {
 
-    return startLocator(port, logFile, false, (InetAddress) null, (Properties) null, true, true,
+    return startLocator(port, logFile, false, (HostAddress) null, (Properties) null, true, true,
```
kamilla1201, line 108: Here and in a few other places in this file (that call startLocator) casting null args to HostAddress or Properties seems redundant. Is there a reason to keep that anyway?

geode-core/src/main/java/org/apache/geode/internal/DistributionLocator.java
```diff
@@ -136,28 +139,27 @@ public static void main(String args[]) {
...
+                DistributionLocator.shutdown(port, faddress.getAddress());
               } catch (IOException e) {
                 e.printStackTrace();
               }
             }));
       }
 
-      lockFile = ManagerInfo.setLocatorStarting(directory, port, address);
+      lockFile = ManagerInfo.setLocatorStarting(directory, port,
+          hostAddress == null ? null : hostAddress.getAddress());
```
kamilla1201, line 154: Looks like this could be simplified by creating the `address` variable a bit earlier
```
final InetAddress address = hostAddress == null ? null : hostAddress.getAddress();
lockFile = ManagerInfo.setLocatorStarting(directory, port, address);
```

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/api/HostAddress.java
```diff
@@ -0,0 +1,106 @@
...
+
+import org.apache.commons.validator.routines.InetAddressValidator;
+
+import org.apache.geode.distributed.internal.tcpserver.HostAndPort;
+
+/**
+ * HostAddress is a holder of a host name/address. It is the primary
+ * way to specify a host address that may or may not have been resolved to an InetAddress.
+ *
+ * @See HostAddress which can hold both a hast name and a port
```
kamilla1201, line 29: Typo in "host" here

kamilla1201 (approved):
Nice work! Just a few small suggestions.

geode-assembly/src/acceptanceTest/java/org/apache/geode/client/sni/DualServerSNIAcceptanceTest.java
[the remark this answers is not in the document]
bschuchardt: comment added

kirklund (approved):

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/api/HostAddress.java
```diff
@@ -0,0 +1,106 @@
...
+
+  /**
+   * Returns an InetSocketAddress for this host and port. An attempt is made to resolve the
+   * host name but if resolution fails an unresolved InetSocketAddress is returned. This return
+   * value will not hold an InetAddress, so calling getAddress() on it will return null.
+   */
+  private InetSocketAddress getInetSocketAddress() {
+    if (address.isUnresolved()) {
+      // note that this leaves the InetAddress null if the hostname isn't resolvable
+      return new InetSocketAddress(address.getHostString(), address.getPort());
```
Bill, line 66: Shouldn't `HostAddress` cache this value before returning it?

Failing to do so causes us to resolve the hostname every single time the socket address is needed.

If this is a problem here, it's also a problem in the sister-class (or perhaps mother-class) `HostAndPort`—it also fails to cache the value.

Also I recommend calling this method `getSocketInetAddress` to match the one in `HostAndPort`.

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/api/HostAddress.java
```diff
@@ -0,0 +1,106 @@
...
+    } else {
+      return this.address;
+    }
+  }
+
+  public String getHostName() {
+    String hostString = address.getHostString();
+    if (hostString != null) {
+      return hostString;
+    }
```
Bill, line 76: I looked at `InetSocketAddress` (docs and implementation) and it looks like `getHostName()` already avoids the reverse lookup if the host string is already known. So I believe lines 73-76 are superfluous.

Bill (changes requested):
I took a quick look focused primarily on the new and excellent `HostAddress` class.

It seems like a caching opportunity was missed there, which points out a corresponding caching opportunity missed our older `HostAndPort` class.

Also, I wonder if there might be an economical way to share implementation between those two classes. You explained why they have to be distinct. But I wonder if they could share implementation somehow.

gradle/warnings.gradle
```diff
@@ -16,6 +16,6 @@
  */
 
 tasks.withType(JavaCompile) {
-  options.compilerArgs << '-Xlint:unchecked' << "-Werror"
-  options.deprecation = true
+//  options.compilerArgs << '-Xlint:unchecked' << "-Werror"
+//  options.deprecation = true
 }
```
jdeppe-pivotal, line 21: You probably didn't mean to leave these commented out?

jdeppe-pivotal (approved):
LG other than the comment I left

gradle/warnings.gradle
bschuchardt: ew, thanks for catching that Jens.  I was struggling with the Gradle upgrade in IJ when I made that change

boglesby (approved):

dickcav (approved):

echobravopapa (approved):
Good stuff, this will be very useful going forward.

geode-tcp-server/src/main/java/org/apache/geode/distributed/internal/tcpserver/InetSocketWrapper.java
```diff
@@ -0,0 +1,97 @@
...
+package org.apache.geode.distributed.internal.tcpserver;
+
+import java.net.InetAddress;
+import java.net.InetSocketAddress;
+import java.util.Objects;
+
+import org.apache.commons.validator.routines.InetAddressValidator;
+
+class InetSocketWrapper {
+  String hostName;
```
Bill, line 24: Why does this class need to track `hostName`? Doesn't `InetSocketAddress` adequately manage that state?

geode-tcp-server/src/main/java/org/apache/geode/distributed/internal/tcpserver/InetSocketWrapper.java
```diff
@@ -0,0 +1,97 @@
...
+    } else {
+      // non-numeric address - resolve hostname when needed
+      inetSocketAddress = InetSocketAddress.createUnresolved(hostName, port);
+    }
+  }
+
+  public String getHostName() {
+    if (hostName != null) {
+      return hostName;
+    }
```
Bill, line 46: This checking is superfluous because `InetSocketAddress.getHostName()` does this same `null` check.

geode-tcp-server/src/main/java/org/apache/geode/distributed/internal/tcpserver/HostAddress.java
```diff
@@ -0,0 +1,53 @@
...
+
+  public HostAddress(HostAndPort hostAndPort) {
+    hostName = hostAndPort.hostName;
+    inetSocketAddress = hostAndPort.inetSocketAddress;
+  }
+
+  @Override
+  public String toString() {
+    if (hostName != null) {
+      return hostName;
```
Bill, line 48: If we get rid of `hostName` in the base class this could change to maybe:

```
final String hostName = getHostName();
if (hostName == null)
  return inetSocketAddress.getAddress().toString();
else
  return hostName;
```

Bill (commented):
Consolidation of `HostAddress` and `HostAndPort` is looking great.

I have a few fine-tuning comments…

geode-tcp-server/src/main/java/org/apache/geode/distributed/internal/tcpserver/InetSocketWrapper.java
bschuchardt: It does not adequately track the original host name.  There is a unit test that failed after setting a bind address of 127.0.0.1 and InetSocketAddress somehow converted that to "localhost".

geode-tcp-server/src/main/java/org/apache/geode/distributed/internal/tcpserver/InetSocketWrapper.java
bschuchardt: I found that this code is necessary to preserve the original string passed in to the constructor.  If you look at the constructor for InetSocketAddress you'll see that it discards the hostname parameter if it's a numeric address.

geode-tcp-server/src/main/java/org/apache/geode/distributed/internal/tcpserver/HostAddress.java
bschuchardt: To do this I would have to delete a gfsh unit test that requires the Locator startup command to have a numeric address as the bind address if that is what was specified by the user.

PR #6314, opened 2021-04-13
GEODE-9141: Hang while shutting down a cache server due to corrupted message

bschuchardt:
The corruption is occurring because another thread was using the same ByteBuffer to
perform a handshake on a new Connection and was still in
SocketChannel.read().  The thread initiating the handshake was
interrupted by CacheServer.close() and released the buffer prematurely,
allowing another thread to grab the still in-use buffer and serialize
a message into it.  The handshake thread's SocketChannel.read()
overwrites some of the content of the buffer before it is transmitted.

The fix is to include the input buffer as state that needs to be
protected by the stateLock and allow the handshake thread to retain
ownership of the buffer, releasing it in it's cleanup code.

I also deleted some very old checks for an IBM JRE that is no longer
required and was interfering with efforts to write a unit test.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/tcp/Connection.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/tcp/Connection.java
@@ -57,7 +57,6 @@ import org.apache.geode.annotations.internal.MakeNotStatic;
 import org.apache.geode.annotations.internal.MutableForTesting;
 import org.apache.geode.cache.CacheClosedException;
 import org.apache.geode.distributed.DistributedSystemDisconnectedException;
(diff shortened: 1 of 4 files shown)


mivanac (approved):

geode-core/src/main/java/org/apache/geode/internal/tcp/Connection.java
```diff
@@ -1938,6 +1942,10 @@ public void setInUse(boolean use, long startTime, long ackWaitThreshold, long ac
     }
   }
 
+  void setReaderThreadForTest(Thread threadForTest) {
```
Bill, line 1945: add `@VisibleForTesting`

PR #6317, opened 2021-04-13
GEODE-7674: Clear on PR with lucene index should throw exception

gesterzhou:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/wan/InternalGatewaySender.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/wan/InternalGatewaySender.java
@@ -40,6 +40,8 @@ public interface InternalGatewaySender extends GatewaySender {
 
   boolean isForwardExpirationDestroy();
 
+  boolean isPartitionedRegionClearUnsupported();
+
   boolean getIsMetaQueue();
 
   InternalCache getCache();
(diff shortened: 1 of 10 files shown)


geode-core/src/main/java/org/apache/geode/internal/cache/wan/InternalGatewaySender.java
```diff
@@ -40,6 +40,8 @@
 
   boolean isForwardExpirationDestroy();
 
+  boolean isPartitionedRegionClearUnsupported();
```
kirklund, line 43: I think it would be clearer to use {{isPartitionedRegionClearSupported()}} instead of the negative "unsupported".

kirklund (approved):
Approved with one comment about `isPartitionedRegionClearUnsupported`.

geode-core/src/main/java/org/apache/geode/internal/cache/wan/InternalGatewaySender.java
dschneider-pivotal: @kirklund That would mean changing the default to true. Does that sound good to you?

dschneider-pivotal (approved):

geode-core/src/main/java/org/apache/geode/internal/cache/wan/InternalGatewaySender.java
gesterzhou: We just changed to "unsupported" based on Darrel's comments.

PR #6322, opened 2021-04-15
GEODE-9064: Configure serial filter during ManagementAgent start (#6196)

kirklund:
The Geode JMX layer will configure the system property
“jmx.remote.rmi.server.serial.filter.pattern" to only accept
JDK classes identified as open-types required for JMX.

If the system property already has a value then Geode will log
a statement and leave the system property alone.

(cherry picked from commit c43dbc59dd54df3538598e85243698840f8b6541)


onichols-pivotal (approved):

kirklund:
DistributedTestOpenJDK11 failed with:
```
> Task :geode-core:distributedTest

timeout exceeded
```

PR #6325, opened 2021-04-15
GEODE-9155: change frequency of passive expiration

nonbinaryprogrammer:
increase the time between checks for keys to expire. this reduces the
cpu utilization but causes longer collections when they do collect.

### For all changes:

### Note:

### DIFF
--- a/geode-apis-compatible-with-redis/src/acceptanceTest/resources/0001-configure-redis-tests.patch
+++ b/geode-apis-compatible-with-redis/src/acceptanceTest/resources/0001-configure-redis-tests.patch
@@ -162,6 +162,33 @@ index de24eabed..aeeb1da7f 100644
  
      test {PERSIST returns 0 against non existing or non volatile keys} {
          r set x foo
+@@ -149,13 +149,13 @@ start_server {tags {"expire"}} {
+-test {Redis should actively expire keys incrementally} {
+-        r flushdb
+-        r psetex key1 500 a
+-        r psetex key2 500 a
+-        r psetex key3 500 a
+-        set size1 [r dbsize]
+-        # Redis expires random keys ten times every second so we are
+-        # fairly sure that all the three keys should be evicted after
+-        # one second.
+-        after 1000
+-        set size2 [r dbsize]
+-        list $size1 $size2
+-    } {3 0}
++test {Redis should actively expire keys incrementally} {
++        r flushdb
++        r psetex key1 500 a
++        r psetex key2 500 a
++        r psetex key3 500 a
++        set size1 [r dbsize]
++        # Redis expires random keys ten times every second so we are
++        # fairly sure that all the three keys should be evicted after
++        # one second.
++        after 1000
++        set size2 [r dbsize]
++        list $size1 $size2
++    } {3 0}
 @@ -154,39 +154,39 @@ start_server {tags {"expire"}} {
          set size1 [r dbsize]
          # Redis expires random keys ten times every second so we are
--- a/geode-apis-compatible-with-redis/src/acceptanceTest/resources/0001-configure-redis-tests.patch
+++ b/geode-apis-compatible-with-redis/src/acceptanceTest/resources/0001-configure-redis-tests.patch
@@ -162,8 +162,8 @@ index de24eabed..aeeb1da7f 100644
  
      test {PERSIST returns 0 against non existing or non volatile keys} {
          r set x foo
-@@ -149,13 +149,13 @@ start_server {tags {"expire"}} {
--test {Redis should actively expire keys incrementally} {
+@@ -149,44 +149,44 @@ start_server {tags {"expire"}} {
+-    test {Redis should actively expire keys incrementally} {
 -        r flushdb
 -        r psetex key1 500 a
 -        r psetex key2 500 a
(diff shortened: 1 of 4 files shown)


sabbey37:
I've been thinking about the way we check out a branch in CI and apply the changes.  I think, rather than trying to edit the patch file manually, the smarter thing to do may be the following:
1. Check out the same branch locally as the branch we check out in `execute_redis_tests.sh` (right now we do `git checkout origin/5.0`, but we should consider checking out a specific release tag vs. a branch, since that 5.0 branch is still being updated and changes to the tests might mess up our patch [lines might be different, etc.])
2. Apply the existing patch file
3. Make desired changes
4. Generate a new patch file based on those changes 
5. Replace the current patch file

sabbey37 (changes requested):
There is also the following test in `AbstractHitsMissesIntegrationTest.java`:

```
  @Test
  public void testPassiveExpiration() {
    runCommandAndAssertNoStatUpdates("hash", (k) -> {
      jedis.expire(k, 1);
      GeodeAwaitility.await().during(Duration.ofSeconds(3)).until(() -> true);
    });
  }
  ```
 
  ...which tests that keyspace hits and misses don't update after the EXPIRE command and after the key passively expires.  However, we should probably increase the time we are waiting ... maybe we could tie that to the `INTERVAL` constant in `PassiveExpirationManager` so it gets updated in conjunction with that?
  
Perhaps we could change the await to be something like this:
```
 GeodeAwaitility.await()
          .atMost(Duration.ofMinutes(PassiveExpirationManager.INTERVAL))
          .until(() -> jedis.keys("hash").isEmpty());
 ```
 
 so we're verifying the key expires before checking the keyspace hits and misses.  We may need to add some time in the await to that `INTERVAL` for a buffer and to avoid potential flakiness.

sabbey37 (changes requested):
I don't think any of my questions/comments from the previous review were addressed.

geode-apis-compatible-with-redis/src/acceptanceTest/resources/0001-configure-redis-tests.patch
```diff
@@ -162,17 +162,33 @@ index de24eabed..aeeb1da7f 100644
...
++#    test {Redis should actively expire keys incrementally} {
++#        r flushdb
++#        r psetex key1 500 a
++#        r psetex key2 500 a
++#        r psetex key3 500 a
++#        set size1 [r dbsize]
++#        # Redis expires random keys ten times every second so we are
++#        # fairly sure that all the three keys should be evicted after
++#        # one second.
++#        after 1000
```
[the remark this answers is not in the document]
ringles: Exposing PassiveExpirationManager.INTERVAL done; I think the other test obviates the need for the TCL test. We're explicitly deviating from Redis behavior, so not worrying about their integration tests in this scenario seems okay. Or have I misunderstood the question?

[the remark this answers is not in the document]
sabbey37: Yeah, that definitely makes sense.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/server/AbstractHitsMissesIntegrationTest.java
```diff
@@ -168,7 +169,7 @@ public void testExpire() {
   @Test
   public void testPassiveExpiration() {
     runCommandAndAssertNoStatUpdates("hash", (k) -> {
-      jedis.expire(k, 1);
+      jedis.expire(k, PassiveExpirationManager.INTERVAL);
       GeodeAwaitility.await().during(Duration.ofSeconds(3)).until(() -> true);
     });
   }
```
sabbey37, line 175: It seems like the intention of this test is to ensure keyspace hits and misses don't update after the `EXPIRE` command and after the key passively expires. We can leave the time we set the key to expire as 1 second (`jedis.expire(k, 1);`).  However, we need to increase the time we are waiting for the key to expire passively since that passive expiration interval has changed.

Perhaps we could change the await to be something like this:

```
 GeodeAwaitility.await()
          .atMost(Duration.ofMinutes(PassiveExpirationManager.INTERVAL))
          .until(() -> jedis.keys("hash").isEmpty());
```
...so we're verifying the key expires before checking the keyspace hits and misses. We will need to add some time in the await to that INTERVAL for a buffer and to avoid potential flakiness.

sabbey37 (changes requested):

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/server/AbstractHitsMissesIntegrationTest.java
ringles: Yeah, I see what you mean now. I'll fix it up!

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/server/AbstractHitsMissesIntegrationTest.java
```diff
@@ -169,8 +169,9 @@ public void testExpire() {
   @Test
   public void testPassiveExpiration() {
     runCommandAndAssertNoStatUpdates("hash", (k) -> {
-      jedis.expire(k, PassiveExpirationManager.INTERVAL);
-      GeodeAwaitility.await().during(Duration.ofSeconds(3)).until(() -> true);
+      jedis.expire(k, 1);
+      GeodeAwaitility.await().during(Duration.ofSeconds(PassiveExpirationManager.INTERVAL))
+          .until(() -> true);
```
sabbey37, line 174: Is there a reason we're not really checking that the key expires here?

We could, for example check the following:
```
GeodeAwaitility.await()
          .atMost(Duration.ofMinutes(PassiveExpirationManager.INTERVAL + 1))
          .until(() -> jedis.keys("hash").isEmpty());
```

...To make sure the key has actually expired before verifying the keyspace hits and misses.  I added 1 minute to the interval to avoid flakiness if passive expiration just ran.  Using `atMost` will also allow the test to move on if the condition evaluates to true before the specified duration.  The `KEYS` command does not update keyspace hits/misses, so we can safely use that to make sure the key has expired.

sabbey37 (changes requested):

sabbey37 (approved):

PR #6076, opened 2021-03-02
GEODE-7309: uplift lucene to 7.1.0

mkevo:
Thank you for submitting a contribution to Apache Geode.

In order to streamline the review of the contribution we ask you
to ensure the following steps have been taken:

### For all changes:
- [x] Is there a JIRA ticket associated with this PR? Is it referenced in the commit message?

- [x] Has your PR been rebased against the latest commit within the target branch (typically `develop`)?

- [x] Is your initial contribution a single, squashed commit?

- [x] Does `gradlew build` run cleanly?

- [x] Have you written or updated unit tests to verify your changes?

- [ ] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)?

### Note:
Please ensure that once the PR is submitted, check Concourse for build issues and
submit an update to your PR as soon as possible. If you need help, please send an
email to dev@geode.apache.org.

PR #6351, opened 2021-04-21
GEODE-7309: uplift lucene to 7.1.0

mkevo:
This is a new PR with rebasing on the latest develop branch.
The last PR was #6076.

Before making a review please reads the following:
[Lucene upgrade dev list](https://markmail.org/message/37swcath75nxxelu#query:+page:1+mid:ygjhsuikdrbuihap+state:results)
[Lucene upgrade CWiki](https://cwiki.apache.org/confluence/display/GEODE/Upgrading+to+Lucene+7.1.0)

### For all changes:

### Note:


geode-lucene/src/upgradeTest/java/org/apache/geode/cache/lucene/LuceneSearchWithRollingUpgradeDUnit.java
```diff
@@ -142,23 +142,17 @@ void executeLuceneQueryWithServerRollOvers(String regionType, String startingVer
 
       server1 = rollServerToCurrentCreateLuceneIndexAndCreateRegion(server1, regionType,
           testingDirs[0], shortcutName, regionName, locatorPorts, reindex);
-      verifyLuceneQueryResultInEachVM(regionName, expectedRegionSize, server1);
       expectedRegionSize += 5;
-      putSerializableObjectAndVerifyLuceneQueryResult(server1, regionName, expectedRegionSize, 5,
-          15, server2, server3);
+      putSerializableObject(server1, regionName, 5, 15);
```
rhoughton-pivot, line 146: My review is mostly for the gradle and pom changes involved in rolling the Lucene dependency version. Seems fine to roll a major-version of our dependency in a minor version of Geode. 

My question here, is why are we not validating the result of this `put` ?

geode-lucene/src/upgradeTest/java/org/apache/geode/cache/lucene/RollingUpgradeQueryReturnsCorrectResultsAfterClientAndServersAreRolledOverAllBucketsCreated.java
```diff
@@ -31,6 +32,7 @@
 public class RollingUpgradeQueryReturnsCorrectResultsAfterClientAndServersAreRolledOverAllBucketsCreated
     extends LuceneSearchWithRollingUpgradeDUnit {
 
+  @Ignore
```
rhoughton-pivot, line 35: Can you comment on why this test is being ignored, and when or how we will re-write or re-enable it again?

rhoughton-pivot (changes requested):

geode-lucene/src/upgradeTest/java/org/apache/geode/cache/lucene/RollingUpgradeQueryReturnsCorrectResultsAfterClientAndServersAreRolledOverAllBucketsCreated.java
mkevo: The main sequence in this test that causes the failure is:


1. start two servers with old version using Lucene 6
2. roll one server to new version server using Lucene 7
3. do puts into primary buckets in new server which creates entries in the fileAndChunk region with Lucene 7 format
4. stop the new version server which causes the old version server to become primary for those buckets
5. do a query which causes the IndexFormatTooNewException to be thrown

So the old server cannot read the new index format. This test can be valid for the version of the Geode which is using Lucene without changes in index format.

geode-lucene/src/upgradeTest/java/org/apache/geode/cache/lucene/LuceneSearchWithRollingUpgradeDUnit.java
mkevo: Between Lucene version happened changes on Lucene format, so it cannot read with old server new indexes. So we agree on the dev list to queue puts until all members are upgraded. You can find it in the description of this PR.

jinmeiliao (approved):
assuming all checks passed

geode-lucene/src/upgradeTest/java/org/apache/geode/cache/lucene/LuceneQueryWithDifferentVersions.java
```diff
@@ -0,0 +1,101 @@
...
+import org.apache.geode.test.dunit.Host;
+import org.apache.geode.test.dunit.NetworkUtils;
+import org.apache.geode.test.dunit.VM;
+
+public class LuceneQueryWithDifferentVersions
+    extends LuceneSearchWithRollingUpgradeDUnit {
+
+  // 2 locator, 2 servers
+  @Test
+  public void luceneQueryCannotBeExecuted()
```
DonalEvans, line 36: This test name could be more descriptive. Indicating specifically when a Lucene query cannot be executed would make it a lot clearer what the test is doing.

geode-lucene/src/upgradeTest/java/org/apache/geode/cache/lucene/LuceneQueryWithDifferentVersions.java
```diff
@@ -0,0 +1,101 @@
...
+      locator1.invoke(
+          invokeStartLocator(hostName, locatorPorts[0], getLocatorPropertiesPre91(locatorString)));
+      locator2.invoke(
+          invokeStartLocator(hostName, locatorPorts[1], getLocatorPropertiesPre91(locatorString)));
+      invokeRunnableInVMs(invokeCreateCache(getSystemProperties(locatorPorts)), server1, server2);
+
+      // Locators before 1.4 handled configuration asynchronously.
+      // We must wait for configuration configuration to be ready, or confirm that it is disabled.
+      locator1.invoke(
+          () -> await().untilAsserted(() -> await().untilAsserted(() -> {
```
DonalEvans, line 64: Why are two `await()` calls being chained here?

geode-lucene/src/upgradeTest/java/org/apache/geode/cache/lucene/LuceneQueryWithDifferentVersions.java
```diff
@@ -0,0 +1,101 @@
...
+            assertThat(InternalLocator.getLocator().isSharedConfigurationRunning()).isTrue();
+          })));
+      locator2.invoke(
+          () -> await()
+              .untilAsserted(() -> await().untilAsserted(() -> {
```
DonalEvans, line 69: Another chained `await()` call.

geode-lucene/src/upgradeTest/java/org/apache/geode/cache/lucene/LuceneQueryWithDifferentVersions.java
```diff
@@ -0,0 +1,101 @@
...
+
+      server1 = rollServerToCurrentCreateLuceneIndexAndCreateRegion(server1, regionType, null,
+          shortcut.name(), regionName, locatorPorts, reindex);
+
+      Throwable thrown = catchThrowable(() -> {
+        putSerializableObjectAndVerifyLuceneQueryResult(server2, regionName, 20, 15,
+            25, server2);
+      });
+
+      assertThat(thrown.getCause()).isInstanceOf(AssertionError.class);
```
DonalEvans, line 94: This is the only assertion that this test makes, and it's very vague. Would it be possible to be more specific in terms of what exactly the test expects to fail, and where? Instead of asserting that some other assertion from `putSerializableObjectAndVerifyLuceneQueryResult()` fails, it would be better to have a specific assertion for this test showing exactly where and how we expect the query verification to fail.

geode-lucene/src/upgradeTest/java/org/apache/geode/cache/lucene/RollingUpgradeQueryReturnsCorrectResultsAfterClientAndServersAreRolledOverAllBucketsCreated.java
DonalEvans: If it's not possible to rewrite this test to allow it to pass in all cases, it would be better that it simply be removed. However, it seems like it should be doable to add some logic to determine the expected behaviour of the query based on the version of Geode under test (and hence the version of Lucene).

geode-lucene/src/main/java/org/apache/geode/cache/lucene/internal/repository/IndexRepositoryImpl.java
```diff
@@ -227,4 +230,31 @@ public int getAsInt() {
       }
     }
   }
+
+  private static class Similarity extends TFIDFSimilarity {
```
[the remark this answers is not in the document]
mkevo: It is similar to the `ClassicSimialrity`, but with some changes to adapt it to Geode. 
You can verify it by running `DistributedScoringJUnitTest`

geode-lucene/src/main/java/org/apache/geode/cache/lucene/internal/LuceneServiceImpl.java
```diff
@@ -238,7 +238,7 @@ public void createIndex(final String indexName, String regionPath, final Analyze
     }
   }
 
-  protected void validateAllMembersAreTheSameVersion(PartitionedRegion region) {
+  public void validateAllMembersAreTheSameVersion(PartitionedRegion region) {
```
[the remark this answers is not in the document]
mkevo: I changed it back to the `protected`.

geode-lucene/src/upgradeTest/java/org/apache/geode/cache/lucene/LuceneQueryWithDifferentVersions.java
```diff
@@ -0,0 +1,101 @@
...
+
+import org.apache.geode.cache.RegionShortcut;
+import org.apache.geode.distributed.internal.InternalLocator;
+import org.apache.geode.internal.AvailablePortHelper;
+import org.apache.geode.test.dunit.DistributedTestUtils;
+import org.apache.geode.test.dunit.Host;
+import org.apache.geode.test.dunit.NetworkUtils;
+import org.apache.geode.test.dunit.VM;
+
+public class LuceneQueryWithDifferentVersions
```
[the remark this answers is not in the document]
mkevo: Done!

geode-lucene/src/upgradeTest/java/org/apache/geode/cache/lucene/LuceneQueryWithDifferentVersions.java
mkevo: Done!

geode-lucene/src/upgradeTest/java/org/apache/geode/cache/lucene/LuceneQueryWithDifferentVersions.java
```diff
@@ -0,0 +1,101 @@
...
+
+    final String regionName = "aRegion";
+    RegionShortcut shortcut = RegionShortcut.PARTITION_REDUNDANT;
+    String regionType = "partitionedRedundant";
+
+    int[] locatorPorts = AvailablePortHelper.getRandomAvailableTCPPorts(2);
+    locator1.invoke(() -> DistributedTestUtils.deleteLocatorStateFile(locatorPorts));
+    locator2.invoke(() -> DistributedTestUtils.deleteLocatorStateFile(locatorPorts));
+
+    String hostName = NetworkUtils.getServerHostName(host);
```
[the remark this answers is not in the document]
mkevo: Done! Thanks!

geode-lucene/src/upgradeTest/java/org/apache/geode/cache/lucene/LuceneQueryWithDifferentVersions.java
mkevo: By mistake! Thanks!

geode-lucene/src/upgradeTest/java/org/apache/geode/cache/lucene/LuceneQueryWithDifferentVersions.java
mkevo: Also by mistake! Thanks!

geode-lucene/src/upgradeTest/java/org/apache/geode/cache/lucene/LuceneQueryWithDifferentVersions.java
mkevo: Thi test checks that it is impossible to do lucene query with different version of the servers. It will failed in ` void verifyLuceneQueryResults(String regionName, int expectedRegionSize)` while doing the following assertion:
`    assertTrue((Boolean) luceneService.getClass()
        .getMethod("waitUntilFlushed", String.class, String.class, long.class, TimeUnit.class)
        .invoke(luceneService, INDEX_NAME, regionName, 60, TimeUnit.SECONDS));`
I'm not sure how to change this assertion in the test to verify this assertion will failed.

geode-core/src/main/java/org/apache/geode/internal/cache/xmlcache/CacheCreation.java
```diff
@@ -1115,6 +1116,11 @@ public void unlockDiskStore(String diskStoreName) {
 
   }
 
+  @Override
+  public boolean hasMemberOlderThan(KnownVersion version) {
+    return false;
```
DonalEvans, line 1121: I wonder if, for this method, it would make sense to follow the approach used elsewhere in this class where we throw an UnsupportedOperationException rather than returning an effectively meaningless value. I think that there's no scenario in which it would make sense to call this method on a `CacheCreation`.

DonalEvans:
Having thought about these changes some more, I'm wondering if it's necessary to be quite so strict with the Geode version check (such as in IndexRepositoryFactory and LuceneEventListener). Certainly on the current develop branch, any older version of Geode will cause problems if a Lucene query is attempted, but once 1.15.0 has been released, there should be no problem with rolling from 1.15.0 to a newer version, providing the Lucene version hasn't changed again. It might be better to instead check that there is no member older than the version in which Lucene was updated (1.15.0, assuming we don't cut the release branch before these changes are merged). I understand that this would be more work, and would mean that RollingUpgrade tests would become more complex, as they'd have to cover both the failing and passing cases, but if you agree that there would be value in only preventing Lucene queries when we know there's a potential problem rather than the more blunt approach currently used, then I'd appreciate the change.

geode-lucene/src/upgradeTest/java/org/apache/geode/cache/lucene/LuceneQueryWithDifferentVersions.java
DonalEvans: It looks like the failure is happening before the Lucene query even begins, since `waitUntilFlushed()` is returning false rather than the expected true. This appears to be because the wait times out, and I also see logging related to stuck Pooled Waiting Message Processor threads, and the message "During normal processing, unsuccessfully dispatched 1 events (batch #0)" being logged repeatedly. This seems like unwanted behaviour, that should be looked into further, as we shouldn't be getting stuck like this if the safeguards in this PR are working as intended.

geode-core/src/main/java/org/apache/geode/internal/cache/xmlcache/CacheCreation.java
```diff
@@ -1124,6 +1125,14 @@ public void setQueryConfigurationServiceCreation(
     this.queryConfigurationServiceCreation = queryConfigurationServiceCreation;
   }
 
+  @Override
+  public boolean hasMemberOlderThan(KnownVersion version) {
+    return getMembers().stream()
```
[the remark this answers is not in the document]
DonalEvans: My apologies for not noticing this the first time around, but it seems like many other methods in CacheCreation throw an UnsupportedOperationException when called if it doesn't really make sense to call them on a Cache that's in the process of being created. I think this method should probably follow the same approach, as it doesn't seem like we're able to actually say whether the value returned is correct or not.

geode-lucene/src/upgradeTest/java/org/apache/geode/cache/lucene/RollingUpgradeQueryReturnsCorrectResultsAfterClientAndServersAreRolledOverAllBucketsCreated.java
mkevo: What did you mean by this logic to determine the expected behaviour of the query based on the Geode version?  In versions before Lucene is upgraded, the test(always doing the upgrade to the newest version) is not changed and it works. Maybe you think that we need this change for the future version of Geode when Lucene format will be the same, and test between that versions should work as before this changes? Or you mean something else?

geode-lucene/src/upgradeTest/java/org/apache/geode/cache/lucene/LuceneQueryWithDifferentVersions.java
mkevo: Yes, I agree that this message "During normal processing,..." should be avoided. I will try to investigate how to fix this part.

mkevo:
> Having thought about these changes some more, I'm wondering if it's necessary to be quite so strict with the Geode version check (such as in IndexRepositoryFactory and LuceneEventListener). Certainly on the current develop branch, any older version of Geode will cause problems if a Lucene query is attempted, but once 1.15.0 has been released, there should be no problem with rolling from 1.15.0 to a newer version, providing the Lucene version hasn't changed again. It might be better to instead check that there is no member older than the version in which Lucene was updated (1.15.0, assuming we don't cut the release branch before these changes are merged). I understand that this would be more work, and would mean that RollingUpgrade tests would become more complex, as they'd have to cover both the failing and passing cases, but if you agree that there would be value in only preventing Lucene queries when we know there's a potential problem rather than the more blunt approach currently used, then I'd appreciate the change.

Yes, there should no be a problem with rolling upgrade to a newer version once 1.15.0 is released, until there are no format changes in new Lucene versions. 
Maybe we can change tests to have some check to see if there is a member older than 1.15.0, and regarding the result run query or ignore it until all members are on a version where Lucene was updated, or newer.

geode-lucene/src/upgradeTest/java/org/apache/geode/cache/lucene/RollingUpgradeQueryReturnsCorrectResultsAfterClientAndServersAreRolledOverAllBucketsCreated.java
DonalEvans: Yeah, I think it would be good to have two paths for the test. One where we know there is a Lucene version mismatch, where we assert that the query fails or throws (or whatever the expected behaviour ends up being in that situation) and one where we know that the Lucene version is the same on both old and new servers (like after 1.15 is released with the Lucene version changed, and we can start rolling from 1.15 to 1.16) where we assert that the query succeeds.

DonalEvans:
> > Having thought about these changes some more, I'm wondering if it's necessary to be quite so strict with the Geode version check (such as in IndexRepositoryFactory and LuceneEventListener). Certainly on the current develop branch, any older version of Geode will cause problems if a Lucene query is attempted, but once 1.15.0 has been released, there should be no problem with rolling from 1.15.0 to a newer version, providing the Lucene version hasn't changed again. It might be better to instead check that there is no member older than the version in which Lucene was updated (1.15.0, assuming we don't cut the release branch before these changes are merged). I understand that this would be more work, and would mean that RollingUpgrade tests would become more complex, as they'd have to cover both the failing and passing cases, but if you agree that there would be value in only preventing Lucene queries when we know there's a potential problem rather than the more blunt approach currently used, then I'd appreciate the change.
> 
> Yes, there should no be a problem with rolling upgrade to a newer version once 1.15.0 is released, until there are no format changes in new Lucene versions.
> Maybe we can change tests to have some check to see if there is a member older than 1.15.0, and regarding the result run query or ignore it until all members are on a version where Lucene was updated, or newer.

Yeah, I think this would be the best course of action.

geode-lucene/src/upgradeTest/java/org/apache/geode/cache/lucene/LuceneSearchWithRollingUpgradeTestBase.java
```diff
@@ -225,12 +224,33 @@ protected static ClientCache createClientCache(Properties systemProperties,
...
+          currentVersionOrdinal = productVersion.ordinal();
+        } else if (vm.getVersion().equals(VersionManager.CURRENT_VERSION)) {
+          currentVersionOrdinal = KnownVersion.CURRENT_ORDINAL;
+        }
+      }
+      if (currentVersionOrdinal < KnownVersion.GEODE_1_15_0.ordinal()) {
+        return true;
+      }
+    }
+    return false;
```
DonalEvans, line 253: This might be simpler as:
```
    return host.getAllVMs().get(0).invoke(() -> {
      List<InternalDistributedMember> allMembers = InternalDistributedSystem
          .getConnectedInstance()
          .getDistributionManager()
          .getViewMembers();

      for (InternalDistributedMember member : allMembers) {
        if (member.getVersionOrdinal() < KnownVersion.GEODE_1_15_0.ordinal()) {
          return true;
        }
      }
      return false;
    });
```

DonalEvans (changes requested):
This PR no longer has any tests added to verify the changed behaviour in `IndexRepositoryFactory`. It's not enough to just skip verification if there is a Lucene version mismatch; there should be tests to assert what the expected behaviour is when Lucene versions are different, otherwise there is a big gap in our testing.

mkevo:
> This PR no longer has any tests added to verify the changed behaviour in `IndexRepositoryFactory`. It's not enough to just skip verification if there is a Lucene version mismatch; there should be tests to assert what the expected behaviour is when Lucene versions are different, otherwise there is a big gap in our testing.

Yes, I'm working on that.

geode-lucene/src/upgradeTest/java/org/apache/geode/cache/lucene/LuceneSearchWithRollingUpgradeTestBase.java
mkevo: The problem with this is on older versions before getDistributionManager is introduced, and it failed with NoSuchMethodError. For the newer version(after 1.5.0) it throws NullPointerException in the same place.

DonalEvans: I see, thanks for the explanation.

mkevo:
Need someone to change a timeout of stressNewTest and re-run it, to have more time to run it all. Thanks!

onichols-pivotal:
> Need someone to change a timeout of stressNewTest and re-run it, to have more time to run it all. Thanks!

Boosted timeouts by 10x and retriggered, hope it doesn't actually take 4 days ;)

kirklund (approved):

DonalEvans (approved):

rhoughton-pivot (approved):

rhoughton-pivot:
@mkevo If you do the rebase and force-push to resolve the `assembly-content.txt` issue listed, then we can get this merged :) The change should be the addition of `lucene-backward-codecs` to the test assembly.

mkevo:
@onichols-pivotal, can you please change the timeout for stressNewTest again?

onichols-pivotal:
> @onichols-pivotal, can you please change the timeout for stressNewTest again?

temporarily increased to 24h

mkevo:
Hi @onichols-pivotal , can you please retrigger a job so we got all green? It seems that the connection is closed by some remote host.

onichols-pivotal:
> Hi @onichols-pivotal , can you please retrigger a job so we got all green? It seems that the connection is closed by some remote host.

I think you are running into some other timeout. I’ve asked around but folks who would know more are on vacation.

mkevo:
Hi @onichols-pivotal, can you please increase the timeout for the StressTest and re-trigger the jobs?

onichols-pivotal:
> please increase the timeout for the StressTest and re-trigger
These checks are not required to merge…

PR #6356, opened 2021-04-21
add 1.12.2 to old versions on develop

onichols-pivotal:


rhoughton-pivot (approved):
This will pass, as soon as the artifact is uploaded.

PR #6359, opened 2021-04-21
GEODE-9158: Add CLUSTER NODES, SLOTS and INFO commands

jdeppe-pivotal:
- None of the commands will return 'replica' information since Geode does
  make the distinction of a primary vs a secondary member.

### For all changes:

### Note:

### DIFF
--- a/geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/cluster/ClusterSlotsAndNodesDUnitTest.java
+++ b/geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/cluster/ClusterSlotsAndNodesDUnitTest.java
@@ -1,0 +1,164 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package org.apache.geode.redis.internal.executor.cluster;
+
+import static org.apache.geode.redis.internal.RegionProvider.REDIS_REGION_BUCKETS;
+import static org.apache.geode.redis.internal.RegionProvider.REDIS_SLOTS_PER_BUCKET;
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.List;
+
+import org.assertj.core.data.Offset;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Test;
+import redis.clients.jedis.Jedis;
+
+import org.apache.geode.cache.control.RebalanceFactory;
+import org.apache.geode.cache.control.ResourceManager;
+import org.apache.geode.redis.ClusterNode;
+import org.apache.geode.redis.ClusterNodes;
+import org.apache.geode.test.awaitility.GeodeAwaitility;
+import org.apache.geode.test.dunit.rules.ClusterStartupRule;
+import org.apache.geode.test.dunit.rules.MemberVM;
+import org.apache.geode.test.dunit.rules.RedisClusterStartupRule;
+
+public class ClusterSlotsAndNodesDUnitTest {
+
+  @ClassRule
+  public static RedisClusterStartupRule cluster = new RedisClusterStartupRule();
+
+  private static final int JEDIS_TIMEOUT =
+      Math.toIntExact(GeodeAwaitility.getTimeout().toMillis());
+  private static final String LOCAL_HOST = "127.0.0.1";
+  private static MemberVM locator;
+  private static MemberVM server1;
+  private static MemberVM server2;
+
+  private static Jedis jedis1;
+  private static Jedis jedis2;
+
+  @BeforeClass
+  public static void classSetup() {
+    locator = cluster.startLocatorVM(0);
+    server1 = cluster.startRedisVM(1, locator.getPort());
+    server2 = cluster.startRedisVM(2, locator.getPort());
+
+    int redisServerPort1 = cluster.getRedisPort(1);
+    int redisServerPort2 = cluster.getRedisPort(2);
+
+    jedis1 = new Jedis(LOCAL_HOST, redisServerPort1, JEDIS_TIMEOUT);
+    jedis2 = new Jedis(LOCAL_HOST, redisServerPort2, JEDIS_TIMEOUT);
+  }
+
+  @Test
+  public void eachServerProducesTheSameNodeInformation() {
+    List<ClusterNode> nodes1 = ClusterNodes.parseClusterNodes(jedis1.clusterNodes()).getNodes();
+    assertThat(nodes1).hasSize(2);
+
+    List<ClusterNode> nodes2 = ClusterNodes.parseClusterNodes(jedis2.clusterNodes()).getNodes();
+    assertThat(nodes2).hasSize(2);
+
+    assertThat(nodes1).containsExactlyInAnyOrderElementsOf(nodes2);
+  }
+
+  @Test
+  public void eachServerProducesTheSameSlotInformation() {
+    List<Object> slots1 = jedis1.clusterSlots();
+    List<Object> slots2 = jedis2.clusterSlots();
+
+    assertThat(slots1).usingRecursiveComparison().isEqualTo(slots2);
+  }
+
+  @Test
+  @SuppressWarnings("unchecked")
+  public void slotInformationIsCorrect() {
+    List<Object> slots = jedis1.clusterSlots();
+
+    assertThat(slots).hasSize(REDIS_REGION_BUCKETS);
+
+    for (int i = 0; i < REDIS_REGION_BUCKETS; i++) {
+      long slotStart = (long) ((List<Object>) slots.get(i)).get(0);
+      long slotEnd = (long) ((List<Object>) slots.get(i)).get(1);
+
+      assertThat(slotStart).isEqualTo((long) i * REDIS_REGION_BUCKETS);
+      assertThat(slotEnd).isEqualTo((long) i * REDIS_REGION_BUCKETS + (REDIS_SLOTS_PER_BUCKET - 1));
+    }
+  }
+
+  @Test
+  public void slotsDistributionIsFairlyUniform() {
+    List<Object> slots = jedis1.clusterSlots();
+    List<ClusterNode> nodes = ClusterNodes.parseClusterSlots(slots).getNodes();
+
+    assertThat(nodes.get(0).slots.size()).isCloseTo(REDIS_REGION_BUCKETS / 2, Offset.offset(2));
+    assertThat(nodes.get(1).slots.size()).isCloseTo(REDIS_REGION_BUCKETS / 2, Offset.offset(2));
+  }
+
+  @Test
+  public void whenAServerIsAddedOrRemoved_slotsAreRedistributed() {
+    cluster.startRedisVM(3, locator.getPort());
+    rebalanceAllRegions(server1);
+
+    List<Object> slots = jedis1.clusterSlots();
+    List<ClusterNode> nodes = ClusterNodes.parseClusterSlots(slots).getNodes();
+
+    assertThat(nodes).as("incorrect number of nodes").hasSize(3);
+    assertThat(nodes.get(0).slots.size()).isCloseTo(REDIS_REGION_BUCKETS / 3, Offset.offset(2));
+    assertThat(nodes.get(1).slots.size()).isCloseTo(REDIS_REGION_BUCKETS / 3, Offset.offset(2));
+    assertThat(nodes.get(2).slots.size()).isCloseTo(REDIS_REGION_BUCKETS / 3, Offset.offset(2));
+
+    String info = jedis1.clusterInfo();
+    assertThat(info).contains("cluster_known_nodes:3");
+    assertThat(info).contains("cluster_size:3");
+
+    cluster.crashVM(3);
+    rebalanceAllRegions(server1);
+
+    slots = jedis1.clusterSlots();
+    nodes = ClusterNodes.parseClusterSlots(slots).getNodes();
+
+    assertThat(nodes).as("incorrect number of nodes").hasSize(2);
+    assertThat(nodes.get(0).slots.size()).isCloseTo(REDIS_REGION_BUCKETS / 2, Offset.offset(2));
+    assertThat(nodes.get(1).slots.size()).isCloseTo(REDIS_REGION_BUCKETS / 2, Offset.offset(2));
+
+    info = jedis1.clusterInfo();
+    assertThat(info).contains("cluster_known_nodes:2");
+    assertThat(info).contains("cluster_size:2");
+  }
+
+  @Test
+  public void clusterSlotsAndClusterNodesResponseIsEquivalent() {
+    List<ClusterNode> nodesFromSlots =
+        ClusterNodes.parseClusterSlots(jedis1.clusterSlots()).getNodes();
+    List<ClusterNode> nodesFromNodes =
+        ClusterNodes.parseClusterNodes(jedis1.clusterNodes()).getNodes();
+
+    assertThat(nodesFromSlots).containsExactlyInAnyOrderElementsOf(nodesFromNodes);
+  }
+
+  private static void rebalanceAllRegions(MemberVM vm) {
+    vm.invoke(() -> {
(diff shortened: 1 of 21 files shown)


onichols-pivotal (approved):

sabbey37 (commented):
I think it looks good. Just some small questions and suggestions.  I am also adding @upthewaterspout and @pivotal-jbarrett as reviewers since they have likely been thinking about and working with Redis CLUSTER mode more than I have.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/ParameterRequirements/ClusterParameterRequirements.java
```diff
@@ -0,0 +1,47 @@
...
+    } else if (numberOfArguments == 2) {
+      confirmKnownSubcommands(command);
+    } else { // numberOfArguments > 3
+      throw new RedisParametersMismatchException(
+          String.format(ERROR_UNKNOWN_CLUSTER_SUBCOMMAND, command.getStringKey()));
+    }
+  }
+
+  private void confirmKnownSubcommands(Command command) {
+    if (!command.getStringKey().equalsIgnoreCase("slots") &&
```
pivotal-jbarrett, line 39: While this command shouldn't be frequently invoked this pattern should be avoided here and elsewhere in the code.

`String.equalsIgnoreCase` has to convert mismatching characters to uppercase to compare them. In this case we invoked this method up to 3 times. An alternative:

```java
final String subCommand = command.getStringKey().toLower();
if (!subCommand.equals("slots") && ...) {
  ...
}
```
At most converts the string to lower case 1 time. 

Extra credit for avoid the string conversation in `getStringKey` and look at the just bytes of this command component directly. 

```java
final byte[] subCommand = getProcessedCommand().get(1);
if (subCommand[0] == 's' || subCommand[0] == 'S') {
  if (subCommand[1] == 'l' || subCommand[1] == 'L') {
    ..
  }
} else if (subCommand[0] == 'I' ...) {
} else {
  throw ...;
}
```

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/cluster/BucketInfoRetrievalFunction.java
```diff
@@ -0,0 +1,108 @@
...
+import org.apache.geode.internal.cache.LocalDataSet;
+import org.apache.geode.internal.cache.execute.InternalFunction;
+import org.apache.geode.internal.inet.LocalHostUtil;
+import org.apache.geode.redis.internal.RegionProvider;
+import org.apache.geode.redis.internal.data.ByteArrayWrapper;
+import org.apache.geode.redis.internal.data.RedisKey;
+
+public class BucketInfoRetrievalFunction implements InternalFunction<Void> {
+
+  public static final String ID = "REDIS_BUCKET_SLOT_FUNCTION";
```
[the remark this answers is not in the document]
pivotal-jbarrett:
```java
public static final String ID = BucketInfoRetrievalFunction.class.getName();
```

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/cluster/ClusterExecutor.java
```diff
@@ -0,0 +1,199 @@
...
+    switch (subCommand.toLowerCase()) {
+      case "info":
+        return getInfo(context);
+      case "nodes":
+        return getNodes(context);
+      case "slots":
+        return getSlots(context);
+      default: {
+        return RedisResponse.error(
+            String.format(ERROR_UNKNOWN_CLUSTER_SUBCOMMAND, subCommand));
```
pivotal-jbarrett, line 70: If we do this check for sub commands here again but do we do them in the parameter check too? Seems like a waste to do it in two places.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/cluster/ClusterExecutor.java
```diff
@@ -0,0 +1,199 @@
...
+
+      slots.add(entry);
+      index++;
+    }
+
+    return RedisResponse.array(slots);
+  }
+
+  @SuppressWarnings("unchecked")
+  private List<MemberBuckets> getMemberBuckets(
```
pivotal-jbarrett, line 115: Is there some way for us to calculate the bucket/slot arrangement upfront and only update it on bucket movement? Is there a listener we could attach to for these events? This way we don't need to generate this data structure for scratch for each client request.

pivotal-jbarrett (changes requested):

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/cluster/ClusterSlotsAndNodesDUnitTest.java
```diff
@@ -0,0 +1,164 @@
...
+  private static MemberVM server2;
+
+  private static Jedis jedis1;
+  private static Jedis jedis2;
+
+  @BeforeClass
+  public static void classSetup() {
+    locator = cluster.startLocatorVM(0);
+    server1 = cluster.startRedisVM(1, locator.getPort());
+    server2 = cluster.startRedisVM(2, locator.getPort());
```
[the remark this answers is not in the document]
jdeppe-pivotal: Fixed.

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/cluster/JedisAndLettuceClusterDUnitTest.java
```diff
@@ -0,0 +1,94 @@
...
+  @ClassRule
+  public static RedisClusterStartupRule cluster = new RedisClusterStartupRule();
+
+  private static final int KEYS = 1000;
+  private static final int JEDIS_TIMEOUT =
+      Math.toIntExact(GeodeAwaitility.getTimeout().toMillis());
+  private static final String LOCAL_HOST = "127.0.0.1";
+  private static MemberVM locator;
+  private static MemberVM server1;
+  private static MemberVM server2;
```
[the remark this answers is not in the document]
jdeppe-pivotal: Fixed.

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/cluster/JedisAndLettuceClusterDUnitTest.java
```diff
@@ -0,0 +1,94 @@
...
+  private static int redisServerPort2;
+
+  @BeforeClass
+  public static void classSetup() {
+    locator = cluster.startLocatorVM(0);
+    server1 = cluster.startRedisVM(1, locator.getPort());
+    server2 = cluster.startRedisVM(2, locator.getPort());
+
+    redisServerPort1 = cluster.getRedisPort(1);
+    redisServerPort2 = cluster.getRedisPort(2);
```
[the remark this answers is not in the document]
jdeppe-pivotal: Fixed.

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/cluster/JedisAndLettuceClusterDUnitTest.java
```diff
@@ -0,0 +1,94 @@
...
+public class JedisAndLettuceClusterDUnitTest {
+
+  @ClassRule
+  public static RedisClusterStartupRule cluster = new RedisClusterStartupRule();
+
+  private static final int KEYS = 1000;
+  private static final int JEDIS_TIMEOUT =
+      Math.toIntExact(GeodeAwaitility.getTimeout().toMillis());
+  private static final String LOCAL_HOST = "127.0.0.1";
+  private static MemberVM locator;
```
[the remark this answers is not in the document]
jdeppe-pivotal: Fixed.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/cluster/BucketInfoRetrievalFunction.java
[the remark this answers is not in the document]
jdeppe-pivotal: Fixed.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/ParameterRequirements/ClusterParameterRequirements.java
jdeppe-pivotal: Removed this check in favor of checking in the `Executor`.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/cluster/ClusterExecutor.java
jdeppe-pivotal: Removing this error response return would require returning `null` which just doesn't look right to me (although functionally it will never get hit). However I've chosen to remove the subcommand check from the param validator in favor of keeping this code. This also means there is just one place to update when subsequent subcommands are added.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/cluster/ClusterExecutor.java
jdeppe-pivotal: I'm planning to add that when we address returning `MOVED` responses. which will definitely require something lighter and more dynamic.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/ParameterRequirements/ClusterParameterRequirements.java
jdeppe-pivotal: Removed all of this in favor of just leaving the check in `ClusterExecutor`.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/RedisConstants.java
```diff
@@ -54,4 +54,6 @@
 
   public static final String ERROR_UNKNOWN_SLOWLOG_SUBCOMMAND =
       "Unknown subcommand or wrong number of arguments for '%s'. Try SLOWLOG HELP.";
+  public static final String ERROR_UNKNOWN_CLUSTER_SUBCOMMAND =
+      "Unknown subcommand or wrong number of arguments for '%s'. Try CLUSTER HELP.";
```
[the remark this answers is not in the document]
jdeppe-pivotal: I've added a story.

rhoughton-pivot (approved):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/cluster/ClusterExecutor.java
```diff
@@ -0,0 +1,196 @@
...
+    // added yet.
+    PartitionRegionHelper.assignBucketsToPartitions(dataRegion);
+
+    Set<DistributedMember> membersWithDataRegion = new HashSet<>();
+    for (PartitionMemberInfo memberInfo : getRegionMembers(ctx)) {
+      membersWithDataRegion.add(memberInfo.getDistributedMember());
+    }
+
+    ResultCollector<MemberBuckets, List<MemberBuckets>> resultCollector =
+        FunctionService.onMembers(membersWithDataRegion).execute(BucketInfoRetrievalFunction.ID);
```
upthewaterspout, line 126: It's somewhat unfortunate we need to use a function at all to get this information. The location of the *buckets* is actually known to all members of the PR, so we can find out where all the primaries are without a function. I think the part we don't know is the redis bind address on the other members.

Is it worth improving the performance of these commands, or are they so infrequent we don't care? If it is worth it, we can add another story to figure out how to distribute the redis bind addresses (or maybe just cache them?).

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/cluster/ClusterExecutor.java
```diff
@@ -0,0 +1,196 @@
...
+  }
+
+
+  private RedisResponse getInfo(ExecutionHandlerContext ctx) {
+    int memberCount = getRegionMembers(ctx).size();
+
+    return RedisResponse.bulkString(
+        "cluster_state:ok\r\n"
+            + "cluster_slots_assigned:" + REDIS_SLOTS + "\r\n"
+            + "cluster_slots_ok:" + REDIS_SLOTS + "\r\n"
```
upthewaterspout, line 178: At some point we may actually have buckets that are offline or missing. I don't know if there is any value in computing these numbers instead of hardcoding them. Maybe not right now.

upthewaterspout (commented):
Looks pretty good! I had a few comments inline.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/cluster/BucketInfoRetrievalFunction.java
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  @Override
+  public void execute(FunctionContext<Void> context) {
+    Region<RedisKey, ByteArrayWrapper> region =
+        context.getCache().getRegion(RegionProvider.REDIS_DATA_REGION);
+
+    String memberId =
+        context.getCache().getDistributedSystem().getDistributedMember().getUniqueId();
+    LocalDataSet localPrimary = (LocalDataSet) PartitionRegionHelper.getLocalPrimaryData(region);
```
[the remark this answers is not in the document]
jdeppe-pivotal: I've updated this to use `getAllLocalPrimaryBucketIds`.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/cluster/ClusterExecutor.java
```diff
@@ -0,0 +1,196 @@
...
+  }
+
+  @SuppressWarnings("unchecked")
+  private List<MemberBuckets> getMemberBuckets(
+      ExecutionHandlerContext ctx) {
+    Region<RedisKey, RedisData> dataRegion = ctx.getRegionProvider().getDataRegion();
+
+    // Really only need this in situations where the cluster is empty and no data has been
+    // added yet.
+    PartitionRegionHelper.assignBucketsToPartitions(dataRegion);
```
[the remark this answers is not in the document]
jdeppe-pivotal: Wrapped in a check now.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/cluster/ClusterExecutor.java
jdeppe-pivotal: ATM I think it's OK since these are called once when a connection is established by a cluster-aware client and then periodically (every few seconds maybe?) to check that the topology has not changed.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/cluster/ClusterExecutor.java
jdeppe-pivotal: I think implementing `MOVED` will introduce a framework to track this detail better. At which point we can improve this.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/cluster/ClusterExecutor.java
upthewaterspout: Ok. Well, maybe we should stick an enhancement in the backlog somewhere. Every few seconds doesn't seem like a lot, until there are 1000s of clients doing that. Doesn't need to hold up this merge.

upthewaterspout (approved):

sabbey37 (approved):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/cluster/ClusterExecutor.java
pivotal-jbarrett: It isn't really periodic though, If a bucket move occurs or a client otherwise has incorrect slot assignments we need to send MOVED, which will need to calculate the current bucket placement. @jdeppe-pivotal said there is a optimization coming to address this though.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/cluster/ClusterExecutor.java
pivotal-jbarrett: This optimization should include some way of sharing hostname and port information for all the Redis endpoints. Perhaps it makes sense for this to go into some Redis metadata replicated region.

pivotal-jbarrett (approved):

PR #6360, opened 2021-04-21
GEODE-9180: heartbeat producer logs a warning when it misses beats

Bill:
[GEODE-9180](https://issues.apache.org/jira/browse/GEODE-9180)

### Problem Description

Sometimes we see a member force-disconnected and we see a preceding gap in the regular sequence of heartbeats generated by the member, but we can't explain why there was a gap. The explanation we are searching for is usually CPU saturation. We look for secondary evidence such as gaps in the regular sequence of statistics e.g. StatSampler sampleCount. When we can't find such secondary evidence, we can't, in good conscience, rule out bugs in the heartbeat generation logic itself.

The heartbeat generation logic consists mainly of a thread that loops forever. Each time through the loop it sleeps for member-timeout / logical-interval. By default that's 5s / 2 = 2.5s. When it wakes up it sends unreliable UDP unicast messages to the coordinator and the two non-coordinator members to its "left" (earlier) in the view. If that heartbeat generation thread oversleeps or doesn't get adequate time slices when it's awake then heartbeats will be delayed. There will be gaps in the regular sequence.

When this ticket is complete, a warning-level message will be logged if the heartbeat generation thread (see GMSHealthMonitor.startHeartbeatThread()) oversleeps by more than the sleep interval (member-timeout / logical-interval), i.e. if it is asleep for more than 2 * (member-timeout / logical-interval), the warning will be logged.

### Solution Description

This was a pretty simple addition to `GMSHealthMonitor.sendPeriodicHeartbeats()` to log a warning if we oversleep in our heartbeat-generation loop.

In an effort to make this new logic testable I turned what was an anonymous inner class into a non-anonymous one, `Heart`. Then I parameterized `sendPeriodicHeartbeats()` on a sleeper object, a timer object and a warner object. With the setup already in place in `GMSHealthMonitorJUnitTest` this was sufficient to let me test that method in isolation, and deterministically. 

I vacillated between using `System.currentTimeMillis()` which can move backward (it isn't monotonic) and is a relatively more expensive than nanos, but is the norm in `GMSHealthMonitor`, vs `System.nanoTime()` which never moves backward and is more efficient than millis. I picked nanos here. In the end I broke with tradition in `GMSHealthMonitor` because I wanted monotonicity and while I was at it, efficiency (even though this path isn't performance sensitive). **I wonder if there are situations e.g. virtualization situations, where `nanoTime()` fails to move forward when it should?**

### Questions for Reviewers

1. We warn if we oversleep by more than one period. Is that the right criterion?
2. What do you think of injecting functions like this to make the method testable, as opposed to adding more mocking?

### Checklist


Bill:
@kamilla1201 will you please review this PR?

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/fd/GMSHealthMonitor.java
```diff
@@ -1538,4 +1463,114 @@ public void run() {
...
+  interface Warner {
+    void warn(String message);
+  }
+
+  class Heart implements Runnable {
+
+    public final long sleepPeriodMillis = memberTimeout / LOGICAL_INTERVAL;
+    public final long sleepPeriodNanos =
+        TimeUnit.NANOSECONDS.convert(sleepPeriodMillis, TimeUnit.MILLISECONDS);
+    public final long sleepLimitNanos = 2 * sleepPeriodNanos;
```
kamilla1201, line 1487: I don't quite understand what `sleepLimitNanos` is for. Why is `sleepPeriodNanos` multiplied by 2 here? Is it because `LOGICAL_INTERVAL`'s default value is 2?

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/fd/GMSHealthMonitor.java
```diff
@@ -1538,4 +1463,114 @@ public void run() {
...
+          sleeper.sleep(sleepPeriodMillis);
+          final long timeAfterSleep = nanoTimer.nanoTime();
+          final long asleepNanos = timeAfterSleep - timeBeforeSleep;
+          if (asleepNanos > sleepLimitNanos) {
+            warner.warn(
+                String.format(
+                    "Failure detection heartbeat-generation thread overslept by more than a full period. Asleep time: %,d nanoseconds. Period: %,d nanoseconds.",
+                    asleepNanos, sleepPeriodNanos));
+          }
+        } catch (InterruptedException e) {
```
kamilla1201, line 1511: looks like we should set the interrupt flag here

kamilla1201 (commented):
I like this idea!

geode-membership/src/integrationTest/java/org/apache/geode/distributed/internal/membership/gms/fd/GMSHealthMonitorJUnitTest.java
```diff
@@ -999,6 +1000,52 @@ public void testDoTCPCheckMemberWithUnkownStatus() throws Exception {
     executeTestDoTCPCheck(GMSHealthMonitor.ERROR + 100, false);
   }
 
+  @Test
+  public void heartbeatOversleepCausesWarning() {
+    testHeartbeatSleepScenario(period -> 2 * period + 1,
```
bschuchardt, line 1005: maybe replace the "2"s in this test with the logical-interval constant from GMSHealthMonitor

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/fd/GMSHealthMonitor.java
```diff
@@ -1538,4 +1463,114 @@ public void run() {
...
+    public final long sleepLimitNanos = 2 * sleepPeriodNanos;
+
+    @Override
+    public void run() {
+      Thread.currentThread().setName("Geode Heartbeat Sender");
+      sendPeriodicHeartbeats(Thread::sleep, System::nanoTime, logger::warn);
+    }
+
+    @VisibleForTesting
+    void sendPeriodicHeartbeats(final Sleeper sleeper,
```
bschuchardt, line 1496: I like this nice injection of time-related functions

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/fd/GMSHealthMonitor.java
Bill: I have somewhat arbitrarily chosen two periods as the sleep limit. I could have e.g. chosen a period and a half as a limit. Or I could have chosen three periods as a limit. Or I could've chosen something else entirely.

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/fd/GMSHealthMonitor.java
Bill: nice catch! coming in next commit…

geode-membership/src/integrationTest/java/org/apache/geode/distributed/internal/membership/gms/fd/GMSHealthMonitorJUnitTest.java
Bill: actually that's a whole different `2`. This `2` is the (arbitrary) oversleep criterion. Will fix in next commit…

bschuchardt (approved):
determined code!

kamilla1201 (approved):

echobravopapa (approved):

PR #6362, opened 2021-04-21
GEODE-9138: Modified hasSeenEvent to log the message if necessary and…

boglesby:
… increment a stat

### For all changes:

### Note:


geode-core/src/integrationTest/java/org/apache/geode/internal/cache/event/DistributedEventTrackerIntegrationTest.java
```diff
@@ -0,0 +1,146 @@
...
+import org.apache.geode.test.junit.rules.serializable.SerializableTestName;
+import org.apache.geode.test.junit.runners.CategoryWithParameterizedRunnerFactory;
+
+@RunWith(Parameterized.class)
+@Parameterized.UseParametersRunnerFactory(CategoryWithParameterizedRunnerFactory.class)
+@Category({RegionsTest.class})
+public class DistributedEventTrackerIntegrationTest {
+
+  @Parameterized.Parameters(name = "possibleDuplicate={0}")
+  public static Collection booleans() {
```
DonalEvans, line 60: Compiler warning here can be fixed by using `Collection<Boolean>`

geode-core/src/integrationTest/java/org/apache/geode/internal/cache/event/DistributedEventTrackerIntegrationTest.java
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+  private void invokeHasSeenEventAndVerifyResults(DistributedRegion region)
+      throws IOException {
+    // Record an event with a high sequence number
+    recordHighSequenceNumberEvent(region);
+
+    // Create an event with a lower sequence number and possibleDuplicate set appropriately
+    EntryEventImpl event =
+        createEvent(region, new EventID(new byte[0], 1l, 0l));
```
DonalEvans, line 111: Since the `threadId` used in the constructor for the `EventID` here needs to be the same as that used in the constructor for the `EventID` in `recordHighSequenceNumberEvent()`, it might be best to extract the value to a constant. Also, very minor nitpick, but IntelliJ creates a warning for me that it's better to use upper-case L to denote long values, so it seems like a simple thing to fix here too.

geode-core/src/integrationTest/java/org/apache/geode/internal/cache/event/DistributedEventTrackerIntegrationTest.java
```diff
@@ -0,0 +1,146 @@
...
+    event.setPossibleDuplicate(possibleDuplicate);
+
+    // Invoke hasSeenEvent
+    boolean hasSeenEvent = region.getEventTracker().hasSeenEvent(event);
+
+    // Assert hasSeenEvent is true
+    assertTrue(hasSeenEvent);
```
DonalEvans, line 118: Could this be replaced with `assertThat(hasSeenEvent).isTrue();` to keep all assertions using AssertJ?

DonalEvans (changes requested):
Just a few small test clean-up requests.

geode-core/src/integrationTest/java/org/apache/geode/internal/cache/event/DistributedEventTrackerIntegrationTest.java
boglesby: Done. Thanks.

geode-core/src/integrationTest/java/org/apache/geode/internal/cache/event/DistributedEventTrackerIntegrationTest.java
boglesby: Done. Thanks.

geode-core/src/integrationTest/java/org/apache/geode/internal/cache/event/DistributedEventTrackerIntegrationTest.java
boglesby: Done. Thanks.

DonalEvans (approved):

kirklund (approved):

echobravopapa (approved):
This is a good add to Geode's stats!

gesterzhou (approved):

nabarunnag:
We can add a bit of documentation  PR : https://github.com/apache/geode/pull/6395

nabarunnag (changes requested):
Addition of documentation, PR : https://github.com/apache/geode/pull/6395

geode-core/src/main/java/org/apache/geode/internal/cache/event/DistributedEventTracker.java
```diff
@@ -356,22 +363,58 @@ public boolean hasSeenEvent(EventID eventID, InternalCacheEvent tagHolder) {
...
+      region.getCachePerfStats().incPreviouslySeenEvents();
+
       return true;
     }
   }
 
+  private boolean shouldLogPreviouslySeenEvent(InternalCacheEvent event,
+      EventSequenceNumberHolder evh) {
+    boolean shouldLogSeenEvent = true;
+    String message = null;
```
Bill, line 388: @boglesby this method is ostensibly deciding whether or not to log the `event`. That being the case, I don't understand how it is possible (or sensible) for `event` to be `null`.

geode-core/src/main/java/org/apache/geode/internal/cache/event/DistributedEventTracker.java
```diff
@@ -356,22 +363,58 @@ public boolean hasSeenEvent(EventID eventID, InternalCacheEvent tagHolder) {
...
+      }
+    }
+    if (!shouldLogSeenEvent && logger.isDebugEnabled()) {
+      if (event == null) {
+        logger.debug("Ignoring previously seen event due to {}", message);
+      } else {
+        logger.debug(
+            "Ignoring previously seen event due to {} for region={}; operation={}; key={}; eventId={}; highestSequenceNumberSeen={}",
+            message, region.getName(), ((EntryEventImpl) event).getKey(), event.getOperation(),
+            event.getEventId().expensiveToString(), evh.getLastSequenceNumber());
```
Bill, line 412: It seems odd that a method whose purpose is to decide whether or not to log something, is itself logging something (as a side-effect). Wouldn't it make more sense to put all the logging in the caller?

The effect is (as I read it) we are logging something either way (whether we `shouldLogSeenEvent` or not).

geode-core/src/integrationTest/java/org/apache/geode/internal/cache/event/DistributedEventTrackerIntegrationTest.java
```diff
@@ -0,0 +1,146 @@
...
+@Category({RegionsTest.class})
+public class DistributedEventTrackerIntegrationTest {
+
+  @Parameterized.Parameters(name = "possibleDuplicate={0}")
+  public static Collection<Boolean> booleans() {
+    return Arrays.asList(true, false);
+  }
+
+  @Parameterized.Parameter
+  public boolean possibleDuplicate;
```
Bill, line 64: It's cool that you're covering the positive and negative case this way.

Bill (commented):
A couple questions @boglesby…

geode-core/src/main/java/org/apache/geode/internal/cache/event/DistributedEventTracker.java
nabarunnag: It's a debug log, hence will not be printed during normal operations, this will be used during debug artifacts in customer issues or hydra.

geode-core/src/main/java/org/apache/geode/internal/cache/event/DistributedEventTracker.java
nabarunnag: This is a case in transactions. This is detailed in the method:
```java
  /**
   * has the Region's event state seen this event? Most checks should use the method that takes an
   * Event, not an ID, but with transactions we do not have an event at the time the check needs to
   * be made. Consequently, this method may cause events to be recorded that would otherwise be
   * ignored.
   *
   * @param eventID the identifier of the event
   * @return true if the Region's event state has seen the event
   */
  @Override
  public boolean hasSeenEvent(EventID eventID) {
    return getEventTracker().hasSeenEvent(eventID);
  }
```

boglesby: I didn't want to take a chance on an NPE since hasSeenEvent can be called with a null event in the case that Naba mentioned. One of the other cases that don't use the event posDup can still apply.

nabarunnag (approved):

PR #6373, opened 2021-04-25
GEODE-9191: PR clear could miss clearing bucket which lost primary

gesterzhou:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/distributedTest/java/org/apache/geode/cache/query/partitioned/PRClearCreateIndexDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/cache/query/partitioned/PRClearCreateIndexDUnitTest.java
@@ -1,0 +1,265 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package org.apache.geode.cache.query.partitioned;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.io.Serializable;
+import java.util.stream.IntStream;
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+
+import org.apache.geode.cache.Region;
+import org.apache.geode.cache.client.ClientCache;
+import org.apache.geode.cache.client.ClientRegionShortcut;
+import org.apache.geode.cache.query.QueryService;
+import org.apache.geode.distributed.internal.ClusterDistributionManager;
+import org.apache.geode.distributed.internal.DistributionMessage;
+import org.apache.geode.distributed.internal.DistributionMessageObserver;
+import org.apache.geode.internal.cache.DistributedClearOperation;
+import org.apache.geode.internal.cache.DistributedClearOperation.ClearRegionMessage;
+import org.apache.geode.internal.cache.PartitionedRegionClearMessage;
+import org.apache.geode.test.dunit.AsyncInvocation;
+import org.apache.geode.test.dunit.rules.ClientVM;
+import org.apache.geode.test.dunit.rules.ClusterStartupRule;
+import org.apache.geode.test.dunit.rules.MemberVM;
+
+public class PRClearCreateIndexDUnitTest implements Serializable {
+  @Rule
+  public ClusterStartupRule cluster = new ClusterStartupRule(4, true);
+
+  private MemberVM primary, secondary;
+  private ClientVM client;
+
+  @Before
+  public void before() throws Exception {
+    int locatorPort = ClusterStartupRule.getDUnitLocatorPort();
+    primary = cluster.startServerVM(0, locatorPort);
+    secondary = cluster.startServerVM(1, locatorPort);
+
+    // create region on server1 first, making sure server1 has the primary bucket
+    primary.invoke(() -> {
+      DistributionMessageObserver.setInstance(new MessageObserver());
+      Region<Object, Object> region =
+          ClusterStartupRule.memberStarter.createPartitionRegion("regionA",
+              f -> f.setTotalNumBuckets(1).setRedundantCopies(1));
+      IntStream.range(0, 100).forEach(i -> region.put(i, "value" + i));
+    });
+
+    // server2 has the secondary bucket
+    secondary.invoke(() -> {
+      DistributionMessageObserver.setInstance(new MessageObserver());
+      ClusterStartupRule.memberStarter.createPartitionRegion("regionA",
+          f -> f.setTotalNumBuckets(1).setRedundantCopies(1));
+    });
+  }
+
+  @After
+  public void after() throws Exception {
+    primary.invoke(() -> {
+      DistributionMessageObserver.setInstance(null);
+    });
+    secondary.invoke(() -> {
+      DistributionMessageObserver.setInstance(null);
+    });
+  }
+
+  // All tests create index on secondary members. These tests are making sure we are requesting
+  // locks for clear on secondary members as well. If we create index on the primary, the clear
+  // and createIndex will run sequentially so there would be no error. But if we create index on
+  // the secondary member and if the secondary member will not
+  // request a lock for clear operation, it will result in an EntryDestroyedException when create
+  // index is happening.
+
+  // Note: OP_LOCK_FOR_CLEAR, OP_CLEAR, OP_UNLOCK_FOR_CLEAR are messages for secondary members
+  // OP_LOCK_FOR_PR_CLEAR, OP_UNLOCK_FOR_PR_CLEAR, OP_PR_CLEAR can be for anybody
+
+  @Test
+  // all local buckets are primary, so only OP_LOCK_FOR_CLEAR and OP_CLEAR messages are sent to the
+  // secondary member
+  // in the end an OP_PR_CLEAR is sent to the secondary for no effect
+  public void clearFromPrimaryMember() throws Exception {
+    AsyncInvocation createIndex = secondary.invokeAsync(PRClearCreateIndexDUnitTest::createIndex);
+    AsyncInvocation clear = primary.invokeAsync(PRClearCreateIndexDUnitTest::clear);
+
+    createIndex.get();
+    clear.get();
+
+    // assert that secondary member received these messages
+    primary.invoke(() -> verifyEvents(false, false, false, false));
+    secondary.invoke(() -> verifyEvents(false, true, true, true));
+  }
+
+  @Test
+  // all local buckets are secondary, so an OP_PR_CLEAR is sent to the primary member, from there
+  // a OP_LOCK_FOR_CLEAR and OP_CLEAR messages are sent back to the secondary
+  public void clearFromSecondaryMember() throws Exception {
+    AsyncInvocation createIndex = secondary.invokeAsync(PRClearCreateIndexDUnitTest::createIndex);
+    AsyncInvocation clear = secondary.invokeAsync(PRClearCreateIndexDUnitTest::clear);
+
+    createIndex.get();
+    clear.get();
+
+    // assert that secondary member received these messages
+    primary.invoke(() -> verifyEvents(false, true, false, false));
+    secondary.invoke(() -> verifyEvents(false, false, true, true));
+  }
+
+  /**
+   * For interested client connecting to secondary member
+   * 1. locks all local primary region
+   * 2. send OP_LOCK_FOR_PR_CLEAR to lock all other members
+   * 3. send OP_PR_CLEAR to primary to clear
+   * 4. primary will send a OP_CLEAR message back to the secondary to clear
+   */
+  @Test
+  public void clearFromInterestedClientConnectingToSecondaryMember() throws Exception {
+    int port = secondary.getPort();
+    client = cluster.startClientVM(2, c -> c.withServerConnection(port).withPoolSubscription(true));
+    AsyncInvocation createIndex = secondary.invokeAsync(PRClearCreateIndexDUnitTest::createIndex);
+
+    AsyncInvocation clear = client.invokeAsync(() -> {
+      Thread.sleep(200);
+      ClientCache clientCache = ClusterStartupRule.getClientCache();
+      Region<Object, Object> regionA =
+          clientCache.createClientRegionFactory(ClientRegionShortcut.PROXY).create("regionA");
+      regionA.registerInterestForAllKeys();
+      regionA.clear();
+    });
+
+    createIndex.get();
+    clear.get();
+    primary.invoke(() -> verifyEvents(true, true, false, false));
+    secondary.invoke(() -> verifyEvents(false, false, true, true));
+  }
+
+  @Test
+  /**
+   * For interested client connecting to primary member, behaves like starting from primary member
+   * except it locks first
+   * 1. locks local primary regions
+   * 2. send OP_LOCK_FOR_PR_CLEAR to lock all other members' primary buckets
+   * 3. send a OP_LOCK_FOR_CLEAR message to lock all secondary buckets
+   * 4. send OP_CLEAR to clear all secondary buckets
+   */
+  public void clearFromInterestedClientConnectingToPrimaryMember() throws Exception {
+    int port = primary.getPort();
+    client = cluster.startClientVM(2, c -> c.withServerConnection(port).withPoolSubscription(true));
+    AsyncInvocation createIndex = secondary.invokeAsync(PRClearCreateIndexDUnitTest::createIndex);
+
+    AsyncInvocation clear = client.invokeAsync(() -> {
+      Thread.sleep(200);
+      ClientCache clientCache = ClusterStartupRule.getClientCache();
+      Region<Object, Object> regionA =
+          clientCache.createClientRegionFactory(ClientRegionShortcut.PROXY).create("regionA");
+      regionA.registerInterestForAllKeys();
+      regionA.clear();
+    });
+
+    createIndex.get();
+    clear.get();
+    primary.invoke(() -> verifyEvents(false, false, false, false));
+    secondary.invoke(() -> verifyEvents(true, true, true, true));
+  }
+
+  private static void clear() throws InterruptedException {
+    // start the clear a bit later that the createIndex operation， to reveal the race condition
+    // comment it out since the test does not need the race condition to happen anymore
+    // Thread.sleep(200);
+    Region region = ClusterStartupRule.getCache().getRegion("/regionA");
+    region.clear();
+  }
+
+  private static void createIndex() {
+    QueryService queryService = ClusterStartupRule.getCache().getQueryService();
+    // run create index multiple times to make sure the clear operation fall inside a
+    // createIndex Operation
+    IntStream.range(0, 10).forEach(i -> {
+      try {
+        queryService.createIndex("index" + i, "name" + i, "/regionA");
+      } catch (Exception e) {
+        throw new RuntimeException(e.getMessage(), e);
+      }
+    });
+  }
+
+  private static void verifyEvents(boolean lockOthers, boolean clearOthers, boolean lockSecondary,
+      boolean clearSecondary) {
+    MessageObserver observer = (MessageObserver) DistributionMessageObserver.getInstance();
+    assertThat(observer.isLock_others())
+        .describedAs("OP_LOCK_FOR_PR_CLEAR received: %s", observer.isLock_others())
+        .isEqualTo(lockOthers);
+    assertThat(observer.isClear_others())
+        .describedAs("OP_PR_CLEAR received: %s", observer.isClear_others()).isEqualTo(clearOthers);
+    assertThat(observer.isLock_secondary())
+        .describedAs("OP_LOCK_FOR_CLEAR received: %s", observer.isLock_secondary())
+        .isEqualTo(lockSecondary);
+    assertThat(observer.isClear_secondary())
+        .describedAs("OP_CLEAR received: %s", observer.isClear_secondary())
+        .isEqualTo(clearSecondary);
+  }
+
+  private static class MessageObserver extends DistributionMessageObserver {
+    private volatile boolean lock_secondary = false;
+    private volatile boolean clear_secondary = false;
+    private volatile boolean clear_others = false;
+    private volatile boolean lock_others = false;
+
+    @Override
+    public void beforeProcessMessage(ClusterDistributionManager dm, DistributionMessage message) {
+      if (message instanceof ClearRegionMessage) {
+        ClearRegionMessage clearMessage = (ClearRegionMessage) message;
+        if (clearMessage
+            .getOperationType() == DistributedClearOperation.OperationType.OP_LOCK_FOR_CLEAR) {
+          lock_secondary = true;
+        }
+        if (clearMessage.getOperationType() == DistributedClearOperation.OperationType.OP_CLEAR) {
+          clear_secondary = true;
+        }
+      }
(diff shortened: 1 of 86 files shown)


geode-core/src/main/java/org/apache/geode/internal/cache/BucketRegion.java
```diff
@@ -567,19 +567,13 @@ public boolean virtualPut(EntryEventImpl event, boolean ifNew, boolean ifOld,
...
-      return;
-    }
-
     // get rvvLock
     Set<InternalDistributedMember> participants =
         getCacheDistributionAdvisor().adviseInvalidateRegion();
     boolean isLockedAlready = this.partitionedRegion.getPartitionedRegionClear()
         .isLockedForListenerAndClientNotification();
 
+    boolean lockedForPrimary = doLockForPrimary(false);
```
nabarunnag, line 576: I was just wondering how will this work if while excuting this command, the bucket is no more primary. How will this method behave in this case?

geode-core/src/main/java/org/apache/geode/internal/cache/InitialImageOperation.java
```diff
@@ -846,12 +846,12 @@ boolean processChunk(List entries, InternalDistributedMember sender)
       }
       List<Entry> entriesToSynchronize = new ArrayList<>();
 
+      if (internalDuringApplyDelta != null && !internalDuringApplyDelta.isRunning
```
dschneider-pivotal, line 849: It looks to me like changing this test hook, by moving it out of the loop, will break the existing TombstoneDUnitTest that uses DuringApplyDelta. Why did you need to change it?

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionClear.java
```diff
@@ -165,8 +165,8 @@ protected void waitForPrimary(PartitionedRegion.RetryTimeKeeper retryTimer) {
                     new RegionEventImpl(localPrimaryBucketRegion, Operation.REGION_CLEAR, null,
                         false, partitionedRegion.getMyId(), regionEvent.getEventId());
                 localPrimaryBucketRegion.cmnClearRegion(bucketRegionEvent, false, true);
+                clearedBuckets.add(localPrimaryBucketRegion.getId());
```
dschneider-pivotal, line 168: How is clearedBuckets used? You are changing the code to only add a bucket to it if the size of the bucket is > 0. I can see how clearedBuckets is used down at line 184 to decide if we will increment stats. This does not seem right. If we find a bunch of empty bucket then we still did a clear and took some time so it seems wrong to not even change the stats.
Also we return clearedBuckets from this method and I can't tell what the callers do with it.
Also this check that the primary bucket is not empty does not seem safe to me. Currently if evict destroy is configured on a PR then the primary and secondary can be out of sync. So the primary could be empty but the secondary could still have entries.

geode-core/src/main/java/org/apache/geode/internal/cache/BucketRegion.java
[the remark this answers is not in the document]
dschneider-pivotal: Perhaps what you really want this method to do on a secondary is to just call clearRegionLocally(regionEvent, cacheWrite, null); All the locking and distribution stuff has been done on the primary

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClearPRDuringGIIRequesterDUnitTest.java
```diff
@@ -0,0 +1,339 @@
...
+          }
+
+          @Override
+          public String description() {
+            return null;
+          }
+        };
+
+        GeodeAwaitility.await().untilAsserted(ev);
+        if (callback == null || !callback.isRunning) {
```
DonalEvans, line 288: It should not be necessary to check these conditions here or throw if they're not met, as the `await()` call above will throw if it times out waiting for the callback to be running.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClearPRDuringGIIRequesterDUnitTest.java
```diff
@@ -0,0 +1,339 @@
...
+          fail("GII tesk hook is not started yet");
+        }
+      }
+    };
+    vm.invoke(waitForCallbackStarted);
+  }
+
+  private static class PauseDuringGIICallback extends InitialImageOperation.GIITestHook {
+    private Object lockObject = new Object();
```
DonalEvans, line 297: A couple of warnings can be cleared up if this object is made `final`

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClearPRDuringGIIRequesterDUnitTest.java
```diff
@@ -0,0 +1,339 @@
...
+
+  public void waitForGIITeskHookStarted(final MemberVM vm,
+      final InitialImageOperation.GIITestHookType callbacktype) {
+    SerializableRunnable waitForCallbackStarted = new SerializableRunnable() {
+      @Override
+      public void run() {
+
+        final InitialImageOperation.GIITestHook callback =
+            getGIITestHookForCheckingPurpose(callbacktype);
+        WaitCriterion ev = new WaitCriterion() {
```
DonalEvans, line 274: `WaitCriterion` is deprecated and should not be used. Instead, `GeodeAwaitility.await()` should be used to assert the conditions, e.g. 
```
GeodeAwaitility.await().untilAsserted(() -> assertThat(getGIITestHookForCheckingPurpose(callbacktype)).satisfies(callback -> {
          assertThat(callback).isNotNull();
          assertThat(callback.isRunning).isTrue();
        }));
```

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClearPRDuringGIIRequesterDUnitTest.java
```diff
@@ -0,0 +1,339 @@
...
+      }
+    }
+
+    @Override
+    public void run() {
+      synchronized (this.lockObject) {
+        try {
+          isRunning = true;
+          this.lockObject.wait();
+        } catch (InterruptedException e) {
```
DonalEvans, line 316: IntelliJ warning here can be fixed if the exception name is changed to "ignore".

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClearPRDuringGIIRequesterDUnitTest.java
```diff
@@ -0,0 +1,339 @@
...
+        }
+      }
+    }
+  } // Mycallback
+
+  private class PauseDuringClearDistributionMessageObserver
+      extends DistributionMessageObserver {
+    private CountDownLatch latch = new CountDownLatch(1);
```
DonalEvans, line 324: This can be `final`

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClearPRDuringGIIRequesterDUnitTest.java
```diff
@@ -0,0 +1,339 @@
...
+
+    @Override
+    public void beforeProcessMessage(ClusterDistributionManager dm, DistributionMessage message) {
+      if (message instanceof PartitionedRegionClearMessage) {
+        PartitionedRegionClearMessage prcm = (PartitionedRegionClearMessage) message;
```
DonalEvans, line 329: The variable `prcm` is never used, so this line can be removed.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClearPRDuringGIIRequesterDUnitTest.java
```diff
@@ -0,0 +1,339 @@
...
+        try {
+          logger.info("before wait for clear message");
+          latch.await();
+          logger.info("after wait for clear message");
+        } catch (InterruptedException ex) {
```
DonalEvans, line 334: IntelliJ warning here can be fixed by renaming this exception "ignore".

DonalEvans (changes requested):
Every comment in `ClearPRDuringGIIRequesterDUnitTest.java` also applies to `ClearPRDuringGIIProviderDUnitTest.java` since they're very similar classes.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClearPRDuringGIIRequesterDUnitTest.java
```diff
@@ -0,0 +1,339 @@
...
+    }
+  }
+
+  @Test
+  public void clearBeforeGIIShouldClearTheRegion() {
+    giiTestHookSyncWithClear(true);
+  }
+
+  @Test
+  public void clearAfterGIIShouldClearTheRegion() {
```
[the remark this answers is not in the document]
gesterzhou: PartitionedRegionPartialClearException is an expected behavior, which will be thrown to the user.

geode-core/src/main/java/org/apache/geode/internal/cache/InitialImageOperation.java
gesterzhou: because I need to make sure processChunk is called. If I did not move out of the loop, if the entryCount is 0, the testHook will not be called. I don't want to introduce another testHook. 

I care if processChunk is call more than if there's any data in chunk.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClearPRDuringGIIRequesterDUnitTest.java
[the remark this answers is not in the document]
DonalEvans: If the PartitionedRegionPartialClearException is expected, then the test should not be asserting that the region is cleared, it should only assert that a PartitionedRegionPartialClearException is thrown, and the test name should be changed to reflect that.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionClear.java
gesterzhou: We definitely should remove the "if (localPrimaryBucketRegion.size() > 0) {"

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionClearMessage.java
```diff
@@ -0,0 +1,289 @@
...
+import org.apache.geode.internal.CopyOnWriteHashSet;
+import org.apache.geode.internal.NanoTimer;
+import org.apache.geode.internal.cache.partitioned.PartitionMessage;
+import org.apache.geode.internal.logging.log4j.LogMarker;
+import org.apache.geode.internal.serialization.DeserializationContext;
+import org.apache.geode.internal.serialization.SerializationContext;
+import org.apache.geode.logging.internal.log4j.api.LogService;
+
+/**
+ * this message is for operations no the partition region level, could be sent by any originating
```
Bill, line 45: typo replace "no" with "on"

rhoughton-pivot (approved):

geode-core/src/main/java/org/apache/geode/internal/cache/partitioned/ClearPRMessage.java
```diff
@@ -0,0 +1,320 @@
...
+import org.apache.geode.internal.cache.EventID;
+import org.apache.geode.internal.cache.ForceReattemptException;
+import org.apache.geode.internal.cache.PartitionedRegion;
+import org.apache.geode.internal.cache.RegionEventImpl;
+import org.apache.geode.internal.logging.log4j.LogMarker;
+import org.apache.geode.internal.serialization.DeserializationContext;
+import org.apache.geode.internal.serialization.SerializationContext;
+import org.apache.geode.logging.internal.log4j.api.LogService;
+
+public class ClearPRMessage extends PartitionMessageWithDirectReply {
```
Bill, line 54: Need a class-level documentation comment explaining what this class is for.

This is especially needed because this PR introduces another, very similarly-named message class `PartitionedRegionClearMessage`. I have no idea why both are needed and how they relate.

Bill (changes requested):
In addition to the changes requested inline, I'd like to ask if you could make the naming a bit more consistent. I see these new names:

```
CLEAR_PARTITIONED_REGION_MESSAGE
PartitionedRegionClearMessage

CLEAR_PARTITIONED_REGION_REPLY_MESSAGE
PartitionedRegionClearMessage.PartitionedRegionClearReplyMessage

PR_CLEAR_MESSAGE
ClearPRMessage

PR_CLEAR_REPLY_MESSAGE
ClearPRMessage.ClearReplyMessage
```

In some cases "partitioned region" is abbreviated. In other cases it is not.

In some cases "partitioned region" (or its abbreviation) comes first. In other cases it does not.

Making the constant names match their class names better, and making all the names follow a consistent pattern will help maintainers know what to expect.

jinmeiliao:
Is it possible to break this PR down to multiple PRs. The changes seem overwhelming. Let's separate functionality change from refactoring (renaming etc.)

DonalEvans:
> Is it possible to break this PR down to multiple PRs. The changes seem overwhelming. Let's separate functionality change from refactoring (renaming etc.)

The reason the PR is so huge is that the GEODE-7665 branch was rebased on develop after this PR was created. The PR needs to be rebased on the GEODE-7665 branch before it can be reviewed. @gesterzhou

geode-core/src/distributedTest/java/org/apache/geode/cache/query/partitioned/PRClearCreateIndexDUnitTest.java
```diff
@@ -0,0 +1,265 @@
...
+
+  // Note: OP_LOCK_FOR_CLEAR, OP_CLEAR, OP_UNLOCK_FOR_CLEAR are messages for secondary members
+  // OP_LOCK_FOR_PR_CLEAR, OP_UNLOCK_FOR_PR_CLEAR, OP_PR_CLEAR can be for anybody
+
+  @Test
+  // all local buckets are primary, so only OP_LOCK_FOR_CLEAR and OP_CLEAR messages are sent to the
+  // secondary member
+  // in the end an OP_PR_CLEAR is sent to the secondary for no effect
+  public void clearFromPrimaryMember() throws Exception {
+    AsyncInvocation createIndex = secondary.invokeAsync(PRClearCreateIndexDUnitTest::createIndex);
```
kirklund, line 98: When there's no return from an invokeAsync, the type should be `AsyncInvocation<Void>`. In general, we should always declare types on everything.

geode-core/src/distributedTest/java/org/apache/geode/cache/query/partitioned/PRClearCreateIndexDUnitTest.java
```diff
@@ -0,0 +1,265 @@
...
+    AsyncInvocation clear = primary.invokeAsync(PRClearCreateIndexDUnitTest::clear);
+
+    createIndex.get();
```
kirklund, line 101: These calls to `AsyncInvocation.get()` should actually be `AsyncInvocation.await()` since there's no return

geode-core/src/distributedTest/java/org/apache/geode/cache/query/partitioned/PRClearCreateIndexDUnitTest.java
```diff
@@ -0,0 +1,265 @@
...
+import org.apache.geode.test.dunit.AsyncInvocation;
+import org.apache.geode.test.dunit.rules.ClientVM;
+import org.apache.geode.test.dunit.rules.ClusterStartupRule;
+import org.apache.geode.test.dunit.rules.MemberVM;
+
+public class PRClearCreateIndexDUnitTest implements Serializable {
+  @Rule
+  public ClusterStartupRule cluster = new ClusterStartupRule(4, true);
+
+  private MemberVM primary, secondary;
```
kirklund, line 47: The spotless format really shouldn't be allowing this. Please declare each field on its own line.

geode-core/src/distributedTest/java/org/apache/geode/cache/query/partitioned/PRClearQueryIndexDUnitTest.java
```diff
@@ -0,0 +1,376 @@
...
+  }
+
+  @Test
+  public void concurrentClearAndPut() throws Exception {
+    AsyncInvocation puts = server1.invokeAsync(() -> {
+      Cache cache = ClusterStartupRule.getCache();
+      Region region = cache.getRegion("cities");
+      for (int i = 0; i < 1000; i++) {
+        // wait for gate to open
+        getBlackboard().waitForGate("proceedToPut", 60, TimeUnit.SECONDS);
```
kirklund, line 298: We shouldn't be using timeouts less than the default that is in GeodeAwaitility. You can simply use:
```
getBlackboard().waitForGate("proceedToPut", GeodeAwaitility.getTimeout().getSeconds(), TimeUnit.SECONDS);
```

geode-core/src/distributedTest/java/org/apache/geode/cache/query/partitioned/PRClearQueryIndexDUnitTest.java
```diff
@@ -0,0 +1,376 @@
...
+    startStopServer.get(60, TimeUnit.SECONDS);
+    putAndClear.get(60, TimeUnit.SECONDS);
+  }
+
+  private static DUnitBlackboard getBlackboard() {
+    if (blackboard == null) {
+      blackboard = new DUnitBlackboard();
+    }
+    return blackboard;
+  }
```
kirklund, line 364: Please use `DistributedBlackboard` instead:
```
  @Rule
```

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClearPRDuringGIIProviderDUnitTest.java
```diff
@@ -0,0 +1,325 @@
...
+
+  public void waitForGIITeskHookStarted(final MemberVM vm,
+      final InitialImageOperation.GIITestHookType callbacktype) {
+    SerializableRunnable waitForCallbackStarted = new SerializableRunnable() {
+      @Override
+      public void run() {
+
+        final InitialImageOperation.GIITestHook callback =
+            getGIITestHookForCheckingPurpose(callbacktype);
+        WaitCriterion ev = new WaitCriterion() {
```
kirklund, line 260: Please don't use `WaitCriterion` anymore. You can just directly specify the content of `done()` within the call to `await()`:
```
GeodeAwaitility.await().until(() -> callback != null && callback.isRunning);
```
Note that you want to use `await().until` instead of `await().untilAsserted` since `callback != null && callback.isRunning` never throws anything.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClearPRDuringGIIProviderDUnitTest.java
```diff
@@ -0,0 +1,325 @@
...
+          @Override
+          public String description() {
+            return null;
+          }
+        };
+
+        GeodeAwaitility.await().untilAsserted(ev);
+        if (callback == null || !callback.isRunning) {
+          fail("GII tesk hook is not started yet");
+        }
```
kirklund, line 276: Please use an assertions instead:
```
assertThat(callback).isNotNull();
assertThat(callback.isRunning).isTrue();
```
In general, I would try to avoid having something like `callback` be null so that you don't need the `isNotNull`.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClearPRDuringGIIProviderDUnitTest.java
```diff
@@ -0,0 +1,325 @@
...
+    }
+
+    @Override
+    public void run() {
+      synchronized (this.lockObject) {
+        try {
+          isRunning = true;
+          this.lockObject.wait();
+        } catch (InterruptedException e) {
+        }
```
kirklund, line 303: Please use ErrorCollector or DistributedErrorCollector instead of eating exceptions:
```
  @Rule
  public DistributedErrorCollector errorCollector = new DistributedErrorCollector();
```
```
  errorCollector.addError(e);
}
```
You'll need to change PauseDuringGIICallback to non-static to use the Rule or change PauseDuringGIICallback to have a field of type DistributedErrorCollector which is passed in.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClearPRDuringGIIProviderDUnitTest.java
```diff
@@ -0,0 +1,325 @@
...
+        if (callback == null || !callback.isRunning) {
+          fail("GII tesk hook is not started yet");
+        }
+      }
+    };
+    vm.invoke(waitForCallbackStarted);
+  }
+
+  private static class PauseDuringGIICallback extends InitialImageOperation.GIITestHook {
+    private Object lockObject = new Object();
```
kirklund, line 283: Any synchronization object should be final, so just add the final keyword.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/ClearPRDuringGIIProviderDUnitTest.java
```diff
@@ -0,0 +1,325 @@
...
+          this.lockObject.wait();
+        } catch (InterruptedException e) {
+        }
+      }
+    }
+  } // Mycallback
+
+  private class PauseDuringClearDistributionMessageObserver
+      extends DistributionMessageObserver {
+    private CountDownLatch latch = new CountDownLatch(1);
```
kirklund, line 310: You'll want to have a tearDown in the test class that counts this down. This will prevent the test from leaving an orphaned thread.

In general, it's cleaner to have the test own the CountDownLatch and pass a reference into this inner class via a constructor.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionAfterClearNotificationDUnitTest.java
```diff
@@ -0,0 +1,372 @@
...
+  protected MemberVM locator;
+  protected MemberVM dataStore1;
+  protected MemberVM dataStore2;
+  protected MemberVM dataStore3;
+  protected MemberVM accessor;
+
+  protected ClientVM client1;
+  protected ClientVM client2;
+
+  private static volatile DUnitBlackboard blackboard;
```
kirklund, line 66: Please use the DistributedBlackboard rule instead:
```
  @Rule
```

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionAfterClearNotificationDUnitTest.java
```diff
@@ -0,0 +1,372 @@
...
+    accessor.invoke(() -> feed(false));
+    verifyServerRegionSize(NUM_ENTRIES);
+
+    dataStore1.invoke(() -> getRegion(false).clear());
+
+    verifyServerRegionSize(0);
+    verifyCacheListenerTriggerCount(dataStore1);
+  }
+
+  @Test(expected = AssertionError.class)
```
kirklund, line 250: Please use AssertJ's `catchThrowable` instead. If any line throws AssertionError, this test will still fail.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionAfterClearNotificationDUnitTest.java
```diff
@@ -0,0 +1,372 @@
...
+    client2.invoke(this::initClientCache);
+
+    accessor.invoke(() -> feed(false));
+    verifyServerRegionSize(NUM_ENTRIES);
+    dataStore2.invoke(() -> DistributionMessageObserver.setInstance(
+        testHookToKillMemberCallingClearBeforeMessageProcessed()));
+
+    AsyncInvocation ds1ClearAsync = dataStore1.invokeAsync(() -> getRegion(false).clear());
+
+    getBlackboard().waitForGate("CLOSE_CACHE", 30, SECONDS);
```
kirklund, line 263: All of these timeouts should change to `GeodeAwaitility.getTimeout().getSeconds()` to avoid intermittent failures on slow machines.

kirklund (changes requested):

PR #6380, opened 2021-04-27
GEODE-9198: Key/trust store watcher follows symlinks

aaronlindsey:
GEODE-9017 introduced a file watching key/trust manager to automatically
reload the key and trust store upon change. However, in the situation
where the key/trust store files are represented by a symbolic links, the
manager would not reload the store correctly. This commit changes the
watcher to follow symbolic links.


upthewaterspout (approved):

Bill (changes requested):
This implementation looks great! (Removing code to add functionality.) But the fact that we have two very similar tests kind of worries me. Would it be possible to have a single test here instead of two?

aaronlindsey:
> Would it be possible to have a single test here instead of two?

Sure, I can try adding an integration test class for the PollingFileWatcher

Bill (approved):
Looks great @aaronlindsey! Thanks for combining the tests.

aaronlindsey:
It looks like the failing D-unit test is a known issue on develop: https://issues.apache.org/jira/browse/GEODE-8990

PR #6384, opened 2021-04-28
force all upgradeTest and integrationTest tasks to depend on installDist

rhoughton-pivot:
Might fix issues we've seen on support/1.14 with classes not in classpath during execution. We'll run a few test iterations and see how this goes...


rhoughton-pivot:
I ran the jobs Naba pointed out (integrationTest and upgradeTest) many times against this PR, and all passed. Not sure if I ran enough times to have confidence that this is, in fact, a fix for the sporadic `class not found` error.

PR #6385, opened 2021-04-28
GEODE-9156: Replace docker-compose-rule with testcontainers in geode-assembly

jdeppe-pivotal:
Something to note when doing SSL testing: testcontainers does not let
you set the container name (using `container_name` in your compose.yml).
This ultimately means that reverse IP lookups produce hostnames that
look something like `project_service_index`. The problem is that these
names are not RFC compliant as they contain underscores. This can break
some aspects of SSL validation. I've had to work around this by renaming
containers in various test classes.

### For all changes:

### Note:


onichols-pivotal (approved):

rhoughton-pivot (approved):

dickcav (approved):

geode-assembly/src/acceptanceTest/java/org/apache/geode/rules/DockerComposeRule.java
```diff
@@ -0,0 +1,202 @@
...
+      dockerClient.execStartCmd(execCreateCmdResponse.getId()).exec(callback).awaitCompletion();
+    } catch (InterruptedException e) {
+      throw new RuntimeException(e);
+    }
+
+    return dockerClient.inspectExecCmd(execCreateCmdResponse.getId()).exec().getExitCodeLong();
+  }
+
+
+  public Integer getExternalPortForService(String serviceName, int port) {
```
bschuchardt, line 143: It would be nice to have javadocs for the methods in 
DockerComposeRule

bschuchardt (approved):
The changes look good to me.  I have one comment.

geode-assembly/src/acceptanceTest/java/org/apache/geode/rules/DockerComposeRule.java
```diff
@@ -0,0 +1,202 @@
...
+  }
+
+  /**
+   * When used with compose, testcontainers does not allow one to have a 'container_name'
+   * attribute in the compose file. This means that container names end up looking something like:
+   * {@code project_service_index}. When a container performs a reverse IP lookup it will get a
+   * hostname that looks something like {@code projectjkh_db_1.my-netwotk}. This can be a problem
+   * since this hostname is not RFC compliant as it contains underscores. This may cause problems
+   * in particular with SSL.
+   */
```
albertogpz, line 88: It is not clear to me with this description what you are doing in the method. Looking at the code, it seems you are only removing the "_1" from the container name but according to the comment, there might be other underscores in the name.
Could you please clarify it?

geode-assembly/src/acceptanceTest/java/org/apache/geode/rules/DockerComposeRule.java
jdeppe-pivotal: Done - thanks for the prompt!

geode-assembly/src/acceptanceTest/java/org/apache/geode/cache/wan/SeveralGatewayReceiversWithSamePortAndHostnameForSendersTest.java
```diff
@@ -92,9 +85,11 @@
   // Docker compose does not work on windows in CI. Ignore this test on windows
```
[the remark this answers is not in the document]
jdeppe-pivotal: Done.

geode-assembly/src/acceptanceTest/java/org/apache/geode/rules/DockerComposeRule.java
jdeppe-pivotal: I've renamed the method to `setContainerName` to make it a bit more explicit. I had to use this in `DualServerSNIAcceptanceTest`. Without this it was producing SSL exceptions indicating that a hostname was malformed.

When `docker-compose` creates containers the names end up looking something like this: `composeowub3t_locator-maeve_1`. The original 'compose' library we're replacing allowed for `container_name:` attributes in the compose yaml, but _testcontainers_ does not. So if a container does a reverse DNS lookup it gets back a hostname that is actually not RFC compliant because it contains underscores:
```
root@server-dolores:/# dig -x 172.22.0.5

; <<>> DiG 9.10.3-P4-Debian <<>> -x 172.22.0.5
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 17690
;; flags: qr rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 0

;; QUESTION SECTION:
;5.0.22.172.in-addr.arpa.	IN	PTR

;; ANSWER SECTION:
5.0.22.172.in-addr.arpa. 600	IN	PTR	composeowub3t_locator-maeve_1.geode-sni-test.
```
Unfortunately, I don't have the stack trace, but this hostname results in an exception being thrown and SSL cannot establish a connection.

Ultimately, this method is here to rename the container exactly as would happen if _testcontainers_ could handle `container_name:`.

albertogpz: Crystal clear now. Thanks.

geode-assembly/src/acceptanceTest/java/org/apache/geode/rules/DockerComposeRule.java
```diff
@@ -0,0 +1,268 @@
...
+    DockerClient dockerClient = DockerClientFactory.instance().client();
+    dockerClient.renameContainerCmd(containerId).withName(newName).exec();
+  }
+
+  /**
+   * Execute a command in a service container
+   *
+   * @return the stdout of the container if the command was successful, else the stderr
+   */
+  public String execForService(String serviceName, String... args) {
```
Bill, line 141: Suggest renaming `args` to `command` for descriptiveness and to conform to next method.

echobravopapa (approved):
Great work Jens! thanks for taking care of this.

geode-assembly/src/acceptanceTest/resources/org/apache/geode/cache/wan/docker-compose.yml
```diff
@@ -17,12 +17,8 @@
 version: '3'
 services:
   locator:
-    container_name: 'locator'
     image: 'geode:develop'
     hostname: locator
-    expose:
```
Bill, line 23: I see you've removed all the `expose` directives in the various `docker-compose.yml` files. I suppose these still work. I guess that's because your Java code is calling testcontainers API to expose the pertinent ports (to other containers).

On the one hand I can see why perhaps this information could be deemed redundant (since it's also in the Java code). On the other hand, if I could have the exposed ports information in only one place, I'd prefer it in the `yml` files since these can be used from the terminal to spin up services for experimentation.

Not only is it useful to have them in the `yml` file for experimentation, their presence there gives us a clear picture (at a glance) of all the exposed ports. That's useful for maintainers.

Would it make sense to reinstate these `expose` directives? Could we rely on them here in the `yml` files and remove the redundant Java code?

geode-assembly/src/acceptanceTest/resources/org/apache/geode/client/sni/dual-server-docker-compose.yml
```diff
@@ -43,11 +37,8 @@ services:
       - ./geode-config:/geode/config:ro
       - ./scripts:/geode/scripts
   server-dolores:
-    container_name: 'server-dolores'
```
Bill, line 46: I guess these container names really were redundant since you can refer to the containers by their service name now. ok.

Bill (changes requested):
A few questions and ideas below…

geode-assembly/src/acceptanceTest/java/org/apache/geode/rules/DockerComposeRule.java
```diff
@@ -0,0 +1,268 @@
...
+      result = container.execInContainer(args);
+    } catch (Exception e) {
+      throw new RuntimeException(e);
+    }
+
+    return result.getExitCode() == 0 ? result.getStdout() : result.getStderr();
+  }
+
+  /**
+   * Execute a commond in a service container, logging the output
```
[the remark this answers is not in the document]
jdeppe-pivotal: Fixed.

geode-assembly/src/acceptanceTest/java/org/apache/geode/rules/DockerComposeRule.java
jdeppe-pivotal: Done.

geode-assembly/src/acceptanceTest/resources/org/apache/geode/cache/wan/docker-compose.yml
jdeppe-pivotal: Totally agree, but I couldn't get _testcontainers_ to work properly with the `expose` directives left in place. Possibly I'm doing something wrong. It does do something a bit weird in that it spins up an _ambassador_ container which proxies the connections to the actual containers. I think that may be part of the need to not expose the ports, but I'm not sure. More details here: https://www.testcontainers.org/modules/docker_compose/#example

geode-assembly/src/acceptanceTest/resources/org/apache/geode/client/sni/dual-server-docker-compose.yml
jdeppe-pivotal: And _testcontainers_ errors out if they are left in place...

geode-assembly/src/acceptanceTest/resources/org/apache/geode/cache/wan/docker-compose.yml
Bill: Apparently `expose` (in a docker compose file) isn't really needed:

https://stackoverflow.com/questions/65784039/why-a-service-configured-with-only-expose-is-able-to-communicate-with-internet/65785558#65785558

At best is was mere documentation. So I'm ok w/ removing these.

PR #6390, opened 2021-04-28
GEODE-9070: Fix flaky behavior in ClientServerSessionCacheDUnit tests

sabbey37:
Wait until cache server is running before attempting to start the client or validate the server.

### DIFF
--- a/extensions/geode-modules/src/main/java/org/apache/geode/modules/util/BootstrappingFunction.java
+++ b/extensions/geode-modules/src/main/java/org/apache/geode/modules/util/BootstrappingFunction.java
@@ -163,6 +163,8 @@ public class BootstrappingFunction implements Function, MembershipListener, Data
     // Create and execute the function
     Cache cache = CacheFactory.getAnyInstance();
     Execution execution = FunctionService.onMember(member);
+    logger.info("BR About to execute boostrapping function ");
+    new Exception().printStackTrace();
     ResultCollector collector = execution.execute(this);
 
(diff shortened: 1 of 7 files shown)


sabbey37:
Feel free to suggest or add other log statements that might be helpful in debugging if the failure happens again.

extensions/geode-modules/src/main/java/org/apache/geode/modules/util/BootstrappingFunction.java
```diff
@@ -97,6 +100,9 @@ protected Cache verifyCacheExists() {
     }
 
     if (cache == null) {
+      if (logger.isDebugEnabled()) {
+        logger.debug("Creating a new cache in BootstrappingFunction");
```
sabbey37, line 104: Is this the proper log level?

jdeppe-pivotal: I'd suggest to setting it to `info` so that it's always visible.

jdeppe-pivotal (approved):

extensions/geode-modules/src/main/java/org/apache/geode/modules/util/BootstrappingFunction.java
sabbey37: Good idea!

sabbey37:
https://github.com/apache/geode/pull/6567

PR #6391, opened 2021-04-28
GEODE-9208: Allow StressNewTest to pick up tests from all directories

sabbey37:
When detecting which tests to run for StressNewTest, we use paths like the following:
```
'*/src/distributedTest/java'
```

This will pick up a file like this one:
```
 geode-assembly/src/distributedTest/java/org/apache/geode/session/tests/Jetty9CachingClientServerTest.java
```

But won't pick up this file since the `src` dir here is more than one directory deep:
```
extensions/geode-modules/src/distributedTest/java/org/apache/geode/modules/util/ClientServerSessionCacheDUnitTest.java 
```

We can use git's pathspec `:glob` keyword described [here](https://git-scm.com/docs/gitglossary) to pick up any number of directories preceding our desired path:
```
':(glob)**/src/distributedTest/java/**'
```

### DIFF
--- a/ci/scripts/repeat-new-tests.sh
+++ b/ci/scripts/repeat-new-tests.sh
@@ -60,11 +60,11 @@ function create_gradle_test_targets() {
   echo $(${JAVA_HOME}/bin/java -cp $(cat /tmp/classpath.txt) org.apache.geode.test.util.StressNewTestHelper $@)
 }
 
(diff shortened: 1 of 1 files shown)


upthewaterspout (approved):

onichols-pivotal (approved):
cool trick, I had not run across this globbing syntax before.  seems to work in my experiments.

dickcav (approved):

PR #6392, opened 2021-04-28
GEODE-9204: thread hung waiting for response

bschuchardt:
A not serializable exception can cause a ServerConnection thread to get stuck
waiting for a reply from another member.  This is caused by ReplyMessage
not handling a non-serializable "returnValue".

The fix is to catch and log the exception in the sending node and then
handle the problem in the receiving node.

I would have liked to serialize the "returnValue" field into a separate buffer
so as not to transmit a partial serialization of the object but that would require
an extra buffer copy for every ReplyMessage sent, and we send a lot of them.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/distributedTest/java/org/apache/geode/distributed/internal/BadCacheLoaderDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/distributed/internal/BadCacheLoaderDUnitTest.java
@@ -1,0 +1,100 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.apache.geode.distributed.internal;
+
+import java.io.NotSerializableException;
+import java.io.Serializable;
+import java.util.Properties;
+
+import org.assertj.core.api.Assertions;
+import org.jetbrains.annotations.NotNull;
+import org.junit.Rule;
+import org.junit.Test;
+
+import org.apache.geode.InternalGemFireException;
+import org.apache.geode.cache.Cache;
+import org.apache.geode.cache.Region;
+import org.apache.geode.cache.RegionShortcut;
+import org.apache.geode.distributed.ConfigurationProperties;
+import org.apache.geode.test.dunit.SerializableRunnableIF;
+import org.apache.geode.test.dunit.VM;
+import org.apache.geode.test.dunit.rules.CacheRule;
+import org.apache.geode.test.dunit.rules.DistributedRule;
+
+public class BadCacheLoaderDUnitTest implements Serializable {
+  public static final String TEST_REGION = "testRegion";
+  public static final String TEST_KEY = "testKey";
+  public static final String TEST_VALUE = "testValue";
+
+  static class NotSerializableTestException extends RuntimeException {
+    Object unserializableField = new Object();
+  }
+
+  @Rule
+  public DistributedRule distributedRule = new DistributedRule(2);
+
+  CacheRule cacheRule = CacheRule.builder().build();
+
+  /**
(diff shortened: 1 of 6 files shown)


geode-core/src/main/java/org/apache/geode/distributed/internal/ReplyMessage.java
```diff
@@ -295,7 +296,14 @@ public void toData(DataOutput out,
       out.writeInt(processorId);
     }
     if (this.returnValueIsException || this.returnValue != null) {
-      DataSerializer.writeObject(this.returnValue, out);
+      try {
+        DataSerializer.writeObject(this.returnValue, out);
```
kamilla1201, line 300: toData/fromData methods in this class could use SerializationContext/DeserializationContext instead of DataSerializer.

kamilla1201 (approved):
Looks good to me

geode-core/src/main/java/org/apache/geode/distributed/internal/ReplyMessage.java
bschuchardt: Good idea - I've pushed an update to do that

geode-core/src/distributedTest/java/org/apache/geode/distributed/internal/BadCacheLoaderDUnitTest.java
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  @NotNull
+  private SerializableRunnableIF createRegionWithBadCacheLoader(Properties properties) {
+    return () -> {
+      final Cache cache = cacheRule.getOrCreateCache(properties);
+      final Region<String, Object> testRegion =
+          cache.<String, Object>createRegionFactory(RegionShortcut.PARTITION)
+              .setCacheLoader(helper -> {
+                throw new NotSerializableTestException();
```
Bill, line 93: it's so cool that it's possible to inject this lambda here that simply throws the exception

geode-core/src/main/java/org/apache/geode/distributed/internal/ReplyMessage.java
```diff
@@ -309,11 +316,17 @@ public void fromData(DataInput in,
...
-      this.returnValue = DataSerializer.readObject(in);
+    try {
+      if (testFlag(status, EXCEPTION_FLAG)) {
+        this.returnValue = context.getDeserializer().readObject(in);
+        this.returnValueIsException = true;
+      } else if (testFlag(status, OBJECT_FLAG)) {
+        this.returnValue = context.getDeserializer().readObject(in);
+        this.returnValueIsException = (returnValue instanceof ReplyException);
+      }
+    } catch (IOException e) {
```
Bill, line 327: In `toData()` the comment says we will get a `WriteAbortedException`.

Would it be better to catch exactly that? Is there a reason why catching the more general class is better? I honestly don't know the answer, but was hoping I might learn something by asking.

geode-core/src/test/java/org/apache/geode/internal/cache/partitioned/PutPutReplyMessageJUnitTest.java
```diff
@@ -48,7 +45,8 @@ protected void toData(OldValueImporter ovi, HeapDataOutputStream hdos) throws IO
   @Override
   protected void fromData(OldValueImporter ovi, byte[] bytes)
       throws IOException, ClassNotFoundException {
-    ((PutReplyMessage) ovi).fromData(new DataInputStream(new ByteArrayInputStream(bytes)), mock(
-        DeserializationContext.class));
+    final DataInputStream dataInputStream = new DataInputStream(new ByteArrayInputStream(bytes));
+    ((PutReplyMessage) ovi).fromData(dataInputStream,
+        InternalDataSerializer.createDeserializationContext(dataInputStream));
```
Bill, line 50: Ah this change is necessitated by the changes in `ReplyMessage.fromData()` that causes it to use the `context` (parameter) instead of static methods on `DataSerializer`. ok ✓

Bill (changes requested):
Some comments and one possible change requested: should we catch `WriteAbortedException` instead of `IOException` (see below)?

geode-core/src/main/java/org/apache/geode/distributed/internal/ReplyMessage.java
bschuchardt: I could narrow it down, but it was very surprising to me that a "read" operation would throw a WriteAbortedException and I wanted to cover all the bases.

mhansonp (approved):

geode-core/src/main/java/org/apache/geode/distributed/internal/ReplyMessage.java
Bill: I like that answer @bschuchardt !

Bill:
The work on this ticket is moving to @kamilla1201 so I'm closing this PR. She'll open a new one…

PR #6402, opened 2021-04-29
GEODE-9092/GEODE-9093 see if revert helps with distributed8 failure rate

onichols-pivotal:
this is exploratory, no sure yet if these commits actually caused the hangs we've started seeing in the last 16 hrs


onichols-pivotal:
it was timeouts-too-short, not this

onichols-pivotal:
since nothing has passed the pipeline since these commits, I'd like to keep the revert option on the table a bit longer

PR #6405, opened 2021-04-29
GEODE-9217: Adding memory overhead tests for redis

upthewaterspout:
Adding integration tests that measure the memory overhead of both native redis
and radish.  These tests measure the memory used before and after adding
entries. They assert that the measured per entry overhead is within 1 byte of
the expected per entry overhead. They are measuring the all of the overhead -
whether it's from geode region entries or our radish data structures.

These are essentially the same tests from the feature/redis-memory-overhead
branch, but with assertions added that our expected overhead didn't change.

### DIFF
--- a/gradle/test.gradle
+++ b/gradle/test.gradle
@@ -173,6 +173,12 @@ gradle.taskGraph.whenReady({ graph ->
         jvmArgs += ["--add-opens", "java.xml/jdk.xml.internal=ALL-UNNAMED"]
         jvmArgs += ["--add-opens", "java.base/jdk.internal.module=ALL-UNNAMED"]
         jvmArgs += ["--add-opens", "java.base/java.lang.module=ALL-UNNAMED"]
+        jvmArgs += ["--add-opens", "java.base/jdk.internal.ref=ALL-UNNAMED"]
+        jvmArgs += ["--add-opens", "java.base/jdk.internal.reflect=ALL-UNNAMED"]
+        jvmArgs += ["--add-opens", "java.base/jdk.internal.loader=ALL-UNNAMED"]
+        jvmArgs += ["--add-opens", "java.base/jdk.internal.misc=ALL-UNNAMED"]
+        jvmArgs += ["--add-opens", "java.base/jdk.internal.util.jar=ALL-UNNAMED"]
+        jvmArgs += ["--add-opens", "jdk.management/com.sun.management.internal=ALL-UNNAMED"]
       }
       if (project.hasProperty('testJVM') && !testJVM.trim().isEmpty()) {
         executable = "${testJVM}/bin/java"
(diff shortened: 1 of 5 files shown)


onichols-pivotal (approved):

gradle/test.gradle
```diff
@@ -173,6 +173,12 @@ gradle.taskGraph.whenReady({ graph ->
         jvmArgs += ["--add-opens", "java.xml/jdk.xml.internal=ALL-UNNAMED"]
         jvmArgs += ["--add-opens", "java.base/jdk.internal.module=ALL-UNNAMED"]
         jvmArgs += ["--add-opens", "java.base/java.lang.module=ALL-UNNAMED"]
+        jvmArgs += ["--add-opens", "java.base/jdk.internal.ref=ALL-UNNAMED"]
+        jvmArgs += ["--add-opens", "java.base/jdk.internal.reflect=ALL-UNNAMED"]
+        jvmArgs += ["--add-opens", "java.base/jdk.internal.loader=ALL-UNNAMED"]
+        jvmArgs += ["--add-opens", "java.base/jdk.internal.misc=ALL-UNNAMED"]
+        jvmArgs += ["--add-opens", "java.base/jdk.internal.util.jar=ALL-UNNAMED"]
+        jvmArgs += ["--add-opens", "jdk.management/com.sun.management.internal=ALL-UNNAMED"]
```
onichols-pivotal, line 181: if these add-opens are needed for test, might they be needed elsewhere like dunit/internal/ProcessManager or session/tests/ServerContainer?

upthewaterspout: What's weird is each one of those things has a different set of add-opens than test.gradle, even before my changes. Should they be the same? In my case I just needed these for an integration test that is doing reflection to measure the memory usage of the cache.

nonbinaryprogrammer (approved):
looks good. we'll need to make sure that we add a test to this for sorted sets as well.

sabbey37 (approved):
Nice! 

Just wondering if we actually need/what the purpose of the `MemoryOverheadNativeRedisAcceptanceTest` is since we're not actually expecting our memory overhead values to be the same as Redis.

upthewaterspout:
@sabbey37 - The main reason I left MemoryOverheadNativeRedisAcceptanceTest in there is so we can compare our numbers with native redis and see how close we are getting. It's also a framework and place to add new measurements - for example to measure the native redis overhead of sorted sets to compare with radish.

So I like having the code checked in somewhere. I guess I'm ambivalent about running it in CI every time. It takes about 4 seconds.

sabbey37:
@upthewaterspout that makes sense and sounds reasonable!

gradle/test.gradle
onichols-pivotal: ahh, so this is test code that needs the add-opens (not product code under test).  looks good then.

dschneider-pivotal (approved):

PR #6423, opened 2021-05-03
GEODE-9227: Performance improvements in the loading of CommandMarkers…

kohlmu-pivotal:
… and Converters

### For all changes:

### Note:


geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/CommandManager.java
```diff
@@ -176,16 +177,15 @@ private void loadPluginCommands() {
...
+    // packagesToScan.add(GfshCommand.class.getPackage().getName());
+    // packagesToScan.add(VersionCommand.class.getPackage().getName());
 
     // Create one scanner to be used everywhere
     try (ClasspathScanLoadHelper scanner = new ClasspathScanLoadHelper(packagesToScan)) {
       loadUserCommands(scanner, userCommandPackages);
       loadPluginCommands();
-      loadGeodeCommands(scanner);
+      // loadGeodeCommands(scanner);
       loadConverters(scanner);
```
jinmeiliao, line 189: since we are using service loader to load all the geode converters, we don't need to pass in the scanner here anymore. Inside the `loadConverters` we can create a scanner only for `"org.springframework.shell.converters"` package and add those converters.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/CommandManager.java
```diff
@@ -176,16 +177,15 @@ private void loadPluginCommands() {
...
     packagesToScan.add("org.springframework.shell.converters");
-    packagesToScan.add(GfshCommand.class.getPackage().getName());
-    packagesToScan.add(VersionCommand.class.getPackage().getName());
+    // packagesToScan.add(GfshCommand.class.getPackage().getName());
+    // packagesToScan.add(VersionCommand.class.getPackage().getName());
 
     // Create one scanner to be used everywhere
     try (ClasspathScanLoadHelper scanner = new ClasspathScanLoadHelper(packagesToScan)) {
       loadUserCommands(scanner, userCommandPackages);
       loadPluginCommands();
```
jinmeiliao, line 187: this code actually loads all geode commands, right? let's rename it.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/CommandManager.java
```diff
@@ -176,16 +177,15 @@ private void loadPluginCommands() {
...
-    packagesToScan.add("org.apache.geode");
     packagesToScan.add("org.springframework.shell.converters");
-    packagesToScan.add(GfshCommand.class.getPackage().getName());
-    packagesToScan.add(VersionCommand.class.getPackage().getName());
+    // packagesToScan.add(GfshCommand.class.getPackage().getName());
+    // packagesToScan.add(VersionCommand.class.getPackage().getName());
 
     // Create one scanner to be used everywhere
     try (ClasspathScanLoadHelper scanner = new ClasspathScanLoadHelper(packagesToScan)) {
       loadUserCommands(scanner, userCommandPackages);
```
jinmeiliao, line 186: we can just pass in the `userCommandsPackages` for this method, and create the scanner inside the method, we don't need all the previous lines.

jinmeiliao (changes requested):

DonalEvans (approved):

geode-core/src/main/java/org/apache/geode/management/internal/util/ClasspathScanLoadHelper.java
```diff
@@ -41,6 +41,12 @@ public ClasspathScanLoadHelper(Collection<String> packagesToScan) {
         .enableAnnotationInfo().scan(1);
   }
 
+  public ClasspathScanLoadHelper(String... packagesToScan) {
+    scanResult = new ClassGraph().whitelistPackages(packagesToScan)
```
jinmeiliao, line 45: use the non-deprecated method `acceptPackage` instead

geode-core/src/main/java/org/apache/geode/management/internal/util/ClasspathScanLoadHelper.java
```diff
@@ -41,6 +41,12 @@ public ClasspathScanLoadHelper(Collection<String> packagesToScan) {
         .enableAnnotationInfo().scan(1);
```
jinmeiliao, line 41: have this call the other constructor instead `this(packagesToScan.toArray(new String[] {}))`

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/CommandManager.java
```diff
@@ -128,15 +134,17 @@ private static void raiseExceptionIfEmpty(Set<?> foundClasses, String errorFor)
...
+    if (userCommandPackages.length == 0) {
       return;
     }
 
     // Load commands found in all of the packages
-    try {
+    try (ClasspathScanLoadHelper scanner = new ClasspathScanLoadHelper(userCommandPackages)) {
       Set<Class<?>> foundClasses = scanner.scanPackagesForClassesImplementing(CommandMarker.class,
-          restrictedToPackages.toArray(new String[] {}));
+          userCommandPackages);
```
jinmeiliao, line 147: I believe you don't need to pass in the 2nd parameter if scanning from all packages

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/CommandManager.java
```diff
@@ -157,59 +165,38 @@ private void loadUserCommands(ClasspathScanLoadHelper scanner, Set<String> restr
...
+    loadUserDefinedCommands();
+  }
 
+  private void loadSpringDefinedConverters() {
+    try (ClasspathScanLoadHelper scanner = new ClasspathScanLoadHelper(SPRING_CONVERTER_PACKAGE)) {
       // Spring shell's converters
-      foundClasses = scanner.scanPackagesForClassesImplementing(Converter.class,
-          "org.springframework.shell.converters");
+      Set<Class<?>> foundClasses = scanner.scanPackagesForClassesImplementing(Converter.class,
+          SPRING_CONVERTER_PACKAGE);
```
jinmeiliao, line 197: same here, no need for the 2nd parameter

jinmeiliao (changes requested):

jinmeiliao (approved):
Just some minor change requests, otherwise, LGTM.

mhansonp (approved):

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/CommandManager.java
```diff
@@ -112,12 +121,10 @@ private static void raiseExceptionIfEmpty(Set<Class<?>> foundClasses, String err
     }
 
     // Find by packages specified in the distribution config
-    if (this.cacheProperties != null) {
-      String cacheUserCmdPackages =
-          this.cacheProperties.getProperty(ConfigurationProperties.USER_COMMAND_PACKAGES);
-      if (cacheUserCmdPackages != null && !cacheUserCmdPackages.isEmpty()) {
-        userCommandSources.add(cacheUserCmdPackages);
-      }
+    String cacheUserCmdPackages =
```
agingade, line 124: I am not sure how the cacheProperties are initialized, will there be a case "cacheUserCmdPackages" could be null.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/CommandManager.java
```diff
@@ -80,16 +88,17 @@ public CommandManager() {
    * this is used when getting the instance in a cache server. We are getting the
    * user-command-package from distribution properties. used by OnlineCommandProcessor.
    */
-  public CommandManager(final Properties cacheProperties, InternalCache cache) {
-    if (cacheProperties != null) {
-      this.cacheProperties = cacheProperties;
+  public CommandManager(final Properties newCacheProperties, InternalCache cache) {
+    if (newCacheProperties != null) {
+      this.cacheProperties.putAll(newCacheProperties);
```
agingade, line 93: If you are planning to do code cleanup; reference to "this" can be removed in many places :)

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/CommandManager.java
```diff
@@ -127,15 +134,17 @@ private static void raiseExceptionIfEmpty(Set<Class<?>> foundClasses, String err
...
       return;
     }
 
     // Load commands found in all of the packages
-    try {
-      Set<Class<?>> foundClasses = scanner.scanPackagesForClassesImplementing(CommandMarker.class,
-          restrictedToPackages.toArray(new String[] {}));
+    try (ClasspathScanLoadHelper scanner = new ClasspathScanLoadHelper(userCommandPackages)) {
+      Set<Class<?>> foundClasses =
+          scanner.scanPackagesForClassesImplementing(CommandMarker.class, userCommandPackages);
```
agingade, line 147: Not sure if there will be any perf benefits, just thought of bringing it here...It looks like we are instantiating the command classed found (below line); can this initialization be done lazily, when the command is used/executed, will this help.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/CommandManager.java
kohlmu-pivotal: `cacheProperties` is initialized to empty Properties on line 75

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/CommandManager.java
kohlmu-pivotal: the removal of the keyword 'this' does not make code any cleaner... in actual fact, it makes it harder to read, as we are spoilt with IDE's that indicate fields and properties in different colors. As for reading this as text, makes it evident that we are referring to the field defined on the CommandManager.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/CommandManager.java
kohlmu-pivotal: Yes, you are correct that the preemptive initialization of the command, without actually using it, would be slightly faster. But also, given that the number of "userCommandPackages" are unknown here and most likely never going to be populated is a safe bet to do this. In addition I don't think this feature has been documented (but I might be wrong, but I'm not going to go through 1000+ pages of documentation to prove it right/wrong), I doubt this path is ever used. BUT I am keeping it here, just in case a user might actually be using this functionality.

jdeppe-pivotal (approved):

agingade (approved):
Approving based on you look at the check requirement comment i posted.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/CommandManager.java
```diff
@@ -112,12 +121,10 @@ private static void raiseExceptionIfEmpty(Set<Class<?>> foundClasses, String err
...
     // Find by packages specified in the distribution config
-    if (this.cacheProperties != null) {
-      String cacheUserCmdPackages =
-          this.cacheProperties.getProperty(ConfigurationProperties.USER_COMMAND_PACKAGES);
-      if (cacheUserCmdPackages != null && !cacheUserCmdPackages.isEmpty()) {
-        userCommandSources.add(cacheUserCmdPackages);
-      }
+    String cacheUserCmdPackages =
+        this.cacheProperties.getProperty(ConfigurationProperties.USER_COMMAND_PACKAGES, "");
+    if (!cacheUserCmdPackages.isEmpty()) {
```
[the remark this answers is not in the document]
kohlmu-pivotal: There is a default on the property look up. So if there is no property by that name, it returns empty String.

PR #6426, opened 2021-05-03
GEODE-9204: thread hung waiting for response

kamilla1201:
The original PR #6392 was closed because I took over the work on this ticket.

A not serializable exception can cause a ServerConnection thread to get stuck
waiting for a reply from another member. This is caused by ReplyMessage
not handling a non-serializable "returnValue".

The fix is to catch and log the exception in the sending node and then
handle the problem in the receiving node.

I would have liked to serialize the "returnValue" field into a separate buffer
so as not to transmit a partial serialization of the object but that would require
an extra buffer copy for every ReplyMessage sent, and we send a lot of them.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/distributedTest/java/org/apache/geode/distributed/internal/BadCacheLoaderDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/distributed/internal/BadCacheLoaderDUnitTest.java
@@ -45,7 +45,8 @@ public class BadCacheLoaderDUnitTest implements Serializable {
   @Rule
   public DistributedRule distributedRule = new DistributedRule(2);
 
-  CacheRule cacheRule = CacheRule.builder().build();
+  @Rule
+  public CacheRule cacheRule = CacheRule.builder().build();
 
   /**
    * Ensure that a cache loader throwing an exception that is not serializable is handled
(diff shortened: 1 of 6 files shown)


mhansonp (approved):

Bill (approved):
Approved!

echobravopapa (approved):

kamilla1201:
@BenjaminPerryRoss @boglesby @gesterzhou @kirklund @nabarunnag @pivotal-eshu Could you please take a look at this PR?

geode-core/src/main/java/org/apache/geode/distributed/internal/ReplyMessage.java
```diff
@@ -295,7 +295,14 @@ public void toData(DataOutput out,
       out.writeInt(processorId);
     }
     if (this.returnValueIsException || this.returnValue != null) {
-      DataSerializer.writeObject(this.returnValue, out);
+      try {
+        context.getSerializer().writeObject(this.returnValue, out);
+      } catch (NotSerializableException e) {
```
dschneider-pivotal, line 300: I think it would be better to have toData serialize returnValue to its own data buffer (you might want to use HeapDataOutputStream) and if that serialization fails. then change the reply to have "returnValueIsException" and set the "returnValue" to the exception you caught during that serialization. You need to do this before "writeByte(status)" is called since you may be changing the status to have the exception flag set. That way the caller gets a nice exception explaining how the reply data was not serializable. The downside is that you have to serialize it to a temp buffer but it should have a short lifetime and you if no exception happens then you can copy it to the stream efficiently using HeapDataOutputStream.sendTo(DataOutput). Doing a partial write to the DataOutput and catching the exception and then depending on that to always give us a WriteAbortedException on the other side seems scary to me. I'm not sure you would always get the behavior. Also catching IOException in fromData and treating it like the other side sent us a reply exception seems like it would lead to confusion.

dschneider-pivotal (changes requested):

agingade:
@kamilla1201 
Other options myself and @dschneider-pivotal were considering about:
Is this issue very specific to cache loader? in that case, let's check if the value is serializable as part of load, if not then send error response back to the caller. 
@dschneider-pivotal Is it a safe assumption to make, anything we load through the cache-loader needs to be serializable...If not, then we may have to see if its serializable based on the thread/caller ?

dschneider-pivotal:
It is possible to have a loader on a local region in which case it would not need to be serializable. It is safer to defer the check until it is being serialized.

gesterzhou (approved):

gesterzhou (commented):
Don't worry of the testEventIdOutOfOrderInPartitionRegionSingleHop failure in precheckin. It will be fixed in other geode tickets: GEODE-9242, GEODE-9373.

geode-core/src/main/java/org/apache/geode/distributed/internal/ReplyMessage.java
```diff
@@ -295,14 +307,7 @@ public void toData(DataOutput out,
...
     if (this.returnValueIsException || this.returnValue != null) {
-      try {
-        context.getSerializer().writeObject(this.returnValue, out);
-      } catch (NotSerializableException e) {
-        // When this happens data has already been written to the output stream for the
-        // non-serializable object. The recipient will get a java.io.WriteAbortedException when
-        // attempting to deserialize the value, so we handle that in fromData()
-        logger.warn("Unable to send a reply to " + getRecipientsDescription(), e);
-      }
+      context.getSerializer().writeObject(this.returnValue, out);
```
dschneider-pivotal, line 310: when you serialized to hdos it has all the serialized bytes of "this.returnValue". So you don't want to serialize them again here. Instead just copy those bytes to "out". So change this line to "hdos.sendTo(out);". You still need to do this here if the serialization above failed so in the NotSerializableException catch set a boolean "failedSerialization" to true that you can test here.

dschneider-pivotal (changes requested):

geode-core/src/distributedTest/java/org/apache/geode/distributed/internal/BadCacheLoaderDUnitTest.java
```diff
@@ -0,0 +1,100 @@
...
+  public static final String TEST_VALUE = "testValue";
+
+  static class NotSerializableTestException extends RuntimeException {
+    Object unserializableField = new Object();
+  }
+
+  @Rule
+  public DistributedRule distributedRule = new DistributedRule(2);
+
+  CacheRule cacheRule = CacheRule.builder().build();
```
kirklund, line 48: Need to annotate this with `@Rule` or it won't be torn down between tests.

geode-core/src/distributedTest/java/org/apache/geode/distributed/internal/BadCacheLoaderDUnitTest.java
```diff
@@ -0,0 +1,100 @@
...
+          cache.<String, Object>createRegionFactory(RegionShortcut.PARTITION)
+              .setCacheLoader(helper -> new Object())
+              .create(TEST_REGION);
+      testRegion.getAttributesMutator().setCacheLoader(helper -> "should not be invoked");
+      testRegion.get(TEST_KEY);
+    };
+  }
+
+  @NotNull
+  private SerializableRunnableIF createRegionWithBadCacheLoader(Properties properties) {
```
kirklund, line 87: Not a request for change, just a note... I recommend not importing and using `SerializableRunnableIF`. Best way to use VM.invoke with methods is to use void as the return type:
```
private void createRegionWithBadCacheLoader(Properties properties)
```
And then have the caller invoke it with a lambda:
```
cacheLoaderVM.invoke("create a region with a bad cache loader",
        () -> createRegionWithBadCacheLoader(properties));
```

kirklund (changes requested):

dschneider-pivotal (approved):

kirklund (approved):

kamilla1201:
Need to run performance tests before merging this PR

PR #6427, opened 2021-05-03
GEODE-8191: update flaky test

mivanac:
It seems that in some cases, creation of buckets of colocated region can take some time to finish. Until it is finished, recovery threads are updating statistics, but since ValueMonitor is not notified to monitor PartitionedRegionStats (this will happen when creation of region is finished), this info is not propagated. When finally creation of region is finished, SampleCollector will interpret stats as old info, and will not propagate it. 

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/statistics/StatArchiveWriter.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/statistics/StatArchiveWriter.java
@@ -508,6 +508,9 @@ public class StatArchiveWriter implements StatArchiveFormat, SampleHandler {
       }
       writeTimeStamp(nanosTimeStamp);
       for (ResourceInstance ri : resourceInstances) {
+        if (!ri.getStatValuesNotified()) {
+          ri.setStatValuesNotified(true);
+        }
         writeSample(ri);
       }
       writeResourceInst(ILLEGAL_RESOURCE_INST_ID);
--- a/geode-core/src/main/java/org/apache/geode/internal/statistics/StatArchiveWriter.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/statistics/StatArchiveWriter.java
@@ -508,9 +508,7 @@ public class StatArchiveWriter implements StatArchiveFormat, SampleHandler {
       }
       writeTimeStamp(nanosTimeStamp);
       for (ResourceInstance ri : resourceInstances) {
-        if (!ri.getStatValuesNotified()) {
-          ri.setStatValuesNotified(true);
-        }
+        ri.setStatValuesNotified(true);
         writeSample(ri);
       }
       writeResourceInst(ILLEGAL_RESOURCE_INST_ID);
--- a/geode-core/src/main/java/org/apache/geode/internal/statistics/ValueMonitor.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/statistics/ValueMonitor.java
@@ -98,6 +98,9 @@ public class ValueMonitor extends StatisticsMonitor {
         if (this.statistics.contains(resource.getStatistics())) {
           ResourceType resourceType = resource.getResourceType();
           StatisticDescriptor[] sds = resourceType.getStatisticDescriptors();
+          if (!resource.getStatValuesNotified()) {
+            resource.setStatValuesNotified(true);
+          }
           int[] updatedStats = resource.getUpdatedStats();
           for (int i = 0; i < updatedStats.length; i++) {
             int idx = updatedStats[i];
--- a/geode-core/src/main/java/org/apache/geode/internal/statistics/ValueMonitor.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/statistics/ValueMonitor.java
@@ -98,9 +98,7 @@ public class ValueMonitor extends StatisticsMonitor {
         if (this.statistics.contains(resource.getStatistics())) {
           ResourceType resourceType = resource.getResourceType();
           StatisticDescriptor[] sds = resourceType.getStatisticDescriptors();
-          if (!resource.getStatValuesNotified()) {
-            resource.setStatValuesNotified(true);
-          }
+          resource.setStatValuesNotified(true);
           int[] updatedStats = resource.getUpdatedStats();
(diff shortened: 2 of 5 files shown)


geode-core/src/distributedTest/java/org/apache/geode/management/MemberMXBeanDistributedTest.java
```diff
@@ -90,27 +90,26 @@ public void testBucketCount() {
...
+    server3.invoke(() -> createBuckets(tempRegioName));
+    server4.invoke(() -> createBuckets(tempRegioName));
 
     await().untilAsserted(() -> {
       final int sumOfBuckets = server1.invoke(() -> getBucketsInitialized()) +
           server2.invoke(() -> getBucketsInitialized()) +
           server3.invoke(() -> getBucketsInitialized()) +
           server4.invoke(() -> getBucketsInitialized());
-      assertThat(sumOfBuckets).isEqualTo(4000);
+      assertThat(sumOfBuckets).isEqualTo(2000);
```
jinmeiliao, line 112: why is this a flaky test and why this change would fix it?

onichols-pivotal: I can answer the first half of that...the latest [mass test run](https://concourse.apachegeode-ci.info/teams/main/pipelines/apache-develop-mass-test-run/jobs/create-mass-test-run-report/builds/13#L60659893:162) identified this test as 1% flaky.

mivanac: We have observed, when manually executing this test, in case we don`t wait for bucket initialization to finish, in some cases, test will fail. So solution is to wait for bucket initialization to finish, after each create region command.

geode-core/src/distributedTest/java/org/apache/geode/management/MemberMXBeanDistributedTest.java
```diff
@@ -90,27 +90,26 @@ public void testBucketCount() {
       assertThat(sumOfBuckets).isEqualTo(1000);
     });
 
-    for (int i = 1; i < 4; i++) {
-      final String tempRegioName = regionName + i;
+    final int i = 1;
+    final String tempRegioName = regionName + i;
```
DonalEvans, line 94: Now that there is no loop, these lines could be simplified a bit:
```
final String secondRegionName = regionName + "2";
```

geode-core/src/distributedTest/java/org/apache/geode/management/MemberMXBeanDistributedTest.java
DonalEvans: The change here doesn't seem to be adding any additional wait for bucket initialization to finish, just creating fewer regions. Also, `await()` already waits for 5 minutes for the value to be correct, so this seems like more than just a timing issue.

Looking at the previous failures, is seems that when this test fails, it's because one server is returning 0 from `getBucketsInitialized()` for one of the regions, as the reported value is consistently 3750 instead of the expected 4000, and each server hosts 250 buckets per region. Is there any explanation for why this is the case?

mivanac: If you look at previous test, then you can see that 3 collocated regions are created repeatedly, without wait. So instead adding wait after each region is created, I just decrease number of collocated regions to 1.

DonalEvans: Ah, I see now, thanks for the explanation.

geode-core/src/main/java/org/apache/geode/internal/statistics/StatArchiveWriter.java
```diff
@@ -508,6 +508,9 @@ public void sampled(long nanosTimeStamp, List<ResourceInstance> resourceInstance
       }
       writeTimeStamp(nanosTimeStamp);
       for (ResourceInstance ri : resourceInstances) {
+        if (!ri.getStatValuesNotified()) {
+          ri.setStatValuesNotified(true);
+        }
```
DonalEvans, line 513: I think that the `if` is unnecessary here, since it's fine to call `setStatValuesNotified(true)` if the value is already true.

geode-core/src/main/java/org/apache/geode/internal/statistics/ValueMonitor.java
```diff
@@ -98,6 +98,9 @@ protected void monitorStatistics(final long millisTimeStamp,
         if (this.statistics.contains(resource.getStatistics())) {
           ResourceType resourceType = resource.getResourceType();
           StatisticDescriptor[] sds = resourceType.getStatisticDescriptors();
+          if (!resource.getStatValuesNotified()) {
+            resource.setStatValuesNotified(true);
+          }
```
DonalEvans, line 103: I think that the `if` is unnecessary here, since it's fine to call `setStatValuesNotified(true)` if the value is already true.

geode-core/src/test/java/org/apache/geode/internal/statistics/SampleCollectorTest.java
```diff
@@ -210,7 +210,7 @@ public void testAddHandlerAfterSamples() {
     ResourceInstance resourceInstance = allocatedResourceInstanceInfo.getResourceInstance();
     assertNotNull(resourceInstance);
     assertEquals(0, resourceInstance.getId());
-    assertEquals(0, resourceInstance.getUpdatedStats().length);
+    assertEquals(1, resourceInstance.getUpdatedStats().length);
```
DonalEvans, line 213: Why has this value changed? It's not immediately obvious from the changes why the behaviour here has been affected.

mivanac: In this test case, handler is added after that statistic is updated. According to previous behavior, handler was not notified with the update.

mhansonp (approved):

mkevo (approved):

PR #6430, opened 2021-05-04
GEODE-9194: Feature/region stats update

mhansonp:


geode-core/src/main/java/org/apache/geode/internal/cache/AbstractRegionMap.java
```diff
@@ -388,9 +378,21 @@ public void close(BucketRegion bucketRegion) {
...
     }
-    LocalRegion lr = _getOwner();
+    final LocalRegion lr = _getOwner();
     RegionVersionVector localRvv = lr.getVersionVector();
-    incClearCount(lr);
+
+    boolean isBucketRegion = false;
+
+    if (!(lr instanceof HARegion)) {
+      CachePerfStats stats = lr.getCachePerfStats();
```
dschneider-pivotal, line 387: stats looks unused so it should be removed

geode-core/src/main/java/org/apache/geode/internal/cache/AbstractRegionMap.java
```diff
@@ -388,9 +378,21 @@ public void close(BucketRegion bucketRegion) {
...
       logger.debug("Clearing entries for {} rvv={}", _getOwner(), rvv);
     }
-    LocalRegion lr = _getOwner();
+    final LocalRegion lr = _getOwner();
     RegionVersionVector localRvv = lr.getVersionVector();
-    incClearCount(lr);
+
+    boolean isBucketRegion = false;
+
+    if (!(lr instanceof HARegion)) {
```
dschneider-pivotal, line 386: I don't think this code needs to check for HARegion. isUsedForPartitionedRegionBucket is all it needs to check for.

geode-core/src/main/java/org/apache/geode/internal/cache/AbstractRegionMap.java
```diff
@@ -378,6 +365,9 @@ public void close(BucketRegion bucketRegion) {
   @Override
   public Set<VersionSource> clear(RegionVersionVector rvv, BucketRegion bucketRegion) {
     Set<VersionSource> result = new HashSet<VersionSource>();
+    long nanoTime = System.nanoTime();
```
dschneider-pivotal, line 368: don't directly call System.nanoTime, Instead add a startBucketClear and endBucketClear to PartitionedRegionStats and have it use the standard start code to compute time stats if they are enabled

geode-core/src/main/java/org/apache/geode/internal/cache/AbstractRegionMap.java
```diff
@@ -388,9 +378,21 @@ public void close(BucketRegion bucketRegion) {
...
+    final LocalRegion lr = _getOwner();
     RegionVersionVector localRvv = lr.getVersionVector();
-    incClearCount(lr);
+
+    boolean isBucketRegion = false;
+
+    if (!(lr instanceof HARegion)) {
+      CachePerfStats stats = lr.getCachePerfStats();
+      if (stats != null) {
+        if (lr.isUsedForPartitionedRegionBucket()) {
```
dschneider-pivotal, line 389: this code would be much cleaner if you added noop methods on LocalRegion named startBucketClear and endBucketClear. Then you can override those methods on BucketRegion to call the corresponding start/end on the prstats. That allows you to get rid of the ifs and the isBucketRegion boolean here.

geode-core/src/main/java/org/apache/geode/internal/cache/AbstractRegionMap.java
```diff
@@ -473,6 +475,11 @@ public void close(BucketRegion bucketRegion) {
         }
       }
     }
+    if (isBucketRegion) {
```
dschneider-pivotal, line 478: this code should be in a finally to make sure it happens

geode-core/src/main/java/org/apache/geode/internal/cache/AbstractRegionMap.java
```diff
@@ -388,9 +378,21 @@ public void close(BucketRegion bucketRegion) {
...
     RegionVersionVector localRvv = lr.getVersionVector();
-    incClearCount(lr);
+
+    boolean isBucketRegion = false;
+
+    if (!(lr instanceof HARegion)) {
+      CachePerfStats stats = lr.getCachePerfStats();
+      if (stats != null) {
+        if (lr.isUsedForPartitionedRegionBucket()) {
+          lr.getPartitionedRegion().getPrStats().incBucketClearCount();
```
dschneider-pivotal, line 390: you should move the incBucketClearCount to the end. That way you inc the op count at the same time you inc the time stat

geode-core/src/main/java/org/apache/geode/internal/cache/CachePerfStats.java
```diff
@@ -1413,36 +1404,30 @@ public void endJob() {
...
+    long clearStartTime = getTime();
+    stats.incLong(clearsInProgressId, 1L);
+    return clearStartTime;
   }
 
-  public void incBucketClearCount() {
-    stats.incLong(bucketClearsId, 1L);
+  public long endClear(long startTime) {
+    long timeTaken = getTime() - startTime;
+    stats.incLong(clearTimeId, timeTaken);
```
dschneider-pivotal, line 1419: you should only call incLong(clearTimeId) if (clock.isEnabled())

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionClear.java
```diff
@@ -229,9 +228,9 @@ protected void waitForPrimary(PartitionedRegion.RetryTimeKeeper retryTimer) {
         } finally {
           partitionedRegion.getDataStore().unlockBucketCreationForRegionClear();
           if (clearedBuckets.size() != 0 && partitionedRegion.getCachePerfStats() != null) {
-            partitionedRegion.getRegionCachePerfStats().incRegionClearCount();
-            partitionedRegion.getRegionCachePerfStats()
-                .incPartitionedRegionClearLocalDuration(System.nanoTime() - clearStartTime);
+            partitionedRegion.getRegionCachePerfStats().stats.incLong(CachePerfStats.clearsId, 1L);
```
dschneider-pivotal, line 231: this does not look correct to me. We definitely should not be doing a CachePerfStat this way. And then the clearLocalDuration seems like it should no longer exist. So I think this method should not be doing stats. The top level clear will be measured by startClear and endClear. And the stats for each bucket will take care of showing how much time spent on buckets during a pr clear. So I don't this this method needs any stats.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionStats.java
```diff
@@ -176,8 +176,20 @@
   private static final int prMetaDataSentCountId;
 
   private static final int localMaxMemoryId;
+  static final int clearLocalDurationId;
```
dschneider-pivotal, line 179: I think clearLocalDuration and clearTotalDuration should no longer exist

dschneider-pivotal (changes requested):

geode-core/src/main/java/org/apache/geode/internal/cache/AbstractRegionMap.java
mhansonp: it is used right below this comment.

geode-core/src/main/java/org/apache/geode/internal/cache/AbstractRegionMap.java
mhansonp: ok.

geode-core/src/main/java/org/apache/geode/internal/cache/AbstractRegionMap.java
mhansonp: ok.

geode-core/src/main/java/org/apache/geode/internal/cache/CachePerfStats.java
mhansonp: Done

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionStats.java
mhansonp: agreed.

geode-core/src/main/java/org/apache/geode/internal/cache/AbstractRegionMap.java
mhansonp: I have pruned out the stats need after re-reviewing that code.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionClear.java
mhansonp: Done

geode-core/src/main/java/org/apache/geode/internal/cache/AbstractRegionMap.java
mhansonp: removed the check

geode-core/src/main/java/org/apache/geode/internal/cache/AbstractRegionMap.java
mhansonp: done

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearDUnitTest.java
```diff
@@ -160,23 +169,15 @@ private void verifyDatastoreStats(MemberVM datastore, boolean isCoordinator) {
...
-      assertThat(stats.getPartitionedRegionClearLocalDuration())
-          .isGreaterThan(0);
-      if (isCoordinator) {
-        assertThat(stats.getPartitionedRegionClearTotalDuration())
-            .isGreaterThan(0);
-      } else {
-        assertThat(stats.getPartitionedRegionClearTotalDuration())
-            .isEqualTo(0);
-      }
+      assertThat(stats.getClearCount()).isEqualTo(1);
```
gesterzhou, line 180: It's better to add a replicated region here and do clear on it too. Then check the stats.getClearCount()

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearDUnitTest.java
```diff
@@ -430,13 +423,54 @@ public void normalClearFromClient() {
...
+  public void testClearTime() {
+    configureServers(false, true);
+    client1.invoke(this::initClientCache);
+    client2.invoke(this::initClientCache);
+
+    accessor.invoke(() -> feed(false));
+    verifyServerRegionSize(NUM_ENTRIES);
+    dataStore1.invoke(() -> {
+      PartitionedRegion partitionedRegion = (PartitionedRegion) getRegion(false);
+      assertThat(partitionedRegion.getCachePerfStats().getClearTime()).isEqualTo(0L);
```
gesterzhou, line 436: clearTime maybe not always 0. It's better to set a expected range, such as less than 10ms.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearDUnitTest.java
```diff
@@ -430,13 +423,54 @@ public void normalClearFromClient() {
...
+    });
+    dataStore1.invoke(() -> getRegion(false).clear());
+    verifyServerRegionSize(0);
+    dataStore1.invoke(() -> {
+      PartitionedRegion partitionedRegion = (PartitionedRegion) getRegion(false);
+      assertThat(partitionedRegion.getCachePerfStats().getClearTime()).isGreaterThan(0L);
```
gesterzhou, line 442: It's better to use "<10" instead of ">0". It maybe 0 in some fast machine.

geode-core/src/integrationTest/java/org/apache/geode/internal/cache/PartitionedRegionStatsJUnitTest.java
```diff
@@ -477,4 +485,139 @@ private long getMemBytes(PartitionedRegion pr) {
...
+
+  @Test
+  public void incBucketClearCountIncrementsClears() {
+    String regionName = "testStats";
+    int localMaxMemory = 100;
+    PartitionedRegion pr = createPR(regionName + 1, localMaxMemory, 0);
+
+    pr.getPrStats().incBucketClearCount();
+
+    assertThat(pr.getPrStats().getStats().getLong(bucketClearsId)).isEqualTo(1L);
```
gesterzhou, line 497: need to also verify partitionedRegion.getCachePerfStats().getClearCount()) is still 0.

geode-core/src/integrationTest/java/org/apache/geode/internal/cache/PartitionedRegionStatsJUnitTest.java
```diff
@@ -477,4 +485,139 @@ private long getMemBytes(PartitionedRegion pr) {
...
+    }
+
+    assertThat(pr.size()).isEqualTo(10000);
+    assertThat(pr.getPrStats().getBucketClearCount()).isEqualTo(0L);
+
+    assertThat(pr.getPrStats().getBucketClearTime()).isEqualTo(0L);
+    pr.clear();
+    assertThat(pr.getPrStats().getBucketClearCount()).isGreaterThan(0L);
+
+    assertThat(pr.getPrStats().getBucketClearTime()).isGreaterThan(0L);
```
gesterzhou, line 567: isGreaterThan 0 is not guaranteed.

geode-core/src/main/java/org/apache/geode/internal/cache/AbstractRegionMap.java
```diff
@@ -388,89 +375,105 @@ public void close(BucketRegion bucketRegion) {
...
-                    re.getVersionStamp());
+    boolean isBucketRegion = false;
+    PartitionedRegionStats prStats = null;
+    long startTime = 0L;
+
+    if (lr.isUsedForPartitionedRegionBucket()) {
+      prStats = lr.getPartitionedRegion().getPrStats();
+      startTime = prStats.startBucketClear();
+      isBucketRegion = true;
+    }
```
gesterzhou, line 389: do you need "else" here? how about replicate region?

geode-core/src/main/java/org/apache/geode/internal/cache/AbstractRegionMap.java
```diff
@@ -388,89 +375,105 @@ public void close(BucketRegion bucketRegion) {
...
-          logger.debug("Size after clearing = {}", getEntryMap().size());
-        }
-        if (isTraceEnabled && getEntryMap().size() < 20) {
-          _getOwner().dumpBackingMap();
-        }
+      }
+    } finally {
+      if (isBucketRegion) {
+        prStats.endBucketClear(startTime);
       }
```
gesterzhou, line 477: same. We might need "else" here

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionStats.java
```diff
@@ -1188,11 +1210,51 @@ public void endPutLocal(long start) {
...
 
   public long getPRMetaDataSentCount() {
     return this.stats.getLong(prMetaDataSentCountId);
   }
+
+  public long getBucketClearCount() {
+    return stats.getLong(bucketClearsId);
+  }
+
+  public void incBucketClearCount() {
```
gesterzhou, line 1237: This code is only used in test. why?

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearDUnitTest.java
```diff
@@ -430,13 +423,54 @@ public void normalClearFromClient() {
...
+      PartitionedRegion partitionedRegion = (PartitionedRegion) getRegion(false);
+      assertThat(partitionedRegion.getCachePerfStats().getClearCount()).isEqualTo(0L);
+    });
+    dataStore1.invoke(() -> getRegion(false).clear());
+    verifyServerRegionSize(0);
+    dataStore1.invoke(() -> {
+      PartitionedRegion partitionedRegion = (PartitionedRegion) getRegion(false);
+      assertThat(partitionedRegion.getCachePerfStats().getClearCount()).isEqualTo(1L);
+    });
+  }
```
gesterzhou, line 464: verifyDatastoreStats() should also be called in this test case.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionStats.java
gesterzhou: increase bucket clear count is actually done in endBucketClear(). Your incBucketClearCount() should be removed, or you can change your endBucketClear() to call incBucketClearCount().

gesterzhou (changes requested):
Besides my other comments, your code needs to add following code changes:

diff --git a/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearDUnitTest.java b/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearDUnitTest.java
index 362c801992..02dbe2a575 100644
--- a/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearDUnitTest.java
@@ -177,7 +177,9 @@ public class PartitionedRegionClearDUnitTest implements Serializable {
 
       CachePerfStats stats = region.getRegionCachePerfStats();
 
-      assertThat(stats.getClearCount()).isEqualTo(1);
+      if (isCoordinator) {
+        assertThat(stats.getClearCount()).isEqualTo(1);
+      }
     });
   }
 
@@ -461,6 +463,9 @@ public class PartitionedRegionClearDUnitTest implements Serializable {
       PartitionedRegion partitionedRegion = (PartitionedRegion) getRegion(false);
       assertThat(partitionedRegion.getCachePerfStats().getClearCount()).isEqualTo(1L);
     });
+    verifyDatastoreStats(dataStore1, true);
+    verifyDatastoreStats(dataStore2, false);
+    verifyDatastoreStats(dataStore3, false);
   }
 
 
diff --git a/geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java b/geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
index 1cb4e5d5bb..21df2aec2d 100755
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
@@ -574,7 +574,7 @@ public class PartitionedRegion extends LocalRegion
   public CachePerfStats getRegionCachePerfStats() {
     if (dataStore != null && dataStore.getAllLocalBucketRegions().size() > 0) {
       BucketRegion bucket = dataStore.getAllLocalBucketRegions().iterator().next();
-      return bucket.getCachePerfStats();
+      return bucket.getPartitionedRegion().getCachePerfStats();
     }
     return null;
   }

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearDUnitTest.java
mhansonp: Hmm. Less than 10 will always be true. This test is basically unnecessary if a fast machine can result in 0 nanoseconds. I find it hard to believe though that a machine can do a clear in 0 nanoseconds. Is that really the case?

geode-core/src/main/java/org/apache/geode/internal/cache/BucketRegion.java
```diff
@@ -2538,4 +2531,9 @@ protected void basicClear(RegionEventImpl regionEvent) {
     basicClear(regionEvent, false);
   }
 
+  @Override
+  public void clear() {
```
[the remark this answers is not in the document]
mhansonp: I don't know. That is how the code was. basicClear is supported. I think it is because that PartitionedRegion controls the clearing of BucketRegions.

geode-core/src/main/java/org/apache/geode/internal/cache/AbstractRegionMap.java
mhansonp: Done.

geode-core/src/main/java/org/apache/geode/internal/cache/AbstractRegionMap.java
mhansonp: Nope that code is gone now.

geode-core/src/main/java/org/apache/geode/internal/cache/AbstractRegionMap.java
mhansonp: That code is gone.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionStats.java
mhansonp: It is used now.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearDUnitTest.java
mhansonp: done

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearDUnitTest.java
dschneider-pivotal: It depends on the resolution of the system clock. If you keep calling nanoTime you will see it does not always change even though time has elapsed. So it is possible that for a quick clear start and end will get the same nanoTime and elapsed will be zero. It is also possible that it will be more than 10 due to a gc pause. This type of test will tend to always be flaky. If you can modify geode to use a clock provided by the test (I think some support for this was added for unit tests) then you could have a clock that always changes by 1 each time it is called and then you could assert it is > 0 safely. But it might be hard to even then assert more because the clock might be used by other parts of geode updating other stats.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionClear.java
```diff
@@ -203,6 +201,7 @@ protected void waitForPrimary(PartitionedRegion.RetryTimeKeeper retryTimer) {
         try {
           boolean retry;
           do {
+            int retryTime = 2 * 60 * 1000;
```
dschneider-pivotal, line 204: I liked the old code better but it should have been static and I'd like retryTime to be RETRY_TIME. Also consider using TimeUnit instead of these magic number. TimeUnit.MINUTES.toMillis(2); With a static import of MINUTES you could just pass this in with no var name: "RetryTimeKeeper(MINUTES.toMillis(2))"

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionStats.java
```diff
@@ -176,8 +176,18 @@
   private static final int prMetaDataSentCountId;
 
   private static final int localMaxMemoryId;
+  static final int bucketClearsId;
+  static final int bucketClearTimeId;
+  static final int bucketClearsInProgressId;
 
   static {
+    final String bucketClearsDesc =
+        "The total number of times a clear has been done on this region and it's bucket regions";
```
dschneider-pivotal, line 185: I think this is better: "The total number of times a bucket of this partitioned region has been cleared."

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionStats.java
```diff
@@ -176,8 +176,18 @@
...
+    final String bucketClearTimeDesc =
+        "The total amount of time spent doing a bucket clear onx bucket regions";
```
dschneider-pivotal, line 187: note the "onx" type. Better would be: "The total amount of time, in nanoseconds, spend clearing buckets on this partitioned region."

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionStats.java
```diff
@@ -176,8 +176,18 @@
...
+    final String bucketClearsInProgressDesc =
+        "The total number bucket region clears in progress";
```
dschneider-pivotal, line 189: Counters are "totals". Gauges are "current". So: "The current number bucket clears on this partitioned region that are in progress."
We want to avoid the term "bucket region" in a descriptions. We have a partitioned region that has buckets

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionStats.java
```diff
@@ -425,8 +435,16 @@
                 false),
 
             f.createLongGauge("localMaxMemory",
-                "local max memory in bytes for this region on this member", "bytes")
-
+                "local max memory in bytes for this region on this member", "bytes"),
+            f.createLongCounter("clearLocalDuration",
```
dschneider-pivotal, line 439: I think clearLocalDuration and clearTotalDuration should no longer exist. If I'm right then you want to delete these two lines

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionStats.java
```diff
@@ -1188,11 +1210,51 @@ public void endPutLocal(long start) {
     stats.incLong(putLocalTimeId, delta);
   }
 
+  public long startBucketClear() {
+    incBucketClearsInProgress();
+    return clock.getTime();
+  }
+
+  public void endBucketClear(long start) {
+    long delta = clock.getTime() - start;
```
dschneider-pivotal, line 1219: check "if (clock.isEnabled)" and only calc delta and inc bucketClearTimeId if it is

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionStats.java
```diff
@@ -1188,11 +1210,51 @@ public void endPutLocal(long start) {
...
+
+  public long getBucketClearTime() {
+    return stats.getLong(bucketClearTimeId);
+  }
+
+  public long getBucketClearsInProgress() {
+    return stats.getLong(bucketClearsInProgressId);
+  }
+
+  public void incBucketClearsInProgress() {
```
dschneider-pivotal, line 1253: now that you have startBucketClear and endBucketClear these methods (inc, dec) should be inlines in start and and end or they should be made private. We want to force external classes to call start and end.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionStats.java
```diff
@@ -1188,11 +1210,51 @@ public void endPutLocal(long start) {
...
+
+  public long getBucketClearCount() {
+    return stats.getLong(bucketClearsId);
+  }
+
+  public void incBucketClearCount() {
+    stats.incLong(bucketClearsId, 1L);
+  }
+
+  public void incBucketClearTime(Long nanoseconds) {
```
dschneider-pivotal, line 1241: with the addition of startBucketClear and endBucketClear this method should no longer be needed.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionStats.java
dschneider-pivotal: It looks like it is only used by tests. Shouldn't this method be removed to force callers to use startBucketClear and endBucketClear?

geode-core/src/main/java/org/apache/geode/internal/cache/RegionPerfStats.java
```diff
@@ -519,16 +519,27 @@ public void incEvictWorkTime(long delta) {
...
   }
 
+  private void startClearLocal() {
+    stats.incLong(clearsInProgressId, 1L);
+  }
+
+  private void endClearLocal(long timeTaken) {
+    stats.incLong(clearsInProgressId, -1L);
+    stats.incLong(clearsId, 1L);
+    stats.incLong(clearTimeId, timeTaken);
```
dschneider-pivotal, line 529: Add a "if (clock.isEnabled())" to the inc of clearTimeId

dschneider-pivotal (changes requested):

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionStats.java
mhansonp: It was done that way in working with Anil to enable testing. I don't think it is necessary. I will pull it.

mhansonp: done

geode-core/src/main/java/org/apache/geode/internal/cache/AbstractRegionMap.java
```diff
@@ -388,90 +375,99 @@ public void close(BucketRegion bucketRegion) {
...
-            VersionSource id = re.getVersionStamp().getMemberID();
-            if (id == null) {
-              id = myId;
-            }
-            if (rvv.contains(id, re.getVersionStamp().getRegionVersion())) {
-              if (isTraceEnabled) {
-                logger.trace("region clear op is removing {} {}", re.getKey(),
-                    re.getVersionStamp());
+    long startTime;
+    startTime = _getOwner().startClear();
```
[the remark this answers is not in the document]
mhansonp: Done

geode-core/src/main/java/org/apache/geode/internal/cache/AbstractRegionMap.java
```diff
@@ -388,90 +375,99 @@ public void close(BucketRegion bucketRegion) {
...
+            logger.debug("Size after clearing = {}", getEntryMap().size());
+          }
+          if (isTraceEnabled && getEntryMap().size() < 20) {
+            _getOwner().dumpBackingMap();
+          }
         }
       }
+    } finally {
+      // if (cacheIsOpen) {
+      _getOwner().endClear(startTime);
```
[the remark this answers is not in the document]
mhansonp: done

geode-core/src/main/java/org/apache/geode/internal/cache/BucketRegion.java
```diff
@@ -2538,4 +2531,20 @@ protected void basicClear(RegionEventImpl regionEvent) {
     basicClear(regionEvent, false);
   }
 
+  @Override
+  public long startClear() {
+    long time = getPartitionedRegion().getPrStats().startBucketClear();
```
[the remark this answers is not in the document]
mhansonp: done

geode-core/src/main/java/org/apache/geode/internal/cache/CachePerfStats.java
```diff
@@ -476,12 +472,9 @@
...
                 "operations"),
-            f.createLongCounter("regionClears", regionClearsDesc, "operations"),
-            f.createLongCounter("bucketClears", bucketClearsDesc, "operations"),
-            f.createLongCounter("partitionedRegionClearLocalDuration",
-                partitionedRegionClearLocalDurationDesc, "nanoseconds"),
-            f.createLongCounter("partitionedRegionClearTotalDuration",
-                partitionedRegionClearTotalDurationDesc, "nanoseconds"),
+            f.createLongCounter("clears", clearsDesc, "operations"),
+            f.createLongGauge("clearsInProgress", clearsInProgressDesc, "operations"),
+            f.createLongCounter("clearTime", clearTimeDesc, "operations"),
```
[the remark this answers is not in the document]
mhansonp: done

geode-core/src/main/java/org/apache/geode/internal/cache/CachePerfStats.java
```diff
@@ -1413,36 +1404,36 @@ public void endJob() {
...
 
-  public long getPartitionedRegionClearTotalDuration() {
-    return stats.getLong(partitionedRegionClearTotalDurationId);
-  }
-
-  public void incRegionClearCount() {
-    stats.incLong(regionClearsId, 1L);
+  public long startClear() {
+    long clearStartTime = 0L;
+    if (clock.isEnabled()) {
```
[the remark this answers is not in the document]
mhansonp: done

geode-core/src/main/java/org/apache/geode/internal/cache/LocalRegion.java
```diff
@@ -8437,8 +8437,12 @@ public void basicBridgeClear(Object callbackArg, final ClientProxyMembershipID c
 
     RegionEventImpl event = new ClientRegionEventImpl(this, Operation.REGION_CLEAR, callbackArg,
         false, client.getDistributedMember(), client, eventId);
-
-    basicClear(event, true);
+    long startTime = startClear();
```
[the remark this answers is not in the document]
mhansonp: done

geode-core/src/main/java/org/apache/geode/internal/cache/LocalRegion.java
```diff
@@ -8631,7 +8635,9 @@ public void clearRegionLocally(RegionEventImpl regionEvent, boolean cacheWrite,
   @Override
   void basicLocalClear(RegionEventImpl rEvent) {
     getDataView().checkSupportsRegionClear();
+    long startTime = startClear();
```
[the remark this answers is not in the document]
mhansonp: done

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
```diff
@@ -10140,12 +10143,24 @@ protected PartitionedRegionClear getPartitionedRegionClear() {
...
+
+  @Override
+  public long startClear() {
+    return getCachePerfStats().startClear();
+  }
+
   @Override
   void cmnClearRegion(RegionEventImpl regionEvent, boolean cacheWrite, boolean useRVV) {
     // Synchronized to avoid other threads invoking clear on this vm/node.
+    long startTime = startClear();
```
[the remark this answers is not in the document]
mhansonp: done

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionClear.java
mhansonp: done

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionStats.java
mhansonp: done

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionStats.java
mhansonp: Done

mhansonp: Done

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionStats.java
mhansonp: Good feedback. Thanks!

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionStats.java
mhansonp: Done

mhansonp: you are right.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionStats.java
mhansonp: will do.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionStats.java
mhansonp: removed.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionStats.java
mhansonp: done

geode-core/src/main/java/org/apache/geode/internal/cache/RegionPerfStats.java
mhansonp: done

geode-core/src/integrationTest/java/org/apache/geode/internal/cache/PartitionedRegionStatsJUnitTest.java
mhansonp: Done

geode-core/src/integrationTest/java/org/apache/geode/internal/cache/PartitionedRegionStatsJUnitTest.java
mhansonp: With 10000 entries to clear is seems like the value should be non-zero always. Otherwise, do you have a better suggestion? I am open to suggestions.

geode-core/src/main/java/org/apache/geode/internal/cache/BucketRegion.java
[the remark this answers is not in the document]
mhansonp: I improved the string.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionStats.java
```diff
@@ -1191,11 +1207,42 @@ public void endPutLocal(long start) {
     stats.incLong(putLocalTimeId, delta);
   }
 
+  public long startBucketClear() {
+    stats.incLong(bucketClearsInProgressId, 1L);
+    return clock.getTime();
+  }
+
+  public void endBucketClear(long start) {
+    long delta = clock.getTime() - start;
```
dschneider-pivotal, line 1216: put this delta line inside the "if (clock.isEnabled"

dschneider-pivotal (changes requested):

gesterzhou (approved):

geode-core/src/integrationTest/java/org/apache/geode/internal/cache/PartitionedRegionStatsJUnitTest.java
dschneider-pivotal: If you could have the test insert its own implementation of StatisticsClock then it would be deterministic. Someone did the work of adding the StatisticsClock interface but probably just for unit testing. GemFireCacheImpl calls the static method StatisticsClockFactory.clock(boolean) and ends up passing this clock into the stats when they are created. Kirk might have some ideas of how we could get tests that are testing time stats to have a deterministic clock instead of a real one. These time stats tend to always fail once in a while so it would be great to find a solution.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearDUnitTest.java
```diff
@@ -430,13 +422,58 @@ public void normalClearFromClient() {
...
+    verifyServerRegionSize(NUM_ENTRIES);
+    dataStore1.invoke(() -> {
+      PartitionedRegion partitionedRegion = (PartitionedRegion) getRegion(false);
+      assertThat(partitionedRegion.getCachePerfStats().getClearTime()).isEqualTo(0L);
+    });
+    dataStore1.invoke(() -> getRegion(false).clear());
+    verifyServerRegionSize(0);
+    dataStore1.invoke(() -> {
+      PartitionedRegion partitionedRegion = (PartitionedRegion) getRegion(false);
+      assertThat(partitionedRegion.getCachePerfStats().getClearTime()).isGreaterThan(0L);
```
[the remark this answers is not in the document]
mhansonp: Yes, I will fix it. I thought I fixed this issue. Maybe it was a different test.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionStats.java
mhansonp: nice.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
```diff
@@ -10149,12 +10147,24 @@ protected PartitionedRegionClear getPartitionedRegionClear() {
...
+  }
+
   @Override
   void cmnClearRegion(RegionEventImpl regionEvent, boolean cacheWrite, boolean useRVV) {
     // Synchronized to avoid other threads invoking clear on this vm/node.
+    final long startTime = startClear();
     synchronized (clearLock) {
       partitionedRegionClear.doClear(regionEvent, cacheWrite);
     }
+    endClear(startTime);
```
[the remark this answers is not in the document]
mhansonp: yup

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionClear.java
```diff
@@ -46,9 +47,7 @@
   private static final Logger logger = LogService.getLogger();
 
   protected static final String CLEAR_OPERATION = "_clearOperation";
-
-  private final int retryTime = 2 * 60 * 1000;
-
+  public static final long RETRY_TIME = TimeUnit.MINUTES.toMillis(2);
```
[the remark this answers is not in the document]
mhansonp: yup

geode-core/src/main/java/org/apache/geode/internal/cache/LocalRegion.java
```diff
@@ -8631,7 +8630,9 @@ public void clearRegionLocally(RegionEventImpl regionEvent, boolean cacheWrite,
   @Override
   void basicLocalClear(RegionEventImpl rEvent) {
     getDataView().checkSupportsRegionClear();
+    final long startTime = startClear();
     cmnClearRegion(rEvent, false/* cacheWrite */, false/* useRVV */);
```
dschneider-pivotal, line 8634: the cmnClearRegion call should be in a try and the endClear should be in a finally

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
```diff
@@ -10149,11 +10147,27 @@ protected PartitionedRegionClear getPartitionedRegionClear() {
...
+    return getCachePerfStats().startClear();
+  }
+
   @Override
   void cmnClearRegion(RegionEventImpl regionEvent, boolean cacheWrite, boolean useRVV) {
     // Synchronized to avoid other threads invoking clear on this vm/node.
-    synchronized (clearLock) {
-      partitionedRegionClear.doClear(regionEvent, cacheWrite);
+    long startTime = 0L;
+    try {
```
dschneider-pivotal, line 10164: why is startClear called inside the try? It seems better to have "final long startTime = startClear(); try...".

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegionStats.java
dschneider-pivotal: Why does incBucketClearTime still exist? It should be removed

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearDUnitTest.java
dschneider-pivotal: can a StatisticsClock be added by the test to this test's cache? Otherwise it could fail intermittently. If not I think it would be better for this test to just test the clear count stat and not the clear time stat. The clear time can be tested in an integration test. I also don't like the idea of doing a bunch of work just to get a stat to change.

dschneider-pivotal (changes requested):

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionClearDUnitTest.java
mhansonp: I am testing this in a JUnit test and so I am just going to pull the Dunit test.

geode-core/src/integrationTest/java/org/apache/geode/internal/cache/PartitionedRegionStatsJUnitTest.java
mhansonp: Using a fake clock now.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
mhansonp: Moved.

geode-core/src/main/java/org/apache/geode/internal/cache/LocalRegion.java
mhansonp: done

mhansonp:
I removed the testClearTime time test, because injecting a Statistics Clock into the test cache would not be easy and this stat is already tested via JUnit. Though mocking is used for the StatisticsClock there.

geode-core/src/main/java/org/apache/geode/internal/cache/LocalRegion.java
```diff
@@ -8631,7 +8630,12 @@ public void clearRegionLocally(RegionEventImpl regionEvent, boolean cacheWrite,
   @Override
   void basicLocalClear(RegionEventImpl rEvent) {
     getDataView().checkSupportsRegionClear();
-    cmnClearRegion(rEvent, false/* cacheWrite */, false/* useRVV */);
+    final long startTime = startClear();;
```
dschneider-pivotal, line 8633: change ";;" to ";"

dschneider-pivotal (approved):

PR #6434, opened 2021-05-04
GEODE-9223: Removing ByteArrayWrapper from hashes 

upthewaterspout:
No longer keeping ByteArrayWrapper classes in memory for RedisHashes or
creating ByteArrayWrapper classes when performing hash operations.

The main change to the RedisHash data structure was to switch from a
HashMap<ByteArrayWrapper, ByteArrayWrapper> to a
Object2ObjectOpenCustomHashMap<byte[], byte[]>. This saves ~50 bytes for each
field in the hash to not have the wrappers. Removing the extra allocations when
performing hash operations also helps performance.

Also removing a couple of extra unneeded allocations during logging and message
parsing. We should always create ArrayLists, etc. with a size if it is known so
those objects don't have to resize during parsing/deserialization.

Moving the hash commands executor to a shared object rather than allocating it
for each operation, which is extra GC pressure we don't need.

These changes BREAK compatibility with geode 1.14. The serialization format for
RedisHash, AddsDeltaInfo, and RemsDeltaInfo has changed to not write
ByteArrayWrapper classes on the wire.

### DIFF
--- a/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisHash.java
+++ b/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisHash.java
@@ -69,7 +69,7 @@ public class RedisHash extends AbstractRedisData {
   // added. if our internal implementation changes, these values may be incorrect. the tests will
   // catch this change. an increase in overhead should be carefully considered.
   protected static final int BASE_REDIS_HASH_OVERHEAD = 336;
-  protected static final int HASH_MAP_VALUE_PAIR_OVERHEAD = 47;
+  protected static final int HASH_MAP_VALUE_PAIR_OVERHEAD = 48;
 
   private static final int defaultHscanSnapshotsExpireCheckFrequency =
       Integer.getInteger("redis.hscan-snapshot-cleanup-interval", 30000);
--- a/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisHash.java
+++ b/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisHash.java
@@ -27,8 +27,8 @@ import java.io.DataOutput;
 import java.io.IOException;
 import java.math.BigDecimal;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
-import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
@@ -38,6 +38,8 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.regex.Pattern;
 
+import it.unimi.dsi.fastutil.bytes.ByteArrays;
+import it.unimi.dsi.fastutil.objects.Object2ObjectOpenCustomHashMap;
 import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.apache.commons.lang3.tuple.Pair;
 
@@ -53,9 +55,10 @@ import org.apache.geode.redis.internal.delta.RemsDeltaInfo;
 import org.apache.geode.redis.internal.netty.Coder;
 
 public class RedisHash extends AbstractRedisData {
-  private HashMap<ByteArrayWrapper, ByteArrayWrapper> hash;
-  private final ConcurrentHashMap<UUID, List<ByteArrayWrapper>> hScanSnapShots;
-  private final ConcurrentHashMap<UUID, Long> hScanSnapShotCreationTimes;
+  private Object2ObjectOpenCustomHashMap<byte[], byte[]> hash;
+  private final ConcurrentHashMap<UUID, List<byte[]>> hScanSnapShots = new ConcurrentHashMap<>();
+  private final ConcurrentHashMap<UUID, Long> hScanSnapShotCreationTimes =
+      new ConcurrentHashMap<>();
   private ScheduledExecutorService HSCANSnapshotExpirationExecutor = null;
 
   private int sizeInBytes;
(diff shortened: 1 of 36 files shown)


onichols-pivotal (approved):

jdeppe-pivotal (approved):
Awesome!

sabbey37 (approved):
This is so beautiful.  Thank you for all the wonderful refactoring.

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisHashTest.java
```diff
@@ -280,58 +278,39 @@ public void constantBaseRedisHashOverhead_shouldEqualCalculatedOverhead() {
...
+      int overHeadPerField = (size - BASE_REDIS_HASH_OVERHEAD - dataSize) / fieldCount;
+      totalOverhead += overHeadPerField;
+    }
 
-    int expectedFirstPairOverhead = oneEntryHashMapSize - emptyHashMapSize
-        - RedisHash.HASH_MAP_VALUE_PAIR_OVERHEAD;
+    long averageOverhead = totalOverhead / totalFields;
 
-    assertThat(RedisHash.SIZE_OF_OVERHEAD_OF_FIRST_PAIR).isEqualTo(expectedFirstPairOverhead);
+    assertThat(RedisHash.HASH_MAP_VALUE_PAIR_OVERHEAD).isEqualTo(averageOverhead);
```
DonalEvans, line 297: This test fails if you use a value of 2000 for `totalFields`, because the actual value of `averageOverhead` is very close to 48 but can vary slightly to be either greater then or less than 48. With `totalFields` = 1000, the value is 47.931, which is rounded down to 47, but with `totalFields` = 2000, the value is 48.0895, which rounds down to 48. This can be remedied by making `totalOverhead` a float and calling `Math.round(totalOverhead / totalFields)` to determine the value of `averageOverhead`, and changing the value of HASH_MAP_VALUE_PAIR_OVERHEAD to 48.

upthewaterspout: Good point! I've updated the PR to use that approach.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/hash/HGetExecutor.java
```diff
@@ -42,11 +41,10 @@ public RedisResponse executeCommand(Command command,
...
     List<byte[]> commandElems = command.getProcessedCommand();
 
-    byte[] byteField = commandElems.get(FIELD_INDEX);
-    ByteArrayWrapper field = new ByteArrayWrapper(byteField);
+    byte[] field = commandElems.get(FIELD_INDEX);
     RedisKey key = command.getKey();
-    RedisHashCommands redisHashCommands = createRedisHashCommands(context);
-    ByteArrayWrapper valueWrapper = redisHashCommands.hget(key, field);
+    RedisHashCommands redisHashCommands = context.getRedisHashCommands();
+    byte[] valueWrapper = redisHashCommands.hget(key, field);
```
DonalEvans, line 47: Small nitpick, but could this be renamed to just "value"?

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisHash.java
```diff
@@ -65,50 +68,43 @@
...
-    this();
+    sizeInBytes = BASE_REDIS_HASH_OVERHEAD;
     HSCAN_SNAPSHOTS_EXPIRE_CHECK_FREQUENCY_MILLISECONDS = hscanSnapShotExpirationCheckFrequency;
     MINIMUM_MILLISECONDS_FOR_HSCAN_SNAPSHOTS_TO_LIVE = minimumLifeForHscanSnaphot;
 
-    Iterator<ByteArrayWrapper> iterator = fieldsToSet.iterator();
+    hash = new Object2ObjectOpenCustomHashMap<>(fieldsToSet.size() / 2, ByteArrays.HASH_STRATEGY);
+    Iterator<byte[]> iterator = fieldsToSet.iterator();
     while (iterator.hasNext()) {
       hashPut(iterator.next(), iterator.next());
```
DonalEvans, line 96: I know it's outside the scope of the changes being made in this PR, but there's a possibility for a NoSuchElementException here if `fieldsToSet` doesn't have an even number of elements. Is it worth putting in a check for that, or something to handle it?

DonalEvans (approved):
Just a couple of minor nitpicks, feel free to disregard them, but other than that, looks good.

rhoughton-pivot (approved):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisHash.java
```diff
@@ -65,50 +68,43 @@
...
-    hash = new HashMap<>();
-    hScanSnapShots = new ConcurrentHashMap<>();
-    hScanSnapShotCreationTimes = new ConcurrentHashMap<>();
-
-    HSCAN_SNAPSHOTS_EXPIRE_CHECK_FREQUENCY_MILLISECONDS = defaultHscanSnapshotsExpireCheckFrequency;
-
-    MINIMUM_MILLISECONDS_FOR_HSCAN_SNAPSHOTS_TO_LIVE = defaultHscanSnapshotsMillisecondsToLive;
-
-    sizeInBytes = BASE_REDIS_HASH_OVERHEAD;
+    // For serialization only!
```
sabbey37, line 107: Maybe not necessary to do in this PR, but `sizeInBytes` was set in a similar way in the empty `RedisSet` constructor. We could remove it there as well.

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisHashTest.java
```diff
@@ -271,67 +269,64 @@ public void hscanSnaphots_shouldExpireAfterExpiryPeriod() {
...
-
-    tempHashmap.put(field, value);
-    int oneEntryHashMapSize = reflectionObjectSizer.sizeof(tempHashmap);
-
-    int expectedFirstPairOverhead = oneEntryHashMapSize - emptyHashMapSize
-        - RedisHash.HASH_MAP_VALUE_PAIR_OVERHEAD;
+      // Compute per field overhead with this number of fields
+      int overHeadPerField = (size - BASE_REDIS_HASH_OVERHEAD - dataSize) / fieldCount;
+      totalOverhead += overHeadPerField;
+    }
```
sabbey37, line 309: Good call on this test, @DonalEvans and @upthewaterspout!  This is great.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisHash.java
sabbey37: That's such a good call out, Donal!  Could `fieldsToSet` ever be odd?  It seems like it could only happen if we intentionally create a new `RedisHash` from a test.  But maybe it's good to put a check anyway just in case?

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisHashTest.java
```diff
@@ -271,67 +269,64 @@ public void hscanSnaphots_shouldExpireAfterExpiryPeriod() {
...
-    ByteArrayWrapper value1 = new ByteArrayWrapper("b".getBytes());
-    ByteArrayWrapper field2 = new ByteArrayWrapper("c".getBytes());
-    ByteArrayWrapper value2 = new ByteArrayWrapper("d".getBytes());
+    // Generate pseudo-random data, but use fixed seed so the test is deterministic
+    Random random = new Random(0);
 
-    tempHashmap.put(field1, value1);
-    int oneEntryHashMapSize = reflectionObjectSizer.sizeof(tempHashmap);
+    // Add 1000 fields and compute the per field overhead after each add
+    for (int fieldCount = 1; fieldCount < totalFields; fieldCount++) {
```
sabbey37, line 295: Good call on this test, @DonalEvans and @upthewaterspout ! This is great.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisHash.java
```diff
@@ -105,7 +104,7 @@ public RedisHash(List<byte[]> fieldsToSet) {
   }
 
   public RedisHash() {
-    sizeInBytes = BASE_REDIS_HASH_OVERHEAD;
+    // For serialization only!
```
pivotal-jbarrett, line 107: If this is used only for serialization:
1. It probably needs an annotation to suppress the unused compile warning.
2. Should be document in the Javadocs so that its visible to the consumer of the class without looking at the source what the intent for the method is.

pivotal-jbarrett (changes requested):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/hash/HGetExecutor.java
upthewaterspout: done.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisHash.java
upthewaterspout: done.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisHash.java
upthewaterspout: I moved it to the field initializer instead. This sizeInBytes is always going to get reset when this constructor is used, because it is for deserialization and we are serializing the size.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisHash.java
upthewaterspout: No compile warnings, so we're good there. I added the javadocs.

pivotal-jbarrett (approved):

PR #6435, opened 2021-05-05
GEODE-9235: Allow binding to all addresses

alb3rtobr:
**AS A** geode contributor
**I WANT TO** be able to start members with option `--bind-address=0.0.0.0`, `--bind-address=::` or `--bind-address=*`
**SO THAT** all TCP endpoints are bound to all interfaces.

It is not currently possible to start cluster members binding to all interfaces. That is using `--bind-address=0.0.0.0` for ipv4 or `--bind-address=::` for ipv6.

Geode binds the locator and server traffic port by default to `0.0.0.0`, but the membership ports are bound by default to the local address.

Changing the default membership binding is not the best option, as it could cause problems in existing Geode deployments. But using `--bind-address=0.0.0.0`, `--bind-address=::` or `--bind-address=*` should be accepted.

There is a use case that needs this binding, that is using Istio with Geode. One of the requirements set by Istio towards applications trying to integrate with it is that the application listening ports need to be bound to 0.0.0.0 address (which listens on all interfaces).

It is currently not possible to start a cluster using wildcard address due to if `--bind-address` is specified, its value will be used for all tcp endpoints and the udp membership endpoint created by JGroups. And as JGroups does not accept binding to all addresses, and exception is thrown. This behavior can be verified with `WildcardAddressBindingDUnitTest.java` tests.


dschneider-pivotal (approved):
I didn't see any problems with this change. Consider the minor refactor I suggested

mhansonp:
Great description!

rhoughton-pivot (approved):

alb3rtobr:
@dschneider-pivotal I have implemented your suggestion. I was waiting some days to do it in case more reviews with suggestions were added. Thanks!

alb3rtobr:
CI failed due to flaky test (https://issues.apache.org/jira/browse/GEODE-9242 )

mhansonp (approved):

jdeppe-pivotal (approved):

jinmeiliao (approved):

kirklund (approved):

alb3rtobr:
rebase done in order to fix conflicts

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/StartLocatorCommand.java
```diff
@@ -183,6 +188,8 @@ ResultModel doStartLocator(
 
     Properties gemfireProperties = new Properties();
 
+    StartMemberUtils.setPropertyIfNotNull(gemfireProperties, ConfigurationProperties.BIND_ADDRESS,
+        bindAddress);
```
Bill, line 192: I thought it was odd that we had to call `setPropertyIfNotNull()` in this method (`doStartLocator()`) but we did not over in `StartServerCommand.doStartServer()`. Looks like it was already being called over in that other method.

Does that mean that before this code change here, we had a bug where gfsh `start locator` would ignore the bind address (before this PR)?

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/fd/GMSHealthMonitor.java
```diff
@@ -877,14 +879,32 @@ public void init(Services<ID> s) throws MembershipConfigurationException {
...
+    startTcpServer(serverSocket);
+    startHeartbeatThread();
+  }
+
+  private void createServerSocket() throws MemberStartupException {
     try {
-      serverSocket = createServerSocket(localAddress.getInetAddress(),
-          services.getConfig().getMembershipPortRange());
+      InetAddress address = localAddress.getInetAddress();
+      String bindAddrStr = services.getConfig().getBindAddress();
```
Bill, line 890: Ooh good catch remembering to fix the health monitor endpoint too!

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/messenger/JGroupsMessenger.java
```diff
@@ -281,7 +281,7 @@ public void init(Services<ID> s) throws MembershipConfigurationException {
 
     String str = config.getBindAddress();
     // JGroups UDP protocol requires a bind address
-    if (str == null || str.length() == 0) {
+    if (str == null || str.length() == 0 || LocalHostUtil.isWildcardAddress(str)) {
```
Bill, line 284: yes!

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/fd/GMSHealthMonitor.java
```diff
@@ -877,14 +879,32 @@ public void init(Services<ID> s) throws MembershipConfigurationException {
...
+        try {
+          if (LocalHostUtil.isWildcardAddress(bindAddrStr)) {
+            address = LocalHostUtil.getAnyLocalAddress();
+          } else {
+            address = InetAddress.getByName(bindAddrStr);
+          }
+        } catch (UnknownHostException e) {
+          logger.error(
+              "Error when configuring {} as bind address in membership, default address will be used. Exception: {}",
+              services.getConfig().getBindAddress(), e.getMessage());
```
Bill, line 901: I was going to recommend logging the full exception but I feel like `e.getMessage()` is sufficient and reduces log clutter ✓

Bill (changes requested):
Great PR @alb3rtobr !! In addition to comments and requests below please update the command-line options help for `LocatorLauncher` (`bind-address`) and `ServerLauncher` (`bind-address`, `server-bind-address`).

The bigger issue is that I don't see a test that verifies that if `--bind-address=*` on a multi-homed host, that both (health monitor) and (client's cache) services bind on all interfaces. I realize this isn't possible in a plain old `DUnit` test. But what about a Dockerized one along the lines of `SingleServerSNIAcceptanceTest`?

FYI and for other reviewers, this is my summary of endpoint configuration before/after this ticket:

image

alb3rtobr:
Thanks for the review @Bill . I will try the dockerized test. In the meanwhile Im moving the PR to draft.

PR #6440, opened 2021-05-05
DRAFT: Add New (Optional) Non-Differential Quorum Algorithm

Bill:
There are problems with Geode's current quorum algorithm. In particular, split-brains are possible if locators are started at the same time, or if they are started during a network partition (and can't reach other locators).

This (draft) PR introduces an alternative quorum maintenance algorithm, activated by a new optional configuration parameter `quorum-absolute-locator-count`. When it's configured to a value greater-than-zero, two things are done differently:

1. when a locator is starting or restarting, and it might potentially become the coordinator, it can become the coordinator only if it has communicated with `quorum-absolute-locator-count` locators (including itself).

2. when a coordinator is considering whether a view change might result in a loss of quorum, rather than basing that decision on the change in membership weight between the current view and the prospective one, the locator count in the prospective view is compared to `quorum-absolute-locator-count`. If the locator count is below the threshold then it's treated as lost quorum, with all the usual processing that happens in that event.

The ability to switch between the legacy quorum algorithm and this new one is accomplished via a new policy object. `GMSJoinLeave` has a reference to a `Quorum` object. Legacy behavior is captured in the `DifferentialQuorum` implementation, and the new algorithm is embodied in the `AbsoluteQuorum` implementation. `GMSJoinLeave` delegates to the `Quorum` object in the `GMSJoinLeave.isLostQuorum()` method. (That method used to be called `isNetworkPartition()` but it was renamed since quorum can be lost due to reasons other than a network partition.)

In the legacy quorum algorithm, no quorum checking is done at cluster startup (when a coordinator is chosen). The new algorithm has to operate there. As a result `GMSJoinLeave.findCoordinator()` was modified to delegate to the `Quorum` object (so that the new algorithm can prevent cluster formation if too few locators are currently reachable).

This PR is very drafty so be kind.

### Checklist:

### DIFF
--- a/geode-core/src/test/java/org/apache/geode/distributed/internal/DistributionConfigJUnitTest.java
+++ b/geode-core/src/test/java/org/apache/geode/distributed/internal/DistributionConfigJUnitTest.java
@@ -104,7 +104,7 @@ public class DistributionConfigJUnitTest {
   @Test
   public void testGetAttributeNames() {
     String[] attNames = AbstractDistributionConfig._getAttNames();
-    assertThat(attNames.length).isEqualTo(169);
+    assertThat(attNames.length).isEqualTo(170);
 
     List boolList = new ArrayList();
     List intList = new ArrayList();
@@ -139,7 +139,7 @@ public class DistributionConfigJUnitTest {
     // TODO - This makes no sense. One has no idea what the correct expected number of attributes
     // are.
     assertEquals(36, boolList.size());
-    assertEquals(35, intList.size());
+    assertEquals(36, intList.size());
     assertEquals(88, stringList.size());
     assertEquals(5, fileList.size());
     assertEquals(5, otherList.size());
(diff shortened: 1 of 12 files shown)


Bill:
Those three latest failures in distributed tests (2 in JDK8, one in JDK11) all have open tickets so they are not caused by work on this branch.

geode-core/src/test/java/org/apache/geode/distributed/internal/DistributionConfigJUnitTest.java
```diff
@@ -104,7 +104,7 @@ public void before() {
   @Test
   public void testGetAttributeNames() {
     String[] attNames = AbstractDistributionConfig._getAttNames();
-    assertThat(attNames.length).isEqualTo(169);
+    assertThat(attNames.length).isEqualTo(170);
```
kirklund, line 107: This syntax provides better failure messages:
```
assertThat(attNames).hasSize(170);
```
I'd go ahead and convert the test to use only AssertJ. There's an IntelliJ plugin that can automate this but it's not very smart, I usually end up having to touch up anything involving collections, arrays, or messages.

kirklund (commented):
I like the approach you have in this PR. Are you still working on it (it's in Draft)? I just recommend some small assertion changes.

onichols-pivotal:

onichols-pivotal (changes requested):
_comment deleted_

onichols-pivotal (approved):

metatype:
Closing this out due to age.

PR #6446, opened 2021-05-06
GEODE-9238: Restructure Hash-related DUnit tests to handle cluster mode

luissson:

### For all changes:

### Note:

### DIFF
--- a/geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/hash/HashesAndCrashesDUnitTest.java
+++ b/geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/hash/HashesAndCrashesDUnitTest.java
@@ -26,6 +26,7 @@ import static org.assertj.core.api.Assertions.assertThat;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 import java.util.Properties;
 import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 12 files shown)


geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/hash/HashesAndCrashesDUnitTest.java
```diff
@@ -147,6 +149,12 @@ public void after() {
     redisClient.shutdown();
   }
 
+  @AfterClass
+  public static void tearDown() {
+    server1.stop();
+    server2.stop();
+    server3.stop();
+  }
```
jdeppe-pivotal, line 157: This seems like the only significant change in this PR. Explicitly stopping the servers shouldn't be required since the rule should take care of cleaning up the servers. If that's not happening, then there's something wrong with the rule. Did you run into issues somewhere?

jdeppe-pivotal (changes requested):

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/hash/HashesAndCrashesDUnitTest.java
luissson: I would get a failure from stress new test when starting up a server, complaining that a server could not be started because the port was in use. Stopping the servers after the tests completed appears to have resolved that. This follows the pattern I saw in other tests for example HScanDUnitTest

luissson: As discussed earlier, the failure in HashesAndCrashes was due to a redis server binding to a port that was already claimed by the locator. Grabbing the port from getRandomAvailableTCPPorts after the locator was started appears to have resolved this failure. Thanks!

luissson: Ah, we set the ephemeral port for the locator in the call to getPort(). I've moved the redis server port grab to be below this line and we're green!

jdeppe-pivotal (approved):

PR #6460, opened 2021-05-10
DRAFT: GEODE-9123: Invoke callback events after PR clear

kirklund:
GEODE-9123: Invoke callback events after PR clear

Note: depends on #6589.

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/cache/asyncqueue/internal/InternalAsyncEventQueue.java
+++ b/geode-core/src/main/java/org/apache/geode/cache/asyncqueue/internal/InternalAsyncEventQueue.java
@@ -28,4 +28,6 @@ public interface InternalAsyncEventQueue extends AsyncEventQueue {
   void stop();
 
   void destroy();
+
+  boolean isPartitionedRegionClearUnsupported();
 }
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
@@ -8079,8 +8079,12 @@ public class PartitionedRegion extends LocalRegion
 
     private final long maxTimeInRetry;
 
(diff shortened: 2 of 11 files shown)


geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
```diff
@@ -8075,8 +8075,12 @@ public DistributionAdvisee getParentAdvisee() {
 
     private final long maxTimeInRetry;
 
-    public RetryTimeKeeper(int maxTime) {
-      this.maxTimeInRetry = maxTime;
+    public RetryTimeKeeper(int maxTimeInRetry) {
+      this.maxTimeInRetry = maxTimeInRetry;
+    }
+
```
gesterzhou, line 8081: why you need two methods here?

geode-core/src/main/java/org/apache/geode/cache/asyncqueue/internal/InternalAsyncEventQueue.java
```diff
@@ -28,4 +28,6 @@
   void stop();
 
   void destroy();
+
+  boolean isPartitionedRegionClearUnsupported();
 }
```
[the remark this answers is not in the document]
kirklund: I added it to the interface so PartitionedRegionClear.requireAsyncEventQueuesSupportPartitionedRegionClear() can depend on just the interface rather than the concrete classes.

geode-core/src/main/java/org/apache/geode/internal/cache/PartitionedRegion.java
kirklund: It's not required. I added `public RetryTimeKeeper(long maxTimeInRetry)` because most millis are in long and the caller has to downcast with `(int)` which causes a warning in IntelliJ, and I was trying to reduce the number of warnings.

PR #6461, opened 2021-05-10
GEODE-9256: Use a stateless hscan implementation

upthewaterspout:
Adding a new subclass, Object2ObjectOpenCustomHashMapWithCursor, that
implements the redis HSCAN operation. This method uses the same algorithm as
redis by incrementing the reverse of the cursor for each step. See
https://github.com/redis/redis/blob/e504583b7806d946da9c3627784d551a742be4d0/src/dict.c#L651

Replacing our stateful hscan implementation with a stateless implementation
based on the same algorithm as redis. This uses the stateless cursor method above.

### DIFF
--- a/geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/hash/AbstractHScanIntegrationTest.java
+++ b/geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/hash/AbstractHScanIntegrationTest.java
@@ -453,7 +453,7 @@ public abstract class AbstractHScanIntegrationTest implements RedisIntegrationTe
     ScanResult<Map.Entry<String, String>> result = jedis.hscan("colors", "5");
 
     assertThat(new HashSet<>(result.getResult()))
-        .containsExactlyInAnyOrderElementsOf(data.entrySet());
+        .isSubsetOf(data.entrySet());
   }
 
   /**** Concurrency ***/
--- a/geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/hash/HScanIntegrationTest.java
+++ b/geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/hash/HScanIntegrationTest.java
@@ -17,17 +17,12 @@ package org.apache.geode.redis.internal.executor.hash;
 
 
 import static org.apache.geode.redis.internal.RedisConstants.ERROR_CURSOR;
-import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 import java.math.BigInteger;
-import java.util.HashMap;
-import java.util.Map;
 
 import org.junit.ClassRule;
 import org.junit.Test;
-import redis.clients.jedis.ScanParams;
-import redis.clients.jedis.ScanResult;
 
 import org.apache.geode.redis.GeodeRedisServerRule;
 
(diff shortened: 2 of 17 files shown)


onichols-pivotal (approved):
junit-quickcheck is MIT license which is compatible with Apache :)

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/hash/AbstractHScanIntegrationTest.java
```diff
@@ -453,7 +453,7 @@ public void should_notErrorGivenNonzeroCursorOnFirstCall() {
     ScanResult<Map.Entry<String, String>> result = jedis.hscan("colors", "5");
 
     assertThat(new HashSet<>(result.getResult()))
-        .containsExactlyInAnyOrderElementsOf(data.entrySet());
+        .isSubsetOf(data.entrySet());
```
upthewaterspout, line 456: The behavior of hscan when you start with a non-zero cursor is undefined, except that it should not return an error. So we can only reasonably expect that we might see some but not necessarily all of the elements of the hash.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/hash/HScanIntegrationTest.java
```diff
@@ -64,27 +59,4 @@ public void givenCursorLessThanIntMinValue_returnsCursorError() {
...
-    scanParams.count(COUNT_PARAM);
-    ScanResult<Map.Entry<byte[], byte[]>> result;
-
-    String cursor = "0";
-
-    result = jedis.hscan("colors".getBytes(), cursor.getBytes(), scanParams);
-
-    assertThat(result.getResult().size()).isEqualTo(COUNT_PARAM);
-  }
-
```
upthewaterspout, line 89: hscan is allowed to return more than `count` number of entries, due to hash collisions. This test would now fail with both native redis and radish.

pivotal-jbarrett (approved):

rhoughton-pivot (approved):

jdeppe-pivotal (approved):
Some really cool stuff here!! I like the 'Quick Tests'.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/collections/Object2ObjectOpenCustomHashMapWithCursor.java
```diff
@@ -0,0 +1,185 @@
...
+    // for a redis style closed addressing hashtable.
+    //
+
+    do {
+      // Emit all of the entries at the cursor. This means looking forward in the hash
+      // table for any non-null entries that might hash to the current cursor and emitting
+      // those as well. This may even wrap around to the front of the hashtable.
+      int position = cursor;
+      while (key[position & mask] != null) {
+        if (elementAtHashesTo(position, cursor & mask)) {
```
nonbinaryprogrammer, line 125: this method name reads kinda weird. maybe `elementAtCursorHashesTo`?

nonbinaryprogrammer (approved):
I like this solution! feel free to ignore my naming comment, it's not a big deal, but if you're making other changes

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/collections/Object2ObjectOpenCustomHashMapWithCursor.java
upthewaterspout: I'm going to merge this and create a new PR with a JMH test for this. I'll rename this method in the next PR. Thanks!

PR #6464, opened 2021-05-11
GEODE-9221: Remove uses of ByteArrayWrapper from RedisString

DonalEvans:
- Replace uses of ByteArrayWrapper in RedisString and NullRedisString
 with byte[]
 - Refactor to remove warnings

### For all changes:

### Note:


geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisString.java
```diff
@@ -718,15 +709,16 @@ protected void handleSetExpiration(SetOptions options) {
...
+    int additionalLength = bytes.length;
+    byte[] combined = new byte[initialLength + additionalLength];
+    System.arraycopy(value, 0, combined, 0, initialLength);
+    System.arraycopy(bytes, 0, combined, initialLength, additionalLength);
+    value = combined;
   }
 
-  protected void valueSetBytes(byte[] bytes) {
-    value.setBytes(bytes);
+  protected void valueSet(byte[] bytes) {
```
jdeppe-pivotal, line 720: This class uses a mix of this method and direct assignment to `value`. I'd prefer just one approach. Don't mind which, but would lean towards just direct `value = ...`

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisString.java
```diff
@@ -718,15 +709,16 @@ protected void handleSetExpiration(SetOptions options) {
   ////// methods that modify the "value" field ////////////
 
   protected void valueAppend(byte[] bytes) {
-    value.append(bytes);
-  }
-
-  protected void valueSet(ByteArrayWrapper newValue) {
-    value = newValue;
+    int initialLength = value.length;
```
jdeppe-pivotal, line 712: Since this is being pulled from `ByteArrayWrapper`, I think it can be deleted from that class now.

jdeppe-pivotal (changes requested):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisString.java
```diff
@@ -233,18 +227,17 @@ public int bitpos(Region<RedisKey, RedisData> region, RedisKey key, int bit,
     if (start > length) {
       start = length - 1;
     }
-    if (end > length) {
+    if (end >= length) {
```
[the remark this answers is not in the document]
sabbey37: Definitely appreciate the find, but I'm worried about scope creep here.  `BITPOS` is still an unsupported command, and there hasn't been a story to fully implement it and test it yet.  I am not sure where it is on our list of priorities.

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisStringTest.java
```diff
@@ -313,6 +317,148 @@ public void setExpirationTimestamp_stores_delta_that_is_stable() throws IOExcept
...
+    RedisString string = new RedisString(new byte[] {0, 1});
+    assertThat(string.bitpos(1, 1, 0)).isEqualTo(-1);
+  }
+
+  @Test
+  public void bitposReturnsNegativeOneWhenBitIsNotZeroOrOne() {
+    RedisString string = new RedisString(new byte[] {0, 1});
+    assertThat(string.bitpos(2, 0, 1)).isEqualTo(-1);
+  }
+
```
sabbey37: Where possible, we favor adding these types of tests at the integration level (see `AbstractBitPosIntegrationTest`) so we can make sure our behavior matches Native Redis. Could we add these there instead?

DonalEvans: Ah, sorry for the mistake. I'll look over the contents of `AbstractBitPosIntegrationTest` and expand the coverage if anything's not there that I added here.

DonalEvans: I left one tase case in the unit test, as it's testing behaviour specific to the implementation in RedisString that can't be exercised via an integration test.

jdeppe-pivotal (approved):

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/string/AbstractBitPosIntegrationTest.java
```diff
@@ -115,11 +123,69 @@ public void bitposWithStart_givenOneInLastByte() {
...
   }
 
+  @Test
+  public void bitposWithStart_givenStartGreaterThanOrEqualToByteArrayLength() {
+    byte[] key = {1, 2, 3};
+    byte[] bytes = {1, 1, 1, 1};
+    jedis.set(key, bytes);
+    assertThat(jedis.bitpos(key, true, new BitPosParams(bytes.length))).isEqualTo(7 + 3 * 8);
+    assertThat(jedis.bitpos(key, true, new BitPosParams(bytes.length + 1))).isEqualTo(7 + 3 * 8);
+  }
```
sabbey37, line 133: Looks like Native Redis doesn't check whether the start is greater than or equal to the length of the value beforehand and just returns -1 if start is greater than end:
```
    /* For empty ranges (start > end) we return -1 as an empty range does
     * not contain a 0 nor a 1. */
    if (start > end) {
        addReplyLongLong(c, -1);
```

I definitely feel like we've gotten beyond the scope of this Jira ticket/PR though.

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisStringTest.java
sabbey37: The reason it can't be exercised via an integration test is because we catch it and throw an error in the `BitPosExecutor`:
```
 if (bit != 0 && bit != 1) {
      return RedisResponse.error(ERROR_BIT);
    }
```
So I'm not sure we need a unit test for it... though unit tests are cheap and more tests can't hurt. I guess this goes back to our testing strategy, which we can discuss with the team.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/string/AbstractBitPosIntegrationTest.java
DonalEvans: Agreed. I'll remove this test case. If/when bitpos becomes a supported command, proper coverage can be established.

sabbey37 (approved):

PR #6467, opened 2021-05-11
GEODE-9265: Introduce redis SlotAdvisor to manage slot information

jdeppe-pivotal:
We're currently using a function to pull bucket/slot info from each
member. However, members already contain the most current view of bucket
locations so we should use that instead. This will avoid an unnecessary
fn call as well as providing more up-to-date information.

This commit introduces a `SlotAdvisor` which is intended to be the
central point of information regarding buckets/slots/members. This will
also be used to support the MOVED response.

### For all changes:

### Note:

### DIFF
--- a/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/GeodeRedisServer.java
+++ b/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/GeodeRedisServer.java
@@ -85,6 +85,7 @@ public class GeodeRedisServer {
 
     CommandFunction.register(regionProvider.getDataRegion(), stripedExecutor, redisStats);
     RenameFunction.register(regionProvider.getDataRegion(), stripedExecutor, redisStats);
+    RedisMemberInfoRetrievalFunction infoFunction = RedisMemberInfoRetrievalFunction.register();
 
     passiveExpirationManager =
         new PassiveExpirationManager(regionProvider.getDataRegion(), redisStats);
@@ -99,7 +100,7 @@ public class GeodeRedisServer {
         this::allowUnsupportedCommands, this::shutdown, port, bindAddress, redisStats,
         redisCommandExecutor, member, commandHelper);
 
-    RedisMemberInfoRetrievalFunction.register(bindAddress, nettyRedisServer.getPort());
+    infoFunction.initialize(bindAddress, nettyRedisServer.getPort());
   }
 
   @VisibleForTesting
--- a/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/GeodeRedisServer.java
+++ b/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/GeodeRedisServer.java
@@ -28,7 +28,8 @@ import org.apache.geode.internal.statistics.StatisticsClock;
 import org.apache.geode.internal.statistics.StatisticsClockFactory;
 import org.apache.geode.logging.internal.executors.LoggingExecutors;
 import org.apache.geode.logging.internal.log4j.api.LogService;
-import org.apache.geode.redis.internal.cluster.BucketInfoRetrievalFunction;
+import org.apache.geode.redis.internal.cluster.RedisMemberInfoRetrievalFunction;
+import org.apache.geode.redis.internal.data.CommandHelper;
 import org.apache.geode.redis.internal.executor.CommandFunction;
 import org.apache.geode.redis.internal.executor.StripedExecutor;
 import org.apache.geode.redis.internal.executor.SynchronizedStripedExecutor;
@@ -79,6 +80,8 @@ public class GeodeRedisServer {
     redisStats = createStats(cache);
     StripedExecutor stripedExecutor = new SynchronizedStripedExecutor();
     regionProvider = new RegionProvider(cache);
+    CommandHelper commandHelper =
+        new CommandHelper(regionProvider.getDataRegion(), redisStats, stripedExecutor);
 
     CommandFunction.register(regionProvider.getDataRegion(), stripedExecutor, redisStats);
     RenameFunction.register(regionProvider.getDataRegion(), stripedExecutor, redisStats);
@@ -94,9 +97,9 @@ public class GeodeRedisServer {
     nettyRedisServer = new NettyRedisServer(() -> cache.getInternalDistributedSystem().getConfig(),
         regionProvider, pubSub,
         this::allowUnsupportedCommands, this::shutdown, port, bindAddress, redisStats,
-        redisCommandExecutor, member);
+        redisCommandExecutor, member, commandHelper);
 
-    BucketInfoRetrievalFunction.register(bindAddress, nettyRedisServer.getPort());
+    RedisMemberInfoRetrievalFunction.register(bindAddress, nettyRedisServer.getPort());
   }
 
   @VisibleForTesting
--- a/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/GeodeRedisServer.java
+++ b/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/GeodeRedisServer.java
@@ -79,13 +79,14 @@ public class GeodeRedisServer {
     pubSub = new PubSubImpl(new Subscriptions());
     redisStats = createStats(cache);
     StripedExecutor stripedExecutor = new SynchronizedStripedExecutor();
+    RedisMemberInfoRetrievalFunction infoFunction = RedisMemberInfoRetrievalFunction.register();
+
     regionProvider = new RegionProvider(cache);
     CommandHelper commandHelper =
         new CommandHelper(regionProvider.getDataRegion(), redisStats, stripedExecutor);
 
     CommandFunction.register(regionProvider.getDataRegion(), stripedExecutor, redisStats);
     RenameFunction.register(regionProvider.getDataRegion(), stripedExecutor, redisStats);
-    RedisMemberInfoRetrievalFunction infoFunction = RedisMemberInfoRetrievalFunction.register();
 
     passiveExpirationManager =
         new PassiveExpirationManager(regionProvider.getDataRegion(), redisStats);
@@ -100,7 +101,7 @@ public class GeodeRedisServer {
         this::allowUnsupportedCommands, this::shutdown, port, bindAddress, redisStats,
         redisCommandExecutor, member, commandHelper);
 
-    infoFunction.initialize(bindAddress, nettyRedisServer.getPort());
+    infoFunction.initialize(member, bindAddress, nettyRedisServer.getPort());
   }
 
   @VisibleForTesting
--- a/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/SlotAdvisor.java
+++ b/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/SlotAdvisor.java
@@ -148,6 +148,7 @@ public class SlotAdvisor {
       resultCollector =
           FunctionService.onRegion(dataRegion).execute(RedisMemberInfoRetrievalFunction.ID);
     } catch (FunctionException e) {
+      logger.warn("Unable to execute {}: {}", RedisMemberInfoRetrievalFunction.ID, e.getMessage());
       return null;
     }
 
--- a/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/SlotAdvisor.java
+++ b/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/SlotAdvisor.java
@@ -29,7 +29,6 @@ import org.apache.logging.log4j.Logger;
 import org.apache.geode.cache.Region;
 import org.apache.geode.cache.execute.FunctionService;
 import org.apache.geode.cache.execute.ResultCollector;
-import org.apache.geode.cache.partition.PartitionRegionHelper;
 import org.apache.geode.distributed.DistributedMember;
 import org.apache.geode.distributed.internal.membership.InternalDistributedMember;
 import org.apache.geode.internal.cache.PartitionedRegion;
@@ -42,13 +41,13 @@ import org.apache.geode.redis.internal.data.RedisKey;
 public class SlotAdvisor {
 
   private static final Logger logger = LogService.getLogger();
-  private static final int HOSTPORT_RETRIEVAL_ATTEMPTS = 10;
-  private static final int HOSTPORT_RETRIEVAL_INTERVAL = 6_000;
+  private static final int HOSTPORT_RETRIEVAL_ATTEMPTS = 20;
+  private static final int HOSTPORT_RETRIEVAL_INTERVAL = 100;
 
   /**
    * Cache of member ids to member IP address and redis listening port
    */
-  private final Map<DistributedMember, Pair<String, Integer>> hostPorts = new HashMap<>();
+  private final Map<DistributedMember, RedisMemberInfo> memberInfos = new HashMap<>();
   private final PartitionedRegion dataRegion;
 
   SlotAdvisor(Region<RedisKey, RedisData> dataRegion) {
@@ -61,12 +60,16 @@ public class SlotAdvisor {
   }
 
   public Pair<String, Integer> getHostAndPortForKey(RedisKey key) {
-    return getHostPort(key.getBucketId());
+    try {
+      RedisMemberInfo memberInfo = getMemberInfo(key.getBucketId());
+      return Pair.of(memberInfo.getHostAddress(), memberInfo.getRedisPort());
+    } catch (InterruptedException ex) {
+      Thread.currentThread().interrupt();
+      throw new RuntimeException(ex);
+    }
   }
 
   public Map<String, List<Integer>> getMemberBuckets() {
-    initializeBucketsIfNecessary();
-
     Map<String, List<Integer>> memberBuckets = new HashMap<>();
     for (int bucketId = 0; bucketId < REDIS_REGION_BUCKETS; bucketId++) {
       String memberId = getOrCreateMember(bucketId).getUniqueId();
@@ -81,15 +84,19 @@ public class SlotAdvisor {
    * the details for a given bucket cannot be determined, that entry will contain {@code null}.
    */
   public synchronized List<MemberBucketSlot> getBucketSlots() {
-    initializeBucketsIfNecessary();
-
     List<MemberBucketSlot> memberBucketSlots = new ArrayList<>(RegionProvider.REDIS_REGION_BUCKETS);
-    for (int bucketId = 0; bucketId < REDIS_REGION_BUCKETS; bucketId++) {
-      Pair<String, Integer> hostPort = getHostPort(bucketId);
-      if (hostPort != null) {
-        memberBucketSlots.add(
-            new MemberBucketSlot(bucketId, hostPort.getLeft(), hostPort.getRight()));
+    try {
+      for (int bucketId = 0; bucketId < REDIS_REGION_BUCKETS; bucketId++) {
+        RedisMemberInfo memberInfo = getMemberInfo(bucketId);
+        if (memberInfo != null) {
+          memberBucketSlots.add(
+              new MemberBucketSlot(bucketId, memberInfo.getHostAddress(),
+                  memberInfo.getRedisPort()));
+        }
       }
+    } catch (InterruptedException ex) {
+      Thread.currentThread().interrupt();
+      throw new RuntimeException(ex);
     }
 
     return memberBucketSlots;
@@ -99,31 +106,20 @@ public class SlotAdvisor {
     return dataRegion.getOrCreateNodeForBucketWrite(bucketId, null);
   }
 
-  private void initializeBucketsIfNecessary() {
-    if (dataRegion.getDataStore() != null &&
-        dataRegion.getDataStore().getAllLocalBucketIds().isEmpty()) {
-      PartitionRegionHelper.assignBucketsToPartitions(dataRegion);
-    }
-  }
-
   /**
    * This method will retry for {@link #HOSTPORT_RETRIEVAL_ATTEMPTS} attempts and return null if
    * no information could be retrieved.
    */
-  private Pair<String, Integer> getHostPort(int bucketId) {
-    Pair<String, Integer> response;
+  private RedisMemberInfo getMemberInfo(int bucketId) throws InterruptedException {
+    RedisMemberInfo response;
 
     for (int i = 0; i < HOSTPORT_RETRIEVAL_ATTEMPTS; i++) {
-      response = getHostPort0(bucketId);
+      response = getMemberInfo0(bucketId);
       if (response != null) {
         return response;
       }
 
-      try {
-        Thread.sleep(HOSTPORT_RETRIEVAL_INTERVAL);
-      } catch (InterruptedException e) {
-        throw new RuntimeException(e);
-      }
+      Thread.sleep(HOSTPORT_RETRIEVAL_INTERVAL);
     }
 
     logger.error("Unable to retrieve host and redis port for member with bucketId: {}", bucketId);
@@ -132,11 +128,11 @@ public class SlotAdvisor {
   }
 
   @SuppressWarnings("unchecked")
-  private Pair<String, Integer> getHostPort0(int bucketId) {
+  private RedisMemberInfo getMemberInfo0(int bucketId) {
     InternalDistributedMember member = getOrCreateMember(bucketId);
 
-    if (hostPorts.containsKey(member)) {
-      return hostPorts.get(member);
+    if (memberInfos.containsKey(member)) {
+      return memberInfos.get(member);
     }
 
     List<RedisMemberInfo> memberInfos;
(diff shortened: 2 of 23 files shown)


sabbey37 (approved):
Looks good!  Would be good for @upthewaterspout to review as well.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/SlotAdvisor.java
```diff
@@ -0,0 +1,230 @@
...
+      return hostPorts.get(memberId);
+    }
+
+    Set<DistributedMember> membersWithDataRegion = new HashSet<>();
+    for (PartitionMemberInfo memberInfo : getRegionMembers(dataRegion)) {
+      membersWithDataRegion.add(memberInfo.getDistributedMember());
+    }
+
+    ResultCollector<RedisMemberInfo, List<RedisMemberInfo>> resultCollector =
+        FunctionService.onMembers(membersWithDataRegion)
```
upthewaterspout, line 148: This should just use onRegion instead of onMembers, and then you don't need to lookup which members have the region.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/SlotAdvisor.java
```diff
@@ -0,0 +1,230 @@
...
+    for (RedisMemberInfo memberInfo : resultCollector.getResult()) {
+      Pair<String, Integer> hostPort =
+          Pair.of(memberInfo.getHostAddress(), memberInfo.getRedisPort());
+      hostPorts.put(memberInfo.getMemberId(), hostPort);
+    }
+
+    if (!hostPorts.containsKey(memberId)) {
+      // There is a very tiny window where this might happen - a member was hosting a bucket and
+      // died before the fn call above could even complete.
+      throw new RuntimeException("Unable to retrieve host and redis port for member: " + memberId);
```
upthewaterspout, line 162: This doesn't seem like a good user experience to get exceptions in this situation - should this trigger a retry?

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/SlotAdvisor.java
```diff
@@ -0,0 +1,230 @@
...
+import org.apache.geode.redis.internal.data.RedisKey;
+
+public class SlotAdvisor {
+
+  private static final Logger logger = LogService.getLogger();
+
+  /**
+   * Mapping buckets to slot information
+   */
+  private final List<MemberBucketSlot> memberBucketSlots;
```
upthewaterspout, line 53: It seems like we don't really need to keep this as a field. For every call it looks like we're updating this list this the new primary locations (or at least we should be!). It would be clearer and cleaner not to have this as a field at all.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/SlotAdvisor.java
```diff
@@ -0,0 +1,230 @@
...
+    return Pair.of(mbs.getPrimaryIpAddress(), mbs.getPrimaryPort());
+  }
+
+  public Map<String, List<Integer>> getMemberBuckets() {
+    initializeBucketsIfNecessary();
+
+    Map<String, List<Integer>> memberBuckets = new HashMap<>();
+    for (int bucketId = 0; bucketId < REDIS_REGION_BUCKETS; bucketId++) {
+      String memberId =
+          dataRegion.getRegionAdvisor().getBucketAdvisor(bucketId).getPrimary().getUniqueId();
```
upthewaterspout, line 89: Maybe it would be better to keep member ids as `InternalDistributedMember` in most places until we actually return a string to the user? I normally don't like converting objects to strings and doing comparisons on the strings.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/SlotAdvisor.java
```diff
@@ -0,0 +1,230 @@
...
+  }
+
+  private void initializeBucketsIfNecessary() {
+    if (dataRegion.getDataStore() != null &&
+        dataRegion.getDataStore().getAllLocalBucketIds().isEmpty()) {
+      PartitionRegionHelper.assignBucketsToPartitions(dataRegion);
+    }
+  }
+
+  @SuppressWarnings("unchecked")
```
upthewaterspout, line 133: Why do we need a suppress warnings?

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/SlotAdvisor.java
```diff
@@ -0,0 +1,230 @@
...
+  }
+
+  private Set<PartitionMemberInfo> getRegionMembers(PartitionedRegion dataRegion) {
+    PartitionRegionInfo info = PartitionRegionHelper.getPartitionRegionInfo(dataRegion);
+    assert info != null; // Mostly to appease IJ since the region is always a PR
+
+    return info.getPartitionMemberInfo();
+  }
+
+  public static class MemberBucketSlot {
```
upthewaterspout, line 175: The `Integer` objects below could be `int` I think.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/cluster/RedisMemberInfoRetrievalFunction.java
```diff
@@ -53,49 +47,31 @@ private BucketInfoRetrievalFunction(String address, int redisPort) {
...
-    PartitionedRegion pr = (PartitionedRegion) region;
-    if (pr.getDataStore() != null) {
-      return new HashSet<>(pr.getDataStore().getAllLocalPrimaryBucketIds());
-    } else {
-      return Collections.EMPTY_SET;
-    }
-  }
-
-  public static class MemberBuckets implements Serializable {
+  public static class RedisMemberInfo implements Serializable {
```
upthewaterspout, line 66: This should probably be BasicSerializable or even DataSerializableFixedID, rather than a java serializable.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/cluster/ClusterExecutor.java
```diff
@@ -145,28 +91,34 @@ private RedisResponse getSlots(ExecutionHandlerContext ctx) {
...
+    List<SlotAdvisor.MemberBucketSlot> memberBucketSlots =
+        ctx.getRegionProvider().getSlotAdvisor().getBucketSlots();
 
     StringBuilder response = new StringBuilder();
-    for (MemberBuckets m : memberBuckets) {
-      response.append(String.format("%s %s:%d@%d master",
-          m.getMemberId(), m.getHostAddress(), m.getPort(), m.getPort()));
+    for (Map.Entry<String, List<Integer>> member : memberBuckets.entrySet()) {
+      List<Integer> buckets = member.getValue();
+      SlotAdvisor.MemberBucketSlot mbs = memberBucketSlots.get(buckets.get(0));
```
upthewaterspout, line 107: Something seems tangled here if we have to assume that buckets have at least 1 entry and we are always calling get(0).

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/cluster/ClusterExecutor.java
```diff
@@ -145,28 +91,34 @@ private RedisResponse getSlots(ExecutionHandlerContext ctx) {
...
+      response.append(String.format("%s %s:%3$d@%3$d master",
+          member.getKey(), mbs.getPrimaryIpAddress(), mbs.getPrimaryPort()));
+
+      if (member.getKey().equals(memberId)) {
         response.append(",myself");
       }
       response.append(" - 0 0 1 connected");
 
-      for (Integer index : m.getPrimaryBucketIds()) {
+      for (Integer bucket : member.getValue()) {
```
upthewaterspout, line 117: `int` instead of `Integer`

geode-apis-compatible-with-redis/src/main/resources/org/apache/geode/redis/internal/sanctioned-geode-apis-compatible-with-redis-serializables.txt
```diff
@@ -1,8 +1,8 @@
 org/apache/geode/redis/internal/ParameterRequirements/RedisParametersMismatchException,true,-643700717871858072
 org/apache/geode/redis/internal/RedisCommandSupportLevel,false
 org/apache/geode/redis/internal/RedisCommandType,false,deferredParameterRequirements:org/apache/geode/redis/internal/ParameterRequirements/ParameterRequirements,executor:org/apache/geode/redis/internal/executor/Executor,parameterRequirements:org/apache/geode/redis/internal/ParameterRequirements/ParameterRequirements,supportLevel:org/apache/geode/redis/internal/RedisCommandSupportLevel
-org/apache/geode/redis/internal/cluster/BucketInfoRetrievalFunction,false,hostAddress:java/lang/String,redisPort:int
-org/apache/geode/redis/internal/cluster/BucketInfoRetrievalFunction$MemberBuckets,false,hostAddress:java/lang/String,memberId:java/lang/String,port:int,primaryBucketIds:java/util/Set
+org/apache/geode/redis/internal/cluster/RedisMemberInfoRetrievalFunction,true,2207969011229079993,hostAddress:java/lang/String,redisPort:int
```
upthewaterspout, line 4: I think this function should really be in the excludedClasses instead. It's not sent over the wire.

upthewaterspout (changes requested):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/SlotAdvisor.java
```diff
@@ -0,0 +1,230 @@
...
+
+  /**
+   * Mapping buckets to slot information
+   */
+  private final List<MemberBucketSlot> memberBucketSlots;
+
+  /**
+   * Cache of member ids to member IP address and redis listening port
+   */
+  private final Map<String, Pair<String, Integer>> hostPorts = new HashMap<>();
```
upthewaterspout, line 58: I think this should be keyed by InternalDistributedMember, and not the member id converted to a string.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/SlotAdvisor.java
jdeppe-pivotal: Done.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/SlotAdvisor.java
jdeppe-pivotal: I've added a retry loop now. Probably even less likely to happen now that we're recreating the buckets.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/SlotAdvisor.java
jdeppe-pivotal: I've added a retry loop now. Probably even less likely to happen now that we're recreating the buckets.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/SlotAdvisor.java
jdeppe-pivotal: Done.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/cluster/RedisMemberInfoRetrievalFunction.java
jdeppe-pivotal: Now implements `DataSerializableFixedID`

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/cluster/ClusterExecutor.java
jdeppe-pivotal: Done.

geode-apis-compatible-with-redis/src/main/resources/org/apache/geode/redis/internal/sanctioned-geode-apis-compatible-with-redis-serializables.txt
jdeppe-pivotal: Right - done.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/SlotAdvisor.java
jdeppe-pivotal: Done.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/cluster/ClusterExecutor.java
jdeppe-pivotal: Actually it assumes that members (nodes) have at least one bucket. I'd prefer to have just one data structure that handles both slots:nodes and nodes:slots but I can't figure out how to avoid a bunch of redundancy.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/GeodeRedisServer.java
```diff
@@ -94,9 +97,9 @@ public GeodeRedisServer(String bindAddress, int port, InternalCache cache) {
     nettyRedisServer = new NettyRedisServer(() -> cache.getInternalDistributedSystem().getConfig(),
         regionProvider, pubSub,
         this::allowUnsupportedCommands, this::shutdown, port, bindAddress, redisStats,
-        redisCommandExecutor, member);
+        redisCommandExecutor, member, commandHelper);
 
-    BucketInfoRetrievalFunction.register(bindAddress, nettyRedisServer.getPort());
+    RedisMemberInfoRetrievalFunction.register(bindAddress, nettyRedisServer.getPort());
```
upthewaterspout, line 102: I think there is a race condition where the region might be created but the function is not registered yet ... that might lead to an error executing the function. Maybe we need register the function earlier and set the bind address & port later, or maybe handle that sort of error when we execute the function?

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/SlotAdvisor.java
jdeppe-pivotal: I've pushed this down to the exact line it suppresses.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/GeodeRedisServer.java
jdeppe-pivotal: This should be handled now.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/SlotAdvisor.java
jdeppe-pivotal: But now it's back again because ... 🤷

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/SlotAdvisor.java
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    return memberBucketSlots;
+  }
+
+  private InternalDistributedMember getOrCreateMember(int bucketId) {
+    return dataRegion.getOrCreateNodeForBucketWrite(bucketId, null);
+  }
+
+  private void initializeBucketsIfNecessary() {
```
upthewaterspout, line 102: This function is no longer necessary since getOrCreateMember will create the buckets as needed.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/SlotAdvisor.java
```diff
@@ -0,0 +1,210 @@
...
+import org.apache.geode.redis.internal.cluster.RedisMemberInfo;
+import org.apache.geode.redis.internal.cluster.RedisMemberInfoRetrievalFunction;
+import org.apache.geode.redis.internal.data.RedisData;
+import org.apache.geode.redis.internal.data.RedisKey;
+
+public class SlotAdvisor {
+
+  private static final Logger logger = LogService.getLogger();
+  private static final int HOSTPORT_RETRIEVAL_ATTEMPTS = 10;
+  private static final int HOSTPORT_RETRIEVAL_INTERVAL = 6_000;
```
upthewaterspout, line 46: Why 6 seconds? Maybe something much smaller, in the 10s of milliseconds range? What is the average timeout for a redis client calling cluster slots I wonder?

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/SlotAdvisor.java
```diff
@@ -0,0 +1,210 @@
...
+    for (int i = 0; i < HOSTPORT_RETRIEVAL_ATTEMPTS; i++) {
+      response = getHostPort0(bucketId);
+      if (response != null) {
+        return response;
+      }
+
+      try {
+        Thread.sleep(HOSTPORT_RETRIEVAL_INTERVAL);
+      } catch (InterruptedException e) {
+        throw new RuntimeException(e);
```
upthewaterspout, line 125: This should probably just throw interrupted exception up to the caller. Wrapping it in a runtime exception means we'll might get some weird errors during a shutdown or something I think. Other option would be to ignore it, finish the loop and then reset the interrupt bit (Call Thread.interrupt()).

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/SlotAdvisor.java
```diff
@@ -0,0 +1,210 @@
...
+      return hostPorts.get(member);
+    }
+
+    List<RedisMemberInfo> memberInfos;
+    try {
+      ResultCollector<RedisMemberInfo, List<RedisMemberInfo>> resultCollector =
+          FunctionService.onRegion(dataRegion).execute(RedisMemberInfoRetrievalFunction.ID);
+      memberInfos = resultCollector.getResult();
+    } catch (Exception e) {
+      logger.warn("Unable to execute {}: {}", RedisMemberInfoRetrievalFunction.ID, e.getMessage());
```
upthewaterspout, line 148: Can distinguish between valid and invalid reasons to retry? If it's valid to retry, I'd rather we didn't log a warning. If it's not valid, I'd rather we through an exception will the full stack, rather than losing the stack trace here. If this fails with an NPE or something we're completely losing the line number here.

upthewaterspout: This might be a reason to register the function early and have the function itself return some sort of error response indicating the host & port aren't available.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/SlotAdvisor.java
```diff
@@ -0,0 +1,210 @@
...
+      return null;
+    }
+
+    hostPorts.clear();
+
+    for (RedisMemberInfo memberInfo : memberInfos) {
+      Pair<String, Integer> hostPort =
+          Pair.of(memberInfo.getHostAddress(), memberInfo.getRedisPort());
+      hostPorts.put(memberInfo.getMember(), hostPort);
```
upthewaterspout, line 157: Huh, should hostPorts just be a map of <String, RedisMemberInfo> ? Might be a bit clearer than Pair<Integer, String>

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/cluster/ClusterExecutor.java
upthewaterspout: There's probably something fun we could do with streams Collectors.groupingBy to take a list an group it by slots or group it by nodes, but anyway this probably works.

upthewaterspout (changes requested):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/SlotAdvisor.java
jdeppe-pivotal: Deleted!

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/SlotAdvisor.java
jdeppe-pivotal: I've changed it to 20 * 100 (2 seconds overall). I'm not sure if it's better to have the client timeout or to return incomplete slot information. I would lean towards the former since I'd expect a timeout to be a much more common situation that clients could handle.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/SlotAdvisor.java
jdeppe-pivotal: I think it's improved now. Exceptions will ultimately be handled in `ExecutionHandlerContext.exceptionCaught`.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/SlotAdvisor.java
jdeppe-pivotal: I'm registering the function early now. It will return a `null` response if it is called before it has been initialized with the host and port.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/SlotAdvisor.java
jdeppe-pivotal: Changed.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/GeodeRedisServer.java
```diff
@@ -79,9 +80,12 @@ public GeodeRedisServer(String bindAddress, int port, InternalCache cache) {
     redisStats = createStats(cache);
     StripedExecutor stripedExecutor = new SynchronizedStripedExecutor();
     regionProvider = new RegionProvider(cache);
+    CommandHelper commandHelper =
+        new CommandHelper(regionProvider.getDataRegion(), redisStats, stripedExecutor);
 
     CommandFunction.register(regionProvider.getDataRegion(), stripedExecutor, redisStats);
     RenameFunction.register(regionProvider.getDataRegion(), stripedExecutor, redisStats);
+    RedisMemberInfoRetrievalFunction infoFunction = RedisMemberInfoRetrievalFunction.register();
```
upthewaterspout, line 88: I think we need to make sure this is registered before the data region is created, so probably before the regionProvider initialization?

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/SlotAdvisor.java
```diff
@@ -0,0 +1,207 @@
...
+   * Cache of member ids to member IP address and redis listening port
+   */
+  private final Map<DistributedMember, RedisMemberInfo> memberInfos = new HashMap<>();
+  private final PartitionedRegion dataRegion;
+
+  SlotAdvisor(Region<RedisKey, RedisData> dataRegion) {
+    this.dataRegion = (PartitionedRegion) dataRegion;
+  }
+
+  public boolean isLocal(RedisKey key) {
```
upthewaterspout, line 57: I think this method is unused?

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/SlotAdvisor.java
upthewaterspout: Hmm, these still look like `Integer`. Maybe forgot to push something?

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/cluster/RedisMemberInfoRetrievalFunction.java
```diff
@@ -0,0 +1,81 @@
...
+ * once the redis server is running. If the function returns null it means that the function has
+ * not been initialized yet and callers, (from java), should retry.
+ */
+public class RedisMemberInfoRetrievalFunction implements InternalFunction<Void> {
+
+  public static final String ID = RedisMemberInfoRetrievalFunction.class.getName();
+  private static final long serialVersionUID = 2207969011229079993L;
+
+  private String hostAddress = null;
+  private Integer redisPort = null;
```
upthewaterspout, line 39: `Integer` again.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/SlotAdvisor.java
```diff
@@ -0,0 +1,207 @@
...
+    if (memberInfos.containsKey(member)) {
+      return memberInfos.get(member);
+    }
+
+    List<RedisMemberInfo> memberInfos;
+    try {
+      ResultCollector<RedisMemberInfo, List<RedisMemberInfo>> resultCollector =
+          FunctionService.onRegion(dataRegion).execute(RedisMemberInfoRetrievalFunction.ID);
+      memberInfos = resultCollector.getResult();
+    } catch (Exception e) {
```
upthewaterspout, line 143: Since you've changed the function to be registered earlier, I don't think it makes sense to catch and ignore function failures anymore.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/SlotAdvisor.java
upthewaterspout: Seems reasonable.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/SlotAdvisor.java
upthewaterspout: Is writing an error back to the client the right way to handle an interrupt though?

upthewaterspout (approved):

Bill (approved):
Approved!

PR #6470, opened 2021-05-12
GEODE-9269: Make the lock holding has the same order.

pivotal-eshu:
* When handling TXLockService, always trying to acquire destroy read lock
    before getting the synchronized batchLocks.
  * If destroy read lock is not acuried, thread will wait until the grantor
    is destoryed.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/distributed/internal/locks/DLockGrantor.java
+++ b/geode-core/src/main/java/org/apache/geode/distributed/internal/locks/DLockGrantor.java
@@ -489,38 +489,38 @@ public class DLockGrantor {
       cleanupSuspendState(request);
       return;
     }
-    if (!acquireDestroyReadLock(0)) {
+    if (acquireDestroyReadLock(0)) {
+      synchronized (batchLocks) { // assures serial processing
+        final boolean isTraceEnabled_DLS = logger.isTraceEnabled(LogMarker.DLS_VERBOSE);
+        if (isTraceEnabled_DLS) {
+          logger.trace(LogMarker.DLS_VERBOSE, "[DLockGrantor.handleLockBatch]");
+        }
+
+        try {
+          checkDestroyed();
+          if (isTraceEnabled_DLS) {
+            logger.trace(LogMarker.DLS_VERBOSE, "[DLockGrantor.handleLockBatch] request: {}",
+                request);
+          }
+
+          DLockBatch batch = (DLockBatch) request.getObjectName();
+          checkIfHostDeparted(batch.getOwner());
+          makeReservation(batch);
+          if (isTraceEnabled_DLS) {
+            logger.trace(LogMarker.DLS_VERBOSE, "[DLockGrantor.handleLockBatch] granting {}",
+                batch.getBatchId());
+          }
+          batchLocks.put(batch.getBatchId(), batch);
+          request.respondWithGrant(Long.MAX_VALUE);
+        } catch (CommitConflictException ex) {
+          request.respondWithTryLockFailed(ex.getMessage());
+        } finally {
+          releaseDestroyReadLock();
+        }
+      }
+    } else {
       waitUntilDestroyed();
       checkDestroyed();
-    }
-
-    synchronized (batchLocks) { // assures serial processing
-      final boolean isTraceEnabled_DLS = logger.isTraceEnabled(LogMarker.DLS_VERBOSE);
-      if (isTraceEnabled_DLS) {
-        logger.trace(LogMarker.DLS_VERBOSE, "[DLockGrantor.handleLockBatch]");
-      }
-
-      try {
-        checkDestroyed();
-        if (isTraceEnabled_DLS) {
-          logger.trace(LogMarker.DLS_VERBOSE, "[DLockGrantor.handleLockBatch] request: {}",
-              request);
-        }
-
-        DLockBatch batch = (DLockBatch) request.getObjectName();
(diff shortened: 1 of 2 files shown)


DonalEvans (changes requested):
A few typos and test cleanup suggestions. Also, the order in which we take the lock and synchronize is the same as the order in which we release the lock and exit the synchronized block. I'm wondering if it might be better to exit the synchronized block first, then release the lock.

geode-core/src/test/java/org/apache/geode/distributed/internal/locks/DLockGrantorTest.java
```diff
@@ -18,28 +18,40 @@
...
+  private final DLockRequestProcessor.DLockRequestMessage request = mock(
+      DLockRequestProcessor.DLockRequestMessage.class);
+  private final InternalDistributedMember owner = mock(InternalDistributedMember.class);
 
   @Before
   public void setup() {
     dLockService = mock(DLockService.class, RETURNS_DEEP_STUBS);
     DistributionManager distributionManager = mock(DistributionManager.class);
     when(dLockService.getDistributionManager()).thenReturn(distributionManager);
+    when(dLockService.getDLockLessorDepartureHandler())
```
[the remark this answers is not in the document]
pivotal-eshu: I thought the RETURNS_DEEP_STUBS should do the trick, but the test runs failed:
https://concourse.apachegeode-ci.info/teams/main/pipelines/apache-develop-pr/jobs/StressNewTestOpenJDK11/builds/1902

geode-core/src/main/java/org/apache/geode/distributed/internal/locks/DLockGrantor.java
```diff
@@ -639,21 +649,23 @@ public void updateLockBatch(Object batchId, DLockBatch newBatch) throws Interrup
...
+
+      synchronized (batchLocks) {
+        try {
+          checkDestroyed();
+          final DLockBatch oldBatch = batchLocks.get(batchId);
+          if (oldBatch != null) {
+            batchLocks.put(batchId, newBatch);
+          }
+        } finally {
+          releaseDestroyReadLock();
```
dschneider-pivotal, line 663: this should be done after the sync is released

geode-core/src/main/java/org/apache/geode/distributed/internal/locks/DLockGrantor.java
```diff
@@ -602,18 +609,20 @@ public DLockBatch getLockBatch(Object batchId) throws InterruptedException {
...
-        releaseDestroyReadLock();
+
+    waitWhileInitializing();
+    if (acquireDestroyReadLock(0)) {
+      synchronized (batchLocks) {
+        try {
+          checkDestroyed();
+          ret = batchLocks.get(batchId);
+        } finally {
+          releaseDestroyReadLock();
```
dschneider-pivotal, line 620: this should be done after the sync is released

geode-core/src/main/java/org/apache/geode/distributed/internal/locks/DLockGrantor.java
```diff
@@ -483,46 +484,51 @@ void handleLockBatch(DLockRequestMessage request) throws InterruptedException {
...
+          if (isTraceEnabled_DLS) {
+            logger.trace(LogMarker.DLS_VERBOSE, "[DLockGrantor.handleLockBatch] granting {}",
+                batch.getBatchId());
+          }
+          batchLocks.put(batch.getBatchId(), batch);
+          request.respondWithGrant(Long.MAX_VALUE);
+        } catch (CommitConflictException ex) {
+          request.respondWithTryLockFailed(ex.getMessage());
+        } finally {
+          releaseDestroyReadLock();
```
dschneider-pivotal, line 518: this should be done after the sync is released

geode-core/src/main/java/org/apache/geode/distributed/internal/locks/DLockGrantor.java
```diff
@@ -674,24 +687,30 @@ public void releaseLockBatch(Object batchId, InternalDistributedMember owner)
...
+    if (acquireDestroyReadLock(0)) {
+      synchronized (batchLocks) {
+        try {
+          checkDestroyed();
+          DLockBatch batch = batchLocks.remove(batchId);
+          if (batch != null) {
+            releaseReservation(batch);
+          }
+        } finally {
+          releaseDestroyReadLock();
```
dschneider-pivotal, line 700: this should be done after the sync is released

dschneider-pivotal (changes requested):
Since you are changing the order we do the readLock and the sync, you should also change the order they are released. Right after you have acquired the read lock you should have the try/finally that has the finally release that read lock. Inside that try block you would have the sync block. Doing it this way makes more sense and allows the code to be refactored (for example you could extra the sync blocks into their own methods). So unless you have a good reason for not have the releases match the acquires, I think you should change them to match.

geode-core/src/test/java/org/apache/geode/distributed/internal/locks/DLockGrantorTest.java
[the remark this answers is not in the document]
DonalEvans: Oh, weird. That must be something to do with how StressNewTest runs things in parallel.

DonalEvans (approved):

dschneider-pivotal (approved):

PR #6472, opened 2021-05-12
GEODE-8772: Assign TCP Server test ports in test JVM

demery-pivotal:
Make TcpServerProductVersionDUnitTest assign the locator port in the
test JVM rather than in child VMs.

An upcoming change to AvailablePortHelper will improve how it assigns
ports to better support tests running in parallel outside of docker.  A
child VM running an older version of Geode (as happens in this test)
will have an out-of-date implementation of AvailablePortHelper, and so
may assign a port that is in use by another test running in parallel.


geode-tcp-server/src/distributedTest/java/org/apache/geode/distributed/internal/tcpserver/TcpServerProductVersionDUnitTest.java
```diff
@@ -251,17 +251,15 @@ private TcpClient getTcpClient() {
         TcpSocketFactory.DEFAULT);
   }
 
-  private int createLocator(VM memberVM, boolean usingOldVersion) {
+  private int createLocator(VM memberVM) {
+    final int port = AvailablePortHelper.getRandomAvailableTCPPort();
```
Bill, line 255: One minor tweak might be to move this definition closer to the first point of use down on line 262.

Bill (approved):
Approved.

Made one optional suggestion below…

PR #6473, opened 2021-05-13
GEODE-9228: Fix batches could be sent with incomplete transactions

albertogpz:
when stopping gateway sender.

When stopping a gateway sender having group-transaction-events
set, there is a grace period in which events
received that belong to transactions are not dropped
if there are events in the queue belonging
to that transaction. This is done to make
sure that the queue does not contain incomplete
transactions.

Given that when the parallel gateway sender
creates a batch with events from the queue,
it removes them from the queue as it adds them
to the batch, it is possible that a transaction event
received by a gateway sender during
the stopping grace period
does not find events belonging to that transaction
in the queue because they have been added to
a current batch and therefore that event will
be dropped, causing that an incomplete transaction
will be finally stored in the queue/batch.
In order to solve the issue, when receiving a
transaction event during the stopping grace period,
the gateway sender must look for events belonging
to that transaction, not only in the queue
but also in the peekedEvents from the queue
(those events retrieved from the queue and
put in batches).

Also, in order to avoid flakyness of the distributed
test cases that test that batches with incomplete
transactions are not sent when stopping the
gateway sender, the number of retries
to get the events from the queue/batches
has been set to high number in the tests by means of
a new setter in GatewaySender which has
also been published as a gfsh and
cache.xml parameter (get-transaction-events-from-queue-retries).

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderEventProcessor.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderEventProcessor.java
@@ -136,14 +136,14 @@ public class ParallelGatewaySenderEventProcessor extends AbstractGatewaySenderEv
     boolean queuedEvent = false;
     try {
       if (getSender().beforeEnqueue(gatewayQueueEvent)) {
+        if (condition != null &&
+            !((ParallelGatewaySenderQueue) this.queue).hasEventsMatching(
+                (GatewaySenderEventImpl) gatewayQueueEvent,
+                condition)) {
+          return false;
+        }
         long start = getSender().getStatistics().startTime();
         try {
-          if (condition != null &&
-              !((ParallelGatewaySenderQueue) this.queue).hasEventsMatching(
-                  (GatewaySenderEventImpl) gatewayQueueEvent,
-                  condition)) {
-            return false;
-          }
           queuedEvent = this.queue.put(gatewayQueueEvent);
         } catch (InterruptedException e) {
           e.printStackTrace();
--- a/geode-core/src/main/java/org/apache/geode/cache/configuration/CacheConfig.java
+++ b/geode-core/src/main/java/org/apache/geode/cache/configuration/CacheConfig.java
@@ -162,6 +162,7 @@ import org.apache.geode.lang.Identifiable;
  *                 &lt;attribute name="dispatcher-threads" type="{http://www.w3.org/2001/XMLSchema}string" />
  *                 &lt;attribute name="order-policy" type="{http://www.w3.org/2001/XMLSchema}string" />
  *                 &lt;attribute name="group-transaction-events" type="{http://www.w3.org/2001/XMLSchema}boolean" />
+ *                 &lt;attribute name="get-transaction-events-from-queue-retries" type="{http://www.w3.org/2001/XMLSchema}string" />
  *               &lt;/restriction>
  *             &lt;/complexContent>
  *           &lt;/complexType>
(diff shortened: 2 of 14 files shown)


mhansonp:
Can you fix the LGTM issue before I re-review this?

I believe I have already been through this once, so I want to minimize re-reviews...

Thanks!

albertogpz:
> Can you fix the LGTM issue before I re-review this?
> 
> I believe I have already been through this once, so I want to minimize re-reviews...
> 
> Thanks!

Done! Sorry about it.

mhansonp:
> > Can you fix the LGTM issue before I re-review this?
> > I believe I have already been through this once, so I want to minimize re-reviews...
> > Thanks!
> 
> Done! Sorry about it.

No problem!

jinmeiliao:
Is the "get-transaction-events-from-queue-retries" added only to help with the flakiness of tests (based on the description of the PR)? the xml attributes is user facing, should avoid adding very obscure meaning attributes if we can help it.

albertogpz:
> Is the "get-transaction-events-from-queue-retries" added only to help with the flakiness of tests (based on the description of the PR)? the xml attributes is user facing, should avoid adding very obscure meaning attributes if we can help it.

Yes. At first, I only added it to the GatewaySender interface to be able to set it in test cases to avoid flakyness. Then I thought I should also add it to the xml file and gfsh interfaces.
But we could just leave it in the GatewaySender interface and as a System property, as it already was.

mhansonp (approved):

albertogpz:
Kind reminder, code owners reviews still pending.

jinmeiliao (approved):

geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderEventProcessor.java
```diff
@@ -136,14 +136,14 @@ protected boolean enqueueEvent(GatewayQueueEvent gatewayQueueEvent,
     boolean queuedEvent = false;
     try {
       if (getSender().beforeEnqueue(gatewayQueueEvent)) {
+        if (condition != null &&
+            !((ParallelGatewaySenderQueue) this.queue).hasEventsMatching(
+                (GatewaySenderEventImpl) gatewayQueueEvent,
+                condition)) {
+          return false;
+        }
```
kirklund, line 144: It might be best to some casts before the conditional just in order to improve readability:
```
ParallelGatewaySenderQueue parallelQueue = (ParallelGatewaySenderQueue) this.queue;
if (condition != null &&
    !(parallelQueue.hasEventsMatching((GatewaySenderEventImpl) gatewayQueueEvent, condition)) {
  return false;
}
```
We also need to consider updating interfaces like `RegionQueue` to avoid having to cast gatewayQueueEvent to GatewaySenderEventImpl. GatewaySenderEventImpl implements `InternalGatewayQueueEvent`. The interfaces start to lose value very fast if we let them become out-of-date. We should strive to use interfaces everywhere instead of concrete implementations.

Add: 
```
boolean hasEventsMatching(GatewaySenderEventImpl event, Predicate<InternalGatewayQueueEvent> condition);
``` 
to `RegionQueue` (if that makes sense... I'm not an expert in Wan code).

Next change that method signature to:
```
boolean hasEventsMatching(InternalGatewayQueueEvent event, Predicate<InternalGatewayQueueEvent> condition);
```
You can start out by casting the arg `event` within the implementation of that method, but it's good to push that change through all of the code that `event` travels to. If any of those methods need implementation specifics, then update the interface `InternalGatewayQueueEvent`.

Then that conditional block becomes:
```
ParallelGatewaySenderQueue parallelQueue = (ParallelGatewaySenderQueue) this.queue;
if (condition != null &&
    !(parallelQueue.hasEventsMatching(gatewayQueueEvent, condition)) {
  return false;
}
```
You could even add `hasNoEventsMatching(...)` to ParallelGatewaySenderQueue.

kirklund (changes requested):

albertogpz:
The original solution for GEODE-8971 is going to be reverted so this PR is not needed anymore given that it added the case for which a fix is provided in this PR.

PR #6478, opened 2021-05-13
GEODE-9229: Avoid possible duplicate events sent with group-transacti…

albertogpz:
…on-events

When group-transaction-events is set to true
the SerialGatewaySenderQueue class stores those events
peeked from the queue to complete transactions
for a batch inside the extraPeekedIds member. This is done
so that, when new events are peeked from the queue
by peekAhead() to be put in a batch, the
events in extraPeekedIds are skipped. Otherwise
those events will be sent several times.
When the remove method of SerialGatewaySenderQueue
is called, the peekedIds member and the head key
of the queue are updated and events from extraPeekedIds
are removed if necessary.
Given that the peekAhead() method first gets the
currentKey (that depends on the peekedIds and the headkey)
and then checks if the event to be removed is in extraPeekedIds,
it could be that in between those two calls,
remove() is called from another thread, causing that
peekAhead() returns an event that
belonged to extraPeekedIds when the currentKey was
obtained but was deleted from extraPeekedIds by remove()
before it was checked that the event belonged to
extraPeekedIds.
This provokes that an event already peeked is returned
again by peekAhead() leading to duplicate events
sent by the gateway sender.

The solution proposed to solve this problem is to make
synchronized the peekAhead() function when group-transaction-events
is set given that remove() is already synchronized. This will
prevent the race condition described above.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/wan/serial/SerialGatewaySenderQueue.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/wan/serial/SerialGatewaySenderQueue.java
@@ -787,6 +787,18 @@ public class SerialGatewaySenderQueue implements RegionQueue {
 
   @VisibleForTesting
   public KeyAndEventPair peekAhead() throws CacheException {
+    // When group-transaction-events is true, peekAhead must
+    // be synchronized in order to get a consistent view
+    // of extraPeekedIds and currentKey as provided by remove()
+    if (mustGroupTransactionEvents()) {
+      synchronized (this) {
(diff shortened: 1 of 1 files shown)


albertogpz:
Kind reminder, code owners reviews still pending.

albertogpz:
The original solution for GEODE-8971 is reverted so this PR is not needed anymore given that it added the case for which a fix is provided in this PR.
Also, the solution proposed in this PR was not valid. Making peekAhead() synchronized would not fix the possible issue because what was required was that remove() was synchronized and that already was so.

PR #6487, opened 2021-05-17
GEODE-9222: Remove ByteArrayWrapper from RedisSet

DonalEvans:
- Replace uses of ByteArrayWrapper with byte[] in RedisSet,
 NullRedisSet and other Set-related classes
 - Replace internal Set in RedisSet with fastutil
 ObjectOpenCustomHashSet
 - Replace the membersAddAll and membersRemoveAll methods with looped
 calls to membersAdd and membersRemove, as ObjectOpenCustomHashSet does
 not implement addAll and removeAll, which causes errors when comparing
 the contents of Sets
 - Replace uses of implementations with interfaces where possible
 - These changes break the sizing tests, but these tests have been
 determined to be inconsistent due to the current implementation of
 sizeable in RedisSet. GEODE-9279 will address the issues with sizeable
 in a future commit

### For all changes:

### Note:

### DIFF
--- a/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/NullRedisSet.java
+++ b/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/NullRedisSet.java
@@ -22,7 +22,11 @@ import static java.util.Collections.emptySet;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Set;
+
+import it.unimi.dsi.fastutil.bytes.ByteArrays;
+import it.unimi.dsi.fastutil.objects.ObjectOpenCustomHashSet;
 
 import org.apache.geode.annotations.VisibleForTesting;
 import org.apache.geode.cache.Region;
@@ -41,18 +45,17 @@ class NullRedisSet extends RedisSet {
   }
 
   @Override
-  Collection<ByteArrayWrapper> spop(Region<RedisKey, RedisData> region,
-      RedisKey key, int popCount) {
+  Collection<byte[]> spop(Region<RedisKey, RedisData> region, RedisKey key, int popCount) {
     return emptyList();
   }
 
   @Override
-  Collection<ByteArrayWrapper> srandmember(int count) {
+  Collection<byte[]> srandmember(int count) {
     return emptyList();
   }
 
   @Override
-  public boolean sismember(ByteArrayWrapper member) {
+  public boolean sismember(byte[] member) {
     return false;
   }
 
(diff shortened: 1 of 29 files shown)


ringles (changes requested):
One minor suggestion, otherwise looks good.

upthewaterspout:
I don't know too much about the TCL test, but I thought this was a refactoring that shouldn't have affected them. What are the changes to the TCL tests about?

sabbey37:
> I don't know too much about the TCL test, but I thought this was a refactoring that shouldn't have affected them. What are the changes to the TCL tests about?

The Native Redis test for `EXPIRE - write on expire should work` were utilizing `lpush`.  We did not have that command/the list data structure implemented, so at some point they were changed to do a `sadd`.  The tests also checks the result of `smembers`.  Since sets don't have a fixed order, it failed (but seems like this happened very infrequently since it was only just caught).  The test was changed to use a `set` on a string (it didn't seem to matter which data set was used, since the test just verified that we could continue to edit the value after an expiration was set).  That was all that was changed in the tcl tests, but the whole patch file was replaced (including all the same changes as before) vs. trying to edit it directly.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/NullRedisSet.java
```diff
@@ -62,21 +65,19 @@ public int scard() {
...
+  long srem(List<byte[]> membersToRemove, Region<RedisKey, RedisData> region, RedisKey key) {
     return 0;
   }
 
   @Override
   @VisibleForTesting
-  Set<ByteArrayWrapper> smembers() {
+  public Set<byte[]> smembers() {
     // some callers want to be able to modify the set returned
     return new HashSet<>();
```
upthewaterspout, line 82: Should this be an Object2Object... set instead of a HashMap<>? If the callers modify this returned set, they won't be getting the right hashing behavior.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSet.java
```diff
@@ -340,7 +306,7 @@ long srem(ArrayList<ByteArrayWrapper> membersToRemove, Region<RedisKey, RedisDat
    * @return a set containing all the members in this set
    */
   @VisibleForTesting
-  Set<ByteArrayWrapper> smembers() {
+  public Set<byte[]> smembers() {
     return new HashSet<>(members);
```
upthewaterspout, line 310: This definitely needs to be an Object2ObjectOpenCustomHashMap, not a HashMap here.

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisSetTest.java
```diff
@@ -193,9 +197,10 @@ public void should_calculateSize_equalToROS_withNoMembers() {
   }
 
   @Test
+  @Ignore("Sizing tests are known to be flaky/incorrect and will be fixed as part of GEODE-9279")
   public void should_calculateSize_equalToROS_withSingleMember() {
-    HashSet<ByteArrayWrapper> members = new HashSet<>();
-    members.add(new ByteArrayWrapper("value".getBytes()));
+    Set<byte[]> members = new HashSet<>();
```
upthewaterspout, line 202: I guess these tests are technically correct, but it still seems weird to use a HashSet<byte[]>() here. That will use the Object.hashCode() of the byte[] array. Maybe make this a ArrayList<> instead to make it clear we don't really care about set semantics on `members` here? Or use Object2ObjectOpenCustomHashMap instead.

upthewaterspout (changes requested):
Overall it looks really good but I had a few comments about lingering usage of HashSet<>, inline.

upthewaterspout:
> > I don't know too much about the TCL test, but I thought this was a refactoring that shouldn't have affected them. What are the changes to the TCL tests about?
> 
> The Native Redis test for `EXPIRE - write on expire should work` was utilizing `lpush`. We did not have that command/the list data structure implemented, so at some point it was changed to do a `sadd`. The test also checks the result of `smembers`. Since sets don't have a fixed order, it failed (but seems like this happened very infrequently since it was only just caught). The test was changed to use a `set` on a string (it didn't seem to matter which data structure was used, since the test just verified that we could continue to edit the value after an expiration was set). That was all that was changed in the tcl tests, but the whole patch file was replaced (including all the same changes as before) vs. trying to edit it directly.

Ok, makes sense.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSet.java
upthewaterspout: I would actually suggest changing this to `Collections.unmodifiableSet(members)` here. This is an expensive operation, so it would be better if we only make a copy of this set where we actually need to modify it. Or only currently supported command, SMEMBERS, doesn't need this extra copying.

DonalEvans: The commit prior to the current head returned an `ObjectOpenCustomHashSet` here (and in NullRedisSet) but this caused serialization issues in DUnit tests as that class wasn't allowed by the serialization filter. After trying to find an alternate solution, I've settled on extending `ObjectOpenCustomHashSet` to allow adding it to the appropriate sanctioned-serializables file, and to allow proper support of `addAll()` and `removeAll()`, since the current implementation of `ObjectOpenCustomHashSet` doesn't employ the hashing strategy set in the constructor for those.

DonalEvans: Correction, the hashing strategy is correctly employed in the current implementation, it's just the serialization issue that's present. Nevertheless, extending the class allows serialization and accurate calculation of size down the line when we revisit sizeable for RedisSet, so it's still the best choice, I think.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/NullRedisSet.java
```diff
@@ -122,29 +117,29 @@ private int doSetOpWhileLocked(SetOp setOp, CommandHelper helper,
...
-      RedisKey destination) {
-    Set<ByteArrayWrapper> result = null;
+  private Set<byte[]> computeSetOp(SetOp setOp, List<Set<byte[]>> nonDestinationSets,
+      CommandHelper helper, RedisKey destination) {
+    ObjectOpenCustomHashSet<byte[]> result = null;
     if (nonDestinationSets.isEmpty()) {
       return emptySet();
     }
-    for (Set<ByteArrayWrapper> set : nonDestinationSets) {
+    for (Set<byte[]> set : nonDestinationSets) {
```
upthewaterspout, line 126: This didn't start with your changes, but these set operations may happen on very large sets - we will probably want to eliminate as much set copying as possible before making these things supported commands.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSetCommandsFunctionExecutor.java
```diff
@@ -42,42 +41,41 @@ private RedisSet getRedisSet(RedisKey key, boolean updateStats) {
...
+  public long srem(RedisKey key, List<byte[]> membersToRemove) {
+    return stripedExecute(key, () -> getRedisSet(key, false)
+        .srem(membersToRemove, getRegion(), key));
   }
 
   @Override
-  public Set<ByteArrayWrapper> smembers(RedisKey key) {
-    return stripedExecute(key, () -> getRedisSet(key, true).smembers());
+  public Set<byte[]> smembers(RedisKey key) {
+    return stripedExecute(key, () -> new HashSet<>(getRedisSet(key, true).smembers()));
```
upthewaterspout, line 73: More set copying here.

upthewaterspout (commented):
Removing my blocking review since I'll be on vacation next week.

DonalEvans:
Regarding the set copying that Dan mentioned, I agree that it's something we want to avoid and get rid of, but the sets used in the diff and intersection operations need to be a class that can handle comparisons between byte arrays in the way we want. However, returning a `ObjectOpenCustomHashSet` from the lambdas in `RedisSetCommandsFunctionExecutor.internalsmembers()` or `RedisSetCommandsFunctionExecutor.smembers()` results in serialization filter errors, so as a compromise, this set copying behaviour is used to allow a `HashSet` to be returned from the lambda and then turned back into a `ObjectOpenCustomHashSet` when needed for diff and intersection. Since these commands aren't currently supported, this bevaiour should be revisited when they're fully implemented and supported.

nonbinaryprogrammer (approved):
Looks solid. I think that @upthewaterspout's comments are valuable, but given the scope of this ticket, I think it would be better to add something to Jira to track that work.

DonalEvans:
> Looks solid. I think that @upthewaterspout's comments are valuable, but given the scope of this ticket, I think it would be better to add something to Jira to track that work.

Agreed. I'll create a ticket.

jdeppe-pivotal (approved):

ringles (approved):
Seems solid at this point. Thanks!

DonalEvans:
Force pushed to rebase on develop, will merge once pre-checkin is finished.

PR #6493, opened 2021-05-17
GEODE-9202: implement PUBSUB CHANNELS subcommand

jhutchison:
-add PUBSUB to supported commands in RedisCommandType
-create subcommands executor in PubSub dir
-implement mehtod for finding channel names in subscriptions class
-implement pattern matching for finding channel names
-add methods to pubsub impl and PubSub interface for returning channel names
-add to PUBSUB to Supported Comands list in SupportedCommandsJunitTest

### For all changes:

### Note:

### DIFF
--- a/geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/pubsub/AbstractSubCommandsIntegrationTest.java
+++ b/geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/pubsub/AbstractSubCommandsIntegrationTest.java
@@ -1,0 +1,199 @@
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ *
+ */
+
+package org.apache.geode.redis.internal.executor.pubsub;
+
+import static org.apache.geode.redis.RedisCommandArgumentsTestHelper.assertAtLeastNArgs;
+import static org.apache.geode.redis.RedisCommandArgumentsTestHelper.assertAtMostNArgsForSubCommand;
+import static org.apache.geode.redis.internal.RedisConstants.ERROR_UNKNOWN_PUBSUB_SUBCOMMAND;
+import static org.apache.geode.redis.internal.executor.pubsub.AbstractPubSubIntegrationTest.JEDIS_TIMEOUT;
+import static org.apache.geode.util.internal.UncheckedUtils.uncheckedCast;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static redis.clients.jedis.Protocol.PUBSUB_CHANNELS;
+
+import java.net.SocketException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.Callable;
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import redis.clients.jedis.Jedis;
+import redis.clients.jedis.Protocol;
+
+import org.apache.geode.redis.RedisIntegrationTest;
+import org.apache.geode.redis.internal.netty.Coder;
+import org.apache.geode.redis.mocks.MockSubscriber;
+import org.apache.geode.test.awaitility.GeodeAwaitility;
+
+public abstract class AbstractSubCommandsIntegrationTest implements RedisIntegrationTest {
+  private Jedis subscriber;
+  private Jedis introspector;
+  private MockSubscriber mockSubscriber;
+
+  @Before
+  public void setup() {
+    mockSubscriber = new MockSubscriber();
+    subscriber = new Jedis("localhost", getPort());
+    introspector = new Jedis("localhost", getPort());
+  }
+
+  @After
+  public void teardown() {
+    if (mockSubscriber.getSubscribedChannels() > 0) {
+      mockSubscriber.unsubscribe();
+    }
+    waitFor(() -> mockSubscriber.getSubscribedChannels() == 0);
+  }
+
+  @Test
+  public void pubsub_shouldError_givenTooFewArguments() {
+    assertAtLeastNArgs(introspector, Protocol.Command.PUBSUB, 1);
+  }
+
+  @Test
+  public void channels_shouldError_givenTooManyArguments() {
+    assertAtMostNArgsForSubCommand(introspector,
+        Protocol.Command.PUBSUB,
+        Coder.stringToBytes(PUBSUB_CHANNELS),
+        1);
+  }
+
+  @Test
+  public void pubsub_shouldReturnError_givenUnknownSubcommand() {
+    String expected = String.format(ERROR_UNKNOWN_PUBSUB_SUBCOMMAND, "nonesuch");
+
+    assertThatThrownBy(() -> introspector.sendCommand(Protocol.Command.PUBSUB, "nonesuch"))
+        .hasMessageContaining(expected);
+  }
+
+  @Test
+  public void channels_shouldReturnListOfAllChannels_withActiveChannelSubscribers_whenCalledWithoutPattern() {
+    List<byte[]> expectedChannels = new ArrayList<>();
+    expectedChannels.add(Coder.stringToBytes("foo"));
+    expectedChannels.add(Coder.stringToBytes("bar"));
+    Runnable runnable =
+        () -> subscriber.subscribe(mockSubscriber, "foo", "bar");
+    Thread subscriberThread = new Thread(runnable);
+
+    subscriberThread.start();
+    waitFor(() -> mockSubscriber.getSubscribedChannels() == 2);
+
+    List<byte[]> result =
+        uncheckedCast(introspector.sendCommand(Protocol.Command.PUBSUB, PUBSUB_CHANNELS));
+
+    assertThat(result).containsExactlyInAnyOrderElementsOf(expectedChannels);
+  }
+
+  @Test
+  public void channels_shouldNeverReturnPsubscribedChannels_givenNoActiveChannelSubscribers() {
+
+    Runnable runnable = () -> subscriber.psubscribe(mockSubscriber, "f*");
+    Thread patternSubscriberThread = new Thread(runnable);
+
+    patternSubscriberThread.start();
+    waitFor(() -> mockSubscriber.getSubscribedChannels() == 1);
+    List<String> result = introspector.pubsubChannels("f*");
+
+    assertThat(result).isEmpty();
+
+    mockSubscriber.punsubscribe();
+  }
+
+
+  @Test
+  public void channels_shouldReturnListOfMatchingChannels_withActiveChannelSubscribers_whenCalledWithPattern() {
+
+    List<String> expectedChannels = new ArrayList<>();
+    expectedChannels.add("foo");
+
+    Runnable runnable =
+        () -> subscriber.subscribe(mockSubscriber, "foo", "bar");
+
+    Thread subscriberThread = new Thread(runnable);
+    subscriberThread.start();
+    waitFor(() -> mockSubscriber.getSubscribedChannels() == 2);
+    List<String> result = introspector.pubsubChannels("fo*");
+
+    assertThat(result).containsExactlyInAnyOrderElementsOf(expectedChannels);
+  }
+
+  @Test
+  public void channels_should_returnEmptyArray_givenPatternWithNoMatches() {
+    List<String> result = introspector.pubsubChannels("fo*");
+
+    assertThat(result).isEmpty();
+  }
+
+  @Test
+  public void channels_shouldOnlyReturnChannelsWithActiveSubscribers() {
+    List<byte[]> expectedChannels = new ArrayList<>();
+    expectedChannels.add(Coder.stringToBytes("bar"));
+    Runnable runnable = () -> subscriber.subscribe(mockSubscriber, "foo", "bar");
+    Thread subscriberThread = new Thread(runnable);
+
+    subscriberThread.start();
+    waitFor(() -> mockSubscriber.getSubscribedChannels() == 2);
+    mockSubscriber.unsubscribe("foo");
+    waitFor(() -> mockSubscriber.getSubscribedChannels() == 1);
+    List<byte[]> result =
+        uncheckedCast(introspector.sendCommand(Protocol.Command.PUBSUB, PUBSUB_CHANNELS));
+
+    assertThat(result).containsExactlyInAnyOrderElementsOf(expectedChannels);
+  }
+
+  @Test
+  public void channels_shouldNotReturnDuplicates_givenMultipleSubscribersToSameChannel_whenCalledWithoutPattern() {
+
+    Jedis subscriber2 = new Jedis("localhost", getPort(), JEDIS_TIMEOUT);
+
+    MockSubscriber mockSubscriber2 = new MockSubscriber();
+    List<byte[]> expectedChannels = new ArrayList<>();
+    expectedChannels.add(Coder.stringToBytes("foo"));
+
+    Runnable runnable = () -> subscriber.subscribe(mockSubscriber, "foo");
+    Thread subscriber1Thread = new Thread(runnable);
+    subscriber1Thread.start();
+
+    Runnable runnable2 = () -> subscriber2.subscribe(mockSubscriber2, "foo");
+    Thread subscriber2Thread = new Thread(runnable2);
+    subscriber2Thread.start();
+
+    waitFor(() -> (mockSubscriber.getSubscribedChannels() == 1)
+        && (mockSubscriber2.getSubscribedChannels() == 1));
+
+    List<byte[]> result =
+        uncheckedCast(introspector
+            .sendCommand(Protocol.Command.PUBSUB, PUBSUB_CHANNELS));
+
+    assertThat(result).containsExactlyInAnyOrderElementsOf(expectedChannels);
+    assertThat(result.size()).isEqualTo(1);
+
+    mockSubscriber2.unsubscribe();
+    waitFor(() -> mockSubscriber2.getSubscribedChannels() == 0);
+
+    subscriber2.close();
+  }
+
+  private void waitFor(Callable<Boolean> booleanCallable) {
+    GeodeAwaitility.await()
+        .ignoreExceptionsInstanceOf(SocketException.class)
+        .until(booleanCallable);
+  }
+}
--- a/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/RedisConstants.java
+++ b/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/RedisConstants.java
@@ -59,4 +59,7 @@ public class RedisConstants {
       "Unknown subcommand or wrong number of arguments for '%s'. Try CLUSTER HELP.";
   public static final String ERROR_INVALID_ZADD_OPTION_NX_XX =
       "XX and NX options at the same time are not compatible";
+
+  public static final String ERROR_UNKNOWN_PUBSUB_SUBCOMMAND =
+      "Unknown subcommand or wrong number of arguments for '%s'";
 }
--- a/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/pubsub/Subscriptions.java
+++ b/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/pubsub/Subscriptions.java
@@ -16,13 +16,19 @@
 
 package org.apache.geode.redis.internal.pubsub;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import it.unimi.dsi.fastutil.bytes.ByteArrays;
+import it.unimi.dsi.fastutil.objects.ObjectOpenCustomHashSet;
 
 import org.apache.geode.annotations.VisibleForTesting;
 import org.apache.geode.redis.internal.executor.GlobPattern;
 import org.apache.geode.redis.internal.netty.Client;
+import org.apache.geode.redis.internal.netty.Coder;
 import org.apache.geode.redis.internal.netty.ExecutionHandlerContext;
 
 /**
@@ -65,6 +71,33 @@ public class Subscriptions {
     return subscriptions.stream()
         .filter(subscription -> subscription.matches(channelOrPattern))
         .collect(Collectors.toList());
+  }
+
+  public List<byte[]> findChannelNames() {
+
+    ObjectOpenCustomHashSet<byte[]> hashSet =
+        new ObjectOpenCustomHashSet<>(ByteArrays.HASH_STRATEGY);
+
+    findChannels()
+        .forEach(channel -> hashSet.add(channel.getSubscriptionName()));
+
+    return new ArrayList<>(hashSet);
+  }
+
+  public List<byte[]> findChannelNames(byte[] pattern) {
+
+    GlobPattern globPattern = new GlobPattern(Coder.bytesToString(pattern));
+
+    return findChannelNames()
+        .stream()
+        .filter(name -> globPattern.matches(Coder.bytesToString(name)))
+        .collect(Collectors.toList());
+  }
+
+  private Stream<ChannelSubscription> findChannels() {
+    return subscriptions.stream()
+        .filter(subscription -> subscription instanceof ChannelSubscription)
+        .map(subscription -> (ChannelSubscription) subscription);
   }
 
   /**
--- a/geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/pubsub/SubscriptionsJUnitTest.java
+++ b/geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/pubsub/SubscriptionsJUnitTest.java
@@ -20,12 +20,15 @@ import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
+import java.util.List;
+
 import io.netty.channel.Channel;
 import io.netty.channel.ChannelFuture;
 import org.junit.Test;
 
 import org.apache.geode.redis.internal.executor.GlobPattern;
 import org.apache.geode.redis.internal.netty.Client;
+import org.apache.geode.redis.internal.netty.Coder;
 import org.apache.geode.redis.internal.netty.ExecutionHandlerContext;
 
 public class SubscriptionsJUnitTest {
@@ -41,10 +44,11 @@ public class SubscriptionsJUnitTest {
     ExecutionHandlerContext context = mock(ExecutionHandlerContext.class);
 
     subscriptions
-        .add(new ChannelSubscription(client, "subscriptions".getBytes(), context, subscriptions));
+        .add(new ChannelSubscription(client, Coder.stringToBytes("subscriptions"), context,
+            subscriptions));
 
-    assertThat(subscriptions.exists("subscriptions".getBytes(), client)).isTrue();
-    assertThat(subscriptions.exists("unknown".getBytes(), client)).isFalse();
+    assertThat(subscriptions.exists(Coder.stringToBytes("subscriptions"), client)).isTrue();
+    assertThat(subscriptions.exists(Coder.stringToBytes("unknown"), client)).isFalse();
   }
 
   @Test
@@ -78,7 +82,8 @@ public class SubscriptionsJUnitTest {
     GlobPattern globPattern2 = new GlobPattern("subscriptions");
 
     subscriptions
-        .add(new ChannelSubscription(client, "subscriptions".getBytes(), context, subscriptions));
+        .add(new ChannelSubscription(client, Coder.stringToBytes("subscriptions"), context,
+            subscriptions));
 
     assertThat(subscriptions.exists(globPattern1, client)).isFalse();
     assertThat(subscriptions.exists(globPattern2, client)).isFalse();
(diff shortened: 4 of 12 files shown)


upthewaterspout (commented):
Overall looks good! I few minor comments inline.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/pubsub/SubCommandExecutor.java
```diff
@@ -0,0 +1,54 @@
...
+
+import java.util.Arrays;
+import java.util.List;
+
+import org.apache.geode.redis.internal.executor.Executor;
+import org.apache.geode.redis.internal.executor.RedisResponse;
+import org.apache.geode.redis.internal.netty.Command;
+import org.apache.geode.redis.internal.netty.ExecutionHandlerContext;
+
+public class SubCommandExecutor implements Executor {
```
sabbey37, line 28: Since there are a lot of commands in Redis that have sub commands, I think it would make more sense to follow the pattern of our other executors and name this after the primary command, so `PubsubExecutor` (or `PubSubExecutor`?) (also, all related test classes would need to be renamed).  Check out `ClusterExecutor`, which is named after the primary command, but has a case switch for sub commands.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/pubsub/SubCommandExecutor.java
```diff
@@ -0,0 +1,54 @@
...
+
+  static final String UNKNOWN_SUBCOMMAND_ERROR_BASE =
+      "Unknown subcommand or wrong number of arguments for '%s'";
```
sabbey37, line 31: I'm guessing we'll add the `Try PUBSUB HELP.` message to the end of this once `PUBSUB HELP` is implemented?  I think we should put this message in the `RedisConstants` file where other similar error messages live and rename it to follow the same pattern as the others, maybe `ERROR_UNKNOWN_PUBSUB_SUBCOMMAND` since it will eventually be specific to PUBSUB.... we could actually refactor those subcommand error messages to all use the same base and insert the relevant primary command, so something like:
```
ERROR_UNKNOWN_SUBCOMMAND =
      "Unknown subcommand or wrong number of arguments for '%1$s'. Try %2$s HELP."

File: geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/pubsub/SubscriptionsJUnitTest.java [hunk 3]
```diff
@@ -195,4 +197,70 @@ public void removeByClientAndPattern() {
...
+    byte[] pattern = "b*".getBytes();
+
+    Channel channel = mock(Channel.class);
+    when(channel.closeFuture()).thenReturn(mock(ChannelFuture.class));
+    Client client = new Client(channel);
+    ExecutionHandlerContext context = mock(ExecutionHandlerContext.class);
+
+    subject.add(new ChannelSubscription(client, "foo".getBytes(), context, subject));
+    subject.add(new ChannelSubscription(client, "bar".getBytes(), context, subject));
+    subject.add(new ChannelSubscription(client, "barbarella".getBytes(), context, subject));
```
sabbey37, line 236: What about `BarbaraAnn`? 😜

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/pubsub/AbstractSubCommandsIntegrationTest.java
```diff
@@ -0,0 +1,156 @@
...
+    expectedChannels.add("foo".getBytes());
+    Runnable runnable =
+        () -> subscriber.subscribe(mockSubscriber, "foo", "bar");
+    Thread subscriberThread = new Thread(runnable);
+
+    subscriberThread.start();
+    waitFor(() -> mockSubscriber.getSubscribedChannels() == 2);
+    ArrayList<byte[]> result =
+        (ArrayList<byte[]>) introspector
+            .sendCommand(Protocol.Command.PUBSUB, "channels", "fo*");
```
sabbey37, line 103: For this test as well as the test below it we could use `introspector.pubsubChannels("fo*")`

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/pubsub/AbstractSubCommandsIntegrationTest.java
```diff
@@ -0,0 +1,156 @@
...
+  @SuppressWarnings("unchecked")
+  public void channels_shouldReturnListOfAllChannels_whenCalledWithoutPattern() {
+    ArrayList<byte[]> expectedChannels = new ArrayList<>();
+    expectedChannels.add("foo".getBytes());
+    expectedChannels.add("bar".getBytes());
+    Runnable runnable =
+        () -> subscriber.subscribe(mockSubscriber, "foo", "bar");
+    Thread subscriberThread = new Thread(runnable);
+
+    subscriberThread.start();
```
sabbey37, line 82: In some of our other tests with subscriberThreads, we have a finally block that cleans up by unsubscribing and waiting for the thread to no longer be alive.  Do we need to do that in these tests?

sabbey37 (changes requested):

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/pubsub/AbstractSubCommandsIntegrationTest.java
```diff
@@ -0,0 +1,156 @@
...
+    waitFor(() -> mockSubscriber.getSubscribedChannels() == 2);
+    ArrayList<byte[]> result = (ArrayList<byte[]>) introspector
+        .sendCommand(Protocol.Command.PUBSUB, "channels");
```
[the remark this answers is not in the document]
sabbey37: It doesn't seem like Jedis has the no args option for pubsubChannels yet.... you have to pass it a pattern, but we could use it in the tests where we pass a pattern.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/pubsub/SubCommandExecutor.java
jdeppe-pivotal: Right - it's unfortunate we already have those `PubSub*` classes. Perhaps even naming these with `PubsubSub*`?

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/pubsub/AbstractSubCommandsIntegrationTest.java
```diff
@@ -0,0 +1,156 @@
...
+    waitFor(() -> mockSubscriber.getSubscribedChannels() == 0);
+  }
+
+  @Test
+  public void Pubsub_shouldError_givenTooFewArguments() {
+    assertAtLeastNArgs(introspector, Protocol.Command.PUBSUB, 1);
+  }
+
+  @Test
+  @SuppressWarnings("unchecked")
```
sabbey37, line 64: We don't need this `SuppressWarnings`

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/pubsub/AbstractSubCommandsIntegrationTest.java
```diff
@@ -0,0 +1,156 @@
...
+  public void channels_shouldReturnListOfAllChannels_whenCalledWithoutPattern() {
+    ArrayList<byte[]> expectedChannels = new ArrayList<>();
+    expectedChannels.add("foo".getBytes());
+    expectedChannels.add("bar".getBytes());
+    Runnable runnable =
+        () -> subscriber.subscribe(mockSubscriber, "foo", "bar");
+    Thread subscriberThread = new Thread(runnable);
+
```
jdeppe-pivotal, line 81: Could you add a second subscriber that subscribes to exactly the same channels to ensure that duplicates are not returned by `PUBSUB CHANNELS`?

jdeppe-pivotal (changes requested):

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/pubsub/AbstractSubCommandsIntegrationTest.java
```diff
@@ -0,0 +1,156 @@
...
+    subscriber = new Jedis("localhost", getPort());
+    introspector = new Jedis("localhost", getPort());
+  }
+
+  @After
+  public void teardown() {
+    if (mockSubscriber.getSubscribedChannels() > 0) {
+      mockSubscriber.unsubscribe();
+    }
+    waitFor(() -> mockSubscriber.getSubscribedChannels() == 0);
```
sabbey37, line 55: Since this is generic, we should also check for pattern subscriptions and `punsubscribe` if needed.  I wonder if we should also wait for the subscriberThread to no longer be alive....

```
waitFor(() -> !subscriberThread.isAlive());
```

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/pubsub/AbstractSubCommandsIntegrationTest.java
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  @Test
+  public void Pubsub_shouldError_givenTooFewArguments() {
+    assertAtLeastNArgs(introspector, Protocol.Command.PUBSUB, 1);
+  }
```
sabbey37, line 61: We should also add a test for too many arguments for `PUBSUB CHANNELS`.  It looks like `CHANNELS` only accepts one argument.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/pubsub/AbstractSubCommandsIntegrationTest.java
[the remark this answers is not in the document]
jhutchison: I changed the ones that pass a parameter.  Too bad the pubSubChannels method doesn't just work as expected if you pass in null...

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/pubsub/SubCommandExecutor.java
```diff
@@ -0,0 +1,54 @@
...
+
+  @Override
+  public RedisResponse executeCommand(Command command, ExecutionHandlerContext context) {
+
+    List<byte[]> response;
+
+    byte[] subCommand = command.getProcessedCommand().get(1);
+
+    if (!Arrays.equals(subCommand, "channels".getBytes())) {
```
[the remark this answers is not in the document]
jhutchison: thanks

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/pubsub/Subscriptions.java
```diff
@@ -67,6 +67,34 @@ boolean exists(Object channelOrPattern, Client client) {
...
+        .filter(subscription -> globPattern.matches(new String(subscription.getSubscriptionName())))
+        .map(ChannelSubscription::getSubscriptionName)
+        .distinct()
+        .collect(Collectors.toList());
+  }
+
+  private List<Subscription> findChannels() {
+    return subscriptions.stream()
+        .filter(subscription -> subscription instanceof ChannelSubscription)
+        .collect(Collectors.toList());
```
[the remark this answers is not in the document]
jhutchison: returning the stream makes sense, thanks. I like the idea of the return type of the findChannels method returning a Stream<ChannelSubscription>,  the downside is that the compiler makes me add this line to do that : `.map(subscription -> (ChannelSubscription) subscription);` 

so the whole thing ends up reading:  `  return subscriptions.stream()
        .filter(subscription -> subscription instanceof ChannelSubscription)
        .map(subscription -> (ChannelSubscription) subscription);`

which then seems a little confusing... not sure which is better?  If you don't have a strong opinion, I would lean towards returning the interface and not casting after filtering in the stream...

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/pubsub/Subscriptions.java
```diff
@@ -67,6 +67,34 @@ boolean exists(Object channelOrPattern, Client client) {
...
+  }
+
+  public List<byte[]> findChannelNames(byte[] pattern) {
+    GlobPattern globPattern = new GlobPattern(new String(pattern));
+
+    return findChannels()
+        .stream()
+        .map(subscription -> (ChannelSubscription) subscription)
+        .filter(subscription -> globPattern.matches(new String(subscription.getSubscriptionName())))
```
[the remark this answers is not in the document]
jhutchison: thanks

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/pubsub/Subscriptions.java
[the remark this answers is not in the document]
jhutchison: Ahh-- I do see that this allows be to not have to cast it in the calling method though... I take it back, I like it better returning the channelSubscription...

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/pubsub/Subscriptions.java
```diff
@@ -67,6 +67,34 @@ boolean exists(Object channelOrPattern, Client client) {
...
+        .map(ChannelSubscription::getSubscriptionName)
+        .distinct()
+        .collect(Collectors.toList());
+  }
```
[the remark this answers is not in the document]
jhutchison: thanks.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/pubsub/SubCommandExecutor.java
jhutchison: thanks-  for some reason I didn't find that file when I looked for it-  thought maybe it had been removed.  

as far as the "help" bit, yeah, when/if we implement help I think we should add that bit back to message.  It's not in the group of stories currently pointed.  I think leaving it off the error message until we implement help makes the most sense?    at that point, maybe it would make sense to genericize all the subcommands error messages....

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/pubsub/SubscriptionsJUnitTest.java
jhutchison: babar the elephant

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/pubsub/AbstractSubCommandsIntegrationTest.java
jhutchison: yeah, the compiler still wants the cast, (this time to list<String>), but we don't need to suppress the unchecked cast warning  :)

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/pubsub/AbstractSubCommandsIntegrationTest.java
```diff
@@ -0,0 +1,156 @@
...
+
+  @Test
+  @SuppressWarnings("unchecked")
+  public void channels_shouldReturnListOfMatchingChannels_whenCalledWithPattern() {
+    ArrayList<byte[]> expectedChannels = new ArrayList<>();
+    expectedChannels.add("foo".getBytes());
+    Runnable runnable =
+        () -> subscriber.subscribe(mockSubscriber, "foo", "bar");
+    Thread subscriberThread = new Thread(runnable);
+
```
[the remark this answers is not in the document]
jhutchison: Jens-  does the test channels_shouldNotReturnChannelsWithOnlyPatternSubscribers() not work to clear that up?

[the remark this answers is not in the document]
jdeppe-pivotal: Kind of, except that that test isn't passing in a pattern as an arg to channels. I'd say you could even get rid of that test if you add a `fo*` subscription here.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/pubsub/SubCommandExecutor.java
jhutchison: yeah, I went back and forth about this.  I had all these files prefaced with pubsub initially, but "pubsubsubcommand*" feels a little like a tongue-twister. ..:).  My reasoning to leave it as subcommand* is that all of these files that start with subcommand live within directories named pubsub, which seems to clarify what command they belong to.   I think there is a little thought that needs to happen to break up the things currently called pubsub into Publish and subscribe.  I'd rather not do all that within this commit, if you all are ok with that, but will make a point of doing it before this "module" is done.  let me know if you all can live with that...  (so, to clarify, the idea I'm pitching is to leave the naming as it is for now,  but before the module is completed, rename stuff currently named pubsub  so that name becomes available, and then look at what to name all this subcommand stuff )

jdeppe-pivotal: Yes, we should definitely fix the other classes now that this command is introduced. I would like, at least, for this PR to have the executor named consistently as the other commands: `PubsubExecutor`

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/pubsub/AbstractSubCommandsIntegrationTest.java
jhutchison: There's only one test right now that psubscribes, and it punsubscribes within that test.  My thought was to give every test one subscriber and to unsubscribe from it-  if a test needs something special, it needs to take care of that itself...  If you're strongly opposed, let me know and I'll change.     


as far as the thread go, I saw that there is a lot of waiting on the threads to die in the subscribe test-  Nothing seems to need it in the subComands test right now as far as I can tell.  I would argue for waiting to add it until we need it?

jhutchison:
sorry for the push -f-  rebased onto master to pull in fastUtils dependency on develop to avoid including it twice.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/pubsub/AbstractSubCommandsIntegrationTest.java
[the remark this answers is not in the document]
upthewaterspout: Makes sense then. Thanks!

jdeppe-pivotal (approved):

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/pubsub/AbstractSubCommandsIntegrationTest.java
```diff
@@ -0,0 +1,203 @@
...
+    subscriber1Thread.start();
+
+    Runnable runnable2 = () -> subscriber2.subscribe(mockSubscriber2, "foo");
+    Thread subscriber2Thread = new Thread(runnable2);
+    subscriber2Thread.start();
+
+    waitFor(() -> {
+      return (mockSubscriber.getSubscribedChannels() == 1)
+          && (mockSubscriber2.getSubscribedChannels() == 1);
+    });
```
sabbey37, line 183: This could be replaced with an expression lambda:
```
waitFor(() -> (mockSubscriber.getSubscribedChannels() == 1)
        && (mockSubscriber2.getSubscribedChannels() == 1));

File: geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/pubsub/AbstractSubCommandsIntegrationTest.java [hunk 17]
```diff
@@ -0,0 +1,203 @@
...
+    expectedChannels.add("foo");
+
+    Runnable runnable =
+        () -> subscriber.subscribe(mockSubscriber, "foo", "bar");
+
+    Thread subscriberThread = new Thread(runnable);
+    subscriberThread.start();
+    waitFor(() -> mockSubscriber.getSubscribedChannels() == 2);
+    ArrayList<String> result =
+        (ArrayList<String>) introspector.pubsubChannels("fo*");
```
sabbey37, line 130: If you replace this with `List<String>`, you won't have to cast it.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/pubsub/AbstractSubCommandsIntegrationTest.java
```diff
@@ -0,0 +1,203 @@
...
+  @Test
+  public void channels_shouldNeverReturnPsubscribedChannels_givenNoActiveChannelSubscribers() {
+
+    Runnable runnable = () -> subscriber.psubscribe(mockSubscriber, "f*");
+    Thread patternSubscriberThread = new Thread(runnable);
+
+    patternSubscriberThread.start();
+    waitFor(() -> mockSubscriber.getSubscribedChannels() == 1);
+    ArrayList<String> result =
+        (ArrayList<String>) introspector.pubsubChannels("f*");
```
sabbey37, line 109: If you replace this with `List<String>`, you won't have to cast it.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/pubsub/AbstractSubCommandsIntegrationTest.java
```diff
@@ -0,0 +1,203 @@
...
+    ArrayList<String> result =
+        (ArrayList<String>) introspector.pubsubChannels("fo*");
+
+    assertThat(result).containsExactlyInAnyOrderElementsOf(expectedChannels);
+  }
+
+  @Test
+  public void channels_should_returnEmptyArry_GivenPatternWithNoMatches() {
+    ArrayList<String> result = (ArrayList<String>) introspector
+        .pubsubChannels("fo*");
```
sabbey37, line 138: If you replace this with `List<String>`, you won't have to cast it.

sabbey37 (changes requested):

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/pubsub/AbstractSubCommandsIntegrationTest.java
sabbey37: If you replace the type with `List<String>` rather than `ArrayList<String>`, you won't have to cast it.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/pubsub/AbstractSubCommandsIntegrationTest.java
sabbey37: In that case, it might make more sense to unsubscribe and wait for there to no longer be any subscribed channels in each test rather than in the after... since we're doing something slightly different for each one (some have multiple mockSubscribers).  

Yeah, I think it's ok to not waitFor subscriber thread to no longer be alive.  Wasn't sure if we added it at some point due to PubSub flakiness, but it seems that thread dies once there are no longer any subscribers so we should be ok just waiting for that condition.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/pubsub/AbstractSubCommandsIntegrationTest.java
```diff
@@ -0,0 +1,203 @@
...
+    subscriberThread.start();
+    waitFor(() -> mockSubscriber.getSubscribedChannels() == 2);
+    ArrayList<String> result =
+        (ArrayList<String>) introspector.pubsubChannels("fo*");
+
+    assertThat(result).containsExactlyInAnyOrderElementsOf(expectedChannels);
+  }
+
+  @Test
+  public void channels_should_returnEmptyArry_GivenPatternWithNoMatches() {
```
DonalEvans, line 136: Couple of typos here, should be "returnEmptyArray" and for consistency, "given" with a lower-case g.

DonalEvans (changes requested):
A few code clean-up suggestions.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/pubsub/AbstractSubCommandsIntegrationTest.java
jhutchison: I'm not sure... If this is something you want to block the merge on let me know and I'll make the change.  Otherwise I kind of can see a case for either way... ?

sabbey37: I think @jdeppe-pivotal had a preference for cleaning up those threads inside each test method, but I don't think either of us would block the merge because of that.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/pubsub/AbstractSubCommandsIntegrationTest.java
```diff
@@ -0,0 +1,203 @@
...
+    expectedChannels.add("foo".getBytes());
+    expectedChannels.add("bar".getBytes());
+    Runnable runnable =
+        () -> subscriber.subscribe(mockSubscriber, "foo", "bar");
+    Thread subscriberThread = new Thread(runnable);
+
+    subscriberThread.start();
+    waitFor(() -> mockSubscriber.getSubscribedChannels() == 2);
+    ArrayList<byte[]> result = (ArrayList<byte[]>) introspector
+        .sendCommand(Protocol.Command.PUBSUB, "channels");
```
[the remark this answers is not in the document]
jhutchison: sweet!

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/pubsub/Subscriptions.java
```diff
@@ -67,6 +73,33 @@ boolean exists(Object channelOrPattern, Client client) {
         .collect(Collectors.toList());
   }
 
+  public List<byte[]> findChannelNames() {
+
+    ObjectOpenCustomHashSet<byte[]> hashSet =
+        new ObjectOpenCustomHashSet<>(ByteArrays.HASH_STRATEGY);
+
+    findChannels()
+        .forEach(channel -> hashSet.add(channel.getSubscriptionName()));
```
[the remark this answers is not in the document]
jhutchison: as discussed, will put in a story to make the underlying collection in Subscriptions a Set instead of a List.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/pubsub/Subscription.java
```diff
@@ -18,6 +18,8 @@
 
 import java.util.List;
 
+import it.unimi.dsi.fastutil.objects.ObjectOpenCustomHashSet;
```
sabbey37, line 21: I don't think this import is currently used and could be removed.  `spA` should handle that though.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/RedisConstants.java
```diff
@@ -45,6 +45,7 @@
   public static final String ERROR_NOT_INTEGER = "value is not an integer or out of range";
   public static final String ERROR_OVERFLOW = "increment or decrement would overflow";
   public static final String ERROR_NAN_OR_INFINITY = "increment would produce NaN or Infinity";
+  public static final String ERROR_OPERATION_PRODUCED_NAN = "resulting score is not a number (NaN)";
```
sabbey37, line 48: Weirdly, it looks like these constants were added as a part of your PR.  Did you rebase with develop?

sabbey37 (approved):
Thank you for making those changes!  I think you might need to spA and rebase with develop.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/RedisConstants.java
jhutchison: I pulled in the conflicting changes in this file-  didn't do a full rebase.  I'll do that and see if it makes git happier

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/pubsub/AbstractSubCommandsIntegrationTest.java
```diff
@@ -0,0 +1,198 @@
...
+
+    mockSubscriber2.unsubscribe();
+    waitFor(() -> mockSubscriber2.getSubscribedChannels() == 0);
+
+    subscriber2.close();
+  }
+
+  private void waitFor(Callable<Boolean> booleanCallable) {
+    GeodeAwaitility.await()
+        .ignoreExceptions() // ignoring socket closed exceptions
```
DonalEvans, line 195: If it's only socket closed exceptions that you want to ignore here, it might be better to use `.ignoreExceptionsInstanceOf(SocketException.class)` rather than ignoring all exceptions.

DonalEvans (approved):
Looks good, thanks for addressing all the feedback. I added one more comment here, but it's fairly minor.

PR #6494, opened 2021-05-18
GEODE-xxxx: Configuration compatible with pre 1.12.0 versions

nabarunnag:
* Configuration send from 1.12.0 and later version locators can be interpreted by pre 1.12.0 version locators

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/management/internal/configuration/domain/Configuration.java
+++ b/geode-core/src/main/java/org/apache/geode/management/internal/configuration/domain/Configuration.java
@@ -19,6 +19,7 @@ import static org.apache.geode.internal.JarDeployer.getArtifactId;
 
 import java.io.DataInput;
 import java.io.DataOutput;
+import java.io.EOFException;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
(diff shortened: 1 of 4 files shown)


geode-core/src/main/java/org/apache/geode/management/internal/configuration/domain/Configuration.java
```diff
@@ -164,9 +165,10 @@ public void toData(DataOutput out) throws IOException {
     DataSerializer.writeString(propertiesFileName, out);
     DataSerializer.writeProperties(gemfireProperties, out);
     // Before 1.12, this code wrote a non-null HashSet of jarnames to the output stream.
-    // As of 1.12, it writes a null HashSet to the stream, so that when we can still read the old
-    // configuration, and will now also write the deployment map.
-    DataSerializer.writeHashSet(null, out);
+    // As of 1.12, it will also send the hashmap with the deployments, so that the new
+    // members can read the deployments while the older members can read just the hashset
+    HashSet<String> jarNames = new HashSet<>(deployments.keySet());
+    DataSerializer.writeHashSet(jarNames, out);
```
jinmeiliao, line 171: do you need to make a change here? Can we continue to write `null` to the stream instead of an empty set? this way, we can say, that if "jarName is not null, then it's definitely pre 1.12).

nabarunnag: We are not sending an empty set but rather a jarFileNames set. The problem is when older version 1.12 locators get this as `null` and it starts throwing NPE and corrupts the config

PR #6495, opened 2021-05-18
GEODE-9289: Configuration compatible with pre-1.12.0 versions

nabarunnag:
* Configuration sent from 1.12.0 locators and later can be interpreted by pre 1.12.0 version locators

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/management/internal/configuration/domain/Configuration.java
+++ b/geode-core/src/main/java/org/apache/geode/management/internal/configuration/domain/Configuration.java
@@ -17,6 +17,7 @@ package org.apache.geode.management.internal.configuration.domain;
 
 import java.io.DataInput;
 import java.io.DataOutput;
+import java.io.EOFException;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
@@ -166,10 +167,10 @@ public class Configuration implements DataSerializable {
     DataSerializer.writeString(cacheXmlContent, out);
     DataSerializer.writeString(propertiesFileName, out);
     DataSerializer.writeProperties(gemfireProperties, out);
-    // Before 1.12, this code wrote a non-null HashSet of jarnames to the output stream.
-    // As of 1.12, it writes a null HashSet to the stream, so that when we can still read the old
-    // configuration, and will now also write the deployment map.
-    DataSerializer.writeHashSet(null, out);
+    // As of 1.12, it writes a jarNames HashSet to the stream, so that pre 1.12.0 members can
+    // read the jarName, and will now also write the deployment map for the post 1.12.0 members.
+    HashSet<String> jarNames = new HashSet<>(deployments.keySet());
+    DataSerializer.writeHashSet(jarNames, out);
     // As of 1.12, this class starting writing the current version
     VersioningIO.writeOrdinal(out, KnownVersion.getCurrentVersion().ordinal(), true);
     DataSerializer.writeHashMap(deployments, out);
@@ -183,16 +184,18 @@ public class Configuration implements DataSerializable {
     propertiesFileName = DataSerializer.readString(in);
     gemfireProperties = DataSerializer.readProperties(in);
     HashSet<String> jarNames = DataSerializer.readHashSet(in);
-    if (jarNames != null) {
-      // we are reading pre 1.12 data. So add each jar name to deployments
-      jarNames.stream()
-          .map(x -> new Deployment(x, null, null))
-          .forEach(deployment -> deployments.put(deployment.getDeploymentName(), deployment));
-    } else {
+    try {
       // version of the data we are reading (1.12 or later)
       final Version version = Versioning.getVersion(VersioningIO.readOrdinal(in));
       if (version.isNotOlderThan(KnownVersion.GEODE_1_12_0)) {
         deployments.putAll(DataSerializer.readHashMap(in));
+      }
+    } catch (EOFException ex) {
+      if (jarNames != null) {
+        // we are reading pre 1.12 data. So add each jar name to deployments
+        jarNames.stream()
+            .map(x -> new Deployment(x, null, null))
+            .forEach(deployment -> deployments.put(deployment.getDeploymentName(), deployment));
       }
     }
   }
@@ -231,4 +234,5 @@ public class Configuration implements DataSerializable {
     return Objects.hash(configName, cacheXmlContent, cacheXmlFileName, propertiesFileName,
         gemfireProperties, deployments);
   }
+
 }
--- a/geode-gfsh/src/upgradeTest/java/org/apache/geode/management/ConfigurationCompatibilityTest.java
+++ b/geode-gfsh/src/upgradeTest/java/org/apache/geode/management/ConfigurationCompatibilityTest.java
@@ -1,0 +1,110 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.apache.geode.management;
+
+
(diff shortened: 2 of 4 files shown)


mhansonp (approved):
This looks good. I would really like to see a unit test of the serialization, just to verify that it does what it says. e.g. serial the thing, deserialize the thing, verify the values.

nabarunnag:
@mhansonp I completely agree with you but in this scenario, the serialization and deserialization code is split between Apache Geode 1.10.0 and Apache Geode 1.12.0+ hence it will be not possible to write the unit test.

geode-core/src/main/java/org/apache/geode/management/internal/configuration/domain/Configuration.java
```diff
@@ -183,20 +185,48 @@ public void fromData(DataInput in) throws IOException, ClassNotFoundException {
...
+      if (jarNames != null) {
+        // we are reading pre 1.12 data. So add each jar name to deployments
+        jarNames.stream()
+            .map(x -> new Deployment(x, null, null))
+            .forEach(deployment -> deployments.put(deployment.getDeploymentName(), deployment));
+      }
     }
   }
 
+  public void toDataPre_GEODE_1_12_0_0(DataOutput out) throws IOException {
```
DonalEvans, line 204: Would it be possible to first call this method in `toData()` and then add the two extra lines that have been added there, to keep the two methods consistent?

geode-core/src/main/java/org/apache/geode/management/internal/configuration/domain/Configuration.java
```diff
@@ -183,20 +185,48 @@ public void fromData(DataInput in) throws IOException, ClassNotFoundException {
     propertiesFileName = DataSerializer.readString(in);
     gemfireProperties = DataSerializer.readProperties(in);
     HashSet<String> jarNames = DataSerializer.readHashSet(in);
-    if (jarNames != null) {
-      // we are reading pre 1.12 data. So add each jar name to deployments
-      jarNames.stream()
-          .map(x -> new Deployment(x, null, null))
-          .forEach(deployment -> deployments.put(deployment.getDeploymentName(), deployment));
-    } else {
+    try {
```
DonalEvans, line 188: Given that we now have a fromDataPre_X method, it should be unnecessary to have this try/catch block, because we can be sure that we'll be reading from a member with the correct version here.

geode-core/src/main/java/org/apache/geode/management/internal/configuration/domain/Configuration.java
```diff
@@ -183,20 +185,48 @@ public void fromData(DataInput in) throws IOException, ClassNotFoundException {
...
+    DataSerializer.writeString(configName, out);
+    DataSerializer.writeString(cacheXmlFileName, out);
+    DataSerializer.writeString(cacheXmlContent, out);
+    DataSerializer.writeString(propertiesFileName, out);
+    DataSerializer.writeProperties(gemfireProperties, out);
+    HashSet<String> jarNames = new HashSet<>(deployments.keySet());
+    DataSerializer.writeHashSet(jarNames, out);
+  }
+
+  public void fromDataPre_GEODE_1_12_0_0(DataInput in) throws IOException, ClassNotFoundException {
```
DonalEvans, line 214: I know that the implementations differ when it comes to how to handle the `jarNames` set, but would it be possible to find a way to have `fromData()` first call this method and then do any additional work that's specific to the post-1.12.0 implementation, just to reduce code duplication?

mhansonp:
> @mhansonp I completely agree with you but in this scenario, the serialization and deserialization code is split between Apache Geode 1.10.0 and Apache Geode 1.12.0+ hence it will be not possible to write the unit test.

I did this recently with Anil's help would it work? GatewaySenderAdvisorTest.java has an example. It may not work for you.. Either way, I still approved the change. It is not necessary, just nice, if the approach works at all...

geode-gfsh/src/upgradeTest/java/org/apache/geode/management/ConfigurationCompatibilityTest.java
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  @Rule
+  public ClusterStartupRule clusterStartupRule = new ClusterStartupRule();
+
+  @Rule
+  public GfshCommandRule gfsh = new GfshCommandRule();
+
+  @Test
+  public void whenConfigurationIsExchangedBetweenMixedVersionLocatorsThenItShouldNotThrowExceptions()
```
DonalEvans, line 58: When I revert the changes in this PR, this test still passes, so I'm not sure that it's testing what you're intending it to test.

DonalEvans (changes requested):

geode-core/src/main/java/org/apache/geode/management/internal/configuration/domain/Configuration.java
nabarunnag: the catch part is when it is reading from a pre 1.12.0 version because it will hit the EOFException is trying to read the Configuration when it is being sent through EntryEventImpl from region (ClusterConfig region) replication.  EntryEventImpl does not send the version in the stream and hence the fromData_pre is not triggered.
But in the case of servers, they request it via function execution, version is sent through the stream and that's when the xxData_pre_XX is triggered.

geode-core/src/main/java/org/apache/geode/management/internal/configuration/domain/Configuration.java
nabarunnag: I think I can completely remove them and let this work via the exception try catch model itself for the function execution too.,

geode-gfsh/src/upgradeTest/java/org/apache/geode/management/ConfigurationCompatibilityTest.java
nabarunnag: I attempted this in the 1.12.0 branch : https://github.com/apache/geode/pull/6494
same code changes and if you revert it in there, the test fails complaining NPEs were encountered, let me check what changed in develop.

geode-core/src/main/java/org/apache/geode/management/internal/configuration/domain/Configuration.java
```diff
@@ -183,20 +185,48 @@ public void fromData(DataInput in) throws IOException, ClassNotFoundException {
...
-          .map(x -> new Deployment(x, null, null))
-          .forEach(deployment -> deployments.put(deployment.getDeploymentName(), deployment));
-    } else {
+    try {
       // version of the data we are reading (1.12 or later)
       final Version version = Versioning.getVersion(VersioningIO.readOrdinal(in));
       if (version.isNotOlderThan(KnownVersion.GEODE_1_12_0)) {
         deployments.putAll(DataSerializer.readHashMap(in));
       }
+    } catch (EOFException ex) {
```
jinmeiliao, line 194: When the xxData_pre_XX is not triggered, the existence of jarNames should be enough to determine whether we are reading a pre1.12 data or not. Do we have to use a try/catch block here?

nabarunnag: Unfortunately, when EntryEventImpl is sent, the version is not sent in the stream, hence in that case pre_XX is not triggered even if it is sent from pre 1.12.0 version

geode-core/src/main/java/org/apache/geode/management/internal/configuration/domain/Configuration.java
```diff
@@ -169,7 +170,8 @@ public void toData(DataOutput out) throws IOException {
     // Before 1.12, this code wrote a non-null HashSet of jarnames to the output stream.
     // As of 1.12, it writes a null HashSet to the stream, so that when we can still read the old
```
upthewaterspout, line 171: This comment seems outdated since you are now writing the hash set of jar names.

geode-gfsh/src/upgradeTest/java/org/apache/geode/management/ConfigurationCompatibilityTest.java
nabarunnag: @DonalEvans @upthewaterspout https://github.com/apache/geode/pull/6526 is created to fix the DUnit runner to detect suspicious messages in logs of VMs when they are restarted.
There was a behavior change in 1.14 hence the tests was not failing in 1.14 but in 1.12

DonalEvans: Thanks for the update. Once that PR is merged, I'll re-check that the test fails without the changes then approve this PR.

DonalEvans (approved):
Confirmed that the fix in #6526 allows the test added in this PR to fail without the changes, so I'm approving this PR.

jinmeiliao (approved):

dschneider-pivotal (approved):

PR #6496, opened 2021-05-18
GEODE-8951: Unnecessary messaging in WAN locator discovery

alb3rtobr:
The ticket states that when a locator joins the cluster, it is notified about itself joining the system.
That problem was already solved by GEODE-6977, in `locatorJoined(...)` method of `LocatorMembershipListenerImpl` class, before starting the thread that will send the notifications about the new locator:

```
    // Remove locators that don't need to be notified (myself, the joining one and the one that
    // notified myself).
    List<DistributionLocatorId> ignoreList = Arrays.asList(locator, localLocatorId, sourceLocator);
    for (Map.Entry<Integer, Set<DistributionLocatorId>> entry : localCopy.entrySet()) {
      for (DistributionLocatorId removeLocId : ignoreList) {
        entry.getValue().remove(removeLocId);
      }
    }
```
Anyway, in order to verify that no more than the expected number of notification messages are sent, a check can be added to three unit tests of `LocatorMembershipListenerImpl` class. These test cases verify that the proper messages are sent, but they would not fail if unexpected messages were sent.


alb3rtobr:
Kind reminder, review by code owners still pending

alb3rtobr:
thanks @boglesby !

PR #6500, opened 2021-05-19
GEODE-8870: Removes GFE_70.

pivotal-jbarrett:
The PR is intentionally broken into multiple commits. The work to just remove the version is in the first commit. 

The followup commits are to cleanup all the compiler and static analyzer warnings in the changed files.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/Oplog.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/Oplog.java
@@ -32,7 +32,6 @@ import java.nio.channels.ClosedChannelException;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
@@ -58,7 +57,6 @@ import org.apache.logging.log4j.Logger;
 import org.apache.geode.CancelException;
 import org.apache.geode.DataSerializer;
 import org.apache.geode.SerializationException;
-import org.apache.geode.annotations.Immutable;
 import org.apache.geode.cache.CacheClosedException;
 import org.apache.geode.cache.CacheWriterException;
 import org.apache.geode.cache.DiskAccessException;
@@ -108,7 +106,6 @@ import org.apache.geode.internal.serialization.KnownVersion;
 import org.apache.geode.internal.serialization.Versioning;
 import org.apache.geode.internal.serialization.VersioningIO;
 import org.apache.geode.internal.shared.NativeCalls;
-import org.apache.geode.internal.util.BlobHelper;
 import org.apache.geode.logging.internal.log4j.api.LogService;
 import org.apache.geode.pdx.internal.PdxWriterImpl;
 import org.apache.geode.util.internal.GeodeGlossary;
@@ -191,7 +188,7 @@ public class Oplog implements CompactableOplog, Flushable {
   private final AtomicLong totalLiveCount = new AtomicLong(0);
 
   private final ConcurrentMap<Long, DiskRegionInfo> regionMap =
-      new ConcurrentHashMap<Long, DiskRegionInfo>();
+      new ConcurrentHashMap<>();
 
   /**
    * Set to true once compact is called on this oplog.
@@ -542,14 +539,14 @@ public class Oplog implements CompactableOplog, Flushable {
           "Too many oplogs. The oplog id can not exceed " + DiskId.MAX_OPLOG_ID);
     }
     this.oplogId = oplogId;
-    this.oplogSet = parent;
+    oplogSet = parent;
     this.parent = parent.getParent();
     this.dirHolder = dirHolder;
     // Pretend we have already seen the first record.
     // This will cause a large initial record to force a switch
     // which allows the maxDirSize to be checked.
-    this.firstRecord = false;
-    this.opState = new OpState();
+    firstRecord = false;
+    opState = new OpState();
     long maxOplogSizeParam = getParent().getMaxOplogSizeInBytes();
     long availableSpace = this.dirHolder.getAvailableSpace();
     if (availableSpace < maxOplogSizeParam) {
@@ -562,22 +559,22 @@ public class Oplog implements CompactableOplog, Flushable {
                 + availableSpace + ", required=" + maxOplogSizeParam),
             getParent());
       }
-      this.maxOplogSize = availableSpace;
+      maxOplogSize = availableSpace;
       if (logger.isDebugEnabled()) {
         logger.debug(
             "Reducing maxOplogSize to {} because that is all the room remaining in the directory.",
             availableSpace);
       }
     } else {
-      this.maxOplogSize = maxOplogSizeParam;
+      maxOplogSize = maxOplogSizeParam;
     }
     setMaxCrfDrfSize();
-    this.stats = getParent().getStats();
-    this.compactOplogs = getParent().getAutoCompact();
+    stats = getParent().getStats();
+    compactOplogs = getParent().getAutoCompact();
 
-    this.closed = false;
+    closed = false;
     String n = getParent().getName();
-    this.diskFile = new File(this.dirHolder.getDir(), oplogSet.getPrefix() + n + "_" + oplogId);
+    diskFile = new File(this.dirHolder.getDir(), oplogSet.getPrefix() + n + "_" + oplogId);
     try {
       createDrf(null);
       createCrf(null);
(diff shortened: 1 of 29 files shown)


geode-core/src/test/java/org/apache/geode/internal/DataSerializableJUnitTest.java
```diff
@@ -1166,13 +1166,13 @@ public void testVersionedDataSerializable() throws Exception {
     VersionedDataSerializableImpl ds = new VersionedDataSerializableImpl(getRandom());
 
     VersionedDataOutputStream v =
-        new VersionedDataOutputStream(this.baos, KnownVersion.GFE_70);
+        new VersionedDataOutputStream(this.baos, KnownVersion.GEODE_1_11_0);
```
DonalEvans: Is there a specific reason for this version to be used here? Similarly for the use of `GEODE_1_12_0` below.

geode-core/src/main/java/org/apache/geode/internal/cache/Oplog.java
```diff
@@ -5723,89 +5590,79 @@ public DirectoryHolder getDirectoryHolder() {
...
     // synchronized block does not attempt to get the backup lock (incorrect lock order)
-    synchronized (this.lock/* crf */) {
-      this.doneAppending = true;
+    synchronized (lock/* crf */) {
+      doneAppending = true;
     }
     handleNoLiveValues();
     // I'm deadcoding the following because it is not safe unless we change to
     // always recover values. If we don't recover values then
     // an oplog we recovered from may still need to fault values in from memory.
```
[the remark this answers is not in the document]
DonalEvans: Sounds good. I'll approve the PR now, on the assumption that the commit will get pushed before it's merged.

DonalEvans (approved):

geode-core/src/main/java/org/apache/geode/internal/cache/persistence/DiskInitFileParser.java
```diff
@@ -567,24 +568,20 @@ private void readEndOfRecord(DataInput di) throws IOException {
...
       throws IOException, ClassNotFoundException {
     ByteArrayInputStream bais = new ByteArrayInputStream(bytes);
     DataInputStream dis = new DataInputStream(bais);
     PersistentMemberID result = new PersistentMemberID();
-    if (KnownVersion.GFE_70.compareTo(gfversion) > 0) {
-      result._fromData662(dis);
-    } else {
-      InternalDataSerializer.invokeFromData(result, dis);
-    }
+    InternalDataSerializer.invokeFromData(result, dis);
```
upthewaterspout, line 584: In general, when we read from disk or the network we should be creating a VersionedDataInputStream. That allows us to make DataSerializableFixedId objects backwards compatible just by adding special fromData_Pre_XXX methods.

Maybe we should create one of those here, using gfVersion.

upthewaterspout (commented):
Thanks for cleaning things up! I reviewed my codeowner files and had one comment, inline.

geode-core/src/main/java/org/apache/geode/internal/cache/persistence/DiskInitFileParser.java
pivotal-jbarrett: I think if this method needs versioning support added it should come when there is a version change here or a separate PR to add versioning support and tests. The only change here was to pull out the obsolete version check which left us with the original un-versioned from data call.

Going through all these cleanups and digging deep into areas of code I haven't been I have noticed there is inconsistent use of `VersionedDataInputStream` and `VersionedDataSerializable`. I wonder if we should have an effort to normalize on the versioned versions (🥁) and deprecate the older stuff.

echobravopapa (approved):
my codeowner files look good, thanks for the cleanup

agingade:
I see lot more refactoring changes than the fix for actual issue. I agree and believe in incremental refactoring as we come across the opportunity; but here I see lot more refactoring changes; even in non-related files (where the actual change is not expected); it will be nice to keep those changes minimal and focused on the issue being fixed...just my thoughts and I am fine with other way too...

pivotal-jbarrett:
> I see lot more refactoring changes than the fix for actual issue. I agree and believe in incremental refactoring as we come across the opportunity; but here I see lot more refactoring changes; even in non-related files (where the actual change is not expected); it will be nice to keep those changes minimal and focused on the issue being fixed...just my thoughts and I am fine with other way too...

The fix and the cleanup on on two separate commits to make reviewing easier. The other cleaned files are files affected by the correction of issues in a dependent class.

dschneider-pivotal (approved):

kirklund (approved):

gesterzhou (approved):

boglesby (approved):

geode-core/src/main/java/org/apache/geode/internal/cache/TXCommitMessage.java
```diff
@@ -1384,153 +1323,145 @@ protected void txApplyEntryOpAdjunctOnly(FarSideEntryOp entryOp) {
...
-                || KnownVersion.GFE_70.compareTo(this.msg.clientVersion) <= 0;
-        if (sendVersionTags) {
-          VersionSource member = this.memberId;
-          if (member == null) {
-            if (this.internalRegion == null) {
-              Assert.assertTrue(this.msg.txState == null);
-            } else {
-              member = this.internalRegion.getVersionMember();
-            }
+        VersionSource member = memberId;
```
nabarunnag, line 1448: Here we have completely removed the check for msg.clientVersion for null. Looking at the code there are some places where this valus is set to nul like in CommitCommand. Now the member version source is always sent, is this behavior change correct?

geode-core/src/main/java/org/apache/geode/internal/cache/TXCommitMessage.java
```diff
@@ -1384,153 +1323,145 @@ protected void txApplyEntryOpAdjunctOnly(FarSideEntryOp entryOp) {
...
         for (int i = 0; i < size; i++) {
-          DataSerializer.writeObject(this.opKeys.get(i), out);
-          if (this.msg.txState != null) {
+          DataSerializer.writeObject(opKeys.get(i), out);
+          if (msg.txState != null) {
             /* we are still on tx node and have the entry state */
-            ((TXEntryState) this.opEntries.get(i)).toFarSideData(out, context, largeModCount,
-                sendVersionTags, useShadowKey);
+            ((TXEntryState) opEntries.get(i)).toFarSideData(out, context, largeModCount,
+                true, useShadowKey);
```
nabarunnag, line 1462: Now sendVersionTags is always set to to true irrespective of if msg.clientVersion is null or not.

nabarunnag (changes requested):
Few concerns regarding the changes in the behavior of toData in TXCommitMessage 
Changes may not be needed if a valid explanation is available.

nabarunnag:
Also, I would request refactorings to be kept in separate PRs. 
- The caption JIRA mentions that this GEODE-8870 ticket is for removal of obsolete GFE_70 code but has GEODE-6588 along with it which has other major refactorings.
- Keeping the CL shorter will have faster approval / thorough review as the reviewer list smaller
- We view the final set of files changed as it is more convenient than looking at each and every commit individually, as the same file may have been modified in multiple commits.

geode-core/src/main/java/org/apache/geode/internal/cache/TXCommitMessage.java
pivotal-jbarrett: It would send the `VersionSource` if it didn't know the `clientVersion` or it was newer than v7.0. In this case the latter part is always true so the check for unknown or `null` `clientVersion` is redundant.

geode-core/src/main/java/org/apache/geode/internal/cache/TXCommitMessage.java
pivotal-jbarrett: For the same reason as above.

nabarunnag (approved):

geode-core/src/main/java/org/apache/geode/internal/cache/TXCommitMessage.java
nabarunnag: Thank you for the explanation

pivotal-jbarrett:
> Also, I would request refactorings to be kept in separate PRs.
> 
> * The caption JIRA mentions that this GEODE-8870 ticket is for removal of obsolete GFE_70 code but has GEODE-6588 along with it which has other major refactorings.
> * Keeping the CL shorter will have faster approval / thorough review as the reviewer list smaller
> * We view the final set of files changed as it is more convenient than looking at each and every commit individually, as the same file may have been modified in multiple commits.

I think you should take this preference to a discussion on dev@geode. We have had similar discussions without an clear consensus. The closest we got was around not forcing squashes on PRs that include explicitly distinguished fix and cleanup commits, like this one.
https://markmail.org/message/hflkinvcxjg6stbo

nabarunnag:
Sure but GEODE-6588 could have been another PRs and limit this PR to GEODE-8870

pivotal-jbarrett:
> Sure but GEODE-6588 could have been another PRs and limit this PR to GEODE-8870

Is the issue that there is cleanup of files touched by the fix in this PR as separate commits or that those commits have a different JIRA issue associated with them?

PR #6510, opened 2021-05-21
GEODE-9298: remove concourse deprecation warnings

rhoughton-pivot:
Merging this PR will wipe Concourse history due to pipeline and/or job
name changes.

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -37,11 +37,11 @@ github:
         strict: false
         # contexts are the names of checks that must pass
         contexts:
-          - concourse-ci/ApiCheckTestOpenJDK11
-          - concourse-ci/Build
-          - concourse-ci/StressNewTestOpenJDK11
-          - concourse-ci/UnitTestOpenJDK11
-          - concourse-ci/UnitTestOpenJDK8
+          - concourse-ci/api-check-test-openjdk11
+          - concourse-ci/build
+          - concourse-ci/stress-new-test-openjdk11
+          - concourse-ci/unit-test-openjdk11
+          - concourse-ci/unit-test-openjdk8
 
       required_pull_request_reviews:
         dismiss_stale_reviews: false
--- a/ci/pipelines/examples/jinja.template.yml
+++ b/ci/pipelines/examples/jinja.template.yml
@@ -96,7 +96,7 @@ resources:
 - name: concourse-metadata-resource
   type: concourse-metadata-resource
   source: {}
-- name: 24h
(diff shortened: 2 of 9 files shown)


rhoughton-pivot:
Test deployment of the `meta` pipeline here: https://concourse.apachegeode-ci.info/teams/main/pipelines/rhoughton-pivot-feature-concourse-de-meta

dickcav (approved):

moleske:
If you don't want to lose history, there is an `old_name` field available which will allow you keep the history https://concourse-ci.org/jobs.html#schema.job.old_name

rhoughton-pivot:
> If you don't want to lose history, there is an `old_name` field available which will allow you keep the history https://concourse-ci.org/jobs.html#schema.job.old_name
> 
> I used it on my current teams concourse when fixing deprecation warnings and it worked well for us. It also exists for resources

I should add that to keep folk happy.

moleske:
@rhoughton-pivot also forgot to mention that fly has a command for rename pipeline https://concourse-ci.org/managing-pipelines.html#fly-rename-pipeline to preserve pipeline history if changing the pipeline name.  Might be harder to do with automation, but it also should be a one time thing

onichols-pivotal (approved):
preserving history is not important to me (Jira links to junit results and test artifacts will still work, regardless).  but if you do want to go to the trouble of adding old_name everywhere, you might consider flying it manually, rather that committing it that way (since it's really a one-time thing)

rhoughton-pivot:
On second look, the way we have the jobs templated, we can't use any simple Jinja tricks to modify the names and keep the old ones, and the full rename-effort is not worth it for the sake of history. Any artifacts are uploaded to GCS as needed anyway.

onichols-pivotal (changes requested):
please also change `_` to `-` in the three benchmarks jobs names for consistency

rhoughton-pivot:
> please also change `_` to `-` in the three benchmarks jobs names for consistency, and update .asf.yaml as well (even though the copy on develop is not yet the source of truth, it will eventually overwrite what's on master when 1.15.0 is released)

Done. Please re-review.

onichols-pivotal (approved):

moleske (approved):
Alas, poor job history!  I knew it, Robert: pages of infinite green and red, of most excellent fancy

demery-pivotal (approved):
Yay! Now I can remove my keyboard's shift key!

PR #6511, opened 2021-05-23
GEODE-7864: Clean up a lot of the IntelliJ warnings in Radish

DonalEvans:
- Rename ParameterRequirements to follow package naming best practices
 - Add missing serialVersionUIDs
 - Add final to fields and variables that can be final
 - Fix various raw uses of parameterized classes
 - Make variables local where possible
 - Remove unused and redundant variables
 - Remove some unused methods
 - Remove uses of toArray() with pre-sized array argument. In older Java
 versions using pre-sized array was recommended, as the reflection call
 which is necessary to create an array of proper size was quite slow.
 However since late updates of OpenJDK 6 this call was intrinsified,
 making the performance of the empty array version the same and
 sometimes even better, compared to the pre-sized version.
 - Fix some typos

These changes reduce the number of inspection warnings in the
geode-apis-compatible-with-redis module from about 450 to about 136

### For all changes:

### Note:


geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/pubsub/PubSubImpl.java
```diff
@@ -66,16 +69,18 @@ public int getSubscriptionCount() {
   public long publish(Region<RedisKey, RedisData> dataRegion, byte[] channel, byte[] message) {
     PartitionRegionInfo info = PartitionRegionHelper.getPartitionRegionInfo(dataRegion);
     Set<DistributedMember> membersWithDataRegion = new HashSet<>();
-    for (PartitionMemberInfo memberInfo : info.getPartitionMemberInfo()) {
-      membersWithDataRegion.add(memberInfo.getDistributedMember());
+    if (info != null) {
```
ringles, line 72: Is there a test that checks the necessity of this code?

ringles (changes requested):
Just a few nitpicks, otherwise looks good!

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/pubsub/PubSubImpl.java
DonalEvans: Looking into it a bit, this null check should be unnecessary, as `PartitionRegionHelper.getPartitionRegionInfo()` only returns null if we hit a `ClassCastException` during it, but nothing in that method actually throws one. It's outside the scope of these changes to go fixing it though, so for now I'll just remove the check and let IntelliJ complain about it.

ringles (approved):
Lots of little things, but does improve the code. Annoyingly, I used a bunch of Longs instead of longs in ZAdd, gonna have to fix that.

DonalEvans:
> Lots of little things, but does improve the code. Annoyingly, I used a bunch of Longs instead of longs in ZAdd, gonna have to fix that.

This is going to need to be rebased on develop before merging, so I can add those changes to the PR after doing that.

DonalEvans:
Closing this as rebasing will be almost impossible. I'll create a new PR based on develop after the last of the big refactors for this module are done.

PR #6524, opened 2021-05-27
GEODE-9318: Implement ZREM command.

pivotal-eshu:

### For all changes:

### Note:


geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZRemIntegrationTest.java
```diff
@@ -0,0 +1,178 @@
...
+      assertThat(score).isNull();
+    }
+    assertThat(jedis.exists(SORTED_SET_KEY)).isFalse();
+  }
+
+  @Test
+  public void zRemCanRemovesMembersConcurrentlyInASortedSet() {
+    int membersCount = 1000;
+    Map<String, Double> map = makeMemberScoreMap(membersCount);
+    addToSortedSet(map);
```
sabbey37, line 131: You can actually just `zadd` all the keys/values in the map directly:
```
jedis.zadd(SORTED_SET_KEY, map);
```

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZRemIntegrationTest.java
```diff
@@ -0,0 +1,178 @@
...
+    Set<String> keys = map.keySet();
+    Long count = 0L;
+
+    for (String member : keys) {
+      Double score = map.get(member);
+      Long res = jedis.zadd(SORTED_SET_KEY, score, member);
+      count += res;
+    }
+    assertThat(count).isEqualTo(keys.size());
+  }
```
sabbey37, line 167: This method seems unnecessary since we can `zadd` all the keys/values in the map directly.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -87,7 +87,9 @@ protected void applyDelta(DeltaInfo deltaInfo) {
       membersAddAll(addsDeltaInfo);
     } else {
       RemsDeltaInfo remsDeltaInfo = (RemsDeltaInfo) deltaInfo;
-      membersRemoveAll(remsDeltaInfo);
+      for (byte[] member : remsDeltaInfo.getRemoves()) {
+        memberRemove(member);
+      }
```
sabbey37, line 92: Could we also update the way we're applying delta for adds so that it's consistent with how we've changed the removes, so eliminating the `memberAddAll` method and moving that logic into this method?

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSetCommandsFunctionExecutor.java
```diff
@@ -43,4 +43,10 @@ public long zadd(RedisKey key, List<byte[]> scoresAndMembersToAdd, ZAddOptions o
   public byte[] zscore(RedisKey key, byte[] member) {
     return stripedExecute(key, () -> getRedisSortedSet(key, true).zscore(member));
   }
+
+  @Override
+  public long zrem(RedisKey key, List<byte[]> membersToRemove) {
+    return stripedExecute(key,
+        () -> getRedisSortedSet(key, false).zrem(getRegion(), key, membersToRemove));
```
sabbey37, line 50: I didn't think about this for ZADD/ZSCORE, but we should also add tests to `AbstractHitsMissesIntegrationTest` verifying that stats should/shouldn't be updated for `ZREM` and all new commands.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/ZRemExecutor.java
```diff
@@ -0,0 +1,40 @@
...
+public class ZRemExecutor extends AbstractExecutor {
+
+  @Override
+  public RedisResponse executeCommand(Command command, ExecutionHandlerContext context) {
+    RedisSortedSetCommands redisSortedSetCommands = context.getRedisSortedSetCommands();;
+
+    List<byte[]> commandElements = command.getProcessedCommand();
+    RedisKey key = command.getKey();
+    ArrayList<byte[]> membersToRemove =
+        new ArrayList<>(commandElements.subList(2, commandElements.size()));
```
sabbey37, line 35: We could use `List<byte[]>` for the variable type here.

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisSortedSetTest.java
```diff
@@ -232,6 +237,45 @@ public void should_calculateSize_closeToROSSize_ofIndividualInstanceWithMultiple
...
+    int originalSize = sortedSet.getSizeInBytes();
+
+    byte[] returnValue = sortedSet.memberRemove(Coder.stringToBytes(member1));
+    int removedSize =
+        sortedSet.calculateSizeOfFieldValuePair(Coder.stringToBytes(member1), returnValue);
+
+    assertThat(sortedSet).isEqualTo(sortedSet2);
+    assertThat(returnValue).isEqualTo(Coder.stringToBytes(score1));
+    assertThat(sortedSet.getSizeInBytes()).isEqualTo(originalSize - removedSize);
+  }
```
sabbey37, line 277: For checking that the sizeInBytes is accurate, I'd prefer we use the `ReflectionObjectSizer.sizeOf` directly in the tests.  I think @nonbinaryprogrammer has been working on a story to fix the sizeable calculations though.  The other assertions in this test seem unnecessary since they were already validated by integration tests.

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisSortedSetTest.java
```diff
@@ -232,6 +237,45 @@ public void should_calculateSize_closeToROSSize_ofIndividualInstanceWithMultiple
...
+    ArrayList<byte[]> membersToRemove = new ArrayList<>();
+    membersToRemove.add(Coder.stringToBytes("nonExisting"));
+    membersToRemove.add(Coder.stringToBytes(member1));
+    membersToRemove.add(Coder.stringToBytes(member3));
+
+    long removed = sortedSet.zrem(region, key, membersToRemove);
+
+    assertThat(removed).isEqualTo(2);
+    verify(sortedSet).storeChanges(eq(region), eq(key), any(RemsDeltaInfo.class));
+  }
```
sabbey37, line 262: This is a cool test!  I'm not sure it's necessary though since all of this should've been validated via the DUnit tests, but it is kind of cool verifying all of this at a lower level.

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisSortedSetTest.java
```diff
@@ -116,7 +121,7 @@ public void equals_returnsTrue_givenDifferentEmptySortedSets() {
 
   @Test
   public void zadd_stores_delta_that_is_stable() throws IOException {
-    Region<RedisKey, RedisData> region = uncheckedCast(Mockito.mock(Region.class));
+    Region<RedisKey, RedisData> region = uncheckedCast(mock(Region.class));
```
sabbey37, line 124: Thanks for making this better!

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/sortedset/ZRemDUnitTest.java
```diff
@@ -0,0 +1,266 @@
...
+  @Test
+  public void zRemCanRemovesMembersConcurrentlyFromSortedSet() {
+    Map<String, Double> memberScoreMap = makeMemberScoreMap(setSize);
+    jedis.zadd(sortedSetKey, memberScoreMap);
+    verifyDataExist(memberScoreMap);
+
+    AtomicInteger totalRemoved = new AtomicInteger();
+    new ConcurrentLoopingThreads(2,
+        (i) -> doZRem(memberScoreMap, totalRemoved),
+        (i) -> doZRem1(totalRemoved)).run();
```
sabbey37, line 120: Since ConcurrentLoopingThreads loops for us, maybe we could loop through the full `setSize` and do the zrem individually:
```
 new ConcurrentLoopingThreads(setSize,
        (i) -> doZRem(i, totalRemoved),
        (i) -> doZRem(i, totalRemoved)).run();
 ```
 ```
 private void doZRem(int i, AtomicInteger totalRemoved) {
      long count = jedis.zrem(SORTED_SET_KEY, baseName + i);
      totalRemoved.addAndGet((int) count);
  }
  ```

File: geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/sortedset/ZRemDUnitTest.java [hunk 10]
```diff
@@ -0,0 +1,266 @@
...
+
+    long removed = jedis.zrem(sortedSetKey, memberScoreMap.keySet().toArray(new String[] {}));
+    assertThat(removed).isEqualTo(setSize);
+
+    verifyDataNotExist(memberScoreMap);
+    assertThat(jedis.exists(sortedSetKey)).isFalse();
+  }
+
+  @Test
+  public void zRemCanRemovesMembersConcurrentlyFromSortedSet() {
```
sabbey37, line 112: Could this be `zRemCanRemove...`?

sabbey37 (changes requested):
Thank you for adding this functionality!

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZRemIntegrationTest.java
```diff
@@ -0,0 +1,178 @@
...
+  @Test
+  public void zRemCanRemovesMembersConcurrentlyInASortedSet() {
+    int membersCount = 1000;
+    Map<String, Double> map = makeMemberScoreMap(membersCount);
+    addToSortedSet(map);
+
+    AtomicInteger totalRemoved = new AtomicInteger();
+    new ConcurrentLoopingThreads(2,
+        (i) -> doZRem(map, totalRemoved),
+        (i) -> doZRem1(membersCount, totalRemoved)).run();
```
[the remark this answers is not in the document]
pivotal-eshu: The issue I see with the above methods is that each thread loop on the same member in order. The second thread always trying to remove the first thread just removed (understandably it is how we use the synchronization to prevent concurrent modification on the sorted set).

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/sortedset/ZRemDUnitTest.java
```diff
@@ -0,0 +1,266 @@
...
+
+  private void doZRem(Map<String, Double> map, AtomicInteger total) {
+    Set<String> keys = map.keySet();
+    for (String key : keys) {
+      long count = jedis.zrem(sortedSetKey, key);
+      total.addAndGet((int) count);
+    }
+  }
+
+  private void doZRem1(AtomicInteger total) {
```
DonalEvans, line 134: Could these methods be renamed to something a little more descriptive? Maybe "doZRemOnAllKeysInMap" and "doZRemOnAllMembers"?

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/sortedset/ZRemDUnitTest.java
```diff
@@ -0,0 +1,266 @@
...
+        assertThat(score).isNull();
+      }
+    } catch (JedisClusterMaxAttemptsException e) {
+      return false;
+    }
+    return true;
+  }
+
+  private void stopNodeWithPrimaryBucketOfTheKey(boolean isCrash) {
+    int numOfServers = 4;
```
DonalEvans, line 225: Rather than hard-coding this value, it might be better to add the servers to a `List<MemberVM>` in the `setUp()` method and iterate through that here.

DonalEvans (changes requested):
A few test clean-up suggestions.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZRemIntegrationTest.java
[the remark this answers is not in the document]
sabbey37: I think that's kind of the point though... that we have two threads trying to remove something from the same set at the same time. We want to make sure we don't hit and concurrent modification exceptions or report we've removed the same thing twice. We also can't be sure which thread will run/complete first.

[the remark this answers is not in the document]
jdeppe-pivotal: The order that the threads execute the commands in `ConcurrentLoopingThreads` is not deterministic. So either one could be 'first' to remove a given entry.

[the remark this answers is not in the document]
pivotal-eshu: At least that is what I see from running the test with above mentioned methods. The second one is always blocked on the synchronization code in Redis -- when that thread is unblocked and working on the key. The first thread is blocked on the key. At least this is what I found. I think we at least want to test that both threads can remove some of the members in the sorted set.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
pivotal-eshu: Actually I decide to keep the removeAll method as it uses synchronized keyword - indicating potentially multiple threads can work on the data set. Use removeAll and addAll, no additional thread will be able to modify members concurrently -all or none.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSetCommandsFunctionExecutor.java
pivotal-eshu: Added the test coverage.

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisSortedSetTest.java
pivotal-eshu: This is just to test in the unit level to verify -- we want to have coverage at lower level if possible.

I added a test case using ReflectionObjectSizer.sizeOf but need to be changed.

DonalEvans (approved):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
sabbey37: Good call.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZRemIntegrationTest.java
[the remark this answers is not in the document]
jdeppe-pivotal: Ideally, the function you pass into `ConcurrentLoopingThreads` shouldn't be doing its own loop. As it is currently set up, the CLT will only loop twice and the second time everything is already gone. Perhaps one slight adjustment to @sabbey37's suggestion is to replace `run()` with `runInLockstep()`. This means that for every iteration of the CLT loop, all threads will wait for completion of a given iteration before moving to the next one. The idea is that it increases the possibility of concurrency errors if they exist.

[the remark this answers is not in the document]
pivotal-eshu: Removed the own loop in the methods.

jdeppe-pivotal (approved):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSetCommandsFunctionExecutor.java
sabbey37: Thank you for adding those tests!

sabbey37 (approved):

PR #6525, opened 2021-05-27
GEODE-9292: Implement ZCARD

luissson:

### For all changes:

### Note:

### DIFF
--- a/geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/sortedset/ZCardDUnitTest.java
+++ b/geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/sortedset/ZCardDUnitTest.java
@@ -1,0 +1,93 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.apache.geode.redis.internal.executor.sortedset;
+
+import static org.apache.geode.distributed.ConfigurationProperties.MAX_WAIT_TIME_RECONNECT;
+import static org.apache.geode.test.dunit.rules.RedisClusterStartupRule.DEFAULT_MAX_WAIT_TIME_RECONNECT;
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Properties;
+import java.util.Random;
+
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Test;
+import redis.clients.jedis.HostAndPort;
+import redis.clients.jedis.Jedis;
+import redis.clients.jedis.JedisCluster;
+
+import org.apache.geode.test.awaitility.GeodeAwaitility;
+import org.apache.geode.test.dunit.rules.MemberVM;
+import org.apache.geode.test.dunit.rules.RedisClusterStartupRule;
+
+public class ZCardDUnitTest {
+
+
+  @ClassRule
+  public static RedisClusterStartupRule clusterStartup = new RedisClusterStartupRule(4);
+
+  private static final String LOCAL_HOST = "127.0.0.1";
+  private static final int JEDIS_TIMEOUT =
+      Math.toIntExact(GeodeAwaitility.getTimeout().toMillis());
+
+  private static JedisCluster jedis;
+
+  @BeforeClass
+  public static void classSetup() {
+    Properties locatorProperties = new Properties();
+    locatorProperties.setProperty(MAX_WAIT_TIME_RECONNECT, DEFAULT_MAX_WAIT_TIME_RECONNECT);
+
+    MemberVM locator = clusterStartup.startLocatorVM(0, locatorProperties);
+    int locatorPort = locator.getPort();
+
+    clusterStartup.startRedisVM(1, locatorPort);
+    clusterStartup.startRedisVM(2, locatorPort);
+    clusterStartup.startRedisVM(3, locatorPort);
+
+    int redisServerPort = clusterStartup.getRedisPort(1);
+
+    jedis = new JedisCluster(new HostAndPort(LOCAL_HOST, redisServerPort), JEDIS_TIMEOUT);
+  }
+
+  @Before
+  public void beforeTest() {
+    try (Jedis conn = jedis.getConnectionFromSlot(0)) {
+      conn.flushAll();
+    }
+  }
+
+  @Test
+  public void shouldReturnSizeOfManySets_OnDifferentServers() {
+    Random random = new Random();
+    for (int i = 0; i < 1000; i++) {
+      int setSize = 1 + random.nextInt(1000);
+      String key = "key" + i;
+      Map<String, Double> memberScoreMap = makeMemberScoreMap("member" + i, setSize);
+      jedis.zadd(key, memberScoreMap);
+      assertThat(jedis.zcard(key)).isEqualTo(setSize);
+    }
+  }
+
+  private Map<String, Double> makeMemberScoreMap(String baseString, int setSize) {
+    Map<String, Double> scoreMemberPairs = new HashMap<>();
+    for (int i = 0; i < setSize; i++) {
+      scoreMemberPairs.put(baseString + i, Double.valueOf(i + ""));
+    }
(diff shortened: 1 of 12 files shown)


geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZCardIntegrationTest.java
```diff
@@ -0,0 +1,80 @@
...
+
+import org.apache.geode.redis.RedisIntegrationTest;
+import org.apache.geode.test.awaitility.GeodeAwaitility;
+
+public abstract class AbstractZCardIntegrationTest implements RedisIntegrationTest {
+
+  private static final int SET_SIZE = 1000;
+  private static final String LOCAL_HOST = "127.0.0.1";
+  private static final int JEDIS_TIMEOUT =
+      Math.toIntExact(GeodeAwaitility.getTimeout().toMillis());
```
DonalEvans, line 38: Instead of defining a new constant here, could you just use `REDIS_CLIENT_TIMEOUT` in `RedisIntegrationTest`?

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZCardIntegrationTest.java
```diff
@@ -0,0 +1,80 @@
...
+import redis.clients.jedis.JedisCluster;
+import redis.clients.jedis.Protocol;
+
+import org.apache.geode.redis.RedisIntegrationTest;
+import org.apache.geode.test.awaitility.GeodeAwaitility;
+
+public abstract class AbstractZCardIntegrationTest implements RedisIntegrationTest {
+
+  private static final int SET_SIZE = 1000;
+  private static final String LOCAL_HOST = "127.0.0.1";
```
DonalEvans, line 36: There's some inconsistency here with other Redis integration tests, which all use a hard-coded value of `"localhost"` to create their JedisClusters. I think it might be best to pull that string up to `RedisIntegrationTest` rather than having it be defined in every test that implements it.

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/sortedset/ZCardDUnitTest.java
```diff
@@ -0,0 +1,123 @@
...
+  public static RedisClusterStartupRule clusterStartup = new RedisClusterStartupRule(4);
+
+  private static final int SET_SIZE = 1000;
+  private static final String LOCAL_HOST = "127.0.0.1";
+  private static final int JEDIS_TIMEOUT =
+      Math.toIntExact(GeodeAwaitility.getTimeout().toMillis());
+
+  private static MemberVM server1;
+  private static MemberVM server2;
+  private static MemberVM server3;
```
DonalEvans, line 50: These three variables can all be safely removed, as they're never used, just assigned.

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/sortedset/ZCardDUnitTest.java
```diff
@@ -0,0 +1,123 @@
...
+  public void shouldReturnSizeOfOneSet() {
+    String key = "key";
+    Map<String, Double> memberScoreMap = makeMemberScoreMap("member1-");
+    jedis.zadd(key, memberScoreMap);
+
+    assertThat(jedis.zcard(key)).isEqualTo(SET_SIZE);
+  }
+
+  @Test
+  public void shouldReturnSizeOfManySets_OnDifferentServers() {
```
DonalEvans, line 89: How are you ensuring that these sets all end up on different servers? It's possible that the keys will hash to slots hosted on the same server, unless I'm missing something.

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/sortedset/ZCardDUnitTest.java
```diff
@@ -0,0 +1,123 @@
...
+
+  @Before
+  public void beforeTest() {
+    try (Jedis conn = jedis.getConnectionFromSlot(0)) {
+      conn.flushAll();
+    }
+  }
+
+  @Test
+  public void shouldReturnSizeOfOneSet() {
```
DonalEvans, line 80: I think it might not be necessary to have a test like this, or `shouldReturnZeroWhenKeyAbsent` for DUnit, since it's not testing anything that the integration test isn't already covering.

DonalEvans (changes requested):
Looks good, just a few test clean-up suggestions.

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/sortedset/ZCardDUnitTest.java
luissson: This could work just for the current implementation. Each key will hash to a different slot by using the hashtag, but on my end I can't guarantee those slots are on different servers. But in this case I know they land in different servers because I added temp logs to check.

DonalEvans: One possible solution to guarantee there are sets stored on every server might be to create a lot of them, so the chances of one server not hosting any of the buckets to which the keys map is effectively zero:
```
    for (int i = 0; i < 1000; ++i) {
      String key = "key" + i;
      Map<String, Double> memberScoreMap = makeMemberScoreMap("member" + i);
      jedis.zadd(key, memberScoreMap);
      assertThat(jedis.zcard(key)).isEqualTo(SET_SIZE);
    }
```
You could also make the set size variable for each set, to confirm that the size returned is distinct and that there's not something weird happening where some other set's size is being returned, just to make the test a little stronger.

luissson: Great suggestion, I've redone the test to randomize the set size and the key. I've also checked the logs and see multiple servers carrying out zcard.

DonalEvans (approved):

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/sortedset/ZCardDUnitTest.java
```diff
@@ -0,0 +1,93 @@
...
+      Math.toIntExact(GeodeAwaitility.getTimeout().toMillis());
+
+  private static JedisCluster jedis;
+
+  @BeforeClass
+  public static void classSetup() {
+    Properties locatorProperties = new Properties();
+    locatorProperties.setProperty(MAX_WAIT_TIME_RECONNECT, DEFAULT_MAX_WAIT_TIME_RECONNECT);
+
+    MemberVM locator = clusterStartup.startLocatorVM(0, locatorProperties);
```
sabbey37, line 55: We don't actually need to set any properties on the locator.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZCardIntegrationTest.java
```diff
@@ -0,0 +1,76 @@
...
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import redis.clients.jedis.HostAndPort;
+import redis.clients.jedis.JedisCluster;
+import redis.clients.jedis.Protocol;
+
+import org.apache.geode.redis.RedisIntegrationTest;
+
+public abstract class AbstractZCardIntegrationTest implements RedisIntegrationTest {
```
sabbey37, line 32: Could we also add a test verifying the proper error message is returned when `ZCARD` is called with a key whose value is not a sorted set?

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSetCommandsFunctionExecutor.java
```diff
@@ -43,4 +43,9 @@ public long zadd(RedisKey key, List<byte[]> scoresAndMembersToAdd, ZAddOptions o
   public byte[] zscore(RedisKey key, byte[] member) {
     return stripedExecute(key, () -> getRedisSortedSet(key, true).zscore(member));
   }
+
+  @Override
+  public long zcard(RedisKey key) {
+    return stripedExecute(key, () -> getRedisSortedSet(key, true).zcard());
```
sabbey37, line 49: Could we add a test to `AbstractHitsMissesIntegrationTest` verifying that stats should be updated for `ZCARD`?

sabbey37 (commented):
Awesome job adding this command, Rene!  Just commenting on a couple tests that would be good to add.

luissson:
> Awesome job adding this command, Rene! Just commenting on a couple tests that would be good to add.

Thanks for the feedback! I've made the changes except for adding zcard to AbstractHitsMissesIntegrationTest; I will wait for the ZREM PR to get merged first to avoid conflict

sabbey37 (approved):

PR #6534, opened 2021-05-27
GEODE-9306: support ZINCRBY command

nonbinaryprogrammer:
-implement the command
-add integration test
-add concurrency test
-support infinity as a valid increment

### For all changes:

### Note:

### DIFF
--- a/geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/server/AbstractHitsMissesIntegrationTest.java
+++ b/geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/server/AbstractHitsMissesIntegrationTest.java
@@ -199,8 +199,7 @@ public abstract class AbstractHitsMissesIntegrationTest implements RedisIntegrat
   /************* Sorted Set related commands *************/
   @Test
   public void testZIncrBy() {
-    runCommandAndAssertNoStatUpdates("key", 100.0,
-        (String k, Double i, String m) -> jedis.zincrby(k, i, m));
+    runCommandAndAssertNoStatUpdates("key", (k, m) -> jedis.zincrby(k, 100.0, m));
   }
 
   /**********************************************
@@ -613,19 +612,6 @@ public abstract class AbstractHitsMissesIntegrationTest implements RedisIntegrat
     String currentMisses = info.get(MISSES);
 
     command.accept(key, key, "42");
-    info = getInfo(jedis);
-
-    assertThat(info.get(HITS)).isEqualTo(currentHits);
-    assertThat(info.get(MISSES)).isEqualTo(currentMisses);
-  }
-
-  private void runCommandAndAssertNoStatUpdates(String key, Double increment,
-      TriConsumer<String, Double, String> command) {
-    Map<String, String> info = getInfo(jedis);
-    String currentHits = info.get(HITS);
-    String currentMisses = info.get(MISSES);
-
-    command.accept(key, increment, "42");
     info = getInfo(jedis);
 
     assertThat(info.get(HITS)).isEqualTo(currentHits);
--- a/geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/server/AbstractHitsMissesIntegrationTest.java
+++ b/geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/server/AbstractHitsMissesIntegrationTest.java
@@ -65,7 +65,7 @@ public abstract class AbstractHitsMissesIntegrationTest implements RedisIntegrat
   /***********************************************
    ************* Supported Commands **************
    **********************************************/
-  // ------------ Key related commands -----------
+  /************* Key related commands *************/
 
   @Test
   public void testKeys() {
@@ -102,7 +102,7 @@ public abstract class AbstractHitsMissesIntegrationTest implements RedisIntegrat
     runCommandAndAssertNoStatUpdates("string", k -> jedis.del(k));
   }
 
-  // ------------ String related commands -----------
+  /************* String related commands *************/
 
   @Test
   public void testGet() {
@@ -124,7 +124,7 @@ public abstract class AbstractHitsMissesIntegrationTest implements RedisIntegrat
     runCommandAndAssertNoStatUpdates("set", (k, v) -> jedis.set(k, v));
   }
 
-  // ------------ Set related commands -----------
+  /************** Set related commands *************/
   @Test
   public void testSadd() {
     runCommandAndAssertNoStatUpdates("set", (k, v) -> jedis.sadd(k, v));
@@ -140,7 +140,7 @@ public abstract class AbstractHitsMissesIntegrationTest implements RedisIntegrat
     runCommandAndAssertHitsAndMisses("set", k -> jedis.smembers(k));
   }
 
-  // ------------ Hash related commands -----------
+  /************* Hash related commands *************/
   @Test
   public void testHset() {
     runCommandAndAssertNoStatUpdates("hash", (k, v, s) -> jedis.hset(k, v, s));
@@ -160,7 +160,7 @@ public abstract class AbstractHitsMissesIntegrationTest implements RedisIntegrat
     runCommandAndAssertNoStatUpdates("key", (k) -> jedis.hmset(k, map));
   }
 
-  // ------------ Key related commands -----------
+  /************* Key related commands *************/
 
   @Test
   public void testExpire() {
@@ -194,6 +194,13 @@ public abstract class AbstractHitsMissesIntegrationTest implements RedisIntegrat
   @Test
   public void testPersist() {
     runCommandAndAssertNoStatUpdates("hash", (k) -> jedis.persist(k));
+  }
+
+  /************* Sorted Set related commands *************/
+  @Test
+  public void testZIncrBy() {
+    runCommandAndAssertNoStatUpdates("key", 100.0,
+        (String k, Double i, String m) -> jedis.zincrby(k, i, m));
   }
 
   /**********************************************
@@ -606,6 +613,19 @@ public abstract class AbstractHitsMissesIntegrationTest implements RedisIntegrat
     String currentMisses = info.get(MISSES);
 
     command.accept(key, key, "42");
+    info = getInfo(jedis);
+
+    assertThat(info.get(HITS)).isEqualTo(currentHits);
+    assertThat(info.get(MISSES)).isEqualTo(currentMisses);
+  }
+
+  private void runCommandAndAssertNoStatUpdates(String key, Double increment,
+      TriConsumer<String, Double, String> command) {
+    Map<String, String> info = getInfo(jedis);
+    String currentHits = info.get(HITS);
+    String currentMisses = info.get(MISSES);
+
+    command.accept(key, increment, "42");
     info = getInfo(jedis);
 
     assertThat(info.get(HITS)).isEqualTo(currentHits);
--- a/geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZIncrByIntegrationTest.java
+++ b/geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZIncrByIntegrationTest.java
@@ -1,0 +1,241 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package org.apache.geode.redis.internal.executor.sortedset;
+
+import static java.lang.Double.NEGATIVE_INFINITY;
+import static java.lang.Double.POSITIVE_INFINITY;
+import static org.apache.geode.redis.RedisCommandArgumentsTestHelper.assertExactNumberOfArgs;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+import java.util.concurrent.ThreadLocalRandom;
+
+import com.google.common.util.concurrent.AtomicDouble;
+import org.assertj.core.data.Offset;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Ignore;
+import org.junit.Test;
+import redis.clients.jedis.HostAndPort;
+import redis.clients.jedis.JedisCluster;
+import redis.clients.jedis.Protocol;
+
+import org.apache.geode.redis.ConcurrentLoopingThreads;
+import org.apache.geode.redis.RedisIntegrationTest;
+import org.apache.geode.redis.internal.RedisConstants;
+
+public abstract class AbstractZIncrByIntegrationTest implements RedisIntegrationTest {
+  JedisCluster jedis;
+  final byte[] KEY = "key".getBytes();
+  final String STRING_KEY = "key";
+  final byte[] MEMBER = "member".getBytes();
+  final String STRING_MEMBER = "member";
+
+  @Before
+  public void setUp() {
+    jedis = new JedisCluster(new HostAndPort("localhost", getPort()), REDIS_CLIENT_TIMEOUT);
+  }
+
+  @After
+  public void tearDown() {
+    flushAll();
+    jedis.close();
+  }
+
+  /************* errors *************/
+  @Test
+  public void shouldError_givenWrongKeyType() {
+    jedis.set(STRING_KEY, "value");
+    assertThatThrownBy(
+        () -> jedis.sendCommand(STRING_KEY, Protocol.Command.ZINCRBY, STRING_KEY, "1", "member"))
+            .hasMessageContaining("WRONGTYPE " + RedisConstants.ERROR_WRONG_TYPE);
+  }
+
+  @Test
+  public void shouldError_givenWrongNumberOfArguments() {
+    jedis.zadd(KEY, 0.0, MEMBER);
+    assertExactNumberOfArgs(jedis, Protocol.Command.ZINCRBY, 3);
+  }
+
+  @Test
+  public void shouldError_givenNonFloatIncrement() {
+    String nonFloatIncrement = "q";
+    jedis.zadd(KEY, 1.0, MEMBER);
+
+    assertThatThrownBy(() -> jedis.sendCommand(STRING_KEY, Protocol.Command.ZINCRBY, STRING_KEY,
+        nonFloatIncrement, STRING_MEMBER))
+            .hasMessageContaining(RedisConstants.ERROR_NOT_A_VALID_FLOAT);
+  }
+
+  @Test
+  public void shouldError_givenNaNIncrement() {
+    String nanIncrement = "NaN";
+    jedis.zadd(STRING_KEY, 1.0, STRING_MEMBER);
+
+    assertThatThrownBy(() -> jedis.sendCommand(STRING_KEY, Protocol.Command.ZINCRBY, STRING_KEY,
+        nanIncrement, STRING_MEMBER))
+            .hasMessageContaining(RedisConstants.ERROR_NOT_A_VALID_FLOAT);
+  }
+
+  /************* infinity *************/
+  @Test
+  public void shouldSetScoreToInfinity_ifIncrementWouldExceedMaxValue() {
+    double increment = Double.MAX_VALUE / 2;
+
+    jedis.zadd(STRING_KEY, Double.MAX_VALUE, STRING_MEMBER);
+    jedis.zincrby(STRING_KEY, increment, STRING_MEMBER);
+    assertThat(jedis.zscore(STRING_KEY, STRING_MEMBER)).isEqualTo(POSITIVE_INFINITY);
+
+    jedis.zadd(STRING_KEY, -Double.MAX_VALUE, STRING_MEMBER);
+    jedis.zincrby(STRING_KEY, -increment, STRING_MEMBER);
+    assertThat(jedis.zscore(STRING_KEY, STRING_MEMBER)).isEqualTo(NEGATIVE_INFINITY);
+  }
+
+  @Test
+  public void shouldSetScoreToInfinity_givenInfiniteIncrement() {
+    jedis.zadd(STRING_KEY, 1.0, STRING_MEMBER);
+    assertThat(jedis.zincrby(STRING_KEY, POSITIVE_INFINITY, STRING_MEMBER))
+        .isEqualTo(POSITIVE_INFINITY);
+
+    jedis.zadd(STRING_KEY, 1.0, STRING_MEMBER);
+    assertThat(jedis.zincrby(STRING_KEY, NEGATIVE_INFINITY, STRING_MEMBER))
+        .isEqualTo(NEGATIVE_INFINITY);
+
+    jedis.zadd(STRING_KEY, 1.0, STRING_MEMBER);
+    jedis.sendCommand(STRING_KEY, Protocol.Command.ZINCRBY, STRING_KEY, "inf", STRING_MEMBER);
+    assertThat(jedis.zscore(STRING_KEY, STRING_MEMBER)).isEqualTo(POSITIVE_INFINITY);
+
+    jedis.zadd(STRING_KEY, 1.0, STRING_MEMBER);
+    jedis.sendCommand(STRING_KEY, Protocol.Command.ZINCRBY, STRING_KEY, "+inf", STRING_MEMBER);
+    assertThat(jedis.zscore(STRING_KEY, STRING_MEMBER)).isEqualTo(POSITIVE_INFINITY);
+
+    jedis.zadd(STRING_KEY, 1.0, STRING_MEMBER);
+    jedis.sendCommand(STRING_KEY, Protocol.Command.ZINCRBY, STRING_KEY, "-inf", STRING_MEMBER);
+    assertThat(jedis.zscore(STRING_KEY, STRING_MEMBER)).isEqualTo(NEGATIVE_INFINITY);
+
+    jedis.zadd(STRING_KEY, 1.0, STRING_MEMBER);
+    jedis.sendCommand(STRING_KEY, Protocol.Command.ZINCRBY, STRING_KEY, "Infinity", STRING_MEMBER);
+    assertThat(jedis.zscore(STRING_KEY, STRING_MEMBER)).isEqualTo(POSITIVE_INFINITY);
+
+    jedis.zadd(STRING_KEY, 1.0, STRING_MEMBER);
+    jedis.sendCommand(STRING_KEY, Protocol.Command.ZINCRBY, STRING_KEY, "+Infinity", STRING_MEMBER);
+    assertThat(jedis.zscore(STRING_KEY, STRING_MEMBER)).isEqualTo(POSITIVE_INFINITY);
+
+    jedis.zadd(STRING_KEY, 1.0, STRING_MEMBER);
+    jedis.sendCommand(STRING_KEY, Protocol.Command.ZINCRBY, STRING_KEY, "-Infinity", STRING_MEMBER);
+    assertThat(jedis.zscore(STRING_KEY, STRING_MEMBER)).isEqualTo(NEGATIVE_INFINITY);
+  }
+
+  @Ignore("blocked by GEODE-9317")
+  @Test
+  public void scoreOfPositiveInfinityShouldRemainInfinite_whenIncrementingOrDecrementing() {
+    jedis.zadd(KEY, POSITIVE_INFINITY, MEMBER);
+
+    jedis.zincrby(KEY, 100.0, MEMBER);
+    assertThat(jedis.zscore(KEY, MEMBER)).isEqualTo(POSITIVE_INFINITY);
+
+    jedis.zincrby(KEY, -1000.0, MEMBER);
+    assertThat(jedis.zscore(KEY, MEMBER)).isEqualTo(POSITIVE_INFINITY);
+  }
+
+  @Ignore("blocked by GEODE-9317")
+  @Test
+  public void scoreOfNegativeInfinityShouldRemainInfinite_whenIncrementingOrDecrementing() {
+    jedis.zadd(KEY, NEGATIVE_INFINITY, MEMBER);
+
+    jedis.zincrby(KEY, 100.0, MEMBER);
+    assertThat(jedis.zscore(KEY, MEMBER)).isEqualTo(NEGATIVE_INFINITY);
+
+    jedis.zincrby(KEY, -1000.0, MEMBER);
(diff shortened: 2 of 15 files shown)


geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/RedisResponse.java
```diff
@@ -71,6 +71,14 @@ public static RedisResponse string(byte[] byteArray) {
...
   }
 
+  public static RedisResponse floatDouble(byte[] byteArray) {
+    return new RedisResponse((buffer) -> Coder.getSimpleStringResponse(buffer, byteArray));
+  }
+
+  public static RedisResponse floatDouble(String stringValue) {
+    return new RedisResponse((buffer) -> Coder.getSimpleStringResponse(buffer, stringValue));
+  }
+
```
jdeppe-pivotal, line 81: I think the semantics of this class and its static methods is intended to mimic the RESP protocol. From that perspective, seeing `RedisResponse.floatDouble` might lead someone to think that RESP can represent a float/double. Prefer instead a `string(double)` method which converts a double into the appropriate RESP response (string in this case).

As an aside, I see that we have `bigDecimal` here which should be changed to adhere to the same semantics (but not in this PR).

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -252,4 +277,28 @@ public String toString() {
...
+        break;
+      case "-inf":
+      case "-infinity":
+        incr = NEGATIVE_INFINITY;
+        break;
+      default:
+        try {
+          incr = Double.parseDouble(stringIncr);
+        } catch (NumberFormatException nfe) {
+          return ERROR_NOT_A_VALID_FLOAT.getBytes();
```
jdeppe-pivotal, line 298: At this point, the error should just be left to bubble up and be processed 'normally'. You can just do:
```
throw new NumberFormatException(ERROR_NOT_A_VALID_FLOAT);
```

jdeppe-pivotal (changes requested):
Just a couple of comments. Really nice test coverage!!

sabbey37 (changes requested):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSetCommandsFunctionExecutor.java
```diff
@@ -43,4 +43,10 @@ public long zadd(RedisKey key, List<byte[]> scoresAndMembersToAdd, ZAddOptions o
   public byte[] zscore(RedisKey key, byte[] member) {
     return stripedExecute(key, () -> getRedisSortedSet(key, true).zscore(member));
   }
+
+  @Override
+  public byte[] zincrby(RedisKey key, byte[] increment, byte[] member) {
+    return stripedExecute(key,
+        () -> getRedisSortedSet(key, false).zincrby(getRegion(), key, increment, member));
+  }
```
sabbey37, line 51: Could we add a test to `AbstractHitsMissesIntegrationTest` verifying that stats should be updated for `ZINCRBY`?

sabbey37 (commented):
Thanks for adding this and noting that we need to handle +/-infinity properly for all commands!

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSetCommandsFunctionExecutor.java
nonbinaryprogrammer: lol I already forgot about it since we talked about it 10 minutes ago 😅 wilco

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -228,6 +230,30 @@ private void validateScoreIsDouble(byte[] score) {
...
 
+  byte[] zincrby(Region<RedisKey, RedisData> region, RedisKey key, byte[] increment,
+      byte[] member) {
+    byte[] score = members.get(member);
+    byte[] incr = Coder.doubleToBytes(
+        processIncrement(Coder.bytesToString(increment).toLowerCase()));
+
+    if (Arrays.equals(incr, ERROR_NOT_A_VALID_FLOAT.getBytes())) {
+      return incr;
+    }
```
jdeppe-pivotal, line 241: Since `processIncrement` is just throwing an exception now, this check and be removed.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSetCommandsFunctionExecutor.java
nonbinaryprogrammer: done

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/ZIncrByExecutor.java
```diff
@@ -0,0 +1,54 @@
...
+    skipCommandAndKey(commandIterator);
+    byte[] increment = commandIterator.next();
+    byte[] member = commandIterator.next();
+
+    byte[] retVal = redisSortedSetCommands.zincrby(command.getKey(), increment, member);
+    if (Arrays.equals(retVal, ERROR_NOT_A_VALID_FLOAT.getBytes())) {
+      return RedisResponse.error(ERROR_NOT_A_VALID_FLOAT);
+    } else if (Arrays.equals(retVal, ERROR_NAN_OR_INFINITY.getBytes())) {
+      return RedisResponse.error(ERROR_NAN_OR_INFINITY);
+    }
```
sabbey37, line 46: You could also remove the checks here since you're throwing the error in `processIncrement`

sabbey37 (changes requested):
I think you might've missed [this comment](https://github.com/apache/geode/pull/6534#discussion_r643279579) I left in a previous review.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZIncrByIntegrationTest.java
```diff
@@ -0,0 +1,272 @@
...
+import org.junit.Test;
+import redis.clients.jedis.HostAndPort;
+import redis.clients.jedis.JedisCluster;
+import redis.clients.jedis.Protocol;
+
+import org.apache.geode.redis.ConcurrentLoopingThreads;
+import org.apache.geode.redis.RedisIntegrationTest;
+import org.apache.geode.redis.internal.RedisConstants;
+
+public abstract class AbstractZIncrByIntegrationTest implements RedisIntegrationTest {
```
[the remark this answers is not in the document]
nonbinaryprogrammer: added!

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/server/AbstractHitsMissesIntegrationTest.java
```diff
@@ -195,6 +195,13 @@ public void testPersist() {
     runCommandAndAssertNoStatUpdates("hash", (k) -> jedis.persist(k));
   }
 
+  /************* Sorted Set related commands *************/
+  @Test
+  public void testZIncrBy() {
+    runCommandAndAssertNoStatUpdates("key", 100.0,
+        (String k, Double i, String m) -> jedis.zincrby(k, i, m));
```
jdeppe-pivotal: I think we'll want to refer to an existing `"sortedset"` here.
Also, consider not creating a specialized `runCommandAssAssertNoStatUpdates` in favor of simply passing a fixed value in the lambda:
```
runCommandAndAssertNoStatUpdates("key", (k, v) -> jedis.zincrby(k, 100.0, v));
```

nonbinaryprogrammer: already changed :)

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZIncrByIntegrationTest.java
```diff
@@ -0,0 +1,290 @@
...
+  @Test
+  public void scoreOfNegativeInfinityShouldRemainInfinite_whenIncrementingOrDecrementing() {
+    jedis.zadd(KEY, NEGATIVE_INFINITY, MEMBER);
+
+    jedis.zincrby(KEY, 100.0, MEMBER);
+    assertThat(jedis.zscore(KEY, MEMBER)).isEqualTo(NEGATIVE_INFINITY);
+
+    jedis.zincrby(KEY, -1000.0, MEMBER);
+    assertThat(jedis.zscore(KEY, MEMBER)).isEqualTo(NEGATIVE_INFINITY);
+  }
```
sabbey37, line 195: We don't actually need to use `ZADD` for these, you could do:
```
    jedis.zincrby(KEY, NEGATIVE_INFINITY, MEMBER);

    assertThatThrownBy(() -> jedis.zincrby(KEY, POSITIVE_INFINITY, MEMBER))
        .hasMessageContaining(RedisConstants.ERROR_OPERATION_PRODUCED_NAN);

    assertThatThrownBy(() -> jedis.zincrby(KEY, POSITIVE_INFINITY, MEMBER))
        .hasMessageContaining(RedisConstants.ERROR_OPERATION_PRODUCED_NAN);
        ```

**Review by sabbey37 — CHANGES_REQUESTED** [review]

File: geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZIncrByIntegrationTest.java [hunk 9]
```diff
@@ -0,0 +1,290 @@
...
+
+  @Ignore("blocked by GEODE-9317")
+  @Test
+  public void incrementingScoreOfNegativeInfinityByPositiveInfinity_throwsNaNError() {
+    jedis.zadd(KEY, NEGATIVE_INFINITY, MEMBER);
+
+    assertThatThrownBy(() -> jedis.zincrby(KEY, POSITIVE_INFINITY, MEMBER))
+        .hasMessageContaining(RedisConstants.ERROR_OPERATION_PRODUCED_NAN);
+  }
```
sabbey37, line 213: We don't actually need to use `ZADD` for these, you could do:
```
    jedis.zincrby(KEY, NEGATIVE_INFINITY, MEMBER);

    assertThatThrownBy(() -> jedis.zincrby(KEY, POSITIVE_INFINITY, MEMBER))
        .hasMessageContaining(RedisConstants.ERROR_OPERATION_PRODUCED_NAN);
 ```

File: geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java [hunk 10]
```diff
@@ -228,6 +230,26 @@ private void validateScoreIsDouble(byte[] score) {
...
 
+  byte[] zincrby(Region<RedisKey, RedisData> region, RedisKey key, byte[] increment,
+      byte[] member) {
+    byte[] score = members.get(member);
+    byte[] incr = Coder.doubleToBytes(
+        processIncrement(Coder.bytesToString(increment).toLowerCase()));
+
+    if (score != null) {
+      incr = Coder.doubleToBytes(Coder.bytesToDouble(score) + Coder.bytesToDouble(incr));
+    }
```
sabbey37, line 241: I imagine you could check the `incr` variable after for `NaN`, something like this, though there might be a better way of doing it:
```
 if(Arrays.equals(incr, "NaN".getBytes())) {
      throw new NumberFormatException(ERROR_OPERATION_PRODUCED_NAN);
  }
 ```

File: geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZIncrByIntegrationTest.java [hunk 11]
```diff
@@ -0,0 +1,290 @@
...
+    jedis.close();
+  }
+
+  /************* errors *************/
+  @Test
+  public void shouldError_givenWrongKeyType() {
+    jedis.set(STRING_KEY, "value");
+    assertThatThrownBy(
+        () -> jedis.sendCommand(STRING_KEY, Protocol.Command.ZINCRBY, STRING_KEY, "1", "member"))
+            .hasMessageContaining("WRONGTYPE " + RedisConstants.ERROR_WRONG_TYPE);
```
DonalEvans, line 64: "WRONGTYPE" is defined as a constant in `Coder` as `Coder.wrongType` so maybe that should be used here. (Also, incidentally, that constant should probably be renamed to follow the all-caps constant naming convention.)

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZIncrByIntegrationTest.java
```diff
@@ -0,0 +1,290 @@
...
+
+    jedis.zadd(STRING_KEY, 1.0, STRING_MEMBER);
+    jedis.sendCommand(STRING_KEY, Protocol.Command.ZINCRBY, STRING_KEY, "-Infinity", STRING_MEMBER);
+    assertThat(jedis.zscore(STRING_KEY, STRING_MEMBER)).isEqualTo(NEGATIVE_INFINITY);
+  }
+
+  @Test
+  public void nullRedisSetShouldCreateNonExistentMember_withIncrementOfPositiveInfinity() {
+    final double increment = POSITIVE_INFINITY;
+    jedis.zadd(KEY, 0.0, "something".getBytes()); // init the key but not the member
```
DonalEvans, line 145: Another call to `getBytes()` that could be replaced with `Coder.stringToBytes()`.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZIncrByIntegrationTest.java
```diff
@@ -0,0 +1,290 @@
...
+
+    jedis.zincrby(KEY, increment, MEMBER);
+    assertThat(jedis.zscore(KEY, MEMBER)).isEqualTo(increment);
+  }
+
+  @Test
+  public void nullRedisSetShouldCreateNonExistentMember_withIncrementOfNegativeInfinity() {
+    final double increment = NEGATIVE_INFINITY;
+    jedis.zadd(KEY, 0.0, "something".getBytes()); // init the key but not the member
```
DonalEvans, line 154: Another call to `getBytes()` that could be replaced with `Coder.stringToBytes()`.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZIncrByIntegrationTest.java
```diff
@@ -0,0 +1,290 @@
...
+    jedis.sendCommand(STRING_KEY, Protocol.Command.ZINCRBY, STRING_KEY, "+Infinity", STRING_MEMBER);
+    assertThat(jedis.zscore(STRING_KEY, STRING_MEMBER)).isEqualTo(POSITIVE_INFINITY);
+
+    jedis.zadd(STRING_KEY, 1.0, STRING_MEMBER);
+    jedis.sendCommand(STRING_KEY, Protocol.Command.ZINCRBY, STRING_KEY, "-Infinity", STRING_MEMBER);
+    assertThat(jedis.zscore(STRING_KEY, STRING_MEMBER)).isEqualTo(NEGATIVE_INFINITY);
+  }
+
+  @Test
+  public void nullRedisSetShouldCreateNonExistentMember_withIncrementOfPositiveInfinity() {
```
DonalEvans, line 143: The name of this test is a little confusing to me. Is it intended to test that `incrby` called on a null set should behave as expected? Because if so, the call to `zadd` should be omitted, since that creates the set, meaning that the code in `NullRedisSortedSet.zincrby()` is never executed.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZIncrByIntegrationTest.java
```diff
@@ -0,0 +1,290 @@
...
+    final double increment = POSITIVE_INFINITY;
+    jedis.zadd(KEY, 0.0, "something".getBytes()); // init the key but not the member
+
+    jedis.zincrby(KEY, increment, MEMBER);
+    assertThat(jedis.zscore(KEY, MEMBER)).isEqualTo(increment);
+  }
+
+  @Test
+  public void nullRedisSetShouldCreateNonExistentMember_withIncrementOfNegativeInfinity() {
```
DonalEvans, line 152: See above comment on test name/intended behaviour related to `NullRedisSortedSet`.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZIncrByIntegrationTest.java
```diff
@@ -0,0 +1,290 @@
...
+  public void incrementingScoreOfNegativeInfinityByPositiveInfinity_throwsNaNError() {
+    jedis.zadd(KEY, NEGATIVE_INFINITY, MEMBER);
+
+    assertThatThrownBy(() -> jedis.zincrby(KEY, POSITIVE_INFINITY, MEMBER))
+        .hasMessageContaining(RedisConstants.ERROR_OPERATION_PRODUCED_NAN);
+  }
+
+  /************* key or member does not exist *************/
+  @Test
+  public void shouldCreateNewKey_whenIncrementedKeyDoesNotExist() {
```
DonalEvans, line 217: This test and the one below it are very similar in what they're testing to the `shouldCreateNonExistentKeyWithInfiniteScore_whenIncrementIs*` tests above. Would it be possible to create parameterized versions of the tests that pass +infinity, -infinity and a finite increment and assert the expected behaviour, rather than having three separate tests?

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZIncrByIntegrationTest.java
```diff
@@ -0,0 +1,290 @@
...
+          jedis.zincrby(STRING_KEY, increment, STRING_MEMBER);
+        },
+        (i) -> {
+          double increment = ThreadLocalRandom.current().nextDouble(-50, 50);
+          expectedValue.addAndGet(increment);
+          jedis.zincrby(STRING_KEY, increment, STRING_MEMBER);
+        }).run();
+
+    Offset<Double> offset = Offset.offset(Math.abs(expectedValue.get()) * 0.05);
+    assertThat(jedis.zscore(STRING_KEY, STRING_MEMBER)).isCloseTo(expectedValue.get(), offset);
```
DonalEvans, line 280: Is there a reason that this can't just be 
```
assertThat(jedis.zscore(STRING_KEY, STRING_MEMBER)).isEqualTo(expectedValue.get());
```
Given the test name, if the score is only close to the expected value rather than exactly equal, we didn't perform all the `zincrby`s and so the test should fail.

DonalEvans (changes requested):
A few code clean-up suggestions and some possible performance improvements.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZIncrByIntegrationTest.java
nonbinaryprogrammer: good point. but I'm wrapping up the zadd work that will enable these tests. I'd rather just wait for that since it'll read a little clearer

nonbinaryprogrammer: good point

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
nonbinaryprogrammer: making that test change you suggested showed that not doing that check is a bug. I implemented the check.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -228,6 +230,26 @@ private void validateScoreIsDouble(byte[] score) {
...
   }
 
+  byte[] zincrby(Region<RedisKey, RedisData> region, RedisKey key, byte[] increment,
+      byte[] member) {
+    byte[] score = members.get(member);
+    byte[] incr = Coder.doubleToBytes(
+        processIncrement(Coder.bytesToString(increment).toLowerCase()));
+
+    if (score != null) {
+      incr = Coder.doubleToBytes(Coder.bytesToDouble(score) + Coder.bytesToDouble(incr));
```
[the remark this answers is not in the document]
nonbinaryprogrammer: haha I just finished implementing that change when I read this. good catch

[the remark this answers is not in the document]
DonalEvans: Great minds think alike!

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/NullRedisSortedSet.java
```diff
@@ -49,4 +51,19 @@ long zadd(Region<RedisKey, RedisData> region, RedisKey key, List<byte[]> members
   byte[] zscore(byte[] member) {
     return null;
   }
+
+  @Override
+  byte[] zincrby(Region<RedisKey, RedisData> region, RedisKey key, byte[] increment,
+      byte[] member) {
+    byte[] incr = Coder.doubleToBytes(processIncrement(Coder.bytesToString(increment)));
```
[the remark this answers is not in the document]
nonbinaryprogrammer: the double that's returned from `processIncrement()` is used in `RedisSortedSet`. Returning a string here would require two implementations of the process method. and I'd be more inclined to make the method void, versus just passing through the string unchanged

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZIncrByIntegrationTest.java
nonbinaryprogrammer: Line 145 uses "something" as the member, then 147 uses the `MEMBER` constant. I have to do that first zadd to initialize the key and sorted set, then the second one creates the member

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZIncrByIntegrationTest.java
nonbinaryprogrammer: yeah I have no idea what that was about. good catch

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZIncrByIntegrationTest.java
nonbinaryprogrammer: done

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/NullRedisSortedSet.java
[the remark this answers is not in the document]
DonalEvans: The String might be changed by the method though, since there's some logic around the +/- infinity values that can take e.g. an input of `"+inf"` and turn it into a double value of `POSITIVE_INFINITY`, which then gets converted by `Coder.doubleToString()` into `"Infinity"`. Two implementations might be worth it, possibly "processIncrementToDouble" for the use in `RedisSortedSet` and "normalizeIncrement" or something for the use here, where we don't actually care about getting a double, just getting bytes that are consistent with the way we handle infinities and trailing `".0"` elsewhere.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZIncrByIntegrationTest.java
DonalEvans: Thanks for the clarification

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -228,6 +231,31 @@ private void validateScoreIsDouble(byte[] score) {
...
+    byte[] byteScore = members.get(member);
+    double incr = processIncrement(Coder.bytesToString(increment).toLowerCase());
+
+    if (byteScore != null) {
+      double score = Coder.bytesToDouble(byteScore);
+      if (Double.isInfinite(score) && Double.isInfinite(incr)) {
+        throw new NumberFormatException(RedisConstants.ERROR_OPERATION_PRODUCED_NAN);
+      }
+      incr = score + incr;
+    }
```
sabbey37, line 245: This would cause the addition of +inf + +inf (or -inf + -inf) to also result in the NAN error, but it shouldn't (should just be +inf or -inf respectively).  I'd prefer it if we checked the `incr` variable after doing the operation.  Double actually has an `isNan` method:
```
if(Double.isNaN(incr)) {
        throw new ArithmeticException(ERROR_OPERATION_PRODUCED_NAN);
      }
 ```

File: geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/ZIncrByExecutor.java [hunk 21]
```diff
@@ -0,0 +1,38 @@
...
+public class ZIncrByExecutor extends AbstractExecutor {
+  @Override
+  public RedisResponse executeCommand(Command command, ExecutionHandlerContext context) {
+    RedisSortedSetCommands redisSortedSetCommands = context.getRedisSortedSetCommands();
+    List<byte[]> commandElements = command.getProcessedCommand();
+
+    byte[] increment = commandElements.get(2);
+    byte[] member = commandElements.get(3);
+
+    byte[] retVal = redisSortedSetCommands.zincrby(command.getKey(), increment, member);
```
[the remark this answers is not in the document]
nonbinaryprogrammer: we return `Infinity` and `-Infinity` because that is the result of converting `POSITIVE_INFINITY` and `NEGATIVE_INFINITY` from double to byte[]. `jedis.zincrby` returns a double, so it will always end up being `Infinity` and `-Infinity` instead of `inf` and `-inf`.
I am also not sure if this is a problem, but it would require returning a byte[] from `jedis.zincrby`, which I don't think is a good idea

nonbinaryprogrammer:
@DonalEvans the stress new test failures are, for example: `expected <295.855182098051[84]> but was:<295.855182098051[96]>`. so that's why I had close to in that test. I guess I thought that that difference was a reasonable margin of error. but the 5% that I had in there is way too forgiving. So I think I'll add back the "close to" with a much smaller offset.

DonalEvans:
> @DonalEvans the stress new test failures are, for example: `expected <295.855182098051[84]> but was:<295.855182098051[96]>`. so that's why I had close to in that test. I guess I thought that that difference was a reasonable margin of error. but the 5% that I had in there is way too forgiving. So I think I'll add back the "close to" with a much smaller offset.

Hmm. Do we know why the numbers aren't matching up exactly? If it's just a test issue then using "close to" is fine as a workaround, but I think it would be good to know exactly what's happening first, in case this is pointing to a possible issue in the implementation.

nonbinaryprogrammer:
> > @DonalEvans the stress new test failures are, for example: `expected <295.855182098051[84]> but was:<295.855182098051[96]>`. so that's why I had close to in that test. I guess I thought that that difference was a reasonable margin of error. but the 5% that I had in there is way too forgiving. So I think I'll add back the "close to" with a much smaller offset.
> 
> Hmm. Do we know why the numbers aren't matching up exactly? If it's just a test issue then using "close to" is fine as a workaround, but I think it would be good to know exactly what's happening first, in case this is pointing to a possible issue in the implementation.

I think that the numbers don't quite match because floating point math is hard for computers. it's accurate to a lot of decimal places. also native redis doesn't pass the test for equality. but they both pass with an offset of 0.00000001. I think that's pretty reasonable.

DonalEvans:
> > > @DonalEvans the stress new test failures are, for example: `expected <295.855182098051[84]> but was:<295.855182098051[96]>`. so that's why I had close to in that test. I guess I thought that that difference was a reasonable margin of error. but the 5% that I had in there is way too forgiving. So I think I'll add back the "close to" with a much smaller offset.
> > 
> > 
> > Hmm. Do we know why the numbers aren't matching up exactly? If it's just a test issue then using "close to" is fine as a workaround, but I think it would be good to know exactly what's happening first, in case this is pointing to a possible issue in the implementation.
> 
> I think that the numbers don't quite match because floating point math is hard for computers. it's accurate to a lot of decimal places. also native redis doesn't pass the test for equality. but they both pass with an offset of 0.00000001. I think that's pretty reasonable.

Ah, if native Redis also fails with `isEqualTo()` then I'm not too concerned. An offset that small is pretty strong, so yeah, seems good.

DonalEvans (approved):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/ZIncrByExecutor.java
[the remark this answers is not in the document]
sabbey37: Discussed this with @DonalEvans, we created a separate ticket for it since the logic was already present in `Coder.java` before this PR: https://issues.apache.org/jira/browse/GEODE-9342

sabbey37 (approved):
Thank you for making all these changes!

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/NullRedisSortedSet.java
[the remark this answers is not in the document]
nonbinaryprogrammer: ok, I'll add that second implementation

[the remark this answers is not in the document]
jdeppe-pivotal: It seems confusing, at first glance, that we have 2 very similar implementations. I'd prefer to have a single one but we can adjust later if necessary. Bear in mind that the usage in `NullRedisSortedSet` is only going to be called once per key (on creation) but the other one will be called for all subsequent `zincrby`s on that key.

jdeppe-pivotal (approved):

PR #6535, opened 2021-05-27
GEODE-9331: remove the threadConnMaps ArrayList

dschneider-pivotal:
The threadConnMaps ArrayList has been removed which will improve performance of new threads
creating their first thread owned connection.
This removal also means that the HashMap is now
only referenced by a ThreadLocal so it no longer
is synchronized which simplifies the code.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/tcp/ConnectionTable.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/tcp/ConnectionTable.java
@@ -83,8 +83,7 @@ public class ConnectionTable {
   private final Map orderedConnectionMap = new ConcurrentHashMap();
 
   /**
-   * ordered connections local to this thread. Note that accesses to the resulting map must be
-   * synchronized because of static cleanup.
+   * ordered connections local to this thread.
    */
   static final ThreadLocal<Map> threadOrderedConnMap = new ThreadLocal<>();
 
--- a/geode-core/src/main/java/org/apache/geode/internal/tcp/ConnectionTable.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/tcp/ConnectionTable.java
@@ -19,8 +19,6 @@ import static java.util.concurrent.TimeUnit.SECONDS;
 
 import java.io.Closeable;
 import java.io.IOException;
-import java.lang.ref.Reference;
-import java.lang.ref.WeakReference;
 import java.net.InetAddress;
 import java.net.Socket;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


geode-core/src/main/java/org/apache/geode/internal/tcp/ConnectionTable.java
```diff
@@ -443,32 +434,18 @@ Connection getThreadOwnedConnection(InternalDistributedMember id, long startTime
...
-        }
-        // check for stale references and remove them.
-        for (Iterator it = threadConnMaps.iterator(); it.hasNext();) {
-          Reference r = (Reference) it.next();
-          if (r.get() == null) {
-            it.remove();
-          }
-        }
-        threadConnMaps.add(new WeakReference(m));
+      if (closed) {
```
pivotal-jbarrett, line 437: I think this check for closing and canceling was here because of the potential blocking at the now removed synchronized block. The calling method checks this before making the call. I don't think checking a few lines later makes any difference, especially since it can change to closed immediately after this check.

pivotal-jbarrett (changes requested):

geode-core/src/main/java/org/apache/geode/internal/tcp/ConnectionTable.java
dschneider-pivotal: nice catch; done

geode-core/src/main/java/org/apache/geode/internal/tcp/ConnectionTable.java
```diff
@@ -443,32 +434,14 @@ Connection getThreadOwnedConnection(InternalDistributedMember id, long startTime
...
-        // check for stale references and remove them.
-        for (Iterator it = threadConnMaps.iterator(); it.hasNext();) {
-          Reference r = (Reference) it.next();
-          if (r.get() == null) {
-            it.remove();
-          }
-        }
-        threadConnMaps.add(new WeakReference(m));
-      }
       threadOrderedConnMap.set(m);
```
pivotal-jbarrett, line 437: Now with all the other stuff gone we could replace this with:
```
static final ThreadLocal<Map<DistributedMember, InternalConnection>> threadOrderedConnMap =
      ThreadLocal.withInitial(HashMap::new);
```

pivotal-jbarrett (changes requested):

geode-core/src/main/java/org/apache/geode/internal/tcp/ConnectionTable.java
dschneider-pivotal: I liked this idea initially but then I found a couple of existing places in the code that we call get on this thread local to see if the current thread has a map and if it does to do something. If we make this change then these calls (they are cleanup methods) can end up creating a new HashMap for threads that didn't have one. So for this PR and like to hold off on making this change.

pivotal-jbarrett (approved):

dschneider-pivotal:
the dunit failure is the known issue: https://issues.apache.org/jira/browse/GEODE-9242

Bill (approved):
Good fix and simplification!

PR #6544, opened 2021-06-02
GEODE-9336: Implement ZADD option CH

luissson:
- Adds support for CH option of redis command ZADD

### For all changes:

### Note:

### DIFF
--- a/geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZAddIntegrationTest.java
+++ b/geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZAddIntegrationTest.java
@@ -169,7 +169,7 @@ public abstract class AbstractZAddIntegrationTest implements RedisIntegrationTes
 
     ZAddParams zAddParam = new ZAddParams();
     zAddParam.ch();
-    addCount = jedis.zadd(SORTED_SET_KEY, initMap);
+    addCount = jedis.zadd(SORTED_SET_KEY, initMap, zAddParam);
     assertThat(addCount).isEqualTo(0);
     assertThat(jedis.zcard(SORTED_SET_KEY)).isEqualTo(2 * INITIAL_MEMBER_COUNT);
   }
--- a/geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZAddIntegrationTest.java
+++ b/geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZAddIntegrationTest.java
@@ -161,6 +161,40 @@ public abstract class AbstractZAddIntegrationTest implements RedisIntegrationTes
   }
 
   @Test
+  public void zaddCountsNewMemberAndExistingMemberChanges_whenCHSpecified() {
+    Map<String, Double> initMap = makeMemberScoreMap(2 * INITIAL_MEMBER_COUNT, 0);
+    Long addCount = (Long) jedis.zadd(SORTED_SET_KEY, initMap);
+    assertThat(addCount).isEqualTo(2 * INITIAL_MEMBER_COUNT);
+    assertThat(jedis.zcard(SORTED_SET_KEY)).isEqualTo(2 * INITIAL_MEMBER_COUNT);
+
+    ZAddParams zAddParam = new ZAddParams();
+    zAddParam.ch();
+
+    addCount = (Long) jedis.zadd(SORTED_SET_KEY, initMap, zAddParam);
+    assertThat(addCount).isEqualTo(0);
+    assertThat(jedis.zcard(SORTED_SET_KEY)).isEqualTo(2 * INITIAL_MEMBER_COUNT);
+
+    Map<String, Double> updateMap = new HashMap<>();
+    for(String member : initMap.keySet()) {
+      Double score = initMap.get(member);
+      score++;
+      updateMap.put(member, score);
+    }
+    addCount = jedis.zadd(SORTED_SET_KEY, updateMap, zAddParam);
+    assertThat(addCount).isEqualTo(2 * INITIAL_MEMBER_COUNT);
+    assertThat(jedis.zcard(SORTED_SET_KEY)).isEqualTo(2 * INITIAL_MEMBER_COUNT);
+
+    final int newMemberCount = 5;
+    for(int i = 2 * INITIAL_MEMBER_COUNT; i < 2 * INITIAL_MEMBER_COUNT + newMemberCount; i++) {
+      updateMap.put("member_" + i, Double.valueOf((i) + ""));
+    }
+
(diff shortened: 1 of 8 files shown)


geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZAddIntegrationTest.java
```diff
@@ -160,6 +160,40 @@ public void zaddCountsOnlyNewMembers_givenMultipleCopiesOfTheSameMember_toAnExis
     assertThat(jedis.zscore(SORTED_SET_KEY, "member")).isEqualTo(3.0);
   }
 
+  @Test
+  public void zaddCountsNewMemberAndExistingMemberChanges_whenCHSpecified() {
```
DonalEvans, line 164: Would it be possible to break this test up into two separate cases, one which tests that the CH option works for updated members, and one which tests that it works with new members? At unit and integration test level, it's generally better to try and test one thing and one thing only per test. I suppose there could also be a third case testing that specifying CH does not return anything when no members are updated or added, but maybe that behaviour could be included in an existing test case for zadd with no new members specified.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -189,18 +189,20 @@ long zadd(Region<RedisKey, RedisData> region, RedisKey key, List<byte[]> members
...
-
       if (options.isNX() && members.containsKey(member)) {
         continue;
       }
       if (options.isXX() && !members.containsKey(member)) {
         continue;
       }
+      if (options.isCH() && members.containsKey(member)) {
+        changesCount = getChangesCount(changesCount, score, member);
+      }
```
DonalEvans, line 205: Since `memberAdd` returns the old value from the internal map, it might be more efficient to simply compare that to the new value and if it's different, increment `changesCount`. That way we're not doing a bunch of the same work that the map does as part of a `put()` call just to determine if the value was updated.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/BaseSetOptions.java
```diff
@@ -47,6 +47,10 @@ public boolean isXX() {
     return exists.equals(Exists.XX);
   }
 
+  public boolean isCH() {
+    return exists.equals(Exists.CH);
+  }
+
```
DonalEvans, line 53: I believe it's possible to set CH and one of NX or XX, so this won't work correctly, as it makes all three options mutually exclusive. I think a boolean flag for `isCH` would be better.

DonalEvans (changes requested):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/BaseSetOptions.java
luissson: Yup that's right! That didn't make sense to group the Exists options with a Change option. I've moved the CH check out of the exists options and on its own (just a bool now).

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -187,21 +187,22 @@ long zadd(Region<RedisKey, RedisData> region, RedisKey key, List<byte[]> members
...
         continue;
       }
       if (options.isXX() && !members.containsKey(member)) {
         continue;
       }
-      memberAdd(member, score);
+      byte[] oldScore = memberAdd(member, score);
+      if (options.isCH() && oldScore != null) {
+        changesCount = getChangesCount(changesCount, oldScore, score);
+      }
```
DonalEvans, line 205: This can be simplified to:
```
      if (options.isCH() && oldScore != null && !Arrays.equals(oldScore, score)) {
        changesCount++;
      }
```

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZAddIntegrationTest.java
```diff
@@ -152,18 +152,72 @@ public void zaddCountsOnlyNewMembers_givenMultipleCopiesOfTheSameMember() {
...
     assertThat(addCount).isEqualTo(1);
     jedis.sendCommand(SORTED_SET_KEY, Protocol.Command.ZADD, SORTED_SET_KEY,
         "1", "member", "2", "member", "3", "member");
     assertThat(jedis.zcard(SORTED_SET_KEY)).isEqualTo(2);
     assertThat(jedis.zscore(SORTED_SET_KEY, "member")).isEqualTo(3.0);
   }
 
+  @Test
+  public void zaddDoesNotCountExistingMembersWithoutChanges_whenCHSpecified() {
+    Map<String, Double> initMap = makeMemberScoreMap(2 * INITIAL_MEMBER_COUNT, 0);
```
DonalEvans, line 165: Can the initial size be just `INITIAL_MEMBER_COUNT` for consistency? It seems arbitrary to double it in these tests.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZAddIntegrationTest.java
```diff
@@ -152,18 +152,72 @@ public void zaddCountsOnlyNewMembers_givenMultipleCopiesOfTheSameMember() {
...
+    Long addCount = jedis.zadd(SORTED_SET_KEY, initMap);
+    assertThat(addCount).isEqualTo(2 * INITIAL_MEMBER_COUNT);
+    assertThat(jedis.zcard(SORTED_SET_KEY)).isEqualTo(2 * INITIAL_MEMBER_COUNT);
```
DonalEvans, line 168: These assertions aren't really necessary here, since that behaviour is tested elsewhere, so they can be removed to make the test simpler. The assertion on `zcard` at the end of the test is still good to have though, to confirm that the CH option isn't making `zadd` behave differently somehow.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/BaseSetOptions.java
```diff
@@ -47,6 +48,14 @@ public boolean isXX() {
     return exists.equals(Exists.XX);
   }
 
+  public void enableChange() {
```
DonalEvans, line 51: This might be better named "setCH" and changed to a true setter that takes an argument, for consistency with a "setX/isX" boolean naming convention.

DonalEvans (changes requested):
A few small clean-up requests.

sabbey37 (approved):
Looks good after Donal's requested changes are addressed!

DonalEvans (approved):

PR #6547, opened 2021-06-02
GEODE-9339: bump json-smart from 2.3 to 2.3.1

onichols-pivotal:
updates json-smart from 2.3 -> 2.3.1

json-smart is used by json-path, not directly by Geode

### DIFF
--- a/boms/geode-all-bom/src/test/resources/expected-pom.xml
+++ b/boms/geode-all-bom/src/test/resources/expected-pom.xml
@@ -273,6 +273,11 @@
         <version>5.8.0</version>
       </dependency>
       <dependency>
+        <groupId>net.minidev</groupId>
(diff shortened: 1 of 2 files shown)


dickcav (approved):

onichols-pivotal:
> This looks like an additional dependency, not a version bump. From

json-smart is used by json-path, not directly by Geode.  you can check `./gradlew :geode-management:dependencies` or the contents of any previous release to confirm that json-path brings in 2.3 by default.  since there is no newer version of json-path available that might bring in the updated json-smart, the only option is to override the version here.
Were this a new dependency, you would expect to also see it newly referenced in at least one gradle file...

rhoughton-pivot:
> Were this a new dependency, you would expect to also see it newly referenced in at least one gradle file...

Makes sense. Thanks!

PR #6549, opened 2021-06-02
GEODE-9335: Organize HitsMissesIntegrationTest

nonbinaryprogrammer:
put all the tests in the right sections of the test class to indicate
which commands of each type are supported or unsupported.

### For all changes:

### Note:

### DIFF
--- a/geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/server/AbstractHitsMissesIntegrationTest.java
+++ b/geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/server/AbstractHitsMissesIntegrationTest.java
@@ -224,6 +224,7 @@ public abstract class AbstractHitsMissesIntegrationTest implements RedisIntegrat
 
     runCommandAndAssertNoStatUpdates("key", (k) -> jedis.hmset(k, map));
   }
+
   @Test
   public void testHdel() {
     runCommandAndAssertNoStatUpdates("hash", (k, v) -> jedis.hdel(k, v));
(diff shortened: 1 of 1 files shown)


DonalEvans (approved):

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/server/AbstractHitsMissesIntegrationTest.java
```diff
@@ -65,8 +65,8 @@ public void teardown() {
   /***********************************************
    ************* Supported Commands **************
    **********************************************/
```
sabbey37, line 67: Do you think we should eliminate the Supported/Unsupported sections altogether?

sabbey37 (approved):

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/server/AbstractHitsMissesIntegrationTest.java
nonbinaryprogrammer: for what reason?

sabbey37: Since the goal is eventually to get rid of Supported/Unsupported, it seems unnecessary to maintain that distinction here.  I recognize we still have that distinction in `RedisCommandType` and the Supported/Unsupported JUnit test, but that is to make sure a user is not able to access those commands.  I don't really see a benefit to maintaining that distinction here.  But maybe since it's already like that we could just move the commands out from the `Unsupported` section over time.

PR #6561, opened 2021-06-03
GEODE-9346: When client received incorrect byte array of PdxType due …

gesterzhou:
…to broken socket, it should be retried

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/PdxMultiThreadQueryDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/PdxMultiThreadQueryDUnitTest.java
@@ -16,7 +16,6 @@ package org.apache.geode.cache.query.dunit;
 
 import static org.apache.geode.internal.Assert.fail;
 import static org.apache.geode.test.awaitility.GeodeAwaitility.await;
-import static org.apache.geode.util.internal.UncheckedUtils.uncheckedCast;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
@@ -167,7 +166,7 @@ public class PdxMultiThreadQueryDUnitTest extends PDXQueryTestBase {
             logger.info("### Executing Query on server: " + queryString[1]);
             Query query = remoteQueryService.newQuery(queryString[1]);
             SelectResults<TestObjectThrowsPdxSerializationException> selectResults =
-                uncheckedCast(query.execute());
+                (SelectResults) query.execute();
             assertThat(selectResults.size()).isEqualTo(numberOfEntries);
           });
     }
@@ -202,7 +201,7 @@ public class PdxMultiThreadQueryDUnitTest extends PDXQueryTestBase {
         logger.info("### Executing Query on server: " + queryString[1]);
         Query query = remoteQueryService.newQuery(queryString[1]);
         SelectResults<TestObjectThrowsPdxSerializationException> selectResults =
-            uncheckedCast(query.execute());
+            (SelectResults) query.execute();
         assertThat(selectResults.size()).isEqualTo(numberOfEntries);
         // the 2 failed try incremented numInstance
         assertThat(numberOfEntries + 2)
--- a/geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/PdxMultiThreadQueryDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/PdxMultiThreadQueryDUnitTest.java
@@ -30,6 +30,7 @@ import org.junit.Before;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
 
+import org.apache.geode.SerializationException;
 import org.apache.geode.cache.CacheException;
 import org.apache.geode.cache.Region;
 import org.apache.geode.cache.client.ClientCache;
@@ -37,6 +38,7 @@ import org.apache.geode.cache.client.ClientCacheFactory;
 import org.apache.geode.cache.client.ClientRegionShortcut;
 import org.apache.geode.cache.client.PoolManager;
 import org.apache.geode.cache.client.ServerConnectivityException;
+import org.apache.geode.cache.client.ServerOperationException;
 import org.apache.geode.cache.query.FunctionDomainException;
 import org.apache.geode.cache.query.NameResolutionException;
 import org.apache.geode.cache.query.Query;
@@ -56,6 +58,7 @@ import org.apache.geode.test.dunit.SerializableRunnableIF;
 import org.apache.geode.test.dunit.VM;
 import org.apache.geode.test.junit.categories.OQLQueryTest;
 import org.apache.geode.test.version.VersionManager;
+import org.apache.geode.util.internal.GeodeGlossary;
 
 @Category({OQLQueryTest.class})
 public class PdxMultiThreadQueryDUnitTest extends PDXQueryTestBase {
@@ -191,6 +194,8 @@ public class PdxMultiThreadQueryDUnitTest extends PDXQueryTestBase {
         new int[] {port0, port1, port2}, true);
 
     client.invoke(() -> {
+      System.setProperty(
+          GeodeGlossary.GEMFIRE_PREFIX + "enableQueryRetryOnPdxSerializationException", "true");
       try {
         TestObjectThrowsPdxSerializationException.throwExceptionOnDeserialization = true;
         QueryService remoteQueryService = (PoolManager.find(poolName)).getQueryService();
@@ -206,6 +211,78 @@ public class PdxMultiThreadQueryDUnitTest extends PDXQueryTestBase {
         assertThat(TestObjectThrowsPdxSerializationException.throwExceptionOnDeserialization)
             .isFalse();
         TestObjectThrowsPdxSerializationException.numInstance.set(0);
+        System.setProperty(
+            GeodeGlossary.GEMFIRE_PREFIX + "enableQueryRetryOnPdxSerializationException", "false");
+      }
+    });
+  }
+
+  @Test
+  public void testRetryNotEnabledBySystemProperty() throws CacheException {
+    // create pdx instance at servers
+    server0.invoke(() -> {
+      Region<Object, Object> region = getRootRegion().getSubregion(regionName);
+      for (int i = 0; i < numberOfEntries; i++) {
+        region.put("key-" + i, new TestObjectThrowsPdxSerializationException());
+      }
+    });
+
+    // Create client pool with only 2 servers to test that retry will not run forever
+    createPool(client, poolName, new String[] {hostName, hostName},
+        new int[] {port0, port1}, true);
+
+    client.invoke(() -> {
+      try {
+        // If the client did not explicitly specify GeodeGlossary.GEMFIRE_PREFIX +
+        // "enableQueryRetryOnPdxSerializationException" to true, retry will not be enabled
+        TestObjectThrowsPdxSerializationException.throwExceptionOnDeserialization = true;
+        QueryService remoteQueryService = (PoolManager.find(poolName)).getQueryService();
+        logger.info("### Executing Query on server: " + queryString[1]);
+        Query query = remoteQueryService.newQuery(queryString[1]);
+        assertThatThrownBy(query::execute).isInstanceOf(ServerOperationException.class)
+            .hasCauseInstanceOf(SerializationException.class);
+        assertThat(TestObjectThrowsPdxSerializationException.numInstance.get()).isEqualTo(1);
+        assertThat(TestObjectThrowsPdxSerializationException.throwExceptionOnDeserialization)
+            .isTrue();
+      } finally {
+        TestObjectThrowsPdxSerializationException.throwExceptionOnDeserialization = false;
+        TestObjectThrowsPdxSerializationException.numInstance.set(0);
+      }
+    });
+  }
+
+  @Test
+  public void testNotToRetryOnRegularSerializationException() throws CacheException {
+    // create pdx instance at servers
+    server0.invoke(() -> {
+      Region<Object, Object> region = getRootRegion().getSubregion(regionName);
+      for (int i = 0; i < numberOfEntries; i++) {
+        region.put("key-" + i, new TestObjectThrowsSerializationException());
+      }
+    });
+
+    // Create client pool with only 2 servers to test that retry will not run forever
+    createPool(client, poolName, new String[] {hostName, hostName},
+        new int[] {port0, port1}, true);
+
+    client.invoke(() -> {
+      try {
+        System.setProperty(
+            GeodeGlossary.GEMFIRE_PREFIX + "enableQueryRetryOnPdxSerializationException", "true");
+        TestObjectThrowsSerializationException.throwExceptionOnDeserialization = true;
+        QueryService remoteQueryService = (PoolManager.find(poolName)).getQueryService();
+        logger.info("### Executing Query on server: " + queryString[1]);
+        Query query = remoteQueryService.newQuery(queryString[1]);
+        assertThatThrownBy(query::execute).isInstanceOf(ServerOperationException.class)
+            .hasCauseInstanceOf(SerializationException.class);
+        assertThat(TestObjectThrowsSerializationException.numInstance.get()).isEqualTo(1);
+        assertThat(TestObjectThrowsSerializationException.throwExceptionOnDeserialization)
+            .isTrue();
+      } finally {
+        TestObjectThrowsSerializationException.throwExceptionOnDeserialization = false;
+        TestObjectThrowsSerializationException.numInstance.set(0);
+        System.setProperty(
+            GeodeGlossary.GEMFIRE_PREFIX + "enableQueryRetryOnPdxSerializationException", "false");
       }
     });
   }
@@ -225,15 +302,22 @@ public class PdxMultiThreadQueryDUnitTest extends PDXQueryTestBase {
         new int[] {port0, port1}, true);
 
     client.invoke(() -> {
-      TestObjectThrowsPdxSerializationException.throwExceptionOnDeserialization = true;
-      QueryService remoteQueryService = (PoolManager.find(poolName)).getQueryService();
-      logger.info("### Executing Query on server: " + queryString[1]);
-      Query query = remoteQueryService.newQuery(queryString[1]);
-      assertThatThrownBy(query::execute).isInstanceOf(ServerConnectivityException.class);
-      assertThat(TestObjectThrowsPdxSerializationException.numInstance.get()).isEqualTo(2);
-      assertThat(TestObjectThrowsPdxSerializationException.throwExceptionOnDeserialization)
-          .isFalse();
-      TestObjectThrowsPdxSerializationException.numInstance.set(0);
+      try {
+        System.setProperty(
+            GeodeGlossary.GEMFIRE_PREFIX + "enableQueryRetryOnPdxSerializationException", "true");
+        TestObjectThrowsPdxSerializationException.throwExceptionOnDeserialization = true;
+        QueryService remoteQueryService = (PoolManager.find(poolName)).getQueryService();
+        logger.info("### Executing Query on server: " + queryString[1]);
+        Query query = remoteQueryService.newQuery(queryString[1]);
+        assertThatThrownBy(query::execute).isInstanceOf(ServerConnectivityException.class);
+        assertThat(TestObjectThrowsPdxSerializationException.numInstance.get()).isEqualTo(2);
+        assertThat(TestObjectThrowsPdxSerializationException.throwExceptionOnDeserialization)
+            .isFalse();
+      } finally {
+        TestObjectThrowsPdxSerializationException.numInstance.set(0);
+        System.setProperty(
+            GeodeGlossary.GEMFIRE_PREFIX + "enableQueryRetryOnPdxSerializationException", "false");
+      }
     });
   }
 
@@ -255,7 +339,30 @@ public class PdxMultiThreadQueryDUnitTest extends PDXQueryTestBase {
           // after retried 2 servers, let the retry to 3rd server succeed
           throwExceptionOnDeserialization = false;
         }
-        throw new PdxSerializationException("Deserialization should not be happening in this VM");
+        throw new PdxSerializationException("Deserialization is expected to fail in this VM");
+      }
+    }
+  }
+
+  public static class TestObjectThrowsSerializationException implements PdxSerializable {
+    private static boolean throwExceptionOnDeserialization = false;
+    public static AtomicInteger numInstance = new AtomicInteger();
+
+    public TestObjectThrowsSerializationException() {
+      numInstance.incrementAndGet();
+    }
+
+    @Override
+    public void toData(PdxWriter writer) {}
+
+    @Override
+    public void fromData(PdxReader reader) {
+      if (throwExceptionOnDeserialization) {
+        if (numInstance.get() >= 2) {
+          // after retried 2 servers, let the retry to 3rd server succeed
+          throwExceptionOnDeserialization = false;
+        }
+        throw new SerializationException("Deserialization is expected to fail in this VM");
       }
     }
   }
--- a/geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/PdxMultiThreadQueryDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/PdxMultiThreadQueryDUnitTest.java
@@ -1,0 +1,271 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.apache.geode.cache.query.dunit;
+
+import static org.apache.geode.test.awaitility.GeodeAwaitility.await;
+import static org.junit.Assert.assertEquals;
+
+import java.util.Arrays;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import org.apache.logging.log4j.Logger;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+
+import org.apache.geode.cache.CacheException;
+import org.apache.geode.cache.Region;
+import org.apache.geode.cache.client.ClientCache;
+import org.apache.geode.cache.client.ClientCacheFactory;
+import org.apache.geode.cache.client.ClientRegionShortcut;
+import org.apache.geode.cache.client.PoolManager;
+import org.apache.geode.cache.query.FunctionDomainException;
+import org.apache.geode.cache.query.NameResolutionException;
+import org.apache.geode.cache.query.Query;
+import org.apache.geode.cache.query.QueryInvocationTargetException;
+import org.apache.geode.cache.query.QueryService;
+import org.apache.geode.cache.query.SelectResults;
+import org.apache.geode.cache.query.TypeMismatchException;
+import org.apache.geode.logging.internal.log4j.api.LogService;
+import org.apache.geode.pdx.PdxReader;
+import org.apache.geode.pdx.PdxSerializable;
+import org.apache.geode.pdx.PdxSerializationException;
+import org.apache.geode.pdx.PdxWriter;
+import org.apache.geode.test.dunit.AsyncInvocation;
+import org.apache.geode.test.dunit.Host;
+import org.apache.geode.test.dunit.NetworkUtils;
+import org.apache.geode.test.dunit.VM;
+import org.apache.geode.test.junit.categories.OQLQueryTest;
+import org.apache.geode.test.version.VersionManager;
+
+@Category({OQLQueryTest.class})
+public class PdxMultiThreadQueryDUnitTest extends PDXQueryTestBase {
+  public static final Logger logger = LogService.getLogger();
+
+  static final int numberOfEntries = 100;
+
+  public PdxMultiThreadQueryDUnitTest() {
+    super();
+  }
+
+  @Test
+  public void testClientServerQuery() throws CacheException {
+    final Host host = Host.getHost(0);
+    final String hostName = NetworkUtils.getServerHostName();
+
+    VM vm0 = host.getVM(VersionManager.CURRENT_VERSION, 0);
+    VM vm1 = host.getVM(VersionManager.CURRENT_VERSION, 1);
+    VM vm2 = host.getVM(VersionManager.CURRENT_VERSION, 2);
+    VM vm3 = host.getVM(VersionManager.CURRENT_VERSION, 3);
+
+    // Start servers
+    for (VM vm : Arrays.asList(vm0, vm1, vm2)) {
+      vm.invoke(() -> {
+        configAndStartBridgeServer();
+      });
+    }
+
+    // create pdx instance at servers
+    vm0.invoke(() -> {
+      Region region = getRootRegion().getSubregion(regionName);
+      for (int i = 0; i < numberOfEntries; i++) {
+        region.put("key-" + i, new TestObject(i, "vmware"));
+      }
+    });
+
+    final int port0 = vm0.invoke(() -> PdxQueryDUnitTest.getCacheServerPort());
+    final int port1 = vm1.invoke(() -> PdxQueryDUnitTest.getCacheServerPort());
+    final int port2 = vm2.invoke(() -> PdxQueryDUnitTest.getCacheServerPort());
+
+    // Create client region
+    vm3.invoke(() -> {
+      ClientCacheFactory cf = new ClientCacheFactory();
+      cf.addPoolServer(hostName, port1);
+      cf.addPoolServer(hostName, port2);
+      cf.addPoolServer(hostName, port0);
+      cf.setPdxReadSerialized(false);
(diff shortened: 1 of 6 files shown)


geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/PdxMultiThreadQueryDUnitTest.java
```diff
@@ -0,0 +1,271 @@
...
+    // Start servers
+    for (VM vm : Arrays.asList(vm0, vm1, vm2)) {
+      vm.invoke(() -> {
+        configAndStartBridgeServer();
+      });
+    }
+
+    // create pdx instance at servers
+    vm0.invoke(() -> {
+      Region region = getRootRegion().getSubregion(regionName);
```
DonalEvans, line 81: Warnings here can be fixed by using `Region<String, TestObject>`.

geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/PdxMultiThreadQueryDUnitTest.java
```diff
@@ -0,0 +1,271 @@
...
+    // Create client region
+    vm3.invoke(() -> {
+      ClientCacheFactory cf = new ClientCacheFactory();
+      cf.addPoolServer(hostName, port1);
+      cf.addPoolServer(hostName, port2);
+      cf.addPoolServer(hostName, port0);
+      cf.setPdxReadSerialized(false);
+      ClientCache clientCache = getClientCache(cf);
+      Region region =
+          clientCache.createClientRegionFactory(ClientRegionShortcut.PROXY).create(regionName);
```
DonalEvans, line 100: If using `Region<String, TestObject>` to fix warnings, this becomes:
```
      Region<String, TestObject> region = clientCache
          .<String, TestObject>createClientRegionFactory(ClientRegionShortcut.PROXY)
          .create(regionName);
```

geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/PdxMultiThreadQueryDUnitTest.java
```diff
@@ -0,0 +1,271 @@
...
+      logger.info("### Executing Query on server: " + queryString[1] + ": from client region: "
+          + region.getFullPath());
+      final int size = 100;
+      int[] array = new int[size];
+      for (int i = 0; i < size; i++) {
+        array[i] = i;
+      }
+      Arrays.stream(array).parallel().forEach(a -> {
+        try {
+          SelectResults rs2 = region.query(queryString[1]);
```
DonalEvans, line 111: To fix warnings about parameterized types, this can be `SelectResults<TestObject>`.

geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/PdxMultiThreadQueryDUnitTest.java
```diff
@@ -0,0 +1,271 @@
...
+          assertEquals(numberOfEntries, rs2.size());
+        } catch (FunctionDomainException e) {
+          e.printStackTrace();
+        } catch (TypeMismatchException e) {
+          e.printStackTrace();
+        } catch (NameResolutionException e) {
+          e.printStackTrace();
+        } catch (QueryInvocationTargetException e) {
+          e.printStackTrace();
```
DonalEvans, line 120: These catch blocks all have the same behaviour, so they can be condensed into one:
```
        } catch (FunctionDomainException | TypeMismatchException | NameResolutionException
            | QueryInvocationTargetException e) {
          e.printStackTrace();
        }
```

geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/PdxMultiThreadQueryDUnitTest.java
```diff
@@ -0,0 +1,271 @@
...
+        }
+      });
+      await().until(() -> TestObject.numInstance.get() == size * numberOfEntries);
+    });
+
+    this.closeClient(vm3);
+    this.closeClient(vm2);
+    this.closeClient(vm1);
+    this.closeClient(vm0);
```
DonalEvans, line 129: This step is repeated at the end of every test case. Could it be moved to an `@After` method?

geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/PdxMultiThreadQueryDUnitTest.java
```diff
@@ -0,0 +1,271 @@
...
+          clientCache.createClientRegionFactory(ClientRegionShortcut.PROXY).create(regionName);
+
+      logger.info("### Executing Query on server: " + queryString[1] + ": from client region: "
+          + region.getFullPath());
+      final int size = 100;
+      int[] array = new int[size];
+      for (int i = 0; i < size; i++) {
+        array[i] = i;
+      }
+      Arrays.stream(array).parallel().forEach(a -> {
```
DonalEvans, line 109: This can be simplified to:
```
IntStream.range(0, size).parallel().forEach(a -> {
```

geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/PdxMultiThreadQueryDUnitTest.java
```diff
@@ -0,0 +1,271 @@
...
+
+  @Test
+  public void testClientServerQuery() throws CacheException {
+    final Host host = Host.getHost(0);
+    final String hostName = NetworkUtils.getServerHostName();
+
+    VM vm0 = host.getVM(VersionManager.CURRENT_VERSION, 0);
+    VM vm1 = host.getVM(VersionManager.CURRENT_VERSION, 1);
+    VM vm2 = host.getVM(VersionManager.CURRENT_VERSION, 2);
+    VM vm3 = host.getVM(VersionManager.CURRENT_VERSION, 3);
```
Bill, line 70: Suggest giving these variables meaningful names. At least `vm3` could have a meaningful name like `client`.

geode-core/src/main/java/org/apache/geode/cache/client/internal/QueryOp.java
```diff
@@ -120,7 +122,11 @@ protected Object processResponse(Message msg) throws Exception {
           queryResult = resultPart.getObject();
         } catch (Exception e) {
           String s = "While deserializing " + getOpName() + " result";
-          exceptionRef[0] = new SerializationException(s, e);
+          if (e instanceof PdxSerializationException) {
+            exceptionRef[0] = new IOException(s, e);
```
Bill, line 126: Can you please explain the reasoning behind converting a `PdxSerializationException` to an `IOException`?

It is a bit puzzling looking at this code (without really understanding the calling context), why we would convert an exception with "serialization" in its name, to an `IOException` but at the same time convert all other exceptions to `SerializationException`.

I assume there is some sort of contract in play here. But there is no method-level comment describing that contract. Nor does the base class `AbstractOp` explain that contract.

It would be very helpful if that reasoning was captured in a method-level doc comment. If the reasoning applies only to `QueryOp` it could go on this method, but I suspect the reasoning actually applies at the level of `AbstractOp`, in which case, the documentation could go there.

geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/PdxMultiThreadQueryDUnitTest.java
```diff
@@ -0,0 +1,271 @@
...
+          + region.getFullPath());
+      final int size = 100;
+      int[] array = new int[size];
+      for (int i = 0; i < size; i++) {
+        array[i] = i;
+      }
+      Arrays.stream(array).parallel().forEach(a -> {
+        try {
+          SelectResults rs2 = region.query(queryString[1]);
+          assertEquals(numberOfEntries, rs2.size());
```
DonalEvans, line 112: Could this use AssertJ assertions instead of JUnit please:
```
assertThat(numberOfEntries).isEqualTo(rs2.size());
```
According to the Geode wiki page for writing DUnit tests, AssertJ is the preferred assertion library to use: https://cwiki.apache.org/confluence/display/GEODE/About+Distributed+Testing

geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/PdxMultiThreadQueryDUnitTest.java
```diff
@@ -0,0 +1,271 @@
...
+    // Start servers
+    for (VM vm : Arrays.asList(vm0, vm1, vm2)) {
+      vm.invoke(() -> {
+        configAndStartBridgeServer();
+      });
+    }
+
+    // create pdx instance at servers
+    vm0.invoke(() -> {
+      Region region = getRootRegion().getSubregion(regionName);
```
DonalEvans, line 152: To fix warnings, this can be `Region<String, TestObject>`

geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/PdxMultiThreadQueryDUnitTest.java
```diff
@@ -0,0 +1,271 @@
...
+      int[] array = new int[size];
+      for (int i = 0; i < size; i++) {
+        array[i] = i;
+      }
+      Arrays.stream(array).parallel().forEach(a -> {
+        try {
+          SelectResults rs2 = region.query(queryString[1]);
+          assertEquals(numberOfEntries, rs2.size());
+        } catch (FunctionDomainException e) {
+          e.printStackTrace();
```
Bill, line 114: If you expect to see these exceptions I recommend asserting that expectation.

If you do not expect to see them then I recommend eliminating these catch blocks and allowing the exceptions to cause the test to fail.

Oh but you can't do that because the functional interface doesn't allow exceptions!ions

Would it make sense to have the catch blocks collect these exceptions e.g. into a concurrent list, so the collection could be asserted empty before the test finishes?

geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/PdxMultiThreadQueryDUnitTest.java
DonalEvans, line 70: Could these be renamed to make it clearer which are the servers and which are the clients?

geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/PdxMultiThreadQueryDUnitTest.java
```diff
@@ -0,0 +1,271 @@
...
+    // Create client pool.
+    final String poolName = "testClientServerQueryPool";
+    createPool(vm3, poolName, new String[] {hostName}, new int[] {port1, port2, port0}, true);
+
+    final int size = 100;
+    AsyncInvocation[] asyncInvocationArray = new AsyncInvocation[size];
+    for (int i = 0; i < size; i++) {
+      asyncInvocationArray[i] =
+          vm3.invokeAsync(() -> {
+            ClientCache clientCache = getClientCache();
```
DonalEvans, line 171: This variable is not used and can be removed.

geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/PdxMultiThreadQueryDUnitTest.java
```diff
@@ -0,0 +1,271 @@
...
+            logger.info("### Executing Query on server: " + queryString[1]);
+            Query query = remoteQueryService.newQuery(queryString[1]);
+            SelectResults rs2 = (SelectResults) query.execute();
+            assertEquals(numberOfEntries, rs2.size());
+          });
+    }
+
+    for (int i = 0; i < size; i++) {
+      asyncInvocationArray[i].await();
+    }
```
DonalEvans, line 182: To clean up some warnings, a List can be used instead of an array here, which would turn this into:
```
    List<AsyncInvocation<Void>> asyncInvocationArray = new ArrayList<>();
    for (int i = 0; i < size; i++) {
      asyncInvocationArray.add(
          vm3.invokeAsync(() -> {
            QueryService remoteQueryService = (PoolManager.find(poolName)).getQueryService();
            logger.info("### Executing Query on server: " + queryString[1]);
            Query query = remoteQueryService.newQuery(queryString[1]);
            SelectResults<TestObject> rs2 = uncheckedCast(query.execute());
            assertThat(numberOfEntries).isEqualTo(rs2.size());
          }));
    }

    for (AsyncInvocation<Void> invocation : asyncInvocationArray) {
      invocation.await();
    }
```

geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/PdxMultiThreadQueryDUnitTest.java
```diff
@@ -0,0 +1,271 @@
...
+
+    this.closeClient(vm3);
+    this.closeClient(vm2);
+    this.closeClient(vm1);
+    this.closeClient(vm0);
+  }
+
+  @Test
+  public void testRetryWithPdxSerializationException()
+      throws CacheException, InterruptedException {
```
DonalEvans, line 195: `InterruptedException` is never thrown here, so that can be removed.

geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/PdxMultiThreadQueryDUnitTest.java
```diff
@@ -0,0 +1,271 @@
...
+      });
+    }
+
+    this.closeClient(vm3);
+    this.closeClient(vm2);
+    this.closeClient(vm1);
+    this.closeClient(vm0);
+  }
+
+  public static class TestPdxSerializable implements PdxSerializable {
```
Bill, line 252: The fact that `TestObject` and `TestObject2` are `PdxSerializable` makes me confused as to why this new object is called `TestPdxSerializable`.

I recommend a more descriptive class name like `TestObjectThrowsPdxSerializationException`.

geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/PdxMultiThreadQueryDUnitTest.java
```diff
@@ -0,0 +1,271 @@
...
+    // Start servers
+    for (VM vm : Arrays.asList(vm0, vm1, vm2)) {
+      vm.invoke(() -> {
+        configAndStartBridgeServer();
+      });
+    }
+
+    // create pdx instance at servers
+    vm0.invoke(() -> {
+      Region region = getRootRegion().getSubregion(regionName);
```
DonalEvans, line 213: To fix some warnings here, this can be `Region<String, TestPdxSerializable>`

geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/PdxMultiThreadQueryDUnitTest.java
```diff
@@ -0,0 +1,271 @@
...
+        region.put("key-" + i, new TestPdxSerializable());
+      }
+    });
+
+    final int port0 = vm0.invoke(() -> PdxQueryDUnitTest.getCacheServerPort());
+    final int port1 = vm1.invoke(() -> PdxQueryDUnitTest.getCacheServerPort());
+    final int port2 = vm2.invoke(() -> PdxQueryDUnitTest.getCacheServerPort());
+
+    // Create client pool.
+    final String poolName = "testClientServerQueryPool";
```
DonalEvans, line 224: This String is used in all the test cases. Could it be extracted to a constant?

geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/PdxMultiThreadQueryDUnitTest.java
```diff
@@ -0,0 +1,271 @@
...
+  public static final Logger logger = LogService.getLogger();
+
+  static final int numberOfEntries = 100;
+
+  public PdxMultiThreadQueryDUnitTest() {
+    super();
+  }
+
+  @Test
+  public void testClientServerQuery() throws CacheException {
```
Bill, line 63: This test does not seem to exercise any of the new code paths in `QueryOp.processResponse()`. So why was this test added?

Same question for next test: `testClientServerQueryUsingRemoteQueryService`?

Were we missing happy-path tests of `QueryOp.processResponse`?

geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/PdxMultiThreadQueryDUnitTest.java
```diff
@@ -0,0 +1,271 @@
...
+        new int[] {port0, port1, port2}, true);
+
+    try {
+      vm3.invoke(() -> {
+        TestPdxSerializable.throwExceptionOnDeserialization = true;
+        QueryService remoteQueryService = (PoolManager.find(poolName)).getQueryService();
+        logger.info("### Executing Query on server: " + queryString[1]);
+        Query query = remoteQueryService.newQuery(queryString[1]);
+        SelectResults rs2 = (SelectResults) query.execute();
+        assertEquals(numberOfEntries, rs2.size());
```
DonalEvans, line 235: Using AssertJ assertions, this would be `assertThat(numberOfEntries).isEqualTo(rs2.size());`

geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/PdxMultiThreadQueryDUnitTest.java
```diff
@@ -0,0 +1,271 @@
...
+    createPool(vm3, poolName, new String[] {hostName, hostName, hostName},
+        new int[] {port0, port1, port2}, true);
+
+    try {
+      vm3.invoke(() -> {
+        TestPdxSerializable.throwExceptionOnDeserialization = true;
+        QueryService remoteQueryService = (PoolManager.find(poolName)).getQueryService();
+        logger.info("### Executing Query on server: " + queryString[1]);
+        Query query = remoteQueryService.newQuery(queryString[1]);
+        SelectResults rs2 = (SelectResults) query.execute();
```
DonalEvans, line 234: Warnings here can be fixed by using: 
```
        SelectResults<TestPdxSerializable> rs2 = uncheckedCast(query.execute());
```

geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/PdxMultiThreadQueryDUnitTest.java
```diff
@@ -0,0 +1,271 @@
...
+        assertEquals(numberOfEntries, rs2.size());
+      });
+    } finally {
+      vm3.invoke(() -> {
+        assertEquals(false, TestPdxSerializable.throwExceptionOnDeserialization);
+        // the failed try will increment numInstance once
+        assertEquals(numberOfEntries + 1, TestPdxSerializable.numInstance.get());
```
DonalEvans, line 241: Using AssertJ assertions here, these would become:
```
        assertThat(TestPdxSerializable.throwExceptionOnDeserialization).isFalse();
        // the failed try will increment numInstance once
        assertThat(numberOfEntries + 1).isEqualTo(TestPdxSerializable.numInstance.get());
```

Bill (changes requested):
It's great to see this test code here!

I've got some questions below. At a minimum I think you need to capture some of your reasoning around exception processing, in javadocs for `QueryOp.processMessage()`.

geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/PdxMultiThreadQueryDUnitTest.java
```diff
@@ -0,0 +1,271 @@
...
+    final int port0 = vm0.invoke(() -> PdxQueryDUnitTest.getCacheServerPort());
+    final int port1 = vm1.invoke(() -> PdxQueryDUnitTest.getCacheServerPort());
+    final int port2 = vm2.invoke(() -> PdxQueryDUnitTest.getCacheServerPort());
+
+    // Create client pool.
+    final String poolName = "testClientServerQueryPool";
+    createPool(vm3, poolName, new String[] {hostName, hostName, hostName},
+        new int[] {port0, port1, port2}, true);
+
+    try {
```
DonalEvans, line 228: What's the purpose of the try/finally block here? When I remove it, the test still passes. If they're not necessary, the two `vm3.invoke()` calls can be combined to one, which simplifies things a bit.

DonalEvans (changes requested):
Some compiler warnings that can be fixed, and some suggestions for simplifying and cleaning up the tests.

geode-core/src/main/java/org/apache/geode/cache/client/internal/QueryOp.java
gesterzhou: The reason is explained in GEODE-9346 ticket. 

In short, IOException will allow the client to retry next server in the connection pool until exhaust all the servers (so it will not retry forever). 

Since the byte array of the pdxInstance is always the same at the server and other clients can deserialize it correctly, even this client can get the pdxInstance before and after the exception, retry is a feasible solution. 

I will add this comment in the code too.

geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/PdxMultiThreadQueryDUnitTest.java
gesterzhou: will change to server* and client

geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/PdxMultiThreadQueryDUnitTest.java
gesterzhou: these 2 tests are using customers' concurrent test cases try to reproduce the issue. 

The negative case is enough to prove it.

geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/PdxMultiThreadQueryDUnitTest.java
gesterzhou: that code was copied from customer's test code.

geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/PdxMultiThreadQueryDUnitTest.java
```diff
@@ -176,45 +177,78 @@ public void testClientServerQueryUsingRemoteQueryService()
...
+        TestObjectThrowsPdxSerializationException.throwExceptionOnDeserialization = true;
+        QueryService remoteQueryService = (PoolManager.find(poolName)).getQueryService();
+        logger.info("### Executing Query on server: " + queryString[1]);
+        Query query = remoteQueryService.newQuery(queryString[1]);
+        SelectResults<TestObjectThrowsPdxSerializationException> selectResults =
+            (SelectResults) query.execute();
+        fail("Expect ServerConnectivityException");
+      } catch (ServerConnectivityException sce) {
+        logger.info("Expect ServerConnectivityException after tried 2 servers");
+      } finally {
```
DonalEvans: Rather than using `fail()` and catching the expected exception here, it would be better to use:
```
        assertThatThrownBy(query::execute).isInstanceOf(ServerConnectivityException.class);
```
Doing this would mean that you don't need to use the try/finally here too, which would simplify the test slightly.

DonalEvans (changes requested):

geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/PdxMultiThreadQueryDUnitTest.java
gesterzhou: I added fail("Unexpected query exception:"+e.getMessage());

geode-core/src/distributedTest/java/org/apache/geode/cache/query/dunit/PdxMultiThreadQueryDUnitTest.java
gesterzhou: done

DonalEvans (approved):

echobravopapa (approved):

geode-core/src/main/java/org/apache/geode/cache/client/internal/QueryOp.java
Bill: Thank you for the extra comment in `processResponse()` @gesterzhou! I now understand that by throwing an `IOException` from any `AbstractOp.processResponse()` (like `QueryOp.processResponse()`) it will cause retrying by `OpExecutorImpl.execute()` and `OpExecutorImpl.handleException()`.

I understand your reasoning, that if those methods in `OpExecutorImpl` are retrying (on a single server, and across servers in the pool), and if _any_ server returns valid chunks (within the retry window), the client will eventually succeed.

What I do not understand, from this PR, and what I do not see represented in tests, is the mechanism by which the corruption of chunks occurs in the first place. You theorize that when a client communicates with a server that is crashing, that we might see corrupt chunks. If that's true, I'd ideally like test code to show that. And if that is infeasible, I'd like a more in-depth explanation of how that could happen.

Without that understanding, I am left wondering: what are the _other_ ways (besides your theorized crashing-server-sends-corrupt-chunk) that the second part of a chunk could be corrupted. And are there ways it could be corrupted on _all_ the servers—not just a single, crashing one.

And if a chunk were corrupted on all the servers, this PR would cause the client to retry across all servers, a total of `servers * retries` tries. If I adopt your theory then perhaps this is not deal. But if I do not adopt your theory, I have a fear that a corrupt chunk could appear on all servers frequently and this could cause a lot of extra client-server traffic.

If I believed your theory, I'd wonder: how (and where, in the code) can we end up with a chunk that is corrupt. That is where the bug seems to lie. The current PR papers over that problem by masking it as an `IOException`. It makes me think there was a deeper `IOException` we failed to handle properly.

gesterzhou: It won't corrupted on all servers. All the servers are hosting the same byte array. Other clients get get it correctly. Even the same client can get it correctly later. And the socket problem only happens occasionally, which is proven by the customer. 

The philosophy of the retry is to handle those occasionally network issue. Geode have retry logic for all client to server operation.

gesterzhou: The problem only happen once every 2 months.

Bill: If there was a bug (a different bug than GEODE-9346) in the code that serialized the PDX object then every copy would be corrupt right? I wonder how likely that sort of bug is and I wonder what the impact would be with the solution in this PR. In that case every server would be retried.

Another concern I have is I wonder if the fix in the PR might be insufficient to fix the problem when other kinds of serialization besides PDX are used (for domain objects).

In the PR scenario, a `PdxSerializationException` is encountered when deserializing domain objects returned by a query. I assume this is because the application happens to be doing this: https://geode.apache.org/docs/guide/12/developing/data_serialization/use_pdx_high_level_steps.html

If instead, the application was using standard Java serialization like https://geode.apache.org/docs/guide/12/developing/data_serialization/java_serialization.html would we expect to see some subclass of `ObjectStreamException` (which happens to be a subclass of `IOException`) instead? see reference here: https://facingissuesonit.com/2019/08/29/java-serialization-exception-handling/

If we believed that we might see these different exceptions for Java serializables then that might suggest that we should catch those too.

It seems to me that this bug (GEODE-9346) is caused by the `Message` class failing to do its job. That class is responsible for deserializing messages from sockets. `readHeaderAndBody()` is responsible for reading complete, correct messages. If the body contains a corrupt payload, doesn't that indicate we have a bug down in `readPayloadFields()`? The `Message` class uses `ByteBuffers`. Are those buffers shared? Because this this seems like the kind of problem that happens when shared buffers are corrupted.

gesterzhou: There're 4 questions. I will answer them one by one:

(1) If there's a real PdxSerialization corruption happened at server side caused by application (it did not happen so far though. Customer has verified all their server pdx types and they kept verifying from time to time). It's user's responsibility to fix their pdx serialzation issue in their application code. Our responsibility is to tell the user that their pdx type is wrong. That's why I add a dunit test to prove that in such a case, the client will fail after retry. My retry logic won't retry forever. It will try all servers then return with exception. You concern if we should fail earlier? I think we are fine to take a few more seconds to double check, because customer has to fix their pdx type in such a case, that will take hours or even days. 

(2) Customer is following the document you mentioned to use pdx type in correct way. That's why the client query correctly even before and after the occasional exception. If they are using the wrong way, they will encounter problem everyday. 

The application is querying the pdx object's byte array from server, then deserialized it at client. This is our default behavior. We offer another non-default usage is: set --read-serialized=true, then the client will not deserialize the byte array and wrap it as a PdxInstance object. It will delay the deserialization into their application logic field-by-field. But that's not our default setting.  

(3) If application is using standard Java serialization, then it will not fail with PdxSerializationException. It will not fail with ObjectStreamException either. It will fail with SerialzationException, which is handled in "else" (see my diff). I purposely did not change SerialzationException into IOException. The reason is as customer support ticket, we only fix what the issue is. If someday we encounter SerialzationException caused by socket corruption, we can fix it later, but not right now. If we concern about SerialzationException at client side should be retried, we can log a new geode ticket. 

(4) I suspected some socket corruption happened, which caused the Message packed wrong byte array. From log, every reproduce of the PdxSerializationException failed in different byte. 
But since it's very hard to reproduce this issue, we have no clue where the trigger is. Maybe in readHeaderAndBody() or readPayloadFields(). Maybe not. Again, if we think these codes have potential issues, we can log a new geode ticket.

geode-core/src/main/java/org/apache/geode/cache/client/internal/QueryOp.java
```diff
@@ -120,7 +122,16 @@ protected Object processResponse(Message msg) throws Exception {
           queryResult = resultPart.getObject();
         } catch (Exception e) {
           String s = "While deserializing " + getOpName() + " result";
-          exceptionRef[0] = new SerializationException(s, e);
+          if (e instanceof PdxSerializationException) {
```
dschneider-pivotal, line 125: The conversion of PdxSerializationException to an IOException while reading a query response seems like a hack to fix a problem that we do not yet understand. What do you think of it being disabled by default and only enabled if a system property "gemfire.enableQueryRetryOnPdxSerializationException" is true? This would give the current customer a workaround if they set the new system property. 

You should also document that this workaround will only work if the client is configured to connect to more than one cache server AND the pool's "retry-attempts" is -1 (the default which means try each server) or > 0. It is possible that if you closed the current connection and got a new connection to the same server and retried the query to it that it would also workaround this issue and it would not have the limitations of needing multiple servers and would not depend on the retry-attempts configuration.

Since we have not been able to reproduce this issue can we get the customer to test this workaround in their environment to confirm it will solve their issue?

gesterzhou: That could be a good idea to introduce the property.

gesterzhou: Our code will not auto retry the failed server. But I will put into the comments for the property.

geode-core/src/main/java/org/apache/geode/cache/client/internal/QueryOp.java
```diff
@@ -122,7 +123,18 @@ protected Object processResponse(Message msg) throws Exception {
...
+          // It only worked when the client is configured to connect to more than one cache server
+          // AND the pool's "retry-attempts" is -1 (the default which means try each server) or > 0.
+          // It is possible that if application closed the current connection and got a new
+          // connection to the same server and retried the query to it, that it would also
+          // workaround this issue and it would not have the limitations of needing multiple servers
+          // and would not depend on the retry-attempts configuration.
+          boolean enableQueryRetryOnPdxSerializationException = Boolean.getBoolean(
+              GeodeGlossary.GEMFIRE_PREFIX + "enableQueryRetryOnPdxSerializationException");
+          if (e instanceof PdxSerializationException
+              && enableQueryRetryOnPdxSerializationException) {
```
dschneider-pivotal, line 137: I think it would also be good to log an info message in this block that a pdx serialization exception is being converted to an ioexception. We don't expect this to happen too often but it would be good in the customer case if they no longer see failures to know that they did reproduce the problem and the workaround saved them.

gesterzhou: done

geode-core/src/main/java/org/apache/geode/cache/client/internal/QueryOp.java
```diff
@@ -120,7 +124,29 @@ protected Object processResponse(Message msg) throws Exception {
...
+            // IOException will allow the client to retry next server in the connection pool until
+            // exhausted all the servers (so it will not retry forever). Why retry:
+            // The byte array of the pdxInstance is always the same at the server. Other clients can
+            // get a correct one from query response message. Even this client can get it correctly
+            // before and after the PdxSerializationException.
+            exceptionRef[0] = new IOException(s, e);
+            LogService.getLogger().warn(
+                "Encountered unexpected PdxSerializationException, retrying on another server");
```
Bill, line 146: good to see the warning issued here

geode-core/src/main/java/org/apache/geode/cache/client/internal/QueryOp.java
```diff
@@ -120,7 +124,29 @@ protected Object processResponse(Message msg) throws Exception {
...
+
+          // Enable the workaround to convert PdxSerializationException into IOException to retry.
+          // It only worked when the client is configured to connect to more than one cache server
+          // AND the pool's "retry-attempts" is -1 (the default which means try each server) or > 0.
+          // It is possible that if application closed the current connection and got a new
+          // connection to the same server and retried the query to it, that it would also
+          // workaround this issue and it would not have the limitations of needing multiple servers
+          // and would not depend on the retry-attempts configuration.
+          boolean enableQueryRetryOnPdxSerializationException = Boolean.getBoolean(
+              GeodeGlossary.GEMFIRE_PREFIX + "enableQueryRetryOnPdxSerializationException");
```
Bill, line 136: good to see this functionality made conditional on the system property

Bill (approved):
Approved!

PR #6567, opened 2021-06-04
GEODE-9070: Improve ClientServerSessionCacheDUnitTest and add additional logging

sabbey37:
This issue occurs when the function tries to execute on a member that has just joined the cluster, but because the new view is not done processing by membership, the member is considered shunned and nonexistent.  This root issue will be fixed by [GEODE-9350](https://issues.apache.org/jira/projects/GEODE/issues/GEODE-9350).  

This PR just makes some small test improvements and adds additional logging to assist in diagnosing the problem if it occurs in the future.

### DIFF
--- a/extensions/geode-modules/src/distributedTest/java/org/apache/geode/modules/util/ClientServerSessionCacheDUnitTest.java
+++ b/extensions/geode-modules/src/distributedTest/java/org/apache/geode/modules/util/ClientServerSessionCacheDUnitTest.java
@@ -49,7 +49,6 @@ import org.apache.geode.internal.cache.execute.metrics.FunctionStats;
 import org.apache.geode.internal.cache.execute.metrics.FunctionStatsManager;
 import org.apache.geode.modules.session.catalina.ClientServerSessionCache;
 import org.apache.geode.modules.session.catalina.SessionManager;
-import org.apache.geode.test.dunit.DistributedTestUtils;
 import org.apache.geode.test.dunit.VM;
 import org.apache.geode.test.dunit.rules.CacheRule;
 import org.apache.geode.test.dunit.rules.ClientCacheRule;
@@ -57,9 +56,9 @@ import org.apache.geode.test.dunit.rules.DistributedRule;
 
 public class ClientServerSessionCacheDUnitTest implements Serializable {
   private static final String SESSION_REGION_NAME = RegionHelper.NAME + "_sessions";
-  private CacheRule cacheRule = new CacheRule();
-  private DistributedRule distributedRule = new DistributedRule();
-  private ClientCacheRule clientCacheRule = new ClientCacheRule();
+  private final CacheRule cacheRule = new CacheRule();
+  private final DistributedRule distributedRule = new DistributedRule();
+  private final ClientCacheRule clientCacheRule = new ClientCacheRule();
 
   @Rule
   public transient RuleChain ruleChain = RuleChain.outerRule(distributedRule)
@@ -83,7 +82,7 @@ public class ClientServerSessionCacheDUnitTest implements Serializable {
   }
 
   @Test
-  public void addServerToExistingClusterCreatesSessionRegion() {
+  public void addServerToExistingClusterCopiesSessionRegion() {
     final VM server0 = VM.getVM(0);
     final VM server1 = VM.getVM(1);
     final VM client = VM.getVM(2);
@@ -98,6 +97,34 @@ public class ClientServerSessionCacheDUnitTest implements Serializable {
 
     // Session region may be created asynchronously on the second server
     server1.invoke(() -> await().untilAsserted(this::validateServer));
+  }
+
+  @Test
+  public void addServerToExistingClusterDoesNotCopyPreCreatedSessionRegion() {
+    final VM server0 = VM.getVM(0);
+    final VM server1 = VM.getVM(1);
+    final VM client = VM.getVM(2);
+
+    server0.invoke(this::startCacheServer);
+
+
+    server0.invoke(this::createSessionRegion);
+
+
+    client.invoke(this::startClientSessionCache);
+    server1.invoke(this::startCacheServer);
+
+    server0.invoke(() -> await().untilAsserted(this::validateBootstrapped));
+    server1.invoke(() -> await().untilAsserted(this::validateBootstrapped));
+
+    // server1 should not have created the session region
+    // If the user precreated the region, they must manually
+    // create it on all servers
+    server1.invoke(() -> {
+      Region<Object, Object> region = cacheRule.getCache().getRegion(SESSION_REGION_NAME);
+      assertThat(region).isNull();
+    });
+
   }
 
   @Test
@@ -124,33 +151,6 @@ public class ClientServerSessionCacheDUnitTest implements Serializable {
 
     server0.invoke(this::validateServer);
     server1.invoke(this::validateServer);
-  }
-
-  @Test
-  public void preCreatedRegionIsNotCopiedToNewlyStartedServers() {
-    final VM server0 = VM.getVM(0);
-    final VM server1 = VM.getVM(1);
-    final VM client = VM.getVM(2);
-
-    server0.invoke(this::startCacheServer);
-
-
-    server0.invoke(this::createSessionRegion);
-
-
-    client.invoke(this::startClientSessionCache);
-    server1.invoke(this::startCacheServer);
-
-    server1.invoke(() -> await().untilAsserted(this::validateBootstrapped));
-
-    // server1 should not have created the session region
-    // If the user precreated the region, they must manually
-    // create it on all servers
-    server1.invoke(() -> {
-      Region<Object, Object> region = cacheRule.getCache().getRegion(SESSION_REGION_NAME);
-      assertThat(region).isNull();
-    });
-
   }
 
   @Test
@@ -189,7 +189,7 @@ public class ClientServerSessionCacheDUnitTest implements Serializable {
           .thenReturn(RegionShortcut.PARTITION_REDUNDANT.toString());
 
       final ClientCacheFactory clientCacheFactory = new ClientCacheFactory();
-      clientCacheFactory.addPoolLocator("localhost", DistributedTestUtils.getLocatorPort());
+      clientCacheFactory.addPoolLocator("localhost", DistributedRule.getLocatorPort());
       clientCacheFactory.setPoolSubscriptionEnabled(true);
       clientCacheRule.createClientCache(clientCacheFactory);
 
@@ -233,7 +233,7 @@ public class ClientServerSessionCacheDUnitTest implements Serializable {
     final RegionAttributes<Object, Object> expectedAttributes =
         cache.getRegionAttributes(RegionShortcut.PARTITION_REDUNDANT.toString());
 
-    final RegionAttributes attributes = region.getAttributes();
+    final RegionAttributes<String, HttpSession> attributes = region.getAttributes();
     assertThat(attributes.getScope()).isEqualTo(expectedAttributes.getScope());
     assertThat(attributes.getDataPolicy()).isEqualTo(expectedAttributes.getDataPolicy());
     assertThat(attributes.getPartitionAttributes())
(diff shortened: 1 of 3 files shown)


extensions/geode-modules/src/distributedTest/java/org/apache/geode/modules/util/ClientServerSessionCacheDUnitTest.java
```diff
@@ -153,6 +127,32 @@ public void preCreatedRegionIsNotCopiedToNewlyStartedServers() {
...
+  public void canPreCreateSessionRegionBeforeStartingClient() {
+    final VM server0 = VM.getVM(0);
+    final VM server1 = VM.getVM(1);
+    final VM client = VM.getVM(2);
+
+    server0.invoke(this::startCacheServer);
+    server1.invoke(this::startCacheServer);
+
+    server0.invoke(this::createSessionRegion);
+    server1.invoke(this::createSessionRegion);
```
Bill, line 148: Are these really happening _before_ the client is started (as the method name states)?

Without some explicit synchronization (maybe via the DUnit blackboard and gates) putting these two `invoke()` lines ahead of the `client.invoke()` line does little to ensure that the session region is created before starting the client.

extensions/geode-modules/src/distributedTest/java/org/apache/geode/modules/util/ClientServerSessionCacheDUnitTest.java
```diff
@@ -233,7 +233,7 @@ private void validateSessionRegion() {
     final RegionAttributes<Object, Object> expectedAttributes =
         cache.getRegionAttributes(RegionShortcut.PARTITION_REDUNDANT.toString());
 
-    final RegionAttributes attributes = region.getAttributes();
+    final RegionAttributes<String, HttpSession> attributes = region.getAttributes();
```
Bill, line 236: yay types!

geode-core/src/main/java/org/apache/geode/distributed/internal/ClusterDistributionManager.java
```diff
@@ -1082,7 +1082,16 @@ public InternalDistributedMember getId() {
...
+      Set<InternalDistributedMember> membersNotSent = sendMessage(msg);
+
+      if (membersNotSent != null && !membersNotSent.isEmpty()) {
+        logger.warn(
+            "The following members: {} were not properly sent the message: {}. If unhandled this may"
+                + " cause the result processor to hang while collecting results.",
+            membersNotSent, msg.getShortClassName());
+      }
+
+      return membersNotSent;
```
Bill, line 1094: I think the only case where `membersNotSent` will have a value is when `DirectChannel.send()` called from `DistributionImpl.directChannelSend()` generates a `ConnectExceptions` (sic).

In that case `DistributionImpl.directChannelSend()` is already logging warnings `"Failed to send message…"`.

Unless there is some other situation in which `membersNotSent` can contain members, I'd say maybe the message content here, belongs down in `DistributionImpl`.

Now the question is: does it make sense to have a message pertaining to "result processor" down in `ClusterDistributionManager.putOutgoing()` (or even down in `DistributionImpl`). Doing so couples (one of) them to "result processor". What is a "result processor" in this context? Is it a concept is part of the `ClusterDistributionManager` design?

extensions/geode-modules/src/distributedTest/java/org/apache/geode/modules/util/ClientServerSessionCacheDUnitTest.java
```diff
@@ -49,17 +49,16 @@
...
 import org.apache.geode.test.dunit.rules.DistributedRule;
 
 public class ClientServerSessionCacheDUnitTest implements Serializable {
   private static final String SESSION_REGION_NAME = RegionHelper.NAME + "_sessions";
-  private CacheRule cacheRule = new CacheRule();
-  private DistributedRule distributedRule = new DistributedRule();
-  private ClientCacheRule clientCacheRule = new ClientCacheRule();
+  private final CacheRule cacheRule = new CacheRule();
+  private final DistributedRule distributedRule = new DistributedRule();
+  private final ClientCacheRule clientCacheRule = new ClientCacheRule();
```
Bill, line 61: 👍

Bill (changes requested):
I've got some questions…

geode-core/src/main/java/org/apache/geode/distributed/internal/ClusterDistributionManager.java
sabbey37: Thanks for the observations, Bill! Right before the `"Failed to send message..."` log in `DistributionImpl.directChannelSend()`, we perform the following check:
```
 if (!membership.hasMember(member) || (th instanceof ShunnedMemberException)) {
          continue;
  }
```
which effective skips the log for those cases.  The `ShunnedMemberException` is the one we encountered here.  We could log before checking those cases or log something different within that if block and remove the log we added to `ClusterDistributionManager`.  What do you think is appropriate?

Regarding the "result processor"-specific message, that is a good question.  The "result processor" in this context is `ReplyProcessor21` (so perhaps that should say `reply processor` instead).  I think that's a good call out though, we probably shouldn't assume that if a message is sent there is always a reply processor involved.  It seems like we should remove this log anyway in favor of logging in `DistributionImpl` instead, so maybe we could just stick with the `"Failed to send message..." log there.

extensions/geode-modules/src/distributedTest/java/org/apache/geode/modules/util/ClientServerSessionCacheDUnitTest.java
sabbey37: That is a very valid point.  I am looking into it now.

geode-core/src/main/java/org/apache/geode/distributed/internal/ClusterDistributionManager.java
jdeppe-pivotal: The log message states - `If unhandled...` what does it then mean to 'handle' the situation?

sabbey37: Yeah, I think we will ditch that message in `ClusterDistributionManager` in favor of the one in `DistributionImpl`.  Do you have any opinions on that log message (if we should move the message up before the check or log a different message in that if block)?

sabbey37: I've removed the log from `ClusterDistributionManager` and moved the `DistributionImpl` log before the if block.  Let me know if y'all think this is the right approach. @jdeppe-pivotal @Bill

jdeppe-pivotal (approved):
This looks OK from my POV.

geode-core/src/main/java/org/apache/geode/distributed/internal/ClusterDistributionManager.java
sabbey37: ...I switched it to just do a debug log in that if block.  Let me know if y'all think that's ok or if we should move the fatal log up (it will also result in failing tests that need to be adjusted due to suspicious string failures).

kirklund (approved):

echobravopapa (approved):
Assuming Bill's comments are addressed.

geode-core/src/main/java/org/apache/geode/distributed/internal/DistributionImpl.java
```diff
@@ -387,14 +387,20 @@ private void checkCancelled() {
         Throwable th = it_causes.next();
 
         if (!membership.hasMember(member) || (th instanceof ShunnedMemberException)) {
+          if (logger.isDebugEnabled()) {
+            logger.debug(String.format("Failed to send message <%s> to member <%s> view, %s",
+                content, member, membership.getView()), th);
```
Bill, line 392: I can see some value in this message, but I have some suggestions.

First, think about the message content. As written, it doesn't tell why it failed. Why not be explicit with something like: "Failed to send message to member <%s> no longer in view."

Second, think about the frequency of this message. I think it'll happen whenever members are leaving (shutting down, crashing, cycling). Not all that frequent so I guess that's ok. OTOH message volume can be very high. So the product of those can be high.

Given we are adding potentially lots of logging, I don't think we need the view in this message. The person reading the logs can search backward for the previous view.

And do we really need the content in this message? I think that can be fairly large too.

Bill (changes requested):
With the test changes, and if there were no product changes at all, e.g. to `DistributionImpl`, do you feel like the test failure (due to GEODE-9350) is easier to understand now? If so I'll approve.

I'm reluctant to add a debugging line to `DistributionImpl` (see my comment below). If your test gives more understandable failures (due to GEODE-9350) with just the test changes I recommend eliding the product change.

geode-core/src/main/java/org/apache/geode/distributed/internal/DistributionImpl.java
sabbey37: The test changes in my opinion don't make the failure easier to understand or diagnose.  They do improve the overall test quality, ensure we are testing the right thing, and reduce flakiness to some degree.  It would be nice if there was a test that more directly tested the GEODE-9350 issue. 

If this issue or something similar happened again (in this test or somewhere else), it would be nice to have a debug log that might help us figure it out.  I've updated it to be more concise, but it does still log the exception.  We could potentially remove that or just get the message from it if it seems like too much:
```
logger.debug(String.format("Failed to send message <%s> to member <%s> no longer in view",
                content.getShortClassName(), member), th);
```

In the logs, this looks like:
```
Failed to send message <MemberFunctionStreamingMessage> to member <26461460c57d(172)<v2>:41002> no longer in view
org.apache.geode.distributed.internal.direct.ShunnedMemberException: Member is being shunned: 26461460c57d(172)<v2>:41002
        at org.apache.geode.distributed.internal.direct.DirectChannel.getConnections(DirectChannel.java:467)
        at org.apache.geode.distributed.internal.direct.DirectChannel.sendToMany(DirectChannel.java:281)
        at org.apache.geode.distributed.internal.direct.DirectChannel.sendToOne(DirectChannel.java:188)
        at org.apache.geode.distributed.internal.direct.DirectChannel.send(DirectChannel.java:537)
        at org.apache.geode.distributed.internal.DistributionImpl.directChannelSend(DistributionImpl.java:354)
        at org.apache.geode.distributed.internal.DistributionImpl.send(DistributionImpl.java:296)
        at org.apache.geode.distributed.internal.ClusterDistributionManager.sendViaMembershipManager(ClusterDistributionManager.java:2067)
        at org.apache.geode.distributed.internal.ClusterDistributionManager.sendOutgoing(ClusterDistributionManager.java:1982)
        at org.apache.geode.distributed.internal.ClusterDistributionManager.sendMessage(ClusterDistributionManager.java:2027)
        at org.apache.geode.distributed.internal.ClusterDistributionManager.putOutgoing(ClusterDistributionManager.java:1084)
        at org.apache.geode.internal.cache.execute.StreamingFunctionOperation.getFunctionResultFrom(StreamingFunctionOperation.java:113)
        at org.apache.geode.internal.cache.execute.MemberFunctionExecutor.executeFunction(MemberFunctionExecutor.java:149)
        at org.apache.geode.internal.cache.execute.MemberFunctionExecutor.executeFunction(MemberFunctionExecutor.java:191)
        at org.apache.geode.internal.cache.execute.AbstractExecution.execute(AbstractExecution.java:397)
        at org.apache.geode.internal.cache.execute.AbstractExecution.execute(AbstractExecution.java:402)
        at org.apache.geode.modules.util.BootstrappingFunction.bootstrapMember(BootstrappingFunction.java:170)
        at org.apache.geode.modules.util.BootstrappingFunction.memberJoined(BootstrappingFunction.java:240)
        at org.apache.geode.distributed.internal.ClusterDistributionManager$MemberJoinedEvent.handleEvent(ClusterDistributionManager.java:2497)
        at org.apache.geode.distributed.internal.ClusterDistributionManager$MemberEvent.handleEvent(ClusterDistributionManager.java:2450)
        at org.apache.geode.distributed.internal.ClusterDistributionManager$MemberEvent.handleEvent(ClusterDistributionManager.java:2439)
        at org.apache.geode.distributed.internal.ClusterDistributionManager.handleMemberEvent(ClusterDistributionManager.java:1405)
        at org.apache.geode.distributed.internal.ClusterDistributionManager.access$200(ClusterDistributionManager.java:109)
        at org.apache.geode.distributed.internal.ClusterDistributionManager$MemberEventInvoker.run(ClusterDistributionManager.java:1437)
        at java.base/java.lang.Thread.run(Thread.java:834)
```

We do check if a shutdown is in progress before getting to this if block and throw an exception there, so that case would not end up logging this message:
```
if (membership.shutdownInProgress()) {
        checkCancelled();
        throw new DistributedSystemDisconnectedException();
      }
```

We also catch `MembershipClosedException` and `DistributedSystemDisconnectedException` and throw exceptions for those (so they'd never log this message).

Bill (approved):
Thanks for honing that log message!

PR #6573, opened 2021-06-07
GEODE-9362: Implement zadd with INCR option.

pivotal-eshu:
* allow zadd to return byte[] if INCR option is specified.
  * fixes a race issue of ZAddExecutorState in ZAddExecutor
  * added an ignored dunit test case showing score incrmented multiple times due to retry

### For all changes:

### Note:

### DIFF
--- a/geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/sortedset/ZAddIncrOptionDUnitTest.java
+++ b/geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/sortedset/ZAddIncrOptionDUnitTest.java
@@ -22,6 +22,7 @@ import static org.assertj.core.api.Assertions.assertThat;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.Future;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.junit.After;
 import org.junit.Before;
@@ -55,7 +56,7 @@ public class ZAddIncrOptionDUnitTest {
   public RedisClusterStartupRule clusterStartUp = new RedisClusterStartupRule(4);
 
   private JedisCluster jedis;
-  private List<MemberVM> servers;
+  private List<MemberVM> servers = new ArrayList<>();
   private static final String sortedSetKey = "key";
   private final String baseMemberName = "member";
   private final int setSize = 1000;
@@ -67,13 +68,9 @@ public class ZAddIncrOptionDUnitTest {
   public void setup() {
     MemberVM locator = clusterStartUp.startLocatorVM(0);
     int locatorPort = locator.getPort();
-    MemberVM server1 = clusterStartUp.startRedisVM(1, locatorPort);
-    MemberVM server2 = clusterStartUp.startRedisVM(2, locatorPort);
-    MemberVM server3 = clusterStartUp.startRedisVM(3, locatorPort);
-    servers = new ArrayList<>();
-    servers.add(server1);
-    servers.add(server2);
-    servers.add(server3);
+    servers.add(clusterStartUp.startRedisVM(1, locatorPort));
+    servers.add(clusterStartUp.startRedisVM(2, locatorPort));
+    servers.add(clusterStartUp.startRedisVM(3, locatorPort));
 
     int redisServerPort = clusterStartUp.getRedisPort(1);
 
@@ -123,9 +120,9 @@ public class ZAddIncrOptionDUnitTest {
     verifyZScores();
   }
 
-  private void doZAddIncrForAllMembers(double increment1, double increment12) {
+  private void doZAddIncrForAllMembers(double increment1, double increment2) {
     for (int i = 0; i < setSize; i++) {
-      doZAddIncr(i, increment1, increment12, false);
+      doZAddIncr(i, increment1, increment2, false);
     }
   }
 
@@ -151,31 +148,31 @@ public class ZAddIncrOptionDUnitTest {
     return true;
   }
 
-  private boolean hitJedisClusterIssue2347 = false;
-
   @Test
   @Ignore("Fails due to GEODE-9310 and/or GEODE-9311")
   public void zAddWithIncrOptionCanIncrementScoresDuringPrimaryIsCrashed() throws Exception {
+    AtomicBoolean hitJedisClusterIssue2347 = new AtomicBoolean(false);
     doZAddIncrForAllMembers(increment1, increment1);
 
-    Future<Void> future1 = executor.submit(this::doZAddIncrForAllMembersDuringCrash);
+    Future<Void> future1 =
+        executor.submit(() -> doZAddIncrForAllMembersDuringCrash(hitJedisClusterIssue2347));
     Future<Void> future2 = executor.submit(() -> stopNodeWithPrimaryBucketOfTheKey(true));
 
     future1.get();
     future2.get();
 
-    if (!hitJedisClusterIssue2347) {
+    if (!hitJedisClusterIssue2347.get()) {
       doZCardWithRetries();
       verifyZScores();
     }
   }
 
-  private void doZAddIncrForAllMembersDuringCrash() {
+  private void doZAddIncrForAllMembersDuringCrash(AtomicBoolean hitJedisClusterIssue2347) {
     for (int i = 0; i < setSize; i++) {
       try {
         doZAddIncr(i, increment2, total, false);
       } catch (JedisClusterMaxAttemptsException ignore) {
-        hitJedisClusterIssue2347 = true;
+        hitJedisClusterIssue2347.set(true);
       }
     }
   }
--- a/geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/sortedset/ZAddIncrOptionDUnitTest.java
+++ b/geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/sortedset/ZAddIncrOptionDUnitTest.java
@@ -1,0 +1,216 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ *
+ */
+package org.apache.geode.redis.internal.executor.sortedset;
+
+import static org.apache.geode.test.dunit.rules.RedisClusterStartupRule.BIND_ADDRESS;
+import static org.apache.geode.test.dunit.rules.RedisClusterStartupRule.REDIS_CLIENT_TIMEOUT;
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.Future;
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Ignore;
+import org.junit.Rule;
+import org.junit.Test;
+import redis.clients.jedis.HostAndPort;
+import redis.clients.jedis.JedisCluster;
+import redis.clients.jedis.Protocol;
+import redis.clients.jedis.exceptions.JedisClusterMaxAttemptsException;
+
+import org.apache.geode.cache.Operation;
+import org.apache.geode.cache.Region;
+import org.apache.geode.internal.cache.PartitionedRegion;
+import org.apache.geode.internal.cache.PartitionedRegionHelper;
+import org.apache.geode.redis.ConcurrentLoopingThreads;
+import org.apache.geode.redis.internal.RegionProvider;
+import org.apache.geode.redis.internal.data.RedisData;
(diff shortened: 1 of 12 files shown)


geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/sortedset/ZAddIncrOptionDUnitTest.java
```diff
@@ -0,0 +1,216 @@
...
+    doZAddIncrForAllMembers(increment1, increment1);
+
+    stopNodeWithPrimaryBucketOfTheKey(false);
+
+    doZCardWithRetries();
+    doZAddIncrForAllMembers(increment2, total);
+    verifyZScores();
+  }
+
+  private void doZAddIncrForAllMembers(double increment1, double increment12) {
```
DonalEvans, line 126: Typo here, I think it should probably be `increment2` not `increment12`.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZAddIntegrationTest.java
```diff
@@ -301,6 +306,106 @@ public void shouldUpdateScore_whenSettingMemberThatAlreadyExists() {
...
+  public void zaddIncrOptionDoseNotAddANewMemberWithXXOption() {
+    Object result = jedis.sendCommand(SORTED_SET_KEY, Protocol.Command.ZADD, SORTED_SET_KEY, "XX",
+        incrOption, "1", member);
+    assertThat(result).isNull();
+    assertThat(jedis.zscore(SORTED_SET_KEY, member)).isNull();
+  }
+
+  private final double initial = 355.681000005;
+  private final double increment = 9554257.921450001;
+  private final double expected = initial + increment;
```
DonalEvans, line 333: Could these fields be moved to the top of the class for consistency with other classes please?

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/ZAddExecutor.java
```diff
@@ -46,11 +45,17 @@ public RedisResponse executeCommand(Command command, ExecutionHandlerContext con
...
 
-    long retVal = redisSortedSetCommands.zadd(command.getKey(),
+    Object retVal = redisSortedSetCommands.zadd(command.getKey(),
         new ArrayList<>(commandElements.subList(optionsFoundCount + 2, commandElements.size())),
         makeOptions(zAddExecutorState));
+    if (zAddExecutorState.incrFound) {
+      if (retVal == null) {
+        return RedisResponse.nil();
+      }
+      return RedisResponse.string((byte[]) retVal);
```
DonalEvans, line 55: The [documentation for ZADD with INCR option](https://redis.io/commands/ZADD#return-value) states that the response will be a Bulk string response, not a simple string. I think this should be therefore be `RedisResponse.bulkString(retVal);` The documentation for ZINCRBY also says the same, so I think we may need to change the `RedisResponse` type in the `ZIncrByExecutor` as well, since it's also currently just using a simple string response.

DonalEvans (changes requested):

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/sortedset/ZAddIncrOptionDUnitTest.java
```diff
@@ -0,0 +1,216 @@
...
+    }
+  }
+
+  private void doZCardWithRetries() {
+    int maxRetryAttempts = 10;
+    int retryAttempts = 0;
+    while (!zCardWithRetries(retryAttempts, maxRetryAttempts)) {
+      retryAttempts++;
+    }
+  }
```
sabbey37, line 138: It might be a bit more straightforward to take out the extra `zCardWithRetries` method and just do the following:
```
  private void doZCardWithRetries() {
    int maxRetryAttempts = 10;
    int retryAttempts = 0;

    while (retryAttempts < maxRetryAttempts) {
      long memberSize;
      try {
        memberSize = jedis.zcard(sortedSetKey);
      } catch (JedisClusterMaxAttemptsException e) {
          retryAttempts++;
          if(retryAttempts == maxRetryAttempts) {
            throw e;
          }
          continue;
      }
      assertThat(memberSize).isEqualTo(setSize);
      break;
    }
  }
 ```
 ...but that might just be a personal preference.

**Review by sabbey37 — COMMENTED** [review]
Looks great, Eric!  Just a few suggestions.

File: geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/ZAddExecutor.java (hunk 3 above)
Comment (sabbey37) [reply geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/ZAddExecutor.java:55]: Good catch!  I wonder if there are other places where we're returning simple string instead of bulk string...

File: geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZAddIntegrationTest.java [hunk 5]
```diff
@@ -35,12 +35,17 @@
...
 import org.apache.geode.redis.internal.RedisConstants;
+import org.apache.geode.redis.internal.netty.Coder;
 import org.apache.geode.test.awaitility.GeodeAwaitility;
 
 public abstract class AbstractZAddIntegrationTest implements RedisIntegrationTest {
   private JedisCluster jedis;
+  private final String member = "member";
+  private final String incrOption = "INCR";
+
   private static final int REDIS_CLIENT_TIMEOUT =
```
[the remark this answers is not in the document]
sabbey37: Oooh we also have the `REDIS_CLIENT_TIMEOUT` variable defined in `RedisClusterStartupRule` which is in the `commonTest` module... so we could potentially remove it from `RedisIntegrationTest` altogether.

On line 57 below this we use "localhost" when starting the JedisCluster, we could import the `BIND_ADDRESS` variable from `RedisClusterStartupRule` and use that instead.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/ZAddExecutor.java
DonalEvans: Do we know if there are any (possibly currently disabled) tests in the `redis-test-openjdk11` precheckin job that would fail due to wrong return format?

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZAddIntegrationTest.java
```diff
@@ -35,12 +35,17 @@
...
 import redis.clients.jedis.params.ZAddParams;
 
+import org.apache.geode.redis.ConcurrentLoopingThreads;
 import org.apache.geode.redis.RedisIntegrationTest;
 import org.apache.geode.redis.internal.RedisConstants;
+import org.apache.geode.redis.internal.netty.Coder;
 import org.apache.geode.test.awaitility.GeodeAwaitility;
 
 public abstract class AbstractZAddIntegrationTest implements RedisIntegrationTest {
   private JedisCluster jedis;
```
nonbinaryprogrammer, line 45: I'd like to see a test for when increment is infinity, and when incrementing a score of infinity, since those are both valid. It would need to test negative and positive infinity, and that you can pass in "Infinity", "+Infinity", "-Infinity", "infinity", "+infinity", "-infinity", "inf", "+inf", and "-inf" as increments.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZAddIntegrationTest.java
```diff
@@ -301,6 +306,106 @@ public void shouldUpdateScore_whenSettingMemberThatAlreadyExists() {
...
+        jedis.sendCommand(SORTED_SET_KEY, Protocol.Command.ZADD, SORTED_SET_KEY, incrOption,
+            Coder.doubleToString(increment), member);
+
+    assertThat(Coder.bytesToDouble((byte[]) result)).isEqualTo(expected);
+    result = jedis.zscore(SORTED_SET_KEY, member);
+    assertThat(result).isEqualTo(expected);
+  }
+
+  @Test
+  public void zaddIncrOptionCanIncrementScoreConcurrently() {
```
nonbinaryprogrammer, line 387: might be a good idea to add a test for concurrent modification of the same key over and over again, making sure that all the increments get applied correctly. I would do that with a random increment every time, then storing that to an atomic integer.

nonbinaryprogrammer (changes requested):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/ZAddExecutor.java
sabbey37: I'm not sure that there are.  I ran a few of the zset.tcl tests against the current implementation (`test "ZINCRBY return value"`, `test "ZINCRBY - increment and decrement`, and `test "ZINCRBY - can create a new sorted set`.. and a few others), but they all passed.

The difference is obvious on the CLI, here's a bulk string returned from native Redis:
```
127.0.0.1:6379> zadd a incr ch 1 a
"3"
```

Here's a simple string returned by us:
```
127.0.0.1:6380> zadd a incr 1 a
1
```
We could write tests at a lower level, but I'm not sure if it's super valuable given that clients are able to handle either one in the same fashion.

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/sortedset/ZAddIncrOptionDUnitTest.java
```diff
@@ -0,0 +1,216 @@
...
+      if (retries < maxRetries) {
+        return false;
+      }
+      throw e;
+    }
+    assertThat(memberSize).isEqualTo(setSize);
+    return true;
+  }
+
+  private boolean hitJedisClusterIssue2347 = false;
```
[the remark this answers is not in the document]
pivotal-eshu: The boolean value hitJedisClusterIssue2347 needs to be referenced by to the caller. There does not seems to be a clean way to do it.

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/sortedset/ZAddIncrOptionDUnitTest.java
```diff
@@ -0,0 +1,216 @@
...
+    verifyZScores();
+  }
+
+  private void verifyZScores() {
+    for (int i = 0; i < setSize; i++) {
+      assertThat(jedis.zscore(sortedSetKey, baseMemberName + i)).isEqualTo(total);
+    }
+  }
+
+  private void doZAddIncr(int i, double increment, double total, boolean isConcurrentExecution) {
```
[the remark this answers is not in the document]
pivotal-eshu: It is being called in doZAddIncrForAllMembers(increment1, increment1);

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/sortedset/ZAddIncrOptionDUnitTest.java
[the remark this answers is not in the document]
sabbey37: You could do it like this:

```
  public void zAddWithIncrOptionCanIncrementScoresDuringPrimaryIsCrashed() throws Exception {
    AtomicBoolean hitJedisClusterIssue2347 = new AtomicBoolean(false);
    doZAddIncrForAllMembers(increment1, increment1);

    Future<Void> future1 = executor.submit(() -> doZAddIncrForAllMembersDuringCrash(hitJedisClusterIssue2347));
    Future<Void> future2 = executor.submit(() -> stopNodeWithPrimaryBucketOfTheKey(true));

    future1.get();
    future2.get();

    if (!hitJedisClusterIssue2347.get()) {
      doZCardWithRetries();
      verifyZScores();
    }
  }

  private void doZAddIncrForAllMembersDuringCrash(AtomicBoolean hitJedisClusterIssue2347) {
    for (int i = 0; i < setSize; i++) {
      try {
        doZAddIncr(i, increment2, total, false);
      } catch (JedisClusterMaxAttemptsException ignore) {
        hitJedisClusterIssue2347.set(true);
      }
    }
  }
```

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZAddIntegrationTest.java
pivotal-eshu: As zadd with incr option internally calls the zincrby methods. I think the test coverage for zincrby should be good enough already.

DonalEvans (approved):

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZAddIntegrationTest.java
nonbinaryprogrammer: good point 👍

nonbinaryprogrammer (approved):
looks good

sabbey37 (approved):

PR #6588, opened 2021-06-08
GEODE-9366: Redis SynchronizedStripeExecutor can throw IndexOutOfBoundsException

jdeppe-pivotal:

### For all changes:

### Note:

### DIFF
--- a/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/SynchronizedStripedExecutor.java
+++ b/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/SynchronizedStripedExecutor.java
@@ -59,10 +59,13 @@ public class SynchronizedStripedExecutor implements StripedExecutor {
 
   private int getStripeIndex(Object stripeId) {
     int hash = stripeId.hashCode();
+
+    hash %= syncs.length;
     if (hash < 0) {
       hash = -hash;
     }
-    return hash % syncs.length;
+
+    return hash;
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/SynchronizedStripedExecutor.java
```diff
@@ -59,10 +59,13 @@ private Object getSync(Object stripeId) {
 
   private int getStripeIndex(Object stripeId) {
     int hash = stripeId.hashCode();
+
+    hash %= syncs.length;
     if (hash < 0) {
       hash = -hash;
```
sabbey37, line 65: Apologies if this is obvious and I've forgotten, but is there a reason we don't do `Math.abs(hash)` here?  Also, it's interesting the way `Math.abs()` handles the `INTEGER.MIN_VALUE`:
```
Note that if the argument is equal to the value of Integer.MIN_VALUE, the most negative representable int value, the result is that same value, which is negative.
```

sabbey37 (approved):
Thank you for finding and fixing this!

DonalEvans (approved):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/SynchronizedStripedExecutor.java
sabbey37: 😲

PR #6590, opened 2021-06-08
GEODE-9301: Implement accurate sizeable for RedisHash

DonalEvans:
- Extend Object2ObjectOpenCustomHashMap to allow direct access to the
 current capacity of the backing arrays
 - Use ReflectionSingleObjectSizer to calculate per-element overhead, as
 the previous implementation in SizeableObjectOpenCustomHashSet was
 effectively just reimplementing it
 - Calculate the overhead associated with resizing the backing arrays
 - Update tests to reflect the new behaviour

### For all changes:

### Note:


geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/collections/SizeableObject2ObjectOpenCustomHashMapWithCursor.java
```diff
@@ -29,52 +33,58 @@
  * The scan method provides the same guarantees as Redis's HSCAN, and in fact
  * uses the same algorithm.
  */
-public class Object2ObjectOpenCustomHashMapWithCursor<K, V>
-    extends Object2ObjectOpenCustomHashMap<K, V> {
+public class SizeableObject2ObjectOpenCustomHashMapWithCursor<K, V>
```
dschneider-pivotal, line 36: since our redis data classes always use the class with K=byte[] and V=byte[] it seems like you could make this class be SizeableByteArray2ByteArrayOpenCustomHashMapWithCursor extends Object2ObjectOpenCustomHashMap<byte[], byte[]>
You could also do this for the HashSet class.
Then in the size code instead of using the ReflectionSingleObjectSizer you can just compute the size of a byte[] which is very simple. ReflectionSingleObjectSizer has to do extra work to handle any type of array. It seems like this would give you fast sizing.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/collections/SizeableObject2ObjectOpenCustomHashMapWithCursor.java
```diff
@@ -29,52 +33,58 @@
...
+    extends Object2ObjectOpenCustomHashMap<K, V> implements Sizeable {
+
+  private static final long serialVersionUID = 9079713776660851891L;
+  public static final int BACKING_ARRAY_OVERHEAD_CONSTANT = 136;
+  public static final int BACKING_ARRAY_LENGTH_COEFFICIENT = 4;
+  private static final ReflectionSingleObjectSizer elementSizer =
+      ReflectionSingleObjectSizer.getInstance();
 
+  private int keysOverhead;
```
dschneider-pivotal, line 45: Currently these "Overhead" ints only include the size of the byte arrays. But if every time you changed the Overhead for a byte array you also included the BACKING_ARRAY_LENGTH_COEFFICIENT (by incing it or decing it) then you would not need to consult the length the arrays later when you compute the total size. Seems like it would make it a little faster.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/collections/SizeableObject2ObjectOpenCustomHashMapWithCursor.java
```diff
@@ -29,52 +33,58 @@
...
+  private int valuesOverhead;
```
dschneider-pivotal, line 46: I'm okay with having two overhead variables but I don't see any need for two. It looks like you could just have one "sizeInBytes" that tracks both keys and values. If you initialized it to BACKING_ARRAY_OVERHEAD_CONSTANT then the getSizeInBytes could just be "return sizeInBytes;"

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/collections/SizeableObject2ObjectOpenCustomHashMapWithCursor.java
DonalEvans: The length of the arrays isn't updated every time an element is added to the collection, only when there are sufficient elements to fill a percentage of it determined by the load factor, which then forces a resize of the backing arrays, so it's necessary to check the length of the arrays to determine the current overhead.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/collections/SizeableObject2ObjectOpenCustomHashMapWithCursor.java
DonalEvans: We currently have [a PR up](https://github.com/apache/geode/pull/6548) that uses a non-byte[] value in the CustomHashMap implementation, so I don't want to conflict with that by changing the underlying data structure.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/collections/SizeableObject2ObjectOpenCustomHashMapWithCursor.java
DonalEvans: Good call, I've combined the two.

kirklund (changes requested):

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/collections/SizeableObjectOpenCustomHashSetTest.java
```diff
@@ -99,13 +50,15 @@ public void backingArrayOverheadCalculationTest() {
 
       // Calculate the overhead associated only with the backing array
       backingArrayOverhead = sizer.sizeof(set) - set.getMemberOverhead();
+      int expected = BACKING_ARRAY_OVERHEAD_CONSTANT
+          + BACKING_ARRAY_LENGTH_COEFFICIENT * set.getBackingArrayLength();
       softly.assertThat(backingArrayOverhead)
```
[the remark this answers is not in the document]
DonalEvans: May I ask why? For this test in particular, the use of soft assertions is helpful in determining in one run what's going wrong with the sizing, as if we just failed on the first assertion, it wouldn't be clear if the size was off by a constant amount or if there was a compounding error in the calculation.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/collections/SizeableObject2ObjectOpenCustomHashMapWithCursor.java
```diff
@@ -183,6 +192,50 @@ private int hash(K key) {
...
+  @Override
+  public int getSizeInBytes() {
+    return arrayContentsOverhead + calculateBackingArraysOverhead();
+  }
+
+  @VisibleForTesting
+  int calculateBackingArraysOverhead() {
+    // This formula determined experimentally using tests.
+    return BACKING_ARRAY_OVERHEAD_CONSTANT
+        + BACKING_ARRAY_LENGTH_COEFFICIENT * (key.length + value.length);
```
dschneider-pivotal, line 226: you could use getTotalBackingArrayLength() here

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/collections/SizeableObjectOpenCustomHashSet.java
```diff
@@ -14,23 +14,23 @@
...
 import org.apache.geode.internal.size.Sizeable;
 
 public class SizeableObjectOpenCustomHashSet<K> extends ObjectOpenCustomHashSet<K>
     implements Sizeable {
   private static final long serialVersionUID = 9174920505089089517L;
-  public static final int MEMBER_OVERHEAD_CONSTANT = 16;
   public static final int BACKING_ARRAY_OVERHEAD_CONSTANT = 92;
   public static final int BACKING_ARRAY_LENGTH_COEFFICIENT = 4;
+  private static final ReflectionSingleObjectSizer elementSizer =
+      ReflectionSingleObjectSizer.getInstance();
```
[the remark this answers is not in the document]
DonalEvans: Done!

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/collections/SizeableObject2ObjectOpenCustomHashMapWithCursorTest.java
```diff
@@ -0,0 +1,349 @@
...
+
+import it.unimi.dsi.fastutil.Hash;
+import it.unimi.dsi.fastutil.bytes.ByteArrays;
+import org.assertj.core.api.SoftAssertions;
+import org.junit.Test;
+
+import org.apache.geode.internal.size.ReflectionObjectSizer;
+
+public class SizeableObject2ObjectOpenCustomHashMapWithCursorTest {
+  private final ReflectionObjectSizer sizer = ReflectionObjectSizer.getInstance();
```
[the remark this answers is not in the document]
DonalEvans: The `ReflectionObjectSizer` is a pre-existing singleton class that doesn't allow instantiation. Removing/refactoring that is outside the scope of this PR, so I'll leave this as is (but change the type to the `ObjectSizer` interface).

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/collections/SizeableObject2ObjectOpenCustomHashMapWithCursorTest.java
```diff
@@ -0,0 +1,349 @@
...
+    }
+  };
+
+  @Test
+  public void scanEntireMap_ReturnsExpectedElements() {
+    SizeableObject2ObjectOpenCustomHashMapWithCursor<Integer, String> map =
+        new SizeableObject2ObjectOpenCustomHashMapWithCursor<>(NATURAL_HASH);
+    IntStream.range(0, 10).forEach(i -> map.put(i, "value-" + i));
+
+    HashMap<Integer, String> scanned = new HashMap<>();
```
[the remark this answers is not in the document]
DonalEvans: Thanks for catching this. This code was just moved, not actually written for this PR, but I'll clean up the class while I'm here.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/collections/SizeableObject2ObjectOpenCustomHashMapWithCursor.java
DonalEvans: Done!

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/collections/SizeableObject2ObjectOpenCustomHashMapWithCursorTest.java
```diff
@@ -0,0 +1,349 @@
...
+    cursor = map.scan(cursor, 3, HashMap::put, scanned);
+    assertThat(scanned).hasSize(6);
+    cursor = map.scan(cursor, 4, HashMap::put, scanned);
+    assertThat(scanned).hasSize(10);
+    cursor = map.scan(cursor, 4, HashMap::put, scanned);
+    assertThat(scanned).hasSize(10);
+    assertThat(cursor).isEqualTo(0);
+
+    assertThat(scanned).isEqualTo(map);
+  }
```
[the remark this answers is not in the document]
DonalEvans: I didn't write this test so I could be misunderstanding what it's trying to do, but I don't think it's possible to have a separate test for each assertion, as the test is showing that successive calls to scan behave as expected. I've refactored the test slightly to simplify it to just two calls.

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisHashTest.java
```diff
@@ -377,8 +335,9 @@ public void testThatSizeIsUpdatedWhenUpdatingValue(final RedisKey key, final Str
 
     hash.hset(region, key, finalData, false);
 
-    int expectedUpdatedRedisHashSize = expectedRedisHash.getSizeInBytes()
-        + (Coder.stringToBytes(finalValue).length - Coder.stringToBytes(initialValue).length);
+    long expectedUpdatedRedisHashSize = expectedRedisHash.getSizeInBytes()
+        + (elementSizer.sizeof(Coder.stringToBytes(finalValue))
+            - elementSizer.sizeof(Coder.stringToBytes(initialValue)));
```
sabbey37, line 340: Apologies that I didn't notice this or comment on it when this test was created, but I was thinking of a few additional potential improvements:
- We could get the size of the RedisHash with the initial data via the elementSizer
- We could get the size of the RedisHash with the final data via elementSizer and assertThat is equal to `hash.getSizeInBytes()` [rather than manually subtracting the initial vs. final value]
- If we still want to assert on the initial/final value difference, we could subtract the elementSizer size of the initial data RedisHash from the `hash.getSizeInBytes()` and assert that is equal to the portion you've improved:
```
(elementSizer.sizeof(Coder.stringToBytes(finalValue))
            - elementSizer.sizeof(Coder.stringToBytes(initialValue)));
```

It just seems better to base our expected values in assertions from that elementSizer size as much as possible.

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisHashTest.java
```diff
@@ -470,7 +428,7 @@ public void putIfAbsent_shouldNotChangeSize_whenPuttingToExistingFields() {
   /******* remove *******/
   @Test
   public void sizeShouldDecrease_whenValueIsRemoved() {
-    final RedisKey key = new RedisKey("key".getBytes());
+    final RedisKey key = new RedisKey(Coder.stringToBytes("key"));
```
sabbey37, line 431: yay!

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisHashTest.java
```diff
@@ -494,42 +452,9 @@ public void sizeShouldDecrease_whenValueIsRemoved() {
 
     redisHash.hdel(region, key, dataToRemove);
 
-    int expectedSize = initialSize - RedisHash.HASH_MAP_VALUE_PAIR_OVERHEAD - field1.length;
-    Offset<Integer> offset = Offset.offset((int) round(expectedSize * 0.05));
+    long expectedSize = initialSize - (elementSizer.sizeof(field1) + elementSizer.sizeof(value1));
```
sabbey37, line 455: I feel similarly about this one...  It just seems like this exactly mimics our underlying implementation...which seems bad.  Could we get the initial size and assert that it's less than the final size... then maybe check that the elementSizer size of `redisHash` after `hdel` is equal to the `getSizeInBytes` of `redisHash`?

sabbey37 (approved):
This is awesome, Donal!  Just a couple potential test improvements.

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisHashTest.java
DonalEvans: Good call, I've updated the test.

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisHashTest.java
DonalEvans: Agreed, I tweaked the assertions.

DonalEvans:
Force pushing after rebasing on latest develop to resolve merge conflicts.

kirklund (approved):
Changes look good, thanks!

PR #6596, opened 2021-06-09
GEODE-9300: write-buffer-size value not used when creating disk store

alb3rtobr:
According to the Geode User Guide, it is possible to set the write buffer size by using the `--write-buffer-size` option when creating a disk store.
Nevertheless, setting a value for that parameter either by using gfsh, cache.xml or the `DiskStoreFactory.setWriteBuffer()` method has no effect.

The current implementation checks if the `WRITE_BUF_SIZE` system property is set, and if that is not the case, the default value of `32768` bytes is used.
This PR changes the behavior so first the disk store write buffer size is obtained (which default value is `32768`). Then, if the `WRITE_BUF_SIZE` system property is set, its value will take precedence over the disk store buffer size.

References:
* `create disk-store` command reference: https://geode.apache.org/docs/guide/113/tools_modules/gfsh/command-pages/create.html
* Question in the dev list about this issue: https://markmail.org/thread/d4ypp7qcw5lfqtf5

### DIFF
--- a/geode-core/src/integrationTest/java/org/apache/geode/internal/cache/OplogRVVJUnitTest.java
+++ b/geode-core/src/integrationTest/java/org/apache/geode/internal/cache/OplogRVVJUnitTest.java
@@ -40,6 +40,7 @@ import org.junit.rules.TemporaryFolder;
 import org.apache.geode.CancelCriterion;
 import org.apache.geode.Statistics;
 import org.apache.geode.StatisticsFactory;
+import org.apache.geode.cache.DiskStoreFactory;
 import org.apache.geode.internal.cache.DiskInitFile.DiskRegionFlag;
 import org.apache.geode.internal.cache.DiskStoreImpl.OplogEntryIdSet;
 import org.apache.geode.internal.cache.persistence.DiskRecoveryStore;
@@ -75,7 +76,7 @@ public class OplogRVVJUnitTest {
     final DiskStoreID m1 = DiskStoreID.random();
     final DiskStoreID m2 = DiskStoreID.random();
     final DiskRecoveryStore drs = mock(DiskRecoveryStore.class);
-    when(parent.getWriteBufferSize()).thenReturn(32768);
+    when(parent.getWriteBufferSize()).thenReturn(DiskStoreFactory.DEFAULT_WRITE_BUFFER_SIZE);
     when(df.getOrCreateCanonicalId(m1)).thenReturn(1);
     when(df.getOrCreateCanonicalId(m2)).thenReturn(2);
     when(df.getOrCreateCanonicalId(ownerId)).thenReturn(3);
--- a/geode-core/src/integrationTest/java/org/apache/geode/internal/cache/OplogRVVJUnitTest.java
+++ b/geode-core/src/integrationTest/java/org/apache/geode/internal/cache/OplogRVVJUnitTest.java
@@ -75,6 +75,7 @@ public class OplogRVVJUnitTest {
     final DiskStoreID m1 = DiskStoreID.random();
     final DiskStoreID m2 = DiskStoreID.random();
     final DiskRecoveryStore drs = mock(DiskRecoveryStore.class);
+    when(parent.getWriteBufferSize()).thenReturn(32768);
     when(df.getOrCreateCanonicalId(m1)).thenReturn(1);
     when(df.getOrCreateCanonicalId(m2)).thenReturn(2);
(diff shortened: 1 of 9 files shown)


DonalEvans:
This PR needs some tests to confirm that the changes are having the desired effects, i.e. that using the  `--write-buffer-size` gfsh option or the `DiskStoreFactory.setWriteBuffer()` method allows a value other than the default to be set, and that setting the system property `WRITE_BUF_SIZE` overrides these calls.

alb3rtobr:
> This PR needs some tests to confirm that the changes are having the desired effects, i.e. that using the `--write-buffer-size` gfsh option or the `DiskStoreFactory.setWriteBuffer()` method allows a value other than the default to be set, and that setting the system property `WRITE_BUF_SIZE` overrides these calls.

Ok, Im moving this PR back to draft while I implement the tests

geode-core/src/integrationTest/java/org/apache/geode/internal/cache/OplogRVVJUnitTest.java
```diff
@@ -75,6 +75,7 @@ public void testRecoverRVV() {
     final DiskStoreID m1 = DiskStoreID.random();
     final DiskStoreID m2 = DiskStoreID.random();
     final DiskRecoveryStore drs = mock(DiskRecoveryStore.class);
+    when(parent.getWriteBufferSize()).thenReturn(32768);
```
DonalEvans, line 78: Could hard-coded uses of this value (when it relates to Oplog buffer sizes) be replaced with references to `DiskStoreFactory.DEFAULT_WRITE_BUFFER_SIZE`?

DonalEvans (approved):
Thanks for adding the tests! I had a thought that might give slightly better test coverage. If the implementation is changed to the below, then you can test that the actual value returned from the system property check is returned from the method, instead of just `null` by mocking the `getWriteBufferSizeProperty()` method to return either null or a non-default value and then asserting that the expected value is returned from `getWriteBufferCapacity()`:
```
  @VisibleForTesting
  Integer getWriteBufferSizeProperty() {
    return Integer.getInteger("WRITE_BUF_SIZE");
  }

  @VisibleForTesting
  Integer getWriteBufferCapacity() {
    Integer writeBufferSizeProperty = getWriteBufferSizeProperty();
    if (writeBufferSizeProperty != null) {
      return writeBufferSizeProperty;
    }
    return getParent().getWriteBufferSize();
  }
```
Just a suggestion, but I think it makes the test a little more robust

alb3rtobr:
> 
> Just a suggestion, but I think it makes the test a little more robust

I agree, I think it looks better now. Thanks for the review!

nabarunnag (changes requested):
Hi Alberto, looking at the commit info you mentioned that the parameter set by gfsh for  --write-buffer-size is ignored by our current implenation, 
It will be nice to see the DUnit test with gfsh commands that fails to set the value and fails but with your fix, the correct value can be seen.

alb3rtobr:
thanks @nabarunnag , I have implemented a test that fails on develop branch and works after my change. In develop branch it shows that although you can set the write buffer size value and get its value successfully after the disk store creation, the oplog file write buffer has the default value (32768), so the introduced value is ignored.

geode-core/src/main/java/org/apache/geode/internal/cache/OverflowOplog.java
```diff
@@ -183,13 +188,27 @@ private void createCrf(OverflowOplog previous) throws IOException {
     this.stats.incOpenOplogs();
   }
 
-  private static ByteBuffer allocateWriteBuf(OverflowOplog previous) {
+  @VisibleForTesting
```
jinmeiliao, line 191: anyway we can avoid duplicating the code?

jinmeiliao (approved):

alb3rtobr:
Branch rebased to solve conflicts with develop branch.

alb3rtobr:
hi @nabarunnag , could you remove the request for changes if the tests I implemented are ok for you? Thanks

DonalEvans:
> hi @nabarunnag , could you remove the request for changes if the tests I implemented are ok for you? Thanks

Not to speak for Naba, but I think his comment was asking for tests that specifically used GFSH commands to set the --write-buffer-size value rather than API calls, as this was part of the problem described in the ticket.

nabarunnag (approved):

nabarunnag:
I agree with @DonalEvans

PR #6601, opened 2021-06-10
GEODE-9369: Command to copy region entries from a WAN site to another

albertogpz:
The command has been implemented as proposed in
https://cwiki.apache.org/confluence/display/GEODE/Geode+Command+to+replicate+region+data+from+one+site+to+another+connected+via+WAN
with some modifications with respect to the initial proposal.

The command will get the entries of a region
in a WAN site and will put them in batches
that will be sent by a gateway sender to a remote
WAN site.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/management/internal/i18n/CliStrings.java
+++ b/geode-core/src/main/java/org/apache/geode/management/internal/i18n/CliStrings.java
@@ -1459,11 +1459,11 @@ public class CliStrings {
   public static final String WAN_COPY_REGION__MSG__EXECUTION__FAILED =
       "Execution failed. Error: {0}";
   public static final String WAN_COPY_REGION__MSG__NO__CONNECTION__POOL =
-      "No connection pool available towards receiver";
+      "No connection pool available to receiver";
   public static final String WAN_COPY_REGION__MSG__COMMAND__NOT__SUPPORTED__AT__REMOTE__SITE =
       "Command not supported at remote site.";
   public static final String WAN_COPY_REGION__MSG__NO__CONNECTION =
-      "No connection available towards receiver after having copied {0} entries";
+      "No connection available to receiver after having copied {0} entries";
   public static final String WAN_COPY_REGION__MSG__ERROR__AFTER__HAVING__COPIED =
       "Error ({0}) in operation after having copied {1} entries";
   public static final String WAN_COPY_REGION__MSG__CANCELED__BEFORE__HAVING__COPIED =
--- a/geode-core/src/main/java/org/apache/geode/management/internal/i18n/CliStrings.java
+++ b/geode-core/src/main/java/org/apache/geode/management/internal/i18n/CliStrings.java
@@ -1426,6 +1426,51 @@ public class CliStrings {
   public static final String EXPORT_DATA__SUCCESS__MESSAGE =
       "Data successfully exported from region : {0} to file : {1} on host : {2}";
 
+  /* 'replicate region' command */
+  public static final String WAN_COPY_REGION = "wan-copy region";
+  public static final String WAN_COPY_REGION__HELP =
+      "Copy a region with a senderId via WAN replication";
+  public static final String WAN_COPY_REGION__REGION = "region";
+  public static final String WAN_COPY_REGION__REGION__HELP =
+      "Region from which data will be exported.";
+  public static final String WAN_COPY_REGION__SENDERID = "sender-id";
+  public static final String WAN_COPY_REGION__SENDERID__HELP =
+      "senderId to use to copy the region.";
+  public static final String WAN_COPY_REGION__MAXRATE = "max-rate";
+  public static final String WAN_COPY_REGION__MAXRATE__HELP =
+      "maximum rate for copying in events per second.";
+  public static final String WAN_COPY_REGION__BATCHSIZE = "batch-size";
+  public static final String WAN_COPY_REGION__BATCHSIZE__HELP =
+      "size for batches of events to be copied.";
+  public static final String WAN_COPY_REGION__CANCEL = "cancel";
+  public static final String WAN_COPY_REGION__CANCEL__HELP =
+      "cancel an ongoing wan-copy region command";
+  public static final String WAN_COPY_REGION__MSG__REGION__NOT__FOUND = "Region {0} not found";
+  public static final String WAN_COPY_REGION__MSG__REGION__NOT__USING_SENDER =
+      "Region {0} is not configured to use sender {1}";
+  public static final String WAN_COPY_REGION__MSG__SENDER__NOT__FOUND = "Sender {0} not found";
+  public static final String WAN_COPY_REGION__MSG__SENDER__SERIAL__AND__NOT__PRIMARY =
+      "Sender {0} is serial and not primary. 0 entries copied.";
+  public static final String WAN_COPY_REGION__MSG__SENDER__NOT__RUNNING =
+      "Sender {0} is not running";
+  public static final String WAN_COPY_REGION__MSG__EXECUTION__CANCELED = "Execution canceled";
+  public static final String WAN_COPY_REGION__MSG__EXECUTION__FAILED =
+      "Execution failed. Error: {0}";
+  public static final String WAN_COPY_REGION__MSG__ENTRIES__COPIED = "Entries copied: {0}";
+  public static final String WAN_COPY_REGION__MSG__NO__CONNECTION__POOL =
+      "No connection pool available towards receiver";
+  public static final String WAN_COPY_REGION__MSG__COMMAND__NOT__SUPPORTED__AT__REMOTE__SITE =
+      "Command not supported at remote site.";
+  public static final String WAN_COPY_REGION__MSG__NO__CONNECTION =
+      "No connection available towards receiver after having copied {0} entries";
+  public static final String WAN_COPY_REGION__MSG__ERROR__AFTER__HAVING__COPIED =
+      "Error ({0}) in operation after having copied {1} entries";
+  public static final String WAN_COPY_REGION__MSG__CANCELED__AFTER__HAVING__COPIED =
+      "Operation canceled after having copied {0} entries";
+  public static final String WAN_COPY_REGION__MSG__COPIED__ENTRIES = "Entries copied: {0}";
+  public static final String WAN_COPY_REGION__MSG__NO__RUNNING__COMMAND =
+      "No running command to be canceled for region {0} and sender {1}";
+
   /* export logs command */
   public static final String EXPORT_LOGS = "export logs";
   public static final String EXPORT_LOGS__HELP = "Export the log files for a member or members.";
--- a/geode-core/src/main/java/org/apache/geode/management/internal/i18n/CliStrings.java
+++ b/geode-core/src/main/java/org/apache/geode/management/internal/i18n/CliStrings.java
@@ -1454,9 +1454,10 @@ public class CliStrings {
   public static final String WAN_COPY_REGION__MSG__SENDER__NOT__RUNNING =
       "Sender {0} is not running";
   public static final String WAN_COPY_REGION__MSG__EXECUTION__CANCELED = "Execution canceled";
+  public static final String WAN_COPY_REGION__MSG__EXECUTIONS__CANCELED =
+      "Executions canceled: {0}";
   public static final String WAN_COPY_REGION__MSG__EXECUTION__FAILED =
       "Execution failed. Error: {0}";
-  public static final String WAN_COPY_REGION__MSG__ENTRIES__COPIED = "Entries copied: {0}";
   public static final String WAN_COPY_REGION__MSG__NO__CONNECTION__POOL =
       "No connection pool available towards receiver";
   public static final String WAN_COPY_REGION__MSG__COMMAND__NOT__SUPPORTED__AT__REMOTE__SITE =
@@ -1465,11 +1466,14 @@ public class CliStrings {
       "No connection available towards receiver after having copied {0} entries";
   public static final String WAN_COPY_REGION__MSG__ERROR__AFTER__HAVING__COPIED =
       "Error ({0}) in operation after having copied {1} entries";
-  public static final String WAN_COPY_REGION__MSG__CANCELED__AFTER__HAVING__COPIED =
-      "Operation canceled after having copied {0} entries";
+  public static final String WAN_COPY_REGION__MSG__CANCELED__BEFORE__HAVING__COPIED =
+      "Operation canceled before having copied all entries";
   public static final String WAN_COPY_REGION__MSG__COPIED__ENTRIES = "Entries copied: {0}";
   public static final String WAN_COPY_REGION__MSG__NO__RUNNING__COMMAND =
       "No running command to be canceled for region {0} and sender {1}";
+  public static final String WAN_COPY_REGION__MSG__ALREADY__RUNNING__COMMAND =
+      "There is already a command running for region {0} and sender {1}";
+
 
   /* export logs command */
   public static final String EXPORT_LOGS = "export logs";
--- a/geode-core/src/main/java/org/apache/geode/management/internal/i18n/CliStrings.java
+++ b/geode-core/src/main/java/org/apache/geode/management/internal/i18n/CliStrings.java
@@ -1438,10 +1438,10 @@ public class CliStrings {
       "Sender Id to use to copy the region.";
   public static final String WAN_COPY_REGION__MAXRATE = "max-rate";
   public static final String WAN_COPY_REGION__MAXRATE__HELP =
-      "Maximum rate for copying in events per second.";
+      "Maximum rate for copying in entries per second.";
   public static final String WAN_COPY_REGION__BATCHSIZE = "batch-size";
   public static final String WAN_COPY_REGION__BATCHSIZE__HELP =
-      "Size for batches of events to be copied.";
+      "Number of entries to be copied in each batch.";
   public static final String WAN_COPY_REGION__CANCEL = "cancel";
   public static final String WAN_COPY_REGION__CANCEL__HELP =
       "Cancel an ongoing wan-copy region command";
--- a/geode-core/src/main/java/org/apache/geode/management/internal/i18n/CliStrings.java
+++ b/geode-core/src/main/java/org/apache/geode/management/internal/i18n/CliStrings.java
@@ -1426,7 +1426,7 @@ public class CliStrings {
   public static final String EXPORT_DATA__SUCCESS__MESSAGE =
       "Data successfully exported from region : {0} to file : {1} on host : {2}";
 
-  /* 'replicate region' command */
+  /* 'wan-copy region' command */
   public static final String WAN_COPY_REGION = "wan-copy region";
   public static final String WAN_COPY_REGION__HELP =
       "Copy a region with a senderId via WAN replication";
@@ -1435,16 +1435,16 @@ public class CliStrings {
       "Region from which data will be exported.";
   public static final String WAN_COPY_REGION__SENDERID = "sender-id";
   public static final String WAN_COPY_REGION__SENDERID__HELP =
-      "senderId to use to copy the region.";
+      "Sender Id to use to copy the region.";
   public static final String WAN_COPY_REGION__MAXRATE = "max-rate";
   public static final String WAN_COPY_REGION__MAXRATE__HELP =
-      "maximum rate for copying in events per second.";
+      "Maximum rate for copying in events per second.";
   public static final String WAN_COPY_REGION__BATCHSIZE = "batch-size";
   public static final String WAN_COPY_REGION__BATCHSIZE__HELP =
-      "size for batches of events to be copied.";
+      "Size for batches of events to be copied.";
   public static final String WAN_COPY_REGION__CANCEL = "cancel";
   public static final String WAN_COPY_REGION__CANCEL__HELP =
-      "cancel an ongoing wan-copy region command";
+      "Cancel an ongoing wan-copy region command";
   public static final String WAN_COPY_REGION__MSG__REGION__NOT__FOUND = "Region {0} not found";
   public static final String WAN_COPY_REGION__MSG__REGION__NOT__USING_SENDER =
       "Region {0} is not configured to use sender {1}";
--- a/geode-docs/tools_modules/gfsh/command-pages/wan_copy_region.html.md.erb
+++ b/geode-docs/tools_modules/gfsh/command-pages/wan_copy_region.html.md.erb
@@ -37,7 +37,7 @@ The main uses of this command are:
 -   Adding a new WAN site to a <%=vars.product_name%> system in which the data in the new WAN site needs to be initially loaded from
     an existing WAN site.
 
-The execution of a currently running instance of this command my be stopped by using
+The execution of a currently running instance of this command may be stopped by using
 this same command with the `--cancel` option.
 
 **Requirements:**
@@ -76,7 +76,7 @@ table th:nth-of-type(3) {
 | &#8209;&#8209;sender-id| <em>Required</em>. The gateway sender to be used to copy the region entries. | |
 | &#8209;&#8209;max-rate| The maximum copy rate in entries per second. If the sender is parallel, the maximum rate limit is applied to each server hosting buckets for the region to be copied. | 0 (unlimited) |
 | &#8209;&#8209;batch-size| The size of the batches, in number of entries, to be used to copy the region entries. | 1000 |
-| &#8209;&#8209;cancel| Cancel a running `wan-copy region` command for the specified sender and region. |  |
+| &#8209;&#8209;cancel| Cancel a running `wan-copy region` command for the specified sender and region. If the `sender-id` and `region` passed are both "*", then all running `wan-copy region` commands will be canceled. |  |
 
 <span class="tablecap">Table 1. Copy Region Parameters</span>
 
@@ -89,6 +89,10 @@ wan-copy region --region=myRegion--sender-id=mySender --max-rate=1000 --batch-si
 
 ``` pre
 wan-copy region --region=/overload --sender-id=sender1 --cancel
+```
+
+``` pre
+wan-copy region --region=* --sender-id=* --cancel
 ```
 
 **Sample Output:**
@@ -173,5 +177,16 @@ gfsh> wan-copy region --region=/region1 --sender-id=sender1 --cancel
     server-sender2 | ERROR  | No running command to be canceled for region /region1 and sender sender1
     server-sender  | ERROR  | No running command to be canceled for region /region1 and sender sender1
     server-sender3 | ERROR  | No running command to be canceled for region /region1 and sender sender1
+```
+
+Example of cancel of all running `wan-copy region` commands:
+
+``` pre
+gfsh> wan-copy region --region=* --sender-id=* --cancel
+        Member     | Status | Message
+    -------------- | ------ | ------------------------------------------------------------------------------------
+    server-sender2 | OK     | Executions canceled: [(myRegion,mySender1), (myRegion,mySender)]
+    server-sender  | OK     | Executions canceled: [(myRegion,mySender1), (myRegion,mySender)]
+    server-sender3 | OK     | Executions canceled: [(myRegion,mySender1), (myRegion,mySender)]
 ```
 
--- a/geode-docs/tools_modules/gfsh/command-pages/wan_copy_region.html.md.erb
+++ b/geode-docs/tools_modules/gfsh/command-pages/wan_copy_region.html.md.erb
@@ -37,11 +37,11 @@ The main uses of this command are:
 -   Recovery of a WAN site after a disaster in which the failed site needs to be put into service again with
     the data from another WAN site.
 
--   Adding a new WAN site to a Geode system in which the data in the new WAN site needs to be initially loaded from
+-   Adding a new WAN site to a <%=vars.product_name%> system in which the data in the new WAN site needs to be initially loaded from
     an existing WAN site.
 
 The execution of a currently running instance of this command my be stopped by using
-this same command with the --cancel option.
+this same command with the `--cancel` option.
 
 **Availability:** Online. You must be connected in `gfsh` to a JMX Manager member to use this command.
 
--- a/geode-docs/tools_modules/gfsh/command-pages/wan_copy_region.html.md.erb
+++ b/geode-docs/tools_modules/gfsh/command-pages/wan_copy_region.html.md.erb
@@ -1,0 +1,171 @@
+---
+title:  wan-copy region
+---
+
+<!--
+Licensed to the Apache Software Foundation (ASF) under one or more
+contributor license agreements.  See the NOTICE file distributed with
+this work for additional information regarding copyright ownership.
+The ASF licenses this file to You under the Apache License, Version 2.0
+(the "License"); you may not use this file except in compliance with
+the License.  You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
+Copy the entries of a region in a WAN site onto the same region in another WAN site, using a gateway sender.
+
+This command requires that a gateway sender is configured and running on the source WAN site and that a gateway
+receiver is configured and running on the remote WAN site.
+The command also requires that the remote WAN site has the region onto which the data will be copied already created.
+Callbacks (cache listeners, replication to other WAN sites) will not be executed in the remote WAN site for the entries copied.
+
+This command copies region entries from a WAN site to another by putting them in batches of configurable
+size that are sent to the remote site by the selected gateway sender.
+
+The command allows to specify a maximum copy rate in order not to stress excessively the sending or receiving WAN sites.
+This rate is configured in entries per second.
+
+The main uses of this command are:
+
+-   Recovery of a WAN site after a disaster in which the failed site needs to be put into service again with
+    the data from another WAN site.
+
+-   Adding a new WAN site to a Geode system in which the data in the new WAN site needs to be initially loaded from
+    an existing WAN site.
+
+The execution of a currently running instance of this command my be stopped by using
+this same command with the --cancel option.
+
+**Availability:** Online. You must be connected in `gfsh` to a JMX Manager member to use this command.
+
+**Syntax:**
+
+``` pre
+wan-copy region --region=value --sender-id=value [--max-rate=value] [--batch-size=value] [--cancel(=value)?]
+```
+
+<a id="wan_copy_region_command_params"></a>
+<style>
+table th:first-of-type {
+    width: 20%;
+}
+table th:nth-of-type(2) {
+    width: 60%;
+}
+table th:nth-of-type(3) {
+    width: 20%;
+}
+</style>
+
+| Name | Description | Default Value |
+|------|-------------|---------------|
+| &#8209;&#8209;region| <em>Required</em>. The region for which the data is to be copied. | |
+| &#8209;&#8209;sender-id| <em>Required</em>. The gateway sender to be used to copy the region entries. | |
+| &#8209;&#8209;max-rate| The maximum copy rate in entries per second. If the sender is parallel, the maximum rate limit is applied to each server hosting buckets for the region to be copied. | 0 (unlimited) |
+| &#8209;&#8209;bath-size| The size of the batches to be used to copy the region entries. | 1000 |
+| &#8209;&#8209;cancel| If specified without a value or if true, it will cancel a running `wan-copy region` command for the passed sender and region. | false |
+
+<span class="tablecap">Table 1. Copy Region Parameters</span>
+
+
+**Example Commands:**
+
+``` pre
+wan-copy region --region=myRegion--sender-id=mySender --max-rate=1000 --batch-size=100
+```
+
+``` pre
+wan-copy region --region=/overload --sender-id=sender1 --cancel
+```
+
+**Sample Output:**
+
+``` pre
+gfsh>wan-copy region --region=/overload --sender-id=myParallelSender --max-rate=100 --batch-size=100
+        Member     | Status | Message
+    -------------- | ------ | -----------------------
+    server-sender  | OK     | Entries copied: 333
+    server-sender3 | OK     | Entries copied: 334
+    server-sender2 | OK     | Entries copied: 333
+
+```
+
+``` pre
+gfsh>wan-copy region --region=/overload --sender-id=mySerialSender --max-rate=100 --batch-size=100
+        Member     | Status | Message
+    -------------- | ------ | ----------------------------------------------------------------------
+    server-sender2 | OK     | Sender mySerialSender is serial and not primary. 0 entries copied.
+    server-sender3 | OK     | Sender mySerialSender is serial and not primary. 0 entries copied.
+    server-sender  | OK     | Entries copied: 1000
+```
+
+
+``` pre
+gfsh>wan-copy region --region=/overload --sender-id=sender1 --cancel
+        Member     | Status | Message
+    -------------- | ------ | ------------------
+    server-sender2 | OK     | Execution canceled
+    server-sender  | OK     | Execution canceled
+    server-sender3 | OK     | Execution canceled
+```
+
+``` pre
+gfsh>wan-copy region --region=myRegion --sender-id=myParallelSender --max-rate=100 --batch-size=10
+        Member     | Status | Message
+    -------------- | ------ | ------------------------------------------------------
+    server-sender2 | OK     | Operation canceled after having copied 10 entries
+    server-sender  | OK     | Operation canceled after having copied 10 entries
+    server-sender3 | OK     | Operation canceled after having copied 10 entries
+```
+
+``` pre
+gfsh>wan-copy region --region=myRegion --sender-id=myParallelSender --max-rate=100 --batch-size=10
+        Member     | Status | Message
+    -------------- | ------ | ----------------------------------------------------------------------
+    server-sender2 | OK     | Sender mySerialSender is serial and not primary. 0 entries copied.
+    server-sender3 | OK     | Sender mySerialSender is serial and not primary. 0 entries copied.
+    server-sender  | OK     | Operation canceled after having copied 4 entries
+```
+
+**Error Messages:**
+
+Example of `wan-copy region` with an invalid region:
+
+``` pre
+gfsh> wan-copy region --region=/regionX --sender-id=sender1
+        Member     | Status | Message
+    -------------- | ------ | -------------------------
+    server-sender  | ERROR  | Region /regionX not found
+    server-sender2 | ERROR  | Region /regionX not found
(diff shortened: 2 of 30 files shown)


geode-gfsh/src/test/resources/expected-pom.xml
```diff
@@ -16,6 +16,11 @@
   See the License for the specific language governing permissions and
   limitations under the License.
   -->
+  <!-- This module was also published with a richer model, Gradle metadata,  -->
+  <!-- which should be used instead. Do not delete the following line which  -->
+  <!-- is to indicate to Gradle or any Gradle module metadata file consumer  -->
+  <!-- that they should prefer consuming it instead. -->
+  <!-- do_not_remove: published-with-gradle-metadata -->
```
onichols-pivotal, line 23: not totally sure what the implications of adding this are, but I would suggest just leaving this file as it was

albertogpz: I do not remember how these lines ended up here. I think it was after adding a dependency to this file that finally was not needed.
I will remove them anyway.

rhoughton-pivot: Gradle started its own alternative to POM files for other Gradle projects. When Geode moved to Gradle 6.x, we started publishing them to our snapshot repository, and this XML tag in the POM is a bread-crumb to Gradle projects to look-aside for that file.

Problem is, that the Gradle metadata does not handle all our dependency declarations (I do not remember the actual failure at the moment) and downstream projects were having an issue when the pulled the metadata file instead of the POM. So we stopped publishing the metadata until there is a push to restore it.

mhansonp:
I am running some testing on this code. I will let people know once that finishes running.

mhansonp:
> I am running some testing on this code. I will let people know once that finishes running.

Nothing of note failed.

echobravopapa:
@albertogpz the RFC is still in discussion, please ping me when this is approved.

albertogpz:
> @albertogpz the RFC is still in discussion, please ping me when this is approved.

@echobravopapa I consider it in approved state as both Dan and Barry have given positive feedback.

mhansonp (commented):
It seems like the code should plan for a shutdown notification. What do you think?

geode-docs/tools_modules/gfsh/command-pages/wan_copy_region.html.md.erb
```diff
@@ -0,0 +1,171 @@
...
+The main uses of this command are:
+
+-   Recovery of a WAN site after a disaster in which the failed site needs to be put into service again with
+    the data from another WAN site.
+
+-   Adding a new WAN site to a Geode system in which the data in the new WAN site needs to be initially loaded from
+    an existing WAN site.
+
+The execution of a currently running instance of this command my be stopped by using
+this same command with the --cancel option.
```
alb3rtobr, line 44: Change format of "--cancel" to be `--cancel`

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/functions/WanCopyRegionFunction.java
```diff
@@ -0,0 +1,480 @@
...
+  }
+
+  final CliFunctionResult cancelWanCopyRegion(FunctionContext<Object[]> context,
+      String regionName, String senderId) {
+    boolean found = false;
+    String threadBaseName = getFunctionThreadName(regionName, senderId);
+    for (Thread t : Thread.getAllStackTraces().keySet()) {
+      if (t.getName().startsWith(threadBaseName)) {
+        found = true;
+        t.interrupt();
```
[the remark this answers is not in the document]
albertogpz: The problem is, where would that list of threads be kept? It could be in a static variable but it does not look like the cleanest solution.
If we open the door to using static variables to identify the command to cancel we could also explore a solution I considered in the past that did not require to run the copy region code in a new thread in order to be able to cancel it later. We could have a static variable that would be used to specify an execution to be canceled (identified by senderId and region) and the function could check after each batch is sent if that variable indicates that the function should be canceled.
This way the identification of the command to cancel would be more efficient than what we are doing currently and we would get rid of the new thread to execute the code.

Anyway, I am not sure which alternative is better.

alb3rtobr (commented):
Just a few minor comments about documentation format

geode-core/src/main/java/org/apache/geode/management/internal/i18n/CliStrings.java
```diff
@@ -1425,6 +1425,51 @@
...
+  /* 'replicate region' command */
+  public static final String WAN_COPY_REGION = "wan-copy region";
+  public static final String WAN_COPY_REGION__HELP =
+      "Copy a region with a senderId via WAN replication";
+  public static final String WAN_COPY_REGION__REGION = "region";
+  public static final String WAN_COPY_REGION__REGION__HELP =
+      "Region from which data will be exported.";
+  public static final String WAN_COPY_REGION__SENDERID = "sender-id";
+  public static final String WAN_COPY_REGION__SENDERID__HELP =
+      "senderId to use to copy the region.";
```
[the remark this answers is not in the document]
albertogpz: ok. I used `Sender Id` to be consistent with the rest of the file.

alb3rtobr (approved):

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/functions/WanCopyRegionFunction.java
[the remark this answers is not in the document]
mhansonp: It will not be the cleanest solution, the idea that you can start stuff from the command line that take time is kind of challenge. If a shutdown of the system happens or a close cache, you are going to want a graceful way to shut this down. From my perspective it would be a protected list of "Threads in progress". I  would be interested to hear what others think as I have not done something like this in this space either.

mhansonp:
> > It seems like the code should plan for a shutdown notification. What do you think?
> 
> What do yo mean by "plan for a shutdown notification"?

Basically what is being talked about in the previous comment.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/functions/WanCopyRegionFunction.java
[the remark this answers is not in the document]
mhansonp: @upthewaterspout  Do you have any thoughts?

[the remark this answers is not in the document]
albertogpz: @mhansonp  Who would be stopping these executions? If it must be done from outside the function when the system is going to be shutdown or the cache to be closed, we could just find all the threads whose name starts with a given prefix (wanCopyRegionFunctionThread_) and send the interrupt signal to them.
That way you would not need to keep a list of threads in progress for this command.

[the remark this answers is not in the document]
mhansonp: That is sufficient. I much rather bear the expense of a couple of Thread objects than a lookup like this, but I am fine with it being a lookup. Other than that, things look good.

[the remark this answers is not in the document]
mhansonp: If you can add someway to cleanly shut this down, I think you have all the necessary bits stuff and the code looks good.
I think it is important to consider what should happen with a partial transfer on the local and the remote when a shutdown or close happens...  I would suggest throwing away the data and finishing the command as a first step others might disagree.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/functions/WanCopyRegionFunction.java
```diff
@@ -0,0 +1,480 @@
...
+   */
+  void doPostSendBatchActions(long startTime, int copiedEntries, long maxRate)
+      throws InterruptedException {
+    if (Thread.currentThread().isInterrupted()) {
+      throw new InterruptedException();
+    }
+    long sleepMs = getTimeToSleep(startTime, copiedEntries, maxRate);
+    if (sleepMs > 0) {
+      logger.info("{}: Sleeping for {} ms to accommodate to requested maxRate",
+          this.getClass().getName(), sleepMs);
```
kirklund, line 419: `getClass().getName()` provides the fully qualified name (FQN) which is full.package.ClassName. `getClass().getSimpleName()` gives you just ClassName. Just making sure you know about that in case you didn't want the package.

geode-gfsh/src/test/java/org/apache/geode/management/internal/cli/functions/WanCopyRegionFunctionTest.java
```diff
@@ -0,0 +1,559 @@
...
+  private Clock clockMock;
+  private WanCopyRegionFunction.ThreadSleeper threadSleeperMock;
+  private InternalCache internalCacheMock;
+  private GatewaySender gatewaySenderMock;
+  private PoolImpl poolMock;
+  private Connection connectionMock;
+  private GatewaySenderEventDispatcher dispatcherMock;
+
+  @SuppressWarnings("unchecked")
+  private final FunctionContext<Object[]> contextMock = mock(FunctionContext.class);
```
kirklund, line 76: Another option to prevent these warnings is:
```
import static org.apache.geode.util.internal.UncheckedUtils.uncheckedCast;

private final FunctionContext<Object[]> contextMock = uncheckedCast(mock(FunctionContext.class));
```

geode-gfsh/src/integrationTest/java/org/apache/geode/management/internal/cli/GfshParserAutoCompletionIntegrationTest.java
```diff
@@ -410,7 +410,7 @@ public void testObtainHintForData() {
     String hintArgument = "data";
     String hintsProvided = gfshParserRule.getCommandManager().obtainHint(hintArgument);
     String[] hintsProvidedArray = hintsProvided.split(lineSeparator());
-    assertThat(hintsProvidedArray.length).isEqualTo(17);
+    assertThat(hintsProvidedArray.length).isEqualTo(18);
```
kirklund, line 413: Better syntax is:
```
assertThat(hintsProvidedArray).hasSize(18);
```
This way if it fails, `hintsProvidedArray` will be fully printed out instead of just the integer of its length.

kirklund (changes requested):

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/WanCopyRegionCommand.java
```diff
@@ -0,0 +1,82 @@
...
+          help = CliStrings.WAN_COPY_REGION__BATCHSIZE__HELP) int batchSize,
+      @CliOption(key = CliStrings.WAN_COPY_REGION__CANCEL,
+          unspecifiedDefaultValue = "false",
+          specifiedDefaultValue = "true",
+          help = CliStrings.WAN_COPY_REGION__CANCEL__HELP) boolean isCancel) {
+
+    authorize(Resource.DATA, Operation.WRITE, regionName);
+    final Object[] args = {regionName, senderId, isCancel, maxRate, batchSize};
+    ResultCollector<?, ?> rc =
+        executeFunction(wanCopyRegionFunction, args, getAllNormalMembers());
```
kirklund, line 62: Please try to avoid using abbreviations and acronyms. Just name vars like this `resultCollector` instead of `rc`.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/functions/WanCopyRegionFunction.java
```diff
@@ -0,0 +1,480 @@
...
+    }
+
+    @Override
+    public CliFunctionResult call() throws Exception {
+      return wanCopyRegion(context, region, sender, maxRate, batchSize);
+    }
+  }
+
+  @VisibleForTesting
+  CliFunctionResult wanCopyRegion(FunctionContext<Object[]> context, Region region,
```
kirklund, line 237: Can you please extract some smaller methods from this? There's actually a maximum byte length for the size of methods in order for the Jit to perform inlining and other optimizations. Basically, if you carve it up into smaller methods, then Jit can do inlining and it will actually perform better.

One way to do this is to extract blocks as methods to avoid having nested try-blocks and while-blocks. So if you see a nested while-block, pull that out as its own method. You can lean on IntelliJ refactoring to make it easy.

kirklund (changes requested):

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/functions/WanCopyRegionFunction.java
```diff
@@ -0,0 +1,480 @@
...
+  @VisibleForTesting
+  public void setClock(Clock clock) {
+    this.clock = clock;
+  }
+
+  @VisibleForTesting
+  public void setThreadSleeper(ThreadSleeper ts) {
+    this.threadSleeper = ts;
+  }
+
```
kirklund, line 120: It would be much better if you make these two fields `final` and delete these setters. Then create two constructors... one that's public with no-args for Geode to use and one that's only for testing that accepts args for both of these fields. Basically you want to always try to make fields `final` to avoid concurrency bugs and to improve performance. Here's an example along with proper constructor chaining:
```
private final Clock clock;
private final ThreadSleeper;

public WanCopyRegionFunction() {
  this(Clock.systemDefaultZone(), new ThreadSleeper());
}

@VisibleForTesting
WanCopyRegionFunction(Clock clock, ThreadSleeper threadSleeper) {
  this.clock = clock;
  this.threadSleeper = threadSleeper;
}
```

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/functions/WanCopyRegionFunction.java
```diff
@@ -0,0 +1,480 @@
...
+    }
+  }
+
+  private CliFunctionResult executeWanCopyRegionFunctionInNewThread(
+      FunctionContext<Object[]> context,
+      Region region, String regionName, GatewaySender sender, long maxRate, int batchSize)
+      throws InterruptedException, ExecutionException {
+    ExecutorService executor = LoggingExecutors
+        .newSingleThreadExecutor(getFunctionThreadName(regionName,
+            sender.getId()), true);
```
kirklund, line 206: Is this Executor providing the same thread that you later look up by name with the following?
```
for (Thread t : Thread.getAllStackTraces().keySet()) {
      if (t.getName().startsWith(threadBaseName)) {
        found = true;
        t.interrupt();
```
This is a really non-standard, unusual way to do any threading in Java. You should try to get it as close to standard use of threading as possible

Can you make `ExecutorService executor` a final field in this Function class and then just later invoke `executor.shutdownNow()`? `shutdownNow()` will interrupt any thread.

If you need ability to replace the ExecutorService at runtime in a thread-safe way then make use AtomicReference:
```
private final AtomicReference<ExecutorService> executor;
```
Then you can always do:
```
executor.set(LoggingExecutors.newSingleThreadExecutor(getFunctionThreadName(regionName, sender.getId()), true));
``` 
PS: "Java Concurrency In Practice" by Brian Goetz is the gold standard for Java threading

kirklund (changes requested):

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/functions/WanCopyRegionFunction.java
albertogpz: @kirklund Thanks for the advice. The problem here is that you could have several instances of this function running and you may want to stop one of them. How could you access the specific instance from a command in order to call the shutdown() method of the `ExecutorService`?
In order to use a more usual way to cancel the threads, I have added static fields to the function class with the `ExecutorService` to be used by any instance of this function and the current executions (`Futures`) so that, if one of them is required to be canceled, you can identify the execution by the senderId and region, get the Future and then invoke cancel on it.
You could also cancel all executions by invoking shutdownNow() on the executor as you suggested. Nevertheless, if you do that, you would not be able to create new executions.

I do not like using static variables but I have not found any better way.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/functions/WanCopyRegionFunction.java
albertogpz: I have done it at the cost of creating a new static class holding the connection status.
But I think it is cleaner now.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/functions/WanCopyRegionFunction.java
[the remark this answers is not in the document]
albertogpz: I have added a change that allows to stop all executions by passing "*" as region name and "*" as senderId to the function.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/functions/WanCopyRegionFunction.java
```diff
@@ -0,0 +1,557 @@
...
+              "Region destroyed",
+              copiedEntries));
+    }
+
+    return new CliFunctionResult(context.getMemberName(), CliFunctionResult.StatusState.OK,
+        CliStrings.format(CliStrings.WAN_COPY_REGION__MSG__COPIED__ENTRIES,
+            copiedEntries));
+  }
+
+  private Optional<CliFunctionResult> sendBatch(FunctionContext<Object[]> context,
```
gesterzhou, line 300: I concern the idea to introduce a whole set of sendBatch in multithread here in this function. Many of these code were duplicated from GatewaySender. This at least caused more efforts to maintain the logic. 

In my idea, we should use current LR.notifyGatewaySender to do it. In that case, we have had mature error handling.

gesterzhou (changes requested):

geode-core/src/main/java/org/apache/geode/management/internal/i18n/CliStrings.java
```diff
@@ -1425,6 +1425,55 @@
...
+  public static final String WAN_COPY_REGION__MSG__SENDER__SERIAL__AND__NOT__PRIMARY =
+      "Sender {0} is serial and not primary. 0 entries copied.";
+  public static final String WAN_COPY_REGION__MSG__SENDER__NOT__RUNNING =
+      "Sender {0} is not running";
+  public static final String WAN_COPY_REGION__MSG__EXECUTION__CANCELED = "Execution canceled";
+  public static final String WAN_COPY_REGION__MSG__EXECUTIONS__CANCELED =
+      "Executions canceled: {0}";
+  public static final String WAN_COPY_REGION__MSG__EXECUTION__FAILED =
+      "Execution failed. Error: {0}";
+  public static final String WAN_COPY_REGION__MSG__NO__CONNECTION__POOL =
```
boglesby, line 1460: Nitpick: Can you change 'towards' to 'to'?
```
No connection available towards receiver after having copied 147,000 entries
```

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/functions/WanCopyRegionFunction.java
```diff
@@ -0,0 +1,557 @@
...
+      try {
+        dispatcher.sendBatch(batch, connectionState.getConnection(),
+            connectionState.getSenderPool(), getAndIncrementBatchId(), true);
+        return Optional.empty();
+      } catch (BatchException70 e) {
+        return Optional.of(new CliFunctionResult(context.getMemberName(),
+            CliFunctionResult.StatusState.ERROR,
+            CliStrings.format(
+                CliStrings.WAN_COPY_REGION__MSG__ERROR__AFTER__HAVING__COPIED,
+                e.getMessage(), copiedEntries)));
```
boglesby, line 317: This result contains the message of the top-level BatchException70.

The BatchException70 contains a List of BatchException70s which are the real exceptions. It would be nice to log one of those, rather than this one.

Instead of:
```Executing - wan-copy region --region=/data_pr_parallel --sender-id=ny

 Member  | Status | Message
-------- | ------ | -----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
server-1 | ERROR  | Error (Exception occurred while processing a batch on the receiver running on DistributedSystem with Id: 1, DistributedMember on which the receiver is running: 10.166.144.101(server-1:87319)<v17>:41002) in operation after having copied 0 entries
```
It would be nice to see:
```
Executing - wan-copy region --region=/data_pr_parallel --sender-id=ny

 Member  | Status | Message
-------- | ------ | -----------------------------------------------------------------------------------------------
server-1 | ERROR  | Error (java.lang.IllegalStateException: test) in operation after having copied 21,000 entries
```

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/functions/WanCopyRegionFunction.java
```diff
@@ -0,0 +1,557 @@
...
+                copiedEntries)));
+      }
+      return Optional.empty();
+    }
+
+    void close() {
+      if (senderPool != null && connection != null) {
+        ((PooledConnection) connection).setShouldDestroy();
+        senderPool.returnConnection(connection);
+      }
```
boglesby, line 548: When the sending members finish copying the region, the receiving members log warnings like this for each connection:
```
[warn 2021/06/24 14:00:53.172 PDT ny-1 <ServerConnection on port 5025 Thread 2> tid=0x69] Server connection from [identity(10.166.144.101(ln-1:76734)<v1>:41003,connection=2; port=64429]: connection disconnect detected by EOF.
```
The `setShouldDestroy` and `returnConnection` call here cause the socket to be closed which causes that warning.

There might need to be another parameter or API that closes the connection (which closes the socket nicely) rather than destroying it.

I hacked `returnConnection` to see if it would get rid of the warn message (it does) but its not a valid fix:
```
private void returnConnection(Connection connection, boolean accessed, boolean addLast) {
  ...
  if (pooledConn.shouldDestroy()) {
    try {pooledConn.close(false);} catch (Exception e) {}
    destroyConnection(pooledConn);
  } ...
}
```

boglesby (commented):
When I initially tried to compile the diff, I go this:
```
> Task :geode-wan:compileDistributedTestJava
geode/geode-wan/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/WanCopyRegionCommandDUnitTest.java:871: error: cannot find symbol
    int receivedBatches = serverInB.invoke(() -> WANTestBase.getReceiverStats().get(2));
                                                            ^
  symbol:   method getReceiverStats()
  location: class WANTestBase
Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
1 error

> Task :geode-wan:compileDistributedTestJava FAILED
```
Thats because this revert removed the `getReceiverStats` method:
```
Revert "GEODE-8971: Add grace period when stopping gateway sender with group-...

This reverts commit 841fa06.
```
So, you'll have to add that back.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/functions/WanCopyRegionFunction.java
albertogpz: I've changed the call to `setShouldDestroy()` to close(). This way the connection is closed gracefully and we get rid of the log warnings in the receiving members.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/functions/WanCopyRegionFunction.java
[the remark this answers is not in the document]
mhansonp: I think an automatic approach is really needed. you cannot connect to geode if it is shutting down go run the command.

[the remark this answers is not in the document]
mhansonp: this should upon checking cancellation and getting a cancel automatically terminate the copy. This assumes that we care if there is a partial copy which I assume we do.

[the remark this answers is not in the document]
albertogpz: @mhansonp I am not sure I understand what you mean.
Would it be ok to check inside the function if there is a cancel in progress, and if so, terminate it, reporting in the output that it was terminated due to the cache being closed?
For example, add the following at the beginning of `doPostSendBatchActions()`:

```
if (cache.getCancelCriterion().isCancelInProgress()) {
<terminate function and report in the output of the command that the cache was closed>
}
```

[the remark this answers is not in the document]
mhansonp: Let me check with some of my peers.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/WanCopyRegionCommand.java
```diff
@@ -0,0 +1,83 @@
...
+          unspecifiedDefaultValue = "1000",
+          help = CliStrings.WAN_COPY_REGION__BATCHSIZE__HELP) int batchSize,
+      @CliOption(key = CliStrings.WAN_COPY_REGION__CANCEL,
+          unspecifiedDefaultValue = "false",
+          specifiedDefaultValue = "true",
+          help = CliStrings.WAN_COPY_REGION__CANCEL__HELP) boolean isCancel) {
+
+    authorize(Resource.DATA, Operation.WRITE, regionName);
+    final Object[] args = {regionName, senderId, isCancel, maxRate, batchSize};
+    ResultCollector<?, ?> resultCollector =
```
jinmeiliao, line 61: Any preconditions to consider? Does the region exists? is the region of the right type? Is the region in the right state to be copied?

geode-wan/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/WanCopyRegionCommandDUnitTest.java
```diff
@@ -0,0 +1,1293 @@
...
+        new Condition<>(s -> s.startsWith("Entries copied:"), "Entries copied");
+    Condition<String> senderNotPrimary = new Condition<>(
+        s -> s.equals("Sender B is serial and not primary. 0 entries copied."),
+        "sender not primary");
+
+    command.hasTableSection(ResultModel.MEMBER_STATUS_SECTION).hasColumn("Message")
+        .asList().haveExactly(1, haveEntriesCopied).haveExactly(2, senderNotPrimary);
+  }
+
+  /**
```
mhansonp, line 781: Nice description

mhansonp (approved):
I have taken a look at this and it all seems fine. Thanks for your patience.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/WanCopyRegionCommand.java
albertogpz: The region's existence as well as the sender's is checked in the function.
I cannot think of other preconditions but if you have any more in mind, please, let me know.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/WanCopyRegionCommand.java
```diff
@@ -0,0 +1,83 @@
...
+
+  @CliCommand(value = CliStrings.WAN_COPY_REGION, help = CliStrings.WAN_COPY_REGION__HELP)
+  @CliMetaData(relatedTopic = {CliStrings.TOPIC_GEODE_DATA, CliStrings.TOPIC_GEODE_REGION})
+  public ResultModel wanCopyRegion(
+      @CliOption(key = CliStrings.WAN_COPY_REGION__REGION, mandatory = true,
+          optionContext = ConverterHint.REGION_PATH,
+          help = CliStrings.WAN_COPY_REGION__REGION__HELP) String regionName,
+      @CliOption(key = CliStrings.WAN_COPY_REGION__SENDERID,
+          optionContext = ConverterHint.GATEWAY_SENDER_ID,
+          help = CliStrings.WAN_COPY_REGION__SENDERID__HELP) String senderId,
```
boglesby, line 47: This is along the same lines as Jinmei's comment below.

If I execute the command with no sender id (which is currently not mandatory), I see:
```
Executing - wan-copy region --region=/data_pr_parallel

Member | Status | Message
------ | ------ | ---------------------
ln-1   | ERROR  | Sender null not found
```
I think the sender id should be mandatory:
```
@CliOption(key = CliStrings.WAN_COPY_REGION__SENDERID, mandatory = true,
    optionContext = ConverterHint.GATEWAY_SENDER_ID,
    help = CliStrings.WAN_COPY_REGION__SENDERID__HELP) String senderId,
```
That provides a better message:
```
Executing - wan-copy region --region=/data_pr_parallel

Jul 01, 2021 1:42:01 PM org.springframework.shell.core.SimpleParser printHintMessage
WARNING: You should specify option (--sender-id, --max-rate, --batch-size, --cancel) for this command
  --sender-id=  is required
Use "help wan-copy region" (without the quotes) for detailed usage information.
```

boglesby (commented):
I also ran this test:

2 servers on site1
1 server on site2
Partitioned region with parallel sender
1,000,000 entries
Wan copy from site1 to site2

I see the correct behavior:
```
Member | Status | Message
------ | ------ | -----------------------
ln-2   | OK     | Entries copied: 504,400
ln-1   | OK     | Entries copied: 495,600
```
If I redo the test and kill one of the site1 servers while the copy is happening, I see:
```
Member | Status | Message
------ | ------ | ---------------------------------------------------------------------------------------------------------------------------
ln-1   | ERROR  | Execution failed. Error: org.apache.geode.distributed.PoolCancelledException: Pool PoolImpl@1539660778 name=ny is shut down
ln-2   | OK     | Entries copied: 673,422
```
or:
```
Member | Status | Message
------ | ------ | -------------------------------------------------------------------------------------------------------------------------------------------------
ln-1   | ERROR  | Execution failed. Error: org.apache.geode.InternalGemFireError: expected int part to be of type BYTE, part = partCode=unknown code 6 partLength=0
ln-2   | OK     | Entries copied: 844,202
```
I'm not sure if this is ok or not. I guess I expected the function to be retried.

boglesby:
You should clean up the raw warnings in both WanCopyRegionCommand and WanCopyRegionFunction.

albertogpz:
> I also ran this test:
> 
> 2 servers on site1
> 1 server on site2
> Partitioned region with parallel sender
> 1,000,000 entries
> Wan copy from site1 to site2
> 
> I see the correct behavior:
> 
> ```
> Member | Status | Message
> ------ | ------ | -----------------------
> ln-2   | OK     | Entries copied: 504,400
> ln-1   | OK     | Entries copied: 495,600
> ```
> 
> If I redo the test and kill one of the site1 servers while the copy is happening, I see:
> 
> ```
> Member | Status | Message
> ------ | ------ | ---------------------------------------------------------------------------------------------------------------------------
> ln-1   | ERROR  | Execution failed. Error: org.apache.geode.distributed.PoolCancelledException: Pool PoolImpl@1539660778 name=ny is shut down
> ln-2   | OK     | Entries copied: 673,422
> ```
> 
> or:
> 
> ```
> Member | Status | Message
> ------ | ------ | -------------------------------------------------------------------------------------------------------------------------------------------------
> ln-1   | ERROR  | Execution failed. Error: org.apache.geode.InternalGemFireError: expected int part to be of type BYTE, part = partCode=unknown code 6 partLength=0
> ln-2   | OK     | Entries copied: 844,202
> ```
> 
> I'm not sure if this is ok or not. I guess I expected the function to be retried.

My idea was that the function is not retried if there is an error in any of the servers. Automatic retrying could get tricky depending on the cases so I would rather have the command fail and let the user retry manually.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/WanCopyRegionCommand.java
albertogpz: Now I understand. I forgot to make sender-id mandatory. Thanks to both!

boglesby (approved):

geode-core/src/main/java/org/apache/geode/management/internal/i18n/CliStrings.java
```diff
@@ -1425,6 +1425,55 @@
...
+      "Region from which data will be exported.";
+  public static final String WAN_COPY_REGION__SENDERID = "sender-id";
+  public static final String WAN_COPY_REGION__SENDERID__HELP =
+      "Sender Id to use to copy the region.";
+  public static final String WAN_COPY_REGION__MAXRATE = "max-rate";
+  public static final String WAN_COPY_REGION__MAXRATE__HELP =
+      "Maximum rate for copying in events per second.";
+  public static final String WAN_COPY_REGION__BATCHSIZE = "batch-size";
+  public static final String WAN_COPY_REGION__BATCHSIZE__HELP =
+      "Size for batches of events to be copied.";
```
davebarnes97, line 1443: Add units to help message for batch-size. (Just as you already did for max-rate.)

davebarnes97:
@albertogpz I find it confusing that the `--cancel` command option can optionally accept a boolean value, when it could be simply present or absent. What is the use case for `--cancel=false`?

geode-core/src/main/java/org/apache/geode/management/internal/i18n/CliStrings.java
albertogpz: Yes, you are right.

albertogpz:
> @albertogpz I find it confusing that the `--cancel` command option can optionally accept a boolean value, when it could be simply present or absent. What is the use case for `--cancel=false`?

I agree. It would not make sense to use --cancel=false.

The normal use would be to either specify --cancel or not use --cancel at all. But the `CliOption` for boolean parameters class does not offer, as far as I know, the possibility to have an options without values.
If you know a way to do it, please, let me know.

davebarnes97:
> > @albertogpz I find it confusing that the `--cancel` command option can optionally accept a boolean value, when it could be simply present or absent. What is the use case for `--cancel=false`?
> 
> I agree. It would not make sense to use --cancel=false.
> 
> The normal use would be to either specify --cancel or not use --cancel at all. But the `CliOption` for boolean parameters class does not offer, as far as I know, the possibility to have an options without values.
> If you know a way to do it, please, let me know.

Thank you for explaining that CliOption allows the boolean parameter to be specified -- that would explain why I've encountered similar constructs when documenting other gfsh commands. In my edit of the docs for this command, I will omit that complication. I find it unhelpful, particularly when trying to describe the default behavior.

albertogpz:
> > > @albertogpz I find it confusing that the `--cancel` command option can optionally accept a boolean value, when it could be simply present or absent. What is the use case for `--cancel=false`?
> > 
> > 
> > I agree. It would not make sense to use --cancel=false.
> > The normal use would be to either specify --cancel or not use --cancel at all. But the `CliOption` for boolean parameters class does not offer, as far as I know, the possibility to have an options without values.
> > If you know a way to do it, please, let me know.
> 
> Thank you for explaining that CliOption allows the boolean parameter to be specified -- that would explain why I've encountered similar constructs when documenting other gfsh commands. In my edit of the docs for this command, I will omit that complication. I find it unhelpful, particularly when trying to describe the default behavior.

We could change the documentation of the command to say:
`[--cancel]` instead of `[--cancel(=value)?]` but the command would allow to set a value anyway.

davebarnes97:
> > > > @albertogpz I find it confusing that the `--cancel` command option can optionally accept a boolean value, when it could be simply present or absent. What is the use case for `--cancel=false`?
> > > 
> > > 
> > > I agree. It would not make sense to use --cancel=false.
> > > The normal use would be to either specify --cancel or not use --cancel at all. But the `CliOption` for boolean parameters class does not offer, as far as I know, the possibility to have an options without values.
> > > If you know a way to do it, please, let me know.
> > 
> > 
> > Thank you for explaining that CliOption allows the boolean parameter to be specified -- that would explain why I've encountered similar constructs when documenting other gfsh commands. In my edit of the docs for this command, I will omit that complication. I find it unhelpful, particularly when trying to describe the default behavior.
> 
> We could change the documentation of the command to say:
> `[--cancel]` instead of `[--cancel(=value)?]` but the command would allow to set a value anyway.

Yes, that's my plan. The user guide does not have to provide an exhaustive list of every conceivable code path. In this case, we can direct the user to the most straightforward usage and leave it at that.

davebarnes97 (changes requested):
@albertogpz 
Thanks for writing the docs, Alberto. I would like to request a few edits.
As discussed, I've modified the description of the cancel option. I also re-ordered some of the text, formatted prerequisites as a bulleted list, and caught a couple of typos. Attached is a zip file containing:

- My edited version of wan_copy_region.html.md.erb
- Changes only, in the form of a diff
[GEODE-9369-dbarnes.zip](https://github.com/apache/geode/files/6804470/GEODE-9369-dbarnes.zip)

albertogpz:
> @albertogpz
> Thanks for writing the docs, Alberto. I would like to request a few edits.
> As discussed, I've modified the description of the cancel option. I also re-ordered some of the text, formatted prerequisites as a bulleted list, and caught a couple of typos. Attached is a zip file containing:
> 
>     * My edited version of wan_copy_region.html.md.erb
> 
>     * Changes only, in the form of a diff
>       [GEODE-9369-dbarnes.zip](https://github.com/apache/geode/files/6804470/GEODE-9369-dbarnes.zip)

@davebarnes97 Thanks for the improvements!

davebarnes97 (approved):
+1 Looks good to me.
Thank you for incorporating my doc suggestions.

jinmeiliao (approved):

echobravopapa (approved):

geode-wan/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/WanCopyRegionCommandDUnitTest.java
```diff
@@ -0,0 +1,1292 @@
...
+        .containsExactly(message, message, message);
+  }
+
+  @Test
+  public void testUnsuccessfulExecutionWithPartitionedRegionAndParallelSender_ExceptionAtReceiver()
+      throws Exception {
+    testUnsuccessfulExecution_ExceptionAtReceiver(true, true);
+  }
+
+  public void testUnsuccessfulExecution_ExceptionAtReceiver(
```
DonalEvans, line 158: This method is only used in one test, so it doesn't seem to need to be extracted here.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/functions/WanCopyRegionFunction.java
```diff
@@ -0,0 +1,562 @@
...
+    String executionName = getExecutionName(regionName, sender.getId());
+    Callable<CliFunctionResult> callable =
+        new WanCopyRegionCallable(this, context, region, sender, maxRate, batchSize);
+    FutureTask<CliFunctionResult> future = new FutureTask<>(callable);
+
+    if (executions.putIfAbsent(executionName, future) != null) {
+      return new CliFunctionResult(context.getMemberName(), CliFunctionResult.StatusState.ERROR,
+          CliStrings.format(CliStrings.WAN_COPY_REGION__MSG__ALREADY__RUNNING__COMMAND,
+              regionName, sender.getId()));
+    }
```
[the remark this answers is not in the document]
albertogpz: > There is no test coverage for this case. Could Unit and possibly DUnit tests be added to confirm that if multiple identical `wan-copy region` commands are executed at the same time, this error message is returned?

Actually, the testDetectOngoingExecution_XXX DUnit tests are testing this case.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/functions/WanCopyRegionFunction.java
```diff
@@ -0,0 +1,562 @@
...
+        throw new InterruptedException();
+      }
+    }
+  }
+
+  private int getAndIncrementBatchId() {
+    if (batchId + 1 == Integer.MAX_VALUE) {
+      batchId = 0;
+    }
+    return batchId++;
```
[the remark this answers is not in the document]
albertogpz: I realized that this variable will not be read/written by several threads. Therefore, we can remove the volatile keyword.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/WANTestBase.java
```diff
@@ -1320,6 +1339,23 @@ public static void checkGatewayReceiverStats(int processBatches, int eventsRecei
     assertEquals(creates, gatewayReceiverStats.getCreateRequest());
   }
 
+  public static List<Integer> getReceiverStats() {
```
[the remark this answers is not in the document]
albertogpz: The problem is that `GatewayReceiverStats` is not serializable and therefore, cannot be sent through the wire from/to cache servers/clients.

geode-wan/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/WanCopyRegionCommandDUnitTest.java
albertogpz: Actually, I forgot to add the corresponding tests for partitionedRegion + serialSender and replicatedRegion + serialSender.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/functions/WanCopyRegionFunction.java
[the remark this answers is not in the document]
DonalEvans: Ah, my mistake. I was looking for other places that reference the String constant, but the tests use just the beginning of this message. Would it be possible to strengthen the tests slightly by ensuring that the region and sender reported in the message are also correct?

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/WANTestBase.java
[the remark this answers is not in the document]
DonalEvans: Ah, that's a pain. This is fine as is then.

DonalEvans (approved):

kirklund (approved):

onichols-pivotal:
Hey @albertogpz it looks like a test in this PR is not passing on Windows.  For example, see https://concourse.apachegeode-ci.info/teams/main/pipelines/apache-develop-main/jobs/windows-unit-test-openjdk8/builds/143 and http://files.apachegeode-ci.info/builds/apache-develop-main/1.15.0-build.0436/test-results/test/1629876775/

Tip: you can add the `Windows` label to any PR (when creating it or before the next commit) to run Windows checks too.

albertogpz:
> Hey @albertogpz it looks like a test in this PR is not passing on Windows. For example, see https://concourse.apachegeode-ci.info/teams/main/pipelines/apache-develop-main/jobs/windows-unit-test-openjdk8/builds/143 and http://files.apachegeode-ci.info/builds/apache-develop-main/1.15.0-build.0436/test-results/test/1629876775/
> 
> Tip: you can add the `Windows` label to any PR (when creating it or before the next commit) to run Windows checks too.

Thanks @onichols-pivotal . I will try to fix it ASAP.

PR #6604, opened 2021-06-10
GEODE-9203:Implement pubsub numsub subcommand

jhutchison:
-rename pre-existing pubsub folders and files to be publishAndSubscribe
 -rename subcommandsIntegrationTest to pubsubIntegrationTest
 -add numsub option to pubsubexecutor
 -add findNumberOfSubscribersByChannel methods to Subscriptions and PubSub interface

### For all changes:

### Note:


DonalEvans:
What's the reasoning behind the rename of the package from "pubsub" to "publishAndSubscribe"? Also, it seems like you only renamed the packages in the "test", "integrationTest" and "acceptanceTest" submodules, but not "main" and "distributedTest." In addition, the 4 classes in the "acceptanceTest" submodule still have their package as "pubsub" in the files, despite being in the "publishAndSubscribe" directory.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/pubsub/PubSubExecutor.java
```diff
@@ -33,28 +33,57 @@
...
-    // specific to subCommands might be a better way to do this
-    if (command.getProcessedCommand().size() > 3) {
-      return RedisResponse
-          .error(String.format(ERROR_UNKNOWN_PUBSUB_SUBCOMMAND, new String(subCommand)));
     }
 
-    List<byte[]> response;
-    if (command.getProcessedCommand().size() > 2) {
-      byte[] pattern = command.getProcessedCommand().get(2);
+    return RedisResponse.array(response);
```
DonalEvans, line 63: To avoid having to change the return type of the `findChannelNames()` methods, you could return directly from each switch case, which would mean that you wouldn't need to define `List<Object> response` above.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/publishAndSubscribe/PubSub.java
```diff
@@ -96,7 +96,9 @@ long publish(
    */
   List<byte[]> findSubscriptionNames(Client client);
 
-  List<byte[]> findChannelNames();
+  List<Object> findChannelNames();
 
-  List<byte[]> findChannelNames(byte[] pattern);
+  List<Object> findChannelNames(byte[] pattern);
```
DonalEvans, line 101: These can (and should, I think) go back to `List<byte[]>` if the change in `PubSubExecutor` is applied.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/publishAndSubscribe/Subscriptions.java
```diff
@@ -73,28 +73,52 @@ boolean exists(Object channelOrPattern, Client client) {
...
+              .filter(subscription -> subscription instanceof ChannelSubscription)
+              .collect(Collectors.toList());
+
+      if (subscriptions.isEmpty()) {
+        result.add(name);
+        result.add(0L);
+      } else {
+        result.add(name);
+        result.add((long) subscriptions.size());
+      }
```
DonalEvans, line 115: This can be simplified to:
```
      Long subscriptions =
          findSubscriptions(name)
              .stream()
              .filter(subscription -> subscription instanceof ChannelSubscription)
              .count();

        result.add(name);
        result.add(subscriptions);
```

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/publishAndSubscribe/SubscriptionsJUnitTest.java
```diff
@@ -318,8 +319,74 @@ public void findChannelNames_shouldNotReturnDuplicates_givenMultipleSubscription
...
+  public void findNumberOfSubscribersByChannel_shouldReturnListOfChannelsAndSubscriberCount_givenListOfActiveChannels() {
+    Subscriptions subject = new Subscriptions();
+
+    Channel channel = mock(Channel.class);
+    when(channel.closeFuture()).thenReturn(mock(ChannelFuture.class));
+    Client client = new Client(channel);
+    ExecutionHandlerContext context = mock(ExecutionHandlerContext.class);
+
+    Subscription subscriptionFoo1 =
+        new ChannelSubscription(client, Coder.stringToBytes("foo"), context, subject);
```
DonalEvans, line 337: Can the byte arrays/strings in these tests be extracted to variables (when they're referred to in multiple places)?

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/publishAndSubscribe/SubscriptionsJUnitTest.java
```diff
@@ -318,8 +319,74 @@ public void findChannelNames_shouldNotReturnDuplicates_givenMultipleSubscription
...
+
+  @Test
+  public void findNumberOfSubscribersByChannel_shouldReturnChannelNameAndZero_givenInactiveActiveChannel() {
+    Subscriptions subject = new Subscriptions();
+
+    List<byte[]> channels = new ArrayList<>();
+    channels.add(Coder.stringToBytes("bar"));
+
+    ArrayList<Object> result =
+        (ArrayList<Object>) subject.findNumberOfSubscribersForChannel(channels);
```
DonalEvans, line 363: The cast here can be removed if this is changed to:
```
    List<Object> result = subject.findNumberOfSubscribersForChannel(channels);
```

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/publishAndSubscribe/SubscriptionsJUnitTest.java
```diff
@@ -318,8 +319,74 @@ public void findChannelNames_shouldNotReturnDuplicates_givenMultipleSubscription
...
+
+    assertThat(result.get(0)).isEqualTo(Coder.stringToBytes("bar"));
+    assertThat(result.get(1)).isEqualTo(0L);
+  }
+
+
+  @Test
+  public void findNumberOfSubscribersByChannel_shouldPatterAndZero_givenPatternSubscription() {
```
DonalEvans, line 371: It seems like there's a typo in this test name.

DonalEvans (changes requested):

DonalEvans:
> > What's the reasoning behind the rename of the package from "pubsub" to "publishAndSubscribe"? Also, it seems like you only renamed the packages in the "test", "integrationTest" and "acceptanceTest" submodules, but not "main" and "distributedTest." In addition, the 4 classes in the "acceptanceTest" submodule still have their package as "pubsub" in the files, despite being in the "publishAndSubscribe" directory.
> 
> Thanks for catching all that-concentration has been a little low this week :)
> 
> As far as the rationale for the change goes:
> When publlsh and subscribe functionality was originally introduced in our codebase, it was referred to as "pubsub" and a lot of stuff was named accordingly. Now that we've added the specific command named "pubsub" it's unclear what the identifier "pubsub" refers to (the specific command, or the over all publish and subscribe functionality). The rename is an attempt to untangle this and start a new convention of referring to the overall functionality by the full name and reserve the term "pubsub" for stuff related to the specific command by that name. hit me up directly if you disagree with that idea, or want to discuss in more depth?

Sounds reasonable, thanks for the explanation!

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/publishAndSubscribe/Subscriptions.java
jhutchison: thank you-  yes, simpler!

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/pubsub/PubSubExecutor.java
jhutchison: yeah, that's a bit better-  thanks

DonalEvans (approved):

jdeppe-pivotal (changes requested):
Would prefer to keep the package name more succinct and not have it be in a non-standard form. Also, would prefer to have a rename be in a separate PR.

jdeppe-pivotal:
I'm going to close and resubmit with just the `numsub` changes.

PR #6605, opened 2021-06-10
GEODE-9303: Implement Radish DUMP/RESTORE commands

jdeppe-pivotal:
- Dump/restore using internal Geode serialization mechanisms meaning
  that dumps can only be used to/from Radish instances.

### For all changes:

### Note:


geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/netty/ExecutionHandlerContext.java
```diff
@@ -234,7 +235,10 @@ private RedisResponse getExceptionResponse(ChannelHandlerContext ctx, Throwable
...
     }
 
-    if (cause instanceof NumberFormatException) {
+
+    if (cause instanceof RedisException) {
+      response = RedisResponse.error(cause.getMessage());
+    } else if (cause instanceof NumberFormatException) {
       response = RedisResponse.error(cause.getMessage());
     } else if (cause instanceof ArithmeticException) {
       response = RedisResponse.error(cause.getMessage());
```
DonalEvans, line 244: These three cases all have the same behaviour (calling `RedisResponse.error(cause.getMessage())`) which is already found further down in this if/else block. Would it make sense to combine all the checks into one case:
```
    } else if (cause instanceof IllegalStateException
        || cause instanceof RedisParametersMismatchException
        || cause instanceof RedisException
        || cause instanceof NumberFormatException
        || cause instanceof ArithmeticException) {
      response = RedisResponse.error(cause.getMessage());
```

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/key/AbstractDumpRestoreIntegrationTest.java
```diff
@@ -0,0 +1,241 @@
...
+
+  @Test
+  public void restore_withAbsTTL() {
+    long absttl = System.currentTimeMillis() + 10000;
+    lettuce.restore("restored", RESTORE_BYTES, new RestoreArgs().ttl(absttl).absttl());
+
+    String response = lettuce.get("restored");
+
+    assertThat(response).isEqualTo(STRING_VALUE);
+    assertThat(lettuce.ttl("restored")).isGreaterThan(1);
```
DonalEvans, line 166: Would it be possible to also add an assertion here that ttl is less than the expected value, similar to what's done in the `restore_withTTL_setsTTL()` test?

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/key/AbstractDumpRestoreIntegrationTest.java
```diff
@@ -0,0 +1,241 @@
...
+  }
+
+  @Test
+  public void restore_withTTL_setsTTL() {
+    lettuce.restore("restored", 2000, RESTORE_BYTES);
+    String response = lettuce.get("restored");
+
+    assertThat(response).isEqualTo(STRING_VALUE);
+    assertThat(lettuce.pttl("restored"))
+        .isLessThan(2000)
```
DonalEvans, line 142: It seems unlikely, but could this test possibly fail due to reaching this assertion before the ttl has had time to decrease from the initially set value? This assertion could then be `.isLessThatOrEqualTo(2000)` to guard against that.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/key/AbstractDumpRestoreIntegrationTest.java
```diff
@@ -0,0 +1,241 @@
...
+
+    assertThat(restored.size()).isEqualTo(hashy.size());
+    assertThat(restoredMap).containsAllEntriesOf(hashy);
+  }
+
+  @Test
+  public void dumpAndRestoreSortedSet() {
+    Set<String> smembers = new HashSet<>();
+    for (int i = 0; i < 100; i++) {
+      lettuce.sadd("set", "member-" + i);
```
DonalEvans, line 230: I think this should probably be `zadd`.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/key/AbstractDumpRestoreIntegrationTest.java
```diff
@@ -0,0 +1,241 @@
...
+    lettuce.restore("restored", RESTORE_BYTES, new RestoreArgs().ttl(absttl).absttl());
+
+    String response = lettuce.get("restored");
+
+    assertThat(response).isEqualTo(STRING_VALUE);
+    assertThat(lettuce.ttl("restored")).isGreaterThan(1);
+  }
+
+  @Test
+  public void restore_withAbsTTL_ofZero() {
```
DonalEvans, line 170: Would there be any value in a test for restore with an AbsTTL that's in the past? I.e. `long absttl = System.currentTimeMillis() - 10000;`

DonalEvans (changes requested):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/NullRedisData.java
```diff
@@ -89,6 +96,25 @@ public boolean rename(Region<RedisKey, RedisData> region, RedisKey oldKey, Redis
     return false;
   }
 
+  @Override
+  public byte[] dump() throws IOException {
```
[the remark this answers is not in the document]
jdeppe-pivotal: Removed.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/NullRedisData.java
```diff
@@ -89,6 +96,25 @@ public boolean rename(Region<RedisKey, RedisData> region, RedisKey oldKey, Redis
...
+    return new byte[0];
+  }
+
+  @Override
+  public RedisData restore(byte[] data, boolean replaceExisting) throws Exception {
```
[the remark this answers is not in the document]
jdeppe-pivotal: Removed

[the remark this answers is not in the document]
jdeppe-pivotal: Removed

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/key/DumpExecutor.java
```diff
@@ -0,0 +1,36 @@
...
+import org.apache.geode.redis.internal.data.RedisKey;
+import org.apache.geode.redis.internal.executor.AbstractExecutor;
+import org.apache.geode.redis.internal.executor.RedisResponse;
+import org.apache.geode.redis.internal.netty.Command;
+import org.apache.geode.redis.internal.netty.ExecutionHandlerContext;
+
+public class DumpExecutor extends AbstractExecutor {
+  @Override
+  public RedisResponse executeCommand(Command command, ExecutionHandlerContext context)
+      throws Exception {
```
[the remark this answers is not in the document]
jdeppe-pivotal: Removed

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/key/RestoreExecutor.java
```diff
@@ -0,0 +1,90 @@
...
+import org.apache.geode.redis.internal.netty.Command;
+import org.apache.geode.redis.internal.netty.ExecutionHandlerContext;
+
+public class RestoreExecutor extends AbstractExecutor {
+
+  private static final int TTL_INDEX = 2;
+
+  @Override
+  public RedisResponse executeCommand(Command command, ExecutionHandlerContext context)
+      throws Exception {
```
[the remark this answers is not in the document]
jdeppe-pivotal: Removed

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/key/RestoreExecutor.java
```diff
@@ -0,0 +1,90 @@
...
+    if (commandElems.size() > 4) {
+      options = parseOptions(commandElems.subList(4, commandElems.size()));
+    } else {
+      options = new RestoreOptions();
+    }
+
+    try {
+      context.getKeyCommands().restore(key, ttl, commandElems.get(3), options);
+    } catch (RedisRestoreKeyExistsException redisRestoreKeyExistsException) {
+      return RedisResponse.busykey(ERROR_RESTORE_KEY_EXISTS);
```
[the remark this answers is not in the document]
jdeppe-pivotal: I'm not sure about that, but I'd prefer to keep all the error messages together.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/netty/ExecutionHandlerContext.java
jdeppe-pivotal: Good suggestion. I've done that and also noticed a couple of other exceptions that can be removed (related to function execution which we are not doing anymore).

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/key/AbstractDumpRestoreIntegrationTest.java
jdeppe-pivotal: Sure.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/key/AbstractDumpRestoreIntegrationTest.java
jdeppe-pivotal: Yes, it is unlikely but I've changed it regardless.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/key/AbstractDumpRestoreIntegrationTest.java
jdeppe-pivotal: Oops - right!

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/key/AbstractDumpRestoreIntegrationTest.java
jdeppe-pivotal: Added

DonalEvans (approved):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/RedisCommandType.java
```diff
@@ -238,6 +240,8 @@
 
   /*************** Keys ******************/
 
+  DUMP(new DumpExecutor(), UNSUPPORTED, new ExactParameterRequirements(2)),
+  RESTORE(new RestoreExecutor(), UNSUPPORTED, new MinimumParameterRequirements(4)),
```
sabbey37, line 244: Did we decide to mark these as supported?

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/AbstractRedisData.java
```diff
@@ -207,6 +210,25 @@ protected void clearDelta() {
     this.deltaInfo = null;
   }
 
+  @Override
+  public byte[] dump() throws IOException {
+    ByteArrayOutputStream baos = new ByteArrayOutputStream();
```
sabbey37, line 215: baos. yum.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/key/AbstractDumpRestoreIntegrationTest.java
```diff
@@ -0,0 +1,257 @@
...
+  @Test
+  public void restoreFails_whenKeyAlreadyExists() {
+    lettuce.set("restored", "already exists");
+
+    assertThatThrownBy(() -> lettuce.restore("restored", 0, RESTORE_BYTES))
+        .hasMessage("BUSYKEY Target key name already exists.");
+  }
+
+  @Test
+  public void restoreFails_whenTTLisNegative() {
```
sabbey37, line 113: Could we also add a test for when TTL is not an integer?

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/key/DumpRestoreIntegrationTest.java
```diff
@@ -0,0 +1,32 @@
...
+ * the License.
+ */
+
+package org.apache.geode.redis.internal.executor.key;
+
+import org.junit.ClassRule;
+
+import org.apache.geode.redis.GeodeRedisServerRule;
+
+public class DumpRestoreIntegrationTest extends AbstractDumpRestoreIntegrationTest {
```
sabbey37, line 22: Could we add a Radish-specific test to verify that we return error messages for `IDLETIME` and `FREQ`?

sabbey37 (commented):
So... I guess I can't use this for my Turnip instances?

Also, is it possible to add some validation of the need for DUMP/RESTORE with Spring-based apps... maybe we can add something to our existing Spring Session tests?

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/RedisCommandType.java
jdeppe-pivotal: Doh!

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/key/AbstractDumpRestoreIntegrationTest.java
jdeppe-pivotal: Done.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/key/DumpRestoreIntegrationTest.java
jdeppe-pivotal: Added.

jdeppe-pivotal:
> So... I guess I can't use this for my Turnip instances?
> 
> Also, is it possible to add some validation of the need for DUMP/RESTORE with Spring-based apps... maybe we can add something to our existing Spring Session tests?

Good call - I've added a story for this.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/key/AbstractDumpRestoreIntegrationTest.java
```diff
@@ -0,0 +1,264 @@
...
+    assertThatThrownBy(() -> lettuce.restore("restored", -1, RESTORE_BYTES))
+        .hasMessage("ERR Invalid TTL value, must be >= 0");
+  }
+
+  @Test
+  public void restoreFails_whenTTLisNotANumber() {
+    assertThatThrownBy(() -> jedis.sendCommand("restored".getBytes(), Protocol.Command.RESTORE,
+        "restored".getBytes(), "not-an-integer".getBytes(), RESTORE_BYTES))
+            .hasMessage("ERR value is not an integer or out of range");
```
sabbey37, line 122: Could we use the `stringToBytes` method from `Coder` for these?

sabbey37 (commented):
Thanks for adding the Spring-related test!  Have you tested this change against our internal tests?

sabbey37 (commented):
Thanks for adding the Spring-related test!  Have you tested this change against our internal tests?

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/key/RestoreExecutor.java
```diff
@@ -0,0 +1,89 @@
...
+
+  private RestoreOptions parseOptions(List<byte[]> options) {
+    RestoreOptions restoreOptions = new RestoreOptions();
+    for (byte[] optionBytes : options) {
+      String option = Coder.bytesToString(optionBytes).toUpperCase();
+      switch (option) {
+        case "REPLACE":
+          restoreOptions.setReplace(true);
+          break;
+        case "ABSTTL":
```
DonalEvans, line 79: With the changes in https://github.com/apache/geode/pull/6599 now merged, could this be changed to make use of the `Coder.equalIgnoreCaseBytes` method, with constants for the byte array representations of "REPLACE" and "ABSTTL" added to the `StringBytesGlossary` class?

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/netty/Coder.java
```diff
@@ -83,6 +83,9 @@
   @MakeImmutable
   public static final byte[] oom = stringToBytes("OOM ");
 
+  @MakeImmutable
+  public static final byte[] BUSYKEY = stringToBytes("BUSYKEY ");
+
```
DonalEvans, line 88: Could this be moved to the `StringBytesGlossary` class after rebasing on develop?

DonalEvans (changes requested):

sabbey37:
Already discussed this with you via Zoom, but putting it here for visibility/posterity.  We are able to dump the contents of a key into a file with Redis (via redis-cli), remove the line feed at the end of that file, then restore the key from the file. We currently cannot do this with Radish.  Do we expect to?  Is this something we should look into further?

```
redis-cli sadd a 1 2 3 4
redis-cli --raw dump a > keyDump
(delete line feed)
cat keyDump | redis-cli -x restore mynewkey 0
redis-cli smembers mynewkey
1) "1"
2) "2"
3) "3"
4) "4"
```

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/key/AbstractDumpRestoreIntegrationTest.java
jdeppe-pivotal: Since this is all client-side (i.e. how a client might call our API) I'd prefer not to imply that we need to use the `Coder` methods here.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/key/RestoreExecutor.java
jdeppe-pivotal: Done

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/netty/Coder.java
jdeppe-pivotal: Done

onichols-pivotal (approved):

jdeppe-pivotal:
@sabbey37 Since you were using the `GeodeServerRunTest` to launch a server, I think the problem was that the correct serialization registration hadn't been configured:
```
    InternalDataSerializer.getDSFIDSerializer().registerDSFID(
        DataSerializableFixedID.REDIS_SET_ID,
        RedisSet.class);
```
I've also added some error logging to the `restore` method so this would have been obvious much sooner :(.

dickcav (approved):

rhoughton-pivot (approved):

DonalEvans (approved):

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/key/AbstractDumpRestoreIntegrationTest.java
sabbey37: I definitely see your point, but I think we decided as a team to switch everything over to use the `stringToBytes` method from Coder.  The only places we don't use that method (other than these dump/restore tests) are when we use a CHARSET other than UTF-8 or a random spot in `RedisString` that must've been overlooked.

sabbey37:
@jdeppe-pivotal Nice!  Thanks for adding that logging.  I just tried it out, and it was very helpful!  Also, I was able to use the raw dump/restore successfully via the CLI once the serializables were properly registered.  Do you think we are calling `registerDataSerializables` at the right point or should it be done further down in `GeodeRedisServer`?

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/key/AbstractDumpRestoreIntegrationTest.java
sabbey37: Marking as resolved per our team conversation.  Thanks for bringing it up!

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/AbstractRedisData.java
```diff
@@ -207,6 +212,22 @@ protected void clearDelta() {
...
+    DataSerializer.writeObject(this, new DataOutputStream(baos));
```
upthewaterspout, line 218: I think we need some extra information for future compatibility. We probably need to include the geode version in the dumped object, so that restore can create a VersionedDataInputStream with that geode version.

I think we also are going to want to be able to `RESTORE` from values that come from native redis at some point. So I think at least the header for our dump format should match redis's - but with an incompatible version or something. So in the future we could read a header and do a switch in the version to restore in our format or there.

jdeppe-pivotal: Unfortunately the redis dump format doesn't contain any distinguishing header information. Although all documentation indicates objects are dumped in RDB format, that format is only per object and doesn't include any envelope information. Basically the dump format is:
- 1 byte for the object type
- serialized object bytes
- 2 bytes for RDB version
- 8 bytes for CRC checksum

We can still create a unique header just for our format though.

upthewaterspout: Interesting. Well, it looks like the native redis format has the version information, just in the footer and not a header. I guess we are relying on native redis to reject our dump files because the checksum won't match? And we will reject native redis dump files because they don't have our magic bytes?

echobravopapa (approved):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/AbstractRedisData.java
jdeppe-pivotal: That's pretty much it. The redis object types are defined here: https://github.com/redis/redis/blob/4fa3e23092d642f0c67116d9f7616f4f1173a1cf/src/rdb.h#L75-L93. For redis, the first byte of the object dump will be one of these values.

PR #6613, opened 2021-06-15
GEODE-9295: Reply sent always  while processing LatestLastAccessTimeM…

nabarunnag:
…essage

	* Even if there any any exception, a reply will be sent back to the sender so that the sender's threads are not stuck.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/LatestLastAccessTimeMessage.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/LatestLastAccessTimeMessage.java
@@ -62,34 +62,34 @@ public class LatestLastAccessTimeMessage<K> extends PooledDistributionMessage
 
   @Override
   protected void process(ClusterDistributionManager dm) {
-    final InternalCache cache = dm.getCache();
-    if (cache == null) {
-      sendReply(dm, 0);
-      return;
-    }
-    final InternalDistributedRegion region =
-        (InternalDistributedRegion) cache.getRegion(this.regionName);
-    if (region == null) {
-      sendReply(dm, 0);
-      return;
-    }
-    final RegionEntry entry = region.getRegionEntry(this.key);
-    if (entry == null) {
-      sendReply(dm, 0);
-      return;
-    }
     long lastAccessed = 0L;
-    // noinspection SynchronizationOnLocalVariableOrMethodParameter
-    synchronized (entry) {
-      if (!entry.isInvalidOrRemoved()) {
-        try {
-          lastAccessed = entry.getLastAccessed();
-        } catch (InternalStatisticsDisabledException ignored) {
-          // last access time is not available
+    try {
+      final InternalCache cache = dm.getCache();
+      if (cache == null) {
+        return;
+      }
+      final InternalDistributedRegion region =
+          (InternalDistributedRegion) cache.getRegion(this.regionName);
+      if (region == null) {
+        return;
+      }
+      final RegionEntry entry = region.getRegionEntry(this.key);
+      if (entry == null) {
+        return;
+      }
+      // noinspection SynchronizationOnLocalVariableOrMethodParameter
+      synchronized (entry) {
+        if (!entry.isInvalidOrRemoved()) {
+          try {
+            lastAccessed = entry.getLastAccessed();
+          } catch (InternalStatisticsDisabledException ignored) {
+            // last access time is not available
+          }
         }
       }
+    } finally {
+      sendReply(dm, lastAccessed);
     }
-    sendReply(dm, lastAccessed);
   }
 
   @VisibleForTesting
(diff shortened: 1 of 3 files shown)


echobravopapa (approved):
Thanks @nabarunnag looks good and I think this addresses @pivotal-jbarrett concerns of stuck threads, but I'll let him concur officially

pivotal-jbarrett (approved):
I would like to see this paired with a change to the caller, `LatestLastAccessTimeOperation`, to use the `waitForReplies(long timeout)` method with a reasonable timeout for this rather ephemeral information. This way if something else buggers up these replies in the future the thread doesn’t hang forever. 1min seems reasonable, 5min would be ok.

sabbey37:
It would be nice to add a DUnit test that replicates the issue we were seeing.

pivotal-jbarrett:
> It would be nice to add a DUnit test that replicates the issue we were seeing.

I agree!

The test doesn't need to setup all the expiration and try get the bucket migration timing to align with an expiration event. We could just have server A invoke this operation on server B for a key we know B does not have. It seems like it should be possible to do the same for the region too by sending the message with a region name that does not exist on B. In both cases without the fix should hang the calling thread on A. You can unstick it in the test, if necessary, by invoking `thread.interrupt()` followed by `LockSupport.unpark`.

boglesby (approved):

geode-core/src/test/java/org/apache/geode/internal/cache/LatestLastAccessTimeMessageTest.java
```diff
@@ -77,6 +78,19 @@ public void processWithNullCacheRepliesZero() {
...
+  @Test
+  public void replyIsSentEvenIfThereIsAnException() {
+    setupMessage();
+    when(dm.getCache()).thenThrow(new RuntimeException());
+    try {
+      lastAccessTimeMessage.process(dm);
+      fail();
+    } catch (RuntimeException ex) {
+      // Expected
+    }
```
DonalEvans, line 90: Rather than using `fail()` and catching the exception just to ignore it, it might be better to extract the `new RuntimeException()` to a variable and then use `assertThatThrownBy()` to confirm that that exception is thrown.

DonalEvans (approved):
Looks good, just one small suggestion, feel free not to take it.

boglesby:
I'll work on the dunit test. Since my test to make it happen for real is too unreliable, I'll simulate the issue by just creating the reply processor and message in one member and sending it to another member in various states.

nabarunnag:
@DonalEvans Thank you for the review, the changes have been implemented.

DonalEvans:
If I revert the changes in this PR, the newly added DUnit test still passes.

nabarunnag:
@DonalEvans for the test to fail we have to revert all the fixes to GEODE-9295, 3b9acf56100da663d4df60417dc389d6c7aea2a9
The previous SHA did not contain a DUnitTest

DonalEvans:
> @DonalEvans for the test to fail we have to revert all the fixes to GEODE-9295, [3b9acf5](https://github.com/apache/geode/commit/3b9acf56100da663d4df60417dc389d6c7aea2a9)
> The previous SHA did not contain a DUnitTest

Aah, I see. Thanks for the explanation, I wasn't aware that there had been a previous commit for this issue.

boglesby:
Yes, the dunit test should definitely pass on current develop. It doesn't pass on 1.13.2.

geode-core/src/main/java/org/apache/geode/internal/cache/LatestLastAccessTimeMessage.java
```diff
@@ -62,34 +62,34 @@ public int getDSFID() {
...
+        if (!entry.isInvalidOrRemoved()) {
+          try {
+            lastAccessed = entry.getLastAccessed();
+          } catch (InternalStatisticsDisabledException ignored) {
+            // last access time is not available
+          }
         }
       }
+    } finally {
+      sendReply(dm, lastAccessed);
```
agingade, line 91: What we want is here to get the last Access time; if the entry is present and send its value....We could write down the path of checking if its not null and then access the time.  You may have thought about it...And logically they are all same...I will leave it to you to see which one you want to pick:

    long lastAccessed = 0L;
    try {
      InternalCache cache = dm.getCache();
      if (cache != null) {
        InternalDistributedRegion region =
            (InternalDistributedRegion) cache.getRegion(regionName);
        if (region != null) {
          RegionEntry entry = region.getRegionEntry(key);
          if (entry != null) {
            try {
              lastAccessed = entry.getLastAccessed();
            } catch (InternalStatisticsDisabledException ignored) {
              // last access time is not available
            }
          }
        }
      }
    } finally {
      sendReply(dm, lastAccessed);
    }

agingade (approved):

PR #6477, opened 2021-05-13
GEODE-9248: Server hosting CQ queue uneccessary fills bucketToTempQue…

jvarenina:
…ueMap

Issue reproduces when following conditions are fulfilled:
   - Redundant partition region must be configured
   - Number of servers must be greater than number of redundant copies of partition region
   - Parallel gateway sender must be configured on partition region
   - Client must register CQs for the region
   - Transactions must be used with put operations
   - Events must be enqueued in parallel gateway senders (remote site is unavailable)

 Server that is hosting primary bucket will send TXCommitMessage to the secondary
 server, and also to the server that is hosting CQ subscription queue (if CQ condition is fulfilled). 
 The problem occurs when the server that is hosting CQ subscription queue does not host
 the bucket for which event is received. In that case the server will store the event in 
 bucketToTempQueueMap, because it assumes that the bucket is in the process of the
 creation, which is not correct.

The fix:
Normal put operation (transaction is not used) is distributed to adjunct
member (only host sub. queue, and not bucket) using seperate
partition.PutMessage which indicates that it is notificationOnly, and
also sets TailKey to -1. Adjunct member will not store event in queue,
because TailKey is always expected to be set in case of parallel
gateway-sender. Members that host secondary bucket will receive the
UpdateOperation message with valid TailKey value.

So this solution implements the similar approach to what is done in
case of normal put operation. TxCommitMessage with TailKey set to -1
will be sent only to adjunct member that host subscription queue, and not
secondary bucket. Other members will get valid TailKey value. Also,
member that host both secondary bucket and subscription queue will get
valid TailKey value.

Thank you for submitting a contribution to Apache Geode.

In order to streamline the review of the contribution we ask you
to ensure the following steps have been taken:

### For all changes:
- [x] Is there a JIRA ticket associated with this PR? Is it referenced in the commit message?

- [x] Has your PR been rebased against the latest commit within the target branch (typically `develop`)?

- [x] Is your initial contribution a single, squashed commit?

- [x] Does `gradlew build` run cleanly?

- [x] Have you written or updated unit tests to verify your changes?

- [ ] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)?

### Note:
Please ensure that once the PR is submitted, check Concourse for build issues and
submit an update to your PR as soon as possible. If you need help, please send an
email to dev@geode.apache.org.

PR #6625, opened 2021-06-17
GEODE-9248: Server hosting CQ queue uneccessary fills bucketToTempQueue

mivanac:
Issue reproduces when following conditions are fulfilled:

    Redundant partition region must be configured
    Number of servers must be greater than number of redundant copies of partition region
    Parallel gateway sender must be configured on partition region
    Client must register CQs for the region
    Transactions must be used with put operations
    Events must be enqueued in parallel gateway senders (remote site is unavailable)

Server that is hosting primary bucket will send TXCommitMessage to the secondary
server, and also to the server that is hosting CQ subscription queue (if CQ condition is fulfilled).
The problem occurs when the server that is hosting CQ subscription queue does not host
the bucket for which event is received. In that case the server will store the event in
bucketToTempQueueMap, because it assumes that the bucket is in the process of the
creation, which is not correct.

The fix:
Normal put operation (transaction is not used) is distributed to adjunct
member (only host sub. queue, and not bucket) using seperate
partition.PutMessage which indicates that it is notificationOnly, and
also sets TailKey to -1. Adjunct member will not store event in queue,
because TailKey is always expected to be set in case of parallel
gateway-sender. Members that host secondary bucket will receive the
UpdateOperation message with valid TailKey value.

So this solution implements the similar approach to what is done in
case of normal put operation. TxCommitMessage with TailKey set to -1
will be sent only to adjunct member that host subscription queue, and not
secondary bucket. Other members will get valid TailKey value. Also,
member that host both secondary bucket and subscription queue will get
valid TailKey value.

### For all changes:

### Note:

For more information see reverted PR https://github.com/apache/geode/pull/6477

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/TXCommitMessage.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/TXCommitMessage.java
@@ -127,7 +127,10 @@ public class TXCommitMessage extends PooledDistributionMessage
    */
   private transient boolean hasReliableRegions = false;
 
-  public transient Set<InternalDistributedMember> notificationOnlyMembers;
+  /**
+   * Member that requires notification only message.
+   */
+  private transient Set<InternalDistributedMember> notificationOnlyMembers = Collections.emptySet();
 
   /**
    * Set of all caching exceptions produced while processing this tx
@@ -196,6 +199,14 @@ public class TXCommitMessage extends PooledDistributionMessage
   protected EventID getEventId(int eventOffset) {
     return new EventID(this.farsideBaseMembershipId, this.farsideBaseThreadId,
         this.farsideBaseSequenceId + eventOffset);
+  }
+
+  public void setNotificationOnlyMembers(Set<InternalDistributedMember> members) {
+    this.notificationOnlyMembers = members;
+  }
+
+  public Set<InternalDistributedMember> getNotificationOnlyMembers() {
+    return this.notificationOnlyMembers;
   }
 
   /**
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/TXCommitMessage.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/TXCommitMessage.java
@@ -441,6 +441,7 @@ public class TXCommitMessage extends PooledDistributionMessage
                   }
                 }
                 tempNotificationOnlyMembers.removeAll(tempTransactionMembers);
+                tempNotificationOnlyMembers.retainAll(recipients);
               }
 
               if (tempNotificationOnlyMembers.isEmpty()) {
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/TXCommitMessage.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/TXCommitMessage.java
@@ -416,12 +416,12 @@ public class TXCommitMessage extends PooledDistributionMessage
                 setRecipientsSendData(Collections.singleton(indivRecip.next()), processor, rcl);
               }
             } else {
-              HashSet tempNotificationOnlyMembers = new HashSet();
-              HashSet tempTransactionMembers = new HashSet();
+              Set<InternalDistributedMember> tempNotificationOnlyMembers = new HashSet();
+              Set<InternalDistributedMember> tempTransactionMembers = new HashSet();
               if (!rcl.isEmpty() && !getNotificationOnlyMembers().isEmpty()) {
                 for (RegionCommit rc : rcl) {
-                  Set getNOM = getNotificationOnlyMembers().get(rc);
-                  Set getTM = getTransactionMembers().get(rc);
+                  Set<InternalDistributedMember> getNOM = getNotificationOnlyMembers().get(rc);
+                  Set<InternalDistributedMember> getTM = getTransactionMembers().get(rc);
                   if (getNOM != null && !getNOM.isEmpty()) {
                     tempNotificationOnlyMembers.addAll(getNOM);
                   }
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/TXCommitMessage.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/TXCommitMessage.java
@@ -134,6 +134,15 @@ public class TXCommitMessage extends PooledDistributionMessage
       new HashMap<>();
 
   /**
+   * Member that the current transaction will be distributed to.
+   * Currently this is any other member who has this region defined.
+   */
+  private final transient Map<RegionCommit, Set<InternalDistributedMember>> transactionMembers =
+      new HashMap<>();
+
+
+
+  /**
    * Set of all caching exceptions produced while processing this tx
    */
   private transient Set processingExceptions = Collections.emptySet();
@@ -205,6 +214,11 @@ public class TXCommitMessage extends PooledDistributionMessage
   Map<RegionCommit, Set<InternalDistributedMember>> getNotificationOnlyMembers() {
     return this.notificationOnlyMembers;
   }
+
+  Map<RegionCommit, Set<InternalDistributedMember>> getTransactionMembers() {
+    return this.transactionMembers;
+  }
+
 
   /**
    * Return the TXCommitMessage we have already received that is associated with id. Note because of
@@ -403,13 +417,19 @@ public class TXCommitMessage extends PooledDistributionMessage
               }
             } else {
               HashSet tempNotificationOnlyMembers = new HashSet();
+              HashSet tempTransactionMembers = new HashSet();
               if (!rcl.isEmpty() && !getNotificationOnlyMembers().isEmpty()) {
                 for (RegionCommit rc : rcl) {
                   Set getNOM = getNotificationOnlyMembers().get(rc);
+                  Set getTM = getTransactionMembers().get(rc);
                   if (getNOM != null && !getNOM.isEmpty()) {
                     tempNotificationOnlyMembers.addAll(getNOM);
                   }
+                  if (getTM != null && !getTM.isEmpty()) {
+                    tempTransactionMembers.addAll(getTM);
+                  }
                 }
+                tempNotificationOnlyMembers.removeAll(tempTransactionMembers);
               }
 
               if (tempNotificationOnlyMembers.isEmpty()) {
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/TXCommitMessage.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/TXCommitMessage.java
@@ -416,8 +416,10 @@ public class TXCommitMessage extends PooledDistributionMessage
                 setRecipientsSendData(Collections.singleton(indivRecip.next()), processor, rcl);
               }
             } else {
-              Set<InternalDistributedMember> tempNotificationOnlyMembers = new HashSet();
-              Set<InternalDistributedMember> tempTransactionMembers = new HashSet();
+              Set<InternalDistributedMember> tempNotificationOnlyMembers =
+                  new HashSet<InternalDistributedMember>();
+              Set<InternalDistributedMember> tempTransactionMembers =
+                  new HashSet<InternalDistributedMember>();
               if (!rcl.isEmpty() && !getNotificationOnlyMembers().isEmpty()) {
                 for (RegionCommit rc : rcl) {
                   Set<InternalDistributedMember> getNOM = getNotificationOnlyMembers().get(rc);
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/TXCommitMessage.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/TXCommitMessage.java
@@ -127,6 +127,8 @@ public class TXCommitMessage extends PooledDistributionMessage
    */
   private transient boolean hasReliableRegions = false;
 
+  public transient Set<InternalDistributedMember> notificationOnlyMembers;
+
   /**
    * Set of all caching exceptions produced while processing this tx
    */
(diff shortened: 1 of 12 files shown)


rhoughton-pivot (approved):

kirklund:
Can you please change the PR title from `Feature/geode 9248` to something more descriptive like:
```
GEODE-9248: Prevent overflowing WAN queue during split-brain
```
Or whatever makes more sense. And then fill in the PR description with details about why this an issue and what the PR does to fix it.

DonalEvans (changes requested):

geode-core/src/main/java/org/apache/geode/internal/cache/TXCommitMessage.java
```diff
@@ -404,15 +402,25 @@ void send(TXLockId lockId) {
                 setRecipientsSendData(Collections.singleton(indivRecip.next()), processor, rcl);
               }
             } else {
-              if (this.notificationOnlyMembers.isEmpty()) {
+              HashSet tempNotificationOnlyMembers = new HashSet();
+              if (!rcl.isEmpty() && !getNotificationOnlyMembers().isEmpty()) {
+                for (RegionCommit rc : rcl) {
+                  Set getNOM = getNotificationOnlyMembers().get(rc);
```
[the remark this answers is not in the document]
mivanac: Updated

geode-core/src/main/java/org/apache/geode/internal/cache/TXRegionState.java
```diff
@@ -319,14 +319,17 @@ void buildMessage(InternalRegion r, TXCommitMessage msg) {
     try {
       if (!r.getScope().isLocal() && !this.entryMods.isEmpty()) {
 
-        msg.startRegion(r, entryMods.size());
+        TXCommitMessage.RegionCommit regionCommit = msg.startRegion(r, entryMods.size());
         Iterator it = this.entryMods.entrySet().iterator();
         Set<InternalDistributedMember> newMemberSet = new HashSet<InternalDistributedMember>();
+        Set<InternalDistributedMember> redundantMemberSet =
+            new HashSet<InternalDistributedMember>();
```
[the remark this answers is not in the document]
mivanac: Updated

geode-core/src/main/java/org/apache/geode/internal/cache/TXRegionState.java
```diff
@@ -335,9 +338,16 @@ void buildMessage(InternalRegion r, TXCommitMessage msg) {
           TXEntryState txes = (TXEntryState) me.getValue();
           txes.buildMessage(r, eKey, msg, this.otherMembers);
           if (txes.getFilterRoutingInfo() != null) {
-            msg.setNotificationOnlyMembers(txes.getAdjunctRecipients());
+            HashSet tempSet = new HashSet(txes.getAdjunctRecipients());
```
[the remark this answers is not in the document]
mivanac: Updated

geode-core/src/main/java/org/apache/geode/internal/cache/TXRegionState.java
```diff
@@ -335,16 +345,26 @@ void buildMessage(InternalRegion r, TXCommitMessage msg) {
...
+            }
+            // exclude members that actually host targeted bucket from notification only list
+            tempSet.removeAll(redundantMemberSet);
+            if (!tempSet.isEmpty()) {
+              if (msg.getNotificationOnlyMembers().get(regionCommit) == null) {
+                msg.getNotificationOnlyMembers().put(regionCommit, tempSet);
+              } else {
+                msg.getNotificationOnlyMembers().get(regionCommit).addAll(tempSet);
+              }
+            }
```
DonalEvans, line 360: If the value returned from `txes.getAdjunctRecipients()` is null, then we know that `tempSet` will be empty, so we can skip the subsequent `removeAll()` and `isEmpty()` checks, because we know the outcome of them. With that in mind, this block can be refactored to:
```
            if (txes.getAdjunctRecipients() != null) {
              Set<InternalDistributedMember> tempSet = new HashSet<>(txes.getAdjunctRecipients());
              // exclude members that actually host targeted bucket from notification only list
              tempSet.removeAll(redundantMemberSet);
              if (!tempSet.isEmpty()) {
                if (msg.getNotificationOnlyMembers().get(regionCommit) == null) {
                  msg.getNotificationOnlyMembers().put(regionCommit, tempSet);
                } else {
                  msg.getNotificationOnlyMembers().get(regionCommit).addAll(tempSet);
                }
              }
            }
            newMemberSet.addAll(txes.getFilterRoutingInfo().getMembers());
```

geode-core/src/main/java/org/apache/geode/internal/cache/TXRegionState.java
```diff
@@ -335,16 +345,26 @@ void buildMessage(InternalRegion r, TXCommitMessage msg) {
           TXEntryState txes = (TXEntryState) me.getValue();
           txes.buildMessage(r, eKey, msg, this.otherMembers);
           if (txes.getFilterRoutingInfo() != null) {
+            Set<InternalDistributedMember> tempSet = new HashSet();
```
DonalEvans, line 348: This should also be `new HashSet<>();`

geode-core/src/main/java/org/apache/geode/internal/cache/TXRegionState.java
```diff
@@ -319,14 +319,24 @@ void buildMessage(InternalRegion r, TXCommitMessage msg) {
...
-        Set<InternalDistributedMember> newMemberSet = new HashSet<InternalDistributedMember>();
+        Set<InternalDistributedMember> newMemberSet = new HashSet();
+        Set<InternalDistributedMember> redundantMemberSet =
+            new HashSet();
 
         if (r.getScope().isDistributed()) {
           DistributedRegion dr = (DistributedRegion) r;
           msg.addViewVersion(dr, dr.getDistributionAdvisor().startOperation());
           newMemberSet.addAll(dr.getCacheDistributionAdvisor().adviseTX());
+          redundantMemberSet.addAll(dr.getCacheDistributionAdvisor().adviseTX());
```
DonalEvans, line 332: Rather than two calls to `dr.getCacheDistributionAdvisor().adviseTX()`, could this be extracted to a local variable?

geode-core/src/main/java/org/apache/geode/internal/cache/TXRegionState.java
```diff
@@ -319,14 +319,24 @@ void buildMessage(InternalRegion r, TXCommitMessage msg) {
     try {
       if (!r.getScope().isLocal() && !this.entryMods.isEmpty()) {
 
-        msg.startRegion(r, entryMods.size());
+        TXCommitMessage.RegionCommit regionCommit = msg.startRegion(r, entryMods.size());
         Iterator it = this.entryMods.entrySet().iterator();
```
DonalEvans, line 323: Can this iterator be moved down to above the while loop that uses it, to make things a little clearer?

DonalEvans (changes requested):

geode-core/src/main/java/org/apache/geode/internal/cache/TXCommitMessage.java
```diff
@@ -416,8 +416,10 @@ void send(TXLockId lockId) {
                 setRecipientsSendData(Collections.singleton(indivRecip.next()), processor, rcl);
               }
             } else {
-              Set<InternalDistributedMember> tempNotificationOnlyMembers = new HashSet();
-              Set<InternalDistributedMember> tempTransactionMembers = new HashSet();
+              Set<InternalDistributedMember> tempNotificationOnlyMembers =
+                  new HashSet<InternalDistributedMember>();
+              Set<InternalDistributedMember> tempTransactionMembers =
+                  new HashSet<InternalDistributedMember>();
```
DonalEvans, line 422: These should just be `new HashSet<>();` with the angle brackets present but empty. It's not necessary to specify the type as part of the constructor, as it's implicit from the `Set<InternalDistributedMember>` declaration, but it's still necessary to use empty angle brackets to show that the `HashSet` is parameterized. This comment also applies to the other changes in this commit.

DonalEvans (changes requested):

geode-core/src/main/java/org/apache/geode/internal/cache/TXCommitMessage.java
mivanac: I hope that now is OK.

DonalEvans (approved):

geode-core/src/main/java/org/apache/geode/internal/cache/TXCommitMessage.java
DonalEvans: Yes, thank you, and sorry for the confusion.

geode-core/src/main/java/org/apache/geode/internal/cache/TXRegionState.java
```diff
@@ -319,32 +319,55 @@ void buildMessage(InternalRegion r, TXCommitMessage msg) {
...
 
+        Iterator it = this.entryMods.entrySet().iterator();
+
         while (it.hasNext()) {
           Map.Entry me = (Map.Entry) it.next();
           Object eKey = me.getKey();
           TXEntryState txes = (TXEntryState) me.getValue();
           txes.buildMessage(r, eKey, msg, this.otherMembers);
           if (txes.getFilterRoutingInfo() != null) {
+            Set<InternalDistributedMember> tempSet = new HashSet<>();
```
gesterzhou, line 350: the tempSet should be renamed to be tempNotificationOnlyMembers.

gesterzhou (approved):

DonalEvans:
Could you please provide some clarification on what the latest commit is doing?

geode-core/src/main/java/org/apache/geode/internal/cache/TXCommitMessage.java
```diff
@@ -391,9 +416,34 @@ void send(TXLockId lockId) {
...
+                  }
+                }
+                tempNotificationOnlyMembers.removeAll(tempTransactionMembers);
+              }
+
+              if (tempNotificationOnlyMembers.isEmpty()) {
+                // Run in normal mode sending to multiple recipients in one shot
+                setRecipientsSendData(recipients, processor, rcl);
+              } else {
+                recipients.removeAll(tempNotificationOnlyMembers);
```
pivotal-eshu, line 450: This still has the same issue as multiple TxCommitMessages (for different rcl) being sent to the same remote server. 
Let's say there are only 2 rcls in the tx. First rcl contains 1 RegionCommit on bucket_1. TxCommitMessage (containing the rcl) should be sent to remote server1 as redundant server. (assume server 2 will be the notificationOnlyMember due to cq/subscription etc). This will cause on first rcl being sent to both server1 and server2.

Second rcl also contains 1 RegionCommit on bucket_2. Server 2 is the redundant server, and server 1 needs to be set as notification only member. This will cause the second TxCommitMessage (containing second rcl) being sent to both server1 and server2. 

As a result, some of the RCLs is not being processed - causing data inconsistency. Each server should only receive one TxCommitMessage for a tx.


When processing second rcl (

pivotal-eshu (changes requested):

geode-core/src/main/java/org/apache/geode/internal/cache/TXCommitMessage.java
mivanac: It seems, that this approach to solve this issue on sending side is generating more and more patches. Maybe this shoulb be solved on receiving side as @jvarenina proposed in https://github.com/apache/geode/pull/6477#issuecomment-848723153

mivanac: I have found what was the problem, and pushed fix, but should we also consider other alternative, mentioned above.

pivotal-eshu: I think the issue is that the fix was based on the region commits list (for different receivers).
I think you may want to consider to use the information available in tx - for each entry event inside the tx.
FilterRoutingInfo filterRoutingInfo is sent to remote receivers through FarSideEntryOp. It can be queried for the adjunct servers. It is calculated on the primary (tx host node).

At remote receiver site, txCallbackEventFactory.createCallbackEvent is invoked to created the callback event to be used for wan etc. You can use the following information to determine whether to set the tailKey for this event.
txEntryState == null indicates this is a farsider receiver (not host)
eventRegion.isUsedForPartitionedRegionBucket() to determine if a bucket region (before setting as a pr).
use the filterRoutingInfo.getMembers to determine the receive is one of the adjunctServers -- to determine if tailKey should be reset.

jvarenina: Hi @pivotal-eshu ,

Thank you for the help on this issue. We really appreciate it.
I have create the draft PR with the solution that it based on what you proposed in above comment: https://github.com/apache/geode/pull/6646

Could you please check it and comment?

Just to share with you why I haven't used some of the proposed checks in the solution:

`internalRegion.getCacheDistributionAdvisor().adviseTX() `-> I haven't used this one because adviseTX only works with BucketRegion. Since bucket region is not available on adjunct server it not possible to get primary and redundant servers with this function there.

`txEntryState == null` -> Not sure that this is needed because impacted part of code is executed only on farside.

`internalRegion.isUsedForPartitionedRegionBucket() `-> I haven't used this check because it returns "false" on adjunct member. I think that this is already checked with `this.internalRegion instanceof PartitionedRegion` few lines before the impact. When server hosts bucket regions then `internalRegion` will be of type `BucketRegion`.

mivanac:
Solution for this issue will be solved with https://github.com/apache/geode/pull/6646

PR #6628, opened 2021-06-17
Do Not Review: trying out in distributed tests and elsewhere

mhansonp:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/distributed/internal/InternalDistributedSystem.java
+++ b/geode-core/src/main/java/org/apache/geode/distributed/internal/InternalDistributedSystem.java
@@ -1866,8 +1866,7 @@ public class InternalDistributedSystem extends DistributedSystem
 
     // Search through the set of all members
     for (InternalDistributedMember member : allMembers) {
-
-      Set<InetAddress> equivalentAddresses = dm.getEquivalents(member.getInetAddress());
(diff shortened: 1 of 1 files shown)


geode-core/src/main/java/org/apache/geode/distributed/internal/InternalDistributedSystem.java
```diff
@@ -1866,8 +1866,7 @@ public InternalDistributedMember getDistributedMember() {
 
     // Search through the set of all members
     for (InternalDistributedMember member : allMembers) {
-
-      Set<InetAddress> equivalentAddresses = dm.getEquivalents(member.getInetAddress());
+      Set<InetAddress> equivalentAddresses = dm.getEquivalents(address);
```
dschneider-pivotal, line 1869: This change looks wrong to me. My understanding of this method is that it is supposed to return all the members that use then given "address". It does this by iterating over the members and checking, first, is address equal to the member's address, and second, is address in the member's equivalent addresses. With your change the second condition should always be true since you are now getting address's equivalents and then checking if address is in them. That should always be true.
At least on the open side I only see this method called from a single test: DistributedMemberDUnitTest#testFindMemberByName. If that is the only use then it seems like you could delete this test (which seems to be testing this method) and delete this method from the product (since it no longer uses it).

PR #6629, opened 2021-06-18
GEODE-9384: Fix exception messages in PdxToJSON and JsonWriter

todmorrison:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/pdx/internal/json/PdxToJSON.java
+++ b/geode-core/src/main/java/org/apache/geode/pdx/internal/json/PdxToJSON.java
@@ -135,17 +135,16 @@ public class PdxToJSON {
       jg.writeString(value.toString());
     } else if (value.getClass().equals(PdxInstanceEnumInfo.class)) {
       jg.writeString(value.toString());
+    } else if (value instanceof PdxInstance) {
+      getJSONString(jg, (PdxInstance) value);
+    } else if (value instanceof Collection) {
+      getJSONStringFromCollection(jg, (Collection<?>) value, pf);
+    } else if (value instanceof Map) {
+      getJSONStringFromMap(jg, (Map) value, pf);
     } else {
-      if (value instanceof PdxInstance) {
-        getJSONString(jg, (PdxInstance) value);
-      } else if (value instanceof Collection) {
-        getJSONStringFromCollection(jg, (Collection<?>) value, pf);
-      } else if (value instanceof Map) {
-        getJSONStringFromMap(jg, (Map) value, pf);
-      } else {
-        throw new IllegalStateException(
-            "PdxInstance returns unknwon pdxfield " + pf + " for type " + value);
-      }
+      throw new IllegalStateException(
+          "PdxInstance returns unsupported type " + value.getClass()
+              + " for field " + pf + " = " + value);
     }
   }
 
@@ -226,7 +225,8 @@ public class PdxToJSON {
       jg.writeEndArray();
     } else {
       throw new IllegalStateException(
-          "PdxInstance returns unknown pdxfield " + pf + " for type " + value);
+          "PdxInstance returns unsupported type " + value.getClass()
+              + " for field " + pf + " = " + value);
     }
   }
 
--- a/geode-core/src/main/java/org/apache/geode/pdx/internal/json/PdxToJSON.java
+++ b/geode-core/src/main/java/org/apache/geode/pdx/internal/json/PdxToJSON.java
@@ -185,7 +185,11 @@ public class PdxToJSON {
   protected void getJSONStringFromArray(JsonGenerator jg, Object value, String pf)
       throws JsonGenerationException, IOException {
 
-    if (value.getClass().getName().equals("[Z")) {
+    if (!value.getClass().isArray()) {
+      throw new IllegalStateException(
+          "Expected an array for pdx field " + pf + ", but got an object of type "
+              + value.getClass());
+    } else if (value.getClass().getName().equals("[Z")) {
       JsonHelper.getJsonFromPrimitiveBoolArray(jg, (boolean[]) value, pf);
     } else if (value.getClass().getName().equals("[B")) {
       JsonHelper.getJsonFromPrimitiveByteArray(jg, (byte[]) value, pf);
@@ -226,15 +230,11 @@ public class PdxToJSON {
         writeValue(jg, obj, pf);
       }
       jg.writeEndArray();
-    } else if (value.getClass().isArray()) {
+    } else {
       throw new IllegalStateException(
           "The pdx field " + pf + " is an array whose component type "
               + value.getClass().getComponentType()
               + " can not be converted to JSON.");
-    } else {
-      throw new IllegalStateException(
-          "Expected an array for pdx field " + pf + ", but got an object of type "
-              + value.getClass());
     }
   }
 
--- a/geode-core/src/main/java/org/apache/geode/pdx/internal/json/PdxToJSON.java
+++ b/geode-core/src/main/java/org/apache/geode/pdx/internal/json/PdxToJSON.java
@@ -28,6 +28,7 @@ import com.fasterxml.jackson.core.JsonGenerationException;
 import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.core.JsonGenerator.Feature;
 
+import org.apache.geode.annotations.VisibleForTesting;
 import org.apache.geode.annotations.internal.MutableForTesting;
 import org.apache.geode.internal.HeapDataOutputStream;
 import org.apache.geode.internal.serialization.KnownVersion;
@@ -94,7 +95,8 @@ public class PdxToJSON {
     return (pdxObj != null ? JSONFormatter.toJSON(pdxObj) : null);
   }
 
-  private void writeValue(JsonGenerator jg, Object value, String pf)
+  @VisibleForTesting
+  protected void writeValue(JsonGenerator jg, Object value, String pf)
       throws JsonGenerationException, IOException {
 
     if (value == null) {
@@ -143,8 +145,8 @@ public class PdxToJSON {
       getJSONStringFromMap(jg, (Map) value, pf);
     } else {
       throw new IllegalStateException(
-          "PdxInstance returns unsupported type " + value.getClass()
-              + " for field " + pf + " = " + value);
+          "The pdx field " + pf + " has a value " + value + " whose type " + value.getClass()
+              + " can not be converted to JSON.");
     }
   }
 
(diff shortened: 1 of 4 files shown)


geode-web-api/src/main/java/org/apache/geode/rest/internal/web/util/JsonWriter.java
```diff
@@ -216,7 +216,8 @@ public static void writeArrayAsJson(JsonGenerator generator, Object value, Strin
       writeObjectArrayAsJson(generator, (Object[]) value, pdxField);
     } else {
       throw new IllegalStateException(
-          "PdxInstance returns unknwon pdxfield " + pdxField + " for type " + value);
+          "PdxInstance returns unsupported type " + value.getClass()
```
dschneider-pivotal, line 219: pdxField can be null. Are you okay with this message just saying "for field null" in that case? I think this message should say that it is trying to write an array as JSON and found a "value" whose "type" can not be written as an array. The only caller I found of it already checked that value.getClass().isArray() is true so we just have an array that does not match all the checks this code does. I think you should test this with an array that will fail instead of a Date. Try a Date[]. 
By the way, the last "else if" on line 215 looks buggy to me. If you look at writeObjectArrayAsJson you will see it just iterates the array and delegates to writeValueAsJson which supports many more types than are check for here (for example an array of Link or an array of Map). If your array was declared as Object[], you could then put Link instances into it and this code would call writeObjectArrayAsJson. But if instead it declared it as Link[] I think this code would throw. It seems like it should work with any type of array as long as it contains values that writeValueAsJson can handle. But I'm not sure how the other direction works (going from JSON back to a java array).

geode-web-api/src/test/java/org/apache/geode/rest/internal/web/util/JsonWriterTest.java
```diff
@@ -0,0 +1,41 @@
...
+
+@Category({RestAPITest.class})
+public class JsonWriterTest {
+  @Test
+  public void testWriteValueAsJsonException() throws IOException {
+    try {
+      JsonWriter.writeArrayAsJson(mock(JsonGenerator.class), new Date(), "test");
+      fail("should not succeed");
+    } catch (IllegalStateException e) {
+      assertThat(e.getMessage()).contains("java.util.Date");
```
dschneider-pivotal, line 38: you should be more specific about the message since that is the main thing you are testing here

geode-core/src/main/java/org/apache/geode/pdx/internal/json/PdxToJSON.java
```diff
@@ -226,7 +225,8 @@ private void getJSONStringFromArray(JsonGenerator jg, Object value, String pf)
       jg.writeEndArray();
     } else {
       throw new IllegalStateException(
-          "PdxInstance returns unknown pdxfield " + pf + " for type " + value);
+          "PdxInstance returns unsupported type " + value.getClass()
```
dschneider-pivotal, line 228: The caller already made sure value.getClass().isArray() was true. So this message should say something about the array element type not being supported by JSON. Given a Class instance that isArray() you can call getComponentType() it to get the class that describes the element type. The message could say that the componentType is not supported for JSON. Also keep in mind that "value" is an array and toString on an array is not very helpful (just prints out the class name and its hashCode but not the elements of the array).
Also you need a unit test for this one

geode-core/src/main/java/org/apache/geode/pdx/internal/json/PdxToJSON.java
```diff
@@ -135,17 +135,16 @@ private void writeValue(JsonGenerator jg, Object value, String pf)
...
-      } else if (value instanceof Collection) {
-        getJSONStringFromCollection(jg, (Collection<?>) value, pf);
-      } else if (value instanceof Map) {
-        getJSONStringFromMap(jg, (Map) value, pf);
-      } else {
-        throw new IllegalStateException(
-            "PdxInstance returns unknwon pdxfield " + pf + " for type " + value);
-      }
+      throw new IllegalStateException(
+          "PdxInstance returns unsupported type " + value.getClass()
```
dschneider-pivotal, line 146: I think this message should make clear what we mean by "unsupported". Pdx supports it but we don't support converting it to JSON. So something like: "The pdx field 'FIELD_NAME' has a value whose type 'TYPE' can not be converted to JSON. The value is 'VALUE'."
You need to add a unit test for this message

dschneider-pivotal (changes requested):

geode-web-api/src/main/java/org/apache/geode/rest/internal/web/util/JsonWriter.java
todmorrison: I am okay with just saying "for field null" if the field doesn't have a name. I've changed the message to:
`          "The pdx field " + pf + " is an array whose type " + value.getClass()
              + " can not be converted to JSON.");
`
Does that work?

As for the last "else if" on line 215, I agree it could be improved (similarly GEODE-9392 suggests there may be room to improve PdxToJSON.writeValue as well), but I'm trying to limit the scope here to making the existing messages more useful.

geode-core/src/main/java/org/apache/geode/pdx/internal/json/PdxToJSON.java
todmorrison: Agreed, though I went with a single sentence:
`          "The pdx field " + pf + " has a value " + value + " whose type " + value.getClass()
              + " can not be converted to JSON.");
`

geode-core/src/main/java/org/apache/geode/pdx/internal/json/PdxToJSON.java
todmorrison: Does this work?
```
          "The pdx field " + pf + " is an array whose component type "
              + value.getClass().getComponentType()
              + " can not be converted to JSON.");
```
I'm not sure I like trusting that the caller made sure that value.getClass().isArray() was true though.

todmorrison: *I* like this better:
```
    } else if (value.getClass().isArray()) {
      throw new IllegalStateException(
          "The pdx field " + pf + " is an array whose component type "
              + value.getClass().getComponentType()
              + " can not be converted to JSON.");
    } else {
      throw new IllegalStateException(
          "Expected an array for pdx field " + pf + ", but got an object of type "
              + value.getClass());
    }
```

dschneider-pivotal: looks good

geode-core/src/main/java/org/apache/geode/pdx/internal/json/PdxToJSON.java
dschneider-pivotal: nice

geode-web-api/src/main/java/org/apache/geode/rest/internal/web/util/JsonWriter.java
dschneider-pivotal: I like your new messages

geode-web-api/src/test/java/org/apache/geode/rest/internal/web/util/JsonWriterTest.java
todmorrison: Agreed and done.

geode-core/src/main/java/org/apache/geode/pdx/internal/json/PdxToJSON.java
todmorrison: Unit test added

geode-core/src/main/java/org/apache/geode/pdx/internal/json/PdxToJSON.java
```diff
@@ -224,9 +226,15 @@ private void getJSONStringFromArray(JsonGenerator jg, Object value, String pf)
...
       jg.writeEndArray();
+    } else if (value.getClass().isArray()) {
+      throw new IllegalStateException(
+          "The pdx field " + pf + " is an array whose component type "
+              + value.getClass().getComponentType()
+              + " can not be converted to JSON.");
     } else {
       throw new IllegalStateException(
-          "PdxInstance returns unknown pdxfield " + pf + " for type " + value);
+          "Expected an array for pdx field " + pf + ", but got an object of type "
```
jinmeiliao, line 236: since only array object will call this method, this "else" statement will never be executed. I think this if/else is redundant.

geode-web-api/src/test/java/org/apache/geode/rest/internal/web/util/JsonWriterTest.java
```diff
@@ -0,0 +1,54 @@
...
+import org.junit.experimental.categories.Category;
+
+import org.apache.geode.test.junit.categories.RestAPITest;
+
+@Category({RestAPITest.class})
+public class JsonWriterTest {
+  @Test
+  public void testWwriteArrayAsJson() throws IOException {
+    Date value = new Date();
+    try {
```
jinmeiliao, line 35: it's probably cleaner to use 'assertThatThrownBy` method

jinmeiliao (changes requested):

geode-core/src/main/java/org/apache/geode/pdx/internal/json/PdxToJSON.java
todmorrison: As in my reply to Darrel's similar earlier comment, I'm not sure I like trusting that the caller made sure that value.getClass().isArray() was true and there's nothing about the method signature (and I don't believe can be) that would prevent this, so my instinct is to go with "defensive coding". (What's "safe" and "obvious" today can get lost in future changes.)

jinmeiliao: Yeah, that's a good point if it's a public api, but this method was intended to be "private". 
I am fine leaving it as is though.

jinmeiliao (approved):

geode-core/src/main/java/org/apache/geode/pdx/internal/json/PdxToJSON.java
todmorrison: Thanks, I did go back and forth about it, since it is a private method, but finally settled on playing it safe since even within a class when the lines of code increases assumptions can get lost. I do agree that the isArrary check should come first and, since I was already going to be doing another commit, went ahead and made the change.

geode-web-api/src/test/java/org/apache/geode/rest/internal/web/util/JsonWriterTest.java
todmorrison: Is this the type of thing you're recommending:
```    assertThatThrownBy(
        () -> JsonWriter.writeArrayAsJson(mock(JsonGenerator.class), value, "myDate"))
            .isInstanceOf(IllegalStateException.class)
            .hasMessage("Expected an array for pdx field myDate, but got an object of type "
              + value.getClass());
```

todmorrison: Done.

jinmeiliao (approved):

mkevo (approved):

PR #6631, opened 2021-06-21
GEODE-9293: Add cause of error to GemFireIOException messages

alb3rtobr:
**AS A** geode user
**I WANT TO** have more information when an IO exception occurs
**SO THAT** I can understand and fix the root cause of the problem

`GemFireIOException` is used to encapsulate a given `IOException` that happen in the Geode code.
For creating a `GemFireIOException` when a `IOException` is catched, this constructor is used:

```
public GemFireIOException(String message, Throwable cause)  {
    super(message, cause);
}
```

The problem is that the `IOException` message is not part of the `GemFireIOException` message. So when a `GemFireIOException` is logged, the root cause of the problem is not shown.

For example, the following message log message:

```
"Cache initialization for GemFireCache[id = 1081136680; isClosing = false; isShutDownAll = false; created = Fri Jun 19 11:42:29 UTC 2020; server = false; copyOnRead = false; lockLease = 120; lockTimeout = 60] failed because: org.apache.geode.GemFireIOException: While starting cache server CacheServer on port=40404 client subscription config policy=none client subscription config capacity=1 client subscription config overflow directory=."
```

It was logged at `GemFireCacheImpl`:

```
logger.error("Cache initialization for " + toString() + " failed because:", throwable);
```

And the `GemFireIOException` was created in `CacheCreation` in this way:

```
} catch (IOException ex) {
     throw new GemFireIOException(format("While starting cache server %s", impl), ex);
}
```

This PR adds the wrapped exception message to the `GemFireIOException` exception message, so when the exception is catched and logged, the root cause will be known at first sight.

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/GemFireIOException.java
+++ b/geode-core/src/main/java/org/apache/geode/GemFireIOException.java
@@ -30,7 +30,9 @@ public class GemFireIOException extends GemFireException {
    * Creates a new <code>GemFireIOException</code>.
    */
   public GemFireIOException(String message, Throwable cause) {
-    super(message, cause);
+    super(message
+        + ((cause != null && cause.getMessage() != null) ? " ( " + cause.getMessage() + " )" : ""),
+        cause);
   }
 
   public GemFireIOException(String message) {
--- a/geode-core/src/test/java/org/apache/geode/GemFireIOExceptionTest.java
+++ b/geode-core/src/test/java/org/apache/geode/GemFireIOExceptionTest.java
@@ -1,0 +1,47 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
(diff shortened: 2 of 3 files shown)


DonalEvans:
Correct me if I'm wrong, but isn't the exception message that you're adding to the `GemFireIOException` already visible in the exception stack trace? With the changes here:
```
org.apache.geode.SerializationException: Serialization failed ( java.lang.Object )
	at org.apache.geode.internal.cache.GemFireCacheImpl.registerPdxMetaData(GemFireCacheImpl.java:5156)
	...
Caused by: java.io.NotSerializableException: java.lang.Object <<< HERE
	at java.io.ObjectOutputStream.writeObject0(ObjectOutputStream.java:1184)
	...
```

and without the changes:
```
org.apache.geode.SerializationException: Serialization failed
	at org.apache.geode.internal.cache.GemFireCacheImpl.registerPdxMetaData(GemFireCacheImpl.java:5156)
	...
Caused by: java.io.NotSerializableException: java.lang.Object <<< HERE
	at java.io.ObjectOutputStream.writeObject0(ObjectOutputStream.java:1184)
	...
```
Are there places where the logged exception doesn't contain the exception stack? If so, I would suggest that the problem is with those places, not with the exception itself.

alb3rtobr:
Thanks @DonalEvans , you are right, I didnt understand the problem correctly. Anyway, I have realized the problem of the missing stacktrace was already solved by GEODE-8124, so Im closing this PR.

PR #6636, opened 2021-06-22
Revert "GEODE-9139: SSLException in starting up a Locator"

onichols-pivotal:
This reverts the merge of #6436 (this fix was only intended for 1.14 and is causing problems in 1.13)


onichols-pivotal:
dup of #6637

PR #6641, opened 2021-06-24
GEODE-9354: Extract and test ArgumentRedactorRegex

kirklund:
1) get ArgumentRedactor under better unit testing
2) refactor ArgumentRedactor: isolate parts of it for better testing, use strategy to try out impls using String and CLI libraries
3) fix lots of issues in regex found by the new unit testing


pivotal-jbarrett (approved):

dschneider-pivotal (approved):

echobravopapa (approved):

Bill (approved):
Approve.

boglesby (approved):

jinmeiliao (approved):

agingade (approved):

mhansonp (approved):
Seems like you should/could have a "sensitive string list" that stores your test strings in one place and uses it everywhere, the reason I say that is that you have a lot of tests that test one string and it might be more interesting to just have a single list, so you modify the list and it runs against both sets of tests automatically.

PR #6658, opened 2021-06-28
GEODE-9342: Handle a double reply of "inf" in the same way as Redis

ringles:
Updates Coder.java to convert doubles with the value of Double.POSITIVE_INFINITY as "inf" and Double.NEGATIVE_INFINITY as "-inf". This matches native Redis behavior.


nonbinaryprogrammer (approved):

DonalEvans (commented):
I think it would also be good to change the `Coder.stringToDouble()` method to reflect these new "official" ways of referring to infinity, since currently it checks "+inf" rather than "inf".

onichols-pivotal:
please include the GEODE ticket in the commit message next time

PR #6659, opened 2021-06-28
GEODE-9404: Do not log error message if sender is not configured.

pivotal-eshu:
* This is normal case for serial wan configuration. Error message should not be
    logged when executing transactions.

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/TXCommitMessage.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/TXCommitMessage.java
@@ -733,7 +733,9 @@ public class TXCommitMessage extends PooledDistributionMessage
     try {
       lastTransactionEvent = getLastTransactionEvent(callbacks);
     } catch (ServiceConfigurationError ex) {
-      logger.error(ex.getMessage());
+      if (!ex.getMessage().startsWith(TXLastEventInTransactionUtils.NO_INFORMATION_FOR_SENDER_ID)) {
+        logger.error(ex.getMessage());
+      }
       isConfigError = true;
     }
 
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/TXCommitMessage.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/TXCommitMessage.java
@@ -733,9 +733,7 @@ public class TXCommitMessage extends PooledDistributionMessage
     try {
       lastTransactionEvent = getLastTransactionEvent(callbacks);
     } catch (ServiceConfigurationError ex) {
-      if (!ex.getMessage().startsWith(TXLastEventInTransactionUtils.NO_INFORMATION_FOR_SENDER_ID)) {
-        logger.error(ex.getMessage());
-      }
+      logger.error(ex.getMessage());
       isConfigError = true;
     }
 
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/TXCommitMessage.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/TXCommitMessage.java
@@ -738,7 +738,8 @@ public class TXCommitMessage extends PooledDistributionMessage
     }
 
     for (EntryEventImpl ee : callbacks) {
-      boolean isLastTransactionEvent = isConfigError || ee.equals(lastTransactionEvent);
+      boolean isLastTransactionEvent =
+          isConfigError || lastTransactionEvent == null || ee.equals(lastTransactionEvent);
       try {
         if (ee.getOperation().isDestroy()) {
           ee.getRegion().invokeTXCallbacks(EnumListenerEvent.AFTER_DESTROY, ee, true,
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/TXCommitMessage.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/TXCommitMessage.java
@@ -738,8 +738,8 @@ public class TXCommitMessage extends PooledDistributionMessage
     }
 
     for (EntryEventImpl ee : callbacks) {
-      boolean isLastTransactionEvent =
-          isConfigError || lastTransactionEvent == null || ee.equals(lastTransactionEvent);
+      boolean isLastTransactionEvent = TXLastEventInTransactionUtils
+          .isLastTransactionEvent(isConfigError, lastTransactionEvent, ee);
       try {
         if (ee.getOperation().isDestroy()) {
           ee.getRegion().invokeTXCallbacks(EnumListenerEvent.AFTER_DESTROY, ee, true,
@@ -766,7 +766,8 @@ public class TXCommitMessage extends PooledDistributionMessage
   }
 
   EntryEventImpl getLastTransactionEvent(List<EntryEventImpl> callbacks) {
-    return TXLastEventInTransactionUtils.getLastTransactionEvent(callbacks, dm.getCache());
+    return TXLastEventInTransactionUtils.getLastTransactionEventInGroupedTxForWANSender(callbacks,
+        dm.getCache());
   }
 
   protected void processCacheRuntimeException(CacheRuntimeException problem) {
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/TXLastEventInTransactionUtils.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/TXLastEventInTransactionUtils.java
@@ -24,7 +24,7 @@ import org.apache.geode.cache.Cache;
 import org.apache.geode.cache.wan.GatewaySender;
 
 public class TXLastEventInTransactionUtils {
-  public static final String NO_INFORMATION_FOR_SENDER_ID = "No information for senderId: ";
+  static final String NO_INFORMATION_FOR_SENDER_ID = "No information for senderId: ";
 
 
   /**
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/TXLastEventInTransactionUtils.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/TXLastEventInTransactionUtils.java
@@ -20,16 +20,10 @@ import java.util.ServiceConfigurationError;
 import java.util.Set;
 import java.util.stream.Collectors;
 
-import org.apache.logging.log4j.Logger;
-
 import org.apache.geode.cache.Cache;
 import org.apache.geode.cache.wan.GatewaySender;
-import org.apache.geode.logging.internal.log4j.api.LogService;
 
 public class TXLastEventInTransactionUtils {
-
-  private static final Logger logger = LogService.getLogger();
-
   /**
    * @param callbacks list of events belonging to a transaction
    *
@@ -40,9 +34,8 @@ public class TXLastEventInTransactionUtils {
    *         events belong have different sets of senders that group transactions
    *         then it throws a ServiceConfigurationError exception.
    */
-  public static EntryEventImpl getLastTransactionEvent(List<EntryEventImpl> callbacks,
-      Cache cache)
-      throws ServiceConfigurationError {
+  public static EntryEventImpl getLastTransactionEventInGroupedTxForWANSender(
+      List<EntryEventImpl> callbacks, Cache cache) throws ServiceConfigurationError {
     if (checkNoSendersGroupTransactionEvents(callbacks, cache)) {
       return null;
     }
@@ -113,5 +106,10 @@ public class TXLastEventInTransactionUtils {
       throw new ServiceConfigurationError("No information for senderId: " + senderId);
     }
     return sender.mustGroupTransactionEvents();
+  }
+
+  static boolean isLastTransactionEvent(boolean isConfigError,
+      EntryEventImpl lastTransactionEvent, EntryEventImpl entryEvent) {
+    return isConfigError || lastTransactionEvent == null || entryEvent.equals(lastTransactionEvent);
   }
 }
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/TXLastEventInTransactionUtils.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/TXLastEventInTransactionUtils.java
@@ -20,14 +20,12 @@ import java.util.ServiceConfigurationError;
 import java.util.Set;
 import java.util.stream.Collectors;
 
-import org.apache.logging.log4j.Logger;
-
 import org.apache.geode.cache.Cache;
 import org.apache.geode.cache.wan.GatewaySender;
-import org.apache.geode.logging.internal.log4j.api.LogService;
 
 public class TXLastEventInTransactionUtils {
-  private static final Logger logger = LogService.getLogger();
+  public static final String NO_INFORMATION_FOR_SENDER_ID = "No information for senderId: ";
+
 
   /**
    * @param callbacks list of events belonging to a transaction
@@ -78,8 +76,7 @@ public class TXLastEventInTransactionUtils {
     for (String senderId : getSenderIdsForEvents(callbacks)) {
       GatewaySender sender = cache.getGatewaySender(senderId);
       if (sender == null) {
-        logger.error("No sender found for {}", senderId);
-        throw new ServiceConfigurationError("No information for senderId: " + senderId);
+        throw new ServiceConfigurationError(NO_INFORMATION_FOR_SENDER_ID + senderId);
       }
       if (sender.mustGroupTransactionEvents()) {
         return false;
@@ -113,7 +110,7 @@ public class TXLastEventInTransactionUtils {
       throws ServiceConfigurationError {
     GatewaySender sender = cache.getGatewaySender(senderId);
     if (sender == null) {
-      throw new ServiceConfigurationError("No information for senderId: " + senderId);
+      throw new ServiceConfigurationError(NO_INFORMATION_FOR_SENDER_ID + senderId);
     }
     return sender.mustGroupTransactionEvents();
   }
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/TXLastEventInTransactionUtils.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/TXLastEventInTransactionUtils.java
@@ -24,9 +24,6 @@ import org.apache.geode.cache.Cache;
 import org.apache.geode.cache.wan.GatewaySender;
 
 public class TXLastEventInTransactionUtils {
-  static final String NO_INFORMATION_FOR_SENDER_ID = "No information for senderId: ";
-
-
   /**
    * @param callbacks list of events belonging to a transaction
    *
@@ -75,10 +72,7 @@ public class TXLastEventInTransactionUtils {
       Cache cache) throws ServiceConfigurationError {
     for (String senderId : getSenderIdsForEvents(callbacks)) {
       GatewaySender sender = cache.getGatewaySender(senderId);
-      if (sender == null) {
-        throw new ServiceConfigurationError(NO_INFORMATION_FOR_SENDER_ID + senderId);
-      }
-      if (sender.mustGroupTransactionEvents()) {
+      if (sender != null && sender.mustGroupTransactionEvents()) {
         return false;
       }
     }
@@ -109,9 +103,6 @@ public class TXLastEventInTransactionUtils {
   private static boolean doesSenderGroupTransactionEvents(Cache cache, String senderId)
       throws ServiceConfigurationError {
     GatewaySender sender = cache.getGatewaySender(senderId);
-    if (sender == null) {
-      throw new ServiceConfigurationError(NO_INFORMATION_FOR_SENDER_ID + senderId);
-    }
-    return sender.mustGroupTransactionEvents();
+    return sender != null && sender.mustGroupTransactionEvents();
   }
 }
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/TXLastEventInTransactionUtils.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/TXLastEventInTransactionUtils.java
@@ -20,10 +20,16 @@ import java.util.ServiceConfigurationError;
 import java.util.Set;
 import java.util.stream.Collectors;
 
+import org.apache.logging.log4j.Logger;
+
 import org.apache.geode.cache.Cache;
 import org.apache.geode.cache.wan.GatewaySender;
+import org.apache.geode.logging.internal.log4j.api.LogService;
 
 public class TXLastEventInTransactionUtils {
+
+  private static final Logger logger = LogService.getLogger();
+
   /**
    * @param callbacks list of events belonging to a transaction
    *
@@ -103,6 +109,9 @@ public class TXLastEventInTransactionUtils {
   private static boolean doesSenderGroupTransactionEvents(Cache cache, String senderId)
       throws ServiceConfigurationError {
     GatewaySender sender = cache.getGatewaySender(senderId);
-    return sender != null && sender.mustGroupTransactionEvents();
+    if (sender == null) {
+      throw new ServiceConfigurationError("No information for sender id: " + senderId);
+    }
+    return sender.mustGroupTransactionEvents();
   }
 }
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/TXLastEventInTransactionUtils.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/TXLastEventInTransactionUtils.java
@@ -110,7 +110,7 @@ public class TXLastEventInTransactionUtils {
       throws ServiceConfigurationError {
     GatewaySender sender = cache.getGatewaySender(senderId);
     if (sender == null) {
-      throw new ServiceConfigurationError("No information for sender id: " + senderId);
+      throw new ServiceConfigurationError("No information for senderId: " + senderId);
     }
     return sender.mustGroupTransactionEvents();
   }
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/TXState.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/TXState.java
@@ -248,7 +248,9 @@ public class TXState implements TXStateInterface {
       lastTransactionEvent =
           TXLastEventInTransactionUtils.getLastTransactionEvent(getPendingCallbacks(), getCache());
     } catch (ServiceConfigurationError ex) {
-      logger.error(ex.getMessage());
+      if (!ex.getMessage().startsWith(TXLastEventInTransactionUtils.NO_INFORMATION_FOR_SENDER_ID)) {
+        logger.error(ex.getMessage());
+      }
       isConfigError = true;
     }
 
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/TXState.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/TXState.java
@@ -248,9 +248,7 @@ public class TXState implements TXStateInterface {
       lastTransactionEvent =
           TXLastEventInTransactionUtils.getLastTransactionEvent(getPendingCallbacks(), getCache());
     } catch (ServiceConfigurationError ex) {
-      if (!ex.getMessage().startsWith(TXLastEventInTransactionUtils.NO_INFORMATION_FOR_SENDER_ID)) {
-        logger.error(ex.getMessage());
-      }
+      logger.error(ex.getMessage());
       isConfigError = true;
     }
 
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/TXState.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/TXState.java
@@ -253,7 +253,8 @@ public class TXState implements TXStateInterface {
     }
 
     for (EntryEventImpl ee : getPendingCallbacks()) {
-      boolean isLastTransactionEvent = isConfigError || ee.equals(lastTransactionEvent);
+      boolean isLastTransactionEvent =
+          isConfigError || lastTransactionEvent == null || ee.equals(lastTransactionEvent);
       if (ee.getOperation().isDestroy()) {
         ee.getRegion().invokeTXCallbacks(EnumListenerEvent.AFTER_DESTROY, ee, true,
             isLastTransactionEvent);
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/TXState.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/TXState.java
@@ -245,16 +245,16 @@ public class TXState implements TXStateInterface {
     boolean isConfigError = false;
     EntryEventImpl lastTransactionEvent = null;
     try {
-      lastTransactionEvent =
-          TXLastEventInTransactionUtils.getLastTransactionEvent(getPendingCallbacks(), getCache());
+      lastTransactionEvent = TXLastEventInTransactionUtils
+          .getLastTransactionEventInGroupedTxForWANSender(getPendingCallbacks(), getCache());
     } catch (ServiceConfigurationError ex) {
       logger.error(ex.getMessage());
       isConfigError = true;
     }
 
     for (EntryEventImpl ee : getPendingCallbacks()) {
-      boolean isLastTransactionEvent =
-          isConfigError || lastTransactionEvent == null || ee.equals(lastTransactionEvent);
+      boolean isLastTransactionEvent = TXLastEventInTransactionUtils
+          .isLastTransactionEvent(isConfigError, lastTransactionEvent, ee);
       if (ee.getOperation().isDestroy()) {
         ee.getRegion().invokeTXCallbacks(EnumListenerEvent.AFTER_DESTROY, ee, true,
             isLastTransactionEvent);
--- a/geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/WANTestBase.java
+++ b/geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/WANTestBase.java
@@ -2308,7 +2308,7 @@ public class WANTestBase extends DistributedTestCase {
     IgnoredException exp2 =
         IgnoredException.addIgnoredException(GatewaySenderException.class.getName());
     try {
-      Region r = cache.getRegion(SEPARATOR + regionName);
+      Region<Object, Object> r = cache.getRegion(SEPARATOR + regionName);
       assertNotNull(r);
       for (long i = 0; i < numPuts; i++) {
         cache.getCacheTransactionManager().begin();
--- a/geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/WANTestBase.java
+++ b/geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/WANTestBase.java
@@ -2303,11 +2303,10 @@ public class WANTestBase extends DistributedTestCase {
   }
 
   public static void doTxPuts(String regionName, int numPuts) {
-    IgnoredException exp1 =
-        IgnoredException.addIgnoredException(InterruptedException.class.getName());
-    IgnoredException exp2 =
-        IgnoredException.addIgnoredException(GatewaySenderException.class.getName());
-    try {
+    try (
+        IgnoredException ignored = IgnoredException.addIgnoredException(InterruptedException.class);
+        IgnoredException ignored1 =
+            IgnoredException.addIgnoredException(GatewaySenderException.class)) {
       Region<Object, Object> r = cache.getRegion(SEPARATOR + regionName);
       assertNotNull(r);
       for (long i = 0; i < numPuts; i++) {
@@ -2315,9 +2314,6 @@ public class WANTestBase extends DistributedTestCase {
         r.put(i, "Value_" + i);
         cache.getCacheTransactionManager().commit();
       }
-    } finally {
-      exp1.remove();
-      exp2.remove();
     }
   }
 
--- a/geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/WANTestBase.java
+++ b/geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/WANTestBase.java
@@ -2302,6 +2302,25 @@ public class WANTestBase extends DistributedTestCase {
     }
   }
 
+  public static void doTxPuts(String regionName, int numPuts) {
+    IgnoredException exp1 =
+        IgnoredException.addIgnoredException(InterruptedException.class.getName());
+    IgnoredException exp2 =
+        IgnoredException.addIgnoredException(GatewaySenderException.class.getName());
+    try {
+      Region r = cache.getRegion(SEPARATOR + regionName);
+      assertNotNull(r);
+      for (long i = 0; i < numPuts; i++) {
+        cache.getCacheTransactionManager().begin();
+        r.put(i, "Value_" + i);
+        cache.getCacheTransactionManager().commit();
+      }
+    } finally {
+      exp1.remove();
+      exp2.remove();
+    }
+  }
+
   public static void doPutsSameKey(String regionName, int numPuts, String key) {
     IgnoredException exp1 =
         IgnoredException.addIgnoredException(InterruptedException.class.getName());
--- a/geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialWANPropagation_PartitionedRegionDUnitTest.java
+++ b/geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialWANPropagation_PartitionedRegionDUnitTest.java
@@ -79,6 +79,41 @@ public class SerialWANPropagation_PartitionedRegionDUnitTest extends WANTestBase
   }
 
   @Test
+  public void testPartitionedSerialPropagationWithTransaction() {
+
+    Integer lnPort = vm0.invoke(() -> WANTestBase.createFirstLocatorWithDSId(1));
+    Integer nyPort = vm1.invoke(() -> WANTestBase.createFirstRemoteLocator(2, lnPort));
+
+    createCacheInVMs(nyPort, vm2, vm3);
+    createReceiverInVMs(vm2, vm3);
+
+    createCacheInVMs(lnPort, vm4, vm5, vm6, vm7);
+
+    vm4.invoke(() -> WANTestBase.createSender("ln", 2, false, 100, 10, false, false, null, true));
+    vm5.invoke(() -> WANTestBase.createSender("ln", 2, false, 100, 10, false, false, null, true));
+
+    vm4.invoke(() -> WANTestBase.createPartitionedRegion(getTestMethodName() + "_PR", "ln", 1, 100,
+        isOffHeap()));
+    vm5.invoke(() -> WANTestBase.createPartitionedRegion(getTestMethodName() + "_PR", "ln", 1, 100,
+        isOffHeap()));
+    vm6.invoke(() -> WANTestBase.createPartitionedRegion(getTestMethodName() + "_PR", "ln", 1, 100,
+        isOffHeap()));
+    vm7.invoke(() -> WANTestBase.createPartitionedRegion(getTestMethodName() + "_PR", "ln", 1, 100,
+        isOffHeap()));
+
+    vm4.invoke(() -> WANTestBase.startSender("ln"));
+
+    vm2.invoke(() -> WANTestBase.createPartitionedRegion(getTestMethodName() + "_PR", null, 1, 100,
+        isOffHeap()));
+    vm3.invoke(() -> WANTestBase.createPartitionedRegion(getTestMethodName() + "_PR", null, 1, 100,
+        isOffHeap()));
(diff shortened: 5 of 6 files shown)


DonalEvans (commented):
It seems like the correct behaviour to not log this error message excessively, but will this change mean that we won't log an error even if none of the nodes have a sender configured?

pivotal-eshu:
> It seems like the correct behaviour to not log this error message excessively, but will this change mean that we won't log an error even if none of the nodes have a sender configured?

For serial wan -- a node performing a tx operation may not have a sender configure. I do not think it is an error.

DonalEvans:
> > It seems like the correct behaviour to not log this error message excessively, but will this change mean that we won't log an error even if none of the nodes have a sender configured?
> 
> For serial wan -- a node performing a tx operation may not have a sender configure. I do not think it is an error.

What I'm wondering about is a situation where none of the nodes have a sender configured. It's normal that a node performing a tx operation may not have a sender, but if none of the nodes in the system do, will we now not be logging an error when it would be useful to do so? Or would that situation never happen/cause other errors to be logged?

pivotal-eshu:
> > > It seems like the correct behaviour to not log this error message excessively, but will this change mean that we won't log an error even if none of the nodes have a sender configured?
> > 
> > 
> > For serial wan -- a node performing a tx operation may not have a sender configure. I do not think it is an error.
> 
> What I'm wondering about is a situation where none of the nodes have a sender configured. It's normal that a node performing a tx operation may not have a sender, but if none of the nodes in the system do, will we now not be logging an error when it would be useful to do so? Or would that situation never happen/cause other errors to be logged?

If there is no node configures the sender, the remote site will not get any data propagation. I think that is the error situation user can detect.
I think the issue (no sender created) should not rely on transaction to detect that issue. We do not log such issue in non-transactional operations.
To log such error, there should be a way from wan to determine there is no sender in the cluster. I don't think this is available.

geode-core/src/main/java/org/apache/geode/internal/cache/TXCommitMessage.java
```diff
@@ -733,7 +733,9 @@ void firePendingCallbacks(List<EntryEventImpl> callbacks) {
     try {
       lastTransactionEvent = getLastTransactionEvent(callbacks);
     } catch (ServiceConfigurationError ex) {
-      logger.error(ex.getMessage());
+      if (!ex.getMessage().startsWith(TXLastEventInTransactionUtils.NO_INFORMATION_FOR_SENDER_ID)) {
+        logger.error(ex.getMessage());
+      }
```
albertogpz, line 738: If this is not necessarily a configuration error, instead of removing the error log, I would update `TXLastEventInTransactionUtils.checkNoSendersGroupTransactionEvents()` to return `false` in case a sender is null as follows:

```
      if (sender == null || sender.mustGroupTransactionEvents()) {
        return false;
      }

```
A similar change would have to be done in `TXLastEventInTransactionUtils.doesSenderGroupTransactionEvents`:

`return sender == null || sender.mustGroupTransactionEvents();`

geode-core/src/main/java/org/apache/geode/internal/cache/TXState.java
```diff
@@ -248,7 +248,9 @@ public void firePendingCallbacks() {
       lastTransactionEvent =
           TXLastEventInTransactionUtils.getLastTransactionEvent(getPendingCallbacks(), getCache());
     } catch (ServiceConfigurationError ex) {
-      logger.error(ex.getMessage());
+      if (!ex.getMessage().startsWith(TXLastEventInTransactionUtils.NO_INFORMATION_FOR_SENDER_ID)) {
+        logger.error(ex.getMessage());
+      }
```
albertogpz, line 253: The same comment as in TXCommitMessage would apply here.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialWANPropagation_PartitionedRegionDUnitTest.java
```diff
@@ -78,6 +78,41 @@ public void testPartitionedSerialPropagation() {
     vm2.invoke(() -> WANTestBase.validateRegionSize(getTestMethodName() + "_PR", 1000));
   }
 
+  @Test
+  public void testPartitionedSerialPropagationWithTransaction() {
```
albertogpz, line 82: I would add in the test case name that the senders are not configured in all servers.

DonalEvans:
> > > > It seems like the correct behaviour to not log this error message excessively, but will this change mean that we won't log an error even if none of the nodes have a sender configured?
> > > 
> > > 
> > > For serial wan -- a node performing a tx operation may not have a sender configure. I do not think it is an error.
> > 
> > 
> > What I'm wondering about is a situation where none of the nodes have a sender configured. It's normal that a node performing a tx operation may not have a sender, but if none of the nodes in the system do, will we now not be logging an error when it would be useful to do so? Or would that situation never happen/cause other errors to be logged?
> 
> If there is no node configures the sender, the remote site will not get any data propagation. I think that is the error situation user can detect.
> I think the issue (no sender created) should not rely on transaction to detect that issue. We do not log such issue in non-transactional operations.
> To log such error, there should be a way from wan to determine there is no sender in the cluster. I don't think this is available.

Thanks for the explanation!

geode-core/src/main/java/org/apache/geode/internal/cache/TXCommitMessage.java
pivotal-eshu: When sender == null, there is no need to get the last transaction event in the node. There is no sender.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/WANTestBase.java
```diff
@@ -2302,6 +2302,25 @@ public static void doPuts(String regionName, int numPuts) {
     }
   }
 
+  public static void doTxPuts(String regionName, int numPuts) {
+    IgnoredException exp1 =
+        IgnoredException.addIgnoredException(InterruptedException.class.getName());
+    IgnoredException exp2 =
+        IgnoredException.addIgnoredException(GatewaySenderException.class.getName());
+    try {
+      Region r = cache.getRegion(SEPARATOR + regionName);
```
DonalEvans, line 2311: Warnings here can be fixed by using `Region<Object, Object>`

DonalEvans (approved):
Looks good, just one small suggestion for code clean-up.

geode-core/src/main/java/org/apache/geode/internal/cache/TXCommitMessage.java
albertogpz: There could be several senders for the same region configured in different servers.

If you opt for returning null in `getLastTransactionEvent` when one of the senders is not configured, I suggest, you 
also change `TXCommitMessage` and `TXState` so that when `getLastTransactionEvent` returns null, it considers that all events for the transaction are the last event for the transaction:
`boolean isLastTransactionEvent = isConfigError || lastTransactionEvent == null || ee.equals(lastTransactionEvent);`

The reason for it is that if another sender is configured and has group-transaction-events set to true, it will not try to get the last event from each transaction in vain delaying unnecessarily the delivery of batches.

pivotal-eshu: I have modified the code so that if the above mentioned situation will set isConfigError to be true.
You are right, lastTransactionEvent == null should be included as well.

agingade:
@pivotal-eshu  
It looks like the Tx code is trying to assess Gateway sender requirement to support GEODE-7971 in TX code. Which doesn't seems right.

The "getLastTransactionEvent" should just return the last event in the transaction; currently it is doing it by "callbacks.get(callbacks.size() - 1);" which also seems to be error prone or inefficient.  
In "firePendingCallbacks" it could  have iterated over the list using size index:
for (int i=0; i<callbaack.size(); i++){
ee.getRegion().invokeTXCallbacks(EnumListenerEvent.AFTER_DESTROY, ee, true, (i+1) == size() /*isLastTrasaction*/);
}

Please correct me if i am missing anyting.

albertogpz:
> @pivotal-eshu
> It looks like the Tx code is trying to assess Gateway sender requirement to support GEODE-7971 in TX code. Which doesn't seems right.
> 
> The "getLastTransactionEvent" should just return the last event in the transaction; currently it is doing it by "callbacks.get(callbacks.size() - 1);" which also seems to be error prone or inefficient.
> In "firePendingCallbacks" it could have iterated over the list using size index:
> for (int i=0; i<callbaack.size(); i++){
> ee.getRegion().invokeTXCallbacks(EnumListenerEvent.AFTER_DESTROY, ee, true, (i+1) == size() /_isLastTrasaction_/);
> }
> 
> Please correct me if i am missing anything.

It was implemented that way to put the logic of getting the last transaction event in another class although it is as simple as getting the last element.
Also, this logic could detect misconfigurations that can only be detected at runtime (second check in `getLastTransactionEvent()`), then the caller of getLastTransactionEvent() could detect it by means of the exception and set all the events as the last in the transaction. Why do this? because if not all events in the transaction go to the same senders (misconfiguration), that would imply that some senders will not get all the events for the transaction but could try if group-transaction-events is set to complete the transaction in the batch, by looking for the last event in the transaction which could never come (delaying the batch sending) or may come but not guaranteeing that the rest of the events for the transaction were sent in the batch (because they may not have reached the sender).

The first check of `getLastTransactionEvent()` was added in order to avoid the cost of the second check if no sender had group-transaction-events set to true. In that case, it is irrelevant to return the last transaction event or null because it will not be used although, it could make more sense to return the last transaction event.

agingade:
> > @pivotal-eshu
> > It looks like the Tx code is trying to assess Gateway sender requirement to support GEODE-7971 in TX code. Which doesn't seems right.
> > The "getLastTransactionEvent" should just return the last event in the transaction; currently it is doing it by "callbacks.get(callbacks.size() - 1);" which also seems to be error prone or inefficient.
> > In "firePendingCallbacks" it could have iterated over the list using size index:
> > for (int i=0; i<callbaack.size(); i++){
> > ee.getRegion().invokeTXCallbacks(EnumListenerEvent.AFTER_DESTROY, ee, true, (i+1) == size() /_isLastTrasaction_/);
> > }
> > Please correct me if i am missing anything.
> 
> It was implemented that way to put the logic of getting the last transaction event in another class although it is as simple as getting the last element.
> Also, this logic could detect misconfigurations that can only be detected at runtime (second check in `getLastTransactionEvent()`), then the caller of getLastTransactionEvent() could detect it by means of the exception and set all the events as the last in the transaction. Why do this? because if not all events in the transaction go to the same senders (misconfiguration), that would imply that some senders will not get all the events for the transaction but could try if group-transaction-events is set to complete the transaction in the batch, by looking for the last event in the transaction which could never come (delaying the batch sending) or may come but not guaranteeing that the rest of the events for the transaction were sent in the batch (because they may not have reached the sender).
> 
> The first check of `getLastTransactionEvent()` was added in order to avoid the cost of the second check if no sender had group-transaction-events set to true. In that case, it is irrelevant to return the last transaction event or null because it will not be used although, it could make more sense to return the last transaction event.

It looks like the GEODE-7971 tried to use grouping with the existing way of "invokeTXCallbacks()", where client, wan, cache-listener notification are done one event from transaction at a time. Instead it could have invoked wan separately from "firePendingCallbacks"; which has all the TX events in one group.
E.g.:
firePendingCallbacks() {
invokeTxCallbackForClientCacheLisener(getPendingCallbacks());
invokeTxCallbackForWan(getPendingCallbacks());
}
Then there was no need to tag last transaction event. 
Please correct me if i am missing anything.

And calling  "configError" for an TX entry not part of specific sender seems to be wrong; why would it be a error; think about the case where TX uses helper/lookup region entry/value to calculate a final result and other site is only interested in the final result, this is not an error, this is the most expected behavior.
E.g.:
getBonusIncreatementFromBonusRegion();
updateNewSalaryForEmployee(); // The region or final value other wan site is interested in.

I think for now we need to change the naming of the methods to be more meaningful (based on what they are doing):
"getLastTransactionEvent" to "getLastTransactionEventForGroupedWANSender()" 
Instead of "ServiceConfigurationError" return null or boolean. As the new change is doing.

geode-core/src/main/java/org/apache/geode/internal/cache/TXLastEventInTransactionUtils.java
```diff
@@ -77,11 +78,7 @@ private static boolean checkNoSendersGroupTransactionEvents(List<EntryEventImpl>
       Cache cache) throws ServiceConfigurationError {
     for (String senderId : getSenderIdsForEvents(callbacks)) {
       GatewaySender sender = cache.getGatewaySender(senderId);
-      if (sender == null) {
-        logger.error("No sender found for {}", senderId);
-        throw new ServiceConfigurationError("No information for senderId: " + senderId);
-      }
-      if (sender.mustGroupTransactionEvents()) {
+      if (sender != null && sender.mustGroupTransactionEvents()) {
```
agingade, line 81: The "ServiceConfigurationError" is thrown in "getLastTransactionEvent" that can also be removed by returning null; as done here...

albertogpz:
> > > @pivotal-eshu
> > > It looks like the Tx code is trying to assess Gateway sender requirement to support GEODE-7971 in TX code. Which doesn't seems right.
> > > The "getLastTransactionEvent" should just return the last event in the transaction; currently it is doing it by "callbacks.get(callbacks.size() - 1);" which also seems to be error prone or inefficient.
> > > In "firePendingCallbacks" it could have iterated over the list using size index:
> > > for (int i=0; i<callbaack.size(); i++){
> > > ee.getRegion().invokeTXCallbacks(EnumListenerEvent.AFTER_DESTROY, ee, true, (i+1) == size() /_isLastTrasaction_/);
> > > }
> > > Please correct me if i am missing anything.
> > 
> > 
> > It was implemented that way to put the logic of getting the last transaction event in another class although it is as simple as getting the last element.
> > Also, this logic could detect misconfigurations that can only be detected at runtime (second check in `getLastTransactionEvent()`), then the caller of getLastTransactionEvent() could detect it by means of the exception and set all the events as the last in the transaction. Why do this? because if not all events in the transaction go to the same senders (misconfiguration), that would imply that some senders will not get all the events for the transaction but could try if group-transaction-events is set to complete the transaction in the batch, by looking for the last event in the transaction which could never come (delaying the batch sending) or may come but not guaranteeing that the rest of the events for the transaction were sent in the batch (because they may not have reached the sender).
> > The first check of `getLastTransactionEvent()` was added in order to avoid the cost of the second check if no sender had group-transaction-events set to true. In that case, it is irrelevant to return the last transaction event or null because it will not be used although, it could make more sense to return the last transaction event.
> 
> It looks like the GEODE-7971 tried to use grouping with the existing way of "invokeTXCallbacks()", where client, wan, cache-listener notification are done one event from transaction at a time. Instead it could have invoked wan separately from "firePendingCallbacks"; which has all the TX events in one group.
> E.g.:
> firePendingCallbacks() {
> invokeTxCallbackForClientCacheLisener(getPendingCallbacks());
> invokeTxCallbackForWan(getPendingCallbacks());
> }
> Then there was no need to tag last transaction event.
> Please correct me if i am missing anything.
> 
> And calling "configError" for an TX entry not part of specific sender seems to be wrong; why would it be a error; think about the case where TX uses helper/lookup region entry/value to calculate a final result and other site is only interested in the final result, this is not an error, this is the most expected behavior.
> E.g.:
> getBonusIncreatementFromBonusRegion();
> updateNewSalaryForEmployee(); // The region or final value other wan site is interested in.
> 
> I think for now we need to change the naming of the methods to be more meaningful (based on what they are doing):
> "getLastTransactionEvent" to "getLastTransactionEventForGroupedWANSender()"
> Instead of "ServiceConfigurationError" return null or boolean. As the new change is doing.

Tagging the last transaction event is needed by `SerialGatewaySenderQueue` and `ParallelGatewaySenderQueue` to make sure that, when group-transaction-events is enabled, batches do not contain incomplete transactions. The way to assure that is to check that for every transaction in the batch (events are tagged with the transaction event), there is an event for that transaction that is the last event in the transaction (there is an event tagged as last transaction event).
I do not see why your changes avoid the need for tagging the last transaction event.

Secondly, having events for a transaction that will not be replicated by all senders (when some of the senders have group-transaction-events set to true) could POSSIBLY be a configuration error.
Let me illustrate it with an example:
Let's suppose that we have a system with two senders, sender1 and sender2, both with group-transaction-events set to true.
Now let's suppose that a transaction (t1) arrives to Geode with two events: event1 (create a) and event2 (create b).
Let's suppose that event1 is to be replicated by sender1 (but not sender2) and event 2 is to be replicated by sender2 (but not sender1).
The last event for the transaction will be event2 and it will be tagged as such.
What we will see happening is that even1 arrives to sender1's queue. When event1 is added to a batch and the batch is to be sent, it will be checked if all events for the transaction to which even1 belongs (t1) are in the batch. As there is only one event and it is not the last event in the the transaction, the batch creator will try to get the missing events for the transaction but it will never get them because they will never arrive, the last event in the transaction will not be on the sender1 queue but it will be on the sender2 queue.
At the same time, event2 will arrive to sender2's queue. When this event is added to a batch and the batch is to be sent, it will be checked if all transactions are complete. For t1, as we have the last event for the transaction (event2), it will be considered that it is so. Nevertheless, as not all events for the transaction are sent to this sender, the assumption would be wrong.

agingade:
> > > > @pivotal-eshu
> > > > It looks like the Tx code is trying to assess Gateway sender requirement to support GEODE-7971 in TX code. Which doesn't seems right.
> > > > The "getLastTransactionEvent" should just return the last event in the transaction; currently it is doing it by "callbacks.get(callbacks.size() - 1);" which also seems to be error prone or inefficient.
> > > > In "firePendingCallbacks" it could have iterated over the list using size index:
> > > > for (int i=0; i<callbaack.size(); i++){
> > > > ee.getRegion().invokeTXCallbacks(EnumListenerEvent.AFTER_DESTROY, ee, true, (i+1) == size() /_isLastTrasaction_/);
> > > > }
> > > > Please correct me if i am missing anything.
> > > 
> > > 
> > > It was implemented that way to put the logic of getting the last transaction event in another class although it is as simple as getting the last element.
> > > Also, this logic could detect misconfigurations that can only be detected at runtime (second check in `getLastTransactionEvent()`), then the caller of getLastTransactionEvent() could detect it by means of the exception and set all the events as the last in the transaction. Why do this? because if not all events in the transaction go to the same senders (misconfiguration), that would imply that some senders will not get all the events for the transaction but could try if group-transaction-events is set to complete the transaction in the batch, by looking for the last event in the transaction which could never come (delaying the batch sending) or may come but not guaranteeing that the rest of the events for the transaction were sent in the batch (because they may not have reached the sender).
> > > The first check of `getLastTransactionEvent()` was added in order to avoid the cost of the second check if no sender had group-transaction-events set to true. In that case, it is irrelevant to return the last transaction event or null because it will not be used although, it could make more sense to return the last transaction event.
> > 
> > 
> > It looks like the GEODE-7971 tried to use grouping with the existing way of "invokeTXCallbacks()", where client, wan, cache-listener notification are done one event from transaction at a time. Instead it could have invoked wan separately from "firePendingCallbacks"; which has all the TX events in one group.
> > E.g.:
> > firePendingCallbacks() {
> > invokeTxCallbackForClientCacheLisener(getPendingCallbacks());
> > invokeTxCallbackForWan(getPendingCallbacks());
> > }
> > Then there was no need to tag last transaction event.
> > Please correct me if i am missing anything.
> > And calling "configError" for an TX entry not part of specific sender seems to be wrong; why would it be a error; think about the case where TX uses helper/lookup region entry/value to calculate a final result and other site is only interested in the final result, this is not an error, this is the most expected behavior.
> > E.g.:
> > getBonusIncreatementFromBonusRegion();
> > updateNewSalaryForEmployee(); // The region or final value other wan site is interested in.
> > I think for now we need to change the naming of the methods to be more meaningful (based on what they are doing):
> > "getLastTransactionEvent" to "getLastTransactionEventForGroupedWANSender()"
> > Instead of "ServiceConfigurationError" return null or boolean. As the new change is doing.
> 

> Tagging the last transaction event is needed by `SerialGatewaySenderQueue` and `ParallelGatewaySenderQueue` to make sure that, when group-transaction-events is enabled, batches do not contain incomplete transactions. The way to assure that is to check that for every transaction in the batch (events are tagged with the transaction event), there is an event for that transaction that is the last event in the transaction (there is an event tagged as last transaction event).
> I do not see why your changes avoid the need for tagging the last transaction event.
> 
Thinking more about this, I see what you are saying...
The other option could have been, separating WAN callback and creating a single event/message for all the entries in transaction. May be this is too much. 

> Secondly, having events for a transaction that will not be replicated by all senders (when some of the senders have group-transaction-events set to true) could POSSIBLY be a configuration error.
> Let me illustrate it with an example:
> Let's suppose that we have a system with two senders, sender1 and sender2, both with group-transaction-events set to true.
> Now let's suppose that a transaction (t1) arrives to Geode with two events: event1 (create a) and event2 (create b).
> Let's suppose that event1 is to be replicated by sender1 (but not sender2) and event 2 is to be replicated by sender2 (but not sender1).
> The last event for the transaction will be event2 and it will be tagged as such.
> What we will see happening is that even1 arrives to sender1's queue. When event1 is added to a batch and the batch is to be sent, it will be checked if all events for the transaction to which even1 belongs (t1) are in the batch. As there is only one event and it is not the last event in the the transaction, the batch creator will try to get the missing events for the transaction but it will never get them because they will never arrive, the last event in the transaction will not be on the sender1 queue but it will be on the sender2 queue.
> At the same time, event2 will arrive to sender2's queue. When this event is added to a batch and the batch is to be sent, it will be checked if all transactions are complete. For t1, as we have the last event for the transaction (event2), it will be considered that it is so. Nevertheless, as not all events for the transaction are sent to this sender, the assumption would be wrong.
I understand why the lastTransactionEvent; but I don't see how its a configuration error...When you say configuration error, are you meaning sender configuration or setting grouping configuration for a sender. In any case, it is different usecase but is not an error. Application have their own requirement of sending/managing data in different sites.

albertogpz:
> > > > > @pivotal-eshu
> > > > > It looks like the Tx code is trying to assess Gateway sender requirement to support GEODE-7971 in TX code. Which doesn't seems right.
> > > > > The "getLastTransactionEvent" should just return the last event in the transaction; currently it is doing it by "callbacks.get(callbacks.size() - 1);" which also seems to be error prone or inefficient.
> > > > > In "firePendingCallbacks" it could have iterated over the list using size index:
> > > > > for (int i=0; i<callbaack.size(); i++){
> > > > > ee.getRegion().invokeTXCallbacks(EnumListenerEvent.AFTER_DESTROY, ee, true, (i+1) == size() /_isLastTrasaction_/);
> > > > > }
> > > > > Please correct me if i am missing anything.
> > > > 
> > > > 
> > > > It was implemented that way to put the logic of getting the last transaction event in another class although it is as simple as getting the last element.
> > > > Also, this logic could detect misconfigurations that can only be detected at runtime (second check in `getLastTransactionEvent()`), then the caller of getLastTransactionEvent() could detect it by means of the exception and set all the events as the last in the transaction. Why do this? because if not all events in the transaction go to the same senders (misconfiguration), that would imply that some senders will not get all the events for the transaction but could try if group-transaction-events is set to complete the transaction in the batch, by looking for the last event in the transaction which could never come (delaying the batch sending) or may come but not guaranteeing that the rest of the events for the transaction were sent in the batch (because they may not have reached the sender).
> > > > The first check of `getLastTransactionEvent()` was added in order to avoid the cost of the second check if no sender had group-transaction-events set to true. In that case, it is irrelevant to return the last transaction event or null because it will not be used although, it could make more sense to return the last transaction event.
> > > 
> > > 
> > > It looks like the GEODE-7971 tried to use grouping with the existing way of "invokeTXCallbacks()", where client, wan, cache-listener notification are done one event from transaction at a time. Instead it could have invoked wan separately from "firePendingCallbacks"; which has all the TX events in one group.
> > > E.g.:
> > > firePendingCallbacks() {
> > > invokeTxCallbackForClientCacheLisener(getPendingCallbacks());
> > > invokeTxCallbackForWan(getPendingCallbacks());
> > > }
> > > Then there was no need to tag last transaction event.
> > > Please correct me if i am missing anything.
> > > And calling "configError" for an TX entry not part of specific sender seems to be wrong; why would it be a error; think about the case where TX uses helper/lookup region entry/value to calculate a final result and other site is only interested in the final result, this is not an error, this is the most expected behavior.
> > > E.g.:
> > > getBonusIncreatementFromBonusRegion();
> > > updateNewSalaryForEmployee(); // The region or final value other wan site is interested in.
> > > I think for now we need to change the naming of the methods to be more meaningful (based on what they are doing):
> > > "getLastTransactionEvent" to "getLastTransactionEventForGroupedWANSender()"
> > > Instead of "ServiceConfigurationError" return null or boolean. As the new change is doing.
> 
> > Tagging the last transaction event is needed by `SerialGatewaySenderQueue` and `ParallelGatewaySenderQueue` to make sure that, when group-transaction-events is enabled, batches do not contain incomplete transactions. The way to assure that is to check that for every transaction in the batch (events are tagged with the transaction event), there is an event for that transaction that is the last event in the transaction (there is an event tagged as last transaction event).
> > I do not see why your changes avoid the need for tagging the last transaction event.
> 
> Thinking more about this, I see what you are saying...
> The other option could have been, separating WAN callback and creating a single event/message for all the entries in transaction. May be this is too much.
I thought about that alternative when I was designing this solution but abandoned it as it implied bigger changes.
> 
> > Secondly, having events for a transaction that will not be replicated by all senders (when some of the senders have group-transaction-events set to true) could POSSIBLY be a configuration error.
> > Let me illustrate it with an example:
> > Let's suppose that we have a system with two senders, sender1 and sender2, both with group-transaction-events set to true.
> > Now let's suppose that a transaction (t1) arrives to Geode with two events: event1 (create a) and event2 (create b).
> > Let's suppose that event1 is to be replicated by sender1 (but not sender2) and event 2 is to be replicated by sender2 (but not sender1).
> > The last event for the transaction will be event2 and it will be tagged as such.
> > What we will see happening is that even1 arrives to sender1's queue. When event1 is added to a batch and the batch is to be sent, it will be checked if all events for the transaction to which even1 belongs (t1) are in the batch. As there is only one event and it is not the last event in the the transaction, the batch creator will try to get the missing events for the transaction but it will never get them because they will never arrive, the last event in the transaction will not be on the sender1 queue but it will be on the sender2 queue.
> > At the same time, event2 will arrive to sender2's queue. When this event is added to a batch and the batch is to be sent, it will be checked if all transactions are complete. For t1, as we have the last event for the transaction (event2), it will be considered that it is so. Nevertheless, as not all events for the transaction are sent to this sender, the assumption would be wrong.
> 
> I understand why the lastTransactionEvent; but I don't see how its a configuration error...When you say configuration error, are you meaning sender configuration or setting grouping configuration for a sender. In any case, it is different usecase but is not an error. Application have their own requirement of sending/managing data in different sites.

I mean both. This is what the documentation says about this feature:

_In order to use transaction event grouping:
- The `group-transaction-events` setting is supported only on serial senders with just one dispatcher thread, or on parallel senders.
- The regions to which the transaction events belong must be replicated by the same set of gateway senders that also have this setting enabled.
- This setting cannot be enabled if `enable-batch-conflation` is in effect._

The first and the third conditions can be (and are) checked at configuration time.
But the second condition can only be checked at run time, when a transaction is received. If you have a transaction with several events that spans several regions, and the gateway senders for the regions are not the same, then group-transaction-events will not work even though whoever configured the gateway senders may think it will. Some senders will receive some events for the transaction and other senders will receive other events.
One may argue that this is not necessarily a configuration error but the fact is that, in this case, a transaction replicated by gateway senders who have group-transaction-events set to true will not be sent in the same batch because senders will not receive all the events for the transaction.

geode-core/src/main/java/org/apache/geode/internal/cache/TXLastEventInTransactionUtils.java
pivotal-eshu: In the flow of the code to getLastTransactionEvent, checkNoSendersGroupTransactionEvents was executed before doesSenderGroupTransactionEvents.

Method checkNoSendersGroupTransactionEvents returns false if there is one op needs to send to a GatewaySender which has mustGroupTransactionEvents set to true. Otherwise doesSenderGroupTransactionEvents method won't be executed (getLastTransactionEvent returns null in that case).

If in a tx, there is one op needs to send to gateway with mustGroupTransactionEvents setting, all ops in the transaction should as well. If not, it means user did not configure the senders correctly (for the transaction executed).

agingade:
> > > > > > @pivotal-eshu
> > > > > > It looks like the Tx code is trying to assess Gateway sender requirement to support GEODE-7971 in TX code. Which doesn't seems right.
> > > > > > The "getLastTransactionEvent" should just return the last event in the transaction; currently it is doing it by "callbacks.get(callbacks.size() - 1);" which also seems to be error prone or inefficient.
> > > > > > In "firePendingCallbacks" it could have iterated over the list using size index:
> > > > > > for (int i=0; i<callbaack.size(); i++){
> > > > > > ee.getRegion().invokeTXCallbacks(EnumListenerEvent.AFTER_DESTROY, ee, true, (i+1) == size() /_isLastTrasaction_/);
> > > > > > }
> > > > > > Please correct me if i am missing anything.
> > > > > 
> > > > > 
> > > > > It was implemented that way to put the logic of getting the last transaction event in another class although it is as simple as getting the last element.
> > > > > Also, this logic could detect misconfigurations that can only be detected at runtime (second check in `getLastTransactionEvent()`), then the caller of getLastTransactionEvent() could detect it by means of the exception and set all the events as the last in the transaction. Why do this? because if not all events in the transaction go to the same senders (misconfiguration), that would imply that some senders will not get all the events for the transaction but could try if group-transaction-events is set to complete the transaction in the batch, by looking for the last event in the transaction which could never come (delaying the batch sending) or may come but not guaranteeing that the rest of the events for the transaction were sent in the batch (because they may not have reached the sender).
> > > > > The first check of `getLastTransactionEvent()` was added in order to avoid the cost of the second check if no sender had group-transaction-events set to true. In that case, it is irrelevant to return the last transaction event or null because it will not be used although, it could make more sense to return the last transaction event.
> > > > 
> > > > 
> > > > It looks like the GEODE-7971 tried to use grouping with the existing way of "invokeTXCallbacks()", where client, wan, cache-listener notification are done one event from transaction at a time. Instead it could have invoked wan separately from "firePendingCallbacks"; which has all the TX events in one group.
> > > > E.g.:
> > > > firePendingCallbacks() {
> > > > invokeTxCallbackForClientCacheLisener(getPendingCallbacks());
> > > > invokeTxCallbackForWan(getPendingCallbacks());
> > > > }
> > > > Then there was no need to tag last transaction event.
> > > > Please correct me if i am missing anything.
> > > > And calling "configError" for an TX entry not part of specific sender seems to be wrong; why would it be a error; think about the case where TX uses helper/lookup region entry/value to calculate a final result and other site is only interested in the final result, this is not an error, this is the most expected behavior.
> > > > E.g.:
> > > > getBonusIncreatementFromBonusRegion();
> > > > updateNewSalaryForEmployee(); // The region or final value other wan site is interested in.
> > > > I think for now we need to change the naming of the methods to be more meaningful (based on what they are doing):
> > > > "getLastTransactionEvent" to "getLastTransactionEventForGroupedWANSender()"
> > > > Instead of "ServiceConfigurationError" return null or boolean. As the new change is doing.
> > 
> > 
> > > Tagging the last transaction event is needed by `SerialGatewaySenderQueue` and `ParallelGatewaySenderQueue` to make sure that, when group-transaction-events is enabled, batches do not contain incomplete transactions. The way to assure that is to check that for every transaction in the batch (events are tagged with the transaction event), there is an event for that transaction that is the last event in the transaction (there is an event tagged as last transaction event).
> > > I do not see why your changes avoid the need for tagging the last transaction event.
> > 
> > 
> > Thinking more about this, I see what you are saying...
> > The other option could have been, separating WAN callback and creating a single event/message for all the entries in transaction. May be this is too much.
> > I thought about that alternative when I was designing this solution but abandoned it as it implied bigger changes.
> > > Secondly, having events for a transaction that will not be replicated by all senders (when some of the senders have group-transaction-events set to true) could POSSIBLY be a configuration error.
> > > Let me illustrate it with an example:
> > > Let's suppose that we have a system with two senders, sender1 and sender2, both with group-transaction-events set to true.
> > > Now let's suppose that a transaction (t1) arrives to Geode with two events: event1 (create a) and event2 (create b).
> > > Let's suppose that event1 is to be replicated by sender1 (but not sender2) and event 2 is to be replicated by sender2 (but not sender1).
> > > The last event for the transaction will be event2 and it will be tagged as such.
> > > What we will see happening is that even1 arrives to sender1's queue. When event1 is added to a batch and the batch is to be sent, it will be checked if all events for the transaction to which even1 belongs (t1) are in the batch. As there is only one event and it is not the last event in the the transaction, the batch creator will try to get the missing events for the transaction but it will never get them because they will never arrive, the last event in the transaction will not be on the sender1 queue but it will be on the sender2 queue.
> > > At the same time, event2 will arrive to sender2's queue. When this event is added to a batch and the batch is to be sent, it will be checked if all transactions are complete. For t1, as we have the last event for the transaction (event2), it will be considered that it is so. Nevertheless, as not all events for the transaction are sent to this sender, the assumption would be wrong.
> > 
> > 
> > I understand why the lastTransactionEvent; but I don't see how its a configuration error...When you say configuration error, are you meaning sender configuration or setting grouping configuration for a sender. In any case, it is different usecase but is not an error. Application have their own requirement of sending/managing data in different sites.
> 
> I mean both. This is what the documentation says about this feature:
> 
> _In order to use transaction event grouping:
> 
> * The `group-transaction-events` setting is supported only on serial senders with just one dispatcher thread, or on parallel senders.
> * The regions to which the transaction events belong must be replicated by the same set of gateway senders that also have this setting enabled.
> * This setting cannot be enabled if `enable-batch-conflation` is in effect._
> 
> The first and the third conditions can be (and are) checked at configuration time.
> But the second condition can only be checked at run time, when a transaction is received. If you have a transaction with several events that spans several regions, and the gateway senders for the regions are not the same, then group-transaction-events will not work even though whoever configured the gateway senders may think it will. Some senders will receive some events for the transaction and other senders will receive other events.

Yes...Thats up to application/use-case, how and what kind data it wants to be sent to other site. The WAN gateway senders 
doesn't have any limitations like that...

> One may argue that this is not necessarily a configuration error but the fact is that, in this case, a transaction replicated by gateway senders who have group-transaction-events set to true will not be sent in the same batch because senders will not receive all the events for the transaction.

The grouping of transaction is some-way of making sure the transaction can be sent as a single group; but it should not be limiting factor (saying error). If it is a limiting factor, the TX should have been rolled back and an exception is thrown to application. Right now what I can see is nothing is rolled back or TX is aborted because of error, it is only logged as message; which is noticed after the event/error has already happened.

geode-core/src/main/java/org/apache/geode/internal/cache/TXState.java
```diff
@@ -253,7 +253,8 @@ public void firePendingCallbacks() {
     }
 
     for (EntryEventImpl ee : getPendingCallbacks()) {
-      boolean isLastTransactionEvent = isConfigError || ee.equals(lastTransactionEvent);
+      boolean isLastTransactionEvent =
+          isConfigError || lastTransactionEvent == null || ee.equals(lastTransactionEvent);
```
kirklund, line 257: This is the 2nd instance of this same code which was also in TXCommitMessage.java. You might want to create a static utils method and have both of these call that method.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/WANTestBase.java
```diff
@@ -2302,6 +2302,25 @@ public static void doPuts(String regionName, int numPuts) {
...
+        cache.getCacheTransactionManager().begin();
+        r.put(i, "Value_" + i);
+        cache.getCacheTransactionManager().commit();
+      }
+    } finally {
+      exp1.remove();
+      exp2.remove();
+    }
+  }
+
```
kirklund, line 2323: Another syntax which is a bit cleaner:
```

try (IgnoredException ie1 = addIgnoredException(InterruptedException.class);
       IgnoredException ie2 = addIgnoredException(GatewaySenderException.class)) {
  // ...
}
```
No need for a `finally` block when you do it as a try-with-resources like this.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/serial/SerialWANPropagation_PartitionedRegionDUnitTest.java
```diff
@@ -78,6 +78,41 @@ public void testPartitionedSerialPropagation() {
     vm2.invoke(() -> WANTestBase.validateRegionSize(getTestMethodName() + "_PR", 1000));
   }
 
+  @Test
+  public void testPartitionedSerialPropagationWithTXWhenSendersNotConfiguredOnAllServers() {
+
+    Integer lnPort = vm0.invoke(() -> WANTestBase.createFirstLocatorWithDSId(1));
+    Integer nyPort = vm1.invoke(() -> WANTestBase.createFirstRemoteLocator(2, lnPort));
```
kirklund, line 85: You can just declare things like `lnPort` as `int` primitive instead of `Integer`.

kirklund (approved):
Approved with some comments and recommendations. No changes actually _needed_ though.

albertogpz:
> This is what the



> > > > > > > @pivotal-eshu
> > > > > > > It looks like the Tx code is trying to assess Gateway sender requirement to support GEODE-7971 in TX code. Which doesn't seems right.
> > > > > > > The "getLastTransactionEvent" should just return the last event in the transaction; currently it is doing it by "callbacks.get(callbacks.size() - 1);" which also seems to be error prone or inefficient.
> > > > > > > In "firePendingCallbacks" it could have iterated over the list using size index:
> > > > > > > for (int i=0; i<callbaack.size(); i++){
> > > > > > > ee.getRegion().invokeTXCallbacks(EnumListenerEvent.AFTER_DESTROY, ee, true, (i+1) == size() /_isLastTrasaction_/);
> > > > > > > }
> > > > > > > Please correct me if i am missing anything.
> > > > > > 
> > > > > > 
> > > > > > It was implemented that way to put the logic of getting the last transaction event in another class although it is as simple as getting the last element.
> > > > > > Also, this logic could detect misconfigurations that can only be detected at runtime (second check in `getLastTransactionEvent()`), then the caller of getLastTransactionEvent() could detect it by means of the exception and set all the events as the last in the transaction. Why do this? because if not all events in the transaction go to the same senders (misconfiguration), that would imply that some senders will not get all the events for the transaction but could try if group-transaction-events is set to complete the transaction in the batch, by looking for the last event in the transaction which could never come (delaying the batch sending) or may come but not guaranteeing that the rest of the events for the transaction were sent in the batch (because they may not have reached the sender).
> > > > > > The first check of `getLastTransactionEvent()` was added in order to avoid the cost of the second check if no sender had group-transaction-events set to true. In that case, it is irrelevant to return the last transaction event or null because it will not be used although, it could make more sense to return the last transaction event.
> > > > > 
> > > > > 
> > > > > It looks like the GEODE-7971 tried to use grouping with the existing way of "invokeTXCallbacks()", where client, wan, cache-listener notification are done one event from transaction at a time. Instead it could have invoked wan separately from "firePendingCallbacks"; which has all the TX events in one group.
> > > > > E.g.:
> > > > > firePendingCallbacks() {
> > > > > invokeTxCallbackForClientCacheLisener(getPendingCallbacks());
> > > > > invokeTxCallbackForWan(getPendingCallbacks());
> > > > > }
> > > > > Then there was no need to tag last transaction event.
> > > > > Please correct me if i am missing anything.
> > > > > And calling "configError" for an TX entry not part of specific sender seems to be wrong; why would it be a error; think about the case where TX uses helper/lookup region entry/value to calculate a final result and other site is only interested in the final result, this is not an error, this is the most expected behavior.
> > > > > E.g.:
> > > > > getBonusIncreatementFromBonusRegion();
> > > > > updateNewSalaryForEmployee(); // The region or final value other wan site is interested in.
> > > > > I think for now we need to change the naming of the methods to be more meaningful (based on what they are doing):
> > > > > "getLastTransactionEvent" to "getLastTransactionEventForGroupedWANSender()"
> > > > > Instead of "ServiceConfigurationError" return null or boolean. As the new change is doing.
> > > 
> > > 
> > > > Tagging the last transaction event is needed by `SerialGatewaySenderQueue` and `ParallelGatewaySenderQueue` to make sure that, when group-transaction-events is enabled, batches do not contain incomplete transactions. The way to assure that is to check that for every transaction in the batch (events are tagged with the transaction event), there is an event for that transaction that is the last event in the transaction (there is an event tagged as last transaction event).
> > > > I do not see why your changes avoid the need for tagging the last transaction event.
> > > 
> > > 
> > > Thinking more about this, I see what you are saying...
> > > The other option could have been, separating WAN callback and creating a single event/message for all the entries in transaction. May be this is too much.
> > > I thought about that alternative when I was designing this solution but abandoned it as it implied bigger changes.
> > > > Secondly, having events for a transaction that will not be replicated by all senders (when some of the senders have group-transaction-events set to true) could POSSIBLY be a configuration error.
> > > > Let me illustrate it with an example:
> > > > Let's suppose that we have a system with two senders, sender1 and sender2, both with group-transaction-events set to true.
> > > > Now let's suppose that a transaction (t1) arrives to Geode with two events: event1 (create a) and event2 (create b).
> > > > Let's suppose that event1 is to be replicated by sender1 (but not sender2) and event 2 is to be replicated by sender2 (but not sender1).
> > > > The last event for the transaction will be event2 and it will be tagged as such.
> > > > What we will see happening is that even1 arrives to sender1's queue. When event1 is added to a batch and the batch is to be sent, it will be checked if all events for the transaction to which even1 belongs (t1) are in the batch. As there is only one event and it is not the last event in the the transaction, the batch creator will try to get the missing events for the transaction but it will never get them because they will never arrive, the last event in the transaction will not be on the sender1 queue but it will be on the sender2 queue.
> > > > At the same time, event2 will arrive to sender2's queue. When this event is added to a batch and the batch is to be sent, it will be checked if all transactions are complete. For t1, as we have the last event for the transaction (event2), it will be considered that it is so. Nevertheless, as not all events for the transaction are sent to this sender, the assumption would be wrong.
> > > 
> > > 
> > > I understand why the lastTransactionEvent; but I don't see how its a configuration error...When you say configuration error, are you meaning sender configuration or setting grouping configuration for a sender. In any case, it is different usecase but is not an error. Application have their own requirement of sending/managing data in different sites.
> > 
> > 
> > I mean both. This is what the documentation says about this feature:
> > _In order to use transaction event grouping:
> > 
> > * The `group-transaction-events` setting is supported only on serial senders with just one dispatcher thread, or on parallel senders.
> > * The regions to which the transaction events belong must be replicated by the same set of gateway senders that also have this setting enabled.
> > * This setting cannot be enabled if `enable-batch-conflation` is in effect._
> > 
> > The first and the third conditions can be (and are) checked at configuration time.
> > But the second condition can only be checked at run time, when a transaction is received. If you have a transaction with several events that spans several regions, and the gateway senders for the regions are not the same, then group-transaction-events will not work even though whoever configured the gateway senders may think it will. Some senders will receive some events for the transaction and other senders will receive other events.
> 
> Yes...Thats up to application/use-case, how and what kind data it wants to be sent to other site. The WAN gateway senders
> doesn't have any limitations like that...
> 
> > One may argue that this is not necessarily a configuration error but the fact is that, in this case, a transaction replicated by gateway senders who have group-transaction-events set to true will not be sent in the same batch because senders will not receive all the events for the transaction.
> 
> The grouping of transaction is some-way of making sure the transaction can be sent as a single group; but it should not be limiting factor (saying error). If it is a limiting factor, the TX should have been rolled back and an exception is thrown to application. Right now what I can see is nothing is rolled back or TX is aborted because of error, it is only logged as message; which is noticed after the event/error has already happened.
You are right. It is not a limiting factor. But if you have configured group-transaction-events for some or all your senders and you have events in transactions being sent by those senders that cannot be grouped then chances are that you might have made a mistake during the configuration. As this cannot be detected at configuration time and it is not necessarily a configuration error, you do not want to rollback the transaction. But at least, you will have a log telling you that transactions are not being grouped so that you can check if there is a configuration error.


> 
> @albertogpz you also mentioned "This is what the documentation says about this feature:"; I believe this is not release feature and is not yet documented in Geode manuals; if this needs to be documented we need to create a jira ticket and tag docs team @davebarnes97
This is documented in: geode-docs/topologies_and_comm/topology_concepts/multisite_overview.html.md.erb
> 
> Also, reading over the code, I see many of the methods in TXLastEventInTransactionUtils.class doesn't have unit level tests. We were wondering what happens with the helper methods in this class, when there are no senders configured, when there are multiple enders on each region with and without grouping option. It will really help Geode product having all the unit test cases covered; do you think you can add additional tests...
Tests could also be added to the non-public methods of `TXLastEventInTransactionUtils` in `TXLastEventInTransactionUtilsTest` but the current tests on the public 'getLastTransactionEvent()` method seemed enough to me when I wrote them to cover the cases you mention.

gesterzhou (approved):

boglesby (approved):

PR #6663, opened 2021-06-30
GEODE-9408: Avoid duplicate events sent by Serial Gateway Sender when…

albertogpz:
… group-transaction-events is true

The logic for removing events from the Serial Gateway Sender queue
when an ack for a batch is received from the gateway receiver
is implemented by the remove() method of the SerialGatewaySenderQueue class.
When the group-transaction-events setting is activated, this logic,
needs to act differently if the event was an event peeked to complete
a transaction in a batch than if it is an event peeked with the peekAhead method.
In the first case, the head key and lastDispatched member must not be
updated with the key of the event because that could provoke that
events from the queue not yet retrieved are never retrieved.

Nevertheless, the head key and lastDispatched member must be
updated for a key peeked to complete a transaction (extraPeekedId)
when the key with the previous id is removed.
This was not being done in some cases by the remove() method and it
has been corrected in this PR.

Several DUnit test cases have been added to test this fix
in a Geode WAN system with a Serial GatewaySender with
group-transaction-events set to true. In the test cases, it is verified that
no duplicated events are sent by means of checking that incomplete transactions
are not sent.

### For all changes:

### Note:


geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/parallel/ParallelWANStatsDUnitTest.java
```diff
@@ -390,6 +391,139 @@ public void testPRParallelPropagationWithoutGroupTransactionEventsSendsBatchesWi
...
+              eventsPerTransaction));
+      asyncInvocations.add(asyncInvocation);
+    }
+
+    try {
+      for (AsyncInvocation asyncInvocation : asyncInvocations) {
+        asyncInvocation.await();
+      }
+    } catch (InterruptedException e) {
+      e.printStackTrace();
```
nabarunnag, line 469: I think it will be better to fail the test here than printing the stacktrace

geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
```diff
@@ -1404,13 +1404,19 @@ void peekEventsFromIncompleteTransactions(List<GatewaySenderEventImpl> batch,
...
+      try {
+        Thread.sleep(GET_TRANSACTION_EVENTS_FROM_QUEUE_WAIT_TIME_MS);
+      } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
+      }
     }
     if (incompleteTransactionIdsInBatch.size() > 0) {
-      logger.warn("Not able to retrieve all events for transactions: {} after {} retries",
-          incompleteTransactionIdsInBatch, retries);
+      logger.warn("Not able to retrieve all events for transactions: {} after {} retries of {}ms" +
```
nabarunnag, line 1418: 3 parameters needed but 2 provided. Maybe the + was a typo

geode-core/src/main/java/org/apache/geode/internal/cache/wan/serial/SerialGatewaySenderEventProcessor.java
```diff
@@ -863,7 +863,7 @@ public void sendBatchDestroyOperationForDroppedEvent(EntryEventImpl dropEvent, i
     } catch (Exception e) {
       if (logger.isDebugEnabled()) {
         logger.debug(
-            "Exception in sending dropped event could be ignored in order not to interrupt sender starting",
+            "Exception in sending dropped event could be ignored in order not to interrupt sender starting: {}",
```
nabarunnag, line 866: I think e.toString needs to be added otherwise it translates it to missing parameter as it translates the method to debug(String message, Throwable ex)

nabarunnag (changes requested):
Added a few comments and additionally the asyncInvocations in the tests can be parameterized using AsyncInvocation<?> or AsyncInvocation<Void> depending on the cases

geode-core/src/main/java/org/apache/geode/cache/wan/GatewaySender.java
```diff
@@ -174,7 +174,15 @@
    */
   int GET_TRANSACTION_EVENTS_FROM_QUEUE_RETRIES =
       Integer.getInteger(GeodeGlossary.GEMFIRE_PREFIX + "get-transaction-events-from-queue-retries",
-          10);
+          3);
+  /**
+   * Milliseconds to wait before retrying to get events for a transaction from the
+   * gateway sender queue when group-transaction-events is true.
+   */
+  int GET_TRANSACTION_EVENTS_FROM_QUEUE_WAIT_TIME_MS =
```
agingade, line 182: Please use SystemPropertyHelper for setting system properties....
E.g.:
SystemPropertyHelper
        .getProductIntegerProperty(SystemPropertyHelper.EVICTION_SCAN_THRESHOLD_PERCENT);

Look for use of VICTION_SCAN_THRESHOLD_PERCENT.

geode-core/src/main/java/org/apache/geode/cache/wan/GatewaySender.java
```diff
@@ -174,7 +174,15 @@
...
+      Integer.getInteger(
+          GeodeGlossary.GEMFIRE_PREFIX + "get-transaction-events-from-queue-wait-time-ms",
```
agingade, line 184: Please use SystemPropertyHelper for setting system properties....
E.g.:
SystemPropertyHelper
        .getProductIntegerProperty(SystemPropertyHelper.EVICTION_SCAN_THRESHOLD_PERCENT);

Look for use of VICTION_SCAN_THRESHOLD_PERCENT.

geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
```diff
@@ -1404,13 +1404,19 @@ void peekEventsFromIncompleteTransactions(List<GatewaySenderEventImpl> batch,
         }
       }
       if (incompleteTransactionIdsInBatch.size() == 0 ||
-          retries++ == GET_TRANSACTION_EVENTS_FROM_QUEUE_RETRIES) {
+          retries >= sender.getRetriesToGetTransactionEventsFromQueue()) {
         break;
       }
+      retries++;
```
agingade, line 1410: I am assuming this is a best effort to find missing transaction events that are getting added to the Queue. If thats the case, can it be done based on wait/notify on Queue? Right now the default wait is for 1ms; but in case if someone sets it longer than wait/notify will be efficient.
Also why is the default 1ms; isn't it too small...

geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
```diff
@@ -1404,13 +1404,19 @@ void peekEventsFromIncompleteTransactions(List<GatewaySenderEventImpl> batch,
...
+        Thread.sleep(GET_TRANSACTION_EVENTS_FROM_QUEUE_WAIT_TIME_MS);
+      } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
+      }
     }
     if (incompleteTransactionIdsInBatch.size() > 0) {
-      logger.warn("Not able to retrieve all events for transactions: {} after {} retries",
-          incompleteTransactionIdsInBatch, retries);
+      logger.warn("Not able to retrieve all events for transactions: {} after {} retries of {}ms" +
+          incompleteTransactionIdsInBatch, retries, GET_TRANSACTION_EVENTS_FROM_QUEUE_WAIT_TIME_MS);
```
agingade, line 1419: Curious, in the code other than logging nothing is done to handle in-complete-trasactions; is it OK to send the events with incomplete transactions...

geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
albertogpz: Good catch. Thanks!

geode-core/src/main/java/org/apache/geode/internal/cache/wan/serial/SerialGatewaySenderEventProcessor.java
albertogpz: I will leave it as it was as I did not know that the method could take a Throwable as a second parameter.

geode-core/src/main/java/org/apache/geode/cache/wan/GatewaySender.java
```diff
@@ -500,4 +516,14 @@
    */
   void setGatewayEventFilters(List<GatewayEventFilter> filters);
 
+  /**
+   * Set the number of retries to get transaction events
+   * for this GatewaySender when GroupTransactionEvents
+   * is set.
+   *
+   * @since Geode 1.15
```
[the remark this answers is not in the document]
albertogpz: If it is not used anymore, I'd rather remove it.

geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
albertogpz: Yes, as you say, this is a best effort.
Given that setting a high value for the wait could lead to delaying too much the replication, which could eventually provoke that queues grow indefinitely, I opted for setting a low value.
Also, in case of a misconfiguration (one not detected at configuration time) due to having transactions with events going to different gateway senders, it could be that it is impossible for the batch to contain the complete transaction. In that case, the maximum time will be consumed (retries*wait_time).

Regarding using wait/notify, I have not thought about it that much. I guess we could call notify when an element is added to the queue and call wait instead of sleep. But, if it makes sense to have a small sleep, then using wait and notify may be inefficient. Do you think it is worth to change it to wait/notify?

geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
albertogpz: Yes, it is as you say. This is a best effort action. If in a "reasonable" time the transaction cannot be completed (and one reason could be because it is impossible due to the contents of the transaction and the configuration of gateway senders), then you go on with sending the batch but you log a warning and increment a counter.

geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
agingade: Wait and notify is preferred over "sleep"; as sleep is not accurate/efficient way of waiting for an event. You can see if its possible to do...

albertogpz: @agingade I have given a bit of thought about using wait and notify instead of sleep but I have come to the conclusion that it is not a good idea in this case.
According to your suggestion, instead of `sleep(time)` we could call `wait(timeout)` and then `notify()` whenever an event is added to the queue. This way, every event added to the queue will awake the thread that must look for events in the queue to complete transactions inside a batch.
The problem is that the event added to the queue that will provoke the awakening of the waiting thread may not be the one required to complete the transaction so the thread could have been awaken in vain.
If a lot of events are entering in the queue at a high rate, it might be that, a lot of awakes will be required before the events to complete the transaction enter the queue. This would require to increase the value of the retries. But this is dangerous as the WAN replication rate could greatly decrease.

The original implementation of this feature did not contemplate any sleeps. It just retried several times the search in the queue to retrieve events to complete transactions when they were not found, assuming that the events for a transaction should arrive to the queue very close in time. I observed that sometimes the retries were done so quickly that even if the number of retries was high, still some batches had to be sent with incomplete transactions and that is why I changed the behavior in this PR to wait a little before retrying by means of sleep.

If you do not agree with my conclusions, I am open to discuss them with you by e-mail, chat or videoconference.

agingade:
@albertogpz wanted to let you know this week most of the geode team members are available scantly; please expect delay in responses.

geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
gesterzhou: I feel it's good to sleep 1ms, but retry number might be better to keep 10 instead of 3.

geode-core/src/main/java/org/apache/geode/cache/wan/GatewaySender.java
```diff
@@ -500,4 +516,12 @@
...
 
+  /**
+   * Set the number of retries to get transaction events
+   * for this GatewaySender when GroupTransactionEvents
+   * is set.
+   *
+   */
+  void setRetriesToGetTransactionEventsFromQueue(int retries);
+
 }
```
boglesby, line 527: This method currently is not invoked by anything.

Other setters in this class are invoked by AlterGatewaySenderFunction. Should this property be alterable?

Also, I don't see any tests that test this property.

geode-core/src/main/java/org/apache/geode/internal/cache/wan/parallel/ParallelGatewaySenderQueue.java
boglesby: GET_TRANSACTION_EVENTS_FROM_QUEUE_WAIT_TIME_MS is 1ms by default.

GET_TRANSACTION_EVENTS_FROM_QUEUE_RETRIES is 3 by default.

So, thats 3 ms of retries by default. As Gester said, I'm not sure thats going to be enough.

geode-core/src/main/java/org/apache/geode/internal/cache/wan/serial/SerialGatewaySenderQueue.java
```diff
@@ -496,13 +489,19 @@ void peekEventsFromIncompleteTransactions(List<AsyncEvent<?, ?>> batch, long las
...
-          retries++ == GET_TRANSACTION_EVENTS_FROM_QUEUE_RETRIES) {
+          retries >= sender.getRetriesToGetTransactionEventsFromQueue()) {
         break;
       }
+      retries++;
+      try {
+        Thread.sleep(GET_TRANSACTION_EVENTS_FROM_QUEUE_WAIT_TIME_MS);
+      } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
+      }
```
boglesby, line 500: Same comment here.

GET_TRANSACTION_EVENTS_FROM_QUEUE_WAIT_TIME_MS is 1ms by default.

GET_TRANSACTION_EVENTS_FROM_QUEUE_RETRIES is 3 by default.

So, thats 3 ms of retries by default. I'm not sure thats going to be enough.

agingade:
@albertogpz 
Thanks to @boglesby, we had extensive internal code review in this area and the changes looks to be fine. There are few comments barry has added that needs to be taken care (or responded); please let us know if you are waiting for anything else on this.

agingade (approved):
Assuming the comments from barry will be addressed.

geode-core/src/main/java/org/apache/geode/cache/wan/GatewaySender.java
albertogpz: Actually, that method is invoked from `WANTestBase::createSender()` when `groupTransactionEvents` is `true`.
I added it to be able to use a high value for the parameter to avoid flakyness in test cases that have `groupTransactionEvents` set to true. Setting this parameter for test cases using a System Property would not be very convenient.
For uses other than test cases, I think that having the System Property would be enough.

geode-core/src/main/java/org/apache/geode/internal/cache/wan/serial/SerialGatewaySenderQueue.java
albertogpz: ok. I will set it to 10 as suggested by @gesterzhou

boglesby (approved):

PR #6665, opened 2021-06-30
GEODE-9407: Handle exception getting region names

aaronlindsey:
GetMemberInformationFunction may throw RegionDestroyedException if a
hosted region is destroyed while the function is checking the region
names. This exception is logged as an "error" level message which is
confusing to users.

Catch the exception and omit the destroyed region from the list of
hosted region names.

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/management/internal/util/ManagementUtils.java
+++ b/geode-core/src/main/java/org/apache/geode/management/internal/util/ManagementUtils.java
@@ -34,11 +34,11 @@ import java.util.Set;
 import java.util.stream.Collectors;
 
 import org.apache.commons.lang3.ArrayUtils;
+import org.apache.logging.log4j.Logger;
 
 import org.apache.geode.annotations.Immutable;
 import org.apache.geode.cache.Cache;
 import org.apache.geode.cache.Region;
-import org.apache.geode.cache.RegionDestroyedException;
 import org.apache.geode.cache.execute.Execution;
 import org.apache.geode.cache.execute.Function;
 import org.apache.geode.cache.execute.FunctionService;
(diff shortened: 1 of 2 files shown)


aaronlindsey:
The failing D-unit test is a known issue: https://issues.apache.org/jira/browse/GEODE-9070.

geode-core/src/main/java/org/apache/geode/management/internal/util/ManagementUtils.java
```diff
@@ -162,14 +163,20 @@ public static DistributedMember getDistributedMemberByNameOrId(String memberName
...
+        Set<Region<?, ?>> subRegions = rootRegion.subregions(true);
 
-      Set<Region<?, ?>> subRegions = rootRegion.subregions(true);
+        for (Region<?, ?> subRegion : subRegions) {
+          regionNames.add(subRegion.getFullPath().substring(1));
+        }
 
-      for (Region<?, ?> subRegion : subRegions) {
-        regionNames.add(subRegion.getFullPath().substring(1));
+      } catch (RegionDestroyedException ignored) {
```
jinmeiliao, line 173: Are we sure this is the only exception this block of code can throw? If we want to prevent this block of code errors out, we should catch a general `Exception`, logs the exception, and then continue.

aaronlindsey: It's not the only exception this block can throw. I like your suggestion of catching general `Exception`. Would it be OK to log the caught exception as "debug" level? (My main purpose with this change is to prevent RegionDestroyedException from showing up in the logs under normal conditions when it is thrown from this function.)

jinmeiliao: would "info" level be ok? not a big stack trace, just enough information about why some region didn't show up in the list.

aaronlindsey: We are performing a health check using the REST API /members endpoint while some regions are being created and destroyed and RegionDestroyedException gets logged quite often. We would like to suppress this exception because it's expected for the call to subregions() to throw this exception after a region is destroyed and it's confusing to users to see this exception in the logs. I am hesitant to log it as an info level exception because it will still show up in the logs under normal circumstances and will still be confusing to users.

kirklund (approved):

geode-core/src/main/java/org/apache/geode/management/internal/util/ManagementUtils.java
aaronlindsey: @jinmeiliao I changed this block of code to catch `Exception` instead of `RegionDestroyedException`, log a debug-level message, and then continue.

jinmeiliao (approved):

PR #6669, opened 2021-07-01
GEODE-9410: User Guide - resolve inconsistencies in gfsh import command description

davebarnes97:
Two write-ups on the gfsh import cluster-configuration command seem to differ with regard to whether the operation is allowed on running cache servers.

The operation is allowed, but with restrictions. The correct details are provided in the gfsh import command description.
This PR modifies the corresponding section under Exporting and Importing Cluster Configurations to be consistent with the gfsh write-up.


dkhopade (approved):
Looks ok to me. Thanks.

alb3rtobr (approved):
LGTM

PR #6676, opened 2021-07-06
GEODE-9416: Bump tomcat8 from 8.5.66 to 8.5.69

onichols-pivotal:
testing whether this hits the same test failures that 8.5.68 did...

### DIFF
--- a/buildSrc/src/main/groovy/org/apache/geode/gradle/plugins/DependencyConstraints.groovy
+++ b/buildSrc/src/main/groovy/org/apache/geode/gradle/plugins/DependencyConstraints.groovy
@@ -56,7 +56,7 @@ class DependencyConstraints implements Plugin<Project> {
     // This would possibly be corrected if they were proper source sets.
     deps.put("tomcat6.version", "6.0.37")
     deps.put("tomcat7.version", "7.0.109")
-    deps.put("tomcat8.version", "8.5.66")
+    deps.put("tomcat8.version", "8.5.69")
     deps.put("tomcat9.version", "9.0.48")
 
     // The jetty version is also hard-coded in geode-assembly:test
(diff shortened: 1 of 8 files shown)


sabbey37:
There was [a change](https://github.com/apache/tomcat/commit/f488fb08f8204126da79fd9d333c04d1ad936567#diff-7a7bd66ef966a17eca71b8f32e9bf763cb7f5ea7d45fad121453011f3824f412R1417) between Tomcat versions 8.5.66 and 8.5.67 as well as between 9.0.20 and 9.0.21 that resulted in an NPE being throw in `setAttribute`.  This issue was found and resolved by [GEODE-7896](https://issues.apache.org/jira/browse/GEODE-7896) and included in these Geode versions: 1.12.2, 1.13.3, 1.14.0+.  However, these upgrade tests pull in the most recent version of the Tomcat dependency (as specified in `TomcatInstall.java`), even though earlier versions of Geode used earlier versions of Tomcat that did not include this change.  This should not be an issue in the product at all since Geode is bundled with a specific Tomcat version and currently all Geode versions with a Tomcat version that includes the commit in question have the fix from GEODE-7896.

@onichols-pivotal if you know of a better way of making sure older versions of Geode pull in their corresponding versions of Tomcat or if you think we should spend some time investigating this, please let us know!  We are working on improving these tests as well as adding the same tests for Tomcat 9.

onichols-pivotal:
We have no control of what version of Tomcat someone might try to use with Geode.  We test several popular versions for compatibility, but we do not bundle Tomcat with Geode.

It sounds like maybe at most a release note is all that is needed to remind users to upgrade Geode before upgrading Tomcat?

sabbey37:
> It sounds like maybe at most a release note is all that is needed to remind users to upgrade Geode before upgrading Tomcat?

Yeah, that's a good call.  We can consult with @davebarnes97 on this.

onichols-pivotal:
closing due to inactivity

PR #6680, opened 2021-07-07
GEODE-9182: Implement Radish ZCOUNT command

DonalEvans:
- Add ZCOUNT to supported commands
 - Modify OrderStatisticsTree.indexOf() to return the index of elements that are
 not present in the collection instead of -1, to make the behaviour of
 OrderStatisticsTree and IndexibleTreeSet match
 - Add constants to the StringBytesGlossary to support parsing Redis
 range values and doing more comparisons on String member names
 - Replace hardcoded Strings with constants in
 AbstractHitsMissesIntegrationTest
 - Fix an inaccurate Javadoc introduced by a previous commit
 - Small code clean-ups in code related to ZRANGE
 - Add tests for ZCOUNT and new behaviour in OrderStatisticsTree

### For all changes:

### Note:

### DIFF
--- a/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/collections/OrderStatisticsSet.java
+++ b/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/collections/OrderStatisticsSet.java
@@ -53,20 +53,20 @@ public interface OrderStatisticsSet<T> extends Set<T> {
   T get(int index);
 
   /**
-   * Returns the index of <code>element</code> in the sorted set.
+   * Returns the index of <code>element</code> in the sorted set. If the element is not present in
+   * the set, returns the index it would have if it was present
    *
    * @param element the query element.
-   * @return the index of the query element or -1 if there is no such element
-   *         in this set.
+   * @return the index of the query element.
    */
   int indexOf(T element);
 
   /**
-   * Returns a range of <code>elements</code> between min and max.
+   * Returns an iterator over a range of <code>elements</code> between min and max.
    *
    * @param min the minimum element.
    * @param max the maximum element.
-   * @return an ArrayList of <code>elements</code>.
+   * @return an Iterator of <code>elements</code>.
    */
   Iterator<T> getIndexRange(int min, int max);
 }
--- a/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/collections/OrderStatisticsSet.java
+++ b/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/collections/OrderStatisticsSet.java
@@ -54,7 +54,7 @@ public interface OrderStatisticsSet<T> extends Set<T> {
 
   /**
    * Returns the index of <code>element</code> in the sorted set. If the element is not present in
-   * the set, returns the index it would have if it was present
+   * the set, returns the index it would have if it were present.
    *
    * @param element the query element.
    * @return the index of the query element.
--- a/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
+++ b/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
@@ -493,6 +493,15 @@ public class RedisSortedSet extends AbstractRedisData {
       this.score = score;
     }
 
+    public int compareTo(AbstractOrderedSetEntry o) {
+      int comparison = score.compareTo(o.score);
+      if (comparison == 0) {
+        // Scores equal, try lexical ordering
+        return compareMembers(member, o.member);
+      }
+      return comparison;
+    }
+
     public byte[] getScoreBytes() {
       return scoreBytes;
     }
@@ -500,6 +509,8 @@ public class RedisSortedSet extends AbstractRedisData {
     public byte[] getMember() {
       return member;
     }
+
+    public abstract int compareMembers(byte[] array1, byte[] array2);
   }
 
   // Entry used to store data in the scoreSet
@@ -510,13 +521,8 @@ public class RedisSortedSet extends AbstractRedisData {
     }
 
     @Override
-    public int compareTo(AbstractOrderedSetEntry o) {
-      int comparison = score.compareTo(o.score);
-      if (comparison == 0) {
-        // Scores equal, try lexical ordering
-        return javaImplementationOfAnsiCMemCmp(member, o.member);
-      }
-      return comparison;
+    public int compareMembers(byte[] array1, byte[] array2) {
+      return javaImplementationOfAnsiCMemCmp(array1, array2);
     }
   }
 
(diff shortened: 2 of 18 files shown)


DonalEvans:
The DUnit test failure is GEODE-9070, a known flaky test unrelated to the changes in this PR.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -289,7 +304,7 @@ long zcard() {
   }
 
   List<byte[]> zrange(int min, int max, boolean withScores) {
-    ArrayList<byte[]> result = new ArrayList<>();
+    List<byte[]> result = new ArrayList<>();
```
ringles, line 307: Thanks for fixing that.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -462,5 +503,19 @@ public int compareTo(OrderedSetEntry o) {
...
+    OrderedSetEntry(double score, boolean isExclusive, boolean isMinimum) {
+      if (isExclusive && isMinimum || !isExclusive && !isMinimum) {
+        // For use in (this < other) and (this >= other) comparisons
+        this.member = bGREATEST_MEMBER_NAME;
+      } else {
+        // For use in (this <= other) and (this > other) comparisons
+        this.member = bLEAST_MEMBER_NAME;
+      }
+      this.scoreBytes = new byte[] {};
+      this.score = score;
```
ringles, line 518: Do we need to allocate an empty array for scoreBytes? Maybe I'm missing a case, but if this doesn't get used could we just set it to null?

ringles (changes requested):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/collections/OrderStatisticsSet.java
```diff
@@ -53,20 +53,20 @@
   T get(int index);
 
   /**
-   * Returns the index of <code>element</code> in the sorted set.
+   * Returns the index of <code>element</code> in the sorted set. If the element is not present in
+   * the set, returns the index it would have if it was present
```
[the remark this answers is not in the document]
DonalEvans: "Incredibly pedantic" is my favourite kind of review :P

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
DonalEvans: Good call. Updated.

ringles (approved):
Looks good!

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -414,6 +429,28 @@ private static double processByteArrayAsDouble(byte[] value) {
     }
   }
 
+  // Comparison to allow the use of bLEAST_MEMBER_NAME and bGREATEST_MEMBER_NAME to always be less
+  // than or greater than respectively any other byte array representing a member name. The use of
+  // "==" is important as it allows us to differentiate between the constant byte arrays defined
+  // in StringBytesGlossary and user-supplied member names which may be equal in content but have
+  // a different memory address.
+  private static int checkDummyMemberNames(byte[] array1, byte[] array2) {
```
jdeppe-pivotal, line 437: I'm a little concerned that this is on a hot path and that the primary code path will require traversing all the conditionals. One alternative I can think of is to have these special sentinels try and actually be min/max values - `byte[] {-128, -128, -128, ...}` and `byte[] {127, 127, 127, ...}` for some largish sized byte array. That way the comparisons can just happen naturally. I realize though that in an extreme corner case a user may actually try and add such a value. I'm OK leaving it like this for now, but at some point I'd like to satisfy myself on the performance aspects.

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/collections/OrderStatisticsTreeTest.java
```diff
@@ -193,16 +194,19 @@ public void testSize() {
...
+    for (int i = 0; i < entries; ++i) {
+      assertThat(tree.add(i * stepSize)).isTrue();
     }
 
-    for (int i = 100; i < 200; ++i) {
-      assertThat(tree.indexOf(i)).isEqualTo(i - 100);
-    }
+    // Test values smaller than the lowest entry and greater than the largest entry
+    assertThat(tree.indexOf(-1)).isEqualTo(0);
+    assertThat(tree.indexOf(1 + entries * stepSize)).isEqualTo(tree.size());
```
jdeppe-pivotal, line 206: The `1 +` is unnecessary here.

jdeppe-pivotal (approved):
Looks good. Some minor comments but not necessary to fix if you don't have any other pending changes.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
DonalEvans: That could potentially work for the least possible member name, which would just be `byte[] {-128}` but to have a true greatest possible member name, it'd have to be `byte[] {127, 127, ...}` for the longest possible byte array, as any byte array longer than it would evaluate to being greater than it. Using such a long byte array in the "normal" comparison method potentially leads to a lot more operations, since we have to compare element by element rather than just do one `==` operation. I see your point though, so I'll refactor this method to do fewer checks if neither of the arrays are equal to the constants. If it's any comfort, Redis uses the approach implemented here, where they have a defined constant as their min/man (they use "minstring" and "maxstring").

DonalEvans:
After sleeping on it, I think I've come up with a better solution, which moves the check for the dummy member names out of the `OrderedSetEntry` class and into a new class, `DummyOrderedSetEntry` which is only used in ZCOUNT, thus moving those comparisons out of the hot path.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -470,10 +470,28 @@ public static int javaImplementationOfAnsiCMemCmp(byte[] array1, byte[] array2)
     // with same name...
   }
 
-  static class OrderedSetEntry implements Comparable<OrderedSetEntry> {
-    private final byte[] member;
-    private final byte[] scoreBytes;
-    private final Double score;
+  abstract static class AbstractOrderedSetEntry implements Comparable<AbstractOrderedSetEntry> {
```
jdeppe-pivotal, line 473: I like this. One small suggestion would be to pull `compareTo` into this class but declare a separate `abstract` method that is the only abstraction difference in subclasses:
```
    protected abstract int internalCompare(byte[] b1, byte[] b2);

    @Override
    public int compareTo(AbstractOrderedSetEntry o) {
      int comparison = score.compareTo(o.score);
      if (comparison == 0) {
        // Scores equal, try lexical ordering
        comparison = internalCompare(member, o.member);
      }
      return comparison;
    }
  }
```

DonalEvans: Ah, good call, that's neater.

PR #6682, opened 2021-07-09
GEODE-9419: Handle IOExceptions when Radish clients disconnect

jdeppe-pivotal:
- This restores some of the behavior of GEODE-9411

### For all changes:

### Note:

### DIFF
--- a/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/netty/ExecutionHandlerContext.java
+++ b/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/netty/ExecutionHandlerContext.java
@@ -16,6 +16,7 @@
 package org.apache.geode.redis.internal.netty;
 
 
+import java.io.IOException;
 import java.math.BigInteger;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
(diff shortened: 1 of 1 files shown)


DonalEvans:
Some small refactoring changes aside, this seems to just be a revert of GEODE-9411. Is there some middle ground solution that would prevent the spurious exception logging from Netty but preserve the behaviour where we don't close the connection if we hit any `IOException` at all, regardless of the origin?

jdeppe-pivotal:
Since `IOException`s are checked, the implication is that they may be recoverable. However, in this case since netty is telling us about this problem, there isn't anything we can do about it. We're being told that the client is gone and we can perform any necessary cleanup.

DonalEvans:
> Since `IOException`s are checked, the implication is that they may be recoverable. However, in this case since netty is telling us about this problem, there isn't anything we can do about it. We're being told that the client is gone and we can perform any necessary cleanup.

Aah, I see now. As long as we can be sure that we only see an `IOException` when the client is gone, and not in other cases, then it should be fine to just revert the entire commit for GEODE-9411. If we think there might be cases where an `IOException` is thrown and we don't want to close the connection, maybe we could check for the "Connection reset by peer" message in the exception and close if we see that?

jdeppe-pivotal:
Closing in favor of a complete revert.

PR #6684, opened 2021-07-09
GEODE-6588: Cleanup multi-user auth.

pivotal-jbarrett:
While spending some time in multi-user auth I cleaned up some of the classes I was spending time in understanding.

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/cache/client/internal/OpExecutorImpl.java
+++ b/geode-core/src/main/java/org/apache/geode/cache/client/internal/OpExecutorImpl.java
@@ -25,6 +25,8 @@ import java.util.List;
 import java.util.Set;
 
 import org.apache.logging.log4j.Logger;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
 import org.apache.geode.CancelCriterion;
 import org.apache.geode.CancelException;
@@ -50,7 +52,6 @@ import org.apache.geode.cache.execute.FunctionException;
 import org.apache.geode.cache.execute.FunctionInvocationTargetException;
 import org.apache.geode.distributed.internal.ServerLocation;
 import org.apache.geode.distributed.internal.ServerLocationAndMemberId;
-import org.apache.geode.internal.cache.PoolManagerImpl;
 import org.apache.geode.internal.cache.PutAllPartialResultException;
 import org.apache.geode.internal.cache.execute.InternalFunctionInvocationTargetException;
 import org.apache.geode.internal.cache.tier.BatchException;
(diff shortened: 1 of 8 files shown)


geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientProxyMembershipID.java
```diff
@@ -176,149 +177,114 @@ public static ClientProxyMembershipID getClientId(DistributedMember member) {
...
   }
 
   public ClientProxyMembershipID() {}
 
-  public ClientProxyMembershipID(DistributedMember member) {
-    this.uniqueId = 1;
-    this.memberId = member;
+  public ClientProxyMembershipID(final DistributedMember member) {
+    this(null, 1, member);
     updateID(member);
```
Bill, line 225: comment: I love what you did to chain the constructors. I see you had to introduce a private pure function `getUniqueId()` to help.

I find it hokey that in this method we set `identity` to `null`, only to turn around and call the (mutating) `updateId()` right after that.

I was about to suggest extracting the guts of `updateID()` into a pure function you could call here to do likewise, until I went and looked at `updateID()` and saw it mutates no fewer than four (4) member variables. Ugh.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientProxyMembershipID.java
```diff
@@ -176,149 +177,114 @@ public static ClientProxyMembershipID getClientId(DistributedMember member) {
...
+    return _toString;
   }
 
   /**
    * returns a string representation of this identifier, ignoring the toString cache
    */
   public String toStringNoCache() {
-    StringBuffer sb = new StringBuffer("identity(").append(getDSMembership()).append(",connection=")
-        .append(uniqueId);
+    StringBuilder sb =
```
Bill, line 254: request: No need for a `StringBuilder` here. If you're bothering to change this code I recommend changing it to use the `+` operator.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientProxyMembershipID.java
```diff
@@ -176,149 +177,114 @@ public static ClientProxyMembershipID getClientId(DistributedMember member) {
...
-   * For Externalizable
-   *
-   * @see Externalizable
-   */
   @Override
   public void readExternal(ObjectInput in) throws IOException, ClassNotFoundException {
     int identityLength = in.readShort();
-    if (identityLength > BYTES_32KB) {
-      throw new IOException(
-          "HandShake identity length is too big");
```
Bill, line 298: comment: dead code. nice catch!

Bill (changes requested):
Mostly congratulations on cleaning up the code and making it easier to understand.

A couple requests to eliminate `StringBuilder`s you introduced.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientProxyMembershipID.java
pivotal-jbarrett: Yeah, I agree. I saw all that and said nope!

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientProxyMembershipID.java
pivotal-jbarrett: Even more so, it was wrong code. Notice the original max constant was off by 1.

geode-core/src/main/java/org/apache/geode/cache/client/internal/OpExecutorImpl.java
```diff
@@ -519,161 +509,161 @@ protected void handleException(Op op, Throwable e, Connection conn, int retryCou
...
-      throw exToThrow;
+    if (exceptionToThrow != null) {
+      throw exceptionToThrow;
     }
   }
 
-  private StringBuffer getExceptionMessage(String exceptionName, int retryCount,
-      boolean finalAttempt, Connection connection) {
-    StringBuffer message = new StringBuffer(200);
+  private StringBuilder getExceptionMessage(final String exceptionName, final int retryCount,
```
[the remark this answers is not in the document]
pivotal-jbarrett: The if statement makes it necessary otherwise internally it constructs two new strings.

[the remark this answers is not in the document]
Bill: I confirmed what @pivotal-jbarrett said by converting the explicit `StringBuilder` to `+` and viewing the byte code. Sure enough, the byte code contains lots of `StringBuilder` allocations and lots of `toString()` calls.

That being said, is this a performance critical path? It is only invoked if there is an exception.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientProxyMembershipID.java
Bill: yep!

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientProxyMembershipID.java
Bill: (see conversation on other `StringBuilder`) I'm resolving this comment because this is called in a `toString()` and who knows, maybe it can affect performance.

Bill (approved):
Approved!

PR #6685, opened 2021-07-09
GEODE-9340: move benchmark baseline

nonbinaryprogrammer:
The new baseline is the commit that caused a slight performance
regression in our most sensitive tests. There is a negligible
performance difference between the previous baseline of 1.13.3 and the
commit before the commit that caused the regression (commit with SHA
217be41f087072026ea3b411b89107700edfb5f7). This means that moving the
baseline will not ignore a bunch of other performance changes.

### For all changes:

### Note:

### DIFF
--- a/ci/pipelines/shared/jinja.variables.yml
+++ b/ci/pipelines/shared/jinja.variables.yml
@@ -17,7 +17,7 @@
 
 benchmarks:
   baseline_branch: ''
-  baseline_version: '1.13.3'
+  baseline_version: '217be41f087072026ea3b411b89107700edfb5f7'
   benchmark_branch: ((geode-build-branch))
   flavors:
   - title: 'base'


dickcav (approved):

onichols-pivotal:
This will be overwritten by the next release (either 1.13.4 or 1.14.0), is that ok?  Or does this need to be pinned in perpetuity?

nonbinaryprogrammer:
> This will be overwritten by the next release (either 1.13.4 or 1.14.0), is that ok? Or does this need to be pinned in perpetuity?

Yes, that is fine. this is a temporary fix to move the baseline past that regression so that tests stop failing in CI. The next release will also be passed that regression, so it will also be ok. We just need the baseline to be something after the dependency version bumps

DonalEvans (approved):

onichols-pivotal:
> We just need the baseline to be something after the dependency version bumps

But 1.13.3 has the same jetty dependency bump...

nonbinaryprogrammer:
> > We just need the baseline to be something after the dependency version bumps
> 
> But 1.13.3 has the same jetty dependency bump...

1.13.3 does not contain the dependency bump, which is a problem. if 1.13.4 does not contain the dependency bump and is the next released version, then we will need to keep the baseline as the above commit until a version that contains it. so definitely 1.14.

nonbinaryprogrammer:
Closing as unnecessary and closing out the ticket.

When I left for vacation two weeks ago, the baseline was 1.13.2, which was causing the performance issue. During those two weeks, the baseline was moved to 1.13.3, which resolves the performance issue. So setting the baseline to this specific commit is no longer necessary. This change was made by the following commit: 5b5748300fb7b783e1a75c54f0dda750b89cc4f5

PR #6686, opened 2021-07-09
GEODE-9368: Changes to support better resiliency for HA scenarios

jdeppe-pivotal:
- Wrap all StripedExecutor.execute calls with
  PartitionedRegion.computeWithPrimaryLocked.
- Handle RegionDestroyedException as a result of trying to lock a bucket
  while the bucket is moving.
- Improve race condition handling in *ScanExecutor calls
- Rename responds with MOVED when the bucket is moving during the call.

### For all changes:

### Note:


jdeppe-pivotal:
I've tried to break this PR into one commit for core changes and another commit for test-related changes

onichols-pivotal (approved):

dickcav (approved):

rhoughton-pivot (approved):

kirklund (commented):
Is it possible to make HScanDunitTest and CrashAndNoRepeatDUnitTest a bit more robust? They're failing the stress-new-test tasks and so it's likely they'll become regular flaky failures.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisKeyCommandsFunctionExecutor.java
```diff
@@ -136,7 +136,11 @@ public void restore(RedisKey key, long ttl, byte[] data, RestoreOptions options)
   @Override
   public boolean rename(RedisKey oldKey, RedisKey newKey) {
     List<RedisKey> orderedKeys = orderForLocking(oldKey, newKey);
-    return stripedExecute(orderedKeys.get(0), () -> rename0(orderedKeys.get(1), oldKey, newKey));
+    return stripedExecute(orderedKeys.get(0), () -> {
+      // Trigger MOVED if necessary
+      getRedisData(newKey);
```
dschneider-pivotal, line 141: I'm not sure I understand why this call of getRedisData(newKey) has been added but here are some questions/comments about it. Calling getRedisData(newKey) seems like a race. My understanding is that you are calling it so that it can throw an exception saying the bucket moved. But couldn't move right after you finish calling it and before rename0 is called? And since the get can go to the primary or secondary but the put that rename0 does on newKey goes to the primary isn't possible for that put to fail anyway? Also at this point of time we don't know that we have locked newKey. So if getRedisData(newKey) is needed shouldn't it be done in rename0 after both newKey and oldKey have been locked?

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/hash/HScanExecutor.java
```diff
@@ -67,12 +67,13 @@ public RedisResponse executeCommand(Command command,
     }
 
     RedisKey key = command.getKey();
-    if (!getDataRegion(context).containsKey(key)) {
+    RedisData value = context.getRegionProvider().getRedisData(key, null);
```
dschneider-pivotal, line 70: you could use "getRedisData(key)" to set value and then "value.isNull()" instead of "value == null".

jdeppe-pivotal:
> Is it possible to make HScanDunitTest and CrashAndNoRepeatDUnitTest a bit more robust? They're failing the stress-new-test tasks and so it's likely they'll become regular flaky failures.

Yes, I'm still working on making them non-flaky

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisKeyCommandsFunctionExecutor.java
jdeppe-pivotal: `stripedExecute` calls `PartitionedRegion.computeWithPrimaryLocked` with the lambda defined here so the bucket shouldn't be able to move as part of this execution sequence.

dschneider-pivotal: Jens and I talked about this and it turns out we now require that the two keys passed to rename are on the same primary bucket. That is why having done the first stripedExecute will have locked the primary even though the first stripedExecute may have been done on oldKey not newKey. We agreed that it would be better to do a check to ensure that both keys are on the same primary  and then this extra getRedisData(newKey) call is not needed.

geode-apis-compatible-with-redis/src/commonTest/java/org/apache/geode/test/dunit/rules/RedisClusterStartupRule.java
```diff
@@ -154,4 +161,47 @@ public void enableDebugLogging(int vmId) {
...
+          () -> new IllegalStateException("No non-hosting member found for key: " + key));
+
+      PartitionRegionHelper.moveBucketByKey(r, primaryMember, targetMember, redisKey);
+
+      // Who is the primary now?
+      return PartitionRegionHelper.getPrimaryMemberForKey(r, redisKey);
+    });
+  }
+
+  public String getKeyOnServer(String keyPrefix, int vmId) {
```
DonalEvans, line 193: This method seems like it could use some documentation around its intended use, since it's not immediately clear how it's supposed to work.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/RedisRestoreKeyExistsException.java
```diff
@@ -17,7 +17,7 @@
 /**
  * An exception thrown when the key being restored already exists.
  */
-public class RedisRestoreKeyExistsException extends RuntimeException {
+public class RedisRestoreKeyExistsException extends RedisException {
```
DonalEvans, line 20: Very minor point, but might this class be better located in the `org.apache.geode.redis.internal.data` package?

DonalEvans (approved):
Looks good, with the understanding that more work is coming to fix the failing stress tests.

geode-apis-compatible-with-redis/src/commonTest/java/org/apache/geode/redis/GeodeRedisServerRule.java
```diff
@@ -49,6 +49,12 @@ protected void before() {
     cache = cacheFactory.create();
     server = new GeodeRedisServer("localhost", 0, (InternalCache) cache);
     server.setAllowUnsupportedCommands(enableUnsupportedCommands);
+
+    // Ensure that buckets are created up front
+    try {
```
dschneider-pivotal, line 54: I think this block should be refactored into a method on GeodeRedisServer.
Also it looks to me like it will be happy even if a bucket can not be created. It looks like we basically need to call SlotAdvisor.getMemberInfo for each bucketId but that method has some retry logic in it and it can fail to get the member info. In that case it logs an error and returns null but this code ends up ignoring those nulls and moving on. I don't know if you want something different here that would fail the startup if one of the memberInfo calls times out.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/set/SScanExecutor.java
```diff
@@ -66,12 +67,13 @@ public RedisResponse executeCommand(Command command, ExecutionHandlerContext con
 
     RedisKey key = command.getKey();
 
-    if (!getDataRegion(context).containsKey(key)) {
+    RedisData value = context.getRegionProvider().getRedisData(key, null);
```
dschneider-pivotal, line 70: this is another place that could be changed to "context.getRegionProvider().getRedisData(key);
      if (value.isNull()"

geode-apis-compatible-with-redis/src/commonTest/java/org/apache/geode/test/dunit/rules/RedisClusterStartupRule.java
jdeppe-pivotal: Done

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/RedisRestoreKeyExistsException.java
jdeppe-pivotal: Done

geode-apis-compatible-with-redis/src/commonTest/java/org/apache/geode/redis/GeodeRedisServerRule.java
jdeppe-pivotal: Regarding the error case where the member info cannot be retrieved - that would result in a 'non-existent' slot (or range of slots) at which point any clients should produce some kind of error if they were trying to write to a missing slot.

Regarding pulling this into `GeodeRedisServer` - are you thinking it should run as part of startup? If so, I'd be wary of that since it would create a bucket imbalance unless we also did a rebalance as part of that lifecycle. As it is, any cluster-aware client will always retrieve the topology as the first thing it does which would implicitly create the buckets. It's still a problem if all members are not up before the very first client starts to connect.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/set/SScanExecutor.java
jdeppe-pivotal: Fixed

PR #6699, opened 2021-07-14
GEODE-9304: Make RedisSortedSet measurement of bytes in use accurate

DonalEvans:
- Add accurate size calulation to RedisSortedSet
 - Make OrderSatisticsSet implement Sizeable
 - Add size tracking to OrderStatisticsTree and rename to
 SizeabelOrderStatisticsTree
 - Modify how SizeableObject2ObjectOpenCustomHashMapWithCursor
 calculates the size of elements to support elements implementing
 Siazeable
 - Make AbstractOrderedSetEntry implement Sizeable
 - Add size calculation to OrderedSetEntry
 - Add unit tests for all new behaviour
 - Add memory overhead integration and acceptance tests for sorted sets

### For all changes:

### Note:

### DIFF
--- a/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/collections/SizableOrderStatisticsTree.java
+++ b/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/collections/SizableOrderStatisticsTree.java
@@ -38,6 +38,9 @@ import java.util.Objects;
 import java.util.Set;
 
 import org.apache.geode.annotations.VisibleForTesting;
+import org.apache.geode.internal.size.ReflectionSingleObjectSizer;
+import org.apache.geode.internal.size.SingleObjectSizer;
+import org.apache.geode.internal.size.Sizeable;
 
 /**
  * This class implements an order statistic tree which is based on AVL-trees.
@@ -46,11 +49,15 @@ import org.apache.geode.annotations.VisibleForTesting;
  * @author Rodion "rodde" Efremov
  * @version 1.6 (Feb 11, 2016)
  */
-public class OrderStatisticsTree<E extends Comparable<? super E>>
+public class SizableOrderStatisticsTree<E extends Comparable<? super E>>
     implements OrderStatisticsSet<E> {
+  public static final int ORDER_STATISTICS_TREE_BASE_SIZE = 32;
+  public static final int PER_ENTRY_OVERHEAD = 40;
   private Node<E> root;
   private int size;
   private int modCount;
+  private int sizeInBytes = ORDER_STATISTICS_TREE_BASE_SIZE;
+  private static final SingleObjectSizer elementSizer = new ReflectionSingleObjectSizer();
 
   @Override
   public Iterator<E> iterator() {
@@ -153,6 +160,7 @@ public class OrderStatisticsTree<E extends Comparable<? super E>>
       root = new Node<>(element);
       size = 1;
       modCount++;
+      updateSizeInBytes(element, true);
       return true;
     }
 
@@ -188,6 +196,7 @@ public class OrderStatisticsTree<E extends Comparable<? super E>>
     newNode.parent = parent;
     size++;
     modCount++;
+    updateSizeInBytes(element, true);
     Node<E> hi = parent;
     Node<E> lo = newNode;
 
@@ -258,6 +267,7 @@ public class OrderStatisticsTree<E extends Comparable<? super E>>
     }
 
     x = deleteNode(x);
+    updateSizeInBytes(x.key, false);
     fixAfterModification(x, false);
     size--;
     modCount++;
@@ -731,6 +741,25 @@ public class OrderStatisticsTree<E extends Comparable<? super E>>
     return leftTreeSize + 1 + rightTreeSize;
   }
 
+  void updateSizeInBytes(E element, boolean isAdd) {
+    int sizeChange;
+    if (element instanceof Sizeable) {
+      sizeChange = ((Sizeable) element).getSizeInBytes() + PER_ENTRY_OVERHEAD;
+    } else {
+      sizeChange = (int) elementSizer.sizeof(element) + PER_ENTRY_OVERHEAD;
+    }
+    if (isAdd) {
+      sizeInBytes += sizeChange;
+    } else {
+      sizeInBytes -= sizeChange;
+    }
+  }
+
+  @Override
+  public int getSizeInBytes() {
+    return sizeInBytes;
+  }
+
   private static final class Node<T> {
     T key;
 
@@ -800,6 +829,7 @@ public class OrderStatisticsTree<E extends Comparable<? super E>>
       checkConcurrentModification();
 
       Node<E> x = deleteNode(previousNode);
+      updateSizeInBytes(x.key, false);
       fixAfterModification(x, false);
 
       if (x == nextNode) {
--- a/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/collections/SizeableObject2ObjectOpenCustomHashMapWithCursor.java
+++ b/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/collections/SizeableObject2ObjectOpenCustomHashMapWithCursor.java
@@ -197,10 +197,10 @@ public class SizeableObject2ObjectOpenCustomHashMapWithCursor<K, V>
     V oldValue = super.put(k, v);
     if (oldValue == null) {
       // A create
-      arrayContentsOverhead += (int) (elementSizer.sizeof(k) + elementSizer.sizeof(v));
+      arrayContentsOverhead += (getElementSize(k) + getElementSize(v));
     } else {
       // An update
-      arrayContentsOverhead += (int) (elementSizer.sizeof(v) - elementSizer.sizeof(oldValue));
+      arrayContentsOverhead += (getElementSize(v) - getElementSize(oldValue));
     }
     return oldValue;
   }
@@ -209,7 +209,7 @@ public class SizeableObject2ObjectOpenCustomHashMapWithCursor<K, V>
   public V remove(Object k) {
     V oldValue = super.remove(k);
     if (oldValue != null) {
-      arrayContentsOverhead -= (elementSizer.sizeof(k) + elementSizer.sizeof(oldValue));
+      arrayContentsOverhead -= (getElementSize(k) + getElementSize(oldValue));
     }
     return oldValue;
   }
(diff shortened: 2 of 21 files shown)


geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -171,20 +170,24 @@ public int getDSFID() {
...
+      // to them appear in both backing collections.
+      sizeInBytesAdjustment += newEntry.getSizeInBytes() + calculateByteArraySize(memberToAdd);
+      return null;
     } else {
-      scoreSet.remove(orderedSetEntry);
+      scoreSet.remove(existingEntry);
       scoreSet.add(newEntry);
-      oldScore = orderedSetEntry.getScoreBytes();
-      sizeInBytes += scoreToAdd.length - oldScore.length;
+      // When updating an entry, the references to the member name array in the backing collections
```
dschneider-pivotal, line 184: It seems to me like we should in this update case canonicalize the memberToAdd byte array so that we reuse the one already stored in the members map instead of this new one. I think you could do this by doing a OrderedSetEntry existingEntry = members.get(memberToAdd) and if it returns non-null then reading existingEntry.getMember() and using that to create the newEntry. If OrderedSetEntry is always read while synchronized then you could make it mutable and in that case you wouldn't need the update to create a new entry each time. I think you would need to do the scoreSet.remove(existingEntry) then mutate existingEntry and then scoreSet.add(existingEntry). This would prevent some gc churn.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/collections/SizableOrderStatisticsTree.java
```diff
@@ -46,11 +49,15 @@
  * @author Rodion "rodde" Efremov
  * @version 1.6 (Feb 11, 2016)
  */
-public class OrderStatisticsTree<E extends Comparable<? super E>>
+public class SizableOrderStatisticsTree<E extends Comparable<? super E>>
```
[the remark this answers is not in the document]
DonalEvans: It probably doesn't need to be renamed, since we don't have a non-sizeable implementation of OrderedStatisticsSet, so it doesn't make sense to distinguish it like this. I'll put the name back how it was.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/collections/SizableOrderStatisticsTree.java
```diff
@@ -731,6 +741,25 @@ private int count(Node<E> node) {
     return leftTreeSize + 1 + rightTreeSize;
   }
 
+  void updateSizeInBytes(E element, boolean isAdd) {
+    int sizeChange;
+    if (element instanceof Sizeable) {
```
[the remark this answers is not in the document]
DonalEvans: I don't think we do this anywhere else, but I did it here to avoid having to use ReflectionSingleObjectSizer for objects that are already keeping track of their size, since that would reduce overhead (I think). Your suggestion is a good one though, so I'll implement it here.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -34,45 +35,44 @@
...
   private OrderStatisticsSet<AbstractOrderedSetEntry> scoreSet;
+  // This field is used to keep track of the size associated with objects that are referenced by
+  // both backing collections, since they will be counted twice otherwise
+  private int sizeInBytesAdjustment = 0;
 
-  protected static final int BASE_REDIS_SORTED_SET_OVERHEAD = 184;
-  protected static final int PER_PAIR_OVERHEAD = 48;
-
-  private int sizeInBytes = BASE_REDIS_SORTED_SET_OVERHEAD;
+  protected static final int BASE_REDIS_SORTED_SET_OVERHEAD = 40;
```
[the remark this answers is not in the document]
DonalEvans: These constants are determined experimentally using the tests found in `RedisSortedSetTest`. I'll add a comment here pointing that out though, so people know which tests to run if they modify these classes.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
DonalEvans: Ah, I had tried to find a solution for that when I was writing this, but I ran out of steam before I could figure one out. Yours works perfectly, thanks!

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/collections/SizeableObject2ObjectOpenCustomHashMapWithCursor.java
```diff
@@ -226,6 +226,14 @@ int calculateBackingArraysOverhead() {
         + BACKING_ARRAY_LENGTH_COEFFICIENT * getTotalBackingArrayLength();
   }
 
+  private <E> int getElementSize(E element) {
+    if (element instanceof Sizeable) {
```
dschneider-pivotal, line 230: could this class also use your new SizeableObjectSizer?

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -173,20 +176,23 @@ public int getDSFID() {
...
+      return null;
     } else {
-      scoreSet.remove(orderedSetEntry);
-      scoreSet.add(newEntry);
-      oldScore = orderedSetEntry.getScoreBytes();
-      sizeInBytes += scoreToAdd.length - oldScore.length;
+      scoreSet.remove(existingEntry);
+      byte[] oldScore = existingEntry.scoreBytes;
+      existingEntry.updateScore(stripTrailingZeroFromDouble(scoreToAdd));
+      members.put(memberToAdd, existingEntry);
```
dschneider-pivotal, line 192: why is this members.put needed when we modified an existing entry? Is it simply to update the size if the byte array form has a different length? We are using extra memory by storing both the byte array form and the double form. We would use less memory if we just stored it as a "double". In this case storing it in both forms uses more CPU since we need to do the extra put. What operations are being optimized by storing it in both forms?

geode-core/src/main/java/org/apache/geode/internal/size/SizeableObjectSizer.java
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * A sizer that allows for efficient sizing of objects that implement the {@link Sizeable}
+ * interface, and delegates to {@link ReflectionSingleObjectSizer} otherwise.
+ */
+public class SizeableObjectSizer implements SingleObjectSizer {
+
+  private final SingleObjectSizer sizer = new ReflectionSingleObjectSizer();
+
+  @Override
+  public long sizeof(Object object) {
```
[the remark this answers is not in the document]
DonalEvans: The `SingleObjectSizer` interface that `SizeableObjectSizer` implements returns a long from its `sizeof()` method, so returning an int here would mean either breaking the inheritance relationship between the two or changing `SingleObjectSizer` to return int instead of long, which will also affect `CachingSingleObjectSizer`, `InstrumentationSingleObjectSizer` and `ReflectionSingleObjectSizer`, with further knock-on effects that are outside the scope of this PR. As a compromise, would it be acceptable to leave the return type as long, but call `Coder.narrowLongToInt()` on the returned value rather than casting it? This will prevent integer overflow in the size, but avoid the need to do refactoring outside the redis module.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/collections/SizeableObject2ObjectOpenCustomHashMapWithCursor.java
DonalEvans: Yes, good catch, thanks

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
DonalEvans: I think that the reason we store both the byte array and double is that some of the SortedSet commands (zadd and zrem) were implemented before we had an actual sorted set implementation, so were just using the member and score byte arrays in a `SizeableObject2ObjectOpenCustomHashMapWithCursor`. With that original implementation, it was faster to leave the scores in their byte array format at all times, but after new commands and the `OrderStatisticsSet` were added to RedisSortedSet, this is no longer the case.

I think that the only time we need a byte array rather than a double is when we're sending the response, so we can eliminate the field from `AbstractOrderedStatisticsEntry` entirely and just change how we return from certain sorted set methods. However, this is a potentially large refactor that's not strictly within the scope of this PR, so I'd prefer to open a new ticket for it rather than do it as part of this one.

dschneider-pivotal: sounds good

dschneider-pivotal: you might want to add a note to that new ticket that this extra members.put call can be removed once this modify no longer changes the size because we are just updating the score stored as a "double".

geode-core/src/main/java/org/apache/geode/internal/size/SizeableObjectSizer.java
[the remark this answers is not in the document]
dschneider-pivotal: Couldn't you just change the new SizeableObjectSizer to not implement the interface? It could be in the redis module instead of core. It is currently only used by redis. I think the classes that use it don't need it to be a SingleObjectSizer. I like the idea of it using Coder.narrowLongToInt but would prefer it to be done in one place instead of all the callers.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
DonalEvans: Will do

geode-core/src/main/java/org/apache/geode/internal/size/SizeableObjectSizer.java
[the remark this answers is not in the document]
DonalEvans: Yeah, that would work. I'll move the Sizer and not have it implement the interface.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
DonalEvans: Filed GEODE-9446 to track the refactoring work.

nonbinaryprogrammer (approved):

dschneider-pivotal (approved):

PR #6700, opened 2021-07-15
GEODE-9378: Implement ZRANGEBYSCORE

ringles:
Addresses GEODE-9378 to implement the ZRANGEBYSCORE command for Geode Radish, the compatible-with-Redis API.


geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSetCommandsFunctionExecutor.java
```diff
@@ -62,6 +62,13 @@ public long zcount(RedisKey key, SortedSetRangeOptions rangeOptions) {
         () -> getRedisSortedSet(key, true).zrange(min, max, withScores));
   }
 
+  @Override
+  public List<byte[]> zrangebyscore(RedisKey key, SortedSetRangeOptions rangeOptions,
+      boolean withScores) {
+    return stripedExecute(key,
+        () -> getRedisSortedSet(key, true).zrangebyscore(rangeOptions, withScores));
```
dschneider-pivotal, line 69: I noticed you are calling getRedisSortedSet with updatesStats set to true. This seems reasonable. But I also noticed that a bunch of other methods on this class (zadd, zincrby, zrange, zrem) call it with false. Do you know when to set it to true vs false?

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -307,6 +308,57 @@ long zcount(SortedSetRangeOptions rangeOptions) {
     return getRange(min, max, withScores, false);
   }
 
+
+  List<byte[]> zrangebyscore(SortedSetRangeOptions rangeOptions, boolean withScores) {
+    List<byte[]> result = new ArrayList<>();
+    AbstractOrderedSetEntry minEntry =
+        new DummyOrderedSetEntry(rangeOptions.getMinDouble(), rangeOptions.isMinExclusive(), true);
+    long minIndex = scoreSet.indexOf(minEntry);
```
dschneider-pivotal, line 316: why is minIndex and maxIndex typed as long? indexOf returns an int and later we cast it to an int. Seems like if you typed it as "int" you could get rid of the cast later.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/ZRangeByScoreExecutor.java
```diff
@@ -0,0 +1,102 @@
...
+    int offset;
+    int count;
+    if (equalsIgnoreCaseBytes(commandElements.get(commandIndex), bRADISH_LIMIT)) {
+      offset = narrowLongToInt(bytesToLong(commandElements.get(commandIndex + 1)));
+      count = narrowLongToInt(bytesToLong(commandElements.get(commandIndex + 2)));
+      if (count < 0) {
+        count = Integer.MAX_VALUE;
+      }
+    } else {
+      throw new Exception();
```
dschneider-pivotal, line 98: can you make this a more specific exception? When the higher level catches Exception it could be catching all kinds of things (like NullPointerException) and reporting them as something else

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSetCommandsFunctionExecutor.java
jdeppe-pivotal: I believe that Redis' approach is to update the stats for commands that don't update data. I would expect that `zrange` _should_ update these stats.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
ringles: They're typed as long because they're essentially the same code as in ZCOUNT. But it makes more sense for them to be int here, as you say.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSetCommandsFunctionExecutor.java
ringles: I see ZRANGE calling it with true. In general, operations that read data update stats, operations that change data do not. So zadd, zrem, zincrby change data and therefore don't update stats.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/ZRangeByScoreExecutor.java
ringles: IllegalArgumentException seems to fit. Done.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZRangeByScoreIntegrationTest.java
```diff
@@ -0,0 +1,269 @@
...
+import org.junit.Before;
+import org.junit.Test;
+import redis.clients.jedis.HostAndPort;
+import redis.clients.jedis.JedisCluster;
+import redis.clients.jedis.Protocol;
+import redis.clients.jedis.Tuple;
+
+import org.apache.geode.redis.RedisIntegrationTest;
+
+public abstract class AbstractZRangeByScoreIntegrationTest implements RedisIntegrationTest {
```
jdeppe-pivotal, line 40: There are a lot of boundary cases tested, but please would you also add a test for exclusivity.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -307,6 +308,57 @@ long zcount(SortedSetRangeOptions rangeOptions) {
...
+    }
+    int skip = 0;
+    int returnedCount = 0;
+
+    while (entryIterator.hasNext() && returnedCount < count) {
+      AbstractOrderedSetEntry entry = entryIterator.next();
+      if (skip < offset) {
+        skip++;
+        continue;
+      }
```
jdeppe-pivotal: You could avoid use of `skip` here and just count down on `offset`:
```
if (offset > 0) {
  offset--;
  continue;
}
```

jdeppe-pivotal (changes requested):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
ringles: Nice; I think I can do that with 'count' as well!

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -307,6 +308,58 @@ long zcount(SortedSetRangeOptions rangeOptions) {
...
+
+    // Okay, if we make it this far there's a potential range of things to return.
+    int offset = 0;
+    int count = Integer.MAX_VALUE;
+    Iterator<AbstractOrderedSetEntry> entryIterator =
+        scoreSet.getIndexRange(minIndex, maxIndex, false);
+    if (rangeOptions.hasLimit()) {
+      count = rangeOptions.getCount();
+      offset = rangeOptions.getOffset();
+    }
```
DonalEvans, line 336: Rather than getting the range iterator and then skipping to the offset, might it be neater to add the offset to `minIndex` and then retrieve the range iterator? Similarly, I think you should be able to pass `Math.min(count, maxIndex - minIndex)` as the second argument of `getIndexRange()` so that you get an iterator that's automatically the right size. Then you can just iterate the entire length and return that with no further checks needed.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZRangeByScoreIntegrationTest.java
DonalEvans: Could you also add tests to confirm the behaviour of ZRANGEBYSCORE with multiple instances of `WITHSCORES` and `LIMIT`, and with incorrectly formatted `LIMIT` arguments?

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/ZRangeByScoreExecutor.java
```diff
@@ -0,0 +1,105 @@
...
+    return RedisResponse.array(result);
+  }
+
+  void parseLimitArguments(SortedSetRangeOptions rangeOptions, List<byte[]> commandElements,
+      int commandIndex) {
+    int offset;
+    int count;
+    if (equalsIgnoreCaseBytes(commandElements.get(commandIndex), bRADISH_LIMIT)) {
+      offset = narrowLongToInt(bytesToLong(commandElements.get(commandIndex + 1)));
+      count = narrowLongToInt(bytesToLong(commandElements.get(commandIndex + 2)));
```
DonalEvans, line 96: With an improperly provided `LIMIT` argument (one or both of `offset` or `count` not supplied) this code will throw an `ArrayIndexOutOfBoundsException` which is not handled, so no syntax error will be returned to the client.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/netty/StringBytesGlossary.java
```diff
@@ -238,6 +238,12 @@
   @MakeImmutable
   public static final byte[] bRADISH_DUMP_HEADER = stringToBytes("RADISH");
 
+  @MakeImmutable
+  public static final byte[] bRADISH_WITHSCORES = stringToBytes("WITHSCORES");
```
DonalEvans, line 242: There is already a byte array constant for `WITHSCORES` further up in this class (line 180), so this one is not necessary.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/netty/StringBytesGlossary.java
```diff
@@ -238,6 +238,12 @@
...
+
+  @MakeImmutable
+  public static final byte[] bRADISH_LIMIT = stringToBytes("LIMIT");
```
DonalEvans, line 245: Could this be renamed "bLIMIT" to match the naming of other byte array constants used in executors, and moved further up in this class to be with the other executor-related constants?

DonalEvans (changes requested):

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZRangeByScoreIntegrationTest.java
```diff
@@ -0,0 +1,269 @@
...
+    double score = 1;
+    map.put("member1", score);
+    map.put("member2", score);
+    map.put("member3", score);
+
+    jedis.zadd(KEY, map);
+
+    // Count 1 <= score <= 1
+    assertThat(jedis.zrangeByScore(KEY, score, score))
+        .containsExactlyInAnyOrderElementsOf(map.keySet());
```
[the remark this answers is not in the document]
ringles: Tweaked; unfortunately the map.keySet() isn't in the order we want, but it's fixable.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZRangeByScoreIntegrationTest.java
ringles: Hope the new one is the kind of thing you have in mind...

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZRangeByScoreIntegrationTest.java
```diff
@@ -0,0 +1,269 @@
...
+  @Test
+  public void shouldReturnZero_givenMinGreaterThanMax() {
+    jedis.zadd(KEY, 1, "member");
+
+    // Count +inf <= score <= -inf
+    assertThat(jedis.zrangeByScore(KEY, "+inf", "-inf")).isEmpty();
+  }
+
+  @Test
+  public void shouldReturnCount_givenRangeIncludingScore() {
```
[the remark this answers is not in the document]
ringles: Yeah, missed one.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZRangeByScoreIntegrationTest.java
```diff
@@ -0,0 +1,269 @@
...
+  @Test
+  public void shouldReturnZero_givenNonExistentKey() {
+    assertThat(jedis.zrangeByScore("fakeKey", "-inf", "inf")).isEmpty();
+  }
+
+  @Test
+  public void shouldReturnZero_givenMinGreaterThanMax() {
+    jedis.zadd(KEY, 1, "member");
+
+    // Count +inf <= score <= -inf
```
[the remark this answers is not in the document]
ringles: Yeah, thought I got all those. Quick fix.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/RedisCommandType.java
```diff
@@ -213,10 +214,10 @@
   ZINCRBY(new ZIncrByExecutor(), SUPPORTED, new ExactParameterRequirements(4)),
   ZRANGE(new ZRangeExecutor(), SUPPORTED, new MinimumParameterRequirements(4)
       .and(new MaximumParameterRequirements(5, ERROR_SYNTAX))),
+  ZRANGEBYSCORE(new ZRangeByScoreExecutor(), SUPPORTED, new MinimumParameterRequirements(4)),
   ZRANK(new ZRankExecutor(), SUPPORTED, new ExactParameterRequirements(3)),
   ZREM(new ZRemExecutor(), SUPPORTED, new MinimumParameterRequirements(3)),
-  ZREVRANGE(new ZRevRangeExecutor(), SUPPORTED, new MinimumParameterRequirements(4)
-      .and(new MaximumParameterRequirements(5, ERROR_SYNTAX))),
```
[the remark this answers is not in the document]
ringles: Fixed.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
ringles: Short answer: Yes.
Long answer:  Yes, and we can eliminate another variable, too.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/ZRangeByScoreExecutor.java
ringles: Tests and fixes added for these.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/netty/StringBytesGlossary.java
ringles: Good catch, fixed!

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/netty/StringBytesGlossary.java
ringles: Also corrected!

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZRangeByScoreIntegrationTest.java
```diff
@@ -0,0 +1,269 @@
...
+
+    map.put("member1", Double.NEGATIVE_INFINITY);
+    map.put("member2", 1.0);
+    map.put("member3", Double.POSITIVE_INFINITY);
+
+    jedis.zadd(KEY, map);
+
+    // Count -inf < score <= +inf
+    assertThat(jedis.zrangeByScore(KEY, "(-inf", "+inf"))
+        .containsExactlyElementsOf(Arrays.asList("member2", "member3"));
```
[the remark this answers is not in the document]
ringles: Got a few of them earlier, all done now!

DonalEvans (approved):

jdeppe-pivotal (approved):
Nice work on this!

PR #6701, opened 2021-07-15
GEODE-9429: Allow Radish HSCAN to handle all valid values for COUNT

DonalEvans:
This PR is intentionally split into two commits; first an initial test clean-up and refactor, and then the commit containing the changes related to the ticket.

 - Refactor and clean up AbstractHScanIntegrationTest
 - Refactor RedisHash.hscan() to handle large COUNT values
 - Add tests to cover new behaviour

### For all changes:

### Note:

### DIFF
--- a/geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/hash/AbstractHScanIntegrationTest.java
+++ b/geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/hash/AbstractHScanIntegrationTest.java
@@ -55,6 +55,7 @@ public abstract class AbstractHScanIntegrationTest implements RedisIntegrationTe
   protected JedisCluster jedis;
 
   public static final String HASH_KEY = "key";
+  public static final int SLOT_FOR_KEY = CRC16.calculate(HASH_KEY) % RegionProvider.REDIS_SLOTS;
   public static final String ZERO_CURSOR = "0";
 
   public static final String FIELD_ONE = "1";
@@ -467,9 +468,8 @@ public abstract class AbstractHScanIntegrationTest implements RedisIntegrationTe
     jedis.hset(HASH_KEY, initialHashData);
     final int iterationCount = 500;
 
-    int slot = getSlotForKey();
-    Jedis jedis1 = jedis.getConnectionFromSlot(slot);
-    Jedis jedis2 = jedis.getConnectionFromSlot(slot);
+    Jedis jedis1 = jedis.getConnectionFromSlot(SLOT_FOR_KEY);
+    Jedis jedis2 = jedis.getConnectionFromSlot(SLOT_FOR_KEY);
 
     new ConcurrentLoopingThreads(iterationCount,
         (i) -> multipleHScanAndAssertOnSizeOfResultSet(jedis1, initialHashData),
@@ -489,9 +489,8 @@ public abstract class AbstractHScanIntegrationTest implements RedisIntegrationTe
     jedis.hset(HASH_KEY, initialHashData);
     final int iterationCount = 500;
 
-    int slot = getSlotForKey();
-    Jedis jedis1 = jedis.getConnectionFromSlot(slot);
-    Jedis jedis2 = jedis.getConnectionFromSlot(slot);
+    Jedis jedis1 = jedis.getConnectionFromSlot(SLOT_FOR_KEY);
+    Jedis jedis2 = jedis.getConnectionFromSlot(SLOT_FOR_KEY);
 
     new ConcurrentLoopingThreads(iterationCount,
         (i) -> multipleHScanAndAssertOnContentOfResultSet(i, jedis1, initialHashData),
@@ -512,9 +511,8 @@ public abstract class AbstractHScanIntegrationTest implements RedisIntegrationTe
     jedis.hset(HASH_KEY, initialHashData);
     final int iterationCount = 500;
 
-    int slot = getSlotForKey();
-    Jedis jedis1 = jedis.getConnectionFromSlot(slot);
-    Jedis jedis2 = jedis.getConnectionFromSlot(slot);
+    Jedis jedis1 = jedis.getConnectionFromSlot(SLOT_FOR_KEY);
+    Jedis jedis2 = jedis.getConnectionFromSlot(SLOT_FOR_KEY);
 
     new ConcurrentLoopingThreads(iterationCount,
         (i) -> multipleHScanAndAssertOnContentOfResultSet(i, jedis1, initialHashData),
@@ -593,11 +591,6 @@ public abstract class AbstractHScanIntegrationTest implements RedisIntegrationTe
       dataSet.put(BASE_FIELD + i, "value_" + i);
     }
     return dataSet;
-  }
-
-  private int getSlotForKey() {
-    int crc = CRC16.calculate(HASH_KEY);
-    return crc % RegionProvider.REDIS_SLOTS;
   }
 
   private static class MapEntryWithByteArraysComparator
--- a/geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/hash/AbstractHScanIntegrationTest.java
+++ b/geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/hash/AbstractHScanIntegrationTest.java
@@ -426,6 +426,39 @@ public abstract class AbstractHScanIntegrationTest implements RedisIntegrationTe
         .isSubsetOf(entryMap.entrySet());
   }
 
+  @Test
+  public void should_notErrorGivenCountEqualToIntegerMaxValue() {
+    Map<byte[], byte[]> entryMap = initializeThreeFieldHashBytes();
+
+    ScanParams scanParams = new ScanParams().count(Integer.MAX_VALUE);
+    scanParams.count(Integer.MAX_VALUE);
+
+    ScanResult<Map.Entry<byte[], byte[]>> result =
+        jedis.hscan(HASH_KEY.getBytes(), ZERO_CURSOR.getBytes(), scanParams);
+    assertThat(result.getResult())
+        .usingElementComparator(new MapEntryWithByteArraysComparator())
+        .containsExactlyInAnyOrderElementsOf(entryMap.entrySet());
+  }
+
+  @Test
+  public void should_notErrorGivenCountGreaterThanIntegerMaxValue() {
+    initializeThreeFieldHash();
+
+    String greaterThanInt = String.valueOf(2L * Integer.MAX_VALUE);
+    List<Object> result =
+        uncheckedCast(jedis.sendCommand(HASH_KEY.getBytes(), Protocol.Command.HSCAN,
+            HASH_KEY.getBytes(), ZERO_CURSOR.getBytes(),
+            "COUNT".getBytes(), greaterThanInt.getBytes()));
+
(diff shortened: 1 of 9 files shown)


geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/hash/AbstractHScanIntegrationTest.java
```diff
@@ -577,16 +569,44 @@ private void multipleHScanAndAssertOnSizeOfResultSet(Jedis jedis,
...
-    for (int i = 0; i < sizeOfDataSet; i++) {
+    for (int i = 0; i < SIZE_OF_ENTRY_MAP; i++) {
       dataSet.put(BASE_FIELD + i, "value_" + i);
     }
     return dataSet;
   }
 
-  private int getSlotForKey(String key) {
-    int crc = CRC16.calculate(key);
+  private int getSlotForKey() {
```
jdeppe-pivotal, line 598: The slot could just be a static field now.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisHash.java
```diff
@@ -237,35 +237,32 @@ public int hstrlen(byte[] field) {
...
 
-  private void addIfMatching(Pattern matchPattern, List<byte[]> resultList, byte[] key,
-      byte[] value) {
+  private void addIfMatching(Pattern matchPattern, List<ImmutablePair<byte[], byte[]>> resultList,
+      byte[] key, byte[] value) {
     if (matchPattern != null) {
       if (matchPattern.matcher(bytesToString(key)).matches()) {
-        resultList.add(key);
-        resultList.add(value);
+        resultList.add(new ImmutablePair<>(key, value));
```
jdeppe-pivotal, line 257: Conceptually I like pairing the key and value here, but it is an additional memory allocation that doesn't add any functional value AFAICT.

jdeppe-pivotal (changes requested):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisHash.java
DonalEvans: It's necessary to wrap the key and value in some object here, since the maximum length of a list returned by `HSCAN` is 2*(number of keys) which could in theory be greater than `Integer.MAX_VALUE`. If we want to assume that a user will never have that many keys in their Redish hash, then we can dispense with this pairing of values to increase efficiency and document that the maximum number of Redis hash keys we support is `Integer.MAX_VALUE / 2`, but if we want to be safe, then this is necessary.

jdeppe-pivotal: Ah, OK I see. Out of curiosity I did some back-of-the-napkin tests and it seems like we'd need a VM with at least 80-100GB of heap to even hold enough data to hit this max (in the 'incorrect' implementation). And twice that for the 'correct' implementation. I'm really torn, but I do feel that it's more pragmatic to err on the side of performance than cater to a use case that seems highly unlikely (famous last words :) ). To think about it a different way, we'd always end up paying the cost of allocation just to avoid the incredibly unlikely error situation. (For which there is a solution - just request fewer than 2 billion keys at a time).

Not to overly complicate this, but technically it shouldn't be difficult to optimise for the most frequent use-case but have a special case for when the requested count does exceed `Integer.MAX_VALUE / 2`. I'm not suggesting you do that now though.

DonalEvans: Yeah, I think I tend to agree with you on this one. I suspect that the vast majority of users would rather their cache be faster than that it be able to handle an improbably large number of keys/values. I'll rework this part of the code and see about maybe introducing some error logging if we hit that unlikely scenario.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisHash.java
```diff
@@ -237,15 +238,21 @@ public int hstrlen(byte[] field) {
...
   }
 
-  public ImmutablePair<Integer, List<byte[]>> hscan(Pattern matchPattern,
-      int count,
-      int cursor) {
-
-    ArrayList<byte[]> resultList = new ArrayList<>(count + 2);
+  public ImmutablePair<Integer, List<byte[]>> hscan(Pattern matchPattern, int count, int cursor) {
+    // No need to allocate more space than it's possible to use given the size of the hash. We need
+    // to add 1 to hash.size() to ensure that if count > hash.size(), we return a cursor of 0
```
dschneider-pivotal, line 243: change hash.size() to hlen()

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisHash.java
```diff
@@ -237,15 +238,21 @@ public int hstrlen(byte[] field) {
...
+    long maximumCapacity = 2L * Math.min(count, hlen() + 1);
+    if (maximumCapacity > Integer.MAX_VALUE) {
+      LogService.getLogger().error(
+          "The size of the data to be returned by hscan, {}, exceeds the maximum capacity of an array",
+          maximumCapacity);
+      throw new OutOfMemoryError("Requested array size exceeds VM limit");
```
dschneider-pivotal, line 249: I think OutOfMemoryError should be reserved for when you actually run out of heap. In this case you are preventing that from happening by not doing a large allocation. Some code will no longer trust the state of the jvm if it sees an OutOfMemoryError and will initiate a shutdown. You could use IllegalArgumentException instead. Whatever exception is thrown here should it be caught in HScanExecutor and converted to RedisResponse.error?

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisHash.java
```diff
@@ -237,15 +238,21 @@ public int hstrlen(byte[] field) {
...
-      int count,
-      int cursor) {
-
-    ArrayList<byte[]> resultList = new ArrayList<>(count + 2);
+  public ImmutablePair<Integer, List<byte[]>> hscan(Pattern matchPattern, int count, int cursor) {
+    // No need to allocate more space than it's possible to use given the size of the hash. We need
+    // to add 1 to hash.size() to ensure that if count > hash.size(), we return a cursor of 0
+    long maximumCapacity = 2L * Math.min(count, hlen() + 1);
+    if (maximumCapacity > Integer.MAX_VALUE) {
+      LogService.getLogger().error(
```
dschneider-pivotal, line 246: I'm not sure what we should do about logging in these cases. Normally we do not log something in a server when a client does something wrong. We just have the server respond to the client with an error message. I know in some cases we are limited by the redis standard error responses in giving the user extra information about what they did wrong. In that case we could add some info in the server log but maybe it would be at the "info" log level. The "error" level is usually reserved for when something unexpected happened on the server but it thinks it can keep running. The "warning" level is if things don't look healthy (but it might be okay) and we want to get the admin's attention. The "fatal" level is if something is wrong and the server needs to shutdown.
In this case I would not log anything here if we can tell the client what was wrong. If we can't get a response back to the client then I think this should be "info" and describe what the client did wrong.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisHash.java
```diff
@@ -237,15 +238,21 @@ public int hstrlen(byte[] field) {
...
-  public ImmutablePair<Integer, List<byte[]>> hscan(Pattern matchPattern,
-      int count,
-      int cursor) {
-
-    ArrayList<byte[]> resultList = new ArrayList<>(count + 2);
+  public ImmutablePair<Integer, List<byte[]>> hscan(Pattern matchPattern, int count, int cursor) {
+    // No need to allocate more space than it's possible to use given the size of the hash. We need
+    // to add 1 to hash.size() to ensure that if count > hash.size(), we return a cursor of 0
+    long maximumCapacity = 2L * Math.min(count, hlen() + 1);
+    if (maximumCapacity > Integer.MAX_VALUE) {
```
dschneider-pivotal, line 245: Where you able to test this in the integration test? I know you did in the unit test but it is hard to do when you need  the hash to have more than 1G entries. If you could use the same hlen() trick in the integration test that you used in the unit test then you could test how this exception will be handled by the higher levels of the code.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisHash.java
DonalEvans: Nice catch, thanks

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisHash.java
DonalEvans: The integration tests are designed to simulate the behaviour of a client connecting to a Redis cluster, and to be agnostic about whether that cluster is a Geode Radish cluster or a native Redis cluster, so it's not really possible to get under the hood and mock out the returns of methods in our classes to force certain behaviours, unfortunately.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisHash.java
DonalEvans: Currently, the exception thrown here will ultimately get handled in `ExecutionHandlerContext.getExceptionResponse()` where it will result in an error message returned to the client with the message "The server had an internal error please try again". This definitely doesn't seem correct, since trying again will just result in the same error, but none of the currently existing error responses we have seem right either, and I'm having a hard time find a list of existing Redis error messages to see if any of them would be a better fit. It's also not practical to test what Redis actually does in this situation, given that would require a Redis hash with over a billion entries. I'll have a look at the Redis source and see if I can figure out what the behaviour would be in this situation.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisHash.java
DonalEvans: I've changed the log level to info and made the message more descriptive/helpful.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisHash.java
dschneider-pivotal: How about ERROR_NOT_INTEGER? In this case the size is "out of range". Also be sure not to use OutOfMemoryError.

DonalEvans: Sounds good to me. I checked Redis' code and they don't hit this issue, since they're able to allocate lists with a maximum length of Long.MAX_VALUE, which is greater than their maximum number of entries in a hash, so this is one case where we're not able to exactly match their behaviour. Thankfully, it's also a case that should never realistically come up, since it's highly unlikely that users will be putting over a billion entries into one Redis hash.

jdeppe-pivotal (approved):

PR #6707, opened 2021-07-20
GEODE-9439: Convert CertificateBuilder to sun/jdk APIs

jdeppe-pivotal:
- Makes it easier to de/serialize created certificates
- Remove the requirement for the Bouncy Castle library
- Java 9 has a very similar utility which we should be able
  to switch to once we move away from 8.

### For all changes:

### Note:

### DIFF
--- a/geode-junit/build.gradle
+++ b/geode-junit/build.gradle
@@ -19,6 +19,10 @@ apply from: "${rootDir}/${scriptDir}/standard-subproject-configuration.gradle"
 
 apply from: "${project.projectDir}/../gradle/publish-java.gradle"
 
+compileJava {
+  options.compilerArgs << '-Xlint:-sunapi'
+  options.compilerArgs << '-XDenableSunApiLintControl'
+}
 
 dependencies {
   api(platform(project(':boms:geode-all-bom')))
--- a/geode-junit/build.gradle
+++ b/geode-junit/build.gradle
@@ -51,7 +51,6 @@ dependencies {
   api('org.apache.logging.log4j:log4j-api')
   
   api('org.awaitility:awaitility')
-  api('org.bouncycastle:bcpkix-jdk15on')
   api('org.hamcrest:hamcrest')
   api('io.micrometer:micrometer-core')
   
--- a/geode-junit/src/main/java/org/apache/geode/cache/ssl/CertificateBuilder.java
+++ b/geode-junit/src/main/java/org/apache/geode/cache/ssl/CertificateBuilder.java
@@ -192,22 +192,22 @@ public class CertificateBuilder {
 
       byte[] keyIdBytes = new KeyIdentifier(publicKey).getIdentifier();
       SubjectKeyIdentifierExtension keyIdentifier = new SubjectKeyIdentifierExtension(keyIdBytes);
-      extensions.set(keyIdentifier.getExtensionId().toString(), keyIdentifier);
+      extensions.set(SubjectKeyIdentifierExtension.NAME, keyIdentifier);
 
       GeneralNames subjectAltNames = san();
       if (!subjectAltNames.isEmpty()) {
         SubjectAlternativeNameExtension altNames =
             new SubjectAlternativeNameExtension(subjectAltNames);
-        extensions.set(altNames.getExtensionId().toString(), altNames);
+        extensions.set(SubjectAlternativeNameExtension.NAME, altNames);
       }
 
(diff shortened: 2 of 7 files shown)


Bill:
🥇

onichols-pivotal (approved):

mhansonp:
I don't remember seeing this as a proprosal in the dev channel. Is this something that should be there? I am just asking the question... I am curious if this affects Spring or others.

jdeppe-pivotal:
It's a change to a test helper class - I didn't think it would warrant any significant discussion.

geode-junit/build.gradle
```diff
@@ -19,6 +19,10 @@ apply from: "${rootDir}/${scriptDir}/standard-subproject-configuration.gradle"
 
 apply from: "${project.projectDir}/../gradle/publish-java.gradle"
 
+compileJava {
+  options.compilerArgs << '-Xlint:-sunapi'
+  options.compilerArgs << '-XDenableSunApiLintControl'
```
Bill, line 24: So we've moved from Bouncy Castle, a high quality open source project, to an internal JDK class. These compiler options suppress what would be compile errors I suppose (or at least warnings) that we are illegally accessing JDK internal classes.

My research shows that in JDK 9 these classes in a `sun` package will move to a `javax` one. I suppose when that happens the JDK8 internal classes accessed here will have direct replacements available so all we'll have to do is change the includes in these files.

geode-junit/src/main/java/org/apache/geode/cache/ssl/CertificateBuilder.java
```diff
@@ -79,16 +77,27 @@ public CertificateBuilder(int days, String algorithm) {
   }
 
   private static GeneralName dnsGeneralName(String name) {
-    return new GeneralName(GeneralName.dNSName, name);
+    try {
+      return new GeneralName(new DNSName(name));
+    } catch (IOException ex) {
+      throw new UncheckedIOException(ex);
+    }
```
Bill, line 84: The value of bundling this as into a runtime exception is that this method can still be used in contexts (e.g. Stream.map(f)`) where a method throwing a checked exception would not be allowed. Ok.

geode-junit/src/main/java/org/apache/geode/cache/ssl/CertificateBuilder.java
```diff
@@ -146,58 +164,69 @@ public CertificateMaterial generate() {
...
+
+      // Update the algorithm, and resign.
+      algo = (AlgorithmId) cert.get(X509CertImpl.SIG_ALG);
+      info.set(CertificateAlgorithmId.NAME + "." + CertificateAlgorithmId.ALGORITHM, algo);
+      cert = new X509CertImpl(info);
+      cert.sign(privateKey, algorithm);
+
+      return cert;
+    } catch (Exception ex) {
+      throw new RuntimeException("Unable to create certificate", ex);
```
Bill, line 229: I'm counting on our test suite to catch any regressions in this method.

Bill (approved):
Nice one @jdeppe-pivotal !

ladyVader (changes requested):
I think you will still need to make the suggested changes to CertificateMaterial (Optional field is not serializable).  See lines marked with =>

public class CertificateMaterial implements java.io.Serializable {
  private final X509Certificate certificate;
  private final KeyPair keyPair;
=>  private final X509Certificate issuer;
  public CertificateMaterial(X509Certificate certificate, KeyPair keyPair, X509Certificate issuer) {
    this.certificate = certificate;
    this.keyPair = keyPair;
=>    this.issuer = issuer;
  }
  public X509Certificate getCertificate() {
    return certificate;
  }
  public PublicKey getPublicKey() {
    return keyPair.getPublic();
  }
  public PrivateKey getPrivateKey() {
    return keyPair.getPrivate();
  }
  public Optional<X509Certificate> getIssuer() {
=>    return Optional.ofNullable(issuer);
  }
}

mhansonp (approved):

mhansonp:
> It's a change to a test helper class - I didn't think it would warrant any significant discussion.

Sounds good. Approved..

jdeppe-pivotal:
@ladyVader Oops - thanks for pointing that out. Just pushed that change.

ladyVader (approved):

dickcav (approved):

PR #6708, opened 2021-07-20
GEODE-9441: Reduce Nested func test thread count

demery-pivotal:
Change `NestedFunctionExecutionDistributedTest` to create fewer threads.

The test creates a number of threads that is a multiple of
`MAX_FE_THREADS`. On a CI machine, the default value of `MAX_FE_THREADS`
can be very high, causing this test to create thousands of threads. This
can leave the system unable to create threads for other processes, which
in turn can cause Gradle processes or other tests to fail or hang.

This test now sets `MAX_FE_THREADS` to a lower value (100), which
suffices for the purpose of the test, and results in greatly reduced
consumption of threads.

Co-authored-by: Dale Emery <demery@vmware.com>

### DIFF
--- a/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/execute/NestedFunctionExecutionDistributedTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/execute/NestedFunctionExecutionDistributedTest.java
@@ -55,7 +55,11 @@ import org.apache.geode.test.junit.rules.serializable.SerializableTestName;
 @Category(FunctionServiceTest.class)
 @SuppressWarnings("serial")
 public class NestedFunctionExecutionDistributedTest implements Serializable {
-
+  // Limit the number of FE threads for this test. On a CI machine, the default value can be very
+  // large, which can cause this test to overwhelm the system and prevent other tests and processes
+  // from creating threads.
+  private static final String TEST_MAX_FE_THREADS = "100";
+  private static final String MAX_FE_THREADS_PROPERTY = "DistributionManager.MAX_FE_THREADS";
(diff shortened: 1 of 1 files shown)


geode-core/src/distributedTest/java/org/apache/geode/internal/cache/execute/NestedFunctionExecutionDistributedTest.java
```diff
@@ -111,6 +116,7 @@ private int createServerCache() throws IOException {
 
   private void createClientCache(int port, int functionTimeout) {
     System.setProperty(GEMFIRE_PREFIX + "CLIENT_FUNCTION_TIMEOUT", String.valueOf(functionTimeout));
+    System.setProperty(MAX_FE_THREADS_PROPERTY, TEST_MAX_FE_THREADS);
```
boglesby, line 119: Its not necessary to set this value in the client. The Function Execution thread pool is not created there.

demery-pivotal: Currently the client does `MAX_FE_THREADS * 2` executions of `ParentFunction`. If I don't set `MAX_FE_THREADS` in the client, I get a swarm of failures with that many functions. Should I directly reduce the number of functions it executes, rather than computing from `MAX_FE_THREADS`?

boglesby: Sorry, you're right. You should set it.

boglesby (approved):

PR #6714, opened 2021-07-22
GEODE-9396: Upgrades using SSL fail with mismatch of hostname in certificates

kamilla1201:
The fix for GEODE-9139 has changed the way hostnames are handled when SNI endpoint validation is enabled. Because of that, when upgrading from some older versions of Geode to a version that contains the fix for GEODE-9139, the SSL handshake may fail to complete. This happens in cases where the call of `MemberData.getHostName()` method on an older member returns an IP address. In the fix for GEODE-9139, `Connection.createIoFilter()` uses the `hostname` for endpoint identification, and if the certificate contains a symbolic name, endpoint identification will fail with "java.security.cert.CertificateException: No subject alternative names matching IP address" error.

This change adds `fromDataPre_GEODE_1_15_0_0` method to `MemberIdentifierImpl` that converts the `hostname` of an older member to a hostname if it contains an IP address. This change also adds `toDataPre_GEODE_1_15_0_0` but it doesn't convert the hostname back to an IP address. We think there is no need to do that because the member is about to be upgraded and will have a hostname in `hostname` field anyway. 

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/distributed/internal/membership/InternalDistributedMember.java
+++ b/geode-core/src/main/java/org/apache/geode/distributed/internal/membership/InternalDistributedMember.java
@@ -456,16 +456,16 @@ public class InternalDistributedMember
   }
 
   @Override
-  public void toDataPre_GEODE_1_15_0_0(DataOutput out, SerializationContext context)
-      throws IOException {
-    memberIdentifier.toDataPre_GEODE_1_15_0_0(out, context);
-  }
-
-  @Override
   public void toData(DataOutput out,
       SerializationContext context)
       throws IOException {
     memberIdentifier.toData(out, context);
+  }
+
+  @Override
+  public void toDataPre_GEODE_1_15_0_0(DataOutput out, SerializationContext context)
+      throws IOException {
+    memberIdentifier.toDataPre_GEODE_1_15_0_0(out, context);
   }
 
   public void toDataPre_GFE_9_0_0_0(DataOutput out,
@@ -475,17 +475,17 @@ public class InternalDistributedMember
   }
 
   @Override
-  public void fromDataPre_GEODE_1_15_0_0(DataInput in, DeserializationContext context)
-      throws IOException, ClassNotFoundException {
-    memberIdentifier.fromDataPre_GEODE_1_15_0_0(in, context);
-    durableClientAttributes = null;
-  }
-
-  @Override
   public void fromData(DataInput in,
       DeserializationContext context)
       throws IOException, ClassNotFoundException {
     memberIdentifier.fromData(in, context);
+    durableClientAttributes = null;
+  }
+
+  @Override
+  public void fromDataPre_GEODE_1_15_0_0(DataInput in, DeserializationContext context)
+      throws IOException, ClassNotFoundException {
+    memberIdentifier.fromDataPre_GEODE_1_15_0_0(in, context);
     durableClientAttributes = null;
   }
 
--- a/geode-core/src/main/java/org/apache/geode/distributed/internal/membership/InternalDistributedMember.java
+++ b/geode-core/src/main/java/org/apache/geode/distributed/internal/membership/InternalDistributedMember.java
@@ -456,6 +456,12 @@ public class InternalDistributedMember
   }
 
   @Override
+  public void toDataPre_GEODE_1_15_0_0(DataOutput out, SerializationContext context)
+      throws IOException {
+    memberIdentifier.toDataPre_GEODE_1_15_0_0(out, context);
+  }
+
+  @Override
   public void toData(DataOutput out,
       SerializationContext context)
       throws IOException {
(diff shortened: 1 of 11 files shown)


mhansonp (approved):

echobravopapa (approved):

jinmeiliao (approved):

geode-core/src/main/java/org/apache/geode/distributed/internal/membership/InternalDistributedMember.java
```diff
@@ -455,6 +455,12 @@ public void readExternal(ObjectInput in) throws IOException, ClassNotFoundExcept
     durableClientAttributes = null;
   }
 
+  @Override
+  public void toDataPre_GEODE_1_15_0_0(DataOutput out, SerializationContext context)
+      throws IOException {
+    memberIdentifier.toDataPre_GEODE_1_15_0_0(out, context);
+  }
+
```
Bill, line 463: sorry to be nitpicky but would you mind putting `toData()` before each of the version-specific to-data methods. And then order the to-data methods from newest to oldest. I believe that's the pattern established in other DSFIDs.

Ordering them consistently helps maintainers.

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/MemberIdentifierImpl.java
```diff
@@ -693,10 +701,43 @@ public void fromData(DataInput in,
...
+    readMemberData(in);
+    convertIpToHostnameIfNeeded();
+  }
+
+  /**
+   * In older versions of Geode {@link MemberData#getHostName()} can return an IP address
+   * if network partition is enabled. Also, older versions of Geode don't use bind-address for
+   * SNI endpoint identification and do a reverse lookup to find the fully qualified hostname.
+   * This can become a problem if TLS certificate doesn't have the fully qualified name in it
+   * as a Subject Alternate Name, therefore this behavior was changed to preserve the bind-address.
```
Bill, line 725: Second sentence confuses me. What do you think of this:

In older versions of Geode {@link MemberData#getHostName()} can return an IP address if network partition detection is enabled. If SSL endpoint identification is enabled, those product versions supply the result of a reverse lookup to the TLS handshake API. Endpoint identification will fail if e.g. the lookup returned a fully-qualified name but the certificate had just a (non-fully-qualified) hostname in a Subject Alternate Name field.

In version 1.15.0 member identifiers were changed so that if a bind address is specified, that exact string will be carried as the host name. That gives the administrator better control over endpoint identification. When upgrading from earlier versions we convert any IP numbers to hostnames via reverse lookup here.

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/MemberIdentifierImpl.java
```diff
@@ -693,10 +701,43 @@ public void fromData(DataInput in,
...
+   *
+   * During a rolling upgrade, we need to ensure that {@link MemberData#getHostName()} returns
+   * a hostname because it might be later used for SNI endpoint identification.
+   */
+  private void convertIpToHostnameIfNeeded() {
+    if (memberData.isNetworkPartitionDetectionEnabled()) {
```
Bill, line 731: should we be guarding also on (SSL endpoint identification is enabled)?

Bill (changes requested):
Minor request for reordering a couple methods. Asking for clarification in some comments. And wondering whether we should add another condition in one place.

Bill (approved):
Great work on this complicated ticket @kamilla1201!

upthewaterspout (commented):
I'm wondering if someone would have a reason to specify an IP address as the bind address that this might break. Cases I could think about:
* Maybe there is no DNS name? The call to `getHost()` might fail
* Maybe the DNS server isn't updated at the time the member starts. This might be the case in a IaaS where VMs or containers are launched as soon as the gemfire process starts. I guess for this case an old member would be less likely to not be resolvable?

I'm not sure if this should reject this approach or not, if there are no better options. Or maybe we just failures from getHost and retain the IP address.

upthewaterspout:
> I'm wondering if someone would have a reason to specify an IP address as the bind address that this might break. Cases I could think about:
> 
>     * Maybe there is no DNS name? The call to `getHost()` might fail
> 
>     * Maybe the DNS server isn't updated at the time the member starts. This might be the case in a IaaS where VMs or containers are launched as soon as the gemfire process starts. I guess for this case an old member would be less likely to not be resolvable?
> 
> 
> I'm not sure if this should reject this approach or not, if there are no better options. Or maybe we just failures from getHost and retain the IP address.

I wonder if it would be possible to fall back to the old logic for hostname validation if the member is from an old member, rather than changing the bind address when deserializing. Would that be doable or any safer?

dschneider-pivotal (approved):

kamilla1201:
Closing this PR because we decided to implement the fix that Dan suggested.

PR #6715, opened 2021-07-22
GEODE-9379: Implement ZREVRANGEBYSCORE

ringles:
Add support for the ZREVRANGEBYSCORE Redis comand.


geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -310,18 +311,62 @@ long zcount(SortedSetRangeOptions rangeOptions) {
...
+    int count = Integer.MAX_VALUE;
+    if (rangeOptions.hasLimit()) {
+      count = rangeOptions.getCount();
+      maxIndex -= rangeOptions.getOffset();
+      if (maxIndex < 0) {
+        return Collections.emptyList();
+      }
+    }
+    Iterator<AbstractOrderedSetEntry> entryIterator =
+        scoreSet.getIndexRange(maxIndex - 1, Math.min(count, maxIndex - minIndex), true);
```
dschneider-pivotal, line 342: It seems like the values passed to getIndexRange here could also be used to compute a good initialize size of the ArrayList result. You could allocate the result here and do it with this better size which will prevent extra copying of the result ArrayList as it grows as we add to it in the loop. Remember to make its initial size *2 if withScores. I think Donal has this same logic in the bylex method.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZRangeByScoreExecutor.java
```diff
@@ -0,0 +1,119 @@
...
+        && commandElements.size() > commandIndex + 2) {
+      offset = narrowLongToInt(bytesToLong(commandElements.get(commandIndex + 1)));
+      count = narrowLongToInt(bytesToLong(commandElements.get(commandIndex + 2)));
+      if (count < 0) {
+        count = Integer.MAX_VALUE;
+      }
+    } else {
+      throw new IllegalArgumentException();
+    }
+    rangeOptions.setLimitValues(offset, count);
```
dschneider-pivotal, line 115: shouldn't we only call setLimitValues if we found the bLIMIT keyword? I guess it works that way (since if we don't find the keyword we throw an exception).  I think the code would be a little cleaner if you declared offset and count inside the if block as you initialize them and made this call at the end of the if block

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZRevRangeByScoreIntegrationTest.java
```diff
@@ -0,0 +1,342 @@
...
+import redis.clients.jedis.Tuple;
+
+import org.apache.geode.redis.RedisIntegrationTest;
+
+@RunWith(JUnitParamsRunner.class)
+public abstract class AbstractZRevRangeByScoreIntegrationTest implements RedisIntegrationTest {
+  private static final String MEMBER_BASE_NAME = "member";
+  private static final String KEY = "key";
+  private JedisCluster jedis;
+  private static final List<Double> scores =
```
DonalEvans, line 52: This field is never used and can be removed.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZRevRangeByScoreIntegrationTest.java
```diff
@@ -0,0 +1,342 @@
...
+
+    assertThat(jedis.zrevrangeByScore(KEY, "(3.0", "(1.0"))
+        .containsExactly("member2");
+    assertThat(jedis.zrevrangeByScore(KEY, "(3.0", "1.0"))
+        .containsExactly("member2", "member1");
+    assertThat(jedis.zrevrangeByScore(KEY, "3.0", "(1.0"))
+        .containsExactly("member3", "member2");
+  }
+
+  private Map<String, Double> getExclusiveTestMap() {
```
DonalEvans, line 192: Could this private method be moved to the bottom of the class? Just personal preference, but it makes things neater, I think.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZRevRangeByScoreIntegrationTest.java
```diff
@@ -0,0 +1,342 @@
...
+    jedis.zadd(KEY, map);
+
+    // Range inf >= score > 0
+    assertThat(jedis.zrevrangeByScore(KEY, "inf", "(")).containsExactly("slightlyMoreThanZero");
+
+    // Range 0 >= score > -inf
+    assertThat(jedis.zrevrangeByScore(KEY, "(", "-inf")).containsExactly("slightlyLessThanZero");
+  }
+
+  private void createZSetRangeTestMap() {
```
DonalEvans, line 239: Could this private method be moved to the bottom of the class too please?

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZRevRangeByScoreIntegrationTest.java
```diff
@@ -0,0 +1,342 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import redis.clients.jedis.HostAndPort;
+import redis.clients.jedis.JedisCluster;
+import redis.clients.jedis.Protocol;
+import redis.clients.jedis.Tuple;
+
+import org.apache.geode.redis.RedisIntegrationTest;
+
+@RunWith(JUnitParamsRunner.class)
```
DonalEvans, line 47: This test doesn't seem to use the `@Parameters` annotation anywhere, so this `@RunWith` is unnecessary. However I think there are probably a few tests that might benefit from parameterization, so maybe it should stay.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZRevRangeByScoreIntegrationTest.java
```diff
@@ -0,0 +1,342 @@
...
+    jedis.close();
+  }
+
+  @Test
+  public void shouldError_givenWrongNumberOfArguments() {
+    assertAtLeastNArgs(jedis, Protocol.Command.ZREVRANGEBYSCORE, 3);
+  }
+
+  @Test
+  public void shouldError_givenInvalidMinOrMax() {
```
DonalEvans, line 72: This test could be simplified by using the `@Parameters` annotation to pass all of the arguments, since the behaviour for each is always the same.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZRevRangeByScoreIntegrationTest.java
```diff
@@ -0,0 +1,342 @@
...
+  public void shouldReturnRange_boundedByLimit() {
+    createZSetRangeTestMap();
+
+    assertThat(jedis.zrevrangeByScore(KEY, "10", "0", 0, 2))
+        .containsExactly("f", "e");
+    assertThat(jedis.zrevrangeByScore(KEY, "10", "0", 2, 3))
+        .containsExactly("d", "c", "b");
+    assertThat(jedis.zrevrangeByScore(KEY, "10", "0", 2, 10))
+        .containsExactly("d", "c", "b");
+    assertThat(jedis.zrevrangeByScore(KEY, "10", "0", 20, 10)).isEmpty();
```
DonalEvans, line 263: This case should probably be broken out into its own test, since it's more of a special case (offset is greater than the size of the range).

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZRevRangeByScoreIntegrationTest.java
```diff
@@ -0,0 +1,342 @@
...
+    assertThat(jedis.zrevrangeByScoreWithScores(KEY, "10", "0", 0, 0))
+        .isEmpty();
+    assertThat(jedis.zrevrangeByScoreWithScores(KEY, "10", "0", 0, 2))
+        .containsExactlyElementsOf(firstExpected);
+    assertThat(jedis.zrevrangeByScoreWithScores(KEY, "10", "0", 2, 3))
+        .containsExactlyElementsOf(secondExpected);
+    assertThat(jedis.zrevrangeByScoreWithScores(KEY, "10", "0", 2, 10))
+        .containsExactlyElementsOf(secondExpected);
+    assertThat(jedis.zrevrangeByScoreWithScores(KEY, "10", "0", 2, -1))
+        .containsExactlyElementsOf(secondExpected);
```
DonalEvans, line 288: This case should probably be broken out into its own test, since it's a special case (negative count returns all elements in range).

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZRevRangeByScoreIntegrationTest.java
```diff
@@ -0,0 +1,342 @@
...
+    firstExpected.add(new Tuple("f", 5d));
+    firstExpected.add(new Tuple("e", 4d));
+
+    Set<Tuple> secondExpected = new LinkedHashSet<>();
+    secondExpected.add(new Tuple("d", 3d));
+    secondExpected.add(new Tuple("c", 2d));
+    secondExpected.add(new Tuple("b", 1d));
+
+    assertThat(jedis.zrevrangeByScoreWithScores(KEY, "10", "0", 0, 0))
+        .isEmpty();
```
DonalEvans, line 280: This case should probably be broken out into its own test, since it's a special case (count of zero returns an empty list).

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZRevRangeByScoreIntegrationTest.java
```diff
@@ -0,0 +1,342 @@
...
+    assertThat(jedis.zrevrangeByScoreWithScores(KEY, "10", "0", 2, 3))
+        .containsExactlyElementsOf(secondExpected);
+    assertThat(jedis.zrevrangeByScoreWithScores(KEY, "10", "0", 2, 10))
+        .containsExactlyElementsOf(secondExpected);
+    assertThat(jedis.zrevrangeByScoreWithScores(KEY, "10", "0", 2, -1))
+        .containsExactlyElementsOf(secondExpected);
+  }
+
+  @Test
+  public void shouldReturnProperError_givenLimitWithWrongFormat() {
```
DonalEvans, line 292: Could we also have a test case where the `LIMIT` argument is misspelled?

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZRevRangeByScoreIntegrationTest.java
```diff
@@ -0,0 +1,342 @@
...
+            "0"))
+                .hasMessageContaining(ERROR_SYNTAX);
+    assertThatThrownBy(
+        () -> jedis.sendCommand(KEY, Protocol.Command.ZREVRANGEBYSCORE, KEY, "10", "0",
+            "LIMIT", "0", "invalid"))
+                .hasMessageContaining(ERROR_NOT_INTEGER);
+  }
+
+  @Test
+  public void shouldReturnRange_givenMultipleCopiesOfWithscoresAndOrLimit() {
```
DonalEvans, line 309: Could we get a test of the behaviour for improperly-specified `LIMIT` with multiple `LIMIT`s? Like, have two `LIMIT`s specified, and assert on the expected error when the first one is wrong, and when the second one is wrong, with and without `WITHSCORES`.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -310,18 +311,62 @@ long zcount(SortedSetRangeOptions rangeOptions) {
     return getRange(min, max, withScores, false);
   }
 
+  List<byte[]> zrevrangebyscore(SortedSetRangeOptions rangeOptions, boolean withScores) {
```
DonalEvans, line 314: This method should also be overridden in `NullRedisSortedSet` to return an empty list.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -310,18 +311,62 @@ long zcount(SortedSetRangeOptions rangeOptions) {
...
+    List<byte[]> result = new ArrayList<>();
```
DonalEvans, line 315: This assignment should be moved closer to the point at which the variable is used, as it's possible that we return from this method before we need it.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -310,18 +311,62 @@ long zcount(SortedSetRangeOptions rangeOptions) {
...
+      }
+    }
+    Iterator<AbstractOrderedSetEntry> entryIterator =
+        scoreSet.getIndexRange(maxIndex - 1, Math.min(count, maxIndex - minIndex), true);
+
+    while (entryIterator.hasNext()) {
+      AbstractOrderedSetEntry entry = entryIterator.next();
+      if (rangeOptions.isStartExclusive() && entry.score == rangeOptions.getStartDouble()) {
+        continue;
+      }
```
DonalEvans, line 348: This check is not necessary since the range returned by `getIndexRange()` will by design not contain elements that it shouldn't.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZRangeByScoreExecutor.java
```diff
@@ -0,0 +1,119 @@
...
+
+import org.apache.geode.redis.internal.executor.AbstractExecutor;
+import org.apache.geode.redis.internal.executor.RedisResponse;
+import org.apache.geode.redis.internal.netty.Command;
+import org.apache.geode.redis.internal.netty.ExecutionHandlerContext;
+
+public abstract class AbstractZRangeByScoreExecutor extends AbstractExecutor {
+  @Override
+  public RedisResponse executeCommand(Command command, ExecutionHandlerContext context) {
+    RedisSortedSetCommands redisSortedSetCommands = context.getSortedSetCommands();
```
DonalEvans, line 37: This should be moved closer to the place that it's used, since it's possible we return before we need it.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZRangeByScoreExecutor.java
```diff
@@ -0,0 +1,119 @@
...
+
+    List<byte[]> commandElements = command.getProcessedCommand();
+
+    SortedSetRangeOptions rangeOptions;
+    boolean withScores = false;
```
DonalEvans, line 42: This should be moved closer to the place that it's used, since it's possible we return before we need it.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/SortedSetRangeOptions.java
```diff
@@ -20,10 +20,10 @@
...
 
 public class SortedSetRangeOptions {
-  private final double minDouble;
-  private final boolean minExclusive;
-  private final double maxDouble;
-  private final boolean maxExclusive;
+  private final double startDouble;
+  private final boolean startExclusive;
+  private final double endDouble;
+  private final boolean endExclusive;
```
DonalEvans, line 26: Rather than renaming all of these fields and methods, would it be simpler to just pass the appropriate arguments to the constructor in `ZRevRangeByScoreExecutor`? That is, if we know that `commandElements.get(2)` is the max, then use that as the second argument in the constructor rather than the first, using the `isRev()` method to switch the behaviour.

DonalEvans (dismissed):
This PR needs to add a test in `AbstractHitsMissesIntegrationTest` to confirm that stats are correctly updated. Also, I think it would be beneficial to wait for [this PR](https://github.com/apache/geode/pull/6716) to be merged and then rebase, since there's some code added there that would be useful to this PR.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZRevRangeByScoreIntegrationTest.java
ringles: Should be gone. And the HitsMisses test has been added.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
ringles: Yup, caught that earlier today!

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZRevRangeByScoreIntegrationTest.java
ringles: Done

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZRevRangeByScoreIntegrationTest.java
ringles: Done

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZRevRangeByScoreIntegrationTest.java
ringles: But of course, monsieur!

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZRevRangeByScoreIntegrationTest.java
ringles: Also added

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
ringles: Tweaked

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
ringles: Leftover from earlier versions, thanks!

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZRangeByScoreExecutor.java
ringles: Shuffled.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZRangeByScoreExecutor.java
ringles: Also tweaked.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZRevRangeByScoreIntegrationTest.java
ringles: Paramified.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZRevRangeByScoreIntegrationTest.java
ringles: Good catch, used below now.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZRevRangeByScoreIntegrationTest.java
ringles: Clarified

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZRevRangeByScoreIntegrationTest.java
ringles: Tweaked

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
ringles: Done, or at least a stab thereto.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZRevRangeByScoreIntegrationTest.java
ringles: Also done.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZRangeByScoreExecutor.java
ringles: Massaged.

DonalEvans (commented):
Looks good, with the understanding that it'll be rebased and modified to make use of new methods once the ZRANGEBYLEX PR has been merged. I'll give it another look over once that's done for final approval.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
ringles: Alphabetized.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -381,6 +381,55 @@ long zrem(Region<RedisKey, RedisData> region, RedisKey key, List<byte[]> members
...
+    int startIndex = maxIndex - 1;
+    int endIndex = Math.min(count, maxIndex - minIndex);
+    int initialCapacity = startIndex - endIndex;
+    if (withScores) {
+      initialCapacity *= 2;
+    }
+
+    Iterator<AbstractOrderedSetEntry> entryIterator =
+        scoreSet.getIndexRange(startIndex, endIndex, true);
+    List<byte[]> result = new ArrayList<>(initialCapacity > 0 ? initialCapacity : 0);
```
dschneider-pivotal, line 421: if initialCapacity is <= 0 does that mean we will return an empty list? If so should we just have checked (up at line 414) if (initialCapacity <= 0) {return Collections.emptyList();}? Then this could just be new ArrayList<>(initialCapacity)

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/AbstractSortedSetRangeOptions.java
```diff
@@ -66,26 +66,31 @@ void parseLimitArguments(List<byte[]> commandElements, int commandIndex) {
 
   // If limit specified but count is zero, or min > max, or min == max and either are exclusive, the
   // range cannot contain any elements
-  boolean isEmptyRange() {
-    int minVsMax = compareMinToMax();
-    return (hasLimit && (count == 0 || offset < 0)) || minVsMax == 1
-        || (minVsMax == 0 && (isMinExclusive || isMaxExclusive));
+  boolean isEmptyRange(boolean isRev) {
+    int startVsEnd = compareStartToEnd();
```
dschneider-pivotal, line 70: Much is now repeated between the if and then else.
I think this code would be clearer if you didn't try to do everything in one boolean expression. For example:
if (hasLimit && (count == 0 || offset < 0)) {
  return true;
}
int startVsEnd = compareStartToEnd();
if (startVsEnd == 0 && (isStartExclusive || isEndExclusive)) {
  return true;
}
if (isRev) {
  return startVsEnd == -1;
} else {
  return startVsEnd == 1;
}

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -310,18 +311,66 @@ long zcount(SortedSetRangeOptions rangeOptions) {
...
+    if (rangeOptions.hasLimit()) {
+      count = rangeOptions.getCount();
+      maxIndex -= rangeOptions.getOffset();
+      if (maxIndex < 0) {
+        return Collections.emptyList();
+      }
+    }
+
+    int startIndex = maxIndex - 1;
+    int endIndex = Math.min(count, maxIndex - minIndex);
```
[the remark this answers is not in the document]
ringles: Using addLimitToRange() now (with some tweaks to handle reverse ranges).

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
ringles: Yeah, per Donal we're using the addLimitToRange (with some tweaks now).

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/AbstractSortedSetRangeOptions.java
ringles: Adjusted the code, and the leading comment to match the test order.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -272,12 +272,12 @@ long zcard() {
   }
 
   long zcount(SortedSetScoreRangeOptions rangeOptions) {
-    AbstractOrderedSetEntry minEntry = new ScoreDummyOrderedSetEntry(rangeOptions.getMinimum(),
-        rangeOptions.isMinExclusive(), true);
+    AbstractOrderedSetEntry minEntry = new ScoreDummyOrderedSetEntry(rangeOptions.getStartRange(),
```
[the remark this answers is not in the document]
ringles: Didn't do it _quite_ that way, but extracted a couple methods that indeed can be used in several places. Good catch; hope this is along the lines of your thinking...

nonbinaryprogrammer (approved):

PR #6716, opened 2021-07-22
GEODE-9388: Implement Radish ZRANGEBYLEX Command

DonalEvans:
- Add ZRANGEBYLEX to supported commands
 - Introduce AbstractSortedSetRangeOptions class to handle behaviour
 common to -BYSCORE and -BYLEX ranges
 - Introduce SortedSetLexRangeOptions to handle parsing lexical ranges
 - Rename SortedSetRangeOptions to SortedSetScoreRangeOptions
 - Introduce MemberDummyOrderedSetEntry to support getting rank by lex
 - Rename DummyOrderedSetEntry to ScoreDummyOrderedSetEntry
 - Add unit tests for MemberDummyOrderedSetEntry
 - Add integration tests for ZRANGEBYLEX
 - Refactor RedisSortedSet.zrangebyscore() to use
 SortedSetScoreRangeOptions
 - Add missing HitsMissesIntegrationTest for ZRANGEBYSCORE
 - Refactor argument parsing in ZRangeByScoreExecutor to allow correct
 error behaviour with multiple LIMIT options, one of which is
 incorrectly formatted
 - Add tests to confirm correct behaviour for ZRANGEBYSCORE for above
 case

### For all changes:

### Note:

### DIFF
--- a/geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZRangeByLexIntegrationTest.java
+++ b/geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZRangeByLexIntegrationTest.java
@@ -104,13 +104,10 @@ public abstract class AbstractZRangeByLexIntegrationTest implements RedisIntegra
   }
 
   @Test
-  public void shouldError_givenNegativeLimitOffset() {
+  public void shouldError_givenNegativeZeroLimitOffset() {
     assertThatThrownBy(() -> jedis.sendCommand(KEY, Protocol.Command.ZRANGEBYLEX, KEY, "-", "+",
         "LIMIT", "-0", "10"))
             .hasMessageContaining(ERROR_NOT_INTEGER);
-
-    assertThatThrownBy(() -> jedis.zrangeByLex(KEY, "-", "+", -1, 10))
-        .hasMessageContaining(ERROR_NOT_INTEGER);
   }
 
   @Test
@@ -123,7 +120,7 @@ public abstract class AbstractZRangeByLexIntegrationTest implements RedisIntegra
     assertThatThrownBy(() -> jedis.sendCommand(KEY, Protocol.Command.ZRANGEBYLEX, KEY, "-", "+",
         "LIMIT", "0",
         "LIMIT", "0", "10"))
-            .hasMessageContaining(ERROR_SYNTAX);
+            .hasMessageContaining(ERROR_NOT_INTEGER);
   }
 
   @Test
@@ -376,6 +373,16 @@ public abstract class AbstractZRangeByLexIntegrationTest implements RedisIntegra
     // Range (v * 4) <= member name <= (v * 6), offset = 0, count = 10
     assertThat(jedis.zrangeByLex(KEY, "[" + min, "[" + max, 0, 10))
         .containsExactlyElementsOf(expected);
+  }
+
+  @Test
+  public void shouldReturnEmptyCollection_givenNonZeroNegativeLimitOffset() {
+    populateSortedSet();
+
+    int offset = -7;
+
+    // Range (v * 1) <= member name <= (v * 3), offset = 7, count = 10
+    assertThat(jedis.zrangeByLex(KEY, "-", "+", offset, 10)).isEmpty();
   }
 
   @Test
--- a/geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZRangeByLexIntegrationTest.java
+++ b/geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZRangeByLexIntegrationTest.java
@@ -1,0 +1,438 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.apache.geode.redis.internal.executor.sortedset;
+
+import static org.apache.geode.redis.RedisCommandArgumentsTestHelper.assertAtLeastNArgs;
+import static org.apache.geode.redis.internal.RedisConstants.ERROR_MIN_MAX_NOT_A_VALID_STRING;
+import static org.apache.geode.redis.internal.RedisConstants.ERROR_NOT_INTEGER;
+import static org.apache.geode.redis.internal.RedisConstants.ERROR_SYNTAX;
+import static org.apache.geode.test.dunit.rules.RedisClusterStartupRule.BIND_ADDRESS;
+import static org.apache.geode.test.dunit.rules.RedisClusterStartupRule.REDIS_CLIENT_TIMEOUT;
+import static org.apache.geode.util.internal.UncheckedUtils.uncheckedCast;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import junitparams.JUnitParamsRunner;
+import junitparams.Parameters;
+import org.apache.commons.lang3.StringUtils;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import redis.clients.jedis.HostAndPort;
+import redis.clients.jedis.JedisCluster;
+import redis.clients.jedis.Protocol;
+
+import org.apache.geode.redis.RedisIntegrationTest;
+import org.apache.geode.redis.internal.netty.Coder;
+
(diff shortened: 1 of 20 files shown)


geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/AbstractSortedSetRangeOptions.java
```diff
@@ -0,0 +1,103 @@
...
+      }
+      // Only set the values for the first limit we parse
+
+      int parsedOffset = narrowLongToInt(bytesToLong(offsetBytes));
+      int parsedCount = narrowLongToInt(bytesToLong(commandElements.get(commandIndex)));
+      if (!hasLimit) {
+        hasLimit = true;
+        this.offset = parsedOffset;
+        this.count = parsedCount;
+        if (this.count < 0) {
```
dschneider-pivotal, line 55: it seems a little better to only set "this.count" once. So do the < 0 check against parsedCount

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/ZRangeByScoreExecutor.java
```diff
@@ -33,35 +30,35 @@
...
     // clause overrides any previous ones
-    if (commandElements.size() >= 5) {
-      int currentCommandElement = 4;
+    // In order to match the error reporting behaviour of native Redis, the argument list must be
+    // parsed in reverse order. Stop parsing at index = 4, since 0 is the command name, 1 is the
+    // key, 2 is the min and 3 is the max
+    boolean withScores = false;
 
-      while (currentCommandElement < commandElements.size()) {
+    if (commandElements.size() >= 5) {
```
dschneider-pivotal, line 52: This whole block of code is much the same as the one in ZRangeByLexExecutor. I think all you need is somehow to tell rangeOptions.parseLimitArguments that it should also parse the withScores flag. You could then ask rangeOptions if isWithScores is true.
You could have the method that does this whole for loop in rangeOptions in a method that returns a RedisResponse. If it returns one then this code would return that response. Otherwise it would continue

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/netty/StringBytesGlossary.java
```diff
@@ -175,12 +175,13 @@
...
   public static final byte[] bABSTTL = stringToBytes("ABSTTL");
 
-  // AbstractZRangeExecutor
+  // Various ZRangeExecutors
   @MakeImmutable
   public static final byte[] bWITHSCORES = stringToBytes("WITHSCORES");
-
   @MakeImmutable
   public static final byte[] bLIMIT = stringToBytes("LIMIT");
+  public static final byte bPLUS = SIMPLE_STRING_ID; // +
```
dschneider-pivotal, line 183: what is the deal with the "b" prefix? At one point I thought it meant the constant was a byte array but here it is just a byte. When should the prefix be used? Also why does it exist? I have not seen this pattern before in java.

DonalEvans: Prior to the refactor that moved constants out of `Coder`, the approach to naming byte/byte array constants was somewhat inconsistent, with some using the "b" prefix and some not. I can't say why this naming convention was originally chosen (commit history shows that it seems to have been present since Geode was first imported to git back in 2015), but when working on https://issues.apache.org/jira/browse/GEODE-9287, I tried to enforce consistency in all the constant names. I think I was overzealous with the single-byte values though, as you're right, many of them don't use the "b" prefix and conceptually it might make more sense to only use it for byte arrays. I can go through and rename the ones are that currently using it so that they're not if you think that'd improve things.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -636,4 +663,48 @@ public int getSizeInBytes() {
...
+    @Override
+    public int compareMembers(byte[] array1, byte[] array2) {
+      int dummyNameComparison = checkDummyMemberNames(array1, array2);
+      if (dummyNameComparison != 0) {
+        return dummyNameComparison;
+      } else {
+        // If not using dummy member names, move on to actual lexical comparison
+        int stringComparison = javaImplementationOfAnsiCMemCmp(array1, array2);
+        if (stringComparison == 0) {
+          if (isMinimum && isExclusive || !isMinimum && !isExclusive) {
```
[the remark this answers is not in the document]
DonalEvans: Thanks for this! I was sure there must be a neater way of doing that but I wasn't sure what it was. I applied the same change to the constructor of `ScoreDummyOrderedSetEntry` which was the other case you saw.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/ZRangeByScoreExecutor.java
DonalEvans: I plan to do a major refactor of this code as part of the ZREVRANGEBYLEX work, once all the other range commands are implemented, as you're right, there's a lot of duplication and potential for simplification. I think we could probably manage to have a single AbstractZRangeExecutor class, extended by the ZRange, ZRevRange, ZRangeBYScore, ZRevRangeBYScore, ZRangeByLex and ZRevRangeByLex executors that populates an AbstractSortedSetRangeOptions class, extended by SortedSetRankRangeOptions, SortedSetScoreRangeOptions and SortedSetLexRangeOptions classes. Maybe I'm overthinking it though.

dschneider-pivotal: sounds good

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/netty/StringBytesGlossary.java
dschneider-pivotal: I can see the danger of having byte arrays that you want to be constants. In fact any array that you want to treat as a constant could be an issue because arrays can always be modified. Should all of these static final byte arrays be marked with the Immutable annotation? For constants I'm used to all upper case and underscores so the lowercase prefix is not helpful to me. I'm open to it if it actually has a specific meaning. When I first saw it I thought we were encoding type info in the symbolic name and if so I don't see a need for that.

dschneider-pivotal: See https://issues.apache.org/jira/browse/GEODE-9449 and https://issues.apache.org/jira/browse/GEODE-9450

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/AbstractSortedSetRangeOptions.java
```diff
@@ -37,25 +39,32 @@
   }
 
   void parseLimitArguments(List<byte[]> commandElements, int commandIndex) {
-    if (!equalsIgnoreCaseBytes(commandElements.get(commandIndex - 2), bLIMIT)) {
+    String element = bytesToString(commandElements.get(commandIndex));
+    if (!equalsIgnoreCaseBytes(commandElements.get(commandIndex), bLIMIT)) {
+      System.out.println("DONAL: throwing because " + element + " is not LIMIT");
```
dschneider-pivotal, line 44: remove the debug logging

dschneider-pivotal: the debug logging is in a couple of places

DonalEvans: Thanks for spotting that. It's gone now.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZRangeByLexIntegrationTest.java
```diff
@@ -0,0 +1,445 @@
...
+import org.junit.runner.RunWith;
+import redis.clients.jedis.HostAndPort;
+import redis.clients.jedis.JedisCluster;
+import redis.clients.jedis.Protocol;
+
+import org.apache.geode.redis.RedisIntegrationTest;
+import org.apache.geode.redis.internal.netty.Coder;
+
+@RunWith(JUnitParamsRunner.class)
+public abstract class AbstractZRangeByLexIntegrationTest implements RedisIntegrationTest {
```
jdeppe-pivotal, line 46: Although, not necessarily a correct dataset to run against, native redis will return all of these entries, but we don't:
```
  @Test
  public void validation() {
    jedis.zadd(KEY, 1, "a1");
    jedis.zadd(KEY, 2, "b1");
    jedis.zadd(KEY, 1, "c1");
    jedis.zadd(KEY, 1, "a2");
    jedis.zadd(KEY, 2, "b2");
    jedis.zadd(KEY, 1, "c2");

    Set<String> result = jedis.zrangeByLex(KEY, "-", "+");
    assertThat(result).containsExactly("a1", "b1", "c1", "a2", "b2", "c2");
  }
```

DonalEvans: Per [the docs for ZRANGEBYLEX](https://redis.io/commands/zrangebylex): "If the elements in the sorted set have different scores, the returned elements are unspecified." so it's fine for us to differ from Redis here, I think.

ringles (approved):
It'll be fun porting this into ZREVRANGEBYSCORE, but it looks solid.

PR #6721, opened 2021-07-24
GEODE-9456, GEODE-9452: Authentication Expiration

jinmeiliao:
* Created the AuthenticationExpirationException
* Created the test SecurityManager to expire users at a definite time
* add serialization conversion to support older clients
* Create tests to verify behavior for current and older clients


onichols-pivotal (approved):

mhansonp (approved):

jmelchio (approved):
LGTM

geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationDUnitTest.java
```diff
@@ -0,0 +1,109 @@
...
+      IntStream.range(0, 100).forEach(i -> {
+        region.put(i, "value" + i);
+      });
+    });
+
+    ExpirableSecurityManager securityManager =
+        (ExpirableSecurityManager) server.getCache().getSecurityService().getSecurityManager();
+
+    // assert that re-authentication happens
+    assertThat(securityManager.isExpired()).isTrue();
```
upthewaterspout, line 102: It seems like this test has the potential to be flaky. This is assuming that expiration happened somewhere in the middle of the puts, based on a 10 millisecond timeout. Maybe it would be clearer just to have an API on the test security manager to *force* expiration and call that in the middle of the puts? I'm not sure this test needs to use timeouts at all.

geode-old-client-support/src/main/java/com/gemstone/gemfire/OldClientSupportProvider.java
```diff
@@ -121,11 +123,21 @@ public Throwable getThrowable(Throwable theThrowable, KnownVersion clientVersion
     if (theThrowable == null) {
       return theThrowable;
     }
+
+    String className = theThrowable.getClass().getName();
+
+    // GEODE-9452, backward compatibility for authentication expiration
+    if (clientVersion.isOlderThan(KnownVersion.GEODE_1_14_0)) {
+      if (className.equals(AuthenticationExpiredException.class.getName())) {
+        return new AuthenticationRequiredException("User authorization attributes not found.");
```
upthewaterspout, line 132: nice!

geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationDUnitTest.java
jinmeiliao: Good idea. The stressNewTest also catches the flakiness.

geode-junit/src/main/java/org/apache/geode/security/ExpirableSecurityManager.java
```diff
@@ -0,0 +1,51 @@
...
+ * this is a test security manager that will authenticate credentials when username matches the
+ * password. It will authorize all operations. It keeps a list of expired users, and will throw
+ * AuthenticationExpiredException if the user is in that list. This security manager is usually used
+ * with NewCredentialAuthInitialize.
+ *
+ * make sure to call reset after each test to clean things up.
+ */
+
+public class ExpirableSecurityManager extends SimpleSecurityManager {
+  private static List<String> EXPIRED_USERS = new ArrayList<>();
```
kirklund, line 33: Statics should be reserved for constants or concepts that pertain to the class itself. Expired users as a concept belongs to the instance. This static variable would be better off as an instance field:
```
private final List<String> expiredUsers = new ArrayList<>();
```
The static methods would become instance methods.

kirklund (changes requested):

geode-junit/src/main/java/org/apache/geode/security/NewCredentialAuthInitialize.java
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+
+public class NewCredentialAuthInitialize implements AuthInitialize {
+  private static int count;
+
+  @Override
+  public Properties getCredentials(Properties securityProps, DistributedMember server,
+      boolean isPeer) throws AuthenticationFailedException {
+    count++;
```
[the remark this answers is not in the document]
jinmeiliao: I changed the int to AtomicInteger. In every VM, there is only one AuthInit object that gets created and used. I tried to make it an instance field, but since there is no easy way to get this object out of the cache and examine its attributes, I made this a static field

geode-junit/src/main/java/org/apache/geode/security/ExpirableSecurityManager.java
jinmeiliao: Same reason as the AuthInitialize. There is only one object every created for every VM, so a static field makes it easier for testing. And since both these classes are testing classes, i figured it's ok to be this way.

echobravopapa (approved):

kirklund (commented):
I think we still need javadocs on the new org.apache.geode.security.AuthenticationExpiredException since it's user-visible. Then I can add my approval.

jinmeiliao:
> I think we still need javadocs on the new org.apache.geode.security.AuthenticationExpiredException since it's user-visible. Then I can add my approval.

Oh, must have been distracted by adding the javadoc to the SecurityManager. Done now!

geode-core/src/integrationTest/java/org/apache/geode/management/internal/security/SecurityWithExpirationIniIntegrationTest.java
```diff
@@ -0,0 +1,66 @@
...
+import org.junit.experimental.categories.Category;
+
+import org.apache.geode.internal.security.SecurityService;
+import org.apache.geode.internal.security.SecurityServiceFactory;
+import org.apache.geode.security.AuthenticationExpiredException;
+import org.apache.geode.security.ExpirableSecurityManager;
+import org.apache.geode.test.junit.categories.SecurityTest;
+
+@Category({SecurityTest.class})
+public class SecurityWithExpirationIniIntegrationTest {
```
pivotal-jbarrett, line 34: `SecurityWithExpirationIni`, what is `Ini`?

geode-core/src/main/java/org/apache/geode/cache/client/internal/OpExecutorImpl.java
```diff
@@ -759,7 +760,8 @@ private Object executeWithPossibleReAuthentication(Connection conn, Op op) throw
       Throwable cause = sce.getCause();
       if ((cause instanceof AuthenticationRequiredException
           && "User authorization attributes not found.".equals(cause.getMessage()))
-          || sce.getMessage().contains("Connection error while authenticating user")) {
+          || sce.getMessage().contains("Connection error while authenticating user")
+          || cause instanceof AuthenticationExpiredException) {
         // (ashetkar) Need a cleaner way of doing above check.
```
pivotal-jbarrett, line 765: Remove the name. Is this todo style comment even relevant anymore?

geode-core/src/main/java/org/apache/geode/security/AuthenticationExpiredException.java
```diff
@@ -0,0 +1,32 @@
...
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package org.apache.geode.security;
+
+/**
+ * This exception is thrown by the SecurityManager's authorize method to indicate that the
+ * authentication has expired
```
pivotal-jbarrett, line 20: 'has expired.'

geode-core/src/main/java/org/apache/geode/security/SecurityManager.java
```diff
@@ -74,6 +74,8 @@ default void init(Properties securityProps) {}
    * @param principal The principal that's requesting the permission
    * @param permission The permission requested
    * @return true if authorized, false if not
+   *
+   *         possibly throw AuthenticationExpiredException
```
pivotal-jbarrett, line 78: Formatting.

Why not list it as a "checked exception". Then use the `@throws` tag to document it.

pivotal-jbarrett (changes requested):

geode-junit/src/main/java/org/apache/geode/security/ExpirableSecurityManager.java
```diff
@@ -0,0 +1,51 @@
...
+  public boolean authorize(Object principal, ResourcePermission permission) {
+    if (EXPIRED_USERS.contains(principal)) {
+      throw new AuthenticationExpiredException("User authentication expired.");
+    }
+    // always authorized
+    return true;
+  }
+
+  public static void addExpiredUser(String user) {
+    ExpirableSecurityManager.EXPIRED_USERS.add(user);
```
pivotal-jbarrett, line 45: Assuming this value is mutated by some other thread this member should be of some thread safe and synchronized  type. Also the class name is redundant here.

geode-old-client-support/src/main/java/com/gemstone/gemfire/OldClientSupportProvider.java
```diff
@@ -121,11 +123,21 @@ public Throwable getThrowable(Throwable theThrowable, KnownVersion clientVersion
     if (theThrowable == null) {
       return theThrowable;
     }
+
+    String className = theThrowable.getClass().getName();
+
+    // GEODE-9452, backward compatibility for authentication expiration
```
pivotal-jbarrett, line 129: Do not use ticket numbers in comments.

geode-old-client-support/src/main/java/com/gemstone/gemfire/OldClientSupportProvider.java
pivotal-jbarrett: To preserver consistency with the `AuthenticationRequiredException` perhaps they should both use the constant string here.

geode-old-client-support/src/main/java/com/gemstone/gemfire/OldClientSupportProvider.java
```diff
@@ -121,11 +123,21 @@ public Throwable getThrowable(Throwable theThrowable, KnownVersion clientVersion
...
+    if (clientVersion.isOlderThan(KnownVersion.GEODE_1_14_0)) {
```
pivotal-jbarrett, line 130: The current version in develop is 1.15. Unless this is back ported in the future to older versions then 1.15 should be used in this check.

pivotal-jbarrett (changes requested):

geode-junit/src/main/java/org/apache/geode/security/NewCredentialAuthInitialize.java
```diff
@@ -0,0 +1,52 @@
...
+ * this is used in conjunction with ExpirableSecurityManager. It will create a new set of
+ * credentials every time getCredentials are called, and they will always be authenticated
+ * and authorized by the ExpirableSecurityManager.
+ *
+ * make sure reset is called after each test to clean things up.
+ */
+
+
+public class NewCredentialAuthInitialize implements AuthInitialize {
+  private static AtomicInteger count = new AtomicInteger(0);
```
pivotal-jbarrett, line 33: make `final`.

geode-junit/src/main/java/org/apache/geode/security/NewCredentialAuthInitialize.java
```diff
@@ -0,0 +1,52 @@
...
+    credentials.put("security-password", "user" + currentCount);
+    return credentials;
+  }
+
+  public static String getCurrentUser() {
+    return "user" + count.get();
+  }
+
+  public static void reset() {
+    count = new AtomicInteger(0);
```
pivotal-jbarrett, line 50: `count.set(0);`

pivotal-jbarrett (changes requested):

geode-junit/src/main/java/org/apache/geode/security/ExpirableSecurityManager.java
pivotal-jbarrett: Perhaps `ConcurrentHashSet<Object>` is more appropriate.

geode-junit/src/main/java/org/apache/geode/security/ExpirableSecurityManager.java
```diff
@@ -0,0 +1,58 @@
...
+ * AuthenticationExpiredException if the user is in that list. This security manager is usually used
+ * with NewCredentialAuthInitialize.
+ *
+ * make sure to call reset after each test to clean things up.
+ */
+
+public class ExpirableSecurityManager extends SimpleSecurityManager {
+  // use static field for ease of testing since there is only one instance of this in each VM
+  // we only need ConcurrentHashSet here, but map is only construct available in the library
+  private static final ConcurrentHashMap<String, String> EXPIRED_USERS = new ConcurrentHashMap<>();
```
pivotal-jbarrett, line 36: If you want a concurrent set only.
```java
private static final Set<String> EXPIRED_USERS = ConcurrentHashMap.newKeySet();
```

geode-junit/src/main/java/org/apache/geode/security/UpdatableUserAuthInitialize.java
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * this is used in conjunction with ExpirableSecurityManager. It will create a new set of
+ * credentials every time getCredentials are called, and they will always be authenticated
+ * and authorized by the ExpirableSecurityManager.
+ *
+ * make sure reset is called after each test to clean things up.
+ */
+
+
```
pivotal-jbarrett, line 30: Detached Javadoc

geode-junit/src/main/java/org/apache/geode/security/ExpirableSecurityManager.java
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * this is a test security manager that will authenticate credentials when username matches the
+ * password. It will authorize all operations. It keeps a list of expired users, and will throw
+ * AuthenticationExpiredException if the user is in that list. This security manager is usually used
+ * with NewCredentialAuthInitialize.
+ *
+ * make sure to call reset after each test to clean things up.
+ */
+
```
pivotal-jbarrett, line 32: Detached javadoc.

pivotal-jbarrett (changes requested):

geode-junit/src/main/java/org/apache/geode/security/UpdatableUserAuthInitialize.java
jinmeiliao: what's a detached javadoc?

pivotal-jbarrett: A detached javadoc has blank lines between in and the symbol it is documenting. Javadocs should not be detached.

geode-core/src/main/java/org/apache/geode/security/SecurityManager.java
```diff
@@ -74,8 +74,11 @@ default void init(Properties securityProps) {}
    * @param principal The principal that's requesting the permission
    * @param permission The permission requested
    * @return true if authorized, false if not
+   *
+   * @throw AuthenticationExpiredException
```
pivotal-jbarrett, line 78: The tag is `@throws`. Explain when the exception is thrown as part of the detail.

```
@throws AuthenticationExpiredException if the principal has expired.
```

pivotal-jbarrett (changes requested):

geode-junit/src/main/java/org/apache/geode/security/UpdatableUserAuthInitialize.java
```diff
@@ -0,0 +1,53 @@
...
+    credentials.put("security-password", user);
+    return credentials;
+  }
+
+  public static String getUser() {
+    return user;
+  }
+
+  public static void setUser(String user) {
+    UpdatableUserAuthInitialize.user = user;
```
pivotal-jbarrett, line 47: If this is going to be set by another thread than those reading the `user` value then `user` should be `volatile` or otherwise be made thread safe and synchronized.

pivotal-jbarrett (changes requested):

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ServerConnection.java
```diff
@@ -100,6 +100,8 @@
   private static final String DISALLOW_INTERNAL_MESSAGES_WITHOUT_CREDENTIALS_NAME =
       "geode.disallow-internal-messages-without-credentials";
 
+  public static final String USER_NOT_FOUND = "User authorization attributes not found.";
```
kirklund, line 103: Please change the constant to `private`.

geode-junit/src/main/java/org/apache/geode/security/UpdatableUserAuthInitialize.java
kirklund: It's probably best to just change the `user` field to be of type AtomicReference<String>:
```
private static final AtomicReference<String> USER = new AtomicReference<>();
```

kirklund (changes requested):

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ServerConnection.java
jinmeiliao: the string is used in other classes as well.

geode-core/src/main/java/org/apache/geode/security/AuthenticationExpiredException.java
```diff
@@ -0,0 +1,32 @@
...
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package org.apache.geode.security;
+
+/**
+ * This exception is thrown by the SecurityManager's authorize method to indicate that the
+ * authentication has expired.
+ */
```
pivotal-jbarrett, line 21: Let's take this small surface area to fix a wrong. Not all exceptions should be unchecked. It is bad pattern to have unchecked exceptions thrown all over the place. It makes it hard for authors to know when and where to catch exceptions. 

`AuthenticationExpiredException` should be a checked exception. To do so we need to correct the wrong documented in the `GemFireException` where it was derived from `RuntimeException`. We should define a new base exception `GeodeException` that derives from the checked `Exception`. This new exception should derive from `GeodeException` and be checked.

pivotal-jbarrett (changes requested):

kirklund (approved):

geode-core/src/main/java/org/apache/geode/security/AuthenticationExpiredException.java
jinmeiliao: wouldn't this break the API?

pivotal-jbarrett: Nope. Exceptions on an interface are optional on the implementation. Checked exceptions are no exception. Current implementations with both compile and run without issue. This particular interface is an SPI which means the only callers of this API are internal to Geode, meaning only Geode code would be subject to the checked exception being handled.

jinmeiliao: @kohlmu-pivotal seems to indicate that it does in the RFC. Will let him comment on this.

jinmeiliao: this is from Oracle java doc site:
13.4.21. Method and Constructor Throws
Changes to the throws clause of methods or constructors do not break compatibility with pre-existing binaries; these clauses are checked only at compile time.

pivotal-jbarrett: > this is from Oracle java doc site:
> 13.4.21. Method and Constructor Throws
> Changes to the throws clause of methods or constructors do not break compatibility with pre-existing binaries; these clauses are checked only at compile time.

Yes, only the calling code at compile time would be exposed to this change and since the only caller is the internals we can safely make this SPI change.

jinmeiliao: There is a `GemfireCheckedExcpetion`, Do we want to do AuthenticationExpiredException -> GemfireCheckedException, or add a `GemfireCheckedSecurityException` in between?

jinmeiliao: Actually, none of the callers needs to handle this exception, it just gets thrown all the way to the client, and when client got it, it gets wrapped in the `ServerOperationException` as a cause, wondering what's really the benefit of making it a checked exception.....

rhoughton-pivot (approved):

geode-core/src/main/java/org/apache/geode/security/AuthenticationExpiredException.java
pivotal-jbarrett: We better be catching it to remove the long ID from the map on the server.

jinmeiliao: Oh, yeah, that's a story we haven't touched upon yet

pivotal-jbarrett: > There is a `GemfireCheckedExcpetion`, Do we want to do AuthenticationExpiredException

I don't see the point in continuing to use the GemFire named symbols. It's about time we start making migration away where we can.

pivotal-jbarrett: > Oh, yeah, that's a story we haven't touched upon yet

And that is why making it checked is important. It forces us to actually deal with the exception and not forget it could be thrown.

jinmeiliao: @jake Turns out the `authorize` has to throw an unchecked exception or internally we have to turn it into an unchecked exception anyway, since this method is called by Shiro's `AuthrizingRealm` (see our implementation `CustomAuthRealm`), it's interface `isPermitted` doesn't throw any exception, only returns a boolean

pivotal-jbarrett: Wow, another point against Shiro. Ok.

pivotal-jbarrett: At the very least then I don't think it makes sense for this exception to extend `AuthenticationRequiredException`. Is there a relationship there I am not thinking of?

pivotal-jbarrett (approved):
I question the inheritance relationship between `AuthenticationExpiredException` and `AuthenticationRequiredException` but otherwise approve.

geode-core/src/main/java/org/apache/geode/security/AuthenticationExpiredException.java
jinmeiliao: Not from a coding perspective, I just though it's a natural extension, "auth expired means auth required". I can change it  to extend `GemfireSecurityException"

jinmeiliao:
Changing the base branch to a feature branch

PR #6727, opened 2021-07-29
GEODE-9493: rework sizing of RedisString, RedisHash, RedisSet, and RedisSortedSet

dschneider-pivotal:
Instead of using a general purpose sizer for sizing the contents of the map and set, now these classes are abstract and the subclasses that know the type of the element have some simple code to size element.
Also the overhead constants are now computed at runtime using JvmSizeUtils.sizeClass which will give us accurate numbers when the sizing is done on different types of jvms. The big difference is when the jvm has a large heap or uses ZGC since both of these disable compressed oops. Some of the existing tests exist to compute these overhead constants; I got them passing but it might make sense to remove them or at least rename them.

Note that RedisSortedSet no longer does extra sizing that it then undoes with the "sizeInBytesAdjustment". Instead it just has a score set that does not size the element at all since it is already sized by the MemberMap. 
Also the computing the byte size of the score set is now simpler getting rid of another int field.
Removal of these fields reduced the memory overhead of RedisSortedSet.

### For all changes:

### Note:


mhansonp (approved):
Nice readability improvements among other things.

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisSortedSetTest.java
```diff
@@ -656,7 +656,9 @@ public void redisSortedSetGetSizeInBytes_isAccurateForZpopmax() {
     for (int i = 0; i < numberOfEntries; ++i) {
       sortedSet.zpopmax(mockRegion, mockKey, 1);
       sortedSet.clearDelta();
-      assertThat(sizer.sizeof(sortedSet)).isEqualTo(sortedSet.getSizeInBytes());
+      int actualSize = sizer.sizeof(sortedSet) - sizer.sizeof(ByteArrays.HASH_STRATEGY);
+      int calculatedSize = sortedSet.getSizeInBytes();
+      assertThat(calculatedSize).isEqualTo(actualSize);
```
nonbinaryprogrammer, line 661: the use of "actual" here is a bit confusing, since in general we do assertThat(actual).isEqualTo(expected). In this case the expected is the "actual" size and actual is "calculated". I think it would be less confusing to use a variable name like "measuredSize". (for this and the above tests)

nonbinaryprogrammer (commented):
looks good. I like this solution better than the constants. this seems less brittle. Just a couple of small suggestions

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/collections/SizeableObject2ObjectOpenCustomHashMapWithCursor.java
```diff
@@ -196,46 +194,60 @@ public V put(K k, V v) {
...
+    // in which case these objects would account for over 100 bytes on a 64-bit jvm,
+    // It is not clear to me why the implementors of Object2ObjectOpenCustomHashMap chose to
+    // cache this immutable object instead of just creating it every time. They have no state
+    // to initialize when constructed so one thing we could do is override the methods that cache it
+    // to either null out the cached field after calling the super method, or have a brand new impl
+    // that just creates and returns an instance without caching. The cache fields themselves would
+    // be a waste of memory in that case but only 24 bytes (12 with compressed oops).
+    // Another possibility would be for our code that uses the map to only call entrySet(), that way
+    // one instance would be cached instead of 3. You can always get keys and values from the
+    // entrySet.
```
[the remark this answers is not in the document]
dschneider-pivotal: thanks it is GEODE-9497

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisSortedSetTest.java
dschneider-pivotal: I changed them to expectedSize and actualSize

upthewaterspout (approved):

kirklund (approved):

dschneider-pivotal:
The acceptance test failure is the known issue: https://issues.apache.org/jira/browse/GEODE-9319

PR #6729, opened 2021-07-29
GEODE-9480: Changing the version with ordinal 121 to be 1.13.2

upthewaterspout:
This protocol change actually happened in 1.13.2. 1.13.1 is still using the
same protocol ordinal as 1.13.0, which is 120.

Evidence
* b2d205459bf is the commit that introduced this version in support/1.13
* b2d205459bf is only contained in rel/v1.13.2 and above
* Running geode 1.13.1, it reports its protocol version as 120.

### DIFF
--- a/geode-serialization/src/main/java/org/apache/geode/internal/serialization/KnownVersion.java
+++ b/geode-serialization/src/main/java/org/apache/geode/internal/serialization/KnownVersion.java
@@ -205,12 +205,12 @@ public class KnownVersion extends AbstractVersion {
       new KnownVersion("GEODE", "1.13.0", (byte) 1, (byte) 13, (byte) 0, (byte) 0,
           GEODE_1_13_0_ORDINAL, true);
 
-  private static final short GEODE_1_13_1_ORDINAL = 121;
+  private static final short GEODE_1_13_2_ORDINAL = 121;
 
   @Immutable
-  public static final KnownVersion GEODE_1_13_1 =
-      new KnownVersion("GEODE", "1.13.1", (byte) 1, (byte) 13, (byte) 1, (byte) 0,
-          GEODE_1_13_1_ORDINAL, true);
+  public static final KnownVersion GEODE_1_13_2 =
+      new KnownVersion("GEODE", "1.13.2", (byte) 1, (byte) 13, (byte) 1, (byte) 0,
+          GEODE_1_13_2_ORDINAL, true);
 
   private static final short GEODE_1_14_0_ORDINAL = 125;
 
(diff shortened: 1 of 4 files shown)


pivotal-jbarrett:
Is there a GEODE ticket associated with these changes?

geode-serialization/src/main/java/org/apache/geode/internal/serialization/KnownVersion.java
```diff
@@ -205,12 +205,12 @@
       new KnownVersion("GEODE", "1.13.0", (byte) 1, (byte) 13, (byte) 0, (byte) 0,
           GEODE_1_13_0_ORDINAL, true);
 
-  private static final short GEODE_1_13_1_ORDINAL = 121;
+  private static final short GEODE_1_13_2_ORDINAL = 121;
```
pivotal-jbarrett, line 208: If we are removing 1.13.1 we need to remove it from the unreleased 1.14 and develop branches also.

pivotal-jbarrett (changes requested):

geode-serialization/src/main/java/org/apache/geode/internal/serialization/KnownVersion.java
pivotal-jbarrett: I take that back, we can't replace this ordinal because we have already released versions with this ordinal defined. It may not be safe to have it suddenly mean something different. The safer thing to do is define a new ordinal 122 for 1.13.2.

upthewaterspout: I'm not sure I follow your reasoning here. 1.13.1 released with ordinal 120. 1.13.2 released with ordinal 121.

The problem here is that the name of this constant is wrong. It should be called GEODE_1_13_2_ORDINAL, because that is the release in which this ordinal was used.

I guess there is a question of whether some external code is using this constant, since there will no longer be a constant named GEODE_1_13_1_ORDINAL. However this is an internal class, so it can be expected to change - in fact this whole class was renamed in 1.14. The most likely project to use these internal constants would be spring data geode, but that project does not use this internal class.

upthewaterspout:
> Is there a GEODE ticket associated with these changes?

Yes! GEODE-9480. Updated the title to reflect that.

geode-serialization/src/main/java/org/apache/geode/internal/serialization/KnownVersion.java
pivotal-jbarrett: If you are comfortable then I am comfortable.

pivotal-jbarrett (approved):

onichols-pivotal (changes requested):
As long as it's clear that this is a cosmetic change, not intended to alter behavior in any way, I'm ok with it.  Just adding a comment might have been equally effective, and probably still a good idea.  Many developers get confused because there is a vestigial version number attached to serialization ordinals that may look similar to Geode product version (which is a totally unrelated concept).  However, if you want to follow the convention that constant names for serialization ordinals should match the product version they were introduced (rather than simply sequential numbering), that's cool.  Just comment that this is a nice convention to make it easier to reason about what versions introduced serialization changes.

echobravopapa (approved):

onichols-pivotal (approved):

jdeppe-pivotal (approved):

kirklund (approved):

PR #6730, opened 2021-07-29
GEODE-9480: Changing the version with ordinal 121 to be 1.13.2

upthewaterspout:
This protocol change actually happened in 1.13.2. 1.13.1 is still using the
same protocol ordinal as 1.13.0, which is 120.

Evidence
* b2d205459bf is the commit that introduced this version in support/1.13
* b2d205459bf is only contained in rel/v1.13.2 and above
* Running geode 1.13.1, it reports it's protocol version as 120.

(cherry picked from commit 2a90a583015352c1cc0f041008ec2a6ea7b5505c)


pivotal-jbarrett (changes requested):
See concerns on similar PR https://github.com/apache/geode/pull/6730

onichols-pivotal (changes requested):
Product version is not the same as serialization version.  Protocol decision should never be based on production version.  KnownVersion tracks serialization version, not product version.  I think no change is necessary here: product versions 1.13.0 and 1.13.1 both used serialization version 1.13.0, while 1.13.2 and later use serialization version 1.13.1

onichols-pivotal (approved):

pivotal-jbarrett (approved):

PR #6731, opened 2021-07-30
GEODE-9483: Make StartLocatorCommandIntegrationTest not start unneeded locator

demery-pivotal:
Two of the tests in `StartLocatorCommandIntegrationTest` launch a locator
process that binds to the default HTTP service port (7070). This can
interfere with the tests that explicitly verify that the system uses the
correct HTTP service port by default.

Neither test actually needs the launched locator for any purpose.
Neither test makes a single assertion about the locator. When the tests
exit, they leave the locator running. In fact, the locator typically
finishes launching well after the tests have all exited.

Rather than fix the port assignment, this commit changes the tests so
that they do not actually launch a locator. Two of the tests in this
class already do this.


demery-pivotal:
The distributed test failure is unrelated to this PR. `RepeatedRebalanceDUnitTest.testSecondRebalanceIsNotNecessaryWithAddedAndRestartedMembers` is known to fail intermittently:  https://issues.apache.org/jira/browse/GEODE-7603

jinmeiliao (approved):

PR #6748, opened 2021-08-09
GEODE-9492: Ignore default port test in stress tests

demery-pivotal:
PROBLEM

`GeodeRedisServerStartupDUnitTest.startupOnDefaultPort()` tests the
system's use of a fixed port. Repeat test tasks run multiple instances
of this test concurrently. Most of the concurrently running instances of
the test are likely to fail when they find that port already in use.

SOLUTION

- Add an `IgnoreInRepeatTestTasks` test category.
- Configure repeat tests to exclude tests with the
  `IgnoreInRepeatTestTasks` category.
- Add the `IgnoreInRepeatTestTasks` category to
  `GeodeRedisServerStartupDUnitTest.startupOnDefaultPort()`

### DIFF
--- a/geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/GeodeRedisServerStartupDUnitTest.java
+++ b/geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/GeodeRedisServerStartupDUnitTest.java
@@ -29,6 +29,7 @@ import java.net.Socket;
 import org.junit.ClassRule;
 import org.junit.Rule;
 import org.junit.Test;
+import org.junit.experimental.categories.Category;
 
 import org.apache.geode.internal.AvailablePortHelper;
 import org.apache.geode.internal.inet.LocalHostUtil;
@@ -37,6 +38,7 @@ import org.apache.geode.redis.internal.GeodeRedisService;
 import org.apache.geode.test.dunit.rules.ClusterStartupRule;
 import org.apache.geode.test.dunit.rules.MemberVM;
 import org.apache.geode.test.dunit.rules.RedisClusterStartupRule;
+import org.apache.geode.test.junit.categories.IgnoreInRepeatTestTasks;
 import org.apache.geode.test.junit.rules.GfshCommandRule;
 
 public class GeodeRedisServerStartupDUnitTest {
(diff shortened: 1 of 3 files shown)


kirklund (approved):

upthewaterspout (approved):
redis file looks good.

mhansonp (approved):
That seems reasonable. Odd that this is the first we have seen of this kind of problem. One would have thought other default ports would have this problem. But perhaps, this test is flawed. Perhaps breaking it into a test of the default port number and a test of the starting up on a specified port would be more appropriate then altering our test framework?

kirklund:
> That seems reasonable. Odd that this is the first we have seen of this kind of problem. One would have thought other default ports would have this problem. But perhaps, this test is flawed. Perhaps breaking it into a test of the default port number and a test of the starting up on a specified port would be more appropriate then altering our test framework?

@mhansonp I think it started appearing because Dale removed use of Docker for these tests.

demery-pivotal:
> Odd that this is the first we have seen of this kind of problem. One would have thought other default ports would have this problem.

It's a consequence of two things. First, we are no longer dockerizing test workers. Until a few weeks ago, each test worker ran in a separate docker container, and within each container fewer processes were competing for ports. Now that dozens of concurrently running tests compete for the same ports, we will see more port conflicts like this.

Second, very few tests use default ports, so there are likely to be few conflicts… except in stress tests, which pit multiple instances of a test against each other.

> But perhaps, this test is flawed. Perhaps breaking it into a test of the default port number and a test of the starting up on a specified port would be more appropriate then altering our test framework?

I chose this approach because it affects only tests that explicitly test the use of default ports. And my guess (or perhaps merely my hope) is that very few tests do that: Only one per service. So very few tests will use this technique to avoid stress. I hope.

As for altering the test framework, my assessment (for now) is that including these few tests in stress test runs isn't worth the extra complexity in the test framework.

onichols-pivotal (approved):

rhoughton-pivot (approved):

mhansonp:
> As for altering the test framework, my assessment (for now) is that including these few tests in stress test runs isn't worth the extra complexity in the test framework.

I suggested that we modify this test to be repeated test capable rather than adding a new category which you have done.

E.g. Pulling the default port in into a default port test. Then modify this code to test that it works fine with a configured port. I am not very attached to this idea, but it seems cleaner. Because then the redis code will get repeat testing.

demery-pivotal:
> I suggested that we modify this test to be repeated test capable rather than adding a new category which you have done.

Ah. I misread "then" as meaning "and then," and my misreading changed the entire meaning of what you wrote. Sorry about that.

PR #6751, opened 2021-08-09
GEODE-9491: Fix REST API test use of ephemeral ports

demery-pivotal:
PROBLEM

A `RestAPICompatibilityTest` backward compatibility test assumed that
two ephemeral ports would remain available while not in use. The test:

1. Launched two locators, each on an ephemeral port.
2. Stopped each locator.
3. Restarted each locator on its original (ephemeral) port.

In between steps 2 and 3, some other process can bind to the port,
causing the restart to fail.

SOLUTION

Change the test to assign a non-ephemeral port to each locator.


demery-pivotal:
The distributed test failure is unrelated to this PR. It is due to a test known to fail intermittently: https://issues.apache.org/jira/browse/GEODE-7739

jinmeiliao (approved):

PR #6752, opened 2021-08-10
GEODE-9489: Assert exception type, not message

demery-pivotal:
PROBLEM

`GeodeRedisServerStartupDUnitTest.startupFailsGivenPortAlreadyInUse()`
asserted that the root cause message of the expected exception was
exactly "Address already in use". The JDK makes no guarantee about the
exception message of a `BindException`, so this assertion is overly
strict.

SOLUTION

Change the test to assert that the thrown exception is a
`BindException`.


demery-pivotal:
The stress test failures are due to another test method in this class. That test method is being addressed in a separate PR: https://github.com/apache/geode/pull/6748

demery-pivotal:
The distributed test failure is from a test known to fail intermittently: https://issues.apache.org/jira/browse/GEODE-7183

demery-pivotal:
Closing as duplicate of: https://github.com/apache/geode/pull/6744

PR #6755, opened 2021-08-10
GEODE-9460: Add testing for mutli-user scenarios

jmelchio:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationDUnitTest.java
+++ b/geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationDUnitTest.java
@@ -18,7 +18,6 @@ import static org.apache.geode.distributed.ConfigurationProperties.SECURITY_CLIE
 import static org.apache.geode.test.version.VersionManager.CURRENT_VERSION;
 import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
 
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
@@ -39,7 +38,6 @@ import org.apache.geode.cache.RegionShortcut;
 import org.apache.geode.cache.client.ClientCache;
 import org.apache.geode.cache.client.ClientRegionFactory;
 import org.apache.geode.cache.client.ClientRegionShortcut;
-import org.apache.geode.cache.client.ServerOperationException;
 import org.apache.geode.test.dunit.rules.ClientVM;
 import org.apache.geode.test.dunit.rules.ClusterStartupRule;
 import org.apache.geode.test.junit.categories.SecurityTest;
@@ -126,7 +124,9 @@ public class AuthExpirationDUnitTest {
             .withServerConnection(serverPort));
 
     clientVM.invoke(() -> {
-      UpdatableUserAuthInitialize.setUser("user1");
+      UpdatableUserAuthInitialize.setUser("serviceUser0");
+      UpdatableUserAuthInitialize.setAttempts(0);
+      UpdatableUserAuthInitialize.setAlternate(false);
       ClientCache clientCache = ClusterStartupRule.getClientCache();
       assert clientCache != null;
       clientCache.createClientRegionFactory(ClientRegionShortcut.PROXY).create("region");
@@ -136,9 +136,11 @@ public class AuthExpirationDUnitTest {
       RegionService regionService = clientCache.createAuthenticatedView(userSecurityProperties);
       Region<Object, Object> region = regionService.getRegion("/region");
       region.put(0, "value0");
+      regionService.close();
     });
 
     clientVM.invoke(() -> {
+      UpdatableUserAuthInitialize.setUser("serviceUser1");
       ClientCache clientCache = ClusterStartupRule.getClientCache();
       Properties userSecurityProperties = new Properties();
       userSecurityProperties.put("security-username", "serviceUser1");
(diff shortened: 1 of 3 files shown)


geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationDUnitTest.java
```diff
@@ -109,4 +116,92 @@ public void clientShouldReAuthenticateWhenCredentialExpiredAndOperationSucceed()
...
+    clientVM.invoke(() -> {
+      ClientCache clientCache = ClusterStartupRule.getClientCache();
+      Properties userSecurityProperties = new Properties();
+      userSecurityProperties.put("security-username", "serviceUser1");
+      userSecurityProperties.put("security-password", "serviceUser1");
+      assert clientCache != null;
+      RegionService regionService = clientCache.createAuthenticatedView(userSecurityProperties);
+      Region<Object, Object> region = regionService.getRegion("/region");
+      region.put(1, "value1");
+      regionService.close();
```
jinmeiliao, line 150: I would think user would continue to use same regionService and expect the next put to succeed. (the whole expire user and re-authenticate happens behind the scene and the client is doing operations like nothing happened).

jmelchio: I agree. I'll see if I can find reuse the same `regionService` from one `clientVM` invocation to another. If you have ideas on how to accomplish this I'm open to suggestions.

jmelchio: I was unable to reuse the `regionService` but came up with a way accomplish the automatic re-authentication behind the scenes with changes to the `AuthInitialize` implementation we're using for the tests.

geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationDUnitTest.java
```diff
@@ -109,4 +114,82 @@ public void clientShouldReAuthenticateWhenCredentialExpiredAndOperationSucceed()
...
+      clientCache.createClientRegionFactory(ClientRegionShortcut.PROXY).create("region");
+      Properties userSecurityProperties = new Properties();
+      userSecurityProperties.put("security-username", "serviceUser0");
+      userSecurityProperties.put("security-password", "serviceUser0");
+      RegionService regionService = clientCache.createAuthenticatedView(userSecurityProperties);
+      Region<Object, Object> region = regionService.getRegion("/region");
+      region.put(0, "value0");
+      regionService.close();
+    });
+
```
mhansonp, line 141: Why are splitting this blocks for the same VM?

mhansonp (commented):
I am confused about the repetition of the clientvm blocks rather than continuing it.

geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationDUnitTest.java
jmelchio: > I am confused about the repetition of the `clientvm blocks rather than continuing it.

You're right, in most cases they can be continued. I organized it this way as logical steps but unless we add a user to the expired user list in between there is no need for separation of the blocks. I will change it.

geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationDUnitTest.java
```diff
@@ -85,9 +86,10 @@ public void clientShouldReAuthenticateWhenCredentialExpiredAndOperationSucceed()
     clientVM.invoke(() -> {
       ClientCache clientCache = ClusterStartupRule.getClientCache();
       UpdatableUserAuthInitialize.setUser("user1");
-      ClientRegionFactory clientRegionFactory =
+      assert clientCache != null;
```
kirklund, line 89: Please use actual Assertions in tests instead of Java asserts.

jmelchio: The reason for the java assert is not to test something but to get rid of a warning in the gutter. I assume that this is a valid use of java asserts?

onichols-pivotal (changes requested):
please resolve conflicts

onichols-pivotal (approved):

jinmeiliao (approved):

geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationDUnitTest.java
mhansonp: I'm of the mindset that an assertThat(clientCache).isNotNull(); would be better than what is there. That is pretty much the standard in tests. Whether it is to clear a warning or not.

geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationDUnitTest.java
```diff
@@ -98,7 +103,8 @@ public void clientShouldReAuthenticateWhenCredentialExpiredAndOperationSucceed()
     clientVM.invoke(() -> {
       UpdatableUserAuthInitialize.setUser("user2");
       ClientCache clientCache = ClusterStartupRule.getClientCache();
-      Region region = clientCache.getRegion("region");
+      assert clientCache != null;
```
mhansonp, line 106: same here.

geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationDUnitTest.java
mhansonp: One note that is most likely irrelevant in this case is that splitting it takes more time for context switching.

mhansonp (approved):
I have changes I would really like to see, but I don't have any that I would require.

PR #6740, opened 2021-08-04
GEODE-4181: Add JUnit 5 Support

demery-pivotal:
Added JUnit 5 support to all Geode projects that use `geode-junit`.

STANDARD TEST TASKS

Updated **./gradle/test.gradle** to configure these standard test tasks
to use JUnit Platform to run tests:
- `test` and `repeatUnitTest`
- `acceptanceTest` and `repeatAcceptanceTest`
- `distributedTest` and `repeatDistributedTest`
- `integrationTest` and `repeatIntegrationTest`
- `performanceTest`
- `uiTest`
- `upgradeTest` and `repeatUpgradeTest`

STANDARD TEST MODULES

Updated **./geode-junit/build.gradle**:
- Added `junit-jupiter-api` and `junit-jupiter-params` as API
  dependencies.
- Added `junit-jupiter-engine` and `junit-vintage-engine` as
  implementation dependencies.

These changes add JUnit 5 support to any source set that depends on
`geode-junit`, either directly or via `geode-dunit`.

OTHER PROJECTS

Added `junit-vintage-engine` dependency directly to each project that
runs tests without `geode-junit` or `geode-dunit`:
- `geode-common`
- `geode-concurrency-test`
- `geode-jmh`
- `geode-modules`
- `geode-rebalancer`
- `static-analysis:pmd-rules`

These changes **do not** add JUnit 5 support to these projects.
Developers who want JUnit 5 support in these projects can declare
dependencies on `junit-jupiter-api`, `junit-jupiter-params`, and
`junit-jupiter-engine`.

SPECIFIC TESTS

Change `ConcurrencyRuleTest` to expect the exception types and exception
messages thrown by AssertJ when `opentest4j` is on the classpath.

PR #6762, opened 2021-08-12
Revert GEODE-4181: Add JUnit 5 Support

onichols-pivotal:
due to large number of Windows CI failures starting after #6740 + #6742 were merged, I want to see if reverting this change makes a difference.  If so, we should go through with this revert and try again to make Junit5 work in a new PR with Windows label.


kirklund (approved):

rhoughton-pivot (approved):
Worth a try. Thanks, Owen!

PR #6763, opened 2021-08-16
GEODE-4181: Reduce ChildVM command line length

demery-pivotal:
PROBLEM

Windows limits command lines to a maximum length of ~32k characters. The
command line for DUnit child VMs is very close to this limit.

Currently this prevents us from adding JUnit 5 support. JUnit 5 adds 8
more jars to the classpath, which adds ~1300 characters to the command
line, which puts the command line over the limit.

SOLUTION

Write the ChildVM's entire classpath to a "pathing" jar, then include
only that pathing jar on the command line. This results in command line
that is shorter by many thousands of characters, and which does not grow
as we add more jars or other classpath elements.


kirklund (approved):
Looks good!

mhansonp (approved):

PR #6764, opened 2021-08-16
GEODE-9365: Getting rid of an unnecessary synchronize

mhansonp:
GEODE-9365: Getting rid of an unnecessary synchronize

- plus general cleanup

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/ha/HARegionQueue.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/ha/HARegionQueue.java
@@ -42,6 +42,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 
 import org.apache.logging.log4j.Logger;
@@ -161,17 +162,17 @@ public class HARegionQueue implements RegionQueue {
    */
   protected volatile Map indexes = Collections.unmodifiableMap(new HashMap());
 
-  private StoppableReentrantReadWriteLock rwLock;
+  private final StoppableReentrantReadWriteLock rwLock;
 
-  private StoppableReentrantReadWriteLock.StoppableReadLock readLock;
+  private final StoppableReentrantReadWriteLock.StoppableReadLock readLock;
 
-  private StoppableWriteLock writeLock;
+  private final StoppableWriteLock writeLock;
 
   /** The name of the {@code Region} backing this queue */
-  private String regionName;
+  private final String regionName;
 
   /** The ClientProxyMembershipID associated with the ha queue */
-  private ClientProxyMembershipID clientProxyID;
+  private final ClientProxyMembershipID clientProxyID;
 
   /**
    * The statistics for this queue
@@ -223,13 +224,13 @@ public class HARegionQueue implements RegionQueue {
   private static QueueRemovalThread qrmThread;
 
   /** protects from modification during GII chunking */
-  private StoppableReentrantReadWriteLock giiLock;
+  private final StoppableReentrantReadWriteLock giiLock;
 
   /** the number of concurrent GII requests being served */
-  private volatile int giiCount;
+  private final AtomicInteger giiCount = new AtomicInteger();
 
   /** queue to hold events during GII transfer so we do not modify the queue during chunking */
-  private Queue giiQueue = new ConcurrentLinkedQueue();
+  private final Queue giiQueue = new ConcurrentLinkedQueue();
 
   /**
    * Constant used to indicate the instance of BlockingHARegionQueue. The static function used for
@@ -282,7 +283,7 @@ public class HARegionQueue implements RegionQueue {
   /** Boolean to indicate whether destruction of queue is in progress. */
   protected volatile boolean destroyInProgress = false;
 
-  private CancelCriterion stopper;
+  private final CancelCriterion stopper;
 
   /** @since GemFire 5.7 */
   protected byte clientConflation = Handshake.CONFLATION_DEFAULT;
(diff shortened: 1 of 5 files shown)


geode-core/src/main/java/org/apache/geode/internal/cache/ha/HARegionQueue.java
```diff
@@ -2251,48 +2237,54 @@ void checkQueueSizeConstraint() throws InterruptedException {
...
+            checkQueueSizeConstraintCore(duration);
+          } // synchronized (this.permitMon)
+        } // if (putPermits <= 0)
+        putPermits.decrementAndGet();
+      }
+    }
+
+    /* Do not call this method directly from anywhere except checkQueueSizeConstraint */
+    private void checkQueueSizeConstraintCore(long duration) throws InterruptedException {
+      if (reconcilePutPermits() <= 0) {
```
dschneider-pivotal, line 2240: since (in my design) the caller has already decd putPermits, then we can stop waiting in this method if reconcilePutPermits() >= 0. So I think you want this comparison to be < 0.

geode-core/src/main/java/org/apache/geode/internal/cache/ha/HARegionQueue.java
```diff
@@ -2251,48 +2237,54 @@ void checkQueueSizeConstraint() throws InterruptedException {
...
-                    throw ex;
-                  }
-                }
-              }
-            } // synchronized (this.permitMon)
-          } // if (putPermits <= 0)
-          --putPermits;
-        } // synchronized (this.putGuard)
+        long startTime = System.currentTimeMillis();
+        if (putPermits.get() <= 0) {
```
dschneider-pivotal, line 2227: instead of putPermits.get() here I think it should be putPermits.decrementAndGet(). You also need to get rid of the decrementAndGet call at the end (line 2247).
Also the System.currentTimeMillis() call should be after the decrementAndGet check.
You can just change this code to if (putPermits.decrementAndGet() >= 0) return; Note that if after the dec it is zero we are still good. We only need to wait if it drops below 0.
So the idea is that it first dec the atomic and checks if the result is > 0. If it is its done. If not it then drops into your current if block that sync and calls checkQueueSizeConstraintCore.

geode-core/src/main/java/org/apache/geode/internal/cache/ha/HARegionQueue.java
```diff
@@ -2251,48 +2237,54 @@ void checkQueueSizeConstraint() throws InterruptedException {
...
+          synchronized (this.permitMon) {
```
dschneider-pivotal, line 2228: since checkQueueSizeConstraintCore does a wait on permitMon, I think it would be better for the sync on permitMon to also be in checkQueueSizeConstraintCore. I would also move all the time calls into checkQueueSizeConstraintCore. Maybe change its name to waitForPermission since we only call it when we run out of permits.

geode-core/src/main/java/org/apache/geode/internal/cache/ha/HARegionQueue.java
```diff
@@ -2302,9 +2294,9 @@ void checkQueueSizeConstraint() throws InterruptedException {
      * @return int current Put permits
      */
     private int reconcilePutPermits() {
-      putPermits += takeSidePutPermits;
-      takeSidePutPermits = 0;
-      return putPermits;
+      putPermits.addAndGet(takeSidePutPermits.get());
```
dschneider-pivotal, line 2284: I think you have a concurrency bug between takeSidePutPermits.get() and the next line takeSidePutPermits.set(0).
If some other thread incs takeSidePutPermits between these two statements then the inc is lost.
I think you want to rewrite this code like so:

int takes = takeSidePutPermits.getAndSet(0);
return putPermits.addAndGet(takes);

geode-core/src/main/java/org/apache/geode/internal/cache/ha/HARegionQueue.java
```diff
@@ -2318,7 +2310,7 @@ private int reconcilePutPermits() {
     @Override
     void incrementTakeSidePutPermitsWithoutNotify() {
       synchronized (this.permitMon) {
```
dschneider-pivotal, line 2299: look into if this sync is still needed. It was before because all writes of takeSidePutPermits did it while synced on permitMon. But now that you have made takeSidePutPermits atomic and this method does not want to notify it might mean that the sync is no longer needed.

geode-core/src/main/java/org/apache/geode/internal/cache/ha/HARegionQueue.java
mhansonp: If I use < 0 this code will fail every single time. It prefers <= 0 so to fail early it would be if (reconcilePutPermits() > 0) { 
return
}
My observations show reconcilePutPermits returns 0, -1, or -2.

geode-core/src/main/java/org/apache/geode/internal/cache/ha/HARegionQueue.java
mhansonp: Yup.

geode-core/src/main/java/org/apache/geode/internal/cache/ha/HARegionQueue.java
mhansonp: Done.

geode-core/src/main/java/org/apache/geode/internal/cache/ha/HARegionQueue.java
mhansonp: Done

geode-core/src/main/java/org/apache/geode/internal/cache/ha/HARegionQueue.java
mhansonp: Done

mhansonp: Since the call of core was pretty thin, I inlined it again. I then renamed the whole function to waitForPermission.

geode-core/src/main/java/org/apache/geode/internal/cache/ha/HARegionQueue.java
mhansonp: The permitmon is what provides the waiting mechanism when the queue is full at 20,000 without it,  it overfills the hrq to 100,000 when the limit is set to 10,000. And the hrq.region.size goes to 200,000 when it should be 20,000. I think we should keep permit mon. All of the other changes look good though.

mhansonp:
Please see https://github.com/apache/geode/pull/6858 I have created a PR off develop.

PR #6765, opened 2021-08-16
GEODE-9514: optimize how Coder reads and writes "int" and "long" values

dschneider-pivotal:
Added convertLongToAsciiDigits to Coder so that ints and longs can more efficiently be converted to bytes. The goal was to reduce
the number of short lived object allocations based on this observation:
  Refactor the getIntegerResponse(ByteBuf buffer, long l) method to write byte string representation to avoid allocating transient char[], String, and byte[]. The existing allocations accounted for 17% of objects requiring GC.

So now when converting an int or long we never convert it to a String and for some values (-99..99) we don't allocate a new byte array. The same is also true when converting a sequence of bytes from command input into a long; we no longer allocate a short lived String.

### For all changes:

### Note:

### DIFF
--- a/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/netty/Coder.java
+++ b/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/netty/Coder.java
@@ -35,6 +35,7 @@ import static org.apache.geode.redis.internal.netty.StringBytesGlossary.bINF;
 import static org.apache.geode.redis.internal.netty.StringBytesGlossary.bINFINITY;
 import static org.apache.geode.redis.internal.netty.StringBytesGlossary.bLOWERCASE_A;
 import static org.apache.geode.redis.internal.netty.StringBytesGlossary.bLOWERCASE_Z;
+import static org.apache.geode.redis.internal.netty.StringBytesGlossary.bMINUS;
 import static org.apache.geode.redis.internal.netty.StringBytesGlossary.bMOVED;
 import static org.apache.geode.redis.internal.netty.StringBytesGlossary.bNIL;
 import static org.apache.geode.redis.internal.netty.StringBytesGlossary.bN_INF;
@@ -156,7 +157,7 @@ public class Coder {
   public static ByteBuf getScanResponse(ByteBuf buffer, BigInteger cursor,
       List<?> scanResult) {
     buffer.writeByte(ARRAY_ID);
-    buffer.writeByte('2');
+    buffer.writeByte(digitToAscii(2));
     buffer.writeBytes(bCRLF);
     byte[] cursorBytes = stringToBytes(cursor.toString());
     writeStringResponse(buffer, cursorBytes);
@@ -480,7 +481,7 @@ public class Coder {
    */
   public static void appendAsciiDigitsToByteBuf(long value, ByteBuf buf) {
     if (value < 0) {
-      buf.writeByte('-');
+      buf.writeByte(bMINUS);
     } else {
       value = -value;
     }
@@ -498,9 +499,9 @@ public class Coder {
     int q = value / 10;
     int r = (q * 10) - value;
     if (q < 0) {
-      buf.writeByte((byte) ('0' - q));
+      buf.writeByte(digitToAscii(-q));
     }
-    buf.writeByte((byte) ('0' + r));
+    buf.writeByte(digitToAscii(r));
   }
 
   private static void appendLargeAsciiDigitsToByteBuf(long value, ByteBuf buf) {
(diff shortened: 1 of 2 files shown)


geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/netty/Coder.java
```diff
@@ -155,12 +156,12 @@ private static void writeCollectionOrString(ByteBuf buffer, Object next) throws
   public static ByteBuf getScanResponse(ByteBuf buffer, BigInteger cursor,
       List<?> scanResult) {
     buffer.writeByte(ARRAY_ID);
-    buffer.writeBytes(intToBytes(2));
+    buffer.writeByte('2');
```
DonalEvans, line 159: Per a comment at the top of `StringBytesGlossary` (which was originally in this class before being pulled out):
```
 * Important note
 * <p>
 * Do not use '' <-- java primitive chars. Redis uses {@link Coder#CHARSET} encoding so we should
 * not risk java handling char to byte conversions, rather just hard code {@link Coder#CHARSET}
 * chars as bytes
 ```
Do we still want to follow this? If so, I think the solution is to introduce a `byte` constant to the `StringBytesGlossary` class for this char. We already have constants for 0 and 1, so this wouldn't be too out of the ordinary.

File: geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/netty/Coder.java [hunk 2]
```diff
@@ -471,4 +472,101 @@ public static int narrowLongToInt(long toBeNarrowed) {
...
   }
+
+  /**
+   * Takes the given "value" and computes the sequence of ASCII digits it represents,
+   * appending them to the given "buf".
+   * This code was adapted from the openjdk Long.java getChars methods.
+   */
+  public static void appendAsciiDigitsToByteBuf(long value, ByteBuf buf) {
+    if (value < 0) {
+      buf.writeByte('-');
```
DonalEvans, line 483: This char can be replaced with the `bMINUS` constant.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/netty/Coder.java
```diff
@@ -471,4 +472,101 @@ public static int narrowLongToInt(long toBeNarrowed) {
...
+    }
+  }
+
+  private static void appendSmallAsciiDigitsToByteBuf(int value, ByteBuf buf) {
+    int q = value / 10;
+    int r = (q * 10) - value;
+    if (q < 0) {
+      buf.writeByte((byte) ('0' - q));
+    }
+    buf.writeByte((byte) ('0' + r));
```
DonalEvans, line 503: These zeroes can be replaced with the `NUMBER_0_BYTE` constant.

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/netty/CoderTest.java
```diff
@@ -176,4 +180,25 @@ public void stripTrailingZeroFromDouble_correctlyStripsTrailingZero(byte[] input
...
+    }
+  }
+
+  private void verify_appendAsciiDigitsToByteBuf(long value) {
+    String expected = Long.toString(value);
+    ByteBuf buf = ByteBufAllocator.DEFAULT.heapBuffer();
+
+    Coder.appendAsciiDigitsToByteBuf(value, buf);
+
+    assertThat(buf.toString(Charset.defaultCharset())).isEqualTo(expected);
```
DonalEvans, line 202: I think we might want to explicitly use UTF-8 as the charset here, as that's what Redis and the Coder class use.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/netty/Coder.java
```diff
@@ -471,4 +472,101 @@ public static int narrowLongToInt(long toBeNarrowed) {
...
+    }
+
+    // We know there are at most two digits left at this point.
+    q2 = i2 / 10;
+    r = (q2 * 10) - i2;
+    bytes[--charPos] = (byte) ('0' + r);
+
+    // Whatever left is the remaining digit.
+    if (q2 < 0) {
+      bytes[--charPos] = (byte) ('0' - q2);
```
DonalEvans, line 540: These zeroes can be replaced with the `NUMBER_0_BYTE` constant.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/netty/Coder.java
```diff
@@ -471,4 +472,101 @@ public static int narrowLongToInt(long toBeNarrowed) {
...
+    }
+    buf.writeBytes(bytes, charPos, MAX_DIGITS - charPos);
+  }
+
+  @Immutable
+  private static final byte[] DIGIT_TENS = {
```
DonalEvans, line 546: If we want to follow the "important note" about not using chars mentioned above, then these arrays will need to use the UTF-8 byte values associated with these chars, which will probably make it look horrible.

DonalEvans (changes requested):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/netty/Coder.java
dschneider-pivotal: these static byte arrays came straight from the jdk so our old code was already using them when it called String.valueOf(long) which called Long.toString(long) which ended up in this code. Even the other code you commented on was for the most part from the jdk and was using '0' and '-'. So either the old code was already broken or the new code is okay.

dschneider-pivotal: Maybe what made the old code safe is that after the jdk converted a long to a String we then called stringToBytes which converted to String chars to the hardcoded UTF-8 charset. In this new code that is no longer happening so if they are using a default charset that does not map '0' to the ASCII code for 0 then this code will be broken. I'll convert it over to using the constants.

DonalEvans (approved):
Looks good. One small suggestion, but it's only cosmetic. For readability, could `q` and `r` be `quotient` and `remainder`, and `q2` and `i2` be `intQuotient` and `intValue` in `appendSmallAsciiDigitsToByteBuf()` and `appendLargeAsciiDigitsToByteBuf()`?

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/netty/Coder.java
dschneider-pivotal: Thanks for pointing this out. It was helpful and I think it is a good idea. I have switch all the code over to using these byte constants instead of java char literals

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/netty/Coder.java
dschneider-pivotal: done

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/netty/CoderTest.java
dschneider-pivotal: done

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/netty/Coder.java
dschneider-pivotal: done

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/netty/Coder.java
dschneider-pivotal: these static arrays are now initialized with some static code using the byte constants.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/netty/Coder.java
dschneider-pivotal: done

jdeppe-pivotal (approved):
After our offline conversation, I think I understand some of the tradeoffs better. Thanks for doing this work!

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/netty/Coder.java
```diff
@@ -332,9 +343,55 @@ public static BigDecimal bytesToBigDecimal(byte[] bytes) {
...
+    }
+    int i = 0;
+    long limit = -Long.MAX_VALUE;
+    boolean negative = false;
+    byte firstByte = bytes[0];
+    if (firstByte < NUMBER_0_BYTE) { // Possible leading "+" or "-"
+      if (firstByte == bMINUS) {
+        negative = true;
+        limit = Long.MIN_VALUE;
+      } else if (firstByte != bPLUS) {
```
DonalEvans, line 369: If we want to match Redis' behaviour exactly, we should also throw an exception if we're given a number that starts with "+" or with "-0". These could be done as separate tickets though, if you feel it's not worth including them here. In fact, we already have [GEODE-9482](https://issues.apache.org/jira/browse/GEODE-9482) to track the second case.

jdeppe-pivotal:
I like these most recent changes.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/netty/Coder.java
dschneider-pivotal: lets do that on another ticket. It does seem we might get some other benefits of having our own code to do the conversion instead of just using the jdk.

PR #6770, opened 2021-08-18
GEODE-9515: Skip setting MBeanServer when JMXConnectorServer will set it

jchen21:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/management/internal/ManagementAgent.java
+++ b/geode-core/src/main/java/org/apache/geode/management/internal/ManagementAgent.java
@@ -52,6 +52,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.Logger;
 
 import org.apache.geode.GemFireConfigException;
+import org.apache.geode.annotations.VisibleForTesting;
 import org.apache.geode.cache.internal.HttpService;
 import org.apache.geode.distributed.internal.DistributionConfig;
 import org.apache.geode.internal.GemFireVersion;
@@ -429,6 +430,19 @@ public class ManagementAgent {
           }
         };
 
+    setMBeanServerForwarder(mbs, env);
+    registerAccessControlMBean();
+    registerFileUploaderMBean();
+
+    jmxConnectorServer.start();
+    if (logger.isDebugEnabled()) {
+      logger.debug("Finished starting jmx manager agent.");
+    }
+  }
+
+  @VisibleForTesting
+  void setMBeanServerForwarder(MBeanServer mbs, HashMap<String, Object> env)
+      throws IOException {
     if (securityService.isIntegratedSecurity()) {
       shiroAuthenticator = new JMXShiroAuthenticator(this.securityService);
       env.put(JMXConnectorServer.AUTHENTICATOR, shiroAuthenticator);
@@ -457,13 +471,6 @@ public class ManagementAgent {
         // if no access control, do not allow mbean creation to prevent Mlet attack
         jmxConnectorServer.setMBeanServerForwarder(new BlockMBeanCreationController());
       }
-    }
-    registerAccessControlMBean();
-    registerFileUploaderMBean();
-
-    jmxConnectorServer.start();
-    if (logger.isDebugEnabled()) {
-      logger.debug("Finished starting jmx manager agent.");
     }
   }
 
@@ -508,6 +515,10 @@ public class ManagementAgent {
 
   public JMXConnectorServer getJmxConnectorServer() {
     return jmxConnectorServer;
+  }
+
+  public void setJmxConnectorServer(JMXConnectorServer jmxConnectorServer) {
+    this.jmxConnectorServer = jmxConnectorServer;
   }
 
   public synchronized RemoteStreamExporter getRemoteStreamExporter() {
--- a/geode-core/src/main/java/org/apache/geode/management/internal/ManagementAgent.java
+++ b/geode-core/src/main/java/org/apache/geode/management/internal/ManagementAgent.java
@@ -449,7 +449,9 @@ public class ManagementAgent {
       if (accessFile != null && accessFile.length() > 0) {
         // Rewire the mbs hierarchy to set accessController
         ReadOpFileAccessController controller = new ReadOpFileAccessController(accessFile);
-        controller.setMBeanServer(mbs);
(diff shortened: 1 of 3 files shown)


geode-core/src/test/java/org/apache/geode/management/internal/ManagementAgentTest.java
```diff
@@ -0,0 +1,123 @@
...
+
+import org.apache.geode.distributed.internal.DistributionConfig;
+import org.apache.geode.internal.cache.InternalCache;
+import org.apache.geode.internal.security.SecurityService;
+
+public class ManagementAgentTest {
+
+  // private ManagementAgent managementAgent = spy(ManagementAgent.class);
+  @Rule
+  public TemporaryFolder temporaryFolder = new TemporaryFolder();
```
kirklund, line 43: Let's move this test to `src/integrationTest` and rename to `ManagementAgentIntegrationTest` since it's touching the file system.

geode-core/src/test/java/org/apache/geode/management/internal/ManagementAgentTest.java
```diff
@@ -0,0 +1,123 @@
...
+    JmxRmiSerialFilter serialFilter = mock(JmxRmiSerialFilter.class);
+    SecurityService securityService = mock(SecurityService.class);
+    when(internalCache.getSecurityService()).thenReturn(securityService);
+    when(securityService.isIntegratedSecurity()).thenReturn(false);
+    File tempFile = temporaryFolder.newFile("testFile");
+    when(distributionConfig.getJmxManagerAccessFile()).thenReturn(tempFile.getCanonicalPath());
+    ManagementAgent managementAgent =
+        new ManagementAgent(distributionConfig, internalCache, serialFilter);
+    MBeanServer mBeanServer = mock(MBeanServer.class);
+    JMXConnectorServer jmxConnectorServerWithMBeanServer = new JMXConnectorServer(mBeanServer) {
```
kirklund, line 58: I think you should move this to a static inner-class or better yet use Mockito to mock it:
```
JMXConnectorServer jmxConnectorServerWithMBeanServer = mock(JMXConnectorServer.class);
```

kirklund (changes requested):

geode-core/src/main/java/org/apache/geode/management/internal/ManagementAgent.java
```diff
@@ -449,20 +463,15 @@ public synchronized void start() throws IOException {
       if (accessFile != null && accessFile.length() > 0) {
         // Rewire the mbs hierarchy to set accessController
         ReadOpFileAccessController controller = new ReadOpFileAccessController(accessFile);
-        controller.setMBeanServer(mbs);
+        if (jmxConnectorServer.getMBeanServer() == null) {
```
jinmeiliao, line 466: Looks like besides re-arranging code, this is the only effective change here. Do we know whey in this case this mbeanServer is not null?

geode-core/src/main/java/org/apache/geode/management/internal/ManagementAgent.java
```diff
@@ -508,6 +517,10 @@ public JMXConnectorServer getJmxConnectorServer() {
     return jmxConnectorServer;
   }
 
+  public void setJmxConnectorServer(JMXConnectorServer jmxConnectorServer) {
+    this.jmxConnectorServer = jmxConnectorServer;
+  }
+
```
kirklund, line 523: Should this method be `private` instead of `public`?

jchen21: This method is added for the test. So it should be non-private. We can remove the `public` scope.

geode-core/src/test/java/org/apache/geode/management/internal/ManagementAgentTest.java
jchen21: Agree.

geode-core/src/test/java/org/apache/geode/management/internal/ManagementAgentTest.java
jchen21: I need two instances of subclass of `JMXConnectorServer`, which an abstract class. One instance with a non-null `mbeanServer` field, the other with a null `mbeanServer`. I don't know why move this to a static inner-class. I have tried to mock `JMXConnectorServer`. But I don't think a mock will work in this test case. I need to set the `mbeanServer` field of `JMXConnectorServer`. There is no setter for it. Only the constructor sets it.

geode-core/src/main/java/org/apache/geode/management/internal/ManagementAgent.java
jchen21: If you look at the implementation of `JMXConnectorServer. setMBeanServerForwarder()`, if the MBeanServer of the connector server is non-null, it will set the MBeanServer for the forwarder, which is the `controller` in our case. We want to avoid setting the MBeanServer twice for the controller, otherwise it will throw `IllegalArgumentException`. 

Here is the code  of `JMXConnectorServer. setMBeanServerForwarder()`:
```
    public synchronized void setMBeanServerForwarder(MBeanServerForwarder mbsf)
    {
        if (mbsf == null)
            throw new IllegalArgumentException("Invalid null argument: mbsf");

        if (mbeanServer !=  null) mbsf.setMBeanServer(mbeanServer);
        mbeanServer = mbsf;
    }
```

jinmeiliao: We know at this point `JMXConnectorServer`'s mbeanServer is definitely not null, so inside the call `setMBeanServerForwarder`, it will call `mbsf.setMBeanServer(mbeanServer)` anyway, should we just delete line 452 then?

jinmeiliao: I looked at the history of the file, the call `jmxConnectorServer.setMBeanServerForwarder(controller);` was added recently. When that call is added, we don't need the previous line `controller.setMBeanServer(mbs);` anymore.

jchen21: Agree.

jinmeiliao (approved):

kirklund (approved):

PR #6774, opened 2021-08-19
GEODE-9523: Fix invalid path in ContainerInstall

demery-pivotal:
PROBLEM

`ContainerInstall` fetches a resource URL, then applies a chain of
conversions before passing the resulting URL to the `ZipURLInstaller`
constructor.  One of the conversions attempts to create a `Path` from a
string.  On Windows, the string value is not a valid path, and
`Paths.get(...)` throws an exception.

SOLUTION

Change `ContainerInstall` to remove all of the conversions. The original
resource URL is acceptable to `ZipURLInstaller` without any conversion.


BenjaminPerryRoss (approved):
LGTM

PR #6776, opened 2021-08-19
Create ThreadLocal instance, use to manage DeltaInfos

ringles:
Using a static ThreadLocal in the RedisData class, we can avoid carrying around an object reference for all instances of RedisData. This saves 4-8 bytes of memory per instance.

### DIFF
--- a/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/AbstractRedisData.java
+++ b/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/AbstractRedisData.java
@@ -60,7 +60,7 @@ public abstract class AbstractRedisData implements RedisData {
    * (i.e. the number of milliseconds since midnight, Jan 1, 1970)
    */
   private volatile long expirationTimestamp = NO_EXPIRATION;
-  private transient DeltaInfo deltaInfo;
+  private static final ThreadLocal<DeltaInfo> deltaInfo = new ThreadLocal<>();
 
   @Override
   public void setExpirationTimestamp(Region<RedisKey, RedisData> region, RedisKey key, long value) {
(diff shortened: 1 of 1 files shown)


DonalEvans:
Is there a JIRA ticket associated with this change? If so, please include it in the PR title, and if not, please open one.

jdeppe-pivotal (approved):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/AbstractRedisData.java
```diff
@@ -168,18 +168,19 @@ public void fromData(DataInput in, DeserializationContext context)
...
   public boolean hasDelta() {
-    return deltaInfo != null;
+    return deltaInfo.get() != null;
   }
 
   @Override
   public void toDelta(DataOutput out) throws IOException {
-    deltaInfo.serializeTo(out);
+    deltaInfo.get().serializeTo(out);
+    deltaInfo.remove();
```
dschneider-pivotal, line 183: don't call remove here. It you look at storeChanges you will see it already calls setDelta(null) in a finally block. By calling ThreadLocal.remove here you just cause setDelta(null) to recreate an entry in the thread local map.
I think we should not call ThreadLocal.remove at all and the current call of setDelta(null) on storeChanges is all we need

dschneider-pivotal (changes requested):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/AbstractRedisData.java
ringles: It's gone!

dschneider-pivotal (approved):
If you were wondering why this change of getting rid of an object reference field in AbstractRedisData did not cause any of our sizing tests to fail, I think it would be because the jvm rounds the size of an instance to a multiple of 8. So a 4 byte obj ref can fit into this extra memory. If our sizing tests were run with compressed oops disabled then I'd expect it not to fit and you would have needed to reduce the expected size.

ringles:
Stores deltas for RedisData in a static ThreadLocal in the RedisData class, which avoids carrying around an object reference for all instances of RedisData. This saves 4-8 bytes of memory per RedisData instance.

PR #6783, opened 2021-08-20
GEODE-9446: Remove unecessary byte arrays from RedisSortedSet

nonbinaryprogrammer:
RedisSortedSet previously stored the score as both a double and a byte
array in order to cut down on conversions using Coder. This refactor
gets rid of the byte array score and instead does a couple of
conversions. Other refactors were done to accomodate this change and
to improve on what was there

### For all changes:

### Note:

### DIFF
--- a/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/AbstractRedisData.java
+++ b/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/AbstractRedisData.java
@@ -25,6 +25,7 @@ import java.io.DataOutput;
 import java.io.DataOutputStream;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.List;
 import java.util.Objects;
 
 import org.apache.logging.log4j.Logger;
@@ -33,6 +34,7 @@ import org.apache.geode.DataSerializer;
 import org.apache.geode.InvalidDeltaException;
 import org.apache.geode.cache.EntryNotFoundException;
 import org.apache.geode.cache.Region;
+import org.apache.geode.internal.InternalDataSerializer;
 import org.apache.geode.internal.cache.BucketRegion;
 import org.apache.geode.internal.serialization.DeserializationContext;
 import org.apache.geode.internal.serialization.KnownVersion;
@@ -45,6 +47,7 @@ import org.apache.geode.redis.internal.delta.DeltaInfo;
 import org.apache.geode.redis.internal.delta.DeltaType;
 import org.apache.geode.redis.internal.delta.RemsDeltaInfo;
 import org.apache.geode.redis.internal.delta.TimestampDeltaInfo;
+import org.apache.geode.redis.internal.delta.ZaddsDeltaInfo;
 
 public abstract class AbstractRedisData implements RedisData {
   private static final BucketRegion.PrimaryMoveReadLockAcquired primaryMoveReadLockAcquired =
@@ -200,6 +203,15 @@ public abstract class AbstractRedisData implements RedisData {
         byte[] byteArray = DataSerializer.readByteArray(in);
         applyDelta(new AppendDeltaInfo(byteArray, sequence));
         break;
+      case ZADDS:
+        int numMembers = InternalDataSerializer.readArrayLength(in);
+        List<byte[]> members = new ArrayList<>();
+        List<Double> scores = new ArrayList<>();
+        for (int i = 0; i < numMembers; i++) {
+          members.add(DataSerializer.readByteArray(in));
+          scores.add(DataSerializer.readDouble(in));
+        }
+        applyDelta(new ZaddsDeltaInfo(members, scores));
     }
   }
 
--- a/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/AbstractRedisData.java
+++ b/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/AbstractRedisData.java
@@ -47,7 +47,7 @@ import org.apache.geode.redis.internal.delta.DeltaInfo;
 import org.apache.geode.redis.internal.delta.DeltaType;
 import org.apache.geode.redis.internal.delta.RemsDeltaInfo;
 import org.apache.geode.redis.internal.delta.TimestampDeltaInfo;
-import org.apache.geode.redis.internal.delta.ZaddsDeltaInfo;
+import org.apache.geode.redis.internal.delta.ZAddsDeltaInfo;
 
 public abstract class AbstractRedisData implements RedisData {
   private static final BucketRegion.PrimaryMoveReadLockAcquired primaryMoveReadLockAcquired =
@@ -204,14 +204,14 @@ public abstract class AbstractRedisData implements RedisData {
         applyDelta(new AppendDeltaInfo(byteArray, sequence));
         break;
       case ZADDS:
-        int numMembers = InternalDataSerializer.readArrayLength(in);
+        int numMembers = InternalDataSerializer.readPrimitiveInt(in);
         List<byte[]> members = new ArrayList<>();
         List<Double> scores = new ArrayList<>();
         for (int i = 0; i < numMembers; i++) {
           members.add(DataSerializer.readByteArray(in));
           scores.add(DataSerializer.readDouble(in));
         }
-        applyDelta(new ZaddsDeltaInfo(members, scores));
+        applyDelta(new ZAddsDeltaInfo(members, scores));
     }
   }
 
--- a/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/AbstractRedisData.java
+++ b/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/AbstractRedisData.java
@@ -34,7 +34,6 @@ import org.apache.geode.DataSerializer;
 import org.apache.geode.InvalidDeltaException;
 import org.apache.geode.cache.EntryNotFoundException;
 import org.apache.geode.cache.Region;
-import org.apache.geode.internal.InternalDataSerializer;
 import org.apache.geode.internal.cache.BucketRegion;
 import org.apache.geode.internal.serialization.DeserializationContext;
 import org.apache.geode.internal.serialization.KnownVersion;
@@ -204,7 +203,7 @@ public abstract class AbstractRedisData implements RedisData {
         applyDelta(new AppendDeltaInfo(byteArray, sequence));
         break;
       case ZADDS:
-        int numMembers = InternalDataSerializer.readPrimitiveInt(in);
+        int numMembers = DataSerializer.readPrimitiveInt(in);
         List<byte[]> members = new ArrayList<>();
         List<Double> scores = new ArrayList<>();
         for (int i = 0; i < numMembers; i++) {
--- a/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/AbstractRedisData.java
+++ b/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/AbstractRedisData.java
@@ -25,7 +25,6 @@ import java.io.DataOutput;
 import java.io.DataOutputStream;
 import java.io.IOException;
 import java.util.ArrayList;
-import java.util.List;
 import java.util.Objects;
 
 import org.apache.logging.log4j.Logger;
@@ -204,13 +203,13 @@ public abstract class AbstractRedisData implements RedisData {
         break;
       case ZADDS:
         int numMembers = DataSerializer.readPrimitiveInt(in);
-        List<byte[]> members = new ArrayList<>(numMembers);
-        List<Double> scores = new ArrayList<>(numMembers);
+        ZAddsDeltaInfo delta = new ZAddsDeltaInfo(numMembers);
         for (int i = 0; i < numMembers; i++) {
-          members.add(DataSerializer.readByteArray(in));
-          scores.add(DataSerializer.readDouble(in));
+          byte[] member = DataSerializer.readByteArray(in);
+          double score = DataSerializer.readPrimitiveDouble(in);
+          delta.add(member, score);
         }
-        applyDelta(new ZAddsDeltaInfo(members, scores));
+        applyDelta(delta);
     }
   }
 
--- a/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/AbstractRedisData.java
+++ b/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/AbstractRedisData.java
@@ -204,8 +204,8 @@ public abstract class AbstractRedisData implements RedisData {
         break;
       case ZADDS:
         int numMembers = DataSerializer.readPrimitiveInt(in);
-        List<byte[]> members = new ArrayList<>();
-        List<Double> scores = new ArrayList<>();
+        List<byte[]> members = new ArrayList<>(numMembers);
+        List<Double> scores = new ArrayList<>(numMembers);
         for (int i = 0; i < numMembers; i++) {
           members.add(DataSerializer.readByteArray(in));
           scores.add(DataSerializer.readDouble(in));
--- a/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
+++ b/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
@@ -67,11 +67,11 @@ public class RedisSortedSet extends AbstractRedisData {
     return REDIS_SORTED_SET_OVERHEAD + members.getSizeInBytes() + scoreSet.getSizeInBytes();
   }
 
-  RedisSortedSet(List<byte[]> members, List<Double> scores) {
+  RedisSortedSet(List<byte[]> members, double[] scores) {
     this.members = new MemberMap(members.size());
 
     for (int i = 0; i < members.size(); i++) {
-      double score = scores.get(i);
+      double score = scores[i];
       byte[] member = members.get(i);
       memberAdd(member, score);
     }
@@ -201,10 +201,10 @@ public class RedisSortedSet extends AbstractRedisData {
    * @return the number of members actually added OR incremented value if INCR option specified
    */
   Object zadd(Region<RedisKey, RedisData> region, RedisKey key, List<byte[]> membersToAdd,
-      List<Double> scoresToAdd, ZAddOptions options) {
+      double[] scoresToAdd, ZAddOptions options) {
     if (options.isINCR()) {
       // if INCR, only one score and member can be added
-      return zaddIncr(region, key, membersToAdd.get(0), scoresToAdd.get(0), options);
+      return zaddIncr(region, key, membersToAdd.get(0), scoresToAdd[0], options);
     }
 
     ZAddsDeltaInfo deltaInfo = null;
@@ -212,7 +212,7 @@ public class RedisSortedSet extends AbstractRedisData {
     int changesCount = 0;
 
     for (int i = 0; i < membersToAdd.size(); i++) {
-      double score = scoresToAdd.get(i);
+      double score = scoresToAdd[i];
       byte[] member = membersToAdd.get(i);
       if (options.isNX() && members.containsKey(member)) {
         continue;
@@ -274,6 +274,7 @@ public class RedisSortedSet extends AbstractRedisData {
 
     if (!(memberAdd(member, score) == MemberAddResult.NO_OP)) {
       ZAddsDeltaInfo deltaInfo = new ZAddsDeltaInfo(1);
+      deltaInfo.add(member, score);
       storeChanges(region, key, deltaInfo);
     }
 
--- a/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
+++ b/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
@@ -229,15 +229,13 @@ public class RedisSortedSet extends AbstractRedisData {
         continue;
       }
       boolean scoreChanged = memberAdd(member, score);
-      if (scoreChanged) {
-        if (options.isCH()) {
-          changesCount++;
-        }
-        if (deltaInfo == null) {
-          deltaInfo = new ZAddsDeltaInfo(member, score);
-        } else {
-          deltaInfo.add(member, score);
-        }
+      if (options.isCH() && scoreChanged) {
+        changesCount++;
+      }
+      if (deltaInfo == null) {
+        deltaInfo = new ZAddsDeltaInfo(member, score);
+      } else {
+        deltaInfo.add(member, score);
       }
     }
 
--- a/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
+++ b/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
@@ -21,9 +21,6 @@ import static org.apache.geode.internal.JvmSizeUtils.memoryOverhead;
 import static org.apache.geode.redis.internal.RedisConstants.ERROR_NOT_A_VALID_FLOAT;
 import static org.apache.geode.redis.internal.data.NullRedisDataStructures.NULL_REDIS_SORTED_SET;
 import static org.apache.geode.redis.internal.data.RedisDataType.REDIS_SORTED_SET;
-import static org.apache.geode.redis.internal.netty.Coder.bytesToDouble;
-import static org.apache.geode.redis.internal.netty.Coder.doubleToBytes;
-import static org.apache.geode.redis.internal.netty.Coder.stripTrailingZeroFromDouble;
 import static org.apache.geode.redis.internal.netty.StringBytesGlossary.bGREATEST_MEMBER_NAME;
 import static org.apache.geode.redis.internal.netty.StringBytesGlossary.bLEAST_MEMBER_NAME;
 import static org.apache.geode.redis.internal.netty.StringBytesGlossary.bZERO;
@@ -32,7 +29,6 @@ import java.io.DataInput;
 import java.io.DataOutput;
 import java.io.IOException;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
@@ -53,11 +49,13 @@ import org.apache.geode.redis.internal.collections.SizeableBytes2ObjectOpenCusto
 import org.apache.geode.redis.internal.delta.AddsDeltaInfo;
 import org.apache.geode.redis.internal.delta.DeltaInfo;
 import org.apache.geode.redis.internal.delta.RemsDeltaInfo;
+import org.apache.geode.redis.internal.delta.ZAddsDeltaInfo;
 import org.apache.geode.redis.internal.executor.sortedset.AbstractSortedSetRangeOptions;
 import org.apache.geode.redis.internal.executor.sortedset.SortedSetLexRangeOptions;
 import org.apache.geode.redis.internal.executor.sortedset.SortedSetRankRangeOptions;
 import org.apache.geode.redis.internal.executor.sortedset.SortedSetScoreRangeOptions;
 import org.apache.geode.redis.internal.executor.sortedset.ZAddOptions;
+import org.apache.geode.redis.internal.netty.Coder;
 import org.apache.geode.redis.internal.executor.sortedset.ZAggregator;
 import org.apache.geode.redis.internal.executor.sortedset.ZKeyWeight;
 import org.apache.geode.redis.internal.netty.Coder;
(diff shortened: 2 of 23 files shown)


geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/delta/ZaddsDeltaInfo.java
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  public void add(byte[] delta, double score) {
+    deltas.add(delta);
+    this.scores.add(score);
+  }
+
+  public void serializeTo(DataOutput out) throws IOException {
+    InternalDataSerializer.writeEnum(ZADDS, out);
+    InternalDataSerializer.writeArrayLength(deltas.size(), out);
```
jdeppe-pivotal, line 52: We should just be using `DataSerializer` and not the internal version. Instead of `writeArrayLength` you can do `DataSerializer.writePrimitiveInt(deltas.size(), out)`

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/AbstractRedisData.java
```diff
@@ -200,6 +203,15 @@ public void fromDelta(DataInput in) throws IOException, InvalidDeltaException {
         byte[] byteArray = DataSerializer.readByteArray(in);
         applyDelta(new AppendDeltaInfo(byteArray, sequence));
         break;
+      case ZADDS:
+        int numMembers = InternalDataSerializer.readArrayLength(in);
+        List<byte[]> members = new ArrayList<>();
+        List<Double> scores = new ArrayList<>();
```
jdeppe-pivotal, line 209: Use `DataSerializer.readPrimitiveInt` (see more comments from `ZAddsDeltaInfo`). Also, this value can then be used to initialize the 2 arrays.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/delta/ZaddsDeltaInfo.java
```diff
@@ -0,0 +1,66 @@
...
+public class ZaddsDeltaInfo implements DeltaInfo {
+  private final ArrayList<byte[]> deltas = new ArrayList<>();
+  private final ArrayList<Double> scores = new ArrayList<>();
+
+  public ZaddsDeltaInfo() {}
+
+  public ZaddsDeltaInfo(List<byte[]> deltas, List<Double> scores) {
+    this.deltas.addAll(deltas);
+    this.scores.addAll(scores);
+  }
```
jdeppe-pivotal, line 38: We should just use the references passed in here instead of copying everything again into our internal arrays.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/netty/Coder.java
```diff
@@ -404,9 +404,14 @@ public static double bytesToDouble(byte[] bytes) {
       return NEGATIVE_INFINITY;
     }
     if (isNaN(bytes)) {
-      return NaN;
+      throw new NumberFormatException();
```
jdeppe-pivotal, line 407: Since `Double.NaN` is a thing, it seems inconsistent that we're able to return `*INFINITY`s but throw an exception with `NaN`. Can this exception be thrown, at a higher level where `NaN` is determined to be invalid?

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -67,14 +65,19 @@ public int getSizeInBytes() {
...
   }
 
-  RedisSortedSet(List<byte[]> members) {
-    this.members = new MemberMap(members.size() / 2);
+  RedisSortedSet(List<byte[]> members, List<Double> scores) {
+    this.members = new MemberMap(members.size() + scores.size());
 
-    Iterator<byte[]> iterator = members.iterator();
+    if (members.size() != scores.size()) {
+      throw new NumberFormatException("there must be the same number of members and scores");
```
jdeppe-pivotal, line 72: I think `IllegalArgumentException` is more appropriate here.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -213,34 +232,42 @@ private synchronized void membersRemoveAll(RemsDeltaInfo remsDeltaInfo) {
...
-      ZAddOptions options) {
+      List<Double> scoresToAdd, ZAddOptions options) {
     if (options.isINCR()) {
-      return zaddIncr(region, key, membersToAdd, options);
+      // if INCR, only one score and member can be added
+      return zaddIncr(region, key, membersToAdd.get(0), scoresToAdd.get(0), options);
+    }
+
+    if (membersToAdd.size() != scoresToAdd.size()) {
+      throw new NumberFormatException("there must be the same number of members and scores");
```
jdeppe-pivotal, line 242: I think IllegalArgumentException is more appropriate here.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -213,34 +232,42 @@ private synchronized void membersRemoveAll(RemsDeltaInfo remsDeltaInfo) {
...
+    Iterator<byte[]> membersIterator = membersToAdd.iterator();
+    Iterator<Double> scoresIterator = scoresToAdd.iterator();
+
     int initialSize = scoreSet.size();
     int changesCount = 0;
-    while (iterator.hasNext()) {
-      byte[] score = iterator.next();
-      byte[] member = iterator.next();
+
+    while (membersIterator.hasNext()) {
```
jdeppe-pivotal, line 252: I think it's more idiomatic to just use a `for(...)` loop here. Previously we had an iterator because members and scores were interleaved in a single list.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -265,30 +292,27 @@ long zcount(SortedSetScoreRangeOptions rangeOptions) {
...
     OrderedSetEntry orderedSetEntry = members.get(member);
-    double incr = processByteArrayAsDouble(increment);
 
     if (orderedSetEntry != null) {
-      double byteScore = orderedSetEntry.getScore();
-      incr += byteScore;
+      score = orderedSetEntry.getScore() + incr;
 
-      if (Double.isNaN(incr)) {
+      if (Double.isNaN(score)) {
```
jdeppe-pivotal, line 302: Since we're not converting a byte array into a double anymore, I'd prefer to see this check removed. In general this type of validation should be occurring at a higher level - in the `Executor`.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/SortedSetScoreRangeOptions.java
```diff
@@ -48,16 +48,10 @@ void parseRangeArguments(List<byte[]> commandElements) {
         score = 0.0;
       } else {
         score = bytesToDouble(Arrays.copyOfRange(bytes, 1, bytes.length));
-        if (Double.isNaN(score)) {
-          throw new NumberFormatException();
-        }
```
jdeppe-pivotal, line 53: If you apply my comments in `Coder` this probably needs to be reverted.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/SortedSetScoreRangeOptions.java
```diff
@@ -48,16 +48,10 @@ void parseRangeArguments(List<byte[]> commandElements) {
...
       }
       return new RangeLimit<>(score, true);
     } else {
       score = bytesToDouble(bytes);
-      if (Double.isNaN(score)) {
```
jdeppe-pivotal, line 58: Ditto the above

jdeppe-pivotal (changes requested):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/delta/ZaddsDeltaInfo.java
nonbinaryprogrammer: I did that to make the fields final. plus even if we just do `this.deltas = deltas`, we still have to convert deltas to an arraylist

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
nonbinaryprogrammer: we can check in the executor if the increment is NaN and exit early, but that check in RedisSortedSet is still required. it handles the case where the score is -Infinity and the increment is +Infinity, which are both valid, but adding them produces NaN

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -433,18 +457,19 @@ private long iteratorRangeRemove(Iterator<AbstractOrderedSetEntry> scoresIterato
     return entriesRemoved;
   }
 
-  private byte[] zaddIncr(Region<RedisKey, RedisData> region, RedisKey key,
-      List<byte[]> membersToAdd, ZAddOptions options) {
+  private byte[] zaddIncr(Region<RedisKey, RedisData> region, RedisKey key, byte[] member,
+      double increment, ZAddOptions options) {
+    if (Double.isNaN(increment)) {
+      throw new NumberFormatException(ERROR_NOT_A_VALID_FLOAT);
+    }
```
DonalEvans, line 464: This check should be unnecessary, as the ZAddExecutor already checks for NaN increments (assuming the `Coder.bytesToDouble()` method keeps its NaN check that throws an Exception).

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -265,30 +292,27 @@ long zcount(SortedSetScoreRangeOptions rangeOptions) {
     return maxIndex - minIndex;
   }
 
-  byte[] zincrby(Region<RedisKey, RedisData> region, RedisKey key, byte[] increment,
-      byte[] member) {
+  double zincrby(Region<RedisKey, RedisData> region, RedisKey key, double incr, byte[] member) {
```
DonalEvans, line 295: For consistency, could either this method return a `byte[]`, or could other methods in this class that return a score also return it as `double`? The former is probably the most consistent and simplest, given we have range commands that return interleaved member and score `byte[]` if "WITHSCORES" is specified and it wouldn't be sensible to change those to return `double` rather than `byte[]` for the score now.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -648,10 +656,9 @@ public int compareScores(double score1, double score2) {
 
     public static final int ORDERED_SET_ENTRY_OVERHEAD = memoryOverhead(OrderedSetEntry.class);
 
-    public OrderedSetEntry(byte[] member, byte[] score) {
+    public OrderedSetEntry(byte[] member, double score) {
       this.member = member;
-      this.scoreBytes = stripTrailingZeroFromDouble(score);
```
DonalEvans, line 653: With the removal of this line, the `stripTrailingZeroFromDouble()` method is not used anywhere outside of tests, so it can be removed.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/delta/ZaddsDeltaInfo.java
```diff
@@ -0,0 +1,66 @@
...
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+import org.apache.geode.DataSerializer;
+import org.apache.geode.internal.InternalDataSerializer;
+
+public class ZaddsDeltaInfo implements DeltaInfo {
+  private final ArrayList<byte[]> deltas = new ArrayList<>();
+  private final ArrayList<Double> scores = new ArrayList<>();
```
DonalEvans, line 31: These should be declared as the interface `List<>` rather that the concrete class, since we're never calling `DataSerializer.writeArrayList()` like in `AddsDeltaInfo`, which is the only reason the field in that class need to be an `ArrayList<>`. Doing this allows the fields to be set in the constructor directly rather than using `addAll()`.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/delta/ZaddsDeltaInfo.java
```diff
@@ -0,0 +1,66 @@
...
+
+  public ZaddsDeltaInfo() {}
```
DonalEvans, line 33: This empty constructor is never used and can be removed, or reworked to have similar behaviour to the 0-arg constructor of `RemsDeltaInfo` or the `AddsDeltaInfo(int size)` constructor for `AddsDeltaInfo`.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/delta/ZaddsDeltaInfo.java
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  public ZaddsDeltaInfo(byte[] delta, Double score) {
+    this.deltas.add(delta);
+    this.scores.add(score);
+  }
+
+  public void add(byte[] delta, double score) {
+    deltas.add(delta);
+    this.scores.add(score);
```
DonalEvans, line 47: Inconsistent use of `this` here. It shouldn't be necessary in this method.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/delta/ZaddsDeltaInfo.java
```diff
@@ -0,0 +1,66 @@
...
+  private final ArrayList<Double> scores = new ArrayList<>();
+
+  public ZaddsDeltaInfo() {}
+
+  public ZaddsDeltaInfo(List<byte[]> deltas, List<Double> scores) {
+    this.deltas.addAll(deltas);
+    this.scores.addAll(scores);
+  }
+
+  public ZaddsDeltaInfo(byte[] delta, Double score) {
```
DonalEvans, line 40: Rather than this being its own method, which basically just reimplements `add()` it might be better to do what `AddsDeltaInfo` and `RemsDeltaInfo` do, and call a constructor that does not need data, then call `add()`.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/ZAddExecutor.java
```diff
@@ -89,6 +101,9 @@ private int findAndValidateZAddOptions(Command command, Iterator<byte[]> command
         optionsFoundCount++;
       } else if (isInfinity(subCommand)) {
         scoreFound = true;
+      } else if (isNaN(subCommand)) {
+        executorState.exceptionMessage = ERROR_NOT_A_VALID_FLOAT;
+        return 0;
```
DonalEvans, line 106: With the call to `Coder.bytesToDouble()` on line 60, this check shouldn't be needed, as the modified `bytesToDouble()` method will now throw an exception with the correct message when we get to that line. We would do a couple of unnecessary allocations if a user has passed NaN as a score with this removed, but we would avoid doing a byte array comparison on every argument in the more likely case where they haven't passed NaN, so it's a net performance improvement, I think.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/netty/Coder.java
DonalEvans: Internally, there is no situation in which NaN is ever a valid double value for us. If we ever get passed NaN by the user or would get NaN from a calculation, then we always return an error and never use the actual "NaN" value. As such, it makes sense to just hard-code it into our conversion method rather than adding checks for it all over the place, since there's no situation in which we would have NaN and be okay with it.

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/collections/SizeableObject2ObjectOpenCustomHashMapWithCursorTest.java
```diff
@@ -338,24 +338,22 @@ public void getSizeInBytesIsAccurateForOrderedSetEntryValues() {
...
+      double oldScore = value.getScore();
+      int scoreDelta =
+          memoryOverhead(Coder.doubleToBytes(i)) - memoryOverhead(Coder.doubleToBytes(oldScore));
 
       int oldSize = hash.getSizeInBytes();
-      value.updateScore(scoreBytes);
+      value.updateScore(i);
       int sizeDelta = hash.getSizeInBytes() - oldSize;
 
       assertThat(sizeDelta).isEqualTo(scoreDelta);
```
DonalEvans, line 359: The scoreDelta here should be 0, since `double` is fixed size and updating its value does not change its size. This section can then be refactored to:
```
      RedisSortedSet.OrderedSetEntry value = hash.get(key);

      int oldSize = hash.getSizeInBytes();
      value.updateScore(i);
      int sizeDelta = hash.getSizeInBytes() - oldSize;

      assertThat(sizeDelta).isZero();
```

DonalEvans (changes requested):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -213,34 +232,42 @@ private synchronized void membersRemoveAll(RemsDeltaInfo remsDeltaInfo) {
...
+      if (options.isCH() && scoreChanged) {
         changesCount++;
       }
 
       if (deltaInfo == null) {
-        deltaInfo = new AddsDeltaInfo(new ArrayList<>());
+        deltaInfo = new ZaddsDeltaInfo(member, score);
+      } else {
+        deltaInfo.add(member, score);
       }
```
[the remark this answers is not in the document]
DonalEvans: I realize now that this suggestion is incorrect, because we actually have three possible cases: a new entry was added, an existing entry was updated or an existing entry was not updated. We need to know if the second case is true when the `CH` option is selected, but if the third case is true, then we don't need to send a delta or update the region. It would be good if we could find a way to cover both cases, since updating the region for no-ops seems wasteful.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -160,40 +152,44 @@ public int getDSFID() {
...
+      }
+      updateScore(memberToAdd, scoreToAdd, entry);
     }
+    return true;
+  }
+
+  private void updateScore(byte[] memberToAdd, double scoreToAdd, OrderedSetEntry entry) {
+    scoreSet.remove(entry);
+    entry.updateScore(scoreToAdd);
+    members.put(memberToAdd, entry);
```
jdeppe-pivotal, line 174: I think this line is not necessary since this is a reference to the value already in the `members` map. Could you also rather inline the remaining 3 lines since this method isn't used anywhere else.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -818,15 +781,11 @@ protected int sizeValue(OrderedSetEntry value) {
...
-          byte[] member = value.getMember();
-          byte[] score = value.getScoreBytes();
-          InternalDataSerializer.writeByteArray(member, out);
-          InternalDataSerializer.writeByteArray(score, out);
-        }
+      for (Map.Entry<byte[], OrderedSetEntry> entry : entrySet()) {
+        byte[] member = entry.getKey();
+        double score = entry.getValue().getScore();
+        InternalDataSerializer.writeByteArray(member, out);
+        InternalDataSerializer.writeDouble(score, out);
```
jdeppe-pivotal, line 788: Prefer `DataSerializer` instead of `InternalDataSerializer`

jdeppe-pivotal (changes requested):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
nonbinaryprogrammer: sorry, thought I'd changed that

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisSortedSetTest.java
```diff
@@ -147,11 +147,13 @@ public void zadd_stores_delta_that_is_stable() {
     Region<RedisKey, RedisData> region = uncheckedCast(mock(Region.class));
     when(region.put(any(), any())).thenAnswer(this::validateDeltaSerialization);
     RedisSortedSet sortedSet1 = createRedisSortedSet("3.14159", "v1", "2.71828", "v2");
-    List<byte[]> adds = new ArrayList<>();
-    adds.add(stringToBytes("1.61803"));
-    adds.add(stringToBytes("v3"));
+    List<byte[]> members = new ArrayList<>();
+    members.add(stringToBytes("v3"));
+    List<Double> scores = new ArrayList<>();
+    scores.add(1.61803);
```
DonalEvans, line 153: This can be simplified to:
```
    List<byte[]> members = singletonList(stringToBytes("v3"));
    List<Double> scores = singletonList(1.61803);
```

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/ZAddExecutor.java
DonalEvans: > With the call to `Coder.bytesToDouble()` on line 60, this check shouldn't be needed, as the modified `bytesToDouble()` method will now throw an exception with the correct message when we get to that line. We would do a couple of unnecessary allocations if a user has passed NaN as a score with this removed, but we would avoid doing a byte array comparison on every argument in the more likely case where they haven't passed NaN, so it's a net performance improvement, I think.

This comment is still relevant, I think.

DonalEvans (changes requested):
Not all of these suggestions are required, but they're at least worth considering, I think. If any of them would be too much of a hassle to implement, I'm happy to let them go.

jdeppe-pivotal (approved):
Looks good. Thanks for sticking with this!

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/ZAddExecutor.java
```diff
@@ -50,8 +52,18 @@ public RedisResponse executeCommand(Command command, ExecutionHandlerContext con
       return RedisResponse.error(zAddExecutorState.exceptionMessage);
     }
 
-    Object retVal = redisSortedSetCommands.zadd(command.getKey(),
-        new ArrayList<>(commandElements.subList(optionsFoundCount + 2, commandElements.size())),
+    int size = (commandElements.size() - optionsFoundCount - 2) / 2;
+    List<byte[]> members = new ArrayList<>(size);
+    List<Double> scores = new ArrayList<>(size);
```
dschneider-pivotal, line 57: You can change this List<Double> to be "double[]"

DonalEvans (approved):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/AbstractRedisData.java
```diff
@@ -200,6 +202,15 @@ public void fromDelta(DataInput in) throws IOException, InvalidDeltaException {
         byte[] byteArray = DataSerializer.readByteArray(in);
         applyDelta(new AppendDeltaInfo(byteArray, sequence));
         break;
+      case ZADDS:
+        int numMembers = DataSerializer.readPrimitiveInt(in);
```
dschneider-pivotal, line 206: If you make the changes I suggested to ZAddsDeltaInfo this could would look like this:
int numMembers = DataSerializer.readPrimitiveInt(in);
ZAddsDeltaInfo delta = new ZAddsDeltaInfo(numMembers);
for (int i = 0; i < numMembers; i++) {
  byte[] member = DataSerializer.readByteArray(in);
  double score = DataSerializer.readPrimitiveDouble(in);
  delta.add(member, score);
}
applyDelta(delta);

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/netty/CoderTest.java
```diff
@@ -105,16 +110,6 @@ public void narrowLongToInt_correctlyConvertsLongToInt() {
     }
   }
 
-  @Test
-  @Parameters(method = "doubleBytes")
-  public void stripTrailingZeroFromDouble_correctlyStripsTrailingZero(byte[] input,
```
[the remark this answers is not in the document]
nonbinaryprogrammer: added

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/delta/ZAddsDeltaInfo.java
```diff
@@ -0,0 +1,68 @@
...
+import static org.apache.geode.redis.internal.delta.DeltaType.ZADDS;
+
+import java.io.DataOutput;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+import org.apache.geode.DataSerializer;
+
+public class ZAddsDeltaInfo implements DeltaInfo {
```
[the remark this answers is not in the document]
nonbinaryprogrammer: done

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/ZAddExecutor.java
nonbinaryprogrammer: done

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/AbstractRedisData.java
nonbinaryprogrammer: done

ringles (approved):
Looks good by now. Sorry it was necessary.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -262,30 +258,26 @@ long zcount(SortedSetScoreRangeOptions rangeOptions) {
...
-    memberAdd(member, byteIncr);
-
-    AddsDeltaInfo deltaInfo = new AddsDeltaInfo(new ArrayList<>());
-    deltaInfo.add(member);
-    deltaInfo.add(byteIncr);
-
-    storeChanges(region, key, deltaInfo);
+    if (!(memberAdd(member, score) == MemberAddResult.NO_OP)) {
+      ZAddsDeltaInfo deltaInfo = new ZAddsDeltaInfo(1);
+      storeChanges(region, key, deltaInfo);
```
DonalEvans, line 277: The failing DUnit test can be fixed by making this
```
      ZAddsDeltaInfo deltaInfo = new ZAddsDeltaInfo(1);
      deltaInfo.add(member, score);
      storeChanges(region, key, deltaInfo);
```

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/ZAddExecutor.java
nonbinaryprogrammer: sorry, guess I didn't change all of them

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
nonbinaryprogrammer: thanks

dschneider-pivotal (approved):

PR #6787, opened 2021-08-23
GEODE-9451: On demand authentication expiration and re-authentication

jinmeiliao:
Merge feature branch to develop.
These 3 commits have already gone through review when merged to feature branch.

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/security/SecurityManager.java
+++ b/geode-core/src/main/java/org/apache/geode/security/SecurityManager.java
@@ -74,8 +74,11 @@ public interface SecurityManager {
    * @param principal The principal that's requesting the permission
    * @param permission The permission requested
    * @return true if authorized, false if not
+   *
+   * @throw AuthenticationExpiredException if the principal has expired.
    */
-  default boolean authorize(Object principal, ResourcePermission permission) {
+  default boolean authorize(Object principal, ResourcePermission permission)
+      throws AuthenticationExpiredException {
     return true;
   }
 
--- a/geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationDUnitTest.java
+++ b/geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationDUnitTest.java
@@ -48,8 +48,8 @@ import org.apache.geode.test.junit.runners.CategoryWithParameterizedRunnerFactor
 @RunWith(Parameterized.class)
 @Parameterized.UseParametersRunnerFactory(CategoryWithParameterizedRunnerFactory.class)
 public class AuthExpirationDUnitTest {
-  static RegionService regionService0;
-  static RegionService regionService1;
+  static RegionService user0Service;
+  static RegionService user1Service;
 
   @Parameterized.Parameter
   public String clientVersion;
@@ -110,9 +110,14 @@ public class AuthExpirationDUnitTest {
 
     // all put operation succeeded
     Region<Object, Object> region = server.getCache().getRegion("/region");
-    assertThat(ExpirableSecurityManager.getExpiredUsers().size()).isEqualTo(1);
-    assertThat(ExpirableSecurityManager.getExpiredUsers().contains("user1")).isTrue();
     assertThat(region.size()).isEqualTo(2);
+    Map<String, List<String>> authorizedOps = ExpirableSecurityManager.getAuthorizedOps();
+    Map<String, List<String>> unAuthorizedOps = ExpirableSecurityManager.getUnAuthorizedOps();
+    assertThat(authorizedOps.keySet().size()).isEqualTo(2);
+    assertThat(authorizedOps.get("user1")).asList().containsExactly("DATA:WRITE:region:0");
+    assertThat(authorizedOps.get("user2")).asList().containsExactly("DATA:WRITE:region:1");
+    assertThat(unAuthorizedOps.keySet().size()).isEqualTo(1);
+    assertThat(unAuthorizedOps.get("user1")).asList().containsExactly("DATA:WRITE:region:1");
   }
 
   @Test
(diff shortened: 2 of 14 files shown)


onichols-pivotal (approved):

geode-core/src/main/java/org/apache/geode/security/SecurityManager.java
```diff
@@ -74,8 +74,11 @@ default void init(Properties securityProps) {}
    * @param principal The principal that's requesting the permission
    * @param permission The permission requested
    * @return true if authorized, false if not
+   *
+   * @throw AuthenticationExpiredException if the principal has expired.
    */
-  default boolean authorize(Object principal, ResourcePermission permission) {
+  default boolean authorize(Object principal, ResourcePermission permission)
+      throws AuthenticationExpiredException {
```
kirklund, line 81: Not sure if this breaks binary compatibility for User's existing impls. I think it probably does.

geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationDUnitTest.java
```diff
@@ -0,0 +1,181 @@
...
+import org.apache.geode.test.junit.categories.SecurityTest;
+import org.apache.geode.test.junit.rules.ServerStarterRule;
+import org.apache.geode.test.junit.runners.CategoryWithParameterizedRunnerFactory;
+
+@Category({SecurityTest.class})
+@RunWith(Parameterized.class)
+@Parameterized.UseParametersRunnerFactory(CategoryWithParameterizedRunnerFactory.class)
+public class AuthExpirationDUnitTest {
+  static RegionService user0Service;
+  static RegionService user1Service;
```
kirklund, line 52: These should be `private`.

geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationDUnitTest.java
```diff
@@ -0,0 +1,181 @@
...
+    int serverPort = server.getPort();
+    ClientVM clientVM = lsRule.startClientVM(0, clientVersion,
+        c -> c.withProperty(SECURITY_CLIENT_AUTH_INIT, UpdatableUserAuthInitialize.class.getName())
+            .withPoolSubscription(true)
+            .withServerConnection(serverPort));
+
+    clientVM.invoke(() -> {
+      ClientCache clientCache = ClusterStartupRule.getClientCache();
+      UpdatableUserAuthInitialize.setUser("user1");
+      assert clientCache != null;
```
kirklund, line 92: Let's remove all `assert` uses. Only use AssertJ Assertions in a test. You probably don't need to assert that a cache is not null; just let it throw an NPE.

geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationDUnitTest.java
```diff
@@ -0,0 +1,181 @@
...
+
+      user0Service.close();
+      user1Service.close();
+    });
+
+    Region<Object, Object> region = server.getCache().getRegion("/region");
+    assertThat(region.size()).isEqualTo(4);
+
+    Map<String, List<String>> authorizedOps = ExpirableSecurityManager.getAuthorizedOps();
+    assertThat(authorizedOps.keySet().size()).isEqualTo(3);
```
kirklund, line 170: `assertThat(authorizedOps.keySet()).hasSize(3);` will provide much better failure messages if this ever fails.

geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationMultiServerDUnitTest.java
```diff
@@ -0,0 +1,136 @@
...
+import org.apache.geode.test.dunit.rules.ClusterStartupRule;
+import org.apache.geode.test.dunit.rules.MemberVM;
+import org.apache.geode.test.junit.categories.SecurityTest;
+import org.apache.geode.test.junit.rules.ClientCacheRule;
+
+@Category({SecurityTest.class})
+public class AuthExpirationMultiServerDUnitTest implements Serializable {
+  public static final String REPLICATE_REGION = "replicateRegion";
+  public static final String PARTITION_REGION = "partitionRegion";
+  private MemberVM locator, server1, server2;
```
kirklund, line 41: Spotless allows this but it's not part of our code standard. Please separate all declarations.

geode-junit/src/main/java/org/apache/geode/security/ExpirableSecurityManager.java
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * this is a test security manager that will authenticate credentials when username matches the
+ * password. It will authorize all operations. It keeps a list of expired users, and will throw
+ * AuthenticationExpiredException if the user is in that list. This security manager is usually used
+ * with NewCredentialAuthInitialize.
+ *
+ * make sure to call reset after each test to clean things up.
+ */
+public class ExpirableSecurityManager extends SimpleSecurityManager {
```
kirklund, line 34: Product APIs that are supported for Users shouldn't extend anything in an `examples` package. `ExpirableSecurityManager` should be moved to `examples` or something needs to be reworked here. I recommend reviewing product User APIs on dev list or with Jake and Dan at a minimum.

kirklund (changes requested):

geode-core/src/main/java/org/apache/geode/security/SecurityManager.java
jinmeiliao: No, adding exception wouldn't. this is from Oracle java doc site:
13.4.21. Method and Constructor Throws
Changes to the throws clause of methods or constructors do not break compatibility with pre-existing binaries; these clauses are checked only at compile time.

geode-junit/src/main/java/org/apache/geode/security/ExpirableSecurityManager.java
jinmeiliao: this is in geode-junit, I thought it's test code, is it product apii?

geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationDUnitTest.java
jinmeiliao: I tried that, The ObjectAssert doesn't have `hasSize` method, if I did a `toList()` first, it complains that the object is not a `list`.

geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationDUnitTest.java
jinmeiliao: done

geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationDUnitTest.java
jinmeiliao: done

mhansonp (commented):
Use of simple security manager does not seem like a good idea. We should not depend on examples code as klund said.

Other than that things look good.

jinmeiliao:
> Use of simple security manager does not seem like a good idea. We should not depend on examples code as klund said.
> 
> Other than that things look good.

ExpirableSecurityManager is a test security manager.

geode-core/src/main/java/org/apache/geode/security/SecurityManager.java
kirklund: Good to know! Thank you

geode-junit/src/main/java/org/apache/geode/security/ExpirableSecurityManager.java
kirklund: That's test code only then.

kirklund (approved):

jdeppe-pivotal (approved):

PR #6789, opened 2021-08-23
GEODE-9458: function execution with expiring authentication

jmelchio:

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationFunctionDUnitTest.java
+++ b/geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationFunctionDUnitTest.java
@@ -1,0 +1,142 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.apache.geode.security;
+
+import static org.apache.geode.cache.execute.FunctionService.onServer;
+import static org.apache.geode.distributed.ConfigurationProperties.SECURITY_CLIENT_AUTH_INIT;
+import static org.apache.geode.distributed.ConfigurationProperties.SECURITY_MANAGER;
+import static org.apache.geode.test.version.VersionManager.CURRENT_VERSION;
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Properties;
+
+import org.junit.After;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+
+import org.apache.geode.cache.RegionShortcut;
+import org.apache.geode.cache.client.ClientCache;
+import org.apache.geode.cache.execute.Function;
+import org.apache.geode.cache.execute.ResultCollector;
+import org.apache.geode.distributed.ConfigurationProperties;
+import org.apache.geode.management.internal.security.TestFunctions;
+import org.apache.geode.test.dunit.rules.ClientVM;
+import org.apache.geode.test.dunit.rules.ClusterStartupRule;
+import org.apache.geode.test.dunit.rules.MemberVM;
+import org.apache.geode.test.junit.categories.SecurityTest;
+import org.apache.geode.test.junit.rules.VMProvider;
+
+@Category({SecurityTest.class})
+public class AuthExpirationFunctionDUnitTest {
+  private static final String RELEASE_VERSION = "1.13.3";
+
+  private static Function<Object> writeFunction;
+
+  private static MemberVM serverVM;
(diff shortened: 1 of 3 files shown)


geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationFunctionDUnitTest.java
```diff
@@ -0,0 +1,142 @@
...
+import org.apache.geode.test.junit.rules.VMProvider;
+
+@Category({SecurityTest.class})
+public class AuthExpirationFunctionDUnitTest {
+  private static final String RELEASE_VERSION = "1.13.3";
+
+  private static Function<Object> writeFunction;
+
+  private static MemberVM serverVM;
+  private static ClientVM currentVersionClientVM, releaseVersionClientVM;
```
kirklund, line 54: Please declare all variables, constants, etc on their own line. It's a bug in Spotless that it allows this.

geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationFunctionDUnitTest.java
```diff
@@ -0,0 +1,142 @@
...
+  public static void beforeClass() throws Exception {
+    Properties properties = new Properties();
+    properties.setProperty(SECURITY_MANAGER, ExpirableSecurityManager.class.getName());
+    properties.setProperty(ConfigurationProperties.SERIALIZABLE_OBJECT_FILTER,
+        "org.apache.geode.management.internal.security.TestFunctions*");
+    serverVM = lsRule.startServerVM(0, properties);
+
+    serverVM.invoke(() -> {
+      Objects.requireNonNull(ClusterStartupRule.getCache())
+          .createRegionFactory(RegionShortcut.REPLICATE).create("region");
```
kirklund, line 69: I assume you're adding all of these assertions and null checks because of an IntelliJ inspection complaining that `ClusterStartupRule.getCache()` can be null. I recommend changing `ClusterStartupRule.getCache()` to use some sort of NonNull annotation within `ClusterStartupRule` instead.

geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationFunctionDUnitTest.java
```diff
@@ -0,0 +1,142 @@
...
+      assertThat(clientCache).isNotNull();
+      UpdatableUserAuthInitialize.setUser("data2");
+      ResultCollector rc = onServer(clientCache.getDefaultPool()).execute(writeFunction);
+      assertThat(((ArrayList) rc.getResult()).get(0))
+          .isEqualTo(TestFunctions.WriteFunction.SUCCESS_OUTPUT);
+    });
+
+    // all put operation succeeded
+    serverVM.invoke(() -> {
+      assertThat(ExpirableSecurityManager.getExpiredUsers().size()).isEqualTo(1);
```
kirklund, line 129: I would change this and the following size assertions to use:
```
assertThat(ExpirableSecurityManager.getExpiredUsers()).hasSize(1);
```
This will work as long as the subject (ie `getExpiredUsers()`) is a collection or array.

Any failure messages will then list all of the entries.

kirklund (changes requested):

geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationFunctionDUnitTest.java
jmelchio: Yes, that's the reason.

Let me see if I can change the rule ...

geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationFunctionDUnitTest.java
jinmeiliao: I tried to use that method, but doesn't seem to work.

jmelchio: I changed it to use that but had to add `.asList()`

geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationFunctionDUnitTest.java
jmelchio: It appears that `null` is a valid return in the rule so adding a `@Notnull` will just move the warning elsewhere without more drastic changes.

geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationFunctionDUnitTest.java
jinmeiliao: Didn't it complain about the object is not a list at runtime?

mhansonp (approved):

geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationFunctionDUnitTest.java
jmelchio: > Didn't it complain about the object is not a list at runtime?

No, works fine.

jinmeiliao:
Instead of using different vm for different version client, a parameterized test would be more appropriate since we should keep adding newer versions to the test. Try add this test to `AuthExpirationDUnitTest` as that is already a parameterized test. This could be just a another test method added there. This tests a single server with different version of client. I would suggest also adding a test to the `AuthExpirationMultiServerDUnitTest`,  and test those `onRegion`, `onServer` and `onServers` calls and verify behavior is as expected.

jmelchio:
> Instead of using different vm for different version client, a parameterized test would be more appropriate since we should keep adding newer versions to the test. Try add this test to `AuthExpirationDUnitTest` as that is already a parameterized test. This could be just a another test method added there. This tests a single server with different version of client. I would suggest also adding a test to the `AuthExpirationMultiServerDUnitTest`, and test those `onRegion`, `onServer` and `onServers` calls and verify behavior is as expected.

I'll see if I can make that work.

kirklund (approved):

jinmeiliao (approved):

PR #6792, opened 2021-08-23
GEODE-9451: add isExpired to SecurityManager interface

jinmeiliao:
We need a method to simply check the expiration without permission check for message dispatcher. Having this method would also make the SecurityManager authorize method not responsible for expiration checks.

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/security/IntegratedSecurityService.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/security/IntegratedSecurityService.java
@@ -42,6 +42,7 @@ import org.apache.geode.internal.security.shiro.SecurityManagerProvider;
 import org.apache.geode.internal.security.shiro.ShiroPrincipal;
 import org.apache.geode.internal.util.BlobHelper;
 import org.apache.geode.logging.internal.log4j.api.LogService;
+import org.apache.geode.security.AuthenticationExpiredException;
 import org.apache.geode.security.AuthenticationFailedException;
 import org.apache.geode.security.AuthenticationRequiredException;
 import org.apache.geode.security.GemFireSecurityException;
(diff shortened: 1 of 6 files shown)


geode-core/src/main/java/org/apache/geode/internal/security/IntegratedSecurityService.java
```diff
@@ -272,6 +260,11 @@ public void authorize(ResourcePermission context, Subject currentUser) {
       return;
     }
 
+    Object principal = currentUser.getPrincipal();
```
pivotal-jbarrett, line 263: Can `principal` ever be `null`?

geode-core/src/main/java/org/apache/geode/security/SecurityManager.java
```diff
@@ -69,19 +69,27 @@ default void init(Properties securityProps) {}
...
+  default boolean authorize(Object principal, ResourcePermission permission) {
     return true;
   }
 
+  /**
+   * checks if the given principal is expired or not
+   *
+   * @return true if the given principal is expired, false if not
+   */
+  default boolean isExpired(Object principal) {
```
pivotal-jbarrett, line 89:
```java
default boolean isExpired(@NotNull Object principal)
```

geode-core/src/main/java/org/apache/geode/internal/security/IntegratedSecurityService.java
```diff
@@ -272,6 +260,11 @@ public void authorize(ResourcePermission context, Subject currentUser) {
...
+    if (securityManager.isExpired(principal)) {
+      throw new AuthenticationExpiredException(principal + " expired.");
```
pivotal-jbarrett, line 265: Given that `Object.toString()` is used to more diagnostic purposes does it make sense to include that in the message? It may result information leak about the credential we don't want serialized. Simply throwing this exception should be information enough.

pivotal-jbarrett (changes requested):

jmelchio (approved):

PR #6804, opened 2021-08-25
GEODE-9453: The server, once a user expires, should clean the user at…

jinmeiliao:
…tributes from the server.

This logout the given subject when an AuthExpiredException is encountered. 

This PR doesn't try to clean out the `ClientUserAuths.uniqueIdVsSubject` map for the following reasons:
1. there is no clean way to get access to it unless using the static variable in `ServerConnection`
2. when a clients connects, multiple threads calls into the server to authenticate, resulting into multiple in that entries in that map, each with different unique id and unique subject/principal, even though it's the same user. Logging out one subject nulls one but didn't null the others. The other subject in the map won't be able to do anything anyway because every attempted operation will logout the user.
3. This map will eventually get cleaned up when connection closes or dies, so there isn't much point trying to clean up the nullified subject in this map (as far I can tell). 

Would love to get more feedback on this.


kirklund (approved):

mhansonp (approved):

jinmeiliao:
Thanks for the review. Closing this PR and incorporate the changes in a coming PR.

PR #6808, opened 2021-08-26
GEODE-9554: Rebalancing a region with multiple redundancy zones can fail

mhansonp:
added the functionality to check to see if a bucket a requesting member is asking to delete on a remote host is in the same redundancy zone.

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/partitioned/rebalance/model/Member.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/partitioned/rebalance/model/Member.java
@@ -66,12 +66,12 @@ public class Member implements Comparable<Member> {
     if (checkZone) {
       if (addressComparor.enforceUniqueZones()
           && !addressComparor.areSameZone(getMemberId(), sourceMember)) {
-        logger.debug(
+        logger.info(
             "Member {} cannot delete {} because it is in a different redundancy zone than {}",
             getMemberId(), bucket, sourceMember);
         return RefusalReason.DIFFERENT_ZONE;
       } else {
-        logger.debug(
+        logger.info(
             "Member {} will delete  {} because it is already on another member with the same redundancy zone",
             this, bucket);
       }
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/partitioned/rebalance/model/Member.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/partitioned/rebalance/model/Member.java
@@ -62,22 +62,19 @@ public class Member implements Comparable<Member> {
 
 
 
-  /**
-   * @param sourceMember the member we will be moving this bucket off of
-   * @param checkZone true if we should not put two copies of a bucket on two nodes with the same
-   */
-  public RefusalReason canDelete(Bucket bucket, InternalDistributedMember sourceMember,
-      boolean checkZone,
-      DistributionManager distributionManager) {
+  public RefusalReason canDelete(Bucket bucket, DistributionManager distributionManager) {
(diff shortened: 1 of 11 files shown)


geode-core/src/main/java/org/apache/geode/internal/cache/partitioned/rebalance/model/PartitionedRegionLoadModel.java
```diff
@@ -470,12 +470,14 @@ public Move findBestRemove(Bucket bucket) {
     Move bestMove = null;
 
     for (Member member : bucket.getMembersHosting()) {
-      float newLoad = (member.getTotalLoad() - bucket.getLoad()) / member.getWeight();
-      if (newLoad > mostLoaded && !member.equals(bucket.getPrimary())) {
-        Move move = new Move(null, member, bucket);
-        if (!this.attemptedBucketRemoves.contains(move)) {
-          mostLoaded = newLoad;
-          bestMove = move;
+      if (member.canDelete(bucket, this.partitionedRegion.getMyId(), true).willAccept()) {
```
upthewaterspout, line 473: This seems to be the only call tot the new canDelete function? If so, the third boolean parameter doesn't need to be there. Also it can just return a boolean, since that is all you are using in the result. No need for a new RefusalReason.

geode-core/src/main/java/org/apache/geode/internal/cache/partitioned/rebalance/model/RefusalReason.java
```diff
@@ -42,6 +48,11 @@ public String formatMessage(Member target, Bucket bucket) {
       case CRITICAL_HEAP:
         return "Target member " + target.getMemberId()
             + " has reached its critical heap percentage, and cannot accept more data";
+      case DIFFERENT_ZONE:
```
upthewaterspout, line 51: All of these other `RefusalReasons` are refusing to create a bucket. I don't think DIFFERENT_ZONE belongs in this enum. Based on my ealier comment, I think you can just remove it entirely since we don't use it.

geode-core/src/main/java/org/apache/geode/internal/cache/partitioned/rebalance/model/Member.java
```diff
@@ -55,6 +55,31 @@
     this.localMaxMemory = localMaxMemory;
   }
 
+  /**
+   * @param sourceMember the member we will be moving this bucket off of
+   * @param checkZone true if we should not put two copies of a bucket on two nodes with the same
+   *        IP address.
+   */
+  public RefusalReason canDelete(Bucket bucket, InternalDistributedMember sourceMember,
+      boolean checkZone) {
```
upthewaterspout, line 64: I'm not sure the logic in this method is correct. What if we have an extra copy of the bucket, and each copy is in a different redundancy zone? It seems like this would prevent us from deleting the extra copy when if fact we should.

Maybe we could keep this method but only throw an exception if we find two copies of the bucket in the same zone *and* we are trying to delete a copy in a different zone, or something like that.

upthewaterspout (changes requested):

geode-core/src/main/java/org/apache/geode/internal/cache/partitioned/rebalance/model/PartitionedRegionLoadModel.java
mhansonp: I have cleaned that up. Since I rewrote the core of the function.

mhansonp:
Reopening with fix and clean testing.

PR #6823, opened 2021-08-31
GEODE-9486: Fix validate-serializable-objects (with Windows fixes)

kirklund:
Rename DistributedSystemService to SanctionedSerializablesService and
remove unused init(DistributedSystem).

Move SanctionedSerializablesService to geode-serialization.

Implement SanctionedSerializablesService in both geode-core and
geode-management to remove special case for each in
InternalDataSerializer.

Fix sanctioned serializables support in geode-management and
geode-apis-compatible-with-redis.

Add sanctioned serializables support to geode-serialization and
geode-pulse.

Add sanctioned serializables support to geode-junit and geode-dunit
to simplify writing tests for validate-serializable-objects and prevent
having to list out DUnit Rules and other test framework classes when
validate-serializable-objects is enabled.

Use ExecutorServiceRule and reformat json strings in
RestRegionAPIIntegrationTest.

Reorganize QueryCommandDUnitTestBase.

Use InvalidClassException instead of Exception in
ObjectInputStreamFilterWrapper fatal log message.

Improve assertion messages in ResourceUtils.

Move loadSanctionedSerializablesServices and loadClassNames to
new SanctionedSerializables in geode-serialization.

Exclude Pulse GemFireAuthentication from sanctioned serializables.

Add SerializationTest Category to all AnalyzeSerializables integration
tests.

Tidy up SANCTIONED_SERIALIZABLES_DEPENDENCIES_PATTERN.

Convert to AssertJ and use BeforeClass in
InternalDataSerializerSerializationAcceptlistTest.

Note: If Git or GitHub is showing invalid file renames in the diffs, you
may need to pull the branch locally and configure diff.renameLimit to
something lower than the default value of 50.


mhansonp (approved):

onichols-pivotal (approved):
did you intentionally re-flow the license headers in the .gradle files?

kirklund:
@onichols-pivotal No, I think that must be an option that's enabled in IntelliJ. I committed from IntelliJ's git plugin. I'll fix it up.

dschneider-pivotal (approved):

jchen21 (approved):

geode-dunit/src/integrationTest/java/org/apache/geode/codeAnalysis/AnalyzeDUnitSerializablesIntegrationTest.java
```diff
@@ -0,0 +1,63 @@
...
+
+@Category(SerializationTest.class)
+public class AnalyzeDUnitSerializablesIntegrationTest extends AnalyzeSerializablesJUnitTestBase {
+  private static final Logger logger = LogService.getLogger();
+
+  private static final Set<Class<?>> IGNORE_CLASSES = new HashSet<>();
+  static {
+    IGNORE_CLASSES.add(Master.class);
+    classForName("org.apache.geode.test.dunit.internal.RemoteDUnitVM")
+        .ifPresent(IGNORE_CLASSES::add);
```
DonalEvans, line 37: Why is this class only optionally added to the set of ignored classes rather than always being added?

kirklund: It's a package-private class, so we couldn't add it in the same way as `Master`. The use of `Optional` was Dale's preferred way for error-handling instead of using `null`. I started out coding this:
```
  private static Class<?> classForName(String className) {
    try {
      return Class.forName(className);
    } catch (ClassNotFoundException e) {
      logger.error("Unable to add class {} to IGNORE_CLASSES", className, e);
    }
    return null;
  }
```
Changing from `null` to `Optional` resulted in:
```
  private static Optional<Class<?>> classForName(String className) {
    try {
      return Optional.of(Class.forName(className));
    } catch (ClassNotFoundException e) {
      logger.error("Unable to add class {} to IGNORE_CLASSES", className, e);
    }
    return Optional.empty();
  }
```
These 2 RMI classes fail the various checks in AnalyzeSerializable, and the cause seems to be the `CompiledClass` library in geode-junit's org.apache.geode.codeAnalysis package (which I think was a 3rd party library that was pulled into our test code). All `RMIObjects` fail deserialization checks with an "invalid RMI" failure because the library doesn't know how to handle RMI. If `Class.forName` failed to load `RemoteDUnitVM`, 2 or 3 tests end up failing. I'd like to fix that library but this is for test-only code anyway, so I decided to have it just skip RMI classes.

One possibility that I considered was to throw `new RuntimeException(e)` wrapping the `ClassNotFoundException` but I think that will end up throwing a `LinkageError` of type `NoClassDefFoundError` when attempting to load `AnalyzeDUnitSerializablesIntegrationTest`.

Yet another approach might be to not use a static constant for the `IGNORE_CLASSES` and just build up a the `Set` every time `ignoreClass` is called.

I would be happy to change this to something better, especially if you have any other ideas.

DonalEvans: Thanks for the comprehensive explanation! I think the current approach is fine.

DonalEvans (approved):

boglesby (approved):

jinmeiliao (approved):

echobravopapa (approved):

PR #6826, opened 2021-09-01
GEODE-9542: Enable SSL client authentication for Radish

jdeppe-pivotal:
- When Geode's ssl-require-authentication is enabled, Redis clients must
  authenticate with a valid certificate.

### For all changes:

### Note:


geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/SSLDUnitTest.java
```diff
@@ -87,8 +87,9 @@ public void givenMutualAuthentication_clientErrorsWithoutKeystore() throws Excep
 
     // Create the client without a keystore
     try (Jedis jedis = createClient(false, false)) {
-      assertThatThrownBy(jedis::ping)
-          .hasMessageContaining("readHandshakeRecord");
+      assertThatThrownBy(jedis::ping).satisfiesAnyOf(
+          e -> assertThat(e.getMessage()).contains("SSLException"),
+          e -> assertThat(e.getMessage()).contains("SSLHandshakeException"));
```
DonalEvans, line 92: These assertions could be combined to a single one:
```
assertThatThrownBy(jedis::ping).isInstanceOf(SSLException.class);
```
Since `SSLHandshakeException` extends `SSLException`. Alternately, you could use `satisfiesAnyOf()` with assertions on the message body rather than exception name, but this might be trickier if you don't know all the different messages that might be expected in exceptions here.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/netty/NettyRedisServer.java
```diff
@@ -180,29 +181,45 @@ public void initChannel(SocketChannel socketChannel) {
...
+            FileWatchingX509ExtendedKeyManager.newFileWatchingKeyManager(sslConfigForServer));
+      }
+
+      TrustManagerFactory trustManagerFactory = null;
+      if (sslConfigForServer.getTruststore() != null) {
+        trustManagerFactory = new TrustManagerFactoryWrapper(
+            FileWatchingX509ExtendedTrustManager.newFileWatchingTrustManager(sslConfigForServer));
+      }
+
+      SslContextBuilder sslContextBuilder = SslContextBuilder.forServer(keyManagerFactory);
```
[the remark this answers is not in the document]
jdeppe-pivotal: Good catch. I cannot think why no keystore would be provided and it seems like a misconfiguration. You *need* a keystore if you want to use SSL/TLS. To that end I think it's OK to throw an exception in that case. WDYT?

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/SSLDUnitTest.java
jdeppe-pivotal: I've removed this block since exceptions are now just coming from the `createClient` call.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/netty/NettyRedisServer.java
```diff
@@ -189,13 +189,15 @@ private void addSSLIfEnabled(SocketChannel ch, ChannelPipeline p) {
       return;
     }
 
+    if (sslConfigForServer.getKeystore() == null) {
+      throw new IllegalStateException(
+          "Cannot start netty as no key manager is configured. Please ensure that the Geode property 'ssl-keystore' is set.");
```
DonalEvans, line 194: I like throwing an exception here, but I'm not certain that "Geode property" is the best way to word it, in case this code gets used in some context that isn't specific to Geode.

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/SSLDUnitTest.java
jdeppe-pivotal: And... now I've added it back since it seems that the `createClient` can also sometime fail (I have a feeling it's platform or JDK specific). But if the client is created, then this ping will definitely fail.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/netty/NettyRedisServer.java
jdeppe-pivotal: I'm not sure that's possible at this moment.

echobravopapa (approved):

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/SSLDUnitTest.java
```diff
@@ -0,0 +1,298 @@
...
+
+  @Test
+  public void givenMutualAuthentication_clientCanConnect() throws Exception {
+    try (Jedis jedis = createClient(true, false)) {
+      assertThat(jedis.ping()).isEqualTo("PONG");
+    }
+  }
+
+  @Test
+  public void givenMutualAuthentication_clientErrorsWithoutKeystore() throws Exception {
```
DonalEvans, line 101: I may just be misunderstanding, but this test name implies mutual authentication, but the boolean `mutualAuthentication` argument passed to the `createClient()` method is false.

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/SSLDUnitTest.java
```diff
@@ -0,0 +1,298 @@
...
+    serverStore.createKeyStore(serverKeyStoreFilename, commonPassword);
+
+    // Wait long enough for the file change to be detected
+    GeodeAwaitility.await().during(Duration.ofSeconds(PollingFileWatcher.PERIOD_SECONDS + 2))
+        .until(() -> true);
+
+    try (Jedis jedis = createClient(true, false)) {
+      assertThat(jedis.ping()).isEqualTo("PONG");
+    }
+    assertThat(hostnameVerifier.getSubjectAltNames()).contains(newServerName);
```
DonalEvans, line 185: Rather than the awaits here before the assertions, would it be possible to just have an `await().untilAsserted()` for the final assertions here? Or would the timing not work properly?

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/netty/NettyRedisServer.java
dschneider-pivotal: even Geode docs refer to these as "gemfire properties" (see: https://geode.apache.org/docs/guide/113/reference/topics/gemfire_properties.html) so I would replace "Geode" with "gemfire" in this message.
Also if you do a google search for "gemfire property" it gets you to docs but "geode property" does not.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/netty/NettyRedisServer.java
```diff
@@ -180,29 +181,47 @@ public void initChannel(SocketChannel socketChannel) {
...
+        trustManagerFactory = new TrustManagerFactoryWrapper(
+            FileWatchingX509ExtendedTrustManager.newFileWatchingTrustManager(sslConfigForServer));
+      }
+
+      SslContextBuilder sslContextBuilder = SslContextBuilder.forServer(keyManagerFactory);
+      sslContextBuilder.trustManager(trustManagerFactory);
 
-    } catch (KeyStoreException | NoSuchAlgorithmException | UnrecoverableKeyException | IOException
-        | CertificateException e) {
+      if (!"any".equals(sslConfigForServer.getCiphers())) {
```
dschneider-pivotal, line 211: other code that compares to "any" uses "equalsIgnoreCase". Also InternalHttpServer treats a blank (StringUtils.isNotBlank) the same as "any> Seems like this code should also. It seems like this should be encapsulated in SSLConfig method like isAnyCiphers and isAnyProtocols instead of doing it in multiple places in our code.

dschneider-pivotal (changes requested):

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/SSLDUnitTest.java
jdeppe-pivotal: The top level exception is always `JedisConnectionException` and the ones here are somewhere in the stack but not even the root cause. That's why I'm just looking at the message since that always contains one of these. I also have a strong suspicion that different plaforms and/or JDK produce slightly different messages or exception stacks. This seems to be one way to cover all bases.

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/SSLDUnitTest.java
jdeppe-pivotal: So the `givenMutualAuthentication` is supposed to refer to how the server is configured. Passing `false` to `createClient` in order to _not_ set it up to perform mutual auth and thus error.

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/SSLDUnitTest.java
jdeppe-pivotal: Thanks for the suggestion - maybe not as readable, but at least it doesn't wait for a fixed time.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/netty/NettyRedisServer.java
jdeppe-pivotal: Switched.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/netty/NettyRedisServer.java
jdeppe-pivotal: Done. Although I didn't change any other locations. I'll open a Jira for that.

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/SSLDUnitTest.java
DonalEvans: Maybe the test name could be "givenMutualAuthenticationOnServer..."? Not super important though.

DonalEvans (approved):

dschneider-pivotal (approved):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/netty/NettyRedisServer.java
```diff
@@ -180,29 +181,47 @@ public void initChannel(SocketChannel socketChannel) {
...
+        sslContextBuilder.protocols(
+            Arrays.asList(sslConfigForServer.getProtocolsAsStringArray()));
+      }
+
+      if (sslConfigForServer.isRequireAuth()) {
+        sslContextBuilder.clientAuth(ClientAuth.REQUIRE);
+      }
+      sslContext = sslContextBuilder.build();
+    } catch (IOException e) {
       throw new RuntimeException(e);
```
upthewaterspout, line 225: I know this wasn't originally your code - but maybe this method should just throw a IOException. Wrapping in a generic RuntimeException seems a bit suspect.

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/SSLDUnitTest.java
```diff
@@ -0,0 +1,301 @@
...
+
+    Jedis jedis;
+    try {
+      // Create the client with a self-signed certificate
+      jedis = createClient(true, true);
+    } catch (JedisConnectionException ignored) {
+      return;
+    }
+
+    // Sometimes the client is created successfully - perhaps this is platform/JDK specific
```
upthewaterspout, line 135: You might be able to write these tests a little more cleanly if you just wrap the both `createClient` and `ping` in one single assertThatThrownBy.

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/SSLDUnitTest.java
```diff
@@ -0,0 +1,301 @@
...
+    serverStore.withCertificate("server", serverCertificate);
+    serverStore.trust("ca", ca);
+
+    // Wait for one second since file timestamp granularity may only be seconds depending on the
+    // platform.
+    GeodeAwaitility.await().during(Duration.ofSeconds(1)).until(() -> true);
+    serverStore.createKeyStore(serverKeyStoreFilename, commonPassword);
+
+    // Try long enough for the file change to be detected
+    GeodeAwaitility.await().atMost(Duration.ofSeconds(PollingFileWatcher.PERIOD_SECONDS * 2))
```
upthewaterspout, line 181: I think maybe we need to up this duration (20 seconds) to make sure this test isn't flaky in CI?

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/SSLDUnitTest.java
```diff
@@ -0,0 +1,301 @@
...
+        .sanDnsName(newServerName)
+        .generate();
+
+    CertStores serverStore = CertStores.serverStore();
+    serverStore.withCertificate("server", serverCertificate);
+    serverStore.trust("ca", ca);
+
+    // Wait for one second since file timestamp granularity may only be seconds depending on the
+    // platform.
+    GeodeAwaitility.await().during(Duration.ofSeconds(1)).until(() -> true);
```
upthewaterspout, line 177: Maybe just a Thread.sleep?

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/SSLDUnitTest.java
```diff
@@ -0,0 +1,301 @@
...
+    GeodeAwaitility.await().atMost(Duration.ofSeconds(PollingFileWatcher.PERIOD_SECONDS * 2))
+        .untilAsserted(() -> {
+          assertThatThrownBy(() -> createClient(true, false))
+              .isInstanceOf(JedisConnectionException.class);
+        });
+
+    IgnoredException.removeAllExpectedExceptions();
+  }
+
+  private Jedis createClient(boolean mutualAuthentication, boolean isSelfSigned) throws Exception {
```
upthewaterspout, line 229: I didn't see `isSelfSigned` being set to true anywhere in the test.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/netty/NettyRedisServer.java
```diff
@@ -180,29 +181,47 @@ public void initChannel(SocketChannel socketChannel) {
...
 
-    } catch (KeyStoreException | NoSuchAlgorithmException | UnrecoverableKeyException | IOException
-        | CertificateException e) {
+      if (!sslConfigForServer.isAnyCiphers()) {
+        sslContextBuilder.ciphers(Arrays.asList(sslConfigForServer.getCiphersAsStringArray()));
+      }
+
+      if (!sslConfigForServer.isAnyProtocols()) {
+        sslContextBuilder.protocols(
+            Arrays.asList(sslConfigForServer.getProtocolsAsStringArray()));
```
upthewaterspout, line 217: I wonder if we should add some tests that our redis server is honoring the ciphers and protocols that the user is specifying.

upthewaterspout (commented):
Overall, looks great! Looks like you also added support for watching the keystore and truststore files.

A few comments, inline.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/netty/NettyRedisServer.java
jdeppe-pivotal: I switched it to an `UncheckIOException` which is probably no better. However, this code is only triggered when a new inbound connection is established and, as such, is totally asynchronous and there is no way to recover from a situation here; like misconfigured SSL params.

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/SSLDUnitTest.java
jdeppe-pivotal: Done.

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/SSLDUnitTest.java
jdeppe-pivotal: I bumped it to 3x (30 seconds)

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/SSLDUnitTest.java
jdeppe-pivotal: Done

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/SSLDUnitTest.java
jdeppe-pivotal: There are a couple of tests that do so.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/netty/NettyRedisServer.java
jdeppe-pivotal: I started down that road but I'm not sure how to best go about it. I started playing with some code I found that probes the port and tries to set various SSL options, but it seems a bit fragile since the JDKs are always adjusting their supported security. Any other ideas?

jdeppe-pivotal: I was able to verify manually, but that doesn't help.

upthewaterspout: If it helps, I see lots of other tests checked in that are picking explicit protocols - eg RestAPIsWithSSLDUnitTest.

I wonder if netty's use of the protocol list exactly matches ours, for example. We allow "any" to show up in the list, and we also *order* the list with the highest priority protocol first.

jdeppe-pivotal: So I added some test code, that I found, that probes a SSL-enabled server and informs what protocols, ciphers and certificates the server has. Originally it was just a standalone-app so I massaged it a bit for this use.

PR #6827, opened 2021-09-01
GEODE-9522: When a server is force disconnected, it should set shutdo…

gesterzhou:
…wn cause for dm to prevent clients recreating server connection.

    Co-authored-by: Barry Oglesby <boglesby@users.noreply.github.com>

see GEODE-9522 for detail. 

### For all changes:

### Note:

### DIFF
--- a/geode-core/src/distributedTest/java/org/apache/geode/cache30/ReconnectWithCacheXMLDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/cache30/ReconnectWithCacheXMLDUnitTest.java
@@ -117,7 +117,7 @@ public class ReconnectWithCacheXMLDUnitTest extends JUnit4CacheTestCase {
       }
     });
     MembershipManagerHelper.crashDistributedSystem(cache.getDistributedSystem());
-    await().untilAsserted(() -> membershipFailed.get());
+    await().until(() -> membershipFailed.get());
     await()
         .until(() -> cache.getReconnectedCache() != null);
 
--- a/geode-core/src/distributedTest/java/org/apache/geode/cache30/ReconnectWithCacheXMLDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/cache30/ReconnectWithCacheXMLDUnitTest.java
@@ -16,7 +16,6 @@ package org.apache.geode.cache30;
 
 import static org.apache.geode.distributed.ConfigurationProperties.LOCATORS;
 import static org.apache.geode.distributed.internal.membership.gms.membership.GMSJoinLeave.BYPASS_DISCOVERY_PROPERTY;
-import static org.apache.geode.internal.Assert.assertTrue;
 import static org.apache.geode.test.awaitility.GeodeAwaitility.await;
 import static org.apache.geode.test.util.ResourceUtils.createTempFileFromResource;
 import static org.junit.Assert.assertEquals;
@@ -118,7 +117,7 @@ public class ReconnectWithCacheXMLDUnitTest extends JUnit4CacheTestCase {
       }
     });
     MembershipManagerHelper.crashDistributedSystem(cache.getDistributedSystem());
-    assertTrue(membershipFailed.get());
+    await().untilAsserted(() -> membershipFailed.get());
     await()
         .until(() -> cache.getReconnectedCache() != null);
 
--- a/geode-core/src/main/java/org/apache/geode/distributed/internal/ClusterDistributionManager.java
+++ b/geode-core/src/main/java/org/apache/geode/distributed/internal/ClusterDistributionManager.java
@@ -2332,6 +2332,15 @@ public class ClusterDistributionManager implements DistributionManager {
     }
 
     @Override
+    public void setShutdownCause(Exception shutdownCause) {
+      if (shutdownCause != null && !(shutdownCause instanceof ForcedDisconnectException)) {
+        logger.info("cluster membership failed due to ", shutdownCause);
+        shutdownCause = new ForcedDisconnectException(shutdownCause.getMessage());
+      }
+      dm.setRootCause(shutdownCause);
+    }
+
+    @Override
     public void newMemberConnected(InternalDistributedMember member) {
       // Do not elect the elder here as surprise members invoke this callback
       // without holding the view lock. That can cause a race condition and
@@ -2407,15 +2416,6 @@ public class ClusterDistributionManager implements DistributionManager {
       if (!dm.getConfig().getDisableAutoReconnect()) {
         dm.getCache().saveCacheXmlForReconnect();
       }
-    }
-
-    @Override
-    public void setShutdownCause(Exception shutdownCause) {
-      if (shutdownCause != null && !(shutdownCause instanceof ForcedDisconnectException)) {
-        logger.info("cluster membership failed due to ", shutdownCause);
-        shutdownCause = new ForcedDisconnectException(shutdownCause.getMessage());
-      }
-      dm.setRootCause(shutdownCause);
     }
   }
 
--- a/geode-core/src/main/java/org/apache/geode/distributed/internal/ClusterDistributionManager.java
+++ b/geode-core/src/main/java/org/apache/geode/distributed/internal/ClusterDistributionManager.java
@@ -2412,11 +2412,6 @@ public class ClusterDistributionManager implements DistributionManager {
         dm.getCache().saveCacheXmlForReconnect();
       }
     }
-
-    @Override
-    public void setShutdownCause(Exception shutdownCause) {
-      dm.setRootCause(shutdownCause);
-    }
   }
 
 
--- a/geode-core/src/main/java/org/apache/geode/distributed/internal/ClusterDistributionManager.java
+++ b/geode-core/src/main/java/org/apache/geode/distributed/internal/ClusterDistributionManager.java
@@ -278,8 +278,6 @@ public class ClusterDistributionManager implements DistributionManager {
    */
   private volatile Throwable rootCause = null;
 
-  private volatile boolean isForceDisconnecting = false;
-
   /**
    * @see #closeInProgress
    */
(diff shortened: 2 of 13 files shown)


kirklund (changes requested):
You should probably remove "DRAFT" from the title so others don't skip reviewing.

Looks good! Just needs some tests.

mhansonp (approved):

geode-core/src/main/java/org/apache/geode/distributed/internal/ClusterDistributionManager.java
```diff
@@ -2335,6 +2330,16 @@ public void membershipFailure(String reason, Throwable t) {
       }
     }
 
+    @Override
+    public Throwable setShutdownCause(Throwable cause) {
+      if (cause != null && !(cause instanceof ForcedDisconnectException)) {
```
Bill, line 2335: Reinterpreting the `cause` sent to this listener method breaks the contract. The listener is a recipient of information (in this case, from membership). What if the reason for shutdown was not a forced disconnect?

Also wrong, is calling this listener method from the `membershipFailure()` method above. This method is a callback that should only be invoked by membership.

In fact before this method is invoked from `membershipFailure()` it has already been invoked by membership.

Bill (changes requested):
Thank you for your work on this PR! It'll be good to have the shutdown cause consistently recorded when this work is complete.

The purpose of the new `MembershipListener` method `setShutdownCause` is unclear. It seems to be a setter, but the implementation in `DMListener` is reinterpreting the value sent to the method (and returning the new value, which in turn is used in the calling context).

In trying to work through the whole messaging flow touched by this PR I see it starting with `GMSMembership.ManagerImpl.forceDisconnect()`. Here is a very rough object interaction diagram showing the flow:

IMG_2180

Please have a look at that diagram and my comments below. There are redundant and in some cases inconsistent values being set in this flow. The new listener method (contract) needs work.

geode-core/src/main/java/org/apache/geode/distributed/internal/ClusterDistributionManager.java
gesterzhou: Here only refactored the existing code into a method, without changing any logic. 

See the replaced code, it changed cause to be ForcedDisconnectException.

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/api/MembershipListener.java
```diff
@@ -71,4 +71,7 @@ void memberSuspect(ID suspect, ID whoSuspected,
    */
   void saveConfig();
 
+  default Throwable setShutdownCause(Throwable t) {
```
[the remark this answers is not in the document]
gesterzhou: I can add javadoc for this method. 
Maybe the method name can be changed to setPossibleShutdownCauseToForceDisconnectException(shutdownCause)

The method signature is from the refactoring above. Since the original code expects to convert throwable into ForceDisconnectionException, thus this setShutdownCause() should return a Throwable. 

From listener point of view, we don't need to return anything. We did this way is only to avoid duplicated code. 

Of cause, I can complete revoked the code changes in ClusterDistributionManager.java, then I can change to void setShutdownCause(Throwable t) here, because in GMSMembership.java, we don't need setShutdownCause() to return anything.  

However, in our code, there will be 2 places with very similar code, one in ClusterDistributionManager, one in GMSMembership.  That's what we want to avoid. 

Conceptually, the whole fix's idea is to move setShutdownCause() a little earlier in force disconnect case. So it makes sense to do the refactoring in ClusterDistributionManager.DMListener into a method, then let listener to make use of this method.

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSMembership.java
```diff
@@ -1775,6 +1775,24 @@ public void memberSuspected(ID initiator, ID suspect, String reason) {
...
+    }
+
+    void uncleanShutdownDS(String reason, Exception shutdownCause) {
+      try {
+        listener.saveConfig();
+      } finally {
+        new LoggingThread("DisconnectThread", false, () -> {
+          lifecycleListener.forcedDisconnect();
+          listener.setShutdownCause(shutdownCause);
+          uncleanShutdown(reason, shutdownCause);
```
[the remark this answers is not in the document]
gesterzhou: We did not change to use logging thread  here. The original code called uncleanShutdown() in logging thread. 

Our fix here is only to add  listener.setShutdownCause(shutdownCause) in both isReconnectingDS() and not reconnecting case.

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSMembership.java
```diff
@@ -1797,19 +1815,9 @@ public void forceDisconnect(final String reason) {
...
-      }
-
-      try {
-        listener.saveConfig();
-      } finally {
-        new LoggingThread("DisconnectThread", false, () -> {
-          lifecycleListener.forcedDisconnect();
-          uncleanShutdown(reason, shutdownCause);
-        }).start();
+        uncleanShutdownReconnectingDS(reason, shutdownCause);
```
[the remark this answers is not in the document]
gesterzhou: services.setShutdownCause(shutdownCause);
is different with listener.setShutdownCause(shutdownCause).

They are 2 separate steps in previous code. 

services.setShutdownCause() set shutdownCause into Services class only, while 
ClusterDistributionManager.DMListener's setShutdownCause(shutdownCause) will convert other cause into ForceDisconnectException and set into DM. 

We call listener.setShutdownCause() in both if and else is because listener.setShutdownCause() has to be called after listener.saveConfig() and lifecycleListener.forcedDisconnect(). 

It cannot be called around line 1808. 

To make thing clearer, maybe we can the listener's method name to be:
setPossibleShutdownCauseToForceDisconnectException(shutdownCause). Then it will not mixed with services.setShutdownCause().

geode-membership/src/main/java/org/apache/geode/distributed/internal/membership/gms/GMSMembership.java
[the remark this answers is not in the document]
Bill: I now understand that the testing impact in making this synchronous outweighs my desire for simplicity in this case.

geode-core/src/distributedTest/java/org/apache/geode/cache30/ReconnectWithCacheXMLDUnitTest.java
```diff
@@ -118,7 +117,7 @@ public void beforeMembershipFailure(String reason, Throwable cause) {
       }
     });
     MembershipManagerHelper.crashDistributedSystem(cache.getDistributedSystem());
-    assertTrue(membershipFailed.get());
+    await().untilAsserted(() -> membershipFailed.get());
```
kirklund, line 120: This line won't do anything as written. It needs to either be:
```
await().until(() -> membershipFailed.get());
```
Or:
```
await().untilAsserted(() -> assertThat(membershipFailed.get()).isTrue());
```

geode-membership/src/integrationTest/java/org/apache/geode/distributed/internal/membership/gms/GMSMembershipJUnitTest.java
```diff
@@ -197,7 +202,54 @@ public void testSendMessage() throws Exception {
...
+    GMSMembership.ManagerImpl managerImpl = (GMSMembership.ManagerImpl) manager.getGMSManager();
+    managerImpl = spy(managerImpl);
+    when(manager.getGMSManager()).thenReturn(managerImpl);
+
+    manager.forceDisconnect(reason);
+    InOrder inOrder = inOrder(managerImpl, directChannelCallback, listener);
+    inOrder.verify(managerImpl, times(1)).uncleanShutdownDS(eq(reason),
+        isA(MemberDisconnectedException.class));
+    inOrder.verify(directChannelCallback, timeout(3000).times(1)).forcedDisconnect();
+    inOrder.verify(listener, timeout(3000).times(1)).forcedDisconnectHappened(eq(reason));
```
kirklund, line 226: Is there a reason you need a timeout and if so, does it need to be less than `GeodeAwaitility.getTimeout`?

geode-core/src/distributedTest/java/org/apache/geode/cache30/ReconnectWithCacheXMLDUnitTest.java
gesterzhou: done

geode-membership/src/integrationTest/java/org/apache/geode/distributed/internal/membership/gms/GMSMembershipJUnitTest.java
gesterzhou: It's running in a thread.

Bill (approved):
This looks great @gesterzhou!

agingade (approved):

geode-membership/src/integrationTest/java/org/apache/geode/distributed/internal/membership/gms/GMSMembershipJUnitTest.java
kirklund: I'm still concerned about what happens if this test runs on a very slow machine and that thread loses CPU time such that it doesn't complete within 3 seconds. This code is the same as doing something like `await().atMost(3, SECONDS).until` some boolean is set by the callback or listener and then immediately performing an assertion. If we really need GeodeAwaitility timeout to be 5 minutes for slow machines, then I think these verify timeouts need to be that long too.

gesterzhou: OK, I changed the timeout.

kirklund (approved):
Looks good now. Thanks Gester!

PR #6831, opened 2021-09-02
GEODE-9519: Implement Radish ZSCAN command

DonalEvans:
This PR is intentionally split into two commits, the first refactors the existing *SCAN implementations and some related code. The second implements the ZSCAN command and adds tests. The changes in the first commit also address GEODE-9427 as a side-effect.

### For all changes:

### Note:

### DIFF
--- a/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisHash.java
+++ b/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisHash.java
@@ -244,10 +244,9 @@ public class RedisHash extends AbstractRedisData {
       throw new IllegalArgumentException("Requested array size exceeds VM limit");
     }
     List<byte[]> resultList = new ArrayList<>((int) maximumCapacity);
-    do {
-      cursor = hash.scan(cursor, 1,
-          (list, key, value) -> addIfMatching(matchPattern, list, key, value), resultList);
-    } while (cursor != 0 && resultList.size() < maximumCapacity);
+
+    cursor = hash.scan(cursor, count,
+        (list, key, value) -> addIfMatching(matchPattern, list, key, value), resultList);
 
     return new ImmutablePair<>(cursor, resultList);
   }
--- a/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
+++ b/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
@@ -65,7 +65,7 @@ import org.apache.geode.redis.internal.netty.Coder;
 
 public class RedisSortedSet extends AbstractRedisData {
   protected static final int REDIS_SORTED_SET_OVERHEAD = memoryOverhead(RedisSortedSet.class);
-  public static final Logger logger = LogService.getLogger();
+  private static final Logger logger = LogService.getLogger();
 
   private MemberMap members;
   private final ScoreSet scoreSet = new ScoreSet();
--- a/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
+++ b/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
@@ -37,6 +37,7 @@ import java.util.Objects;
 import java.util.regex.Pattern;
 
 import org.apache.commons.lang3.tuple.ImmutablePair;
+import org.apache.logging.log4j.Logger;
 
 import org.apache.geode.DataSerializer;
 import org.apache.geode.annotations.VisibleForTesting;
@@ -64,6 +65,7 @@ import org.apache.geode.redis.internal.netty.Coder;
 
 public class RedisSortedSet extends AbstractRedisData {
   protected static final int REDIS_SORTED_SET_OVERHEAD = memoryOverhead(RedisSortedSet.class);
+  public static final Logger logger = LogService.getLogger();
 
   private MemberMap members;
   private final ScoreSet scoreSet = new ScoreSet();
@@ -391,7 +393,7 @@ public class RedisSortedSet extends AbstractRedisData {
     // need to add 1 to zcard() to ensure that if count > members.size(), we return a cursor of 0
     long maximumCapacity = 2L * Math.min(count, zcard() + 1);
     if (maximumCapacity > Integer.MAX_VALUE) {
-      LogService.getLogger().info(
+      logger.info(
           "The size of the data to be returned by zscan, {}, exceeds the maximum capacity of an array. A value for the ZSCAN COUNT argument less than {} should be used",
           maximumCapacity, Integer.MAX_VALUE / 2);
       throw new IllegalArgumentException("Requested array size exceeds VM limit");
--- a/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
+++ b/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
@@ -397,10 +397,9 @@ public class RedisSortedSet extends AbstractRedisData {
       throw new IllegalArgumentException("Requested array size exceeds VM limit");
     }
     List<byte[]> resultList = new ArrayList<>((int) maximumCapacity);
-    do {
-      cursor = members.scan(cursor, 1,
-          (list, key, value) -> addIfMatching(matchPattern, list, key, value.score), resultList);
-    } while (cursor != 0 && resultList.size() < maximumCapacity);
+
+    cursor = members.scan(cursor, count,
+        (list, key, value) -> addIfMatching(matchPattern, list, key, value.score), resultList);
 
     return new ImmutablePair<>(cursor, resultList);
   }
--- a/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
+++ b/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
@@ -20,6 +20,8 @@ import static java.lang.Double.compare;
 import static org.apache.geode.internal.JvmSizeUtils.memoryOverhead;
 import static org.apache.geode.redis.internal.data.NullRedisDataStructures.NULL_REDIS_SORTED_SET;
 import static org.apache.geode.redis.internal.data.RedisDataType.REDIS_SORTED_SET;
+import static org.apache.geode.redis.internal.netty.Coder.bytesToString;
+import static org.apache.geode.redis.internal.netty.Coder.doubleToBytes;
 import static org.apache.geode.redis.internal.netty.StringBytesGlossary.bGREATEST_MEMBER_NAME;
 import static org.apache.geode.redis.internal.netty.StringBytesGlossary.bLEAST_MEMBER_NAME;
 
@@ -32,6 +34,9 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.regex.Pattern;
+
+import org.apache.commons.lang3.tuple.ImmutablePair;
 
 import org.apache.geode.DataSerializer;
 import org.apache.geode.annotations.VisibleForTesting;
@@ -40,6 +45,7 @@ import org.apache.geode.internal.serialization.DeserializationContext;
 import org.apache.geode.internal.serialization.KnownVersion;
 import org.apache.geode.internal.serialization.SerializationContext;
 import org.apache.geode.internal.size.Sizeable;
+import org.apache.geode.logging.internal.log4j.api.LogService;
 import org.apache.geode.redis.internal.RedisConstants;
 import org.apache.geode.redis.internal.RegionProvider;
 import org.apache.geode.redis.internal.collections.OrderStatisticsTree;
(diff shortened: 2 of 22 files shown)


geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/key/ScanExecutor.java
```diff
@@ -103,7 +104,7 @@ public RedisResponse executeCommand(Command command, ExecutionHandlerContext con
         scan(getDataRegion(context).keySet(), matchPattern, count, cursor);
     context.setScanCursor(scanResult.getLeft());
 
-    return RedisResponse.scan(scanResult.getLeft(), scanResult.getRight());
+    return RedisResponse.scan(scanResult.getLeft().intValue(), scanResult.getRight());
   }
 
   private Pair<BigInteger, List<Object>> scan(Collection<RedisKey> list,
```
dschneider-pivotal, line 110: why does this BigInteger flavor still exist

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -377,6 +382,25 @@ long zrevrank(byte[] member) {
...
+    long maximumCapacity = 2L * Math.min(count, zcard() + 1);
+    if (maximumCapacity > Integer.MAX_VALUE) {
+      LogService.getLogger().info(
+          "The size of the data to be returned by zscan, {}, exceeds the maximum capacity of an array. A value for the ZSCAN COUNT argument less than {} should be used",
+          maximumCapacity, Integer.MAX_VALUE / 2);
+      throw new IllegalArgumentException("Requested array size exceeds VM limit");
+    }
+    List<byte[]> resultList = new ArrayList<>((int) maximumCapacity);
+    do {
+      cursor = members.scan(cursor, 1,
```
dschneider-pivotal, line 397: Something I find out about our scan code is that we implemented SizeableObject2ObjectOpenCustomHashMapWithCursor#scan to take a "count" parameter that allows it to loop but then when we call it we always give it "1" and the caller does a loop. It seems like we could refactor it so that the shared scan method on our HashMap does all the looping.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/key/ScanExecutor.java
DonalEvans: I didn't want to have this PR get too big, so I haven't changed the existing implementations for SCAN or SSCAN. They're currently not supported commands, so I've left that work for whoever eventually implements them in the future.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
DonalEvans: Good call. It seems that all the tests pass with this change, so I've applied it to simplify things.

jdeppe-pivotal: I think this breaks things a bit. @DonalEvans the assertion that was removed in `RedisHashTest.hscanOnlyReturnsElementsMatchingPattern` should remain. If that test is translated into an integration test you'll see it passes in native redis (with a cursor result of `0`).
The javadoc for `SizeableBytes2ObjectOpenCustomHashMapWithCursor.scan` is slightly misleading (for me at least) in that it says:
> This method will scan until at least count entries are returned...
It should preferably say: _"until at least count entries are consumed"_. The caller of this method,`RedisHash.hscan`, has a different semantic for `count` in that it wants to _return_ that many entries that _match_ the pattern not just process that many entries in the hash.

jdeppe-pivotal: This is the integration test I tried:
```
  @Test
  public void wat() {
    Map<String, String> hashy = new HashMap<>();
    hashy.put("ak1", "v1");
    hashy.put("k2", "v2");
    hashy.put("ak3", "v3");
    hashy.put("k4", "v4");
    jedis.hset(HASH_KEY, hashy);

    ScanResult<Map.Entry<byte[], byte[]>> result =
        jedis.hscan(HASH_KEY.getBytes(), "0".getBytes(), new ScanParams().count(3).match("a*"));

    assertThat(result.getCursor()).isEqualTo("0");
  }
```

DonalEvans: The difficulty here is that Redis has different behaviour depending on the size of the sorted set/hash. For sizes less than a certain amount, they internally use a different data structure, which results in different behaviour:

>"When iterating Sets encoded as intsets (small sets composed of just integers), or Hashes and Sorted Sets encoded as ziplists (small hashes and sets composed of small individual values), usually all the elements are returned in the first SCAN call regardless of the COUNT value."

from https://redis.io/commands/scan#the-count-option

But for larger sizes, a scan command may return no results (except a cursor value) when used with MATCH and COUNT. I created a hash with ~500 entries, three of which matched the pattern `a*`, and called HSCAN with `MATCH a*` and various COUNT values. Redis returned an empty array for COUNT values up to 106, and only one result for COUNT values up to 243, so it's not accurate to say that the COUNT argument indicates the number of matching entries that should be returned.

Given the inconsistent behaviour of Redis' scan commands, I'm not sure that we want to try to exactly match behaviour for situations where a full iteration isn't guaranteed, because our internal implementation will not behave the same for small set/hash sizes. This can be tested by creating a hash with 20 entries, then calling HSCAN with a COUNT value of 5. Redis will return all entries, but our implementation will return 5.

jdeppe-pivotal: That's pretty annoying :) I'd still argue for restoring the prior behavior (or some variant) since it should be more performant overall in that it will reduce the potential number of client calls. If we only process `count` entries for every `scan` call, the client is going to make ~`set_size / count` requests until they are done. The downside of the original implementation is that the key is locked for the duration of the call as more entries are iterated. Since the semantics are so loose, it would be awesome if the scan could be interrupted by a write operation, but that's just crazy talk...

Even Redis requires a lot of calls when the matches are sparse. With a set of 10003 entries where 3 entries will match the `zscan` call (and count=3), Redis still takes 2769 calls before the whole set is processed, so I guess leaving this change is OK for now.

jdeppe-pivotal (approved):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
DonalEvans: Yeah the COUNT/MATCH interaction isn't ideal. From Redis' SCAN documentation:
> It is important to note that the MATCH filter is applied after elements are retrieved from the collection, just before returning data to the client. This means that if the pattern matches very little elements inside the collection, SCAN will likely return no elements in most iterations.

so it'd be a significant departure from Redis' way of doing things if we kept the old implementation (even if it's better).

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -377,6 +382,24 @@ long zrevrank(byte[] member) {
     return scoreSet.size() - scoreSet.indexOf(orderedSetEntry) - 1;
   }
 
+  ImmutablePair<Integer, List<byte[]>> zscan(Pattern matchPattern, int count, int cursor) {
+    // No need to allocate more space than it's possible to use given the size of the sorted set. We
+    // need to add 1 to zcard() to ensure that if count > members.size(), we return a cursor of 0
+    long maximumCapacity = 2L * Math.min(count, zcard() + 1);
+    if (maximumCapacity > Integer.MAX_VALUE) {
+      LogService.getLogger().info(
```
kirklund, line 390: `LogService.getLogger()` constructs a new instance of `FastLogger` every time it's called. You should always store it in a `static final` reference instead of inlining calls like this.
```
```

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -568,6 +591,19 @@ private int getMaxElementsToReturn(AbstractSortedSetRangeOptions<?> rangeOptions
...
+    if (matchPattern != null) {
+      if (matchPattern.matcher(bytesToString(key)).matches()) {
+        resultList.add(key);
+        resultList.add(doubleToBytes(score));
+      }
+    } else {
+      resultList.add(key);
+      resultList.add(doubleToBytes(score));
+    }
+  }
```
kirklund, line 605: This change should be unit tested.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/key/AbstractScanExecutor.java
```diff
@@ -15,17 +15,120 @@
...
+import org.apache.geode.redis.internal.netty.ExecutionHandlerContext;
 
 public abstract class AbstractScanExecutor extends AbstractExecutor {
-
+  private static final Logger logger = LogService.getLogger();
   protected final BigInteger UNSIGNED_LONG_CAPACITY = new BigInteger("18446744073709551615");
   protected final int DEFAULT_COUNT = 10;
 
+  @Override
+  public RedisResponse executeCommand(Command command, ExecutionHandlerContext context) {
```
kirklund, line 55: There are a lot of changes here that should be unit tested.

kirklund (changes requested):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisHash.java
```diff
@@ -243,10 +243,9 @@ public int hstrlen(byte[] field) {
       throw new IllegalArgumentException("Requested array size exceeds VM limit");
     }
     List<byte[]> resultList = new ArrayList<>((int) maximumCapacity);
-    do {
-      cursor = hash.scan(cursor, 1,
-          (list, key, value) -> addIfMatching(matchPattern, list, key, value), resultList);
-    } while (cursor != 0 && resultList.size() < maximumCapacity);
+
+    cursor = hash.scan(cursor, count,
+        (list, key, value) -> addIfMatching(matchPattern, list, key, value), resultList);
```
upthewaterspout, line 248: I think this is going to underpopulate the resultList if there is a matchPattern. The scan method is only going to call `addIfMatching` for `maximumCapacity` iterations, rather than waiting until resultList has reached `maximumCapacity`.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/key/AbstractScanExecutor.java
```diff
@@ -15,25 +15,158 @@
...
     if (pattern == null) {
       return null;
     }
     return GlobPattern.compile(pattern);
   }
+  // Redis allows values for CURSOR up to UNSIGNED_LONG_CAPACITY, but internally it only makes sense
+  // for us to use values up to Integer.MAX_VALUE, so safely narrow any cursor values to int here
+
+  @VisibleForTesting
+  public int getIntCursor(String cursorString) {
```
upthewaterspout, line 148: How about we just use Integer.parseInt here (or Long.parseLong, if you prefer). Using BigInteger is a performance impact. Since the behavior is undetermined if the user passes as a cursor value other than 0 or a number that we previously returned, it seems like we could just treat numbers that are too large as 0 or integer max value or whatever. https://redis.io/commands/scan#calling-scan-with-a-corrupted-cursor

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisHash.java
DonalEvans: That's the behaviour that Redis has. See this comment: https://github.com/apache/geode/pull/6831#discussion_r702144208

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/key/AbstractScanExecutor.java
DonalEvans: In order to match Redis' error behaviour, we need to be able to tell the difference between a value for cursor that's not an integer of any kind ("123abc" for example), and a value that's larger than `Long.MAX_VALUE` but smaller than the maximum capacity of an unsigned long. Redis returns an error for the first case, but doesn't for the second (even though it's a nonsense value for the cursor). In order to check if the cursor value is in that second range, BigInteger is needed.

The alternative is to deviate from Redis' error behaviour here, but that's something we've avoided doing everywhere else. It's definitely an obscure corner case, since users realistically shouldn't ever be passing values for CURSOR greater than `Integer.MAX_VALUE`, let alone as big as an unsigned long, so maybe we're okay with having a difference here that we can document, where a cursor value greater than `Long.MAX_VALUE` returns an error, since we have no way to differentiate between a value that's too big and a value that's gibberish.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -69,6 +70,7 @@
 
 public class RedisSortedSet extends AbstractRedisData {
   protected static final int REDIS_SORTED_SET_OVERHEAD = memoryOverhead(RedisSortedSet.class);
+  public static final Logger logger = LogService.getLogger();
```
kirklund, line 73: You should probably make this private.

kirklund (approved):
One comment about the `Logger` scope if you want to follow up on it. Looks good!

If you ever want ideas about moving away from partial mocking with spies, hit me up.

PR #6832, opened 2021-09-02
GEODE:9556: make GETSET a supported command

ezoerner:


geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/RedisCommandType.java
```diff
@@ -187,6 +187,7 @@
   SET(new SetExecutor(), SUPPORTED, new MinimumParameterRequirements(3)),
   SETNX(new SetNXExecutor(), SUPPORTED, new ExactParameterRequirements(3)),
   STRLEN(new StrlenExecutor(), SUPPORTED, new ExactParameterRequirements(2)),
+  GETSET(new GetSetExecutor(), SUPPORTED, new ExactParameterRequirements(3)),
```
DonalEvans, line 190: Could this be moved to between `GET` and `INCRBY` to preserve the alphabetical ordering of commands?

dschneider-pivotal (approved):

DonalEvans (changes requested):
There's a test for GETSET in `AbstractHitsMissesIntegrationTest` that is currently in the "unsupported commands" section which should be moved up to be with the rest of the supported String commands.

Also, as a small clean-up, could the `REDIS_CLIENT_TIMEOUT` constant defined in `AbstractGetSetIntegrationTest` be removed and instead the constant with the same name in `RedisClusterStartupRule` be used? Almost all integration tests were updated to use the one in `RedisClusterStartupRule` but this one got missed I guess. For similar reasons, it would be good to use the `RedisClusterStartupRule#BIND_ADDRESS` constant in place of the hard-coded "localhost" in the `setUp()` method.

Finally, in the `AbstractGetSetIntegrationTest.testGetSet_shouldBeAtomic()` test, could the `ConcurrentLoopingThreads` class be used instead of manually creating a new executor pool? This class was introduced after this test was originally written, but all new tests use it now, so it would be good to update the test here to use it.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/RedisCommandType.java
ezoerner: Yes. By the way, I also updated the README file as I had missed that.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/string/AbstractGetSetIntegrationTest.java
```diff
@@ -120,41 +114,20 @@ public void testGetSet_whenWrongType_shouldReturnError() {
...
+
+    final AtomicInteger sumHolder = new AtomicInteger(0);
+    new ConcurrentLoopingThreads(ITERATION_COUNT,
+        (ignored) -> jedis.incr(contestedKey),
+        (i) -> {
+          // continue fetching with getSet until caught up with incrementing
+          // noinspection StatementWithEmptyBody
+          while (sumHolder.getAndAdd(Integer.parseInt(jedis.getSet(contestedKey, "0"))) <= i);
+        }).run();
+    assertThat(sumHolder.get()).isEqualTo(ITERATION_COUNT);
```
jdeppe-pivotal, line 130: It took me a moment to figure out the while here. Consider this as an alternative perhaps?
```
    final AtomicInteger sumHolder = new AtomicInteger(0);
    new ConcurrentLoopingThreads(ITERATION_COUNT,
        i -> jedis.incr(contestedKey),
        i -> sumHolder.getAndAdd(Integer.parseInt(jedis.getSet(contestedKey, "0"))))
        .runInLockstep();
    sumHolder.getAndAdd(Integer.parseInt(jedis.getSet(contestedKey, "0")));
    assertThat(sumHolder.get()).isEqualTo(ITERATION_COUNT);
```
The `runInLockstep` will ensure that each thread is in lockstep on the iteration so no one thread can "run ahead". Still needs a final `getSet` since either thread could complete first.

ezoerner: The while loop is there pretty much to make sure there is the final getSet, prevents the two threads from getting completely out of sync, and also to fit in better with the iterations of the ConcurrentLoopingThreads.   Keeping them in lockstep might be ok too, though it seems hard to say whether it would make it more or less likely to try and break the atomicity--i.e. for the incrementing thread to theoretically try to intervene between the get and set. I was hoping the comment above would make it a little clearer why the while loop is there.

DonalEvans (approved):

PR #6833, opened 2021-09-02
GEODE-9369: Command to copy region entries from a WAN site to another

albertogpz:
This PR contains the original commit of PR #6601 plus some improvements mainly on the testing as suggested by @kirklund .

### For all changes:

### Note:


geode-wan/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/WanCopyRegionCommandDUnitTest.java
```diff
@@ -0,0 +1,1417 @@
...
+
+    Integer senderLocatorPort = create3WanSitesAndClient(true, vm0,
+        vm1, vm2, serversInA, serverInB, serverInC, client,
+        senderIdInA, senderIdInB);
+
+    int wanCopyRegionBatchSize = 20;
+    String regionName = "foo";
+
+    // Execute wan-copy region command
+    GfshCommandRule gfsh = new GfshCommandRule();
```
kirklund, line 101: This should be declared as a `@Rule` and hoisted up to the top of the class. That way the `before` and `after` are invoked correctly. This mode of usage won't clean up properly.

If changing this to a `@Rule` doesn't seem to work properly in a WAN dunit test, then I would recommend trying to use non-GFSH command code here to set it up and then move these lines to a different test class that is specifically about testing the GFSH command.

geode-wan/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/WanCopyRegionCommandDUnitTest.java
```diff
@@ -0,0 +1,1417 @@
...
+  @Test
+  public void testUnsuccessfulExecution_SenderNotFound() throws Exception {
+    List<VM> serversInA = Arrays.asList(vm5, vm6, vm7);
+    VM serverInB = vm3;
+    VM serverInC = vm4;
+    VM client = vm8;
+    String senderIdInA = "B";
+    String senderIdInB = "C";
+
+    Integer senderLocatorPort = create3WanSitesAndClient(true, vm0,
```
kirklund, line 128: Here's another unnecessary boxing of an `int`.

geode-wan/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/WanCopyRegionCommandDUnitTest.java
```diff
@@ -0,0 +1,1417 @@
...
+  public void testUnsuccessfulExecution_ExceptionAtReceiver(
+      boolean isPartitionedRegion, boolean isParallelGatewaySender) throws Exception {
+    List<VM> serversInA = Arrays.asList(vm5, vm6, vm7);
+    VM serverInB = vm3;
+    VM serverInC = vm4;
+    VM client = vm8;
+    String senderIdInA = "B";
+    String senderIdInB = "C";
+
+    Integer senderLocatorPort = create3WanSitesAndClient(isPartitionedRegion, vm0,
```
kirklund, line 180: Another unnecessary boxing of `int`.

geode-wan/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/WanCopyRegionCommandDUnitTest.java
```diff
@@ -0,0 +1,1417 @@
...
+          "sender not primary");
+      command.hasTableSection(ResultModel.MEMBER_STATUS_SECTION).hasColumn("Message")
+          .asList().haveExactly(1, exceptionError).haveExactly(2, senderNotPrimary);
+    }
+  }
+
+  @Test
+  public void testSuccessfulExecution_WithReplicatedRegionAndSerialGatewaySender()
+      throws Exception {
+    testSuccessfulExecution(false, false);
```
kirklund, line 239: You might want to consider using `@RunWith(JUnitParamsRunner.class)` and then parameterizing some of these tests instead of using this old pattern. If you search for tests using `JUnitParamsRunner` you should find god examples.

geode-wan/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/WanCopyRegionCommandDUnitTest.java
```diff
@@ -0,0 +1,1417 @@
...
+    IgnoredException.addIgnoredException(
+        "Error closing the connection used to wan-copy region entries");
+  }
+
+  private void addIgnoredExceptionsForReceiverConnectedToSenderInUseWentDown() {
+    IgnoredException.addIgnoredException(
+        "Exception org.apache.geode.cache.client.internal.pooling.ConnectionDestroyedException in sendBatch. Retrying");
+    IgnoredException.addIgnoredException(
+        "Exception org.apache.geode.cache.client.ServerConnectivityException in sendBatch. Retrying");
+    IgnoredException.addIgnoredException("DistributedSystemDisconnectedException");
```
kirklund, line 421: This is ok but you can also use:
```

addIgnoredException(DistributedSystemDisconnectedException.class);
```

geode-wan/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/WanCopyRegionCommandDUnitTest.java
```diff
@@ -0,0 +1,1417 @@
...
+              .addOption(WAN_COPY_REGION__SENDERID, senderIdInA)
+              .addOption(WAN_COPY_REGION__BATCHSIZE, String.valueOf(wanCopyRegionBatchSize))
+              .addOption(WAN_COPY_REGION__MAXRATE, "50")
+              .getCommandString();
+          GfshCommandRule gfsh = new GfshCommandRule();
+          try {
+            gfsh.connectAndVerify(locatorAPort, GfshCommandRule.PortType.locator);
+          } catch (Exception e) {
+            e.printStackTrace();
+          }
```
kirklund, line 523: Don't use a catch-block like this. Try adding the `DistributedErrorCollector` rule instead and use it:
```
@Rule
public DistributedErrorCollector errorCollector = new DistributedErrorCollector();
```
```
try {
  gfsh.connectAndVerify(locatorAPort, GfshCommandRule.PortType.locator);
} catch (Exception e) {
  errorCollector.addError(e);
}
```
When the test ends (success or failure), JUnit will provide every exception stack collected in the `ErrorCollector` and any normal assertion failure as the cause of a test failure.

geode-wan/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/WanCopyRegionCommandDUnitTest.java
```diff
@@ -0,0 +1,1417 @@
...
+          wanCopyCommandStartLatch.countDown();
+          return gfsh.executeAndAssertThat(command);
+        });
+    LoggingExecutors.newSingleThreadExecutor(getTestMethodName(), true)
+        .submit(wanCopyCommandFuture);
```
kirklund, line 528: You'll need to do a bit more with this Executor in order to shut it down at the end of test. Rather than trying to code it yourself, I recommend using the `DistributedExecutorServiceRule` instead:
```
@Rule
public DistributedExecutorServiceRule executorServiceRule = new DistributedExecutorServiceRule();
```
Note sure which syntax will work best for submitting `wanCopyCommandFuture`. Probably just simply:
```
executorServiceRule.submit(wanCopyCommandFuture);
```
The rule will automatically shutdown and perform some error handling. It'll also dump executor threads if something hangs.

geode-wan/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/WanCopyRegionCommandDUnitTest.java
```diff
@@ -0,0 +1,1417 @@
...
+              .addOption(WAN_COPY_REGION__REGION, regionName)
+              .addOption(WAN_COPY_REGION__SENDERID, senderIdInA)
+              .addOption(WAN_COPY_REGION__BATCHSIZE, String.valueOf(wanCopyRegionBatchSize))
+              .addOption(WAN_COPY_REGION__MAXRATE, "5")
+              .getCommandString();
+          GfshCommandRule gfsh = new GfshCommandRule();
+          try {
+            gfsh.connectAndVerify(locatorAPort, GfshCommandRule.PortType.locator);
+          } catch (Exception e) {
+            e.printStackTrace();
```
kirklund, line 609: Another place to use `DistributedErrorCollector`.

geode-wan/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/WanCopyRegionCommandDUnitTest.java
```diff
@@ -0,0 +1,1417 @@
...
+          }
+          wanCopyCommandStartLatch.countDown();
+          return gfsh.executeAndAssertThat(command);
+        });
+    LoggingExecutors.newSingleThreadExecutor(getTestMethodName(), true)
+        .submit(wanCopyCommandFuture);
+
+    // Wait for the wan-copy command to start
+    wanCopyCommandStartLatch.await();
+    Thread.sleep(1000);
```
kirklund, line 532: These thread sleeps are actually pretty error prone. Your use of `CountDownLatch` is really good. See if you can figure out a way to add in one more latch or a `CyclicBarrier` or something like that.

geode-wan/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/WanCopyRegionCommandDUnitTest.java
```diff
@@ -0,0 +1,1417 @@
...
+            e.printStackTrace();
+          }
+          wanCopyCommandStartLatch.countDown();
+          return gfsh.executeAndAssertThat(command);
+        });
+    LoggingExecutors.newSingleThreadExecutor(getTestMethodName(), true)
+        .submit(wanCopyCommandFuture);
+
+    // Wait for the wan-copy command to start
+    wanCopyCommandStartLatch.await();
```
kirklund, line 531: You need to change all `CountDownLatch` awaits to use a timeout:
```
latch.await(GeodeAwaitility.getTimeout().getSeconds(), TimeUnit.SECONDS);
```
`import static` can help make it more readable.

geode-wan/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/WanCopyRegionCommandDUnitTest.java
```diff
@@ -0,0 +1,1417 @@
...
```
kirklund, line 618: Add the `GeodeAwaitility` timeout.

geode-wan/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/WanCopyRegionCommandDUnitTest.java
```diff
@@ -0,0 +1,1417 @@
...
```
kirklund, line 714: Add the `GeodeAwaitility` timeout.

geode-wan/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/WanCopyRegionCommandDUnitTest.java
```diff
@@ -0,0 +1,1417 @@
...
+    AsyncInvocation<Object> asyncOps3 =
+        client.invokeAsync(() -> sendRandomOpsFromClient(regionName, keySet, 10));
+
+    // Check command status and output
+    verifyStatusIsOk(gfsh.executeAndAssertThat(command));
+
+    // Wait for random operations to finish
+    asyncOps1.await();
+    asyncOps2.await();
+    asyncOps3.await();
```
kirklund, line 1080: Note: These are ok because `AsyncInvocation` is our code and `await()` actually uses the `GeodeAwaitility` timeout under the covers.

geode-wan/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/WanCopyRegionCommandDUnitTest.java
```diff
@@ -0,0 +1,1417 @@
...
+            e.printStackTrace();
+          }
+          wanCopyCommandStartLatch.countDown();
+          return gfsh.executeAndAssertThat(command).statusIsError();
+        });
+    LoggingExecutors.newSingleThreadExecutor(getTestMethodName(), true)
+        .submit(wanCopyCommandFuture);
+
+    // Wait for the wan-copy command to start
+    wanCopyCommandStartLatch.await();
```
kirklund, line 1199: Add the `GeodeAwaitility` timeout.

geode-wan/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/WanCopyRegionCommandDUnitTest.java
```diff
@@ -0,0 +1,1417 @@
...
+
+    return senderLocatorPort;
+  }
+
+  private String getRegionName(boolean isPartitionedRegion) {
+    return getTestMethodName() + (isPartitionedRegion ? "_PR" : "RR");
+  }
+
+  public static void removeEntry(String regionName, long key) {
+    Region<?, ?> r = ClientCacheFactory.getAnyInstance().getRegion(SEPARATOR + regionName);
```
kirklund, line 1329: Singleton call.

geode-wan/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/WanCopyRegionCommandDUnitTest.java
```diff
@@ -0,0 +1,1417 @@
...
+    assertNotNull(r);
+    r.remove(key);
+  }
+
+  public void sendRandomOpsFromClient(String regionName, Set<Long> keySet, int iterations) {
+    Region<Long, Integer> r = ClientCacheFactory.getAnyInstance().getRegion(SEPARATOR + regionName);
```
kirklund, line 1335: Singleton call.

kirklund (changes requested):

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/WanCopyRegionCommand.java
```diff
@@ -0,0 +1,84 @@
...
+      @CliOption(key = CliStrings.WAN_COPY_REGION__CANCEL,
+          unspecifiedDefaultValue = "false",
+          specifiedDefaultValue = "true",
+          help = CliStrings.WAN_COPY_REGION__CANCEL__HELP) boolean isCancel) {
+
+    authorize(Resource.DATA, Operation.WRITE, regionName);
+    final Object[] args = {regionName, senderId, isCancel, maxRate, batchSize};
+    ResultCollector<?, ?> resultCollector =
+        executeFunction(wanCopyRegionFunction, args, getAllNormalMembers());
+    @SuppressWarnings("unchecked")
```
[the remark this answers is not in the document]
albertogpz: Not anymore.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/WANTestBase.java
```diff
@@ -429,7 +434,7 @@ public static void createReplicatedRegion(String regionName, String senderIds, B
       }
 
       fact.setOffHeap(offHeap);
-      Region r = fact.create(regionName);
+      Region<Object, Object> r = fact.create(regionName);
```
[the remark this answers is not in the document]
albertogpz: I removed the method altogether as it was not needed.

geode-wan/src/distributedTest/java/org/apache/geode/internal/cache/wan/WANTestBase.java
```diff
@@ -2340,31 +2387,40 @@ public static void doPutsSameKey(String regionName, int numPuts, String key) {
...
+    Region<Object, Object> r = cache.getRegion(SEPARATOR + regionName);
+    assertNotNull(r);
+    for (long i = from; i < numPuts; i++) {
+      r.put(i, "Value_" + i);
+    }
+  }
+
+  public static void doClientPutsFrom(String regionName, int from, int numPuts) {
+    Region<Object, Object> r =
+        ClientCacheFactory.getAnyInstance().getRegion(SEPARATOR + regionName);
```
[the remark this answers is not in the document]
albertogpz: I have finally removed the method because it was not needed.

geode-wan/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/WanCopyRegionCommandDUnitTest.java
albertogpz: Good point. Done.
In `testSuccessfulCancelCommand()` I had to create an instance of `GfshCommandRule` to run the canceling of the command while an execution was ongoing. Using the same instance (the member instance) did not work in this case because the output of both commands did not come always in the same order. However I did not have to explicitely close the created instance because the member rule would close it (closes all) when the method finished.
Using two member rules did not work because there was a double close that provoked some `NullPointerException`.

geode-wan/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/WanCopyRegionCommandDUnitTest.java
albertogpz: Thanks for the suggestion!

geode-wan/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/WanCopyRegionCommandDUnitTest.java
albertogpz: I have opted for using a method in the `WanCopyRegionFunction` that provides the number of current executions.
That way I do not need the sleeps and neither the `CountDownLatch`.

geode-wan/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/WanCopyRegionCommandDUnitTest.java
albertogpz: Not applicable as I have removed the `CountDownLatch`.

geode-wan/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/WanCopyRegionCommandDUnitTest.java
albertogpz: Not applicable as there is no `CountDownLatch` anymore.

geode-wan/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/WanCopyRegionCommandDUnitTest.java
albertogpz: Not applicable as there is no `CountDownLatch` anymore.

onichols-pivotal (approved):

davebarnes97 (approved):

mhansonp (approved):

DonalEvans (approved):
Approving changes in files for which I'm a codeowner.

geode-wan/src/main/java/org/apache/geode/management/internal/cli/functions/WanCopyRegionFunction.java
```diff
@@ -0,0 +1,581 @@
...
+
+  private CliFunctionResult executeWanCopyRegionFunctionInNewThread(
+      FunctionContext<Object[]> context,
+      Region<?, ?> region, String regionName, GatewaySender sender, long maxRate, int batchSize)
+      throws InterruptedException, ExecutionException, CancellationException {
+    String executionName = getExecutionName(regionName, sender.getId());
+    CompletableFuture<CliFunctionResult> future = null;
+    Executor executor = executorFactory.getExecutor((InternalCache) context.getCache());
+    try {
+      synchronized (executions) {
```
kirklund, line 217: Since `executions` would be moved to the service, I think you should only synchronize on that field over in the service. You could potentially pass a `Runnable` and a `String executionName` over to the service to be executed under synchronization. The service would then handle all use of the `executions` and probably the `executor`. Then the only thing this function would need is to invoke the method for executing tasks over in the service.

The code is then divided into two places. Everything that can be performed by multiple threads and multiple instances lives in `WanCopyRegionFunction`, while everything that must be one instance or fixed threads per Cache would live in `WanCopyRegionFunctionService`.

kirklund (changes requested):
I've pointed out some things you can change to move all of your changes into geode-wan instead of having any code changes in geode-core and geode-gfsh.

I also think you should change the function to delegate execution tasks to the service.

geode-wan/src/main/java/org/apache/geode/management/internal/cli/functions/WanCopyRegionFunction.java
kirklund: An example of submitting `Runnables` to an execute method can be found in `AlertingAction` in geode-core. The purpose of `AlertingAction` is to surround the call to `run()` within a `ThreadLocal` but this could just as easily submit the `Runnable` to an `Executor` as well with related usage of your `executions` map.

jinmeiliao (approved):

albertogpz:
> I've pointed out some things you can change to move all of your changes into geode-wan instead of having any code changes in geode-core and geode-gfsh.
> 
> I also think you should change the function to delegate execution tasks to the service.

Thanks for the suggestions. They are very good.

geode-core/src/distributedTest/java/org/apache/geode/management/internal/security/MultiGfshDUnitTest.java
```diff
@@ -52,6 +52,9 @@ public void testMultiUser() throws Exception {
     IgnoredException.addIgnoredException("java.util.zip.ZipException: zip file is empty");
     IgnoredException
         .addIgnoredException("java.lang.IllegalStateException: WAN service is not available.");
+    IgnoredException
+        .addIgnoredException(
+            "org.apache.geode.internal.cache.wan.GatewaySenderConfigurationException: Locators must be configured before starting gateway-sender.");
     int jmxPort = server.getJmxPort();
```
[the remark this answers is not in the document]
albertogpz: It was needed at some point in the development. But not anymore. I have removed it.

geode-wan/src/main/java/org/apache/geode/cache/wan/internal/WanCopyRegionFunctionService.java
```diff
@@ -0,0 +1,68 @@
...
+      }
+    } catch (InterruptedException ie) {
+      wanCopyRegionFunctionExecutionPool.shutdownNow();
+      Thread.currentThread().interrupt();
+    }
+  }
+
+  public Executor getExecutor() {
+    return wanCopyRegionFunctionExecutionPool;
+  }
```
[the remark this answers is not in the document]
albertogpz: I've made it volatile, which I think achieves the same purpose.

geode-wan/src/main/java/org/apache/geode/management/internal/cli/functions/WanCopyRegionFunction.java
```diff
@@ -0,0 +1,581 @@
...
+  }
+
+  @VisibleForTesting
+  WanCopyRegionFunction(Clock clock, ThreadSleeper threadSleeper,
+      WanCopyRegionFunctionExecutorFactory executorFactory, EventCreator eventCreator) {
+    this.clock = clock;
+    this.threadSleeper = threadSleeper;
+    this.executorFactory = executorFactory;
+    this.eventCreator = eventCreator;
+  }
```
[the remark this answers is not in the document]
albertogpz: I like the change but there were a problems when deserializing the two lambdas passed. So, in order to avoid them, I had to create a ThreadSleeperImpl and a WanCopyRegionFuncionServiceProviderImpl classes with the above's labmdas behaviors.

geode-wan/src/main/java/org/apache/geode/management/internal/cli/functions/WanCopyRegionFunction.java
albertogpz: Nice change!
I have also split the `WanCopyRegionFunction` into two so that the new one `WanCopyRegionFunctionDelegate` is in charge of copying the entries.
The original one makes all the checks before invoking the new class and is also in charge of executing cancels via de new service provider.

rhoughton-pivot (approved):
Build changes seem good.

kirklund:
@albertogpz Quick note to let you I'm reviewing this.

geode-wan/src/distributedTest/java/org/apache/geode/cache/wan/internal/cli/commands/WanCopyRegionCommandDUnitTest.java
```diff
@@ -0,0 +1,1267 @@
...
+      Condition<String> exceptionError =
+          new Condition<>(s -> s.startsWith("Error ("), "Error");
+      Condition<String> senderNotPrimary = new Condition<>(
+          s -> s.equals(CliStrings
+              .format(WAN_COPY_REGION__MSG__SENDER__SERIAL__AND__NOT__PRIMARY,
+                  senderIdInA)),
+          "sender not primary");
+      command.hasTableSection(ResultModel.MEMBER_STATUS_SECTION).hasColumn("Message")
+          .asList().haveExactly(1, exceptionError).haveExactly(2, senderNotPrimary);
+    }
```
kirklund, line 236: [comment] Not a request for changes, but I generally try to format these chained APIs with linefeeds like:
```
command
    .hasTableSection(ResultModel.MEMBER_STATUS_SECTION)
    .hasColumn("Message")
    .asList()
    .haveExactly(1, exceptionError)
    .haveExactly(2, senderNotPrimary);
```
I'm not sure if it's easier to read for everyone, but it is for me.

geode-wan/src/distributedTest/java/org/apache/geode/cache/wan/internal/cli/commands/WanCopyRegionCommandDUnitTest.java
```diff
@@ -0,0 +1,1267 @@
...
+        .addOption(WAN_COPY_REGION__BATCHSIZE, String.valueOf(wanCopyRegionBatchSize))
+        .getCommandString();
+
+    // While the command is running, send some random operations over the same keys
+    AsyncInvocation<Object> asyncOps1 =
+        client.invokeAsync(() -> sendRandomOpsFromClient(regionName, keySet, 10));
+    AsyncInvocation<Object> asyncOps2 =
+        client.invokeAsync(() -> sendRandomOpsFromClient(regionName, keySet, 10));
+    AsyncInvocation<Object> asyncOps3 =
+        client.invokeAsync(() -> sendRandomOpsFromClient(regionName, keySet, 10));
```
kirklund, line 706: Since you don't actually need a return from these (you just call `await()` down below), you should just use the `Void` type: `AsyncInvocation<Void>`.

geode-wan/src/distributedTest/java/org/apache/geode/cache/wan/internal/cli/commands/WanCopyRegionCommandDUnitTest.java
```diff
@@ -0,0 +1,1267 @@
...
+    command.hasTableSection(ResultModel.MEMBER_STATUS_SECTION).hasColumn("Message")
+        .asList().haveExactly(1, haveEntriesCopied).haveExactly(2, senderNotPrimary);
+  }
+
+  private String getRegionName(boolean isPartitionedRegion) {
+    return getTestMethodName() + (isPartitionedRegion ? "_PR" : "RR");
+  }
+
+  public static void removeEntry(String regionName, long key) {
+    Region<?, ?> r = cache.getRegion(SEPARATOR + regionName);
```
kirklund, line 1133: You should go ahead and name any variables with a full word like `region`.

geode-wan/src/test/java/org/apache/geode/cache/wan/internal/WanCopyRegionFunctionServiceTest.java
```diff
@@ -0,0 +1,184 @@
...
+    };
+
+    String regionName = "myRegion";
+    String senderId = "mySender";
+    CompletableFuture
+        .supplyAsync(() -> {
+          try {
+            return service.execute(firstExecution, regionName, senderId);
+          } catch (Exception e) {
+            return null;
```
kirklund, line 59: [comment] Returning null is probably ok. Or you could use an `ErrorCollector` JUnit rule here as well.

geode-wan/src/test/java/org/apache/geode/management/internal/cli/functions/WanCopyRegionFunctionDelegateTest.java
```diff
@@ -0,0 +1,526 @@
...
+  private Connection connectionMock;
+  private GatewaySenderEventDispatcher dispatcherMock;
+
+  private final Region<?, ?> regionMock =
+      uncheckedCast(mock(InternalRegion.class));
+
+  private EventCreator eventCreatorMock;
+
+  @Before
+  public void setUp() throws InterruptedException {
```
kirklund, line 84: [comment] Some of these big methods are easier to read if you group all of the mocks together at the top, and then all of the stubbings (when statements) down lower.

geode-wan/src/test/java/org/apache/geode/management/internal/cli/functions/WanCopyRegionFunctionDelegateTest.java
```diff
@@ -0,0 +1,526 @@
...
+      throws InterruptedException, BatchException70 {
+    Object[] options = new Object[] {"myRegion", "mySender", false, 1L, 2};
+    WanCopyRegionFunctionDelegate.ThreadSleeper sleeperMock =
+        mock(WanCopyRegionFunctionDelegate.ThreadSleeper.class);
+    doNothing().when(sleeperMock).sleep(anyLong());
+
+    executeWanCopyRegionFunction(options, sleeperMock);
+    ArgumentCaptor<Long> captor = ArgumentCaptor.forClass(Long.class);
+    verify(sleeperMock, times(1)).sleep(captor.capture());
+    assertThat(captor.getValue()).isGreaterThan(0).isLessThanOrEqualTo(2000);
```
kirklund, line 144: [comment] Another way to organize these tests for (hopefully) easier reading is to group it in arrange/act/assert sections:
```
    // arrange
    Object[] options = new Object[] {"myRegion", "mySender", false, 1L, 2};
    WanCopyRegionFunctionDelegate.ThreadSleeper sleeperMock =
        mock(WanCopyRegionFunctionDelegate.ThreadSleeper.class);
    doNothing().when(sleeperMock).sleep(anyLong());

    // act
    executeWanCopyRegionFunction(options, sleeperMock);

    // assert
    ArgumentCaptor<Long> captor = ArgumentCaptor.forClass(Long.class);
    verify(sleeperMock, times(1)).sleep(captor.capture());
    assertThat(captor.getValue()).isGreaterThan(0).isLessThanOrEqualTo(2000);
```

geode-wan/src/test/java/org/apache/geode/management/internal/cli/functions/WanCopyRegionFunctionDelegateTest.java
```diff
@@ -0,0 +1,526 @@
...
+    when(poolMock.acquireConnection()).thenThrow(NoAvailableServersException.class)
+        .thenThrow(NoAvailableServersException.class);
+    Set<Region.Entry<Object, Object>> entries = new HashSet<>();
+    entries.add(uncheckedCast(mock(Region.Entry.class)));
+    when(regionMock.entrySet()).thenReturn(uncheckedCast(entries));
+
+    assertThatThrownBy(
+        () -> function.wanCopyRegion(internalCacheMock, "member1", regionMock, gatewaySenderMock, 1,
+            10))
+                .isInstanceOf(NoAvailableServersException.class);
```
kirklund, line 238: [comment] If you decide you like arrange/act/assert then you can use this syntax if you want:
```
// arrange
...

// act
Throwable thrown = catchThrowable(() -> {
  function.wanCopyRegion(internalCacheMock, "member1", regionMock, gatewaySenderMock, 1, 10);
});

// assert
assertThat(thrown).isInstanceOf(NoAvailableServersException.class);
```

geode-wan/src/test/java/org/apache/geode/cache/wan/internal/WanCopyRegionFunctionServiceTest.java
```diff
@@ -0,0 +1,184 @@
...
+import org.junit.Before;
+import org.junit.Test;
+
+import org.apache.geode.internal.cache.InternalCache;
+import org.apache.geode.management.internal.functions.CliFunctionResult;
+
+public class WanCopyRegionFunctionServiceTest {
+
+  private WanCopyRegionFunctionService service;
+  InternalCache cache = mock(InternalCache.class);
```
kirklund, line 35: `private InternalCache cache = mock(InternalCache.class);`

geode-wan/src/test/java/org/apache/geode/cache/wan/internal/WanCopyRegionFunctionServiceTest.java
```diff
@@ -0,0 +1,183 @@
...
+            try {
+              return service.execute(firstExecution, regionName, "mySender1");
+            } catch (Exception e) {
+              return null;
+            }
+          });
+    }
+
+    // Wait for the functions to start execution
+    await().until(() -> service.getNumberOfCurrentExecutions() == executions);
```
kirklund, line 176: Another one where I'd use `untilAsserted` with an assertion so you see the exact number if it fails.

geode-wan/src/test/java/org/apache/geode/cache/wan/internal/WanCopyRegionFunctionServiceTest.java
```diff
@@ -0,0 +1,183 @@
...
+    await().untilAsserted(() -> assertThat(service.getNumberOfCurrentExecutions()).isEqualTo(0));
+  }
+
+  @Test
+  public void severalExecuteWithDifferentRegionOrSenderAreAllowed() {
+    int executions = 5;
+    CountDownLatch latch = new CountDownLatch(executions);
+    for (int i = 0; i < executions; i++) {
+      Callable<CliFunctionResult> firstExecution = () -> {
+        latch.await();
```
kirklund, line 160: Since `CountDownLatch` await waits forever by default, you should change it to:
```
latch.await(GeodeAwaitility.getTimeout().getSeconds(), TimeUnit.SECONDS);
```

geode-wan/src/test/java/org/apache/geode/management/internal/cli/functions/WanCopyRegionFunctionDelegateTest.java
```diff
@@ -0,0 +1,582 @@
...
+    threadSleeperMock = mock(WanCopyRegionFunctionDelegate.ThreadSleeper.class);
+    gatewaySenderMock = mock(AbstractGatewaySender.class);
+    poolMock = mock(PoolImpl.class);
+    connectionMock = mock(PooledConnection.class);
+    dispatcherMock = mock(GatewaySenderEventDispatcher.class);
+    eventCreatorMock = mock(EventCreator.class);
+    AbstractGatewaySenderEventProcessor eventProcessorMock =
+        mock(AbstractGatewaySenderEventProcessor.class);
+    RegionAttributes<?, ?> attributesMock = mock(RegionAttributes.class);
+    Set<?> idsMock = mock(Set.class);
```
kirklund, line 96: [comments] This is ok but `Collections.emptySet()` might be better to use since the general recommendation is to not mock classes from other libraries.

geode-wan/src/test/java/org/apache/geode/management/internal/cli/functions/WanCopyRegionFunctionDelegateTest.java
```diff
@@ -0,0 +1,582 @@
...
+    function = new WanCopyRegionFunctionDelegate(clockMock, threadSleeperMock, eventCreatorMock);
+
+    doNothing().when(threadSleeperMock).sleep(anyLong());
+    when(gatewaySenderMock.getId()).thenReturn("mySender");
+    when(connectionMock.getWanSiteVersion()).thenReturn(KnownVersion.GEODE_1_15_0.ordinal());
+    when(eventCreatorMock.createGatewaySenderEvent(any(), any(), any(), any()))
+        .thenReturn(uncheckedCast(mock(GatewayQueueEvent.class)));
+    when(eventProcessorMock.getDispatcher()).thenReturn(dispatcherMock);
+    when(((AbstractGatewaySender) gatewaySenderMock).getEventProcessor())
+        .thenReturn(eventProcessorMock);
```
kirklund, line 108: I think the interface `InternalGatewaySender` should work here instead of `AbstractGatewaySender`.

geode-wan/src/test/java/org/apache/geode/management/internal/cli/functions/WanCopyRegionFunctionDelegateTest.java
```diff
@@ -0,0 +1,582 @@
...
+    long maxRate = 1;
+    long elapsedTime = 20L;
+
+    // arrange
+    when(clockMock.millis()).thenReturn(startTime + elapsedTime);
+    // Sleep so that the sleep can be interrupted
+    doAnswer(invocation -> {
+      Thread.sleep(1);
+      return null;
+    }).when(threadSleeperMock).sleep(anyLong());
```
kirklund, line 158: This test is actually validating that Thread.sleep throws InterruptedException when interrupted, but you don't need (or want) to test the Java library, just your own classes. So this would be a place where it's better to just stub threadSleeperMock to throw InterruptedException:
```
  @Test
  public void doPostSendBatchActions_ThrowInterruptedIfInterruptedTimeToSleepNotZero()
      throws InterruptedException {
    // arrange
    long maxRate = 1;
    long elapsedTime = 20L;
    InterruptedException thrownByThreadSleeper = new InterruptedException("thrownByThreadSleeper");
    when(clockMock.millis()).thenReturn(startTime + elapsedTime);
    doThrow(thrownByThreadSleeper).when(threadSleeperMock).sleep(anyLong());

    // act
    Throwable thrown =
        catchThrowable(() -> function.doPostSendBatchActions(startTime, entries, maxRate));

    // assert
    assertThat(thrown).isInstanceOf(InterruptedException.class);
  }
```

kirklund (changes requested):
You should change any unit test that is validating that Thread.sleep throws InterruptedException to only test your class' interaction with ThreadSleeper. I inlined an example for one of the tests.

That `Thread.sleep(1)` is very prone to flakiness.

PR #6835, opened 2021-09-02
GEODE-9457: re-authentication in event dispatcher

jinmeiliao:
GEODE-9453: clean up the old subject when new subject is authenticated.

* authorize the message before dispatching the message to the client
* add a new message type to be sent to the client for re-authentiate
* message dispatcher will wait for a certain time for client to reauthenticate before terminating the client
* when re-authenticate, the new subject will re-use the original uniqueId. Credentials will be sent to the user along with the old uniqueId if exists.
* old subject will be cleaned out when new subject re-authenticate back.
* re-auth multi-user mode in event dispatching is not supported (yet).

I will continue to write unit tests for those refactored classes

### DIFF
--- a/geode-core/src/distributedTest/java/org/apache/geode/management/internal/security/MultiUserAuthenticationDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/management/internal/security/MultiUserAuthenticationDUnitTest.java
@@ -16,8 +16,12 @@
 package org.apache.geode.management.internal.security;
 
 import static org.apache.geode.cache.Region.SEPARATOR;
+import static org.apache.geode.test.awaitility.GeodeAwaitility.await;
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Properties;
 
 import org.junit.BeforeClass;
@@ -31,6 +35,14 @@ import org.apache.geode.cache.RegionService;
 import org.apache.geode.cache.client.ClientCache;
 import org.apache.geode.cache.client.ClientRegionShortcut;
 import org.apache.geode.cache.client.ServerOperationException;
+import org.apache.geode.cache.query.CqAttributesFactory;
+import org.apache.geode.cache.query.CqEvent;
+import org.apache.geode.cache.query.CqException;
+import org.apache.geode.cache.query.CqExistsException;
+import org.apache.geode.cache.query.CqListener;
+import org.apache.geode.cache.query.CqQuery;
+import org.apache.geode.cache.query.QueryService;
+import org.apache.geode.cache.query.RegionNotFoundException;
 import org.apache.geode.examples.SimpleSecurityManager;
 import org.apache.geode.test.dunit.IgnoredException;
 import org.apache.geode.test.dunit.rules.ClusterStartupRule;
@@ -98,6 +110,66 @@ public class MultiUserAuthenticationDUnitTest {
       regionService1.close();
       regionService2.close();
       cache.close();
+    }
+  }
+
+  @Test
+  public void multiUserCQ() throws Exception {
+    int locatorPort = locator.getPort();
+    client.withCacheSetup(f -> f.setPoolSubscriptionEnabled(true)
+        .setPoolMultiuserAuthentication(true)
+        .addPoolLocator("localhost", locatorPort))
+        .createCache();
+
+    // both are able to read data
+    RegionService regionService1 = client.createAuthenticatedView("data", "data");
+    RegionService regionService2 = client.createAuthenticatedView("dataRead", "dataRead");
+
+    EventsCqListner listener1 = createAndExecuteCQ(regionService1.getQueryService(), "cq1",
+        "select * from /region r where r.length<=2");
+    EventsCqListner listener2 = createAndExecuteCQ(regionService2.getQueryService(), "cq2",
+        "select * from /region r where r.length>=2");
+
+    // put 3 data in the region
+    gfsh.executeAndAssertThat("put --region=region --key=1 --value=1");
+    gfsh.executeAndAssertThat("put --region=region --key=11 --value=11");
+    gfsh.executeAndAssertThat("put --region=region --key=111 --value=111");
+
+    await().untilAsserted(
+        () -> assertThat(listener1.getKeys())
+            .containsExactly("1", "11"));
+
+    await().untilAsserted(
+        () -> assertThat(listener2.getKeys())
+            .containsExactly("11", "111"));
+
+  }
+
+  private static EventsCqListner createAndExecuteCQ(QueryService queryService, String cqName,
+      String query)
+      throws CqExistsException, CqException, RegionNotFoundException {
+    CqAttributesFactory cqaf = new CqAttributesFactory();
+    EventsCqListner listener = new EventsCqListner();
+    cqaf.addCqListener(listener);
+
+    CqQuery cq = queryService.newCq(cqName, query, cqaf.create());
+    cq.execute();
+    return listener;
+  }
+
+  private static class EventsCqListner implements CqListener {
+    private List<String> keys = new ArrayList<>();
+
+    @Override
+    public void onEvent(CqEvent aCqEvent) {
+      keys.add(aCqEvent.getKey().toString());
+    }
+
+    @Override
+    public void onError(CqEvent aCqEvent) {}
+
+    public List<String> getKeys() {
+      return keys;
     }
   }
 }
--- a/geode-core/src/main/java/org/apache/geode/cache/client/internal/AuthenticateUserOp.java
+++ b/geode-core/src/main/java/org/apache/geode/cache/client/internal/AuthenticateUserOp.java
@@ -23,7 +23,6 @@ import org.apache.geode.DataSerializer;
 import org.apache.geode.InternalGemFireError;
 import org.apache.geode.cache.client.ServerOperationException;
 import org.apache.geode.distributed.DistributedMember;
-import org.apache.geode.distributed.DistributedSystem;
 import org.apache.geode.distributed.internal.InternalDistributedSystem;
 import org.apache.geode.distributed.internal.ServerLocation;
 import org.apache.geode.distributed.internal.membership.InternalDistributedMember;
@@ -54,6 +53,7 @@ import org.apache.geode.security.NotAuthorizedException;
  * @since GemFire 6.5
  */
 public class AuthenticateUserOp {
+  public static final long NOT_A_USER_ID = -1L;
 
   /**
    * Sends the auth credentials to the server. Used in single user mode of authentication.
@@ -63,8 +63,21 @@ public class AuthenticateUserOp {
    * @return Object unique user-id.
    */
   public static Long executeOn(Connection con, ExecutablePool pool) {
-    AbstractOp op = new AuthenticateUserOpImpl(con);
+    AbstractOp op = new AuthenticateUserOpImpl();
     return (Long) pool.executeOn(con, op);
+  }
+
+  /**
+   * Sends the auth credentials to the server. Used in single user mode of authentication.
+   *
+   * @param location The ServerLocation instance whose connection instance will be used to perform
+   *        the operation.
+   * @param pool The connection pool to use for this operation.
+   * @return Object unique user-id.
+   */
+  public static Long executeOn(ServerLocation location, ExecutablePool pool) {
+    AbstractOp op = new AuthenticateUserOpImpl();
+    return (Long) pool.executeOn(location, op);
   }
 
   /**
@@ -87,31 +100,12 @@ public class AuthenticateUserOp {
   }
 
   static class AuthenticateUserOpImpl extends AbstractOp {
-
     private Properties securityProperties = null;
     private boolean needsServerLocation = false;
 
-    AuthenticateUserOpImpl(Connection con) {
+    AuthenticateUserOpImpl() {
       super(MessageType.USER_CREDENTIAL_MESSAGE, 1);
-      byte[] credentialBytes;
-      DistributedMember server = new InternalDistributedMember(con.getSocket().getInetAddress(),
-          con.getSocket().getPort(), false);
-      DistributedSystem sys = InternalDistributedSystem.getConnectedInstance();
-      String authInitMethod = sys.getProperties().getProperty(SECURITY_CLIENT_AUTH_INIT);
-      Properties tmpSecurityProperties = sys.getSecurityProperties();
-
-      // LOG: following passes the DS API LogWriters into the security API
-      Properties credentials = Handshake.getCredentials(authInitMethod, tmpSecurityProperties,
-          server, false, sys.getLogWriter(), sys.getSecurityLogWriter());
-
       getMessage().setMessageHasSecurePartFlag();
-      try (HeapDataOutputStream heapdos = new HeapDataOutputStream(KnownVersion.CURRENT)) {
-        DataSerializer.writeProperties(credentials, heapdos);
-        credentialBytes = ((ConnectionImpl) con).encryptBytes(heapdos.toByteArray());
-      } catch (Exception e) {
-        throw new ServerOperationException(e);
-      }
-      getMessage().addBytesPart(credentialBytes);
     }
 
     AuthenticateUserOpImpl(Properties securityProps) {
@@ -123,50 +117,74 @@ public class AuthenticateUserOp {
       super(MessageType.USER_CREDENTIAL_MESSAGE, 1);
       securityProperties = securityProps;
       needsServerLocation = needsServer;
-
       getMessage().setMessageHasSecurePartFlag();
     }
 
     @Override
-    protected void sendMessage(Connection cnx) throws Exception {
-      HeapDataOutputStream hdos = new HeapDataOutputStream(KnownVersion.CURRENT);
-      byte[] secureBytes;
-      hdos.writeLong(cnx.getConnectionID());
-      if (securityProperties != null) {
-        DistributedMember server = new InternalDistributedMember(cnx.getSocket().getInetAddress(),
-            cnx.getSocket().getPort(), false);
-        DistributedSystem sys = InternalDistributedSystem.getConnectedInstance();
-        String authInitMethod = sys.getProperties().getProperty(SECURITY_CLIENT_AUTH_INIT);
+    protected void sendMessage(Connection connection) throws Exception {
+      if (securityProperties == null) {
+        securityProperties = getConnectedSystem().getSecurityProperties();
+      }
+      byte[] credentialBytes = getCredentialBytes(connection, securityProperties);
+      getMessage().addBytesPart(credentialBytes);
 
-        Properties credentials = Handshake.getCredentials(authInitMethod, securityProperties,
-            server, false, sys.getLogWriter(), sys.getSecurityLogWriter());
-        byte[] credentialBytes;
-        try (HeapDataOutputStream heapdos = new HeapDataOutputStream(KnownVersion.CURRENT)) {
-          DataSerializer.writeProperties(credentials, heapdos);
-          credentialBytes = ((ConnectionImpl) cnx).encryptBytes(heapdos.toByteArray());
-        }
-        getMessage().addBytesPart(credentialBytes);
+      try (HeapDataOutputStream hdos = new HeapDataOutputStream(16, KnownVersion.CURRENT)) {
+        hdos.writeLong(connection.getConnectionID());
+        hdos.writeLong(getUserId(connection));
+        getMessage().setSecurePart(((ConnectionImpl) connection).encryptBytes(hdos.toByteArray()));
       }
-      try {
-        secureBytes = ((ConnectionImpl) cnx).encryptBytes(hdos.toByteArray());
-      } finally {
-        hdos.close();
-      }
-      getMessage().setSecurePart(secureBytes);
       getMessage().send(false);
     }
 
+    protected long getUserId(Connection connection) {
+      // single user mode
+      if (UserAttributes.userAttributes.get() == null) {
+        return connection.getServer().getUserId();
+      }
+      // multi user mode
+      Long id = UserAttributes.userAttributes.get().getServerToId().get(connection.getServer());
+      if (id == null) {
+        return NOT_A_USER_ID;
+      }
+      return id;
+    }
+
+    protected InternalDistributedSystem getConnectedSystem() {
+      return InternalDistributedSystem.getConnectedInstance();
+    }
+
+    protected byte[] getCredentialBytes(Connection connection, Properties securityProperties)
+        throws Exception {
+      InternalDistributedSystem distributedSystem = getConnectedSystem();
+      DistributedMember server =
+          new InternalDistributedMember(connection.getSocket().getInetAddress(),
+              connection.getSocket().getPort(), false);
+      String authInitMethod =
+          distributedSystem.getProperties().getProperty(SECURITY_CLIENT_AUTH_INIT);
+
+      Properties credentials = Handshake.getCredentials(authInitMethod, securityProperties,
+          server, false, distributedSystem.getLogWriter(),
+          distributedSystem.getSecurityLogWriter());
+      byte[] credentialBytes;
+      try (HeapDataOutputStream heapdos = new HeapDataOutputStream(KnownVersion.CURRENT)) {
+        DataSerializer.writeProperties(credentials, heapdos);
+        credentialBytes = ((ConnectionImpl) connection).encryptBytes(heapdos.toByteArray());
+      }
+      return credentialBytes;
+    }
+
+
     @Override
-    public Object attempt(Connection cnx) throws Exception {
-      if (cnx.getServer().getRequiresCredentials()) {
-        return super.attempt(cnx);
+    public Object attempt(Connection connection) throws Exception {
+      if (connection.getServer().getRequiresCredentials()) {
+        return super.attempt(connection);
       } else {
         return null;
       }
     }
 
     @Override
-    protected Object processResponse(Message msg, Connection cnx) throws Exception {
+    protected Object processResponse(Message msg, Connection connection) throws Exception {
       byte[] bytes;
       Part part = msg.getPart(0);
       final int msgType = msg.getMessageType();
@@ -174,16 +192,16 @@ public class AuthenticateUserOp {
       if (msgType == MessageType.RESPONSE) {
         bytes = (byte[]) part.getObject();
         if (bytes.length == 0) {
-          cnx.getServer().setRequiresCredentials(false);
+          connection.getServer().setRequiresCredentials(false);
         } else {
-          cnx.getServer().setRequiresCredentials(true);
-          byte[] decrypted = ((ConnectionImpl) cnx).decryptBytes(bytes);
+          connection.getServer().setRequiresCredentials(true);
+          byte[] decrypted = ((ConnectionImpl) connection).decryptBytes(bytes);
           try (ByteArrayDataInput dis = new ByteArrayDataInput(decrypted)) {
             userId = dis.readLong();
           }
         }
         if (needsServerLocation) {
-          return new Object[] {cnx.getServer(), userId};
+          return new Object[] {connection.getServer(), userId};
         } else {
           return userId;
         }
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/CacheClientProxy.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/CacheClientProxy.java
@@ -216,11 +216,14 @@ public class CacheClientProxy implements ClientSession {
    */
   boolean keepalive = false;
 
+  /**
+   * for single user environment
+   */
   private AccessControl postAuthzCallback;
   private Subject subject;
 
   /**
-   * For multiuser environment..
+   * used for cq name to subject/auth mapping, always initialized in single/multi user casees
    */
   private ClientUserAuths clientUserAuths;
 
@@ -345,19 +348,15 @@ public class CacheClientProxy implements ClientSession {
     initializeClientAuths();
   }
 
-  private void initializeClientAuths() {
-    if (AcceptorImpl.isPostAuthzCallbackPresent()) {
-      clientUserAuths = ServerConnection.getClientUserAuths(proxyID);
-    }
+  void initializeClientAuths() {
+    clientUserAuths = ServerConnection.getClientUserAuths(proxyID);
   }
 
   private void reinitializeClientAuths() {
-    if (clientUserAuths != null && AcceptorImpl.isPostAuthzCallbackPresent()) {
-      synchronized (clientUserAuthsLock) {
-        ClientUserAuths newClientAuth = ServerConnection.getClientUserAuths(proxyID);
-        newClientAuth.fillPreviousCQAuth(clientUserAuths);
-        clientUserAuths = newClientAuth;
-      }
+    synchronized (clientUserAuthsLock) {
+      ClientUserAuths newClientAuth = ServerConnection.getClientUserAuths(proxyID);
+      newClientAuth.fillPreviousCQAuth(clientUserAuths);
+      clientUserAuths = newClientAuth;
     }
   }
 
@@ -372,7 +371,6 @@ public class CacheClientProxy implements ClientSession {
   }
 
   public void setSubject(Subject subject) {
-    // TODO:hitesh synchronization
     synchronized (clientUserAuthsLock) {
       if (this.subject != null) {
         this.subject.logout();
@@ -381,13 +379,14 @@ public class CacheClientProxy implements ClientSession {
     }
   }
 
-  public void setCQVsUserAuth(String cqName, long uniqueId, boolean isDurable) {
-    if (postAuthzCallback == null) // only for multiuser
-    {
-      if (clientUserAuths != null) {
-        clientUserAuths.setUserAuthAttributesForCq(cqName, uniqueId, isDurable);
-      }
+  protected Subject getSubject(String cqName) {
+    synchronized (clientUserAuthsLock) {
+      return clientUserAuths.getSubject(cqName);
     }
+  }
+
+  public void setCQVsUserAuth(String cqName, long uniqueId, boolean isDurable) {
+    clientUserAuths.setUserAuthAttributesForCq(cqName, uniqueId, isDurable);
   }
 
   private void initializeTransientFields(Socket socket, ClientProxyMembershipID pid, boolean ip,
@@ -782,7 +781,8 @@ public class CacheClientProxy implements ClientSession {
         if (postAuthzCallback != null) {// for single user
           postAuthzCallback.close();
           postAuthzCallback = null;
-        } else if (clientUserAuths != null) {// for multiple users
+        }
+        if (clientUserAuths != null) {// for multiple users
           clientUserAuths.cleanup(true);
           clientUserAuths = null;
         }
@@ -1219,14 +1219,6 @@ public class CacheClientProxy implements ClientSession {
     _cacheClientNotifier.deliverInterestChange(proxyID, message);
   }
 
-  /*
-   * protected void addFilterRegisteredClients(String regionName, Object keyOfInterest) { try {
-   * this._cacheClientNotifier.addFilterRegisteredClients(regionName, this.proxyID); } catch
-   * (RegionDestroyedException e) {
-   * logger.warn(LocalizedStrings.CacheClientProxy_0_INTEREST_REG_FOR_0_FAILED, regionName + "->" +
-   * keyOfInterest, e); } }
-   */
-
   /**
    * Registers interest in the input region name and key
    *
@@ -1526,8 +1518,6 @@ public class CacheClientProxy implements ClientSession {
    *
    */
   protected void deliverMessage(Conflatable conflatable) {
-    ThreadState state = securityService.bindSubject(subject);
-
     final ClientUpdateMessage clientMessage;
     if (conflatable instanceof HAEventWrapper) {
       clientMessage = ((HAEventWrapper) conflatable).getClientUpdateMessage();
@@ -1537,12 +1527,21 @@ public class CacheClientProxy implements ClientSession {
 
     _statistics.incMessagesReceived();
 
-    // post process
-    if (securityService.needPostProcess()) {
-      Object oldValue = clientMessage.getValue();
-      Object newValue = securityService.postProcess(clientMessage.getRegionName(),
-          clientMessage.getKeyOfInterest(), oldValue, clientMessage.valueIsObject());
-      clientMessage.setLatestValue(newValue);
+    // post process for single-user mode. We don't do post process for multi-user mode
+    if (subject != null) {
+      ThreadState state = securityService.bindSubject(subject);
+      try {
+        if (securityService.needPostProcess()) {
+          Object oldValue = clientMessage.getValue();
+          Object newValue = securityService.postProcess(clientMessage.getRegionName(),
+              clientMessage.getKeyOfInterest(), oldValue, clientMessage.valueIsObject());
+          clientMessage.setLatestValue(newValue);
+        }
+      } finally {
+        if (state != null) {
+          state.clear();
+        }
+      }
     }
 
     if (clientMessage.needsNoAuthorizationCheck() || postDeliverAuthCheckPassed(clientMessage)) {
@@ -1567,10 +1566,6 @@ public class CacheClientProxy implements ClientSession {
       }
 
       _statistics.incMessagesFailedQueued();
-    }
-
-    if (state != null) {
-      state.clear();
     }
   }
 
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/CacheClientUpdater.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/CacheClientUpdater.java
@@ -47,6 +47,7 @@ import org.apache.geode.cache.Operation;
 import org.apache.geode.cache.RegionDestroyedException;
 import org.apache.geode.cache.client.ServerRefusedConnectionException;
 import org.apache.geode.cache.client.SocketFactory;
+import org.apache.geode.cache.client.internal.AuthenticateUserOp;
 import org.apache.geode.cache.client.internal.ClientUpdater;
 import org.apache.geode.cache.client.internal.Endpoint;
 import org.apache.geode.cache.client.internal.EndpointManager;
@@ -1655,6 +1656,9 @@ public class CacheClientUpdater extends LoggingThread implements ClientUpdater, 
             case MessageType.LOCAL_DESTROY:
               handleDestroy(clientMessage);
               break;
+            case MessageType.CLIENT_RE_AUTHENTICATE:
+              handleAuthenticate();
+              break;
             case MessageType.LOCAL_DESTROY_REGION:
               handleDestroyRegion(clientMessage);
               break;
@@ -1761,6 +1765,18 @@ public class CacheClientUpdater extends LoggingThread implements ClientUpdater, 
       // 2. if there is some other race condition with continueProcessing flag
       this.qManager.checkEndpoint(this, this.endpoint);
     }
+  }
+
+  private void handleAuthenticate() {
+    // if client is in multi-user mode, the CacheClientUpdater (at this point)
+    // can't differentiate which user this message is intended to. so throw exception for now
+    // one possible solution is re-authenticate all users in this client
+    if (qManager.getPool().getMultiuserAuthentication()) {
+      throw new UnsupportedOperationException(
+          "Multi-user mode doesn't support re-authentication. This client will be closed.");
+    }
+    Long userId = AuthenticateUserOp.executeOn(location, qManager.getPool());
+    location.setUserId(userId);
   }
 
   /**
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientReAuthenticateMessage.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientReAuthenticateMessage.java
@@ -1,0 +1,106 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.apache.geode.internal.cache.tier.sockets;
+
+import java.io.DataInput;
+import java.io.DataOutput;
+import java.io.IOException;
+
+import org.apache.geode.annotations.Immutable;
+import org.apache.geode.internal.cache.EventID;
+import org.apache.geode.internal.cache.tier.MessageType;
+import org.apache.geode.internal.serialization.DeserializationContext;
+import org.apache.geode.internal.serialization.KnownVersion;
+import org.apache.geode.internal.serialization.SerializationContext;
+
+public class ClientReAuthenticateMessage implements ClientMessage {
+  @Immutable
+  public static final KnownVersion RE_AUTHENTICATION_START_VERSION = KnownVersion.GEODE_1_15_0;
+  /**
+   * This {@code ClientMessage}'s {@code EventID}
+   */
+  private final EventID eventId;
+
+  public ClientReAuthenticateMessage() {
+    this(new EventID());
+  }
+
+  public ClientReAuthenticateMessage(EventID eventId) {
+    this.eventId = eventId;
+  }
+
+  @Override
+  public Message getMessage(CacheClientProxy proxy, boolean notify) throws IOException {
+    Message message = new Message(1, KnownVersion.CURRENT);
+    message.setMessageType(MessageType.CLIENT_RE_AUTHENTICATE);
+    message.setTransactionId(0);
+    message.addObjPart(eventId);
+    return message;
+  }
+
+  @Override
+  public boolean shouldBeConflated() {
+    return true;
+  }
+
+  @Override
+  public void toData(DataOutput out,
+      SerializationContext context) throws IOException {
+    eventId.toData(out, context);
+  }
+
+  @Override
+  public int getDSFID() {
+    return CLIENT_RE_AUTHENTICATE;
+  }
+
+  @Override
+  public void fromData(DataInput in,
+      DeserializationContext context) throws IOException, ClassNotFoundException {
+    eventId.fromData(in, context);
+  }
+
+  @Override
+  public EventID getEventId() {
+    return eventId;
+  }
+
+  @Override
+  public String getRegionToConflate() {
+    return "gemfire_reserved_region_name_for_client_re_authenticate";
+  }
+
+  @Override
+  public Object getKeyToConflate() {
+    // This method can be called by HARegionQueue.
+    // Use this to identify the message type.
+    return "re_authenticate";
+  }
+
+  @Override
+  public Object getValueToConflate() {
+    // This method can be called by HARegionQueue
+    // Use this to identify the message type.
+    return "re_authenticate";
+  }
+
+  @Override
+  public void setLatestValue(Object value) {}
+
+  @Override
+  public KnownVersion[] getSerializationVersions() {
+    return null;
+  }
+}
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientUserAuths.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientUserAuths.java
@@ -14,6 +14,8 @@
  */
 package org.apache.geode.internal.cache.tier.sockets;
 
+import static org.apache.geode.cache.client.internal.AuthenticateUserOp.NOT_A_USER_ID;
+
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Map;
@@ -37,6 +39,7 @@ public class ClientUserAuths {
   private final ConcurrentMap<String, UserAuthAttributes> cqNameVsUserAuth =
       new ConcurrentHashMap<>();
   private final ConcurrentMap<Long, Subject> uniqueIdVsSubject = new ConcurrentHashMap<>();
+  private final ConcurrentMap<String, Long> cqNameVsUniqueId = new ConcurrentHashMap<>();
 
   private final int m_seed;
 
@@ -49,8 +52,16 @@ public class ClientUserAuths {
     return newId;
   }
 
-  public Long putSubject(Subject subject) {
-    final Long newId = getNextID();
+  public Long putSubject(Subject subject, long existingUniqueId) {
+    final Long newId;
+    if (existingUniqueId == 0 || existingUniqueId == NOT_A_USER_ID) {
+      newId = getNextID();
+    } else {
+      // the user re-authenticate back, remove the existing subject, reuse the uniqueId
+      removeSubject(existingUniqueId);
+      newId = existingUniqueId;
+    }
+
     uniqueIdVsSubject.put(newId, subject);
     logger.debug("Subject of {} added.", newId);
     return newId;
@@ -62,7 +73,7 @@ public class ClientUserAuths {
     m_firstId = uniqueIdGenerator.nextLong();
   }
 
-  private synchronized long getNextID() {
+  synchronized long getNextID() {
     final long uniqueId = uniqueIdGenerator.nextLong();
     if (uniqueId == m_firstId) {
       uniqueIdGenerator = new Random(m_seed + System.currentTimeMillis());
@@ -95,12 +106,24 @@ public class ClientUserAuths {
       return false;
     }
 
+    if (subject.getPrincipal() == null) {
+      return false;
+    }
+
     subject.logout();
     return true;
   }
 
   public UserAuthAttributes getUserAuthAttributes(final String cqName) {
     return cqNameVsUserAuth.get(cqName);
+  }
+
+  public Subject getSubject(final String cqName) {
+    Long uniqueId = cqNameVsUniqueId.get(cqName);
+    if (uniqueId != null) {
+      return uniqueIdVsSubject.get(uniqueId);
+    }
+    return null;
   }
 
   public void setUserAuthAttributesForCq(final String cqName, final Long uniqueId,
@@ -122,6 +145,7 @@ public class ClientUserAuths {
         }
       }
     }
+    cqNameVsUniqueId.put(cqName, uniqueId);
   }
 
   public void removeUserAuthAttributesForCq(final String cqName, final boolean isDurable) {
@@ -129,6 +153,7 @@ public class ClientUserAuths {
     if (uaa != null && isDurable) {
       uaa.unsetDurable();
     }
+    cqNameVsUniqueId.remove(cqName);
   }
 
   public void removeUserId(final Long userId, final boolean keepAlive) {
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/MessageDispatcher.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/MessageDispatcher.java
@@ -14,6 +14,9 @@
  */
 package org.apache.geode.internal.cache.tier.sockets;
 
+import static org.apache.geode.internal.cache.EntryEventImpl.deserialize;
+import static org.apache.geode.internal.cache.tier.sockets.ClientReAuthenticateMessage.RE_AUTHENTICATION_START_VERSION;
+import static org.apache.geode.internal.lang.SystemPropertyHelper.RE_AUTHENTICATE_WAIT_TIME;
 import static org.apache.geode.util.internal.UncheckedUtils.uncheckedCast;
 
 import java.io.IOException;
@@ -28,6 +31,8 @@ import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 
 import org.apache.logging.log4j.Logger;
+import org.apache.shiro.subject.Subject;
+import org.jetbrains.annotations.NotNull;
 
 import org.apache.geode.CancelException;
 import org.apache.geode.DataSerializer;
@@ -40,16 +45,22 @@ import org.apache.geode.distributed.internal.InternalDistributedSystem;
 import org.apache.geode.internal.cache.ClientServerObserver;
 import org.apache.geode.internal.cache.ClientServerObserverHolder;
 import org.apache.geode.internal.cache.Conflatable;
+import org.apache.geode.internal.cache.EventID;
 import org.apache.geode.internal.cache.InternalCache;
 import org.apache.geode.internal.cache.ha.HAContainerWrapper;
 import org.apache.geode.internal.cache.ha.HARegionQueue;
 import org.apache.geode.internal.cache.ha.HARegionQueueAttributes;
 import org.apache.geode.internal.cache.ha.HARegionQueueStats;
+import org.apache.geode.internal.lang.SystemPropertyHelper;
 import org.apache.geode.internal.logging.log4j.LogMarker;
+import org.apache.geode.internal.security.SecurityService;
 import org.apache.geode.internal.serialization.ByteArrayDataInput;
 import org.apache.geode.internal.statistics.StatisticsClock;
 import org.apache.geode.logging.internal.executors.LoggingThread;
 import org.apache.geode.logging.internal.log4j.api.LogService;
+import org.apache.geode.security.AuthenticationExpiredException;
+import org.apache.geode.security.NotAuthorizedException;
+import org.apache.geode.security.ResourcePermission;
 
 /**
  * Class <code>MessageDispatcher</code> is a <code>Thread</code> that processes messages bound for
@@ -68,6 +79,12 @@ public class MessageDispatcher extends LoggingThread {
    * Key in the system property from which the slow starting time value will be retrieved
    */
   private static final String KEY_SLOW_START_TIME_FOR_TESTING = "slowStartTimeForTesting";
+
+
+  /**
+   * Default value in milliseconds for waiting for re-authentication
+   */
+  private static final long DEFAULT_RE_AUTHENTICATE_WAIT_TIME = 5000;
 
   /**
    * The queue of messages to be sent to the client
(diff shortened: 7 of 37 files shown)


geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/MessageDispatcher.java
```diff
@@ -362,31 +385,63 @@ protected void runDispatcher() {
           }
           waitForResumption();
         }
-        try {
-          clientMessage = (ClientMessage) _messageQueue.peek();
-        } catch (RegionDestroyedException skipped) {
-          break;
+
+        if (clientMessage == null) {
```
agingade, line 389: Can we add comment saying client-message from previous peek is not yet dispatched and waiting for re-auth; to make it clear why it may not be null...

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/MessageDispatcher.java
```diff
@@ -362,31 +385,63 @@ protected void runDispatcher() {
...
-          // Process the message
-          long start = getStatistics().startTime();
-          //// BUGFIX for BUG#38206 and BUG#37791
-          boolean isDispatched = dispatchMessage(clientMessage);
-          getStatistics().endMessage(start);
-          if (isDispatched) {
+
+        if (clientMessage == null) {
+          _messageQueue.remove();
+          break;
```
agingade, line 404: The code differs from what it used to be previously; it used to be only removed. 
Can this check be moved to right after the peek is done? and continue the loop if its null...

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/MessageDispatcher.java
```diff
@@ -362,31 +385,63 @@ protected void runDispatcher() {
...
+            getStatistics().endMessage(start);
             _messageQueue.remove();
             if (clientMessage instanceof ClientMarkerMessageImpl) {
               getProxy().setMarkerEnqueued(false);
             }
           }
-        } else {
+          clientMessage = null;
+          wait_for_re_auth_start_time = -1;
+        } catch (NotAuthorizedException notAuthorized) {
```
agingade, line 419: The auth exception handling logic be moved to separate block. E.g. handleAuthExceptions(){}

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/MessageDispatcher.java
```diff
@@ -362,31 +385,63 @@ protected void runDispatcher() {
...
+          logger.info("skip delivering message: " + clientMessage, notAuthorized);
           _messageQueue.remove();
+          clientMessage = null;
+        } catch (AuthenticationExpiredException expired) {
+          if (wait_for_re_auth_start_time == -1) {
+            wait_for_re_auth_start_time = System.currentTimeMillis();
+            // only send the message to clients who can handle the message
+            if (getProxy().getVersion().isNewerThanOrEqualTo(RE_AUTHENTICATION_START_VERSION)) {
+              EventID eventId = new EventID(getCache().getDistributedSystem());
+              sendMessageDirectly(new ClientReAuthenticateMessage(eventId));
```
agingade, line 430: There could be IOException during send that needs to be handled as done in IOException catch block.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/MessageDispatcher.java
```diff
@@ -362,31 +385,63 @@ protected void runDispatcher() {
...
+            }
+            // for older client, we still wait, just in case client will perform some operations to
+            // trigger credential refresh on its own.
+            Thread.sleep(500);
```
agingade, line 434: 500 may be a large number?
Is there way to do wait/notify...timely waiting in re-auth?

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/MessageDispatcher.java
```diff
@@ -362,31 +385,63 @@ protected void runDispatcher() {
...
+          } else {
+            long elapsedTime = System.currentTimeMillis() - wait_for_re_auth_start_time;
+            if (elapsedTime > reAuthenticateWaitTime) {
+              pauseOrUnregisterProxy(expired);
```
agingade, line 438: adding warn/debug level log here may help here...

agingade (commented):
This is changing the messaging between client/server; it will be nice to have client backward compatibility related unit tests.

kirklund (changes requested):
Thanks for writing AuthenticateUserOpTest!

We should really unit test the PR changes in the following classes. No need to test the entire class, just the spots that are changed in the PR. I would be happy to pair or help in any way with unit testing these. I think I even created some initial unit tests for a couple of these:
- CacheClientProxy
- ClientUserAuths
- MessageDispatcher
- PutUserCredentials
- ServerConnection

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/CacheClientProxy.java
```diff
@@ -371,8 +370,11 @@ public void setPostAuthzCallback(AccessControl authzCallback) {
     }
   }
 
+  public boolean isMultiUserMode() {
+    return subject == null;
```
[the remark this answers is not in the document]
jinmeiliao: got rid of this method.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/MessageDispatcher.java
```diff
@@ -362,31 +385,63 @@ protected void runDispatcher() {
...
             if (clientMessage instanceof ClientMarkerMessageImpl) {
               getProxy().setMarkerEnqueued(false);
             }
           }
-        } else {
+          clientMessage = null;
+          wait_for_re_auth_start_time = -1;
+        } catch (NotAuthorizedException notAuthorized) {
+          // behave as if the message is dispatched, remove from the queue
+          logger.info("skip delivering message: " + clientMessage, notAuthorized);
```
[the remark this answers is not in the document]
jinmeiliao: this exception should not happen in the first place, if it does, an info level message would make it clear what happened since this exception is not bubbled up to the caller.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/MessageDispatcher.java
jinmeiliao: done

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/MessageDispatcher.java
jinmeiliao: done

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/MessageDispatcher.java
jinmeiliao: the method itself handled it already

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/MessageDispatcher.java
jinmeiliao: change it to 200ms.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/MessageDispatcher.java
jinmeiliao: done.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/CacheClientProxy.java
```diff
@@ -1537,12 +1531,18 @@ protected void deliverMessage(Conflatable conflatable) {
...
+      ThreadState state = securityService.bindSubject(subject);
+      if (securityService.needPostProcess()) {
+        Object oldValue = clientMessage.getValue();
+        Object newValue = securityService.postProcess(clientMessage.getRegionName(),
+            clientMessage.getKeyOfInterest(), oldValue, clientMessage.valueIsObject());
+        clientMessage.setLatestValue(newValue);
+      }
+      if (state != null) {
+        state.clear();
+      }
```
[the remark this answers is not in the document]
jinmeiliao: done

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/MessageDispatcher.java
jinmeiliao: unable to do this, multiple exit points

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/CacheClientUpdater.java
```diff
@@ -1763,6 +1767,19 @@ private void processMessages() {
     }
   }
 
+  private void handleAuthenticate(Message clientMessage) {
+    // if client is in multi-user mode, the CacheClientUpdator (at this point)
+    // can't differentiate which user this message is intended to. so throw exception for now
+    // one possible solution is re-authenticate all users in this client
+    if (qManager.getPool().getMultiuserAuthentication()) {
+      throw new UnsupportedOperationException(
```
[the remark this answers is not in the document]
jinmeiliao: The test shows the connection gets closed, may need to investigate what eventually triggered the close event.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientUserAuths.java
```diff
@@ -49,10 +50,18 @@ public Long putUserAuth(UserAuthAttributes userAuthAttr) {
...
+      newId = getNextID();
+    } else {
+      // the user re-authenticate back, remove the existing subject, reuse the uniqueId
+      removeSubject(existingUniqueId);
+      newId = existingUniqueId;
+    }
+
     uniqueIdVsSubject.put(newId, subject);
-    logger.debug("Subject of {} added.", newId);
+    logger.info("Subject of {} added.", newId);
```
mhansonp, line 64: Do we want this to always log? It was debug before.

mhansonp (approved):

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientUserAuths.java
jinmeiliao: Oh, no, accidental change.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/MessageDispatcher.java
jinmeiliao: Tried this out. The method needs to return a wait_start_time to be used for later iterations. It doesn't make the logic easier to read. I think it's much clear to handle it in place. Also we don't have problem to unit test this part of the logic.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientReAuthenticateMessage.java
```diff
@@ -0,0 +1,103 @@
...
+import org.apache.geode.internal.serialization.DeserializationContext;
+import org.apache.geode.internal.serialization.KnownVersion;
+import org.apache.geode.internal.serialization.SerializationContext;
+
+public class ClientReAuthenticateMessage implements ClientMessage {
+
+  /**
+   * This <code>ClientMessage</code>'s <code>EventID</code>
+   */
+  private EventID eventId;
```
dschneider-pivotal, line 33: A more efficient way of having this message serialize "eventId" also allows you to make this field final. You would make the following changes:
1. make the field "final".
2. add to the default constructor  "eventId = new EventID();"
3. change toData to be "eventId.toData(out, context);"
4. change fromData to be "eventId.fromData(in, context);"

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/MessageDispatcher.java
```diff
@@ -69,6 +78,17 @@
    */
   private static final String KEY_SLOW_START_TIME_FOR_TESTING = "slowStartTimeForTesting";
 
+
+  /**
+   * Default value for waitinig for re-authentication
```
dschneider-pivotal, line 83: typo "waitining". Also describe the units (milliseconds)

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/MessageDispatcher.java
```diff
@@ -69,6 +78,17 @@
...
+   */
+  private static final long DEFAULT_RE_AUTHENTICATE_WAIT_TIME = 5000;
+
+  /**
+   * Key in the system property from which the slow starting time value will be retrieved
+   */
+  protected static final String RE_AUTHENTICATE_WAIT_TIME_KEY = "geode.reauthenticate.wait.time";
```
dschneider-pivotal, line 90: Consider using SystemPropertyHelper for you new system properties. It has support for both prefixes (geode. or gemfire.).

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/MessageDispatcher.java
```diff
@@ -362,31 +390,67 @@ protected void runDispatcher() {
           }
           waitForResumption();
         }
-        try {
-          clientMessage = (ClientMessage) _messageQueue.peek();
-        } catch (RegionDestroyedException skipped) {
-          break;
+
+        // if message is not delivered due to authentication expiation, this clientMessage
```
dschneider-pivotal, line 394: typo "expiation" should be "expiration"?

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/MessageDispatcher.java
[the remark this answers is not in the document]
dschneider-pivotal: But if it does happen (i.e. the info log message) do we want the server to keep logging it for every message it attempts to send to that unauthenticated client? Will the server close this dispatcher in this case or retry reauth? If not it seems like we could log once (maybe it should be a warning) for that client.

dschneider-pivotal (approved):

geode-core/src/main/java/org/apache/geode/cache/client/internal/AuthenticateUserOp.java
```diff
@@ -123,39 +119,61 @@ private AuthenticateUserOp() {
...
-        secureBytes = ((ConnectionImpl) cnx).encryptBytes(hdos.toByteArray());
-      } finally {
-        hdos.close();
+      byte[] credentialBytes = getCredentialBytes(cnx, securityProperties);
+      getMessage().addBytesPart(credentialBytes);
+
+      long userId = getUserId(cnx);
+      try (HeapDataOutputStream hdos = new HeapDataOutputStream(KnownVersion.CURRENT)) {
+        hdos.writeLong(cnx.getConnectionID());
+        hdos.writeLong(userId);
```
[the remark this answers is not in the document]
jinmeiliao: having this `getUserId(cnx)` method makes it easier to unit test. I don't understand your suggestion of "move the declaration into...", move what to where?

geode-core/src/main/java/org/apache/geode/cache/client/internal/AuthenticateUserOp.java
```diff
@@ -123,39 +119,61 @@ private AuthenticateUserOp() {
...
       }
-      try {
-        secureBytes = ((ConnectionImpl) cnx).encryptBytes(hdos.toByteArray());
-      } finally {
-        hdos.close();
+      byte[] credentialBytes = getCredentialBytes(cnx, securityProperties);
+      getMessage().addBytesPart(credentialBytes);
+
+      long userId = getUserId(cnx);
+      try (HeapDataOutputStream hdos = new HeapDataOutputStream(KnownVersion.CURRENT)) {
```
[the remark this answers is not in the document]
jinmeiliao: Good call.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientReAuthenticateMessage.java
jinmeiliao: done, thanks for the suggestion

geode-core/src/distributedTest/java/org/apache/geode/management/internal/security/MultiUserAuthenticationDUnitTest.java
```diff
@@ -16,9 +16,14 @@
 package org.apache.geode.management.internal.security;
 
 import static org.apache.geode.cache.Region.SEPARATOR;
+import static org.apache.geode.test.awaitility.GeodeAwaitility.await;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
```
kirklund, line 21: I think you meant to use:
```
```

geode-core/src/distributedTest/java/org/apache/geode/management/internal/security/MultiUserAuthenticationDUnitTest.java
```diff
@@ -100,4 +113,67 @@ public void multiAuthenticatedView() throws Exception {
...
+        "select * from /region r where r.length<=2");
+    EventsCqListner listener2 = createAndExecuteCQ(regionService2.getQueryService(), "cq2",
+        "select * from /region r where r.length>=2");
+
+    // put 4 data in the region
+    gfsh.executeAndAssertThat("put --region=region --key=1 --value=1");
+    gfsh.executeAndAssertThat("put --region=region --key=11 --value=11");
+    gfsh.executeAndAssertThat("put --region=region --key=111 --value=111");
+
+    await().atMost(100, TimeUnit.SECONDS).untilAsserted(
```
kirklund, line 139: These `await()` calls should just use the default timeout. If this test runs on a slow or overloaded machine, it may fail here.

geode-core/src/main/java/org/apache/geode/cache/client/internal/AuthenticateUserOp.java
```diff
@@ -123,39 +119,61 @@ private AuthenticateUserOp() {
...
+      try (HeapDataOutputStream hdos = new HeapDataOutputStream(16, KnownVersion.CURRENT)) {
+        hdos.writeLong(cnx.getConnectionID());
+        hdos.writeLong(userId);
+        getMessage().setSecurePart(((ConnectionImpl) cnx).encryptBytes(hdos.toByteArray()));
       }
-      getMessage().setSecurePart(secureBytes);
       getMessage().send(false);
     }
 
+    protected long getUserId(Connection cnx) {
```
kirklund, line 142: Can we name the parameter `connection` instead of `cnx`?

geode-core/src/main/java/org/apache/geode/cache/client/internal/AuthenticateUserOp.java
```diff
@@ -123,39 +119,61 @@ private AuthenticateUserOp() {
...
+        return -1L;
+      }
+      return id;
+    }
+
+    protected InternalDistributedSystem getConnectedSystem() {
+      return InternalDistributedSystem.getConnectedInstance();
+    }
+
+    protected byte[] getCredentialBytes(Connection cnx, Properties securityProperties)
```
kirklund, line 159: `connection` instead of `cnx`? I know the old uses abbreviations, but we really want to escape that trend.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/CacheClientProxy.java
```diff
@@ -345,19 +348,15 @@ protected CacheClientProxy(InternalCache cache, CacheClientNotifier ccn, Socket
     initializeClientAuths();
   }
 
-  private void initializeClientAuths() {
-    if (AcceptorImpl.isPostAuthzCallbackPresent()) {
-      clientUserAuths = ServerConnection.getClientUserAuths(proxyID);
-    }
+  void initializeClientAuths() {
+    clientUserAuths = ServerConnection.getClientUserAuths(proxyID);
```
kirklund, line 352: One way to make the use of `CacheClientProxy` more testable is to wrap the`ServerConnection.getClientUserAuths` calls in a ` java.util.function.Function` or custom inner-interface.

You basically add a new field:
```
private final Function<ClientProxyMembershipID, ClientUserAuths> getClientUserAuthsFunction;
```
Then update the 2nd `CacheClientProxy` constructor marked with `@VisibleForTesting` to include one more parameter:
```
MessageDispatcherFactory messageDispatcherFactory,
Function<ClientProxyMembershipID, ClientUserAuths> getClientUserAuthsFunction) {
```
Change the 1st constructor to pass in a method reference:
```
DEFAULT_MESSAGEDISPATCHERFACTORY,
ServerConnection::getClientUserAuths);
```
And save the the parameter in the 2nd constructor to the field:
```
this.getClientUserAuthsFunction = getClientUserAuthsFunction;
```
And finally use it anywhere the class calls the static `ServerConnection.getClientUserAuths`:
```
clientUserAuths = getClientUserAuths.apply(proxyID);
```

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientReAuthenticateMessage.java
```diff
@@ -0,0 +1,104 @@
...
+import org.apache.geode.internal.cache.EventID;
+import org.apache.geode.internal.cache.tier.MessageType;
+import org.apache.geode.internal.serialization.DeserializationContext;
+import org.apache.geode.internal.serialization.KnownVersion;
+import org.apache.geode.internal.serialization.SerializationContext;
+
+public class ClientReAuthenticateMessage implements ClientMessage {
+
+  /**
+   * This <code>ClientMessage</code>'s <code>EventID</code>
```
kirklund, line 30: Please use the newer syntax `{@code xxx}` instead of `<code>`:
```
* This {@code ClientMessage}'s {@code EventID}
```

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientReAuthenticateMessage.java
```diff
@@ -0,0 +1,104 @@
...
+   */
+  private final EventID eventId;
+
+  public ClientReAuthenticateMessage() {
+    eventId = new EventID();
+  }
+
+  public ClientReAuthenticateMessage(EventID eventId) {
+    this.eventId = eventId;
+  }
```
kirklund, line 40: You should really use constructor chaining here:
```
public ClientReAuthenticateMessage() {
  this(new EventID());
}
```

kirklund (changes requested):

geode-core/src/distributedTest/java/org/apache/geode/management/internal/security/MultiUserAuthenticationDUnitTest.java
jinmeiliao: yeah, thanks!. I was wondering why some of the methods were not available.

geode-core/src/distributedTest/java/org/apache/geode/management/internal/security/MultiUserAuthenticationDUnitTest.java
jinmeiliao: done

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/CacheClientProxy.java
jinmeiliao: That's a useful trick, I will see if I need more tests that would utilize this.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientReAuthenticateMessage.java
jinmeiliao: done

geode-core/src/main/java/org/apache/geode/cache/client/internal/AuthenticateUserOp.java
jinmeiliao: done

geode-core/src/main/java/org/apache/geode/cache/client/internal/AuthenticateUserOp.java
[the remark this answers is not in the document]
dschneider-pivotal: You declare "long userId" right before the try block that uses it. You could narrow the scope of this variable by declaring and initializing it inside the try block. I was not suggesting you get rid of the getUserId method
but you could in this code just have "hdos.writeLong(getUserId(connection));"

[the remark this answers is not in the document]
jinmeiliao: ah, I see. thanks!

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/CacheClientProxy.java
jinmeiliao: If I just add a constructor that takes a ClientUserAuth object as the extra parameter instead of the supplier function, would there be any difference as far as testability goes?

kirklund: If `getClientUserAuths` was non-static then you could pass an instance of it into the constructor and that would be even better. Since it's static you need to wrap it within a `Function` interface.

kirklund: PS: this was just sharing a pattern that might help with some of the testing... use it in later PRs instead of this one if you don't need it here

kirklund (approved):
Looks great!

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/CacheClientProxy.java
jinmeiliao: Great! Thanks!

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/CacheClientUpdater.java
```diff
@@ -1763,6 +1767,19 @@ private void processMessages() {
...
+          "Multi-user mode doesn't support re-authentication. This client will be closed.");
+    }
+    Long userId = AuthenticateUserOp
+        .executeOn(location, qManager.getPool(), this.system.getSecurityProperties());
```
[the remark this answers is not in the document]
jinmeiliao: I will continue to write more dunit tests in the coming PR.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/MessageDispatcher.java
```diff
@@ -362,31 +387,67 @@ protected void runDispatcher() {
...
+        } catch (AuthenticationExpiredException expired) {
+          if (wait_for_re_auth_start_time == -1) {
+            wait_for_re_auth_start_time = System.currentTimeMillis();
+            // only send the message to clients who can handle the message
+            if (getProxy().getVersion().isNewerThanOrEqualTo(RE_AUTHENTICATION_START_VERSION)) {
+              EventID eventId = createEventId();
+              sendMessageDirectly(new ClientReAuthenticateMessage(eventId));
+            }
+            // for older client, we still wait, just in case client will perform some operations to
+            // trigger credential refresh on its own.
```
Bill, line 437: The comment makes me think that this `Thread.sleep()` should be in an `else` clause so the sleep if clients don't understand the re-authenticate message.

Consider putting this sleep into an `else` block.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/MessageDispatcher.java
```diff
@@ -335,10 +356,14 @@ protected void runDispatcher() {
       logger.debug("{}: Beginning to process events", this);
     }
 
+    long reAuthenticateWaitTime =
+        getSystemProperty(RE_AUTHENTICATE_WAIT_TIME, DEFAULT_RE_AUTHENTICATE_WAIT_TIME);
+
     ClientMessage clientMessage = null;
+    long wait_for_re_auth_start_time = -1;
```
Bill, line 363: Please consider moving this variable declaration closer to where it's needed. It can moved inside the `while` body: down to just before the `try` block at line 413.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/MessageDispatcher.java
```diff
@@ -362,31 +387,67 @@ protected void runDispatcher() {
...
+            // trigger credential refresh on its own.
+            Thread.sleep(200);
+          } else {
+            long elapsedTime = System.currentTimeMillis() - wait_for_re_auth_start_time;
+            if (elapsedTime > reAuthenticateWaitTime) {
+              logger.warn("Client did not re-authenticate back successfully in " + elapsedTime
+                  + "ms. Unregister this client proxy.");
+              pauseOrUnregisterProxy(expired);
+            } else {
+              Thread.sleep(200);
```
Bill, line 446: I don't understand why the `sleep()` is needed here. Please explain.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/MessageDispatcher.java
```diff
@@ -581,6 +656,33 @@ protected boolean dispatchMessage(ClientMessage clientMessage) throws IOExceptio
         logger.trace(msg.toString());
       }
 
+      // authorize the message before dispatching
+      String regionName = clientUpdateMessage.getRegionName();
+      Object key = clientUpdateMessage.getKeyOfInterest();
+      SecurityService securityService = getCache().getSecurityService();
+      ResourcePermission permission = new ResourcePermission(ResourcePermission.Resource.DATA,
+          ResourcePermission.Operation.READ,
+          regionName, key == null ? null : key.toString());
```
Bill, line 665: `permission` isn't needed outside the body of the next `if`. And even then it is only needed conditionally (sometimes it isn't needed at all). And `regionName` and `key` are used only to compute `permission`.

Please consider replacing lines 660, 661, 663-665 w/ a method that takes `clientUpdateMessage` and returns `ResourcePermission` and then call that method in the two places `permission` is currently used below. This will eliminate some spurious object allocation and make this code a little simpler.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ServerConnection.java
```diff
@@ -1164,6 +1190,20 @@ public void removeUserAuth(Message message, boolean keepAlive) {
     }
   }
 
+  @VisibleForTesting
+  long putSubject(Subject subject, long existingUniqueId) {
+    long uniqueId;
+    uniqueId = clientUserAuths.putSubject(subject, existingUniqueId);
```
Bill, line 1196: Please combine the assignment on this line with the declaration on the previous one. 

Incidentally, `uniqueId` can be `final`.

Bill (changes requested):
Great PR @jinmeiliao. Thanks for the added unit testing! (I verified that lots of the new/changed logic is covered by new tests).

I found one problem (an unused variable). There are a few suggestions which I hope you'll agree with. And there is a question or two where I'd like some clarification so I can understand better.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientReAuthenticateMessage.java
```diff
@@ -0,0 +1,104 @@
...
+
+  @Override
+  public Object getValueToConflate() {
+    // This method can be called by HARegionQueue
+    // Use this to identify the message type.
+    return "re_authenticate";
+  }
+
+  @Override
+  public void setLatestValue(Object value) {}
```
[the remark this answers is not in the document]
jinmeiliao: I pretty much copied the implementation of this message from `ClientPingMessageImpl`. I guess it doesn't mean it should fail, it just won't do anything.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/MessageDispatcher.java
jinmeiliao: updated the comment to be more clear of the intention here.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/MessageDispatcher.java
jinmeiliao: It has to be outside of the `while` loop. The same message will be processed by multiple loops if authentication expired and user hasn't re-authenticated back.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/MessageDispatcher.java
jinmeiliao: the `sleep()` is basically an interval between two authorization checks.  We don't want the check to be too busy. So the logic is like this
`
while(I am not stopped) {
   if (last message is not delivered), keep process the same message
   else take the message out of the queue
  try{
      dispatch the message (which includes authorization)
      message dispatched
  } catch (AuthExpirationException e) {
     if (this is the first time I am processing this message) {
          send the re-auth message to the client
          wait for 200ms before go back to the loop
     } else {
         if ( I waited  long enough) {
              stop the dispatcher and terminate the connection 
         } else {
             wait for 200ms and then go back to thee loop
         }
     }
  }
}
`

geode-core/src/main/java/org/apache/geode/cache/client/internal/AuthenticateUserOp.java
```diff
@@ -87,31 +101,13 @@ private AuthenticateUserOp() {
   }
 
   static class AuthenticateUserOpImpl extends AbstractOp {
-
+    private static final Logger logger = LogService.getLogger();
```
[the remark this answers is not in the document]
jinmeiliao: nice catch!

geode-core/src/main/java/org/apache/geode/cache/client/internal/AuthenticateUserOp.java
```diff
@@ -123,67 +119,89 @@ private AuthenticateUserOp() {
...
 
+    protected long getUserId(Connection connection) {
+      // single user mode
+      if (UserAttributes.userAttributes.get() == null) {
+        return connection.getServer().getUserId();
+      }
+      // multi user mode
+      Long id = UserAttributes.userAttributes.get().getServerToId().get(connection.getServer());
+      if (id == null) {
+        return -1L;
```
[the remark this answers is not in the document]
jinmeiliao: done

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/MessageDispatcher.java
jinmeiliao: done

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ServerConnection.java
jinmeiliao: done

boglesby:
I do see an issue when I run this test:

- 1 server
- 1 client registers interest in all keys
- 1 client does puts continuously

Cause the SecurityManager to throw an AuthenticationExpiredException after a puts have been occurring.

Add a sleep in the PutUserCredentials cmdExecute method so that the MessageDispatcher fails. I see an exception when the client attempts to reconnect that I'm not sure should be happening. Its not 100% reproducible, but it happens a lot.

I added some logging on the client and server to show the order of events.

The SecurityManager authorize method expires the credentials and throws AuthenticationExpiredException:
```
[warn 2021/09/14 16:53:17.224 PDT server1 <Client Message Dispatcher for boglesbymac(client-register:31501:loner):62522:364cbae6:client-register> tid=0x48] XXX TestExpirableSecurityManager.authorize isExpired returning isExpired=true; key=152104
```
The Message Dispatcher catches the AuthenticationExpiredException:
```
[warn 2021/09/14 16:53:17.224 PDT server1 <Client Message Dispatcher for boglesbymac(client-register:31501:loner):62522:364cbae6:client-register> tid=0x48] XXX MessageDispatcher.runDispatcher caught AuthenticationExpiredException processing key=152104
```
The CacheClientUpdater.handleAuthenticate on the client sends the AuthenticateUserOp:
```
[warn 2021/09/14 16:53:17.268 PDT client-register <Cache Client Updater Thread  on boglesbymac(server1:31479)<v81>:41001 port 62511> tid=0x1e] XXX CacheClientUpdater.handleAuthenticate about to re-authenticate
```
The server executes the PutUserCredentials.cmdExecute:
```
[warn 2021/09/14 16:53:17.270 PDT server1 <ServerConnection on port 62511 Thread 2> tid=0x43] XXX PutUserCredentials.cmdExecute existingUniqueId=-49945521082788877
```
I've added a sleep in PutUserCredentials.cmdExecute.

Then, I see pairs of these messages every 200ms for 5 seconds (added in the `catch (AuthenticationExpiredException expired)` block of MessageDispatcher runDispatcher):
```
[warn 2021/09/14 16:53:17.429 PDT server1 <Client Message Dispatcher for boglesbymac(client-register:31501:loner):62522:364cbae6:client-register> tid=0x48] XXX MessageDispatcher.runDispatcher caught AuthenticationExpiredException processing key=152104

[warn 2021/09/14 16:53:17.429 PDT server1 <Client Message Dispatcher for boglesbymac(client-register:31501:loner):62522:364cbae6:client-register> tid=0x48] XXX MessageDispatcher.runDispatcher waiting for reauthentication key=152104
```
After 5 seconds, the Message Dispatcher gives up and closes:
```
[warn 2021/09/14 16:53:22.289 PDT server1 <Client Message Dispatcher for boglesbymac(client-register:31501:loner):62522:364cbae6:client-register> tid=0x48] Client did not re-authenticate back successfully in 5065ms. Unregister this client proxy.

[warn 2021/09/14 16:53:22.291 PDT server1 <Client Message Dispatcher for boglesbymac(client-register:31501:loner):62522:364cbae6:client-register> tid=0x48] ClientHealthMonitor: Unregistering client with member id identity(127.0.0.1(client-register:31501:loner):62522:364cbae6:client-register,connection=2 due to: Unknown reason
```
The CacheClientUpdater closes on the client:
```
[info 2021/09/14 16:53:22.294 PDT client-register <Cache Client Updater Thread  on boglesbymac(server1:31479)<v81>:41001 port 62511> tid=0x1e] Primary subscription endpoint boglesbymac.local:62511,boglesbymac(server1:31479)<v81>:41001 crashed. Scheduling recovery.

[info 2021/09/14 16:53:22.296 PDT client-register <Cache Client Updater Thread  on boglesbymac(server1:31479)<v81>:41001 port 62511> tid=0x1e] Cache client updater for Queue on endpoint boglesbymac.local:62511,boglesbymac(server1:31479)<v81>:41001 exiting. Scheduling recovery.
```
The poolTimer sends the AuthenticateUserOp:
```
[warn 2021/09/14 16:53:22.321 PDT client-register <poolTimer-pool-3> tid=0x1b] XXX AuthenticateUserOp.executeOn about to send op
java.lang.Exception
	at org.apache.geode.cache.client.internal.AuthenticateUserOp.executeOn(AuthenticateUserOp.java:68)
	at org.apache.geode.cache.client.internal.ConnectionFactoryImpl.authenticateIfRequired(ConnectionFactoryImpl.java:152)
	at org.apache.geode.cache.client.internal.ConnectionFactoryImpl.createClientToServerConnection(ConnectionFactoryImpl.java:120)
	at org.apache.geode.cache.client.internal.ConnectionFactoryImpl.createClientToServerConnection(ConnectionFactoryImpl.java:223)
	at org.apache.geode.cache.client.internal.pooling.ConnectionManagerImpl.createPooledConnection(ConnectionManagerImpl.java:196)
	at org.apache.geode.cache.client.internal.pooling.ConnectionManagerImpl.createPooledConnection(ConnectionManagerImpl.java:190)
	at org.apache.geode.cache.client.internal.pooling.ConnectionManagerImpl.prefillConnection(ConnectionManagerImpl.java:608)
	at org.apache.geode.cache.client.internal.pooling.ConnectionManagerImpl.prefill(ConnectionManagerImpl.java:576)
	at org.apache.geode.cache.client.internal.pooling.ConnectionManagerImpl.access$200(ConnectionManagerImpl.java:70)
	at org.apache.geode.cache.client.internal.pooling.ConnectionManagerImpl$PrefillConnectionsTask.run2(ConnectionManagerImpl.java:684)
	at org.apache.geode.cache.client.internal.PoolImpl$PoolTask.run(PoolImpl.java:1336)

The queueTimer also sends the AuthenticateUserOp:

[warn 2021/09/14 16:53:22.322 PDT client-register <queueTimer-pool1> tid=0x1f] XXX AuthenticateUserOp.executeOn about to send op
java.lang.Exception
	at org.apache.geode.cache.client.internal.AuthenticateUserOp.executeOn(AuthenticateUserOp.java:68)
	at org.apache.geode.cache.client.internal.ConnectionFactoryImpl.authenticateIfRequired(ConnectionFactoryImpl.java:152)
	at org.apache.geode.cache.client.internal.ConnectionFactoryImpl.createClientToServerConnection(ConnectionFactoryImpl.java:120)
	at org.apache.geode.cache.client.internal.QueueManagerImpl.createNewPrimary(QueueManagerImpl.java:793)
	at org.apache.geode.cache.client.internal.QueueManagerImpl.recoverPrimary(QueueManagerImpl.java:911)
	at org.apache.geode.cache.client.internal.QueueManagerImpl.access$600(QueueManagerImpl.java:77)
	at org.apache.geode.cache.client.internal.QueueManagerImpl$RedundancySatisfierTask.run2(QueueManagerImpl.java:1448)
	at org.apache.geode.cache.client.internal.PoolImpl$PoolTask.run(PoolImpl.java:1336)
```
The server processes both of these AuthenticateUserOps and returns the unique ids:
```
[warn 2021/09/14 16:53:22.323 PDT server1 <ServerConnection on port 62512 Thread 3> tid=0x51] XXX ServerConnection.setCredentials credentials={security-username=boglesby, security-password=boglesby}

[warn 2021/09/14 16:53:22.323 PDT server1 <ServerConnection on port 62512 Thread 2> tid=0x50] XXX ServerConnection.setCredentials credentials={security-username=boglesby, security-password=boglesby}

[warn 2021/09/14 16:53:22.325 PDT server1 <ServerConnection on port 62512 Thread 2> tid=0x50] XXX ServerConnection.setCredentials uniqueId=858200770272636520

[warn 2021/09/14 16:53:22.325 PDT server1 <ServerConnection on port 62512 Thread 3> tid=0x51] XXX ServerConnection.setCredentials uniqueId=-8065196947535721026
```
The client receives these unique ids:
```
[warn 2021/09/14 16:53:22.325 PDT client-register <queueTimer-pool1> tid=0x1f] XXX AuthenticateUserOp.executeOn uniqueId=-8065196947535721026

[warn 2021/09/14 16:53:22.325 PDT client-register <poolTimer-pool-3> tid=0x1b] XXX AuthenticateUserOp.executeOn uniqueId=858200770272636520
```
Then the queueTimer creates a new CacheClientUpdater:
```
[warn 2021/09/14 16:53:22.325 PDT client-register <queueTimer-pool1> tid=0x1f] XXX CacheClientUpdater.<init>

[info 2021/09/14 16:53:22.346 PDT client-register <Cache Client Updater Thread  on boglesbymac(server1:31479)<v81>:41001 port 62512> tid=0x22] Cache Client Updater Thread  on boglesbymac(server1:31479)<v81>:41001 port 62512 (boglesbymac.local:62512) : ready to process messages.
```
Then, the queueTimer attempts to re-register interest, but fails due to `AuthenticationRequiredException: Failed to find the authenticated user` on the server:
```
[warn 2021/09/14 16:53:22.364 PDT client-register <queueTimer-pool1> tid=0x1f] QueueManagerImpl failed to recover interest to server boglesbymac.local:62512
org.apache.geode.cache.client.ServerOperationException: remote server on boglesbymac(client-register:31501:loner):62522:364cbae6:client-register: org.apache.geode.security.AuthenticationRequiredException: Failed to find the authenticated user.
	at org.apache.geode.cache.client.internal.OpExecutorImpl.handleException(OpExecutorImpl.java:559)
	at org.apache.geode.cache.client.internal.OpExecutorImpl.handleException(OpExecutorImpl.java:623)
	at org.apache.geode.cache.client.internal.OpExecutorImpl.handleException(OpExecutorImpl.java:512)
	at org.apache.geode.cache.client.internal.OpExecutorImpl.executeOn(OpExecutorImpl.java:475)
	at org.apache.geode.cache.client.internal.OpExecutorImpl.executeOn(OpExecutorImpl.java:486)
	at org.apache.geode.cache.client.internal.PoolImpl.executeOn(PoolImpl.java:857)
	at org.apache.geode.cache.client.internal.RegisterInterestOp.executeOn(RegisterInterestOp.java:103)
	at org.apache.geode.cache.client.internal.ServerRegionProxy.registerInterestOn(ServerRegionProxy.java:496)
	at org.apache.geode.cache.client.internal.QueueManagerImpl.recoverSingleKey(QueueManagerImpl.java:1190)
	at org.apache.geode.cache.client.internal.QueueManagerImpl.recoverSingleRegion(QueueManagerImpl.java:1158)
	at org.apache.geode.cache.client.internal.QueueManagerImpl.recoverSingleList(QueueManagerImpl.java:1103)
	at org.apache.geode.cache.client.internal.QueueManagerImpl.recoverInterestList(QueueManagerImpl.java:1223)
	at org.apache.geode.cache.client.internal.QueueManagerImpl.recoverAllInterestTypes(QueueManagerImpl.java:1237)
	at org.apache.geode.cache.client.internal.QueueManagerImpl.recoverInterest(QueueManagerImpl.java:1069)
	at org.apache.geode.cache.client.internal.QueueManagerImpl.recoverPrimary(QueueManagerImpl.java:928)
	at org.apache.geode.cache.client.internal.QueueManagerImpl.access$600(QueueManagerImpl.java:77)
	at org.apache.geode.cache.client.internal.QueueManagerImpl$RedundancySatisfierTask.run2(QueueManagerImpl.java:1448)
Caused by: org.apache.geode.security.AuthenticationRequiredException: Failed to find the authenticated user.
	at org.apache.geode.internal.security.IntegratedSecurityService.getSubject(IntegratedSecurityService.java:123)
	at org.apache.geode.internal.security.IntegratedSecurityService.authorize(IntegratedSecurityService.java:249)
	at org.apache.geode.internal.security.IntegratedSecurityService.authorize(IntegratedSecurityService.java:244)
	at org.apache.geode.internal.security.IntegratedSecurityService.authorize(IntegratedSecurityService.java:230)
	at org.apache.geode.internal.cache.tier.sockets.command.RegisterInterest61.cmdExecute(RegisterInterest61.java:174)
	at org.apache.geode.internal.cache.tier.sockets.BaseCommand.execute(BaseCommand.java:184)
```
The server disconnects the client:
```
[warn 2021/09/14 16:53:22.365 PDT server1 <ServerConnection on port 62512 Thread 3> tid=0x51] Server connection from [identity(127.0.0.1(client-register:31501:loner):62522:364cbae6:client-register,connection=2; port=62544]: connection disconnect detected by EOF.
```
There are several additional retries to create the CacheClientUpdater after this that fail.

geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationDUnitTest.java
```diff
@@ -56,7 +70,7 @@
   @Parameterized.Parameters(name = "{0}")
   public static Collection<String> data() {
     // only test the current version and the latest released version
```
DonalEvans, line 72: This comment is inaccurate, as once we're on version 1.16.0, this test will use versions 1.14.0, 1.15.0 and 1.16.0. A more accurate comment would be "only test versions greater than or equal to 1.14.0"

geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationDUnitTest.java
```diff
@@ -70,15 +84,70 @@
...
+  private static void doPutAndExpectFailure(Region<Object, Object> region, int times) {
+    for (int i = 1; i < times; i++) {
+      try {
+        region.put(1, "value1");
+        fail("Exception expected");
+      } catch (Exception e) {
+        assertThat(e).isInstanceOf(ServerOperationException.class);
+        assertThat(e.getCause()).isInstanceOfAny(AuthenticationFailedException.class,
+            AuthenticationRequiredException.class);
+      }
```
DonalEvans, line 138: This can be simplified to:
```
      assertThatThrownBy(() -> region.put(1, "value1"))
          .isInstanceOf(ServerOperationException.class)
          .getCause().isInstanceOfAny(AuthenticationFailedException.class,
          AuthenticationRequiredException.class);
```

jinmeiliao:
@boglesby thanks for the findings, but I haven't started working on registered interests yet. This story is only about CQs. I will work on Registered interests in my next story, but definitely will use your test scenario.

geode-core/src/distributedTest/java/org/apache/geode/management/internal/security/MultiUserAuthenticationDUnitTest.java
```diff
@@ -100,4 +112,65 @@ public void multiAuthenticatedView() throws Exception {
       cache.close();
     }
   }
+
+  @Test
+  public void multiUserCQ() throws Exception {
```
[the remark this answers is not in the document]
jinmeiliao: that's covered by other tests in the same class. I added this specific test to verify a bug I fixed in this PR. (before this test would fail with "subject not found" exception)

geode-core/src/test/java/org/apache/geode/internal/cache/tier/sockets/ServerConnectionTest.java
```diff
@@ -57,25 +67,37 @@
...
 
   private ServerConnection serverConnection;
+  private Command command;
+  private InetAddress inetAddress;
+  private Socket socket;
+  private InternalCacheForClientAccess cache;
+  private CachedRegionHelper cachedRegionHelper;
+  private InternalDistributedSystem internalDistributedSystem;
+  private DistributionManager distributionManager;
+  private ThreadsMonitoring threadsMonitoring;
```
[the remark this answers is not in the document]
jinmeiliao: I make it a habit to always declare them as class variables just in case any future tests would use them.

geode-core/src/test/java/org/apache/geode/internal/cache/tier/sockets/ServerConnectionTest.java
```diff
@@ -124,4 +146,97 @@ public void nonSecureShouldThrow() {
         .isExactlyInstanceOf(AuthenticationRequiredException.class)
         .hasMessage("No security credentials are provided");
   }
+
+  @Test
+  public void bindSubjectDoesNothingIfNotIntegratedService() throws Exception {
```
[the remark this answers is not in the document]
jinmeiliao: it's added by the IDE auto code block just in case any test would. Maybe I should change that default behavior.

geode-junit/src/main/java/org/apache/geode/test/version/VersionManager.java
```diff
@@ -143,6 +143,14 @@ public String getInstall(String version) {
     return unmodifiableList(testVersions);
   }
 
+  public List<String> getVersionsLaterThanAndEqualTo(String version) {
+    checkForLoadFailure();
+    List<String> result = new ArrayList<>(testVersions);
+    result.removeIf(s -> TestVersion.compare(s, version) < 0);
+    result.add(CURRENT_VERSION);
```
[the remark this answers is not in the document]
jinmeiliao: Yeah, that's true, but it's test code, hopefully the tester would know they passed in something invalid.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientReAuthenticateMessage.java
```diff
@@ -0,0 +1,104 @@
...
+  public ClientReAuthenticateMessage() {
+    this(new EventID());
+  }
+
+  public ClientReAuthenticateMessage(EventID eventId) {
+    this.eventId = eventId;
+  }
+
+  @Override
+  public Message getMessage(CacheClientProxy proxy, boolean notify) throws IOException {
```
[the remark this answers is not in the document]
jinmeiliao: it's a signature of the interface, I guess it's added by the IDE.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/OldClientSupportService.java
```diff
@@ -18,13 +18,16 @@
...
+import org.apache.geode.annotations.Immutable;
 import org.apache.geode.internal.cache.CacheService;
 import org.apache.geode.internal.serialization.KnownVersion;
 
 /**
  * Support for old GemFire clients
  */
 public interface OldClientSupportService extends CacheService {
+  @Immutable
+  KnownVersion RE_AUTHENTICATION_START_VERSION = KnownVersion.GEODE_1_15_0;
```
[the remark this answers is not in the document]
jinmeiliao: Good idea!

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ServerConnection.java
```diff
@@ -894,6 +876,50 @@ void doNormalMessage() {
...
+    }
+
+    if (isInternalMessage(requestMessage, allowInternalMessagesWithoutCredentials)) {
+      return null;
+    }
+
+    long uniqueId = getUniqueId();
+    String messageType = MessageType.getString(requestMessage.getMessageType());
+    if (uniqueId == 0 || uniqueId == -1) {
+      logger.debug("No unique ID yet. {}, {}", messageType, getName());
```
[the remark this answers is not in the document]
jinmeiliao: always not sure what's the benefit of that, I was just following what was done before. I hate to clutter up the code with more `if` statements if it's not necessary.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/command/PutUserCredentials.java
```diff
@@ -39,45 +39,44 @@ public void cmdExecute(final Message clientMessage, final ServerConnection serve
...
     boolean isSecureMode = clientMessage.isSecureMode();
 
-    // if (!isSecureMode)
-    // client has not send secuirty header, need to send exception and log this in security (file)
-
-    if (isSecureMode) {
-
-      int numberOfParts = clientMessage.getNumberOfParts();
+    if (!isSecureMode) {
+      // client has not send secuirty header, need to send exception and log this in security (file)
```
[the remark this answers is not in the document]
jinmeiliao: yeah, I just preserved the old comments when refactoring the nested ifs. I don't want to change the original behavior (at least not in this PR).

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/command/PutUserCredentials.java
```diff
@@ -39,45 +39,44 @@ public void cmdExecute(final Message clientMessage, final ServerConnection serve
...
-                    serverConnection.getProxyID(), ""),
-                ex);
-          }
-          writeException(clientMessage, ex, false, serverConnection);
-        } finally {
-          serverConnection.setAsTrue(RESPONDED);
-        }
+    int numberOfParts = clientMessage.getNumberOfParts();
+    if (numberOfParts != 1) {
+      // need to throw exception
```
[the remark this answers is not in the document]
jinmeiliao: again, do not want to change the old behavior on simple refatoring.

geode-core/src/test/java/org/apache/geode/internal/cache/tier/sockets/CacheClientProxyTest.java
```diff
@@ -0,0 +1,123 @@
...
+  private KnownVersion version;
+  private SecurityService securityService;
+  private Subject subject;
+  private StatisticsClock clock;
+  private InternalCache cache;
+  private StatisticsFactory statsFactory;
+  private CacheClientProxyStatsFactory proxyStatsFactory;
+  private MessageDispatcherFactory dispatcherFactory;
+  private InetAddress inetAddress;
+  private CacheServerStats stats;
```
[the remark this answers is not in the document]
jinmeiliao: Same as my other junit tests, I prefer to put them as class variable for future tests

geode-core/src/test/java/org/apache/geode/internal/cache/tier/sockets/MessageDispatcherTest.java
```diff
@@ -0,0 +1,167 @@
...
+  private ClientUpdateMessage message;
+  private InternalCache cache;
+  private SecurityService securityService;
+  private Subject subject;
+  private CacheClientNotifier notifier;
+  private ClientProxyMembershipID proxyId;
+  private HARegionQueue messageQueue;
+  private HARegionQueueStats queueStats;
+  private CancelCriterion cancelCriterion;
+  private CacheClientProxyStats proxyStats;
```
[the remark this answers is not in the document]
jinmeiliao: same reason.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ServerConnection.java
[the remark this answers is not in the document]
DonalEvans: I just checked and it seems that wrapping the log statement in that check has no effect on the behaviour. I guess that it was necessary at one point, because I see that pattern all over the place, but it seems that it's fine not to do it here. I learned something new today!

geode-core/src/test/java/org/apache/geode/internal/cache/tier/sockets/ServerConnectionTest.java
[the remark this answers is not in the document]
DonalEvans: That's reasonable

DonalEvans (approved):

boglesby:
I see the same basic exception in the CQ case, just a slightly different stack from the server:
```
[warn 2021/09/15 12:55:30.583 PDT client-cq <queueTimer-pool1> tid=0x1f] QueueManagerImpl failed to recover interest to server boglesbymac.local:59654
org.apache.geode.cache.client.ServerOperationException: remote server on boglesbymac(client-cq:11999:loner):59663:660c07eb:client-cq: org.apache.geode.security.AuthenticationRequiredException: Failed to find the authenticated user.
	at org.apache.geode.cache.client.internal.OpExecutorImpl.handleException(OpExecutorImpl.java:559)
	at org.apache.geode.cache.client.internal.OpExecutorImpl.handleException(OpExecutorImpl.java:623)
	at org.apache.geode.cache.client.internal.OpExecutorImpl.handleException(OpExecutorImpl.java:512)
	at org.apache.geode.cache.client.internal.OpExecutorImpl.executeOn(OpExecutorImpl.java:475)
	at org.apache.geode.cache.client.internal.OpExecutorImpl.executeOn(OpExecutorImpl.java:486)
	at org.apache.geode.cache.client.internal.PoolImpl.executeOn(PoolImpl.java:857)
	at org.apache.geode.cache.query.cq.internal.ops.CreateCQOp.executeOn(CreateCQOp.java:68)
	at org.apache.geode.cache.query.cq.internal.ops.ServerCQProxyImpl.createOn(ServerCQProxyImpl.java:76)
	at org.apache.geode.cache.query.cq.internal.ClientCQImpl.createOn(ClientCQImpl.java:574)
	at org.apache.geode.cache.client.internal.QueueManagerImpl.recoverCqs(QueueManagerImpl.java:1121)
	at org.apache.geode.cache.client.internal.QueueManagerImpl.recoverAllInterestTypes(QueueManagerImpl.java:1239)
	at org.apache.geode.cache.client.internal.QueueManagerImpl.recoverInterest(QueueManagerImpl.java:1069)
	at org.apache.geode.cache.client.internal.QueueManagerImpl.recoverPrimary(QueueManagerImpl.java:928)
	at org.apache.geode.cache.client.internal.QueueManagerImpl.access$600(QueueManagerImpl.java:77)
	at org.apache.geode.cache.client.internal.QueueManagerImpl$RedundancySatisfierTask.run2(QueueManagerImpl.java:1448)
	at org.apache.geode.cache.client.internal.PoolImpl$PoolTask.run(PoolImpl.java:1336)
Caused by: org.apache.geode.security.AuthenticationRequiredException: Failed to find the authenticated user.
	at org.apache.geode.internal.security.IntegratedSecurityService.getSubject(IntegratedSecurityService.java:123)
	at org.apache.geode.internal.security.IntegratedSecurityService.authorize(IntegratedSecurityService.java:249)
	at org.apache.geode.internal.security.IntegratedSecurityService.authorize(IntegratedSecurityService.java:244)
	at org.apache.geode.internal.security.IntegratedSecurityService.authorize(IntegratedSecurityService.java:230)
	at org.apache.geode.cache.query.cq.internal.command.ExecuteCQ61.lambda$cmdExecute$0(ExecuteCQ61.java:138)
	at java.lang.Iterable.forEach(Iterable.java:75)
	at java.util.Collections$UnmodifiableCollection.forEach(Collections.java:1080)
	at org.apache.geode.cache.query.cq.internal.command.ExecuteCQ61.cmdExecute(ExecuteCQ61.java:137)
	at org.apache.geode.internal.cache.tier.sockets.BaseCommand.execute(BaseCommand.java:184)
```
The server is processing the ExecuteCQ61. It looks like it binds the Subject in bindSubject, but later on it fails to find it.

Here is the exception on the server:
```
[warn 2021/09/15 12:55:30.563 PDT server1 <ServerConnection on port 59654 Thread 5> tid=0x4a] XXX ServerConnection.doNormalMessage about to bindSubject command=ExecuteCQ61

[warn 2021/09/15 12:55:30.564 PDT server1 <ServerConnection on port 59654 Thread 5> tid=0x4a] XXX ServerConnection.bindSubject command=org.apache.geode.cache.query.cq.internal.command.ExecuteCQ61; uniqueId=8287195257237307674; subject=org.apache.shiro.subject.support.DelegatingSubject@492e352e

[warn 2021/09/15 12:55:30.564 PDT server1 <ServerConnection on port 59654 Thread 5> tid=0x4a] XXX ServerConnection.doNormalMessage done bindSubject command=ExecuteCQ61

[warn 2021/09/15 12:55:30.565 PDT server1 <ServerConnection on port 59654 Thread 5> tid=0x4a] XXX ExecuteCQ61.cmdExecute caught:
org.apache.geode.security.AuthenticationRequiredException: Failed to find the authenticated user.
	at org.apache.geode.internal.security.IntegratedSecurityService.getSubject(IntegratedSecurityService.java:123)
	at org.apache.geode.internal.security.IntegratedSecurityService.authorize(IntegratedSecurityService.java:249)
	at org.apache.geode.internal.security.IntegratedSecurityService.authorize(IntegratedSecurityService.java:244)
	at org.apache.geode.internal.security.IntegratedSecurityService.authorize(IntegratedSecurityService.java:230)
	at org.apache.geode.cache.query.cq.internal.command.ExecuteCQ61.lambda$cmdExecute$0(ExecuteCQ61.java:138)
	at java.lang.Iterable.forEach(Iterable.java:75)
	at java.util.Collections$UnmodifiableCollection.forEach(Collections.java:1080)
	at org.apache.geode.cache.query.cq.internal.command.ExecuteCQ61.cmdExecute(ExecuteCQ61.java:137)
	at org.apache.geode.internal.cache.tier.sockets.BaseCommand.execute(BaseCommand.java:184)
	at org.apache.geode.internal.cache.tier.sockets.ServerConnection.doNormalMessage(ServerConnection.java:883)
	at org.apache.geode.internal.cache.tier.sockets.ServerConnection.doOneMessage(ServerConnection.java:1072)
	at org.apache.geode.internal.cache.tier.sockets.ServerConnection.run(ServerConnection.java:1341)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at org.apache.geode.internal.cache.tier.sockets.AcceptorImpl.lambda$initializeServerConnectionThreadPool$3(AcceptorImpl.java:690)
	at org.apache.geode.logging.internal.executors.LoggingThreadFactory.lambda$newThread$0(LoggingThreadFactory.java:120)
	at java.lang.Thread.run(Thread.java:745)
```
I do see that it depends on where I put the sleep.

If I put it in `CacheClientUpdater handleAuthenticate` on the client before sending the `AuthenticateUserOp`, the test works fine.

If I put it in `PutUserCredentials cmdExecute` on the server before the command is processed, the test fails.

boglesby (approved):

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientReAuthenticateMessage.java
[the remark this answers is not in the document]
Bill: ok

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/MessageDispatcher.java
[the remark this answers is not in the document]
agingade: Jinmei, Darrel has additional comment on this...I am changing the state to unresolved.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/MessageDispatcher.java
agingade: Can the sleep be replaced with wait/notify?
Here it is expecting client to do a re-auth...Can a wait be done for re-auth to happen from client and then notify this wait when re-auth is received from client (from the re-auth command handling code).

Bill: Thanks for that explanation. It helps a lot.

Whenever the dispatcher catches an `AuthExpirationException`, if it doesn't terminate the connection, all message processing is paused for 200ms.

I'm wondering how often does authentication expire generally? The product doesn't ship with a security manager that implements expiration (other than a test class), so it would be up to each particular customer's security manager implementation right?

Do you think it's ok to sleep for 200ms each time or do you think it would be worth the effort to shorten that time window by introducing say, a latch (we'd do an `await()` at the points where we're currently sleeping, and we'd issue a `countDown()` at the point where the client re-authenticates). **Is there an obvious place in the product where we could issue that `countDown()`?**

PS: In evaluating this change I looked for how `ClientReAuthenticateMessage` is used but I don't see it used anywhere in the product except where it's thrown in this class (and registered as a DSFID). That leads me to believe that a user's security manager would have to register a handler for that message. `ClientReAuthenticateMessage` is not mentioned in the product documentation nor in examples so I'm just guessing here.

jinmeiliao: > Can the sleep be replaced with wait/notify?
> Here it is expecting client to do a re-auth...Can a wait be done for re-auth to happen from client and then notify this wait when re-auth is received from client (from the re-auth command handling code).

We are actually not waiting for client to re-auth, 'cause 1) the client might be an old client that will never re-authenticate back, 2) even the new client maybe not re-authenticate back successfully. As far as the server is concerned, I am just keep trying to "authorize" the message for a certain period of time (5 seconds by default), if in the end of this period, I still can not deliver this message, then I am closing this connection.

jinmeiliao: > Thanks for that explanation. It helps a lot.
> 
> Whenever the dispatcher catches an `AuthExpirationException`, if it doesn't terminate the connection, all message processing is paused for 200ms.
> 
> I'm wondering how often does authentication expire generally? The product doesn't ship with a security manager that implements expiration (other than a test class), so it would be up to each particular customer's security manager implementation right?
> 
> Do you think it's ok to sleep for 200ms each time or do you think it would be worth the effort to shorten that time window by introducing say, a latch (we'd do an `await()` at the points where we're currently sleeping, and we'd issue a `countDown()` at the point where the client re-authenticates). **Is there an obvious place in the product where we could issue that `countDown()`?**
> 
> PS: In evaluating this change I looked for how `ClientReAuthenticateMessage` is used but I don't see it used anywhere in the product except where it's thrown in this class (and registered as a DSFID). That leads me to believe that a user's security manager would have to register a handler for that message. `ClientReAuthenticateMessage` is not mentioned in the product documentation nor in examples so I'm just guessing here.

Like I replied to Anil, there is not a definite event that the server needs to "wait" for,  the client may never re-authenticate back, so all the server can rely on is to keep trying to deliver the message with the "subject" it knows about (with 200 ms interval) for 5 seconds (configurable by system property).

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/MessageDispatcher.java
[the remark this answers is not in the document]
jinmeiliao: Notice this is NotAuthorizedException (not AuthenticationExpiredException), when this exception happens, we only log it, drop it from the queue and continue on to the next message (that's what the old authorization model does to the un-authorized messages), we don't do retrys for any "un-authorized" messages.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/CacheClientUpdater.java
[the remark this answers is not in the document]
agingade: This is getting processed by CacheClientUpdater thread...As you can see from the thread name, this thread is used only for local cache operation; its not used to do client-to-server operation. I am worried about the exception it could see by doing server side operation....And its impact on overall closing CCU thread and re-creating it. In the past we had several issues with older and new CCU threads running simultaneously. Without sufficient tests its hard to see its impact. It will be nice to process this in separate thread than the CCU thread.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/MessageDispatcher.java
agingade: As we know sleep is not the best way to wait for certain state to be changed; its a last resort if there is no option. Based on the explanation, what I see is the code is expecting a state change from two different client:
1. Old client:
In this case, if it is an old client disconnect the server to client connection immediately. At this level we know the clients versions (from the client proxy); if it is an old version, the connection could be closed immediately without any wait.
2. New Client: 
In this case, the expectation is the new client will re-auth; for java clients the re-auth is processed by the command or security service; using the client info the wait in the dispatch could be notified. 
These are just thoughts, not sure if there are any hurdle in doing this way. It will be good investigate to see if this is possible.

jinmeiliao: Sure， we can add a task to investigate this, but I don't want to further muddle up this PR anymore.

jinmeiliao: About the client, we don't immediately close the connection to old client and still choose to wait because any active client interaction will trigger a re-auth automatically, so there is still chance that the message could be delivered to the old client, we have a test specifically for that: `AuthExpirationDUnitTest.cqOlderClientWithClientInteractionWillDeliverEventEventually`

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/CacheClientUpdater.java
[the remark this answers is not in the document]
jinmeiliao: I didn't put this "re-authenticate" in my initial implementation because
1) if we get this message from the server, no other message will be delivered to this client until we re-authenticate back successfully.
2) even if we put it in a separate thread, we will need to see if any exception happens on that thread and if so, we need to close the thread anyway.

I haven't seen much benefit of putting this in a different thread yet, but I don't see much harm in putting in a different thread either. Since it can go either way, I much rather wait for this PR to settle before I put in more "improvement".

jinmeiliao:
the acceptance test failure happens on the develop pipeline as well.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/MessageDispatcher.java
Bill: Adding wait/notify semantics between the client re-authenticating and the `MessageDispatcher` waiting to start trying (to dispatch) again will add some complexity to the existing `SecurityManager` framework.

I now see that this PR (and associated ticket) are one task in a larger effort [GEODE-9451](https://issues.apache.org/jira/browse/GEODE-9451) covered by an [RFC](https://cwiki.apache.org/confluence/display/GEODE/On+Demand+Geode+Authentication+Expiration+and+Re-authentication).

In that context, I think @jinmeiliao's answer: "we can add a task to investigate this, but I don't want to further muddle up this PR anymore" is acceptable.

Bill (approved):
Approved.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/CacheClientUpdater.java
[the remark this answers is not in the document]
agingade: @jinmeiliao my concerns are:
Point 1:
Yes that is true.
Point 2:
If there is exception, then thread can handle according to the exception. And if it needs to close the connection, it can. But what I am trying to convey here is the CacheClientUpdater thread is not meant to be used for server cache operation. And currently its handling the exception based on the expectation of server to client communication. I am not seeing any tests/code changes in CacheClientUpdater exception handling to handle the exception from client to server communication which this PR is introducing.

PR #6837, opened 2021-09-02
GEODE-9561:  jmx-manager-ssl-enabled supersedes jmx-manager-ssl

davebarnes97:
jmx-manager-ssl, which no longer exists, was superseded by jmx-manager-ssl-enabled, which does exist but has subsequently been deprecated.

For now, we’ll fix the small issue in front of us by updating the property to jmx-manager-ssl-enabled.


kohlmu-pivotal (approved):
LGTM

PR #6844, opened 2021-09-06
GEODE-9546: Integrate Security Manager into Radish AUTH flow

jdeppe-pivotal:

### For all changes:

### Note:

### DIFF
--- a/geode-apis-compatible-with-redis/README.md
+++ b/geode-apis-compatible-with-redis/README.md
@@ -124,6 +124,8 @@ The `SimpleSecurityManager` is only to be used **for demonstration purposes**. I
 
 Note that the `compatible-with-redis-username` property is only needed if `AUTH` commands are issued without a username. In this case, the Security Manager will need to respond to authentication requests using this username.
 
+Note also that _any_ `AUTH` requests will fail if no Security Manager has been configured.
+
 ## <a name="application-developement"></a>Application Developement
 
 ### Things to know before you begin
--- a/geode-apis-compatible-with-redis/README.md
+++ b/geode-apis-compatible-with-redis/README.md
@@ -4,32 +4,38 @@ Note: This feature is experimental and is subject to change in future releases o
 
 [Introduction](#introduction)  
 [How To Try It](#how-to-try-it)  
-[Building Apache Geode](#building-apache-geode)  
-[Starting a Geode Server with Redis Enabled](#starting-a-server)  
-[Adding an Additional Geode Server Compatible with Redis](#adding-a-server)  
+[Add an Additional Geode Server Compatible with Redis](#adding-a-server)  
 [Shutting Down](#shutting-down)  
+[Security](#security)  
 [Redis Commands](#redis-commands)
 
 ## <a name="introduction"></a>Introduction
 
-The Geode APIs compatible with Redis allow Geode to function as a drop-in replacement for a highly-available Redis data store, letting Redis applications take advantage of Geode’s scaling capabilities without changing their client code. 
+The Geode APIs compatible with Redis allows Redis clients to use Geode as a highly-available Redis data store. Redis applications can take advantage of Geode’s scaling capabilities with minimal changes to the client code (Redis clients must be compatible with Redis CLUSTER commands).
 
-Redis clients connect to a Geode server in the same way they connect to a Redis server, using a hostname and a port number, with optional username/password authentication.
+Redis clients connect to a Geode server in the same way they connect to a Redis server, using a hostname and a port number.
 
-### <a name="redis-commands"></a>Supported Redis Commands
+## <a name="redis-commands"></a>Supported Redis Commands
 
-Not all Redis commands are supported. See [Supported Redis Commands](#supported-redis-commands) for the implemented subset.
+Not all Redis commands are supported at this time. See [Supported Redis Commands](#supported-redis-commands) for the implemented subset.
 
 ## <a name="how-to-try-it"></a>How To Try It
 
-Install and configure Geode v1.14 or later.
+Install and configure Geode v1.15 or later.
 
-Use gfsh to start at least one server with a command of the form:
+In a terminal, start Geode using the `gfsh` command.
 
+With `gfsh` running, in the same terminal, start one [locator](https://geode.apache.org/docs/guide/12/configuring/running/running_the_locator.html) 
+
+```
+gfsh> start locator
+```
+
+Once the locator has started, start at least one server with a command of the form:
 ```console
-start server \
+gfsh> start server \
   --name=<serverName> \
-  --locators=<locatorPort> \
+  --locators=localhost[<locatorPort>] \
   --compatible-with-redis-port=<compatibleWithRedisPort> \
   --compatible-with-redis-bind-address=<compatibleWithRedisBindAddress>
 ```
@@ -53,56 +59,30 @@ Your Geode instance should now be up and running (1 locator and 1 server) and re
 To confirm the server is listening, in a separate terminal run:
 
 ```console
-redis-cli -h <compatibleWithRedisBindAddress> -p <compatibleWithRedisPort> ping
+$> redis-cli -c -h <compatibleWithRedisBindAddress> -p <compatibleWithRedisPort> ping
 ```
 
 - Replace `<compatibleWithRedisBindAddress>`, `<compatibleWithRedisPort>`, and `<compatibleWithRedisPassword>` with the same values as the server.
 
 If the server is functioning properly, you should see a response of `PONG`.
 
-## <a name="security"></a>Security
+### <a name="adding-a-server"></a> Add an additional Geode server compatible with Redis APIs
+If you’re interested in testing Geode scalability, in gfsh run the `start server` command again **BUT**
+make sure you change the `--name=` and `--redis-port=` parameters.
 
-Security is implemented slightly differently to Redis. To enable security, a Security Manager needs to be configured. This Security Manager will authenticate `AUTH username password` commands. Instead of a default, system-wide password, a default username can be set using the `geode-compatible-with-redis-username` parameter. This username is used when `AUTH` commands are sent with only a password.
-
-For example, the following gfsh command will configure a `SimpleSecurityManager`:
-
-```console
-start server \
-  --name=<serverName> \
-  --locators=<locatorPort> \
-  --compatible-with-redis-port=<compatibleWithRedisPort> \
-  --compatible-with-redis-bind-address=<compatibleWithRedisBindAddress> \
-  --compatible-with-redis-username=<compatibleWithRedisUsername> \
-  --J=-Dgemfire.security-manager=org.apache.geode.examples.SimpleSecurityManager
+For example:
+```commandLine
+gfsh> start server --name=redisServer2 --locators=localhost[10334] --server-port=0 --compatible-with-redis-port=6380
 ```
 
-To confirm that the server is working, in a separate terminal run:
-
-```console
-redis-cli -h <compatibleWithRedisBindAddress> -p <compatibleWithRedisPort> --user <compatibleWithRedisUsername> -a <compatibleWithRedisUsername> ping
-```
-
-The `SimpleSecurityManager` is only to be used for demonstration purposes and will successfully authenticate if the provided password is the same as the user.
-
-Note that the `compatible-with-redis-username` property is only needed if `AUTH` commands are issued without a username. In this case, the Security Manager will need to respond to authentication requests using this username.
-
-### <a name="adding-a-server"></a>Optional - Adding an additional Geode server with compatible with Redis APIs
-If you’re interested in testing Geode scalability, in gfsh run the `start server` command again BUT 
-make sure you change the `--name=` and `--redis-port=` parameters. 
-
-For example: 
-   ```commandLine
-   $ start server --name=redisServer2 --locators=localhost[10334] --server-port=0 --compatible-with-redis-port=6380
-   ```
-
-### <a name="shutting-down"></a>Shutting Down 
+### <a name="shutting-down"></a>Shutting Down
 To shut down the Geode instance you started, in the terminal with gfsh running type the following command
 
 ```commandLine
-$ shutdown --include-locators=true
+gfsh> shutdown --include-locators=true
 ```
-	
-This command shuts down the entire Geode instance/cluster. You are prompted with the following choice: 
+
+This command shuts down the entire Geode instance/cluster. You are prompted with the following choice:
 
 ```commandline
 As a lot of data in memory will be lost, including possibly events in queues, do you really want to shutdown the entire distributed system? (Y/n)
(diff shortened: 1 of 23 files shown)


onichols-pivotal (approved):

jdeppe-pivotal:
@jomartin-999 Please would you review the README changes.

geode-apis-compatible-with-redis/README.md
```diff
@@ -56,14 +53,40 @@ Your Geode instance should now be up and running (1 locator and 1 server) and re
...
+  --compatible-with-redis-port=<compatibleWithRedisPort> \
+  --compatible-with-redis-bind-address=<compatibleWithRedisBindAddress> \
+  --compatible-with-redis-username=<compatibleWithRedisUsername> \
+  --J=-Dgemfire.security-manager=org.apache.geode.examples.SimpleSecurityManager
+```
+
+To confirm that the server is working, in a separate terminal run:
+
+```console
+redis-cli -h <compatibleWithRedisBindAddress> -p <compatibleWithRedisPort> --user <compatibleWithRedisUsername> -a <compatibleWithRedisUsername> ping
```
jomartin-999, line 82: I believe they will need to add "-c" when starting the cli in cluster mode.

We should also link to or have an example of how one would actually use a security manager in production.  For example, users would include the jar in their server startup...and do they have to include it in every startup?  Also noting that that if they don't start with a security manager, they'll have to....delete the instance and start a new one?

jomartin-999 (commented):
Added come feedback on the Readme.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/connection/AbstractAuthIntegrationTest.java
```diff
@@ -0,0 +1,183 @@
...
+import io.lettuce.core.RedisClient;
+import io.lettuce.core.RedisURI;
+import org.junit.Test;
+import redis.clients.jedis.Jedis;
+import redis.clients.jedis.Protocol;
+
+
+public abstract class AbstractAuthIntegrationTest {
+
+  static final String USERNAME = "default";
```
DonalEvans, line 29: Could this be set as `DistributionConfig#DEFAULT_REDIS_USERNAME` to avoid the test failing if the default value changes?

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/connection/AbstractAuthIntegrationTest.java
```diff
@@ -0,0 +1,183 @@
...
+  // Since we're going to use a SimpleSecurityManager where password == username means success
+  static final String PASSWORD = USERNAME;
+  Jedis jedis;
+
+  protected abstract void setupCacheWithSecurity(boolean withSecurityManager) throws Exception;
```
DonalEvans, line 34: The value passed for `withSecurityManager` is always `true`. Do we need this method to take an argument if we also have the `setupCacheWithoutSecurity()` method? Alternately, are tests needed for the case where `withSecurityManager` is false?

DonalEvans (approved):

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/connection/AbstractAuthIntegrationTest.java
jdeppe-pivotal: This is also used to authenticate against Redis' default username of `default`. Applying the suggestion would still cause tests to fail, in the native acceptance tests, if the Geode default changes.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/connection/AbstractAuthIntegrationTest.java
jdeppe-pivotal: I've removed the parameter.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/connection/AuthExecutor.java
```diff
@@ -15,35 +15,55 @@
...
-      ExecutionHandlerContext context) {
+  public RedisResponse executeCommand(Command command, ExecutionHandlerContext context) {
     List<byte[]> commandElems = command.getProcessedCommand();
 
-    byte[] password = context.getAuthPassword();
-    if (password == null) {
-      return RedisResponse.error(RedisConstants.ERROR_NO_PASS);
+    SecurityManager securityManager = context.getSecurityManager();
+    Properties props = new Properties();
+    if (commandElems.size() == 2) {
```
jinmeiliao, line 40: nitpick: probably do early returns to avoid nested if/elses.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/connection/AuthExecutor.java
```diff
@@ -15,35 +15,55 @@
...
   @Override
-  public RedisResponse executeCommand(Command command,
-      ExecutionHandlerContext context) {
+  public RedisResponse executeCommand(Command command, ExecutionHandlerContext context) {
     List<byte[]> commandElems = command.getProcessedCommand();
 
-    byte[] password = context.getAuthPassword();
-    if (password == null) {
-      return RedisResponse.error(RedisConstants.ERROR_NO_PASS);
+    SecurityManager securityManager = context.getSecurityManager();
```
jinmeiliao, line 38: Is it possible to get the securtyService out of context? SecurityManager is supposed to be an internal component used by the SecurityService, the product should interact only with SecurityService.

echobravopapa (approved):

dschneider-pivotal (changes requested):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/connection/AuthExecutor.java
```diff
@@ -15,35 +15,55 @@
...
+      if (securityManager == null) {
+        return RedisResponse.error(ERROR_AUTH_CALLED_WITHOUT_PASSWORD_CONFIGURED);
+      }
+      props.setProperty("security-username", context.getRedisUsername());
+      props.setProperty("security-password", bytesToString(commandElems.get(1)));
+    } else {
+      if (securityManager == null) {
+        String receivedUsername = new String(commandElems.get(1));
+        if (receivedUsername.equalsIgnoreCase(context.getRedisUsername())) {
+          return RedisResponse.ok();
```
upthewaterspout, line 50: Yesterday we discussed that we think we ought to deviate for the native redis behavior here. If there is no security manager, *any* attempt to call auth should fail. Hopefully that can make the logic in this method a bit simpler.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/connection/AuthExecutor.java
```diff
@@ -15,35 +15,55 @@
...
-    if (password == null) {
-      return RedisResponse.error(RedisConstants.ERROR_NO_PASS);
+    SecurityManager securityManager = context.getSecurityManager();
+    Properties props = new Properties();
+    if (commandElems.size() == 2) {
+      if (securityManager == null) {
+        return RedisResponse.error(ERROR_AUTH_CALLED_WITHOUT_PASSWORD_CONFIGURED);
+      }
+      props.setProperty("security-username", context.getRedisUsername());
+      props.setProperty("security-password", bytesToString(commandElems.get(1)));
```
upthewaterspout, line 45: I think you can use the constants from `SecurityManager` for strings like "security-username"

geode-core/src/main/java/org/apache/geode/distributed/internal/AbstractDistributionConfig.java
```diff
@@ -1317,7 +1317,7 @@ static Class _getAttributeType(String attName) {
         "Specifies the address on which the Redis API for Geode is listening. If set to the empty string or this property is not specified, localhost is requested from the operating system.");
     m.put(REDIS_ENABLED,
         "When the default value of false, the Redis API for Geode is not available.  Set to true to enable the Redis API for Geode.");
-    m.put(REDIS_PASSWORD,
+    m.put(REDIS_USERNAME,
         "Specifies the password that the server uses when a client attempts to authenticate. The default is none and no authentication will be required.");
```
upthewaterspout, line 1321: I think you need to change the description here as well - it still refers to a password.

mhansonp (approved):
There are some requested changes by others that are needed, but beyond that, it looks good.

geode-apis-compatible-with-redis/README.md
jdeppe-pivotal: One is the username and one is the password (-a). Since we're saying to use the `SimpleSecurityManager` they should be the same.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/connection/AuthExecutor.java
jdeppe-pivotal: Changed a lot since we're returning an error now when there is no SecurityManager.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/connection/AuthExecutor.java
jdeppe-pivotal: Done.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/netty/NettyRedisServer.java
```diff
@@ -86,18 +85,20 @@
   private final Channel serverChannel;
   private final int serverPort;
   private final DistributedMember member;
+  private final SecurityManager securityManager;
 
   public NettyRedisServer(Supplier<DistributionConfig> configSupplier,
       RegionProvider regionProvider, PubSub pubsub, Supplier<Boolean> allowUnsupportedSupplier,
       Runnable shutdownInvoker, int port, String requestedAddress, RedisStats redisStats,
-      DistributedMember member) {
+      DistributedMember member, SecurityManager securityManager) {
```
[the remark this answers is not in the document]
jdeppe-pivotal: Done

geode-core/src/main/java/org/apache/geode/distributed/internal/AbstractDistributionConfig.java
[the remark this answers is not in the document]
jdeppe-pivotal: Done

geode-core/src/main/java/org/apache/geode/distributed/internal/DistributionConfig.java
```diff
@@ -3515,24 +3515,24 @@ static InetAddress _getDefaultMcastAddress() {
...
 
   /**
-   * Returns the value of the {@link ConfigurationProperties#REDIS_PASSWORD} property
+   * Returns the value of the {@link ConfigurationProperties#REDIS_USERNAME} property
    * <p>
    * Returns the value of the
-   * {@link ConfigurationProperties#REDIS_PASSWORD} property
+   * {@link ConfigurationProperties#REDIS_USERNAME} property
    *
    * @return the authentication password for GeodeRedisServer
```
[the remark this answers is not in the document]
jdeppe-pivotal: Done

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/StartServerCommand.java
```diff
@@ -290,7 +290,7 @@ ResultModel doStartServer(String memberName, Boolean assignBuckets, String bindA
         redisPort);
     StartMemberUtils.setPropertyIfNotNull(gemfireProperties,
         ConfigurationProperties.REDIS_BIND_ADDRESS, redisBindAddress);
-    StartMemberUtils.setPropertyIfNotNull(gemfireProperties, ConfigurationProperties.REDIS_PASSWORD,
+    StartMemberUtils.setPropertyIfNotNull(gemfireProperties, ConfigurationProperties.REDIS_USERNAME,
         redisPassword);
```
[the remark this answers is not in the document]
jdeppe-pivotal: Done

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/connection/AuthExecutor.java
jdeppe-pivotal: I made this change. In this case the returned error is `AUTH called without a Security Manager configured.`

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/connection/AuthExecutor.java
jdeppe-pivotal: Done

geode-core/src/main/java/org/apache/geode/distributed/internal/AbstractDistributionConfig.java
jdeppe-pivotal: Done

geode-apis-compatible-with-redis/README.md
```diff
@@ -91,68 +94,139 @@ To confirm that everything shut down correctly, if you execute a Redis command i
...
 ```
-### <a name="redis-commands"></a>Redis Commands
+
+## <a name="security"></a>Security
+
+Security is implemented slightly differently to OSS Redis. Redis stores password information in plain text in the redis.conf file.     
+
+When using Apache Geode, to enable security, a Security Manager needs to be configured on the server(s). This Security Manager will authenticate `AUTH <password>` commands and `AUTH <username> <password>` commands. Users can set a custom `default` username using the `geode-compatible-with-redis-username` parameter. This username will be used when `AUTH <password>` commands are sent without a `<username>`. 
+
+The following gfsh command will configure a `SimpleSecurityManager`:
```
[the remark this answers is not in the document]
jdeppe-pivotal: We can add that in the next bit of work which is for authorization.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/connection/AuthExecutor.java
```diff
@@ -15,35 +15,54 @@
...
+      props.setProperty(SecurityManager.USER_NAME, context.getRedisUsername());
+      props.setProperty(SecurityManager.PASSWORD, bytesToString(commandElems.get(1)));
     } else {
-      return RedisResponse.error(RedisConstants.ERROR_INVALID_PWD);
+      props.setProperty(SecurityManager.USER_NAME, bytesToString(commandElems.get(1)));
+      props.setProperty(SecurityManager.PASSWORD, bytesToString(commandElems.get(2)));
     }
+
+    try {
+      Subject subject = securityService.login(props);
```
jinmeiliao, line 59: to avoid resource leak, a login should have a corresponding logout when we know the subject has come to an end. we should either call `securiityService.logout()` if the subject is bound to the thread, or `subject.logout()` sometime. So probably it's best to save the subject to the context instead of just the principal, so that we can call logout in the end.

geode-apis-compatible-with-redis/README.md
jinmeiliao: but for documentation, shouldn't it point out -a takes password? The fact that it uses the username here is only for a specific security manager.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/connection/AuthExecutor.java
jdeppe-pivotal: Ah - thanks that's a good point.

jinmeiliao (approved):

dschneider-pivotal (approved):

PR #6847, opened 2021-09-08
GEODE-9585: adding support for the KEYSLOT command

upthewaterspout:
Adding support for the keyslot command. refactoring the RedisKey code a little
bit to move the slot calculation into a static method that can be used and
tested separately. Changing RedisKey.crc16 to slot, because that is only thing
we are using the crc16 field for.

### For all changes:

### Note:

### DIFF
--- a/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/RegionProvider.java
+++ b/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/RegionProvider.java
@@ -206,7 +206,7 @@ public class RegionProvider {
 
   private RedisDataMovedException createRedisDataMovedException(RedisKey key) {
     RedisMemberInfo memberInfo = getRedisMemberInfo(key);
-    Integer slot = key.getSlot();
+    int slot = key.getSlot();
     return new RedisDataMovedException(slot, memberInfo.getHostAddress(),
         memberInfo.getRedisPort());
   }
--- a/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/RegionProvider.java
+++ b/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/RegionProvider.java
@@ -206,7 +206,7 @@ public class RegionProvider {
 
   private RedisDataMovedException createRedisDataMovedException(RedisKey key) {
     RedisMemberInfo memberInfo = getRedisMemberInfo(key);
-    Integer slot = key.getCrc16() & (REDIS_SLOTS - 1);
+    Integer slot = key.getSlot();
     return new RedisDataMovedException(slot, memberInfo.getHostAddress(),
         memberInfo.getRedisPort());
   }
--- a/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/KeyHashUtil.java
+++ b/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/KeyHashUtil.java
@@ -1,0 +1,51 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.apache.geode.redis.internal.data;
+
+import static org.apache.geode.redis.internal.RegionProvider.REDIS_SLOTS;
+
+import org.apache.geode.redis.internal.executor.cluster.CRC16;
+
+/**
+ * Utility for parsing the hashtags in a redis key and computing the redis slot of the key.
+ */
+public class KeyHashUtil {
+
+  /**
+   * Compute the slot of a given redis key
+   *
+   * @return the slot, a value between 0 and 16K.
+   */
+  public static int slotForKey(byte[] value) {
+    int startHashtag = Integer.MAX_VALUE;
+    int endHashtag = 0;
+
+    for (int i = 0; i < value.length; i++) {
+      if (value[i] == '{' && startHashtag == Integer.MAX_VALUE) {
+        startHashtag = i;
+      } else if (value[i] == '}') {
+        endHashtag = i;
+        break;
+      }
+    }
+
+    if (endHashtag - startHashtag <= 1) {
+      startHashtag = -1;
+      endHashtag = value.length;
+    }
+
+    return CRC16.calculate(value, startHashtag + 1, endHashtag) & (REDIS_SLOTS - 1);
+  }
+}
--- a/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/KeyHashUtil.java
+++ b/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/KeyHashUtil.java
@@ -15,6 +15,8 @@
 package org.apache.geode.redis.internal.data;
 
 import static org.apache.geode.redis.internal.RegionProvider.REDIS_SLOTS;
+import static org.apache.geode.redis.internal.netty.StringBytesGlossary.LEFT_BRACE_ID;
+import static org.apache.geode.redis.internal.netty.StringBytesGlossary.RIGHT_BRACE_ID;
 
 import org.apache.geode.redis.internal.executor.cluster.CRC16;
 
@@ -28,24 +30,36 @@ public class KeyHashUtil {
    *
    * @return the slot, a value between 0 and 16K.
    */
-  public static int slotForKey(byte[] value) {
-    int startHashtag = Integer.MAX_VALUE;
-    int endHashtag = 0;
+  public static short slotForKey(byte[] key) {
+    int startHashtag;
+    int endHashtag;
 
-    for (int i = 0; i < value.length; i++) {
-      if (value[i] == '{' && startHashtag == Integer.MAX_VALUE) {
-        startHashtag = i;
-      } else if (value[i] == '}') {
-        endHashtag = i;
+    for (startHashtag = 0; startHashtag < key.length; startHashtag++) {
+      if (key[startHashtag] == LEFT_BRACE_ID) {
         break;
       }
     }
 
-    if (endHashtag - startHashtag <= 1) {
-      startHashtag = -1;
-      endHashtag = value.length;
+    // No starting hashtag, hash the whole thing
+    if (startHashtag == key.length) {
+      return crc16(key, 0, key.length);
     }
 
-    return CRC16.calculate(value, startHashtag + 1, endHashtag) & (REDIS_SLOTS - 1);
+    for (endHashtag = startHashtag + 1; endHashtag < key.length; endHashtag++) {
+      if (key[endHashtag] == RIGHT_BRACE_ID) {
+        break;
+      }
+    }
+
+    // No ending hashtag, or empty hashtag. Hash the whole thing.
+    if (endHashtag == key.length || endHashtag == startHashtag + 1) {
+      return crc16(key, 0, key.length);
+    }
+
+    return crc16(key, startHashtag + 1, endHashtag);
+  }
+
+  private static short crc16(byte[] key, int start, int end) {
+    return (short) (CRC16.calculate(key, start, end) & (REDIS_SLOTS - 1));
   }
 }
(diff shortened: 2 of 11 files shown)


geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisKey.java
```diff
@@ -29,41 +28,23 @@
...
-      startHashtag = -1;
-      endHashtag = value.length;
-    }
-
-    crc16 = CRC16.calculate(value, startHashtag + 1, endHashtag);
+    slot = KeyHashUtil.slotForKey(value);
   }
 
   public int getBucketId() {
     // & (REDIS_SLOTS - 1) is equivalent to % REDIS_SLOTS but supposedly faster
```
dschneider-pivotal, line 46: remove this comment since it no longer does & (REDIS_SLOTS - 1)

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/KeyHashUtil.java
```diff
@@ -0,0 +1,51 @@
...
+ * Utility for parsing the hashtags in a redis key and computing the redis slot of the key.
+ */
+public class KeyHashUtil {
+
+  /**
+   * Compute the slot of a given redis key
+   *
+   * @return the slot, a value between 0 and 16K.
+   */
+  public static int slotForKey(byte[] value) {
```
dschneider-pivotal, line 31: since your comments refer to a "redis key" should the parameter be named "key" instead of "value"?

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/KeyHashUtil.java
```diff
@@ -0,0 +1,51 @@
...
+    int startHashtag = Integer.MAX_VALUE;
+    int endHashtag = 0;
+
+    for (int i = 0; i < value.length; i++) {
```
dschneider-pivotal, line 35: If you have a key that looks like this "aaa}bbb{tag}ccc" this parsing will ignore the "{tag}" and say the whole key is the hash tag. Is that how native redis also does it?

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/RegionProvider.java
```diff
@@ -206,7 +206,7 @@ public RedisData getRedisData(RedisKey key, RedisData notFoundValue) {
 
   private RedisDataMovedException createRedisDataMovedException(RedisKey key) {
     RedisMemberInfo memberInfo = getRedisMemberInfo(key);
-    Integer slot = key.getCrc16() & (REDIS_SLOTS - 1);
+    Integer slot = key.getSlot();
```
dschneider-pivotal, line 209: change "Integer" to "int" here and on RedisDataMovedException

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/parameters/ClusterParameterRequirements.java
```diff
@@ -27,7 +27,7 @@ public void checkParameters(Command command, ExecutionHandlerContext context) {
 
     if (numberOfArguments < 2) {
       throw new RedisParametersMismatchException(command.wrongNumberOfArgumentsErrorMessage());
-    } else if (numberOfArguments > 2) {
+    } else if (numberOfArguments > 3) {
```
dschneider-pivotal, line 30: It is not clear to me why this change was made. If it was wrong before should a test have been broken? Can we add a test to the integration test that passes CLUSTER the wrong number of parameters?

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisKeyJUnitTest.java
```diff
@@ -39,43 +40,45 @@ public static void classSetup() {
...
     key = new RedisKey(new byte[] {});
-    assertThat(key.getCrc16()).isEqualTo(CRC16.calculate(""));
+    assertThat(key.getSlot()).isEqualTo(CRC16.calculate("") % RegionProvider.REDIS_SLOTS);
   }
 
   @Test
   public void testSerialization_withPositiveSignedShortCRC16() throws Exception {
     RedisKey keyOut = new RedisKey(stringToBytes("012345"));
-    assertThat((short) keyOut.getCrc16()).isPositive();
+    assertThat((short) keyOut.getSlot()).isPositive();
```
dschneider-pivotal, line 81: should this just be "testSerialization" and not worry about the slot being positive as a short? If we wanted to test that slot is always >= 0 it seems like it should be in its own test method.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/KeyHashUtil.java
```diff
@@ -0,0 +1,51 @@
...
+      if (value[i] == '{' && startHashtag == Integer.MAX_VALUE) {
```
DonalEvans, line 36: In the `StringBytesGlossary` class, there's a comment (originally in the Coder class before the constants were pulled out into their own class) stating:
> Do not use '' <-- java primitive chars. Redis uses `{@link Coder#CHARSET}` encoding so we should
not risk java handling char to byte conversions, rather just hard code `{@link Coder#CHARSET}`
chars as bytes
Is this a concern here?

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisKey.java
```diff
@@ -29,41 +28,23 @@
 import org.apache.geode.internal.serialization.DeserializationContext;
 import org.apache.geode.internal.serialization.KnownVersion;
 import org.apache.geode.internal.serialization.SerializationContext;
-import org.apache.geode.redis.internal.executor.cluster.CRC16;
 import org.apache.geode.redis.internal.executor.cluster.RedisPartitionResolver;
 
 public class RedisKey implements DataSerializableFixedID {
 
-  private int crc16;
+  private int slot;
```
dschneider-pivotal, line 35: did you consider changing this field to "short" since its value will always be in the range 0..REDIS_SLOTS-1?

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisKey.java
```diff
@@ -92,15 +73,15 @@ public int getDSFID() {
...
+    out.writeShort(slot);
     DataSerializer.writeByteArray(value, out);
   }
 
   @Override
   public void fromData(DataInput in, DeserializationContext context)
       throws IOException {
     // Need to convert a signed short to unsigned
-    crc16 = in.readShort() & 0xffff;
+    slot = in.readShort() & 0xffff;
```
dschneider-pivotal, line 84: since slot will never be negative I think you can drop the "& 0xffff".

DonalEvans (approved):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/KeyHashUtil.java
```diff
@@ -0,0 +1,51 @@
...
+package org.apache.geode.redis.internal.data;
+
+import static org.apache.geode.redis.internal.RegionProvider.REDIS_SLOTS;
+
+import org.apache.geode.redis.internal.executor.cluster.CRC16;
+
+/**
+ * Utility for parsing the hashtags in a redis key and computing the redis slot of the key.
+ */
+public class KeyHashUtil {
```
kirklund, line 24: I think a unit test for `KeyHashUtil` would be valuable and pretty easy to write.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/KeyHashUtil.java
upthewaterspout: Oooh, good catch! BTW - I didn't write this hashing logic, I just moved it to a new class. But I added a test and it looks like native redis handles this differently!

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/parameters/ClusterParameterRequirements.java
upthewaterspout: I've added some more tests to the `AbstractClusterIntegrationTest`. I changed this `ClusterParameterRequirements` class because the command `CLUSTER KEYSLOT XXX` has three arguments. I added additional validation into the `ClusterExecutor`

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisKey.java
upthewaterspout: done

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/KeyHashUtil.java
upthewaterspout: done

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/KeyHashUtil.java
upthewaterspout: I fixed this. I changed the hash implementation to be closer to native redis.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/RegionProvider.java
upthewaterspout: done

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisKey.java
upthewaterspout: done

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisKey.java
upthewaterspout: done

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/KeyHashUtil.java
upthewaterspout: I added a new test. This code was previously covered by the RedisKey unit test. I moved that testing into KeyHashUtilTest and added more coverage.

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/data/RedisKeyJUnitTest.java
upthewaterspout: done

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/KeyHashUtil.java
upthewaterspout: I put a couple of constants in StringBytesGlossary.

However, I don't think it is necessary for that class to hard code the decimal value of these chars ( eg 56, 48, etc.). `char` literals in java are always UTF-16. So converting them to a byte will always have the same value. That is not true for String.getBytes(), which uses a OS platform specific encoding.

So what I changed is that the constants are in still StringBytesGlossary but I changed that class to use char literals, which I think makes it easier to read and maintain.

DonalEvans: Ah, thanks for the explanation. Yeah, that definitely makes the class more readable. If it's not too much trouble, there are some other single-char constants in that file starting on line 243 that could also be simplified in this way.

upthewaterspout: Done!

upthewaterspout:
@ezoerner - added the randomized integration test.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/netty/StringBytesGlossary.java
```diff
@@ -32,27 +32,37 @@
   /**
    * byte identifier of a bulk string
    */
-  public static final byte BULK_STRING_ID = 36; // '$'
+  public static final byte BULK_STRING_ID = '$';
```
dschneider-pivotal, line 35: Is a java character literal's ordinal value stable? Or can it change based on the locale? If you know the answer can you send a link?

dschneider-pivotal (approved):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/netty/StringBytesGlossary.java
upthewaterspout: The only link I have is the java language spec - https://docs.oracle.com/javase/specs/jls/se16/html/jls-3.html#jls-3.10.4. The character literals are UTF-16.

PR #6858, opened 2021-09-10
GEODE-9365: Removing unnecessary synchronization

mhansonp:
Plus some cleanup

### DIFF
--- a/geode-core/src/integrationTest/java/org/apache/geode/internal/cache/ha/BlockingHARegionJUnitTest.java
+++ b/geode-core/src/integrationTest/java/org/apache/geode/internal/cache/ha/BlockingHARegionJUnitTest.java
@@ -20,12 +20,16 @@ import static org.apache.geode.internal.cache.ha.HARegionQueue.BLOCKING_HA_QUEUE
 import static org.apache.geode.internal.cache.ha.HARegionQueue.getHARegionQueueInstance;
 import static org.apache.geode.internal.statistics.StatisticsClockFactory.disabledClock;
 import static org.apache.geode.test.dunit.ThreadUtils.join;
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Properties;
 
+import org.apache.logging.log4j.Logger;
 import org.junit.Before;
 import org.junit.Ignore;
 import org.junit.Test;
@@ -35,6 +39,7 @@ import org.apache.geode.cache.CacheFactory;
 import org.apache.geode.distributed.DistributedSystem;
 import org.apache.geode.internal.cache.EventID;
 import org.apache.geode.internal.cache.InternalCache;
+import org.apache.geode.logging.internal.log4j.api.LogService;
 import org.apache.geode.test.awaitility.GeodeAwaitility;
 import org.apache.geode.test.dunit.ThreadUtils;
 import org.apache.geode.test.dunit.WaitCriterion;
@@ -111,49 +116,17 @@ public class BlockingHARegionJUnitTest {
     hrq.setPrimary(true);// fix for 40314 - capacity constraint is checked for primary only.
     final Thread thread1 = new DoPuts(hrq, 2);
     thread1.start();
-    WaitCriterion ev = new WaitCriterion() {
-      @Override
-      public boolean done() {
-        return hrq.region.size() == 2;
-      }
-
-      @Override
-      public String description() {
-        return null;
-      }
-    };
-    GeodeAwaitility.await().untilAsserted(ev);
+    GeodeAwaitility.await().until(() -> hrq.region.size() == 2);
     assertTrue(thread1.isAlive()); // thread should still be alive (in wait state)
 
     Thread thread2 = new DoTake(hrq, 1);
     thread2.start(); // start take thread
-    ev = new WaitCriterion() {
-      @Override
-      public boolean done() {
-        return hrq.region.size() == 3;
-      }
 
-      @Override
-      public String description() {
-        return null;
-      }
-    };
     // sleep. take will proceed and so will sleeping put
-    GeodeAwaitility.await().untilAsserted(ev);
+    GeodeAwaitility.await().until(() -> hrq.region.size() == 3);
 
     // thread should have died since put should have proceeded
-    ev = new WaitCriterion() {
-      @Override
-      public boolean done() {
-        return !thread1.isAlive();
-      }
-
-      @Override
-      public String description() {
-        return "thread1 still alive";
-      }
-    };
-    GeodeAwaitility.await().untilAsserted(ev);
+    GeodeAwaitility.await().alias("thread1 still alive").until(() -> !thread1.isAlive());
 
     join(thread1, 30 * 1000); // for completeness
     join(thread2, 30 * 1000);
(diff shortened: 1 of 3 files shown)


geode-core/src/main/java/org/apache/geode/internal/cache/ha/HARegionQueue.java
```diff
@@ -2292,66 +2270,79 @@ public void destroy() throws CacheWriterException {
...
       }
+
+      if (Thread.interrupted()) {
+        throw new InterruptedException();
+      }
+
+      long startTime = System.currentTimeMillis();
+      synchronized (this.permitMon) {
+
+        if (putPermits.decrementAndGet() >= 0) {
```
jchen21, line 2285: getAndDecrement()?

mhansonp: Nope that is correct.

geode-core/src/integrationTest/java/org/apache/geode/internal/cache/ha/BlockingHARegionJUnitTest.java
```diff
@@ -281,26 +231,15 @@ public void testConcurrentPutsTakesNotExceedingLimit() throws Exception {
...
-      }
+    GeodeAwaitility.await().until(() -> hrq.region.size() == 20000);
 
-      @Override
-      public String description() {
-        return null;
-      }
-    };
-    GeodeAwaitility.await().untilAsserted(ev);
+    assertThat(thread1.isAlive()).isTrue();
```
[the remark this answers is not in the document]
mhansonp: The test is disabled. I chose not to modify it.

geode-core/src/main/java/org/apache/geode/internal/cache/ha/HARegionQueue.java
```diff
@@ -2292,66 +2270,79 @@ public void destroy() throws CacheWriterException {
      * in the HARegionQueue.
      */
     @Override
-    @edu.umd.cs.findbugs.annotations.SuppressWarnings("TLW_TWO_LOCK_WAIT")
```
mhansonp, line 2295: This is the main body of the change.

geode-core/src/main/java/org/apache/geode/internal/cache/ha/HARegionQueue.java
```diff
@@ -2292,66 +2270,79 @@ public void destroy() throws CacheWriterException {
...
 
+    /* Do not call this method directly from anywhere except checkQueueSizeConstraint */
+
     /**
      * This function should always be called under a lock on putGuard & permitMon obejct
      *
      * @return int current Put permits
      */
     private int reconcilePutPermits() {
-      putPermits += takeSidePutPermits;
```
mhansonp, line 2352: additional pieces required.

geode-core/src/main/java/org/apache/geode/internal/cache/ha/HARegionQueue.java
```diff
@@ -2211,22 +2194,16 @@ public Object updateHAEventWrapper(InternalDistributedMember sender,
...
-     */
-    private final Object putGuard = new Object();
-
-    private final int capacity;
-
     /**
      * Current put permits available
      */
-    private int putPermits;
+    private final AtomicInteger putPermits = new AtomicInteger(0);
```
mhansonp, line 2200: Variable changes that were required.

kirklund (changes requested):
I think you should retrofit the test to use `ExecutorServiceRule` (geode-junit) and `ErrorCollector` (JUnit) instead.

Change all the catch-blocks in threads to hand off the error to the `ErrorCollector` instead of appending to a `StringBuffer`:
```
@Rule
public ErrorCollector errorCollector = new ErrorCollector();
```
```
} catch (Exception e) {
  errorCollector.addError(e);
}
```
And replace all the `Thread` uses with simple uses of `ExecutorServiceRule`.

mhansonp:
> I think you should retrofit the test to use `ExecutorServiceRule` (geode-junit) and `ErrorCollector` (JUnit) instead.
> 
> Change all the catch-blocks in threads to hand off the error to the `ErrorCollector` instead of appending to a `StringBuffer`:
> 
> ```
> @Rule
> public ErrorCollector errorCollector = new ErrorCollector();
> ```
> 
>  ```
> } catch (Exception e) {
>   errorCollector.addError(e);
> }
> ```
> 
> And replace all the `Thread` uses with simple uses of `ExecutorServiceRule`.
> 
> Submitting tasks to `ExecutorServiceRule` will return futures the test can wait on (with built in use of GeodeAwaitility timeout) so you can ditch the waiting, any looping recommended by others, etc.

I could use some guidance on this. I will talk to you directly.

kirklund (approved):
Test needs more love at a later date.

geode-core/src/main/java/org/apache/geode/internal/cache/ha/HARegionQueue.java
```diff
@@ -2292,66 +2270,79 @@ public void destroy() throws CacheWriterException {
...
+      long startTime = System.currentTimeMillis();
+      synchronized (this.permitMon) {
+
+        if (putPermits.decrementAndGet() >= 0) {
+          return;
+        }
+
+        long duration = (System.currentTimeMillis() - startTime);
+
+        if (reconcilePutPermits() >= 0) {
```
jchen21, line 2291: I think this should be `if (reconcilePutPermits() > 0) {` without the `=`.  Because the original code that you deleted is `if (reconcilePutPermits() <= 0) {`.

dschneider-pivotal: I think the new code is different here because this thread already decremented putPermits and it went negative. In the old code we waited until putPermits was > 0 BEFORE we decremented it. But now we decrement it even if it is already < 0. 
reconcilePutPermits  updates and returns putPermits so if we see it go positive (>= 0) then this thread now has a permit to add to the queue.

dschneider-pivotal (changes requested):

mhansonp:
Updated with a suggested performance change.

jchen21 (approved):

DonalEvans (approved):
Approving the changes in `HARegion`, which is the only file in the PR for which I'm codeowner.

dschneider-pivotal (approved):

geode-core/src/main/java/org/apache/geode/internal/cache/ha/HARegionQueue.java
```diff
@@ -2292,66 +2270,79 @@ public void destroy() throws CacheWriterException {
...
+    void waitForPermission() throws InterruptedException {
+      if (!(this.haContainer instanceof HAContainerMap && isPrimary())) {
+        return;
       }
+
+      if (Thread.interrupted()) {
+        throw new InterruptedException();
+      }
+
+      if (putPermits.decrementAndGet() >= 0) {
```
nabarunnag, line 2282: Was wondering on the behavior change for value 0, previously if the value was zero, we allowed the code block to execute. Is it ok in the new code not to?

geode-core/src/main/java/org/apache/geode/internal/cache/ha/HARegionQueue.java
nabarunnag: just trying to understand the change, previously we alway decremented the putPermits, now we don't allow the block to be executed if the putPermits is zero after decrement but we allowed it earlier.

nabarunnag:
Is there a behavior change in this code ? just wondering if there is a need for a test to be written?

nabarunnag:
also please do write a better commit message during the final commit explaining what has changed : finals added , refactored generics, cleanups of dead comments, new atomic counters.

mhansonp:
> also please do write a better commit message during the final commit explaining what has changed : finals added , refactored generics, cleanups of dead comments, new atomic counters.

I kind of pointed that stuff out in the comments I added when I created this PR, but point taken. I will do a better job of that in the future.

boglesby (approved):

agingade:
It is about the number of items to keep in the HARegion and then throttle; which can go beyond set limit. 
Can we just replace the whole code using capacity and region’s size? and keep it simple...
sync(this) {
  if (capacity - region.size()) > 0) {
    wait(x time); // Notify when take is done while still waiting...
   }
}

PR #6860, opened 2021-09-10
GEODE-9521: add more multi-server test scenario

jinmeiliao:
* remove the static methods in ExpirableSecurityManager

### DIFF
--- a/geode-junit/src/main/java/org/apache/geode/security/ExpirableSecurityManager.java
+++ b/geode-junit/src/main/java/org/apache/geode/security/ExpirableSecurityManager.java
@@ -15,6 +15,7 @@
 
 package org.apache.geode.security;
 
+import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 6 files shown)


jmelchio (approved):

mhansonp (approved):

geode-junit/src/main/java/org/apache/geode/security/ExpirableSecurityManager.java
```diff
@@ -31,44 +32,44 @@
  *
  * make sure to call reset after each test to clean things up.
  */
-public class ExpirableSecurityManager extends SimpleSecurityManager {
+public class ExpirableSecurityManager extends SimpleSecurityManager implements Serializable {
```
kirklund, line 35: This is `Serializable only` because of lambda style VM invocations, right?

Since this is in the product code and doesn't really need to go on the wire, you should instead subclass `ExpirableSecurityManager` as an inner-class within the dunit test and then use that subclass in the tests.
```
private static class SerializableExpirableSecurityManager 
    extends ExpirableSecurityManager
    implements Serializable {
...
}
```

kirklund (changes requested):

geode-junit/src/main/java/org/apache/geode/security/ExpirableSecurityManager.java
jinmeiliao: No, this is not production code. `ExpirableSecurityManager` is a test security manager in `geode-junit` module

jdeppe-pivotal (approved):

geode-junit/src/main/java/org/apache/geode/security/ExpirableSecurityManager.java
kirklund: Oh yeah I see `geode-junit` now. Thanks!

kirklund (approved):

PR #6861, opened 2021-09-13
GEODE-9516: add ZINTERSTORE command

nonbinaryprogrammer:
ZINTERSTORE command takes the intersection of N sorted sets and stores
the results, either summing, taking the max, or taking the min of all
the scores for each member. The provided destination key will be
overridden if it already exists.

### For all changes:

### DIFF
--- a/geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZInterStoreIntegrationTest.java
+++ b/geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZInterStoreIntegrationTest.java
@@ -702,15 +702,48 @@ public abstract class AbstractZInterStoreIntegrationTest implements RedisIntegra
     return tuples;
   }
 
+  @Test
+  public void test_assertThatActualScoresAreVeryCloseToExpectedScores() {
+    Set<Tuple> actualResult = new HashSet<>(3);
+    Set<Tuple> expectedResult = new HashSet<>(2);
+
+    actualResult.add(new Tuple("element1", 1.0));
+    expectedResult.add(new Tuple("element1", 1.0));
+
+    actualResult.add(new Tuple("element2", 2.0));
+    expectedResult.add(new Tuple("element2", 2.0));
+
+    actualResult.add(new Tuple("element3", 3.0));
+
+    // actual has more elements than expected
+    assertThatThrownBy(
+        () -> assertThatActualScoresAreVeryCloseToExpectedScores(expectedResult, actualResult))
+            .isInstanceOf(AssertionError.class);
+
+    expectedResult.add(new Tuple("element3", 3.0));
+    expectedResult.add(new Tuple("element4", 4.0));
+
+    // actual has fewer elements than expected
+    assertThatThrownBy(
+        () -> assertThatActualScoresAreVeryCloseToExpectedScores(expectedResult, actualResult))
+            .isInstanceOf(AssertionError.class);
+  }
+
   private void assertThatActualScoresAreVeryCloseToExpectedScores(
       Set<Tuple> expectedResults, Set<Tuple> results) {
+    assertThat(expectedResults.size()).isEqualTo(results.size());
+
     for (Tuple expectedResult : expectedResults) {
+      boolean resultFound = false;
       for (Tuple actualResult : results) {
         if (Objects.equals(actualResult.getElement(), expectedResult.getElement())) {
           assertThat(actualResult.getScore()).isCloseTo(expectedResult.getScore(),
               Offset.offset(0.0001D));
+          resultFound = true;
+          break;
         }
       }
+      assertThat(resultFound).isTrue();
     }
   }
 }
--- a/geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZInterStoreIntegrationTest.java
+++ b/geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZInterStoreIntegrationTest.java
@@ -575,45 +575,6 @@ public abstract class AbstractZInterStoreIntegrationTest implements RedisIntegra
   }
 
   @Test
-  public void ensureSetConsistency_whenRunningConcurrently() {
-    int scoreCount = 1000;
-    Map<String, Double> scores1 = buildMapOfMembersAndScores(1, 15);
-    Map<String, Double> scores2 = buildMapOfMembersAndScores(3, 18);
-    Map<String, Double> scores3 = buildMapOfMembersAndScores(4, 14);
-
-    jedis.zadd(KEY1, scores1);
-    jedis.zadd(KEY2, scores2);
-    jedis.zadd(KEY3, scores3);
-
-    Set<Tuple> expectedMinScores = new HashSet<>();
-    Set<Tuple> expectedMaxScores = new HashSet<>();
-    for (int i = 4; i <= 14; i++) {
-      expectedMinScores.add(tupleMinOfScores("player" + i, scores1, scores2, scores3));
-      expectedMaxScores.add(tupleMaxOfScores("player" + i, scores1, scores2, scores3));
-    }
-
-    // new ConcurrentLoopingThreads(1000,
-    // i -> jedis.zadd(KEY1, (double) i, "member-" + i),
-    // i -> jedis.zadd(KEY2, (double) i, "member-" + i),
-    // i -> jedis.zadd(KEY3, (double) i, "member-" + i),
-    // i -> jedis.zinterstore("maxSet", new ZParams().aggregate(ZParams.Aggregate.MAX),
-    jedis.zinterstore("maxSet", new ZParams().aggregate(ZParams.Aggregate.MAX),
-        KEY1, KEY2, KEY3);
-    // This ensures that the lock ordering for keys is working
-    // i -> jedis.zinterstore("minSet", new ZParams().aggregate(ZParams.Aggregate.MIN),
-    jedis.zinterstore("minSet", new ZParams().aggregate(ZParams.Aggregate.MIN),
-        KEY1, KEY2, KEY3);// )
-    // .runWithAction(() -> {
-    Set<Tuple> maxSet = jedis.zrangeWithScores("maxSet", 0, scoreCount);
-    Set<Tuple> minSet = jedis.zrangeWithScores("minSet", 0, scoreCount);
-    assertThat(maxSet).hasSize(scoreCount);
-    assertThat(minSet).hasSize(scoreCount);
-    assertThatActualScoresAreVeryCloseToExpectedScores(expectedMaxScores, maxSet);
-    assertThatActualScoresAreVeryCloseToExpectedScores(expectedMinScores, minSet);
-    // });
-  }
-
-  @Test
   public void ensureSetConsistency_andNoExceptions_whenRunningConcurrently() {
     int scoreCount = 1000;
     jedis.zadd("{A}ones", buildMapOfMembersAndScores(0, scoreCount - 1));
--- a/geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZInterStoreIntegrationTest.java
+++ b/geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZInterStoreIntegrationTest.java
@@ -175,6 +175,14 @@ public abstract class AbstractZInterStoreIntegrationTest implements RedisIntegra
   }
 
   @Test
+  public void shouldError_givenWeightsFollowedByCorrectNumberOfArgumentsIncludingAggregate() {
+    assertThatThrownBy(
+        () -> jedis.sendCommand(NEW_SET, Protocol.Command.ZINTERSTORE, NEW_SET, "3",
+            KEY1, KEY2, KEY3, "WEIGHTS", "1", "AGGREGATE", "SUM"))
+                .hasMessage("ERR " + RedisConstants.ERROR_WEIGHT_NOT_A_FLOAT);
+  }
+
+  @Test
   public void shouldStoreIntersection_givenWeightOfOne_andOneRedisSortedSet() {
     Map<String, Double> scores = buildMapOfMembersAndScores();
     Set<Tuple> expectedResults = convertToTuples(scores, (ignore, value) -> value);
@@ -592,6 +600,16 @@ public abstract class AbstractZInterStoreIntegrationTest implements RedisIntegra
     Set<Tuple> results = jedis.zrangeWithScores(KEY1, 0, 10);
 
     assertThatActualScoresAreVeryCloseToExpectedScores(expectedResults, results);
+  }
+
+  @Test
+  public void shouldNotError_givenKeysNamedWeightsOrAggregate() {
+    String weights = "WEIGHTS";
+    jedis.zadd(weights, 1, "member");
+    String aggregate = "AGGREGATE";
+    jedis.zadd(aggregate, 1, "member");
+    jedis.zinterstore(weights, weights, weights);
+    jedis.zinterstore(aggregate, aggregate, aggregate);
   }
 
   @Test
--- a/geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZInterStoreIntegrationTest.java
+++ b/geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZInterStoreIntegrationTest.java
@@ -183,6 +183,13 @@ public abstract class AbstractZInterStoreIntegrationTest implements RedisIntegra
   }
 
   @Test
+  public void shouldError_givenNumKeysNotAnInteger() {
+    assertThatThrownBy(
+        () -> jedis.sendCommand(NEW_SET, Protocol.Command.ZINTERSTORE, NEW_SET, "fish", KEY1, KEY2))
+            .hasMessage("ERR " + RedisConstants.ERROR_NOT_INTEGER);
+  }
+
+  @Test
   public void shouldStoreIntersection_givenWeightOfOne_andOneRedisSortedSet() {
     Map<String, Double> scores = buildMapOfMembersAndScores();
     Set<Tuple> expectedResults = convertToTuples(scores, (ignore, value) -> value);
--- a/geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZInterStoreIntegrationTest.java
+++ b/geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZInterStoreIntegrationTest.java
@@ -115,7 +115,7 @@ public abstract class AbstractZInterStoreIntegrationTest implements RedisIntegra
   }
 
   @Test
-  public void shouldError_givenWeightNotANumber() {
+  public void shouldError_givenWeightNotAFloat() {
     assertThatThrownBy(
         () -> jedis.sendCommand(NEW_SET, Protocol.Command.ZINTERSTORE, NEW_SET, "1",
             KEY1, "WEIGHTS", "not-a-number"))
@@ -502,7 +502,7 @@ public abstract class AbstractZInterStoreIntegrationTest implements RedisIntegra
 
     Set<Tuple> expected = new HashSet<>();
     for (int i = 6; i <= 10; i++) {
-      expected.add(tupleMaxOfScores("player" + i, scores1, scores2, scores3));
+      expected.add(new Tuple("player" + i, score));
     }
 
     Set<Tuple> actual = jedis.zrangeWithScores(NEW_SET, 0, 10);
--- a/geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZInterStoreIntegrationTest.java
+++ b/geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZInterStoreIntegrationTest.java
@@ -101,6 +101,20 @@ public abstract class AbstractZInterStoreIntegrationTest implements RedisIntegra
   }
 
   @Test
+  public void shouldError_givenNumKeysOfZero() {
+    assertThatThrownBy(
+        () -> jedis.sendCommand(NEW_SET, Protocol.Command.ZINTERSTORE, NEW_SET, "0", KEY1, KEY2))
+            .hasMessage("ERR " + RedisConstants.ERROR_KEY_REQUIRED);
+  }
+
+  @Test
+  public void shouldError_givenNegativeNumKeys() {
+    assertThatThrownBy(
+        () -> jedis.sendCommand(NEW_SET, Protocol.Command.ZINTERSTORE, NEW_SET, "-2", KEY1, KEY2))
+            .hasMessage("ERR " + RedisConstants.ERROR_KEY_REQUIRED);
+  }
+
+  @Test
   public void shouldError_givenTooManyWeights() {
     assertThatThrownBy(
         () -> jedis.sendCommand(NEW_SET, Protocol.Command.ZINTERSTORE, NEW_SET, "1",
@@ -302,16 +316,13 @@ public abstract class AbstractZInterStoreIntegrationTest implements RedisIntegra
   }
 
   @Test
-  public void shouldStoreIntersection_givenOneSetDoesNotExist() {
+  public void shouldStoreEmptyIntersection_givenOneSetDoesNotExist() {
     Map<String, Double> scores = buildMapOfMembersAndScores(1, 10);
-    Set<Tuple> expectedResults = convertToTuples(scores, (i, x) -> x);
     jedis.zadd(KEY1, scores);
 
-    assertThat(jedis.zinterstore(NEW_SET, KEY1, KEY2)).isEqualTo(scores.size());
+    assertThat(jedis.zinterstore(NEW_SET, KEY1, KEY2)).isEqualTo(0);
 
-    Set<Tuple> results = jedis.zrangeWithScores(NEW_SET, 0, 10);
-
-    assertThatActualScoresAreVeryCloseToExpectedScores(expectedResults, results);
+    assertThat(jedis.zrangeWithScores(NEW_SET, 0, 10)).isEmpty();
   }
 
   @Test
@@ -347,6 +358,15 @@ public abstract class AbstractZInterStoreIntegrationTest implements RedisIntegra
     Set<Tuple> actual = jedis.zrangeWithScores(NEW_SET, 0, 10);
 
     assertThatActualScoresAreVeryCloseToExpectedScores(expected, actual);
+  }
+
+  @Test
+  public void shouldStoreSumOfZero_whenAddingNegativeAndPositiveInfinity() {
+    String member = "member";
+    jedis.zadd(KEY1, Double.POSITIVE_INFINITY, member);
+    jedis.zadd(KEY2, Double.NEGATIVE_INFINITY, member);
+    assertThat(jedis.zinterstore(NEW_SET, KEY1, KEY2)).isOne();
+    assertThat(jedis.zscore(NEW_SET, member)).isZero();
   }
 
   @Test
--- a/geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZInterStoreIntegrationTest.java
+++ b/geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZInterStoreIntegrationTest.java
@@ -1,0 +1,735 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.apache.geode.redis.internal.executor.sortedset;
+
+import static java.lang.Math.max;
+import static java.lang.Math.min;
+import static org.apache.geode.redis.RedisCommandArgumentsTestHelper.assertAtLeastNArgs;
+import static org.apache.geode.test.dunit.rules.RedisClusterStartupRule.BIND_ADDRESS;
+import static org.apache.geode.test.dunit.rules.RedisClusterStartupRule.REDIS_CLIENT_TIMEOUT;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+import java.util.HashSet;
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Random;
+import java.util.Set;
+import java.util.function.BiFunction;
+
+import org.assertj.core.data.Offset;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import redis.clients.jedis.HostAndPort;
+import redis.clients.jedis.JedisCluster;
+import redis.clients.jedis.Protocol;
+import redis.clients.jedis.Tuple;
+import redis.clients.jedis.ZParams;
+
+import org.apache.geode.redis.ConcurrentLoopingThreads;
+import org.apache.geode.redis.RedisIntegrationTest;
+import org.apache.geode.redis.internal.RedisConstants;
+
+public abstract class AbstractZInterStoreIntegrationTest implements RedisIntegrationTest {
+
+  private static final String NEW_SET = "{user1}new";
+  private static final String KEY1 = "{user1}sset1";
+  private static final String KEY2 = "{user1}sset2";
+  private static final String KEY3 = "{user1}sset3";
+
+  private JedisCluster jedis;
+
+  @Before
+  public void setUp() {
+    jedis = new JedisCluster(new HostAndPort(BIND_ADDRESS, getPort()), REDIS_CLIENT_TIMEOUT);
+  }
+
+  @After
+  public void tearDown() {
+    flushAll();
+    jedis.close();
+  }
+
+  @Test
+  public void shouldError_givenTooFewArguments() {
+    assertAtLeastNArgs(jedis, Protocol.Command.ZINTERSTORE, 3);
+  }
+
+  @Test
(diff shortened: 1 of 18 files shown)


geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -578,6 +625,91 @@ private int getMaxElementsToReturn(AbstractSortedSetRangeOptions<?> rangeOptions
     return result;
   }
 
+  private RedisSortedSet getIntersection(List<RedisSortedSet> sets, ZAggregator aggregator) {
+    RedisSortedSet retVal = new RedisSortedSet(Collections.emptyList());
+
+    for (RedisSortedSet set : sets) {
+      for (OrderedSetEntry entry : set.members.values()) {
```
dschneider-pivotal, line 632: Since all you care about seeing in this for loop is each member in "members" I think you can just do this "for (byte[] member: set.members.keySet())".
Or you could use "set.members.fastForEachKey(key->{...});"

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -578,6 +625,91 @@ private int getMaxElementsToReturn(AbstractSortedSetRangeOptions<?> rangeOptions
...
+      Double runningTotal) {
+    if (sets.isEmpty()) {
+      return runningTotal;
+    }
+
+    if (runningTotal.isNaN()) {
+      return runningTotal;
+    }
+
+    if (sets.get(0).members.containsKey(member)) {
```
dschneider-pivotal, line 663: instead of doing containsKey followed by get, just do get and check for it returning null. This does one map lookup instead of two

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -578,6 +625,91 @@ private int getMaxElementsToReturn(AbstractSortedSetRangeOptions<?> rangeOptions
...
+          newScore = recursivelyGetMaxScoreForMember(sets, entry.member, Double.MIN_VALUE);
+        } else {
+          newScore = recursivelyGetMinScoreForMember(sets, entry.member, Double.MAX_VALUE);
+        }
+
+        if (newScore != null) {
+          if (newScore.isNaN()) {
+            throw new ArithmeticException(ERROR_OPERATION_PRODUCED_NAN);
+          }
+          retVal.memberAdd(entry.getMember(), Coder.doubleToBytes(newScore));
```
dschneider-pivotal, line 646: It seems like if you passed retVal in to recursively*ScoreForMember then it could do the retVal.memberAdd if needed. That would also allow us to get rid of this use of "Double". If you get rid of the recursion then these methods could just take a member, input sets, and a set to fill.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -578,6 +625,91 @@ private int getMaxElementsToReturn(AbstractSortedSetRangeOptions<?> rangeOptions
...
+        }
+      }
+    }
+    return retVal;
+  }
+
+  private Double recursivelySumScoresForMember(List<RedisSortedSet> sets, byte[] member,
```
dschneider-pivotal, line 653: I think the recursion in all three of these methods should be replaced with a simple iteration of sets checking each one for member and computing the sum/min/max. If it finds member in each set then it can store that value on the retVal set.
Recursion is usually slower, harder for some to understand, and could have a stack overflow.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -661,7 +793,7 @@ public static int javaImplementationOfAnsiCMemCmp(byte[] array1, byte[] array2)
       Sizeable {
     byte[] member;
     byte[] scoreBytes;
-    double score;
+    Double score;
```
dschneider-pivotal, line 796: Why does this need to be an Object (Double)? The old "double" saves memory.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -578,6 +625,91 @@ private int getMaxElementsToReturn(AbstractSortedSetRangeOptions<?> rangeOptions
...
+    if (sets.isEmpty()) {
+      return runningTotal;
+    }
+
+    if (runningTotal.isNaN()) {
+      return runningTotal;
+    }
+
+    if (sets.get(0).members.containsKey(member)) {
+      Double score = sets.get(0).members.get(member).score;
```
dschneider-pivotal, line 664: use "double" here

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/ZInterStoreExecutor.java
```diff
@@ -0,0 +1,126 @@
...
+          } catch (NumberFormatException nex) {
+            return RedisResponse.error(ERROR_WEIGHT_NOT_A_FLOAT);
+          }
+        }
+        continue;
+      }
+
+      if (Arrays.equals(arg, bAGGREGATE)) {
+        try {
+          aggregator = ZAggregator.valueOf(Coder.bytesToString(argIterator.next()));
```
dschneider-pivotal, line 92: do you need a hasNext check here also?

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/ZInterStoreExecutor.java
```diff
@@ -0,0 +1,126 @@
...
+      return RedisResponse.error(ERROR_SYNTAX);
+    }
+
+    // Rough validation so that we can use numKeys to initialize the array sizes below.
+    if (numKeys > commandElements.size()) {
+      return RedisResponse.error(ERROR_SYNTAX);
+    }
+
+    List<RedisKey> sourceKeys = new ArrayList<>((int) numKeys);
+    List<Double> weights = new ArrayList<>((int) numKeys);
```
dschneider-pivotal, line 61: could weights be made a "double[]"?

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/ZInterStoreExecutor.java
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    int bucket = command.getKey().getBucketId();
+    for (RedisKey key : sourceKeys) {
+      if (key.getBucketId() != bucket) {
+        return RedisResponse.crossSlot(ERROR_WRONG_SLOT);
+      }
+    }
+
+    List<ZKeyWeight> keyWeights = new ArrayList<>();
```
dschneider-pivotal, line 114: this list can be presized.
How about just having this List and get rid of the sourceKeys list and and the weights list/array. Create the initial ZKeyWeight with just the key (its double can default to 1). Then when you parse the weights you can call setWeight on the correct ZKeyWeight.

dschneider-pivotal (changes requested):

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZInterStoreIntegrationTest.java
```diff
@@ -0,0 +1,696 @@
...
+    jedis.zadd(KEY2, scores2);
+    jedis.zadd(KEY3, scores3);
+
+    ZParams zParams = new ZParams().aggregate(ZParams.Aggregate.MAX);
+
+    assertThat(jedis.zinterstore(NEW_SET, zParams, KEY1, KEY2, KEY3)).isEqualTo(5);
+
+    Set<Tuple> expected = new HashSet<>();
+    for (int i = 6; i <= 10; i++) {
+      expected.add(tupleMaxOfScores("player" + i, scores1, scores2, scores3));
```
DonalEvans, line 505: This can be simplified to just
```
expected.add(new Tuple("player" + i, score));
```

DonalEvans (changes requested):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -578,6 +625,91 @@ private int getMaxElementsToReturn(AbstractSortedSetRangeOptions<?> rangeOptions
     return result;
   }
 
+  private RedisSortedSet getIntersection(List<RedisSortedSet> sets, ZAggregator aggregator) {
+    RedisSortedSet retVal = new RedisSortedSet(Collections.emptyList());
+
+    for (RedisSortedSet set : sets) {
```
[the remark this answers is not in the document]
nonbinaryprogrammer: oh, yeah that's much better. I meant to change that and forgot.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
nonbinaryprogrammer: I'm not entirely sure what you mean. it's a get followed by a contains, not the other way around

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/ZInterStoreExecutor.java
nonbinaryprogrammer: No, Iterator.next() throws a NoSubElementException when there isn't a next element, which we subsequently catch and send a syntax error as the response. Does that seem sufficient?

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/ZInterStoreExecutor.java
nonbinaryprogrammer: it cannot. the weights must be able to be null, and we need to be able to check if the list is empty, and since it's pre-sized the length was always `numKeys`

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -288,6 +289,54 @@ long zcount(SortedSetScoreRangeOptions rangeOptions) {
...
+          if (weight == 0) {
+            score = 0;
+          } else if (weight == 1) {
+            score = entry.getScore();
+          } else if (Double.isInfinite(weight) && entry.score == 0D) {
+            score = 0D;
+          } else {
+            double newScore = entry.score * weight;
+            if (Double.isNaN(newScore)) {
+              throw new ArithmeticException(ERROR_OPERATION_PRODUCED_NAN);
```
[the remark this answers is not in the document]
nonbinaryprogrammer: I get that native redis doesn't have the same behavior, but in Java positive infinity plus negative infinity results in NaN, not 0

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -578,6 +627,91 @@ private int getMaxElementsToReturn(AbstractSortedSetRangeOptions<?> rangeOptions
     return result;
   }
 
+  private RedisSortedSet getIntersection(List<RedisSortedSet> sets, ZAggregator aggregator) {
+    RedisSortedSet retVal = new RedisSortedSet(Collections.emptyList());
```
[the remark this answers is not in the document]
nonbinaryprogrammer: I made some changes to this area of the code, so please take a look and we can talk more about it, if you don't mind.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -607,6 +650,84 @@ private void addIfMatching(GlobPattern matchPattern, List<byte[]> resultList, by
...
+    }
+    return retVal;
+  }
+
+  private Double getSumOfScoresForMember(List<RedisSortedSet> sets, byte[] member,
+      RedisSortedSet retVal) {
+    double runningTotal = 0;
+    for (RedisSortedSet set : sets) {
+      if (set.members.containsKey(member)) {
+        runningTotal += set.members.get(member).score;
```
dschneider-pivotal, line 688: use getScore()
Also instead of calling both containsKey and get (which does two map lookups) just call get and if its result is not null compute the runningTotal.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -607,6 +650,84 @@ private void addIfMatching(GlobPattern matchPattern, List<byte[]> resultList, by
...
+        if (newScore.isNaN()) {
+          throw new ArithmeticException(ERROR_OPERATION_PRODUCED_NAN);
+        }
+        retVal.memberAdd(member, newScore);
+      }
+    }
+    return retVal;
+  }
+
+  private Double getSumOfScoresForMember(List<RedisSortedSet> sets, byte[] member,
```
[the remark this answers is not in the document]
dschneider-pivotal: Oh you needed to use Double so you could return null. I think with the idea that we are computing the intersection on "this" (i.e. the retVal parameter goes away) and since these three methods will become one (by passing the aggregator into it) instead of it returning a Double/double it can return void and when it would have returned a non-null total it can instead check it for NAN and then just store it in this (i.e. memberAdd(member, total)).

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSetCommandsFunctionExecutor.java
```diff
@@ -164,6 +174,17 @@ public long zrevrank(RedisKey key, byte[] member) {
   @Override
   public long zunionstore(RedisKey destinationKey, List<ZKeyWeight> keyWeights,
       ZAggregator aggregator) {
+    List<RedisKey> keysToLock = lockKeys(destinationKey, keyWeights);
+
+    return stripedExecute(destinationKey, keysToLock,
+        () -> new RedisSortedSet(Collections.emptyList(), new double[] {}).zunionstore(
```
dschneider-pivotal, line 180: If you add that new RedisSortedSet constructor that just takes a size then you should use it here with a size of 0.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/ZStoreExecutor.java
```diff
@@ -0,0 +1,129 @@
...
+      byte[] arg = argIterator.next();
+
+      if (keyWeights.size() < numKeys) {
+        keyWeights.add(new ZKeyWeight(new RedisKey(arg), 1D));
+        continue;
+      }
+
+      arg = toUpperCaseBytes(arg);
+      if (Arrays.equals(arg, bWEIGHTS)) {
+        if (!allWeightsAreOne(keyWeights)) {
```
dschneider-pivotal, line 73: This seems like overkill (allWeightsAreOne). When I first saw it I thought "why do all the weights have to be 1". 
 It only seems possible if they supply more than one  WEIGHTS list. For example "ZINTERSTORE dst 2 set1 set2 WEIGHTS 1 2 WEIGHTS 2 3". But isn't that illegal any way? If you are trying to make sure that WEIGHTS does not happen more than once wouldn't a simple boolean that it has been seen let you do that? AGGREGATE may have the same issue (does native redis let you repeat it?). Does native redis enforce the order (i.e. WEIGHTS first then AGGREGATE)? If so we should also.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/ZStoreExecutor.java
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    if (keyWeights.size() != numKeys) {
+      return RedisResponse.error(ERROR_SYNTAX);
+    }
+
+    int bucket = command.getKey().getBucketId();
+    for (ZKeyWeight keyWeight : keyWeights) {
+      if (keyWeight.getKey().getBucketId() != bucket) {
+        return RedisResponse.crossSlot(ERROR_WRONG_SLOT);
```
dschneider-pivotal, line 109: I saw another place in your code were you checked that all the sets are local. That probably is not needed since you check that they are all on the same bucket here.

dschneider-pivotal (changes requested):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -607,6 +650,84 @@ private void addIfMatching(GlobPattern matchPattern, List<byte[]> resultList, by
...
+    RedisSortedSet smallestSet = sets.get(0);
+
+    for (RedisSortedSet set : sets) {
+      if (set.getSortedSetSize() < smallestSet.getSortedSetSize()) {
+        smallestSet = set;
+      }
+    }
+
+    for (byte[] member : smallestSet.members.keySet()) {
+      Double newScore;
```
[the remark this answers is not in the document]
nonbinaryprogrammer: it needs to be nullable

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/ZStoreExecutor.java
```diff
@@ -0,0 +1,129 @@
...
+          return RedisResponse.error(ERROR_SYNTAX);
+        }
+        continue;
+      }
+
+      // End up here if we have more keys than weights
+      return RedisResponse.error(ERROR_SYNTAX);
+    }
+
+    if (keyWeights.size() != numKeys) {
```
DonalEvans, line 102: This check might be unnecessary if the check on line 56 is modified slightly to check if `numKeys` is greater than `commandElements.size() - 2` rather than being greater than `commandElements.size()`. What this line is really checking is that we parsed `numKeys` elements before running out of things to parse. By checking that (excluding the command name and the destination key, hence the -2) there are at minimum enough elements in `commandElements` for us to parse `numKeys` worth, we can know that if we get to this line, then `keyWeights.size()` is guaranteed to be equal to `numKeys`, because we only ever add to `keyWeights()` on line 67, and we stop adding there as soon as we've added `numKeys` elements.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/ZStoreExecutor.java
nonbinaryprogrammer: order is not enforced and multiple aggregates is valid and only the last one is applied. good call on the boolean

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/ZStoreExecutor.java
nonbinaryprogrammer: do you mean this? https://github.com/apache/geode/pull/6861/files#diff-390c7c2da5112f714107592eaa475143c72fd69d19b1d1d3d8bb2ac25ed0f3d5R190

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/ZStoreExecutor.java
```diff
@@ -0,0 +1,129 @@
...
+      return RedisResponse.error(ERROR_SYNTAX);
+    }
+
+    int bucket = command.getKey().getBucketId();
```
DonalEvans, line 106: This check should be on the slot ID of the key, not the bucket ID, since each bucket has multiple slots in it. As this is, you could provide keys with non-matching slots and the command would succeed if they happened to be in the same bucket, which would be a significant difference from native Redis' behaviour.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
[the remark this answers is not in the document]
DonalEvans: The test that I included here is still failing when run against Geode Redis and passing when run against native Redis. I think that what's needed is a check in the `getSumOfScoresForMember()` method to see if we're about to add positive infinity to negative infinity and if so, not return NaN, but return 0, so that we can match Redis' behaviour. We're already avoiding doing Java math if we know the result would be different from Redis (see the comments just above here about `INFINITY * 0`) so adding in this check isn't deviating from what we've already decided we want to do.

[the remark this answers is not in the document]
nonbinaryprogrammer: I can handle this case specifically, but I don't like it. with the special case of `-infinity + infinity = 0`, this sequence of events `-infinity + infinity + 3` is not equal to `-infinity + 3 + infinity`, which means that the result of our sum will be dependent on the ordering of the sorted sets we're looking through. But that's a problem that native redis has, so we must as well.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/ZStoreExecutor.java
nonbinaryprogrammer: without this check we accept fewer keys than we specified. i.e `zinterstore 3 set1 set2`. The test that fails when this check is removed is `AbstractZInterStoreIntegrationTest.shouldError_givenNumKeysTooLarge`. the test passes on native redis

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -288,6 +288,49 @@ long zcount(SortedSetScoreRangeOptions rangeOptions) {
...
+          regionProvider.getTypedRedisData(REDIS_SORTED_SET, keyWeight.getKey(), false);
+
+      if (set == NULL_REDIS_SORTED_SET) {
+        continue;
+      }
+
+      double weight = keyWeight.getWeight();
+      RedisSortedSet weightedSet = new RedisSortedSet(Collections.emptyList(), new double[] {});
+
+      for (AbstractOrderedSetEntry entry : set.members.values()) {
```
jdeppe-pivotal, line 305: Sorry for the late comments and also if this has already been discussed, but it seems like we're potentially doing a lot of work here and creating a lot of allocations for sets that are just going to be tossed away. Would it be possible to first determine whether a member is intersecting and then perform any aggregation/weighting? Essentially what is already happening in `getIntersection` but include the weighting there instead of this loop. I like how that method is picking the smallest set to iterate over as the base.

jdeppe-pivotal (changes requested):

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/sortedset/AbstractZInterStoreIntegrationTest.java
```diff
@@ -0,0 +1,696 @@
...
+    for (Map.Entry<String, Double> e : map.entrySet()) {
+      tuples.add(new Tuple(e.getKey().getBytes(), function.apply(x++, e.getValue())));
+    }
+
+    return tuples;
+  }
+
+  private void assertThatActualScoresAreVeryCloseToExpectedScores(
+      Set<Tuple> expectedResults, Set<Tuple> results) {
+    for (Tuple expectedResult : expectedResults) {
```
jdeppe-pivotal, line 687: I think this method should also be testing whether all members in `expectedResults` are also exactly what is in `results`. As it is now, passing in `[a, b, c]` and `[d, e, f]` as members would never cause the assertion to be checked.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/RegionProvider.java
```diff
@@ -345,6 +345,14 @@ public RedisKeyCommands getKeyCommands() {
     return (Set<DistributedMember>) (Set<?>) partitionedRegion.getRegionAdvisor().adviseDataStore();
   }
 
+  /**
+   * A means to consistently order multiple keys for locking to avoid typical deadlock situations.
+   * Note that the list of keys is sorted in place.
+   */
+  public void orderForLocking(List<RedisKey> keys) {
+    keys.sort(stripedCoordinator::compareStripes);
+  }
```
jdeppe-pivotal, line 354: You don't need to use this anymore since locking (and sorting) is automatically handled in the `LockingStripedCoordinator` now.

jdeppe-pivotal (changes requested):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -607,6 +669,87 @@ private void addIfMatching(GlobPattern matchPattern, List<byte[]> resultList, by
...
+    }
+
+    this.memberAdd(memberName, runningTotal);
+    return runningTotal;
+  }
+
+  private Double getMaxScoreForMember(List<RedisSortedSet> sets, byte[] member) {
+    double runningMax = Double.MIN_VALUE;
+    for (RedisSortedSet set : sets) {
+      if (set.members.containsKey(member)) {
```
dschneider-pivotal, line 724: You had a question about an old comment I made about containsKey. In this code instead of calling members.containsKey on line 724 followed by members.get on line 725 you could just call members.get and test for null. That way you do one map lookup instead of two.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/ZInterStoreExecutor.java
dschneider-pivotal: sounds good

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/ZStoreExecutor.java
dschneider-pivotal: I'm not sure but I also saw a comment from Jens that seemed related to this. He would be a good one to ask about this too

jdeppe-pivotal (approved):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -607,6 +669,87 @@ private void addIfMatching(GlobPattern matchPattern, List<byte[]> resultList, by
     }
   }
 
+  private void computeIntersection(List<RedisSortedSet> sets, ZAggregator aggregator) {
```
dschneider-pivotal, line 672: should this method be removed? I don't see any callers

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -607,6 +669,87 @@ private void addIfMatching(GlobPattern matchPattern, List<byte[]> resultList, by
...
+      if (aggregator.equals(ZAggregator.SUM)) {
+        newScore = getSumOfScoresForMember(sets, member);
+      } else if (aggregator.equals(ZAggregator.MAX)) {
+        newScore = getMaxScoreForMember(sets, member);
+      } else {
+        newScore = getMinScoreForMember(sets, member);
+      }
+
+      if (newScore != null) {
+        if (newScore.isNaN()) {
```
dschneider-pivotal, line 693: the new "zinterstore" impl does not do this NaN check. Did you decide it was not needed?

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
```diff
@@ -607,6 +669,87 @@ private void addIfMatching(GlobPattern matchPattern, List<byte[]> resultList, by
...
+          throw new ArithmeticException(ERROR_OPERATION_PRODUCED_NAN);
+        }
+        retVal.memberAdd(member, newScore);
+      }
+    }
+  }
+
+  private Double getSumOfScoresForMember(List<RedisSortedSet> sets, byte[] memberName) {
```
dschneider-pivotal, line 701: are these three methods also dead code now?

dschneider-pivotal (changes requested):
I think some dead code needs to be removed and I had one question about a NaN check.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSet.java
nonbinaryprogrammer: Yes, I believe it is now unnecessary because the ZAggregator function that is applied doesn't ever return null.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/ZStoreExecutor.java
```diff
@@ -0,0 +1,133 @@
...
+        // found WEIGHTS keyword; parse weights
+      } else if (Arrays.equals(toUpperCaseBytes(argument), bWEIGHTS)) {
+        for (int i = 0; i < numKeys; i++) {
+          if (!argIterator.hasNext()) {
+            return syntaxErrorResponse;
+          }
+          argument = argIterator.next();
+          if (Arrays.equals(toUpperCaseBytes(argument), bAGGREGATE)) {
+            return syntaxErrorResponse; // there must be # weights between 'WEIGHTS' & 'AGGREGATE'
+          }
```
DonalEvans, line 105: This behaviour doesn't match native Redis. The below test fails with Radish but passes with native Redis:
```
  @Test
  public void shouldReturnWeightNotAValidFloat_givenWeightsFollowedByCorrectNumberOfArgumentsIncludingAggregate() {
    assertThatThrownBy(
        () -> jedis.sendCommand(NEW_SET, Protocol.Command.ZINTERSTORE, NEW_SET, "3",
            KEY1, KEY2, KEY3, "WEIGHTS", "1", "AGGREGATE", "SUM"))
                .hasMessage("ERR " + RedisConstants.ERROR_WEIGHT_NOT_A_FLOAT);
  }
```
I think that this check is not needed, because we only care if there are enough weights specified (if there aren't, `argIterator.hasNext()` returns false and we return a syntax error) and that those weights are valid floats (if they're not, we get a `NumberFormatException` below and return `ERROR_WEIGHT_NOT_A_FLOAT`).

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/ZStoreExecutor.java
DonalEvans: With the current state of the code, when I comment out this check, `AbstractZInterStoreIntegrationTest.shouldError_givenNumKeysTooLarge()` passes. If we specify too many keys, we catch that on line 54 and return there.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/ZStoreExecutor.java
```diff
@@ -0,0 +1,133 @@
...
+    // get all the keys
+    for (int i = 0; i < numKeys; i++) {
+      if (!argIterator.hasNext()) {
+        return syntaxErrorResponse;
+      }
+      argument = argIterator.next();
+      if (Arrays.equals(toUpperCaseBytes(argument), bWEIGHTS)
+          || Arrays.equals(toUpperCaseBytes(argument), bAGGREGATE)) {
+        return syntaxErrorResponse;
+      }
```
DonalEvans, line 76: This behaviour is incorrect. "WEIGHTS" and "AGGREGATE" are both valid Redis key names. The below test fails with our implementation, but passes on native Redis:
```
@Test
  public void shouldNotReturnError_givenKeysNamedWeightsOrAggregate() {
    String weights = "WEIGHTS";
    jedis.zadd(weights, 1, "member");
    String aggregate = "AGGREGATE";
    jedis.zadd(aggregate, 1, "member");
    jedis.zinterstore(weights, weights, weights);
    jedis.zinterstore(aggregate, aggregate, aggregate);
  }
```

DonalEvans (changes requested):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSetCommandsFunctionExecutor.java
```diff
@@ -164,7 +174,7 @@ public long zrevrank(RedisKey key, byte[] member) {
   @Override
   public long zunionstore(RedisKey destinationKey, List<ZKeyWeight> keyWeights,
       ZAggregator aggregator) {
-    List<RedisKey> keysToLock = new ArrayList<>(keyWeights.size());
+    List<RedisKey> keysToLock = getKeysToLock(destinationKey, keyWeights);
```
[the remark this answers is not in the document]
nonbinaryprogrammer: thank you

dschneider-pivotal (approved):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/sortedset/ZStoreExecutor.java
nonbinaryprogrammer: good point, thank you.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisSortedSetCommandsFunctionExecutor.java
```diff
@@ -175,10 +174,7 @@ public long zrevrank(RedisKey key, byte[] member) {
   public long zunionstore(RedisKey destinationKey, List<ZKeyWeight> keyWeights,
       ZAggregator aggregator) {
     List<RedisKey> keysToLock = getKeysToLock(destinationKey, keyWeights);
-    for (ZKeyWeight kw : keyWeights) {
-      getRegionProvider().ensureKeyIsLocal(kw.getKey());
-      keysToLock.add(kw.getKey());
-    }
+
     getRegionProvider().ensureKeyIsLocal(destinationKey);
     keysToLock.add(destinationKey);
```
DonalEvans, line 179: These two lines are also duplicated in the `getKeysToLock()`  method.

DonalEvans (approved):

PR #6862, opened 2021-09-13
GEODE-8192: Make Radish MSET command atomic

jdeppe-pivotal:
- Replace SynchronizedStripedCoordinator with LockingStripedCoordinator
  which locks against a list of locks instead of recurisively locking.
- Perform the MSET operation within a transacation. This will avoid
  partial results in the case of a server shutdown.
- Add DUnit tests to repeatedly crash the primary server while
  performing concurrent MSETs.

### For all changes:


geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/services/LockingStripedCoordinator.java
```diff
@@ -0,0 +1,103 @@
...
+    for (int i = 0; i < concurrencyLevel; i++) {
+      locks[i] = new ReentrantLock();
+    }
+  }
+
+  @Override
+  public <T> T execute(RedisKey stripeId, Callable<T> callable) {
+    Lock lock = getLock(stripeId);
+    try {
+      lock.lock();
```
dschneider-pivotal, line 49: if lock throws an exception then this method's finally block will still call unlock. That doesn't seem correct.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/services/LockingStripedCoordinator.java
```diff
@@ -0,0 +1,103 @@
...
+      lock.unlock();
+    }
+  }
+
+  @Override
+  public <T> T execute(List<RedisKey> stripeIds, Callable<T> callable) {
+    Lock[] locks = getLocks(stripeIds);
+    try {
+      for (int i = 0; i < locks.length; i++) {
+        locks[i].lock();
```
dschneider-pivotal, line 65: once again if lock throws an exception we should not call unlock in finally.
Also we might have only gotten through part of this lock array when an exception is thrown and then we should only unlock the subset of locks we locked.

dschneider-pivotal (changes requested):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/RegionProvider.java
```diff
@@ -124,11 +125,13 @@ public RegionProvider(InternalCache cache, StripedCoordinator stripedCoordinator
...
     partitionedRegion = (PartitionedRegion) dataRegion;
 
-    stringCommands = new RedisStringCommandsFunctionExecutor(this);
-    setCommands = new RedisSetCommandsFunctionExecutor(this);
-    hashCommands = new RedisHashCommandsFunctionExecutor(this);
-    sortedSetCommands = new RedisSortedSetCommandsFunctionExecutor(this);
-    keyCommands = new RedisKeyCommandsFunctionExecutor(this);
+    CacheTransactionManager txManager = cache.getCacheTransactionManager();
+
+    stringCommands = new RedisStringCommandsFunctionExecutor(this, txManager);
```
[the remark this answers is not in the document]
jdeppe-pivotal: Yes, I like that idea!

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/services/LockingStripedCoordinator.java
jdeppe-pivotal: Yes.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/services/LockingStripedCoordinator.java
jdeppe-pivotal: I did wonder about that - since `ReentrantLock.lock()` isn't supposed to throw anything. But better to be safer.

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/string/MSetDUnitTest.java
```diff
@@ -0,0 +1,190 @@
...
+            .runWithAction(() -> {
+              int count = 0;
+              List<String> values = jedis.mget(keys);
+              for (String v : values) {
+                if (v == null) {
+                  continue;
+                }
+                count += v.startsWith("valueOne") ? 1 : -1;
+              }
+              assertThat(Math.abs(count)).isEqualTo(KEY_COUNT);
```
DonalEvans, line 114: A similar verification process to the one described in `AbstractMSetIntegrationTest` could be used here for added clarity.

DonalEvans (changes requested):
One test that needs to be updated, and some other optional suggestions.

DonalEvans (approved):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/services/LockingStripedCoordinator.java
dschneider-pivotal: I think you have a good point. No need to make this code more complicated than it needs to be. I think any exception thrown means we are in a lot of trouble (like out of memory or stack overflow, etc.)

dschneider-pivotal: But I think you should move the "lock()" calls out of the try/finally so that the finally block will only be called if we actually did the lock calls.

jdeppe-pivotal: Done

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/RegionProvider.java
```diff
@@ -172,6 +176,42 @@ public RedisStats getRedisStats() {
...
+  }
+
+  private <T> Callable<T> getTxWrappedCallable(Callable<T> callable) {
+    Callable<T> txWrappedCallable = () -> {
+      T result;
+      try {
+        txManager.begin();
+        result = callable.call();
+        txManager.commit();
+      } finally {
```
dschneider-pivotal, line 204: I think you want this finally block to complete the tx. So move the begin outside the try and then in the finally block you either commit or rollback.
It would look something like this:
boolean success = false;
txManager.begin();
try {
  result = callable.call();
  success = true;
} finally {
  if (success) {
    txManager.commit();
  } else {
    txManager.rollback();
  }
}

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisStringCommandsFunctionExecutor.java
```diff
@@ -144,6 +143,26 @@ public int setbit(RedisKey key, long offset, int value) {
...
+    }
+
+    // Pass a key in so that the bucket will be locked. Since all keys are already guaranteed to be
+    // in the same bucket we can use any key for this.
+    return stripedExecuteInTransaction(keysToLock.get(0), keysToLock, () -> mset0(keys, values));
+  }
+
+  private Void mset0(List<RedisKey> keys, List<byte[]> values) {
+    for (int i = 0; i < keys.size(); i++) {
+      setRedisString(getRegionProvider(), keys.get(i), values.get(i));
```
dschneider-pivotal, line 161: do you need to also clear any pending expiration in the case of mset being called on an existing key that has expiration scheduled? I think the easiest way to do this is call the "set" method with SetOptions param set to "null".

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/services/LockingStripedCoordinator.java
```diff
@@ -0,0 +1,103 @@
...
+      locks[i].lock();
+    }
+    try {
+      return callable.call();
+    } catch (RuntimeException re) {
+      throw re;
+    } catch (Exception e) {
+      throw new RuntimeException(e);
+    } finally {
+      for (int i = locks.length - 1; i >= 0; i--) {
```
dschneider-pivotal, line 73: its okay to unlock in reverse order but not needed. For lock we need to do it inorder but once we hold all the locks we can unlock in any order without getting in trouble.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/services/LockingStripedCoordinator.java
```diff
@@ -0,0 +1,103 @@
...
+
+  private Lock getLock(RedisKey stripeId) {
+    return locks[getStripeIndex(stripeId)];
+  }
+
+  private Lock[] getLocks(List<RedisKey> stripeIds) {
+    // Sort multiple keys in a consistent manner so that locking always happens in the same order
+    // across multiple keys.
+    stripeIds.sort(this::compareStripes);
+    Lock[] locks = new Lock[stripeIds.size()];
```
dschneider-pivotal, line 87: we really don't need this array of locks. Instead we could have sorted the stripeIds once and then just iterated it for both lock and unlock. For each RedisKey we would call getLock(RedisKey) in these iterations. I'm not sure which is best but unless you think the getLock call is expensive its probably better to call it twice (instead of once) for each key to save the extra ArrayList allocation. If you do get rid of this extra array of locks you could then have a more concise iteration (for example: stripeIds.foreach(item -> getLock(item).lock())

dschneider-pivotal (changes requested):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/RegionProvider.java
jdeppe-pivotal: I used this as a guide for the current code: https://gemfire.docs.pivotal.io/910/geode/developing/transactions/directed_example.html.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisStringCommandsFunctionExecutor.java
jdeppe-pivotal: Yes! Good call.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/services/LockingStripedCoordinator.java
jdeppe-pivotal: Good point. I've removed that extra allocation.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/RegionProvider.java
jdeppe-pivotal: After our discussion, I've made the changes you suggested.

dschneider-pivotal (approved):

PR #6863, opened 2021-09-13
GEODE-9595: Exclude circular integrationTest classpath dependencies

kirklund:
The bug GEODE-9595 results in Analyze*Module*SerializablesIntegrationTests failing with the following when an actual serializable is missing from the sanctioned serializables and the excluded list:
```
java.nio.file.FileSystemNotFoundException
	at com.sun.nio.zipfs.ZipFileSystemProvider.getFileSystem(ZipFileSystemProvider.java:171)
	at com.sun.nio.zipfs.ZipFileSystemProvider.getPath(ZipFileSystemProvider.java:157)
	at java.nio.file.Paths.get(Paths.java:143)
	at org.apache.geode.codeAnalysis.AnalyzeDataSerializablesJUnitTestBase.mainResourceToSourcePath(AnalyzeDataSerializablesJUnitTestBase.java:260)
	at org.apache.geode.codeAnalysis.AnalyzeSerializablesJUnitTestBase.testSerializables(AnalyzeSerializablesJUnitTestBase.java:102)
```
Debugging reveals that the above is caused by the integrationTest classpath loading classes and resources from geode-*module*/build/libs/geode-*module*.jar instead of from geode-*module*/build/classes and geode-*module*/build/resources.

Example in geode-core:

Execute `$ ./gradlew geode-core:dependencies --configuration integrationTestRuntimeClasspath | less`

Search for `geode-core` and note that the following dependencies will add the geode-core jar file as a dependency. Integration tests will then load classes and resources from that jar file instead of the filesystem which is incorrect behavior unless a testing target is specifically written to use the jars:
```
+--- project :geode-gfsh
|    +--- <snip>
|    +--- project :geode-core (*)
```
```
+--- project :geode-dunit
|    +--- <snip>
|    +--- project :geode-core (*)
|    +--- <snip>
|    +--- project :geode-cq
|    |    +--- <snip>
|    |    +--- project :geode-core (*)
|    +--- project :geode-log4j
|    |    +--- <snip>
|    |    +--- project :geode-core (*)
```
```
+--- project :geode-deployment:geode-deployment-legacy
|    +--- <snip>
|    +--- project :geode-core (*)
```
The fix involves excluding `geode-core` from the above identified integrationTestImplementation dependencies:
```
    integrationTestImplementation(project(':geode-gfsh')) {
        exclude module: 'geode-core'
    }
    integrationTestImplementation(project(':geode-junit')) {
        exclude module: 'geode-core'
    }
    integrationTestImplementation(project(':geode-dunit')) {
        exclude module: 'geode-core'
    }
    integrationTestImplementation(project(':geode-log4j')) {
        exclude module: 'geode-core'
    }
```
This PR fixes the circular dependencies that add geode-*module*.jars to the classpath of tests in modules:
- geode-core
- geode-cq
- geode-gfsh
- geode-maangement
- geode-wan
- geode-web-api

### DIFF
--- a/geode-core/build.gradle
+++ b/geode-core/build.gradle
@@ -233,7 +233,9 @@ dependencies {
         exclude group: '*'
     }
 
-    runtimeOnly(project(':geode-deployment:geode-deployment-legacy'))
+    runtimeOnly(project(':geode-deployment:geode-deployment-legacy')) {
+        exclude module: 'geode-core'
+    }
 
     //Commons lang is used in many different places in core
     implementation('org.apache.commons:commons-lang3')
@@ -349,10 +351,18 @@ dependencies {
 
     // Needed for JDK8, not JDK11, after nebula.facet v7.0.9
     integrationTestImplementation(files("${System.getProperty('java.home')}/../lib/tools.jar"))
-    integrationTestImplementation(project(':geode-gfsh'))
-    integrationTestImplementation(project(':geode-junit'))
-    integrationTestImplementation(project(':geode-dunit'))
-    integrationTestImplementation(project(':geode-log4j'))
+    integrationTestImplementation(project(':geode-gfsh')) {
+        exclude module: 'geode-core'
+    }
+    integrationTestImplementation(project(':geode-junit')) {
+        exclude module: 'geode-core'
+    }
+    integrationTestImplementation(project(':geode-dunit')) {
+        exclude module: 'geode-core'
+    }
+    integrationTestImplementation(project(':geode-log4j')) {
+        exclude module: 'geode-core'
+    }
     integrationTestImplementation(project(':geode-concurrency-test'))
     integrationTestImplementation('org.apache.bcel:bcel')
     integrationTestImplementation('org.apache.logging.log4j:log4j-core')
@@ -364,7 +374,9 @@ dependencies {
     integrationTestRuntimeOnly('xerces:xercesImpl')
     integrationTestRuntimeOnly('com.fasterxml.jackson.datatype:jackson-datatype-jsr310')
 
-    distributedTestImplementation(project(':geode-gfsh'))
+    distributedTestImplementation(project(':geode-gfsh')) {
+        exclude module: 'geode-core'
+    }
     distributedTestImplementation(project(':geode-junit')) {
         exclude module: 'geode-core'
     }
--- a/geode-management/build.gradle
+++ b/geode-management/build.gradle
@@ -49,7 +49,9 @@ dependencies {
   testCompileOnly(platform(project(':boms:geode-all-bom')))
   testCompileOnly('io.swagger:swagger-annotations')
 
-  integrationTestImplementation(project(':geode-dunit'))
+  integrationTestImplementation(project(':geode-dunit')) {
+    exclude module: 'geode-management'
+  }
   integrationTestImplementation(project(':geode-junit'))
   integrationTestImplementation('junit:junit')
   integrationTestImplementation('org.awaitility:awaitility')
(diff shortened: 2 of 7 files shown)


demery-pivotal (approved):
The need for this seems like a sign of a deeper problem. We should create a ticket to address the deeper problem.

geode-web-api/build.gradle
```diff
@@ -104,6 +104,7 @@ dependencies {
   }
   integrationTestImplementation(project(':geode-dunit')) {
     exclude module: 'geode-core'
+    exclude module: 'geode-web-api'
```
rhoughton-pivot, line 107: geode-dunit does not bring in geode-web-api.

geode-wan/build.gradle
```diff
@@ -62,7 +64,9 @@ dependencies {
 
   upgradeTestImplementation(project(':geode-gfsh'))
   upgradeTestImplementation(project(':geode-junit'))
-  upgradeTestImplementation(project(':geode-dunit'))
+  upgradeTestImplementation(project(':geode-dunit')) {
```
rhoughton-pivot, line 67: geode-dunit does not bring in geode-wan

geode-wan/build.gradle
```diff
@@ -45,7 +45,9 @@ dependencies {
 
 
   distributedTestImplementation(project(':geode-gfsh'))
-  distributedTestImplementation(project(':geode-dunit'))
+  distributedTestImplementation(project(':geode-dunit')) {
+    exclude module: 'geode-wan'
```
rhoughton-pivot, line 49: geode-dunit does not bring in geode-wan

geode-management/build.gradle
```diff
@@ -49,7 +49,9 @@ dependencies {
   testCompileOnly(platform(project(':boms:geode-all-bom')))
   testCompileOnly('io.swagger:swagger-annotations')
 
-  integrationTestImplementation(project(':geode-dunit'))
+  integrationTestImplementation(project(':geode-dunit')) {
```
rhoughton-pivot, line 52: geode-dunit needs geode-core, which needs geode-management. What is the actual concern in the testing here?

geode-core/build.gradle
```diff
@@ -233,7 +233,9 @@ dependencies {
         exclude group: '*'
     }
 
-    runtimeOnly(project(':geode-deployment:geode-deployment-legacy'))
+    runtimeOnly(project(':geode-deployment:geode-deployment-legacy')) {
+        exclude module: 'geode-core'
```
rhoughton-pivot, line 237: What is the actual problem being addressed by this change?

`geode-core` needs `geode-deployment:geode-deployment-legacy`, which needs `geode-core` to exist on the classpath. At runtime, this will be true. Why not declare it?

rhoughton-pivot (changes requested):

geode-core/build.gradle
kirklund: You should be pairing with Dale and me then :) This was hard to debug.

I'm trying to back GEODE-9486. AnalyzeSerializables integration tests load the `sanctioned-geode-*-serializables.txt` resource and then generates a failure message to point the person running the test to overwrite the `sanctioned-geode-*-serializables.txt` with the generated actual serializables file. However, the error message was broken. After fixing it, it fails only in the modules that have these circular dependencies. We discovered that when the test runs (only in modules with circular dependencies) the classes and resources are loaded from the geode-*module*/build/libs/geode-*module*.jar which causes the failure message to blow up.

How about we pair from 10am to noon tomorrow?

kirklund:
Replaced by #6867.

PR #6865, opened 2021-09-14
GEODE-9458: Enhance function execution testing auth expiry

jmelchio:
Added multi server testing
Exercise all client executable FunctionService interfaces

### DIFF
--- a/geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationFunctionDUnitTest.java
+++ b/geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationFunctionDUnitTest.java
@@ -17,9 +17,10 @@ package org.apache.geode.security;
 import static org.apache.geode.cache.execute.FunctionService.onRegion;
 import static org.apache.geode.cache.execute.FunctionService.onServer;
 import static org.apache.geode.cache.execute.FunctionService.onServers;
-import static org.apache.geode.distributed.ConfigurationProperties.GROUPS;
 import static org.apache.geode.distributed.ConfigurationProperties.SECURITY_CLIENT_AUTH_INIT;
 import static org.apache.geode.distributed.ConfigurationProperties.SECURITY_MANAGER;
+import static org.apache.geode.security.ClientAuthenticationTestUtils.combineSecurityManagerResults;
+import static org.apache.geode.security.ClientAuthenticationTestUtils.getSecurityManager;
 import static org.apache.geode.security.SecurityManager.PASSWORD;
 import static org.apache.geode.security.SecurityManager.USER_NAME;
 import static org.apache.geode.test.version.VersionManager.CURRENT_VERSION;
@@ -28,8 +29,6 @@ import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
-import java.util.HashMap;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -96,7 +95,6 @@ public class AuthExpirationFunctionDUnitTest {
     serverProperties.setProperty(SECURITY_MANAGER, ExpirableSecurityManager.class.getName());
     serverProperties.setProperty(ConfigurationProperties.SERIALIZABLE_OBJECT_FILTER,
         "org.apache.geode.management.internal.security.TestFunctions*");
-    serverProperties.setProperty(GROUPS, "group");
     serverProperties.setProperty(USER_NAME, "test");
     serverProperties.setProperty(PASSWORD, "test");
 
@@ -141,23 +139,21 @@ public class AuthExpirationFunctionDUnitTest {
     assertThat(((ArrayList) rc.getResult()).get(0))
         .isEqualTo(TestFunctions.WriteFunction.SUCCESS_OUTPUT);
 
-    // all function invocation authorizations are recorded
-    List<Object> resultsVM0 = collectSecurityManagerResults(serverVM0);
-    List<Object> resultsVM1 = collectSecurityManagerResults(serverVM1);
-    List<Object> resultsVM2 = collectSecurityManagerResults(serverVM2);
 
-    Set<String> combinedExpiredUsers = combineExpiredUsers(resultsVM0, resultsVM1, resultsVM2);
+    ExpirableSecurityManager consolidated =
+        combineSecurityManagerResults(serverVM0, serverVM1, serverVM2);
+    Set<String> combinedExpiredUsers = consolidated.getExpiredUsers();
 
     assertThat(combinedExpiredUsers.size()).isEqualTo(1);
     assertThat(combinedExpiredUsers.contains("data1")).isTrue();
-    Map<String, List<String>> authorizedOps = collectVMOps(1, resultsVM0, resultsVM1, resultsVM2);
+    Map<String, List<String>> authorizedOps = consolidated.getAuthorizedOps();
 
     assertThat(authorizedOps.get("data1")).asList().hasSize(1);
     assertThat(authorizedOps.get("data1")).asList().containsExactly("DATA:WRITE");
     assertThat(authorizedOps.get("data2")).asList().hasSize(1);
     assertThat(authorizedOps.get("data2")).asList().containsExactly("DATA:WRITE");
 
-    Map<String, List<String>> unauthorizedOps = collectVMOps(2, resultsVM0, resultsVM1, resultsVM2);
+    Map<String, List<String>> unauthorizedOps = consolidated.getUnAuthorizedOps();
 
     assertThat(unauthorizedOps.get("data1")).asList().hasSize(1);
     assertThat(unauthorizedOps.get("data1")).asList().containsExactly("DATA:WRITE");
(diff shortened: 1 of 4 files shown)


geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationFunctionDUnitTest.java
```diff
@@ -65,77 +76,327 @@
...
+    MemberVM locatorVM =
+        lsRule.startLocatorVM(0, l -> l.withSecurityManager(ExpirableSecurityManager.class));
+    int locatorPort = locatorVM.getPort();
+
+    Properties serverProperties = new Properties();
+    serverProperties.setProperty(SECURITY_MANAGER, ExpirableSecurityManager.class.getName());
+    serverProperties.setProperty(ConfigurationProperties.SERIALIZABLE_OBJECT_FILTER,
         "org.apache.geode.management.internal.security.TestFunctions*");
-    serverVM = lsRule.startServerVM(0, properties);
+    serverProperties.setProperty(GROUPS, "group");
```
jinmeiliao, line 99: you don't need to specify group, right?

geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationFunctionDUnitTest.java
```diff
@@ -65,77 +76,327 @@
...
+        serverVM0, serverVM1, serverVM2);
+
+    // do a second function execution, if this is successful, it means new credentials are provided
+    UpdatableUserAuthInitialize.setUser("data2");
+    rc = onRegion(region).execute(writeFunction);
+    assertThat(((ArrayList) rc.getResult()).get(0))
+        .isEqualTo(TestFunctions.WriteFunction.SUCCESS_OUTPUT);
+
+    // all function invocation authorizations are recorded
+    List<Object> resultsVM0 = collectSecurityManagerResults(serverVM0);
```
jinmeiliao, line 236: seems like every test needs to combine the results, see `AuthExpirationMultiServerDUnitTest.gatherAuthorizedAndUnauthorizedOps` method, maybe we can make that method as a utility test so that all others can use it.

jinmeiliao (changes requested):

geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationFunctionDUnitTest.java
jmelchio: > seems like every test needs to combine the results, see `AuthExpirationMultiServerDUnitTest.gatherAuthorizedAndUnauthorizedOps` method, maybe we can make that method as a utility test so that all others can use it.

Yes, let me do some refactoring to reduce code.

geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationFunctionDUnitTest.java
jmelchio: > you don't need to specify group, right?

You're right. That can be removed.

geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationFunctionDUnitTest.java
```diff
@@ -65,77 +76,327 @@
...
+        serverVM0, serverVM1, serverVM2);
+
+    // do a second function execution, if this is successful, it means new credentials are provided
+    UpdatableUserAuthInitialize.setUser("data2");
+    rc = onServer(regionService).execute(writeFunction);
+    assertThat(((ArrayList) rc.getResult()).get(0))
+        .isEqualTo(TestFunctions.WriteFunction.SUCCESS_OUTPUT);
+
+    // all function invocation authorizations are recorded
+    List<Object> resultsVM0 = collectSecurityManagerResults(serverVM0);
```
jchen21, line 285: There is some code duplication here, compared with previous test. You might want to consider extracting the code to a method.

mhansonp:
Can you cleanup the spotless errors so we can have a clean run of the tests before I review this PR?

geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationFunctionDUnitTest.java
```diff
@@ -65,77 +76,327 @@
...
 
-  private MemberVM serverVM;
-  private ClientVM clientVM;
+  private MemberVM serverVM0;
+  private MemberVM serverVM1;
+  private MemberVM serverVM2;
 
   @Rule
-  public ClusterStartupRule lsRule = new ClusterStartupRule();
+  public ClusterStartupRule lsRule = new ClusterStartupRule(4);
```
kirklund, line 83: Not sure what `ls` refers to in `lsRule`. Maybe rename it to `clusterRule`.

geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationFunctionDUnitTest.java
```diff
@@ -65,77 +76,327 @@
...
-    properties.setProperty(SECURITY_MANAGER, ExpirableSecurityManager.class.getName());
-    properties.setProperty(ConfigurationProperties.SERIALIZABLE_OBJECT_FILTER,
+  public void setup() {
+    MemberVM locatorVM =
+        lsRule.startLocatorVM(0, l -> l.withSecurityManager(ExpirableSecurityManager.class));
+    int locatorPort = locatorVM.getPort();
+
+    Properties serverProperties = new Properties();
+    serverProperties.setProperty(SECURITY_MANAGER, ExpirableSecurityManager.class.getName());
+    serverProperties.setProperty(ConfigurationProperties.SERIALIZABLE_OBJECT_FILTER,
```
kirklund, line 96: I would go ahead and `import static` the `SERIALIZABLE_OBJECT_FILTER` to match `SECURITY_MANAGER`. I generally use `import static` as much as possible for any statics that read better using it and only skip doing this for anything that reads more clearly without it.

geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationFunctionDUnitTest.java
```diff
@@ -65,77 +76,327 @@
...
+    List<Object> resultsVM0 = collectSecurityManagerResults(serverVM0);
+    List<Object> resultsVM1 = collectSecurityManagerResults(serverVM1);
+    List<Object> resultsVM2 = collectSecurityManagerResults(serverVM2);
 
-    // all put operation succeeded
-    serverVM.invoke(() -> {
+    Set<String> combinedExpiredUsers = combineExpiredUsers(resultsVM0, resultsVM1, resultsVM2);
+
+    assertThat(combinedExpiredUsers.size()).isEqualTo(1);
+    assertThat(combinedExpiredUsers.contains("data1")).isTrue();
```
kirklund, line 233: These two assertions are used in several places. It would be better if you use the `Collection` assertion everywhere instead of the `boolean` assertion:
```
assertThat(combinedExpiredUsers).hasSize(1);
assertThat(combinedExpiredUsers).contains("data1");
```
Or:
```
assertThat(combinedExpiredUsers)
    .hasSize(1)
    .contains("data1");
```
Or:
```
assertThat(combinedExpiredUsers).containsExactly("data1");
```

geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationFunctionDUnitTest.java
```diff
@@ -65,77 +76,327 @@
...
+  @Test
+  public void clientShouldReAuthenticateWhenCredentialExpiredAndFunctionExecutionOnServerSucceed()
+      throws Exception {
+    ClientCache clientCache = clientCacheRule.createCache();
+    UpdatableUserAuthInitialize.setUser("data1");
+    writeFunction = new TestFunctions.WriteFunction();
+
+    ResultCollector rc = onServer(clientCache.getDefaultPool()).execute(writeFunction);
+    assertThat(((ArrayList) rc.getResult()).get(0))
+        .isEqualTo(TestFunctions.WriteFunction.SUCCESS_OUTPUT);
```
kirklund, line 130: You should always try to cast to an interface rather than a concrete implementation:
```
List<?> result = (List) rc.getResult();
assertThat(result.get(0))
    .isEqualTo(TestFunctions.WriteFunction.SUCCESS_OUTPUT);
```

kirklund (changes requested):

mhansonp (approved):

geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationFunctionDUnitTest.java
```diff
@@ -14,19 +14,26 @@
...
+import static org.apache.geode.cache.execute.FunctionService.onServers;
 import static org.apache.geode.distributed.ConfigurationProperties.SECURITY_CLIENT_AUTH_INIT;
 import static org.apache.geode.distributed.ConfigurationProperties.SECURITY_MANAGER;
+import static org.apache.geode.distributed.ConfigurationProperties.SERIALIZABLE_OBJECT_FILTER;
+import static org.apache.geode.security.ClientAuthenticationTestUtils.combineSecurityManagerResults;
+import static org.apache.geode.security.ClientAuthenticationTestUtils.getSecurityManager;
+import static org.apache.geode.security.SecurityManager.PASSWORD;
+import static org.apache.geode.security.SecurityManager.USER_NAME;
 import static org.apache.geode.test.version.VersionManager.CURRENT_VERSION;
 import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
```
jinmeiliao, line 28: I think you meant to use `org.assertj.core.api.AssertThat` instead

jinmeiliao (approved):

kirklund (approved):

geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationFunctionDUnitTest.java
jmelchio: > I think you meant to use `org.assertj.core.api.AssertThat` instead

Good catch!

PR #6870, opened 2021-09-15
GEODE-9582: make radish pattern match compatible with redis

dschneider-pivotal:
GlobPattern now uses code derived from native redis's glob matching.
It should be completely compatible with native redis unless I broke something
in the conversion to java. This has the benefit of being able to do the match with
a byte array (the jdk matcher required a conversion to a String).
This also cleaned up some of our other code that handled the exception
thrown from the jdk when compiling a pattern.

### For all changes:

### DIFF
--- a/geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/executor/GlobPatternTest.java
+++ b/geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/executor/GlobPatternTest.java
@@ -1,0 +1,88 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.apache.geode.redis.internal.executor;
+
+import static org.apache.geode.redis.internal.netty.Coder.stringToBytes;
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.Test;
+
+public class GlobPatternTest {
+  private GlobPattern createPattern(String pattern) {
+    return new GlobPattern(stringToBytes(pattern));
+  }
+
+  @Test
+  public void verifyEmptyInput() {
+    assertThat(createPattern("*").matches(stringToBytes(""))).isFalse();
+    assertThat(createPattern("**").matches(stringToBytes(""))).isFalse();
+    assertThat(createPattern("?").matches(stringToBytes(""))).isFalse();
+    assertThat(createPattern("").matches(stringToBytes(""))).isTrue();
+  }
+
+  @Test
+  public void verifyOneByteInput() {
+    assertThat(createPattern("*").matches(stringToBytes("a"))).isTrue();
+    assertThat(createPattern("**").matches(stringToBytes("b"))).isTrue();
+    assertThat(createPattern("?").matches(stringToBytes("@"))).isTrue();
+    assertThat(createPattern("").matches(stringToBytes("a"))).isFalse();
+  }
+
+  @Test
+  public void patternThatEndsWithWildcards() {
+    assertThat(createPattern("foo****").matches(stringToBytes("foo"))).isTrue();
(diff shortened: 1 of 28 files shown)


DonalEvans (commented):
Would it be possible to add some integration tests (or modify existing ones) to verify that Radish and native Redis behave the same for some of the more unusual patterns? The unit tests for `GlobPattern` are great, but they don't guarantee that what we're doing is the same as what Redis is doing. I'm thinking about situations like this:
```
127.0.0.1:6379> psubscribe \c
Reading messages... (press Ctrl-C to quit)
1) "psubscribe"
2) "\\c"
3) (integer) 1

127.0.0.1:6379> psubscribe "\c"
Reading messages... (press Ctrl-C to quit)
1) "psubscribe"
2) "c"
3) (integer) 1

127.0.0.1:6379> psubscribe "\\c"
Reading messages... (press Ctrl-C to quit)
1) "psubscribe"
2) "\\c"
3) (integer) 1
```
where wether or not the pattern is in quotes changes how it's handled.

dschneider-pivotal:
@DonalEvans The native redis code that parses the glob pattern does nothing with quotes. So I think what you are seeing is something redis-cli adds. If a double or single quote end up getting passed in as part of the pattern it will just be treated as a literal character. The ones I added to the unit test were to make sure all the parsing code had coverage. I can add those also to the existing glob integration test

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/executor/GlobPatternTest.java
```diff
@@ -0,0 +1,84 @@
...
+  public void caseSensitive() {
+    assertThat(createPattern("foo****").matches(stringToBytes("Foo"))).isFalse();
+    assertThat(createPattern("foo****").matches(stringToBytes("foo"))).isTrue();
+    assertThat(createPattern("[a-z]oo").matches(stringToBytes("Foo"))).isFalse();
+    assertThat(createPattern("[a-z]oo").matches(stringToBytes("goo"))).isTrue();
+  }
+
+  @Test
+  public void escapeMakesSpecialsLiteral() {
+    assertThat(createPattern("b\\*r").matches(stringToBytes("b*r"))).isTrue();
```
jdeppe-pivotal, line 68: Also add:
```
    assertThat(createPattern("b\\*r").matches(stringToBytes("bar"))).isFalse();
```

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/executor/GlobPatternTest.java
```diff
@@ -0,0 +1,84 @@
...
+    assertThat(createPattern("[z-a]").matches(stringToBytes("z"))).isTrue();
+    assertThat(createPattern("[z-a]").matches(stringToBytes("a"))).isTrue();
+    assertThat(createPattern("[z-a]").matches(stringToBytes("m"))).isTrue();
+    assertThat(createPattern("[^a-a]").matches(stringToBytes("a"))).isFalse();
+    assertThat(createPattern("[a-a]").matches(stringToBytes("a"))).isTrue();
+    assertThat(createPattern("[a-a]").matches(stringToBytes("b"))).isFalse();
+    assertThat(createPattern("[a\\-z]").matches(stringToBytes("-"))).isTrue();
+    assertThat(createPattern("[a").matches(stringToBytes("a"))).isTrue();
+  }
+}
```
jdeppe-pivotal, line 84: Would you mind adding some tests with multiple `*`s something like this:
```
    assertThat(createPattern("*bar*").matches(stringToBytes("foobarbaz"))).isTrue();
    assertThat(createPattern("*bar*joe").matches(stringToBytes("foorbarbazjoe"))).isTrue();
    assertThat(createPattern("*bar*joe").matches(stringToBytes("foorbarbazjoey"))).isFalse();
```

jdeppe-pivotal (changes requested):
I also ran some basic JMH tests against a simple `sometext*` pattern (which is probably very common) and the new code appears to be slightly faster! (than previously).

DonalEvans:
> @DonalEvans The native redis code that parses the glob pattern does nothing with quotes. So I think what you are seeing is something redis-cli adds. If a double or single quote end up getting passed in as part of the pattern it will just be treated as a literal character. The ones I added to the unit test were to make sure all the parsing code had coverage. I can add those also to the existing glob integration test

Ah, thanks for the explanation. Given that, I think the existing coverage is sufficient.

DonalEvans (approved):

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/executor/GlobPatternTest.java
dschneider-pivotal: yes

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/executor/GlobPatternTest.java
dschneider-pivotal: done

jdeppe-pivotal (approved):

PR #6871, opened 2021-09-15
GEODE-9552: Change how Radish server shutdown/disconnect is handled

DonalEvans:
- Close client connection on CancelException
 - Fix handling of RedisCommandParserException
 - Remove unused constants

### For all changes:


upthewaterspout (approved):
I wondering why we never saw a problem with CacheClosedException, honestly. I looked at the redis source code, and I don't see that "The server is shutting down" error message sent back to a client anywhere, so I would be surprised if that was being handled correctly by clients.

DonalEvans:
> I wondering why we never saw a problem with CacheClosedException, honestly. I looked at the redis source code, and I don't see that "The server is shutting down" error message sent back to a client anywhere, so I would be surprised if that was being handled correctly by clients.

It's surprising, yeah. I wonder if it's possible we got lucky and were just never doing anything that would cause a CacheClosedException to be thrown to here.

dschneider-pivotal (approved):

PR #6873, opened 2021-09-15
GEODE-9575: improve how publish executes a function

dschneider-pivotal:
- getRegionMembers now uses the advisor to compute
the set of members hosting the redis region.
- PubSubImpl now has a PublishFunction in place of the old
anonymous inner class.
- PubSubDUnitTest had a timing issue after killing a server.
The next test would fail due to the state left around by
the test that did the kill. Now this test gives each test
method a new clean slate and is passing consistently.

### For all changes:

### DIFF
--- a/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/RegionProvider.java
+++ b/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/RegionProvider.java
@@ -14,6 +14,7 @@
  */
 package org.apache.geode.redis.internal;
 
+import static java.util.Collections.singleton;
 import static org.apache.geode.redis.internal.data.NullRedisDataStructures.NULL_REDIS_DATA;
 import static org.apache.geode.redis.internal.data.NullRedisDataStructures.NULL_REDIS_HASH;
 import static org.apache.geode.redis.internal.data.NullRedisDataStructures.NULL_REDIS_SET;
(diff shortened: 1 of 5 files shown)


geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/RegionProvider.java
```diff
@@ -300,14 +299,18 @@ public RedisKeyCommands getKeyCommands() {
...
-      membersWithDataRegion.add(memberInfo.getDistributedMember());
+    InternalDistributedMember myId =
+        partitionedRegion.getDistributionManager().getDistributionManagerId();
+    Set<InternalDistributedMember> otherIds =
+        partitionedRegion.getRegionAdvisor().adviseDataStore();
+    if (otherIds.isEmpty()) {
+      return singleton(myId);
+    } else {
+      otherIds.add(myId);
+      return (Set<DistributedMember>) (Set<?>) otherIds;
```
jdeppe-pivotal, line 312: Not very important, but I think you can call `adviseDataStore(true)` here and then avoid this empty check. It's the same thing essentially but would avoid folks wondering why it's done this way.

jdeppe-pivotal (approved):
I left one comment, but not a big deal if you want to leave it as is.

DonalEvans (approved):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/RegionProvider.java
dschneider-pivotal: I reworked this so that this method only returns remote members and the caller of it executes the publish directly for the local member and only using the FunctionService for remote members.

PR #6874, opened 2021-09-16
GEODE-9602: QueryObserver improvements.

albertogpz:
- Make QueryObserverHolder thread-safe
- Allow having an observer per query by means of setting the observer
  in the query at the start of the execution.
- Invoke beforeIterationEvaluation and afterIterationEvaluation callbacks when
  query is using indexes.

### For all changes:


geode-core/src/integrationTest/java/org/apache/geode/cache/query/internal/QueryObserverCallbacksTest.java
```diff
@@ -243,6 +245,59 @@ public void beforeAggregationsAndGroupByShouldBeCalledForAggregateFunctions() th
...
   }
 
+  @Test
+  public void testBeforeAndAfterIterationEvaluateNoWhere() throws Exception {
+    Query query = queryService.newQuery(
+        "select count(*) from " + SEPARATOR + "portfolio p");
+
+    query.execute();
+    verify(myQueryObserver, times(0)).beforeIterationEvaluation(any(), any());
+    verify(myQueryObserver, times(0)).afterIterationEvaluation(any());
```
kirklund, line 255: You can also use `never()`:
```java
    verify(myQueryObserver, never()).beforeIterationEvaluation(any(), any());
    verify(myQueryObserver, never()).afterIterationEvaluation(any());
```

geode-core/src/main/java/org/apache/geode/cache/query/internal/QueryObserverHolder.java
```diff
@@ -54,26 +54,26 @@
...
   }
 
-  public static boolean hasObserver() {
+  public static synchronized boolean hasObserver() {
     return _instance != NO_OBSERVER;
   }
 
   /** Return the current QueryObserver instance */
-  public static QueryObserver getInstance() {
+  public static synchronized QueryObserver getInstance() {
```
kirklund, line 69: I've wanted to make this class thread-safe too. There is one problem with doing so though. `getInstance()` is in the path of every query execution and index evaluate call. If you make it `synchronized`, then there will be a significant performance penalty inflicted on every query and index evaluation.

To see this, look at the call hierarchy of constructing a new ExecutionConext. Every new instance will invoke:
```java
private QueryObserver observer = QueryObserverHolder.getInstance();
```

In order to make this class thread-safe, I think we need to solve the problem of how to get queries and index evaluations from hitting those `synchronized` blocks. I'm not as familiar with querying or indexing so you may need to chat about it further with Anil or others.

Another problem is that the new `private QueryObserver observer` field in `ExecutionContext` is itself not thread-safe, so the setter and getter you added are not thread-safe either.

I think your intention is to have a test invoke `setObserver` for one or more direct invocations of a query. If that's true, then you're probably better off introducing a `ThreadLocal` to `ExecutionContext` for this purpose. Then the execution would check that `ThreadLocal` for a non-null value and then use it. If it's null, the execution would fallback to using the usual `observer` stored in the field.

Then you could make the field itself `final` in the hopes that it would be used multiple times. But again, I suspect that every query invocation creates a new `ExecutionContext` which will perform poorly because of the `synchronized` in `QueryObserverHolder`.

geode-core/src/main/java/org/apache/geode/cache/query/internal/QueryObserverHolder.java
```diff
@@ -54,26 +54,26 @@
   /**
    * Set the given observer to be notified of query events. Returns the current observer.
    */
-  public static QueryObserver setInstance(QueryObserver observer) {
+  public static synchronized QueryObserver setInstance(QueryObserver observer) {
```
kirklund, line 57: Rather than using `synchronized`, you'd be better off making `_instance` a `private static final AtomicReference<QueryObserver>`. The use of `synchronized` will cause at least some JVM implementations to pin every thread hitting it to the same CPU.

Using atomics is generally always a better approach if you can use it.

geode-core/src/main/java/org/apache/geode/cache/query/internal/index/AbstractIndex.java
```diff
@@ -1826,9 +1828,14 @@ void addValuesToCollection(Collection result, CompiledValue iterOp, RuntimeItera
...
-                    runtimeItr.setCurrent(o1);
-                    ok = QueryUtils.applyCondition(iterOp, context);
+                  try {
+                    if (ok && runtimeItr != null) {
+                      runtimeItr.setCurrent(o1);
+                      observer.beforeIterationEvaluation(iterOp, o1);
+                      ok = QueryUtils.applyCondition(iterOp, context);
+                    }
+                  } finally {
+                    observer.afterIterationEvaluation(ok);
```
kirklund, line 1838: I think you should try to extract the blocks that look like this to their own private methods. There is some repetition involved that you may be able to reduce down to one method. The code calling these methods will also become more readable.

geode-core/src/main/java/org/apache/geode/cache/query/internal/index/CompactRangeIndex.java
```diff
@@ -803,8 +809,13 @@ private void addToResultsFromEntries(Object lowerBoundKey, Object upperBoundKey,
...
               }
-              if (ok && runtimeItr != null && iterOps != null) {
-                ok = QueryUtils.applyCondition(iterOps, context);
+              try {
+                if (ok && runtimeItr != null) {
+                  observer.beforeIterationEvaluation(iterOps, value);
+                  ok = QueryUtils.applyCondition(iterOps, context);
+                }
+              } finally {
+                observer.afterIterationEvaluation(ok);
```
kirklund, line 818: This class also has several try-finally blocks like that this may benefit by being extracted as a new private method both for readability and to hopefully reduce some of the repetition in these blocks.

kirklund (changes requested):

geode-core/src/integrationTest/java/org/apache/geode/cache/query/internal/QueryTraceJUnitTest.java
```diff
@@ -296,8 +319,11 @@ public void testTraceOnLocalRegionWithTracePrefixNoComments() throws Exception {
     assertTrue(((DefaultQuery) query).isTraced());
 
     SelectResults results = (SelectResults) query.execute();
-    assertTrue(QueryObserverHolder.getInstance() instanceof IndexTrackingQueryObserver);
-    // The query should return all elements in region.
```
DonalEvans, line 300: This comment seems to have been accidentally removed.

DonalEvans (approved):
I'm approving this PR, but I share Kirk's concerns over query performance.

geode-core/src/main/java/org/apache/geode/cache/query/internal/QueryObserverHolder.java
agingade: It looks like this is not going to achieve what you are looking for:
> The QueryObserverHolder class is not thread-safe. 
As Kirk pointed out, you have to make this as a ThreadLocal and set/reset while in use.
Other option could be making this as non-static...If the query context is available in all the places where query-observer is getting used, you can think of creating a new QueryObserver and setting it on the context. 
QueryContext.setObserver(new QueryObserver()); And no static implementation/use of query observer.

I did not see any test changes where concurrent queries are using QueryObserver...It will show if the query observers are cross referenced...

agingade (changes requested):

geode-core/src/main/java/org/apache/geode/cache/query/internal/QueryObserverHolder.java
albertogpz: > I've wanted to make this class thread-safe too. There is one problem with doing so though. `getInstance()` is in the path of every query execution and index evaluate call. If you make it `synchronized`, then there will be a significant performance penalty inflicted on every query and index evaluation.
> 
> To see this, look at the call hierarchy of constructing a new ExecutionConext. Every new instance will invoke:
> 
> ```java
> private QueryObserver observer = QueryObserverHolder.getInstance();
> ```
> 
> In order to make this class thread-safe, I think we need to solve the problem of how to get queries and index evaluations from hitting those `synchronized` blocks on every invocation. I'm not as familiar with querying or indexing so you may need to chat about it further with Anil or others.
> 
I would expect this to be a problem if the above calls are done many times per query. But in this PR, it is proposed to invoke it just once, to get the value and then set it in the `ExecutionContext`, right before the query is executed. Later, when the query needs the value of the observer, it would get it from the `ExecutionContext` using a non-synchronized call given that all the calls to get the observer have been changed from `QueryObserverHolder.getInstance()` to `context.getObserver()`.

> Another problem is that the new `private QueryObserver observer` field in `ExecutionContext` is itself not thread-safe, so the setter and getter you added are not thread-safe either.
> 
I am assuming that the `ExecutionContext` is not shared between threads and therefore it does not need to be thread-safe. I am wrong here?
> I think your intention is to have a test invoke `setObserver` for one or more direct invocations of a query. If that's true, then you're probably better off introducing a `ThreadLocal` to `ExecutionContext` for this purpose. Then the execution would check that `ThreadLocal` for a non-null value and then use it. If it's null, the execution would fallback to using the usual `observer` stored in the field.
> 
Actually, my intention was just to make this class thread-safe but not extend it to have new usages. Nevertheless, the class is documented and used so that the observer can be set from any thread and be available to any thread in the VM.
I have proposed the current solution but have also considered the use of a `ThreadLocal` variable, not in the `ExecutionContext` class (which did not have an `observer` field) but in the `QueryObserverHolder` class to be used in a way similar to what you are proposing.
But I think the lifecycle of the `ThreadLocal` variable would be complex. It should be reset after the query execution. But who would do it?

> Then you could make the field itself `final` in the hopes that it would be used multiple times. But again, I suspect that every query invocation creates a new `ExecutionContext` which will perform poorly because of the `synchronized` in `QueryObserverHolder`.

albertogpz: > It looks like this is not going to achieve what you are looking for:
> 
> > The QueryObserverHolder class is not thread-safe.
> 
> As Kirk pointed out, you have to make this as a ThreadLocal and set/reset while in use.
> Other option could be making this as non-static...If the query context is available in all the places where query-observer is getting used, you can think of creating a new QueryObserver and setting it on the context.
> QueryContext.setObserver(new QueryObserver()); And no static implementation/use of query observer.
> 
That's actually what I tried to do with this proposal. Am I missing something here?
> I did not see any test changes where concurrent queries are using QueryObserver...It will show if the query observers are cross referenced...

geode-core/src/integrationTest/java/org/apache/geode/cache/query/internal/QueryTraceJUnitTest.java
```diff
@@ -366,7 +396,11 @@ public void testTraceOnLocalRegionWithSmallTracePrefixNoComments() throws Except
     assertTrue(((DefaultQuery) query).isTraced());
 
     SelectResults results = (SelectResults) query.execute();
-    assertTrue(QueryObserverHolder.getInstance() instanceof IndexTrackingQueryObserver);
+
```
mhansonp, line 399: These tests seem awfully repetitive. Is there a reason to check the test hook in every test?  From my perspective it seems like a waste of CPU. Just my opinion though.

geode-core/src/main/java/org/apache/geode/cache/query/internal/QueryObserverHolder.java
mhansonp: It looks to me like you have that covered Alberto. By passing around context, it looks to me like you have addressed many if not all the thread safety issues. I tend to agree with Kirk about preferring and atomic to a synchronize, though.

mhansonp (changes requested):
I think that the use of Atomics is preferable to Synchronize. This takes one extra step elsewhere but encapsulates and manages synchronization challenges. Other than that I think this looks good.

geode-core/src/integrationTest/java/org/apache/geode/cache/query/internal/QueryTraceJUnitTest.java
albertogpz: It is the way to check that the `trace` directive in the query has taken effect when set or that trace is not being done when it is not specified.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/functions/DataCommandFunction.java
```diff
@@ -237,6 +238,9 @@ private DataCommandResult select(InternalCache cache, Object principal, String q
       if (queryObserver != null) {
         QueryObserverHolder.reset();
       }
+      if (tracedQuery.isTraced()) {
```
jinmeiliao, line 241: this block is already inside the `if (tracedQuery.isTraced())` condition, is this `if` necessary?

albertogpz: It is not. This finally block is outside any `if (tracedQuery.isTraced())` condition.

mhansonp (approved):

kirklund:
Maybe the way to go here is to rewrite `QueryObserverHolder` as a class that hold a `ThreadLocal<QueryObserverHolder>` and then have ALL other product classes such as `DefaultQuery` just use `QueryObserverHolder` as an API for accessing that ThreadLocal.

albertogpz:
> Maybe the way to go here is to rewrite `QueryObserverHolder` as a class that hold a `ThreadLocal<QueryObserverHolder>` and then have ALL other product classes such as `DefaultQuery` just use `QueryObserverHolder` as an API for accessing that ThreadLocal.

I like that idea. The problem is that there are test cases that count on setting the observer globally and then running a query from the client. Take a look for example at: `SelectStarQueryDUnitTets::testSelectStarQueryForPartitionedRegion`. In these cases, you cannot set the thread local observer in the thread that will executy the query because the query is executed remotely from the client.

I have been able to change the test cases in `QueryUsingFunctionContextDUnitTest` and set the observer in the function executing the queries instead of globally in the servers but this approach is not valid for the test cases previously mentioned.

I am also worried about the use done by the `QueryObserver` to provide `trace` information in queries. It is not possible to set a query observer and at the same time have trace information.

onichols-pivotal:

albertogpz:
> this PR appears to be abandoned, can it be closed?

It is not yet abandoned. It is pending from a discussion with @kirklund about it.

JinwooHwang:
@albertogpz, please let us know if there’s anything we can do to help move it forward. Thanks.

JinwooHwang:
Hi @kirklund, Just checking in regarding the pending item—there’s been some discussion around it, and I understand a decision may hinge on your input. When you have a moment, could you share your thoughts or let me know if there's anything you'd like to align on before we proceed?
@albertogpz , Do you have any comments to @kirklund ?

semioticrobotic:
@JinwooHwang: I sent @kirklund a note seeking his input here.

JinwooHwang:
Thank you for helping out, @semioticrobotic

PR #6877, opened 2021-09-17
GEODE-8840: Support RENAMENX command

ezoerner:
### For all changes:


geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/key/AbstractRenamenxIntegrationTest.java
```diff
@@ -0,0 +1,385 @@
...
+    assertThat(result).isEqualTo(0L);
+  }
+
+  @Test
+  public void testConcurrentSets() throws ExecutionException, InterruptedException {
+    Set<String> stringsForK1 = new HashSet<>();
+
+    int numOfStrings = 500000;
+    Callable<Long> callable1 =
+        () -> addStringsToKeys(stringsForK1, "{user1}k1", numOfStrings, jedis);
```
DonalEvans, line 135: This method could be reworked a little, since we don't use the contents of `stringsForK1`.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/key/AbstractRenamenxIntegrationTest.java
```diff
@@ -0,0 +1,385 @@
...
+    Callable<Long> callable2 = () -> jedis.renamenx("{user1}k1", "{user1}k2");
+
+    ExecutorService pool = Executors.newFixedThreadPool(3);
+    Future<Long> future1 = pool.submit(callable1);
+    Thread.sleep(rand.nextInt(1000));
```
DonalEvans, line 140: If this sleep is absolutely necessary, it probably shouldn't be for a random amount of time, since a low value will make the test flaky.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/key/AbstractRenamenxIntegrationTest.java
```diff
@@ -0,0 +1,385 @@
...
+    future1.get();
+    try {
+      future2.get();
+      assertThat(jedis.scard("{user1}k2")).isEqualTo(numOfStrings);
+      assertThat(jedis.exists("{user1}k1")).isFalse();
+    } catch (Exception e) {
+      assertThat(e.getMessage()).contains(RedisConstants.ERROR_NO_SUCH_KEY);
+      assertThat(jedis.scard("{user1}k1")).isEqualTo(numOfStrings);
+      assertThat(jedis.exists("{user1}k2")).isFalse();
+    }
```
DonalEvans, line 152: I'm not sure that we should be expecting to see `ERROR_NO_SUCH_KEY` here. It could be avoided entirely by calling `jedis.sadd("{user1}k1", "initialEntry")` at the start of the test to ensure that "{user1}k1" definitely exists before we try to do the rename.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/key/AbstractRenamenxIntegrationTest.java
```diff
@@ -0,0 +1,385 @@
...
+  @Test
+  public void testRenameKeyToExistingKey() {
+    jedis.set("{user1}foo1", "bar1");
+    jedis.set("{user1}foo12", "bar2");
+    long result = jedis.renamenx("{user1}foo1", "{user1}foo12");
+    assertThat(result).isEqualTo(0L);
+  }
+
+  @Test
+  public void testConcurrentSets() throws ExecutionException, InterruptedException {
```
DonalEvans, line 130: This test should probably be using `ConcurrentLoopingThreads` to achieve what it's trying to test (that a rename and a set do not cause problems if they happen simultaneously) rather than just doing one call of sadd and one call of renamenx and hoping that the timing is right.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/key/AbstractRenamenxIntegrationTest.java
```diff
@@ -0,0 +1,385 @@
...
+      assertThat(jedis.exists("{user1}k1")).isFalse();
+    } catch (Exception e) {
+      assertThat(e.getMessage()).contains(RedisConstants.ERROR_NO_SUCH_KEY);
+      assertThat(jedis.scard("{user1}k1")).isEqualTo(numOfStrings);
+      assertThat(jedis.exists("{user1}k2")).isFalse();
+    }
+  }
+
+  @Test
+  public void should_succeed_givenTwoKeysOnDifferentStripes() {
```
DonalEvans, line 156: This test and others testing stripe-related stuff seem to be testing internal behaviour rather than something a user would have access to or knowledge of (there is no concept of a "stripe" in Redis, that's an internal implementation that we're using) so it should probably be a unit test rather than an integration test. Also, this stripe-related behaviour is not unique to the RENAMENX command, so tests around it should probably be in the `SynchronizedStripedCoordinatorTest` class.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/key/AbstractRenamenxIntegrationTest.java
```diff
@@ -0,0 +1,385 @@
...
+
+    for (int i = 0; i < 5; i++) {
+      doConcurrentRenamesDifferentKeys(
+          Arrays.asList(keysOnStripe1.get(0), keysOnStripe2.get(0)),
+          Arrays.asList(keysOnStripe2.get(1), keysOnStripe1.get(1)));
+    }
+  }
+
+  @Test
+  public void shouldThrowError_givenKeyDeletedDuringRename()
```
DonalEvans, line 195: This test should also probably be using `ConcurrentLoopingThreads` rather than the `ExecutorService`. It could use `runWithAction()` and assert that either we see the key renamed successfully or that we hit the `ERROR_NO_SUCH_KEY` error, and only stop looping when we do hit that error. The current test may pass without ever actually seeing the error.

DonalEvans (changes requested):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/key/RenamenxExecutor.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.List;
+
+import org.apache.geode.redis.internal.data.RedisKey;
+import org.apache.geode.redis.internal.data.RedisKeyExistsException;
+import org.apache.geode.redis.internal.executor.AbstractExecutor;
+import org.apache.geode.redis.internal.executor.RedisResponse;
+import org.apache.geode.redis.internal.netty.Command;
+import org.apache.geode.redis.internal.netty.ExecutionHandlerContext;
+
+public class RenamenxExecutor extends AbstractExecutor {
```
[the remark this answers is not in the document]
ezoerner: Not sure if this small amount of code duplication warrants and additional class, but I will go ahead with this change.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/key/AbstractRenamenxIntegrationTest.java
ezoerner: The random sleep just introduces more variability as to when the rename actually occurs, and doesn't affect the reliability of the test.

ezoerner: (i.e. it is not necessary for the test to pass)

geode-apis-compatible-with-redis/src/acceptanceTest/java/org/apache/geode/redis/internal/executor/key/RenameNxNativeRedisAcceptanceTest.java
```diff
@@ -18,7 +18,7 @@
 
 import org.apache.geode.redis.NativeRedisClusterTestRule;
 
-public class RenamenxNativeRedisAcceptanceTest extends AbstractRenamenxIntegrationTest {
+public class RenameNxNativeRedisAcceptanceTest extends AbstractRenameNxIntegrationTest {
```
DonalEvans, line 21: Sorry to be nitpicky, but could this be "RenameNX" with a capital N and capital X? That's the naming pattern that has been used for other commands with NX versions.

ezoerner: Ah yes, of course...

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/key/AbstractRenamenxIntegrationTest.java
ezoerner: The unit tests we have don't use Jedis and use mocked Regions. I'm not sure if setting up a Jedis cluster with real regions is a good idea for unit tests, and I'm also not sure using a mocked region with these concurrency tests for stripes is a good idea either -- maybe these should be a dunit test?

ezoerner: In fact, I see there are already similar dunit tests in `RenameDUnitTest`, so maybe these tests here are redundant anyway?

DonalEvans: Yeah, I think the stripe tests apply in theory to any command that involves multiple keys, so duplicating them for different commands isn't really necessary, as we don't expect to have any different behaviour. It's not ideal that the current tests around stripe behaviour are living in the RENAME test classes, but there's no need to try to address that as part of this PR. Ideally, I think we'd have a test class specifically for stripe-related behaviour that was agnostic about what commands were being used to test it.

ezoerner:
The failure is due to GEODE-7138

DonalEvans (approved):

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/key/AbstractRenameIntegrationTest.java
```diff
@@ -188,38 +244,39 @@ public void shouldNotDeadlock_concurrentRenames_givenStripeContention()
...
+                RuntimeException e = renameException.get();
+                if (e != null) {
+                  throw e;
+                }
+                assertThat(jedis.get("{user1}newKey")).isEqualTo("foo");
+                assertThat(jedis.exists("{user1}oldKey")).isFalse();
+                jedis.set("{user1}oldKey", "foo");
+              });
+    } catch (RuntimeException e) {
+      assertThat(e).hasMessageContaining(ERROR_NO_SUCH_KEY);
```
jdeppe-pivotal, line 274: I'm not sure that catching these exceptions and re-throwing from the `runWithAction` is necessary. This sample never runs the 'action':
```
    new ConcurrentLoopingThreads(1,
        i -> {throw new RuntimeException("BANG!");}
    ).runWithAction(() -> System.out.println("running action"));
```
Regardless, this change also causes the loop to bail out after only 1 or 2 iterations instead of the full 2000.
I do think that the original code should include a `jedis.del("{user1}newKey")` so as to get to a consistent state between loop iterations.

jdeppe-pivotal (changes requested):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisKeyCommandsFunctionExecutor.java
```diff
@@ -135,12 +137,22 @@ public void restore(RedisKey key, long ttl, byte[] data, RestoreOptions options)
...
     return stripedExecute(oldKey, lockOrdering,
-        () -> getRedisData(oldKey).rename(getRegionProvider().getLocalDataRegion(), oldKey,
-            newKey));
+        () -> {
+          RedisData sourceData = getRedisData(oldKey);
+          // nonexistent source takes priority over nonexistent target
+          if (!sourceData.exists()) {
+            return false;
+          }
+          if (ifTargetNotExists && getRedisData(newKey).exists()) {
```
dschneider-pivotal, line 150: I think it would be better if you pushed this logic down into AbstractRedisData.rename by passing it "ifTargetNotExists". Down at that level you can use putIfAbsent or create on the region instead of doing an extra get up at this level. This will also allow you to get rid of the check of sourceData.exists() at this level since you get that for free from the NullRedisData implementation.
I also prefer we keep these *FunctionExecutor classes focused on locking and striped execution.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/key/AbstractRenameIntegrationTest.java
ezoerner: The intention with this test would be that it would exit as successful as soon as the no such key exception was thrown.

I did go back and re-verify that if the exception is not captured and rethrown by the action then you get `expected:<"foo"> but was:<null>`, but you don't if the exception is captured and rethrown. This is hard evidence that the exception is not propagated to the caller.

Also, I ran your code above and got the following output:

```
running action

java.util.concurrent.ExecutionException: java.lang.RuntimeException: BANG!

```

... which shows that the action **IS** called before the exception is propagated to the caller.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/data/RedisKeyCommandsFunctionExecutor.java
ezoerner: We still need to check for the existence of the source key first because that takes priority over the target key.

ezoerner: For some reason the Region interface doesn't support `putIfAbsent` with a callback argument, but using `create` should work.

dschneider-pivotal (approved):

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/key/AbstractRenameIntegrationTest.java
```diff
@@ -188,38 +245,39 @@ public void shouldNotDeadlock_concurrentRenames_givenStripeContention()
...
-        future1.get();
-        assertThat(jedis.get("{user1}newKey")).isEqualTo("foo");
-      } catch (Exception e) {
-        assertThat(e).hasMessageContaining("no such key");
-      }
-      future2.get();
-
-      assertThat(jedis.get("{user1}oldKey")).isNull();
+    try {
+      new ConcurrentLoopingThreads(iterations,
```
jdeppe-pivotal, line 256: This test still ends up only running a single iteration since an exception causes it to end. That wasn't the intent of the original test.

ezoerner: I changed the semantics of the test to do this because of the suggestion from Donal here:
https://github.com/apache/geode/pull/6877#discussion_r714212553

ezoerner: His interpretation of the original test was that the point was to get the error to be thrown and that's it so the test can exit with success as soon as that exception is thrown.   The test does not necessarily only run a single iteration because the rename can occur before the delete in which case the next iteration is executed.

ezoerner: I accepted his interpretation of that and went with it

jdeppe-pivotal: OK, I guess that makes sense - especially given the name of the test method too.

jdeppe-pivotal (approved):

PR #6879, opened 2021-09-17
Revert "GEODE-9495: Limit Thread Sleep In PubSubNativeRedisAcceptance…

ringles:
This reverts commit ed902f6404a1c2b06b980973a7a9d74f045fee8b. Not
working reliably in the pipeline.

### DIFF
--- a/geode-apis-compatible-with-redis/src/acceptanceTest/java/org/apache/geode/redis/internal/executor/pubsub/PubSubNativeRedisAcceptanceTest.java
+++ b/geode-apis-compatible-with-redis/src/acceptanceTest/java/org/apache/geode/redis/internal/executor/pubsub/PubSubNativeRedisAcceptanceTest.java
@@ -15,14 +15,8 @@
 
 package org.apache.geode.redis.internal.executor.pubsub;
 
-import java.io.BufferedReader;
-import java.io.IOException;
-import java.io.InputStreamReader;
-
-import org.apache.commons.lang3.SystemUtils;
 import org.apache.logging.log4j.Logger;
 import org.junit.AfterClass;
-import org.junit.BeforeClass;
 import org.junit.ClassRule;
 
 import org.apache.geode.NativeRedisTestRule;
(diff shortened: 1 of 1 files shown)


DonalEvans (approved):

geode-apis-compatible-with-redis/src/acceptanceTest/java/org/apache/geode/redis/internal/executor/pubsub/PubSubNativeRedisAcceptanceTest.java
```diff
@@ -31,52 +25,20 @@
...
     // need to wait for that to clear up. It shouldn't take more than a minute or so.
-    // For now a thread sleep is the simplest way to wait for the sockets to be out of the TIME_WAIT
-    // state. The default timeout of 240 sec was chosen because that is the default duration for
-    // TIME_WAIT on Windows. The timeouts for both mac and linux are significantly shorter.
-    Thread.sleep(socketTimeWaitMsec);
+    // There will be a better solution for this from GEODE-9495, but for now a thread sleep is the
+    // simplest way to wait for the sockets to be out of the TIME_WAIT state. The timeout of 240 sec
+    // was chosen because that is the default duration for TIME_WAIT on Windows. The timeouts for
+    // both mac and linux are significantly shorter.
+    Thread.sleep(240000);
```
pivotal-jbarrett, line 41: How about configuring the host machines to cycle out of `TIME_WAT` faster and/or increasing the ephemeral port range?

BenjaminPRoss: I think the plan is to reintroduce the changes being reverted in this PR shortly, this commit has broken our acceptance tests so we're just removing it from develop to unblock the pipeline.

dschneider-pivotal (approved):

PR #6880, opened 2021-09-17
GEODE-9607: prevent pubsub from running server out of memory

dschneider-pivotal:
Moved all of the publish code from PubSubImpl to Publisher.
This new class is able to batch the publish requests instead
of always sending them one at a time to other servers.

Also the ExecutorService used for publishing is now correctly configured (thanks to Jens for figuring this out) so that it can use more than one thread for publishing.

Both the publish and subscribe executor now do a critical memory check and will throw a  LowMemoryException

### For all changes:

### DIFF
--- a/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/pubsub/Publisher.java
+++ b/geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/pubsub/Publisher.java
@@ -1,0 +1,330 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ *
+ */
+package org.apache.geode.redis.internal.pubsub;
+
+import static org.apache.geode.logging.internal.executors.LoggingExecutors.newCachedThreadPool;
+import static org.apache.geode.redis.internal.netty.Coder.bytesToString;
+
+import java.io.DataInput;
+import java.io.DataOutput;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.TimeUnit;
+
+import org.apache.logging.log4j.Logger;
+
+import org.apache.geode.DataSerializer;
+import org.apache.geode.annotations.VisibleForTesting;
+import org.apache.geode.cache.execute.FunctionContext;
+import org.apache.geode.cache.execute.FunctionService;
+import org.apache.geode.cache.execute.ResultCollector;
+import org.apache.geode.distributed.DistributedMember;
+import org.apache.geode.internal.cache.execute.InternalFunction;
+import org.apache.geode.internal.serialization.DataSerializableFixedID;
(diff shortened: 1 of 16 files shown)


echobravopapa (approved):

upthewaterspout (approved):

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/pubsub/PublisherTest.java
```diff
@@ -0,0 +1,207 @@
...
+import org.apache.geode.redis.internal.netty.Client;
+
+public class PublisherTest {
+  private final RegionProvider regionProvider = mock(RegionProvider.class);
+  private final Subscriptions subscriptions = mock(Subscriptions.class);
+  private final ExecutorService executorService = createExecutorService();
+  private final Publisher publisher = new Publisher(regionProvider, subscriptions, executorService);
+
+  private ExecutorService createExecutorService() {
+    // This service will do synchronous executioo
```
DonalEvans, line 47: Typo here, should be "execution"

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/pubsub/Publisher.java
```diff
@@ -0,0 +1,330 @@
...
+
+    @Override
+    public void toData(DataOutput out, SerializationContext context) throws IOException {
+      DataSerializer.writeByteArray(getChannel(), out);
+      DataSerializer.writeByteArray(getMessage(), out);
+    }
+
+    @Override
+    public void fromData(DataInput in, DeserializationContext context)
+        throws IOException, ClassNotFoundException {
```
DonalEvans, line 201: `ClassNotFoundException` is never thrown from this method.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/pubsub/Publisher.java
```diff
@@ -0,0 +1,330 @@
...
+    }
+  }
+
+  /**
+   * Manages all the publish requests from a particular client.
+   * The order of these requests is maintained, and they will be
+   * delivered to subscribers in that order.
+   * Since the delivery to subscribers can be slower than the rate
+   * at which publish requests arrive, this publisher supports
+   * batching so allow multiple requests to be sent with a single
```
DonalEvans, line 226: Typo here, I think, should "so allow" be "to allow"?

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/pubsub/Publisher.java
```diff
@@ -0,0 +1,330 @@
...
+        active = false;
+      }
+    }
+
+    private void publishBatch() {
+      executor.execute(() -> {
+        try {
+          internalPublish(batch);
+        } finally {
+          fillBatch();
```
DonalEvans, line 283: In situations with high publish throughput from one client, is it possible that these recursive calls to `fillBatch()` might cause a stack overflow?

DonalEvans (approved):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/pubsub/Publisher.java
dschneider-pivotal: I think this could only happen if the executor is one like PublisherTest has that does not handoff the execute to another thread. But the executor the Publisher uses actually hand it off to another thread (which has its own stack). It could be rewritten to have the current thread loop around and call internalPublish again. That might be better from a context switch point of view (the executor thread does not need to go back to executor's queue but instead goes straight to the ClientPublisher queue) but it does not seem as fair as the current solution. But as far as stack overflow goes the thread will never have more than two frames in fillBatch (because the second one will handoff to another thread if it calls publishBatch). We can talk more about this to clear things up if needed.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/pubsub/Publisher.java
dschneider-pivotal: done

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/pubsub/Publisher.java
dschneider-pivotal: done

geode-apis-compatible-with-redis/src/test/java/org/apache/geode/redis/internal/pubsub/PublisherTest.java
dschneider-pivotal: done

PR #6892, opened 2021-09-22
GEODE-9625: Only serialize transaction metadata when grouping enabled.

pivotal-jbarrett:
### For all changes:


DonalEvans (approved):

geode-core/src/main/java/org/apache/geode/internal/cache/wan/GatewaySenderEventImpl.java
```diff
@@ -1354,4 +1355,20 @@ protected GatewaySenderEventImpl makeCopy() {
   public void setAcked(boolean acked) {
     isAcked = acked;
   }
+
+  public enum TransactionMetadataDisposition {
```
albertogpz, line 1359: You need to update `sanctioned-geode-core-serializables.txt` after adding this enum to `GatewaySenderEventImpl`.

albertogpz (approved):

geode-core/src/main/java/org/apache/geode/internal/cache/wan/GatewaySenderEventImpl.java
```diff
@@ -1354,4 +1355,20 @@ protected GatewaySenderEventImpl makeCopy() {
...
+    Exclude,
+    /**
+     * Transaction metadata should be included in the event.
+     */
+    Include,
+    /**
+     * Transaction metadata should be included in the event and this is the last event in the
+     * transaction.
+     */
+    IncludeLastEvent,
```
gesterzhou, line 1372: This name is misleading. We actually don't have a case not to include the last event. 

When tx id != null, we should always include it. The original name is "isLastEventInTransaction" which makes sense.

Maybe we should change it to "LastEventInTX"?

geode-core/src/main/java/org/apache/geode/internal/cache/wan/GatewaySenderEventImpl.java
```diff
@@ -322,9 +321,11 @@ public GatewaySenderEventImpl(EnumListenerEvent operation, CacheEvent<?, ?> ce,
     }
     isConcurrencyConflict = event.isConcurrencyConflict();
 
-    transactionId = event.getTransactionId();
-    this.isLastEventInTransaction = isLastEventInTransaction;
-
+    if (transactionMetadataDisposition != Exclude) {
```
gesterzhou, line 324: BTW, I did not see toData() and fromData() are changed. Why serialization and replication would save data?

geode-core/src/main/java/org/apache/geode/internal/cache/wan/GatewaySenderEventImpl.java
pivotal-jbarrett: I would be more include to renaming it `IncludeAsLastEvent` implying both the inclusion of the `TransactionId` and the setting of the `isLastEvent` flag.

geode-core/src/main/java/org/apache/geode/internal/cache/wan/GatewaySenderEventImpl.java
pivotal-jbarrett: The disposition is now only set to anything other than `Exclude` by when TX grouping is enabled. The `transactionId` is `null` when `transactionMetadataDisposition` is `Exclude`. In `toData` the `transactionId` part is only serialized if it isn't `null`.

geode-core/src/main/java/org/apache/geode/internal/cache/wan/GatewaySenderEventImpl.java
```diff
@@ -1354,4 +1355,20 @@ protected GatewaySenderEventImpl makeCopy() {
   public void setAcked(boolean acked) {
     isAcked = acked;
   }
+
+  public enum TransactionMetadataDisposition {
+    /**
+     * Transaction metadata should be excluded from the event.
+     */
+    Exclude,
```
dschneider-pivotal, line 1363: I think the geode style guide says constants should be all upper case. The members of an enum are constants. See https://google.github.io/styleguide/javaguide.html#s5.2-specific-identifier-names
I think you should follow this convention for this new enum

geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySenderEventProcessor.java
```diff
@@ -1225,6 +1228,13 @@ public void closeProcessor() {
     }
   }
 
+  protected GatewaySenderEventImpl.TransactionMetadataDisposition getTransactionMetadataDisposition(
+      final boolean isLastEventInTransaction) {
+    return getSender().mustGroupTransactionEvents()
```
dschneider-pivotal, line 1233: I think this would be easier to understand if you changed it to use if statements.

geode-core/src/main/java/org/apache/geode/internal/cache/wan/GatewaySenderEventImpl.java
pivotal-jbarrett: Yes, you are correct. :(

kirklund (approved):
Thanks for all the unit tests too!

PR #6894, opened 2021-09-23
GEODE-9628: Fix setting client credential documentation

nabarunnag:
* The current document mentions that we can choose between setting the properties or the AuthInitializer.
	* After a few experiments we were able to confirm that both the settings are needed for successful client verification.

### For all changes:


geode-docs/managing/security/implementing_authentication.html.md.erb
```diff
@@ -95,7 +95,7 @@ sender or receiver.
 In order to connect with a locator or a server that does authentication,
 a client will need to set its credential, composed of the two properties
 `security-username` and `security-password`.
-Choose one of these two ways to accomplish this:
+Follow these two steps to accomplish this:
```
jinmeiliao, line 98: Actually, only "AuthInitialize" is required for client authentication, depending on the implementation of "AuthtInitialize", if it's looking for these properties in the gemfire properties, then gemfire properties needs to have these properties, but it's solely depends on the implementation of that interface.

davebarnes97 (approved):

geode-docs/managing/security/implementing_authentication.html.md.erb
davebarnes97: @jinmeiliao Perhaps we should add that subtlety to the description. So I know what wording to suggest, would you say that the gemfire properties dependency is the most common use-case, or is it relatively rare?

nabarunnag: @jinmeiliao fair point, I have made the change to reflect that. Previously, it assumed that if the properties are set then there is no need for implementing the AuthInitialize inteface

PR #6895, opened 2021-09-23
GEODE-9624: Fix 'status redundancy' result for empty regions

alb3rtobr:
When a partition redundant region is empty, `status redundancy` command shows a wrong redundancy value for that region.

```
gfsh>status redundancy
Number of regions with zero redundant copies = 0
Number of regions with partially satisfied redundancy = 1
Number of regions with fully satisfied redundancy = 0

Redundancy is partially satisfied for regions: 
  test redundancy status: NOT_SATISFIED. Desired redundancy is 1 and actual redundancy is 2147483647.
```

Problem is in `SerializableRegionRedundancyStatusImpl` class. The `calculateLowestRedundancy` initializes `minRedundancy` variable with `Integer.MAX_VALUE`. But for empty regions, all buckets have redundancy -1, so `minRedundancy` is not modified, and returned with that value.

I propose to check `minRedundancy` value before returning it. If its value is `Integer.MAX_VALUE`, the region is empty so I think the method should return `0`.

Notice that after this change, the redundancy status of an empty region will be reported as `RedundancyStatus.NO_REDUNDANT_COPIES` instead of `RedundancyStatus.NOT_SATISFIED`.

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/control/SerializableRegionRedundancyStatusImpl.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/control/SerializableRegionRedundancyStatusImpl.java
@@ -53,7 +53,7 @@ public class SerializableRegionRedundancyStatusImpl extends
    * @param region The region for which the lowest redundancy should be calculated.
    * @return The redundancy of the least redundant bucket in the region.
    */
-  private int calculateLowestRedundancy(PartitionedRegion region) {
+  int calculateLowestRedundancy(PartitionedRegion region) {
     int numBuckets = region.getPartitionAttributes().getTotalNumBuckets();
     int minRedundancy = Integer.MAX_VALUE;
     for (int i = 0; i < numBuckets; i++) {
@@ -64,7 +64,7 @@ public class SerializableRegionRedundancyStatusImpl extends
         minRedundancy = bucketRedundancy;
       }
     }
-    return minRedundancy;
+    return minRedundancy == Integer.MAX_VALUE ? 0 : minRedundancy;
   }
 
   /**
--- a/geode-gfsh/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/StatusRedundancyCommandDUnitTest.java
+++ b/geode-gfsh/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/StatusRedundancyCommandDUnitTest.java
@@ -76,6 +76,7 @@ public class StatusRedundancyCommandDUnitTest {
   private static final String ZERO_COPIES_REGION = "zeroRedundantCopies";
   private static final List<String> regionNames = Arrays.asList(SATISFIED_REGION,
       UNSATISFIED_REGION, NO_CONFIGURED_REDUNDANCY_REGION, ZERO_COPIES_REGION);
+  private static final String EMPTY_REGION = "emptyRegion";
 
   @Before
   public void setUp() throws Exception {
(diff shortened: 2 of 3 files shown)


geode-core/src/main/java/org/apache/geode/internal/cache/control/SerializableRegionRedundancyStatusImpl.java
```diff
@@ -53,7 +53,7 @@ public SerializableRegionRedundancyStatusImpl(PartitionedRegion region) {
    * @param region The region for which the lowest redundancy should be calculated.
    * @return The redundancy of the least redundant bucket in the region.
    */
-  private int calculateLowestRedundancy(PartitionedRegion region) {
+  int calculateLowestRedundancy(PartitionedRegion region) {
     int numBuckets = region.getPartitionAttributes().getTotalNumBuckets();
     int minRedundancy = Integer.MAX_VALUE;
```
mhansonp, line 58: This seems complicated to me. Why not use 0? If I am wrong that is fine. I am happy to be wrong, but it doesn't seem like you need the MAX_VALUE construct. You are handling the -1 inside the loop.

mhansonp (changes requested):
Logic *seems* a little odd to me in calculateLowestRedundancy method. Happy to be wrong though...

DonalEvans:
Nice catch on this. However, it's not strictly accurate to say that the behaviour seen here is for empty regions, but rather regions for which no buckets have been created yet. If at least one bucket exists for the region, redundancy will be correctly reported, regardless of whether there are any entries in the region. It's a small distinction, but doing a put followed by a remove on one key, then calling `status redundancy` on that region should show that an empty region can correctly report its redundancy.

geode-core/src/main/java/org/apache/geode/internal/cache/control/SerializableRegionRedundancyStatusImpl.java
DonalEvans: I was the original author of this, so I hope Alberto doesn't mind me answering for him. We're calculating the minimum redundancy level for a partitioned region, defined as the minimum redundancy level among all of its buckets, so given that any bucket may have a redundancy value between -1 (bucket does not exist) and N, the only sensible way to find the minimum is to start at the max value and reduce it as we find smaller values. If we started with the minimum redundancy at 0, we'd have no way of keeping track of what the lowest value for redundancy was among all the buckets.

geode-gfsh/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/StatusRedundancyCommandDUnitTest.java
```diff
@@ -195,6 +196,23 @@ public void statusRedundancyWithIncludeRegionReturnsErrorWhenAtLeastOneIncludedR
         Collections.singletonList(includedRegion));
   }
 
+  @Test
+  public void statusRedundancyReturnsEmptyRegionStatusAsNonRedundantCopies() {
```
DonalEvans, line 200: Typo here, should be "StatusAsNoRedundantCopies"

DonalEvans (approved):

alb3rtobr:
> doing a put followed by a remove on one key, then calling `status redundancy` on that region should show that an empty region can correctly report its redundancy.

Good catch, thanks! Im going to rename the test case's names accordingly.

geode-gfsh/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/StatusRedundancyCommandDUnitTest.java
```diff
@@ -197,14 +197,24 @@ public void statusRedundancyWithIncludeRegionReturnsErrorWhenAtLeastOneIncludedR
...
+  public void statusRedundancyReturnsRegionWithNoBucketsCreatedStatusAsNoRedundantCopies() {
     createRegion();
     String command = new CommandStringBuilder(STATUS_REDUNDANCY).getCommandString();
     CommandResultAssert commandResult = gfsh.executeAndAssertThat(command).statusIsSuccess();
     verifyGfshOutput(commandResult, Collections.singletonList(EMPTY_REGION), new ArrayList<>(),
         new ArrayList<>());
   }
 
+  @Test
+  public void statusRedundancyReturnsEmptyRegionStatusAsSatisfied() {
```
DonalEvans, line 209: To be more accurate, this test name should just say that an empty region reports redundancy status correctly, since an empty region will not report as "satisfied" if there aren't enough servers to host that many copies, or if redundancy isn't actually satisfied due to restarted servers etc. Really, what the test is showing is that an empty region behaves no different from a region with entries in it.

alb3rtobr: I agree, good catch

geode-core/src/main/java/org/apache/geode/internal/cache/control/SerializableRegionRedundancyStatusImpl.java
alb3rtobr: thanks for your answer Donal

alb3rtobr: @mhansonp is it ok for you to remove the request for changes?

mhansonp (approved):

geode-core/src/main/java/org/apache/geode/internal/cache/control/SerializableRegionRedundancyStatusImpl.java
mhansonp: @alb3rtobr I have approved your pull request.

PR #6896, opened 2021-09-23
GEODE-9581: Fix possible sending of duplicate events

albertogpz:
When group-transaction-events is set for a serial
gateway sender there is a small chance that duplicated
events are sent by the sender.

The reason behind is a bug in the serial gateway sender.

When the remove() method of the SerialGatewaySenderQueue
was invoked for an event that was peeked to complete
a transaction, it was also removed from the extraPeekedIds
structure.
This provoked that events already peeked, were peeked
again by SerialGatewaySenderQueue.peekAhead() when
they should not have been.

A change has been made in the remove() method so
that events that were peeked to complete a transaction
are only removed from the extraPeekedIds when a call
to remove() has been made previously for that event
(the events is in extraPeekedIdsRemovedButPreviousIdNotRemoved)
and the previous event is the lastDispatchedKey.

### For all changes:


gesterzhou (approved):
I expect to see some test cases

albertogpz:
> I expect to see some test cases


> I expect to see some test cases

I did not write a new test case given that with the fault, there was a test case already failing, although it was not always.
I have added a unit test for this specific case.

alb3rtobr (approved):

PR #6902, opened 2021-09-24
GEODE-9479: Document receivers & regions start order

alb3rtobr:
This PR adds a note on the user guide about the correct order to start regions and gateway receivers to avoid the issue described in [this dev list thread](https://markmail.org/thread/qq32z5hducjoqndz)

### DIFF
--- a/geode-docs/tools_modules/gfsh/command-pages/create.html.md.erb
+++ b/geode-docs/tools_modules/gfsh/command-pages/create.html.md.erb
@@ -368,9 +368,10 @@ unless the `--groups` or `--members` option is specified.
 If the gateway receiver creation succeeds on at least one member,
 this `gfsh` command exits with an exit code indicating success.
 
-
 Outputs a tabular format status of each member's gateway receiver,
 independent of the success or failure of the creation.
+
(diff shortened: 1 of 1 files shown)


davebarnes97 (changes requested):
The correct order of creation is already documented, and in greater detail. The problem is that the section is deeply buried under Topologies and easily overlooked by users.
I propose a two-part solution, of which you have addressed one part:
- add a Note here in the gfsh create command page as you have done. I would augment this with a reference to the Topology write-up and Anil's 3-part ordering description from the dev thread.
- Consider moving the Note to the top of the page, after the list of all create commands, just before "create async-event-queue".

Here is source code for the augmented note, including Anil's text and the cross reference:
```
**Note:** The order in which components are created matters. For example, the recommendation for WAN setup is:

- Create/start WAN Senders first
- Create Regions
- Create/Start WAN receivers last

This assures that when WAN receivers are started, their associated regions are in place. Otherwise,
the `create region` command may fail if events are received before the region exists.
For more on this topic, see [Configuring a Multi-site (WAN) System](../../../topologies_and_comm/multi_site_configuration/setting_up_a_multisite_system.html).
```

alb3rtobr:
Thanks for the suggestion and the tip about the documented order @davebarnes97 
I have added your note to the top of the page.

PR #6908, opened 2021-09-27
GEODE-9650: LOLWUT command

ringles:
This implements a version of the Redis LOLWUT command, which is supposed to (a) algorithmically generate computer art, and (b) output the program version. Optionally, the LOLWUT command can take arguments to affect the output.

This implementation generates arbitrary-size mazes. By default, it generates a 40-cell-wide maze of height 10, but given arguments it can generate a valid maze of any width and height, while consuming a small and fixed amount of memory. After the maze, it outputs the Geode version.

### DIFF
--- a/geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/server/LolWutIntegrationTest.java
+++ b/geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/server/LolWutIntegrationTest.java
@@ -1,0 +1,170 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package org.apache.geode.redis.internal.executor.server;
+
+import static org.apache.geode.distributed.ConfigurationProperties.LOG_LEVEL;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.ClassRule;
+import org.junit.Test;
+import redis.clients.jedis.Jedis;
+import redis.clients.jedis.commands.ProtocolCommand;
+import redis.clients.jedis.util.SafeEncoder;
+
+import org.apache.geode.internal.serialization.KnownVersion;
+import org.apache.geode.redis.GeodeRedisServerRule;
+import org.apache.geode.redis.RedisIntegrationTest;
+import org.apache.geode.redis.internal.netty.Coder;
+import org.apache.geode.test.awaitility.GeodeAwaitility;
+
+public class LolWutIntegrationTest implements RedisIntegrationTest {
+  private static final int REDIS_CLIENT_TIMEOUT =
+      Math.toIntExact(GeodeAwaitility.getTimeout().toMillis());
+  private static final String GEODE_VERSION_STRING = KnownVersion.getCurrentVersion().toString();
+  private static final int DEFAULT_MAZE_HEIGHT = 10 + 2; // Top & bottom walls, version string
+  private static final int MAX_MAZE_HEIGHT = (1024 * 1024) + 2;
+  private Jedis jedis;
+
+  @ClassRule
+  public static GeodeRedisServerRule server = new GeodeRedisServerRule()
+      .withProperty(LOG_LEVEL, "info");
+
+  @Override
+  public int getPort() {
+    return server.getPort();
+  }
+
+  @Before
+  public void setUp() {
+    jedis = new Jedis("localhost", getPort(), REDIS_CLIENT_TIMEOUT);
+  }
+
+  @After
+  public void tearDown() {
+    jedis.flushAll();
+    jedis.close();
+  }
+
+  @Test
+  public void shouldReturnGeodeVersion() {
+    String actualResult = Coder.bytesToString((byte[]) jedis.sendCommand(new ProtocolCommand() {
+      @Override
+      public byte[] getRaw() {
+        return SafeEncoder.encode("lolwut");
+      }
+    }));
+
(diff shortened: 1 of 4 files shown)


jdeppe-pivotal (approved):
Totally amazing! (someone had to say it) Seriously though, this is so cool.

upthewaterspout (approved):

dschneider-pivotal (commented):
Do you have any concerns that this could cause a geode server to run out of memory? How much memory will it use if width and height are MAX_INT? Could we have a fixed max width and height and if numbers are given bigger than that then we ignore them and use our fixed value?

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/server/LolWutExecutor.java
```diff
@@ -0,0 +1,167 @@
...
+  public RedisResponse executeCommand(Command command,
+      ExecutionHandlerContext context) {
+
+    long inputWidth = -1;
+    long inputHeight = -1;
+
+    List<byte[]> commands = command.getProcessedCommand();
+    if (commands.size() > 1) {
+      for (int i = 1; i < commands.size(); i++) {
+        if (Coder.bytesToString(commands.get(i)).equalsIgnoreCase("version")) {
```
DonalEvans, line 48: Could this instead use the pattern seen elsewhere in executors and compare the argument byte array to a constant defined in `StringBytesGlossary` using the `Coder.equalsIgnoreCaseBytes()` method?

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/server/LolWutExecutor.java
```diff
@@ -0,0 +1,167 @@
...
+          i += 1; // skip next arg, we only have one version for now
+        } else {
+          try {
+            if (inputWidth < 0) {
+              inputWidth = Coder.bytesToLong(commands.get(i));
```
DonalEvans, line 53: Rather than defining this as a long then casting it to an int, which could cause overflow errors, it should be defined as an in from the start and the `Coder.narrowLongToInt()` method should be used here and for `inputHeight`.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/server/LolWutExecutor.java
```diff
@@ -0,0 +1,167 @@
...
+    }
+    if (inputWidth >= 0) {
+      width = (int) inputWidth;
+    }
+
+    return RedisResponse.bulkString(makeArbitraryHeightMaze());
+  }
+
+  // Adapted from code here: https://tromp.github.io/maze.html
+  public static String makeArbitraryHeightMaze() {
```
DonalEvans, line 76: This method should probably take a `height` and `width` argument rather than relying on the static fields in this class, or it should be renamed, as otherwise, calling it from outside the class will not result in an arbitrary sized maze, but rather one of a fixed size based on `DEFAULT_WIDTH` and `DEFAULT_HEIGHT`.

DonalEvans (changes requested):
This is really neat! Since LOLWUT also returns the version of the Redis server, could we get a test to validate just that part of the output? Also that the command doesn't return an error if the user doesn't provide any arguments, or if they provide the "version" argument, and that expected errors are returned for non-integer arguments.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/server/LolWutExecutor.java
ringles: Yeah, that's a better approach for sure.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/server/LolWutExecutor.java
ringles: Actually, it _should_ be a private method. Originally the sub-methods took width and height arguments but it made the arg list a bit unwieldy.

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/server/LolWutExecutor.java
ringles: Will do.

jinmeiliao (approved):

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/server/LolWutIntegrationTest.java
```diff
@@ -0,0 +1,170 @@
...
+
+import org.apache.geode.internal.serialization.KnownVersion;
+import org.apache.geode.redis.GeodeRedisServerRule;
+import org.apache.geode.redis.RedisIntegrationTest;
+import org.apache.geode.redis.internal.netty.Coder;
+import org.apache.geode.test.awaitility.GeodeAwaitility;
+
+public class LolWutIntegrationTest implements RedisIntegrationTest {
+  private static final int REDIS_CLIENT_TIMEOUT =
+      Math.toIntExact(GeodeAwaitility.getTimeout().toMillis());
```
DonalEvans, line 38: This can be removed, and the constant from `RedisClusterStartupRule` with the same name can be used instead.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/server/LolWutIntegrationTest.java
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  @Test
+  public void shouldReturnGeodeVersion() {
+    String actualResult = Coder.bytesToString((byte[]) jedis.sendCommand(new ProtocolCommand() {
+      @Override
+      public byte[] getRaw() {
+        return SafeEncoder.encode("lolwut");
+      }
+    }));
```
DonalEvans, line 71: This can be simplified to:
```
    String actualResult = Coder.bytesToString((byte[]) jedis.sendCommand(
        () -> SafeEncoder.encode("lolwut")));
```
Similarly, other uses of `new ProtocolCommand` in this class can be replaced with lambdas.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/server/LolWutIntegrationTest.java
ringles: I'm gonna make a PR that gets rid of all the old ones. Too easy to miss.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/server/LolWutIntegrationTest.java
ringles: Done. Was working from old code before.

DonalEvans (approved):

PR #6914, opened 2021-09-29
GEODE-8200: After checking locator presence store the status in OperationStateStore

agingade:
After checking locator presence store the status in OperationStateStore

### For all changes:

- [Y] Is your initial contribution a single, squashed commit?

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/management/internal/operation/OperationHistoryManager.java
+++ b/geode-core/src/main/java/org/apache/geode/management/internal/operation/OperationHistoryManager.java
@@ -106,8 +106,7 @@ public class OperationHistoryManager {
   }
 
   private boolean isLocatorOffline(OperationState operationState) {
-    return operationState.getOperationEnd() == null
-        && (operationState.getLocator() != null)
+    return (operationState.getLocator() != null)
         && !cache.getMyId().toString().equals(operationState.getLocator())
         && (!cache.getDistributedSystem().getAllOtherMembers().stream().map(Object::toString)
             .collect(Collectors.toSet()).contains(operationState.getLocator()));
--- a/geode-core/src/main/java/org/apache/geode/management/internal/operation/OperationHistoryManager.java
+++ b/geode-core/src/main/java/org/apache/geode/management/internal/operation/OperationHistoryManager.java
@@ -73,7 +73,7 @@ public class OperationHistoryManager {
     final long expirationTime = now() - keepCompletedMillis;
     Set<String> expiredKeys = operationStateStore.list()
         .stream()
-        .map(this::validateLocator)
+        .map(this::validateLocatorAndUpdateOperationState)
         .filter(operationInstance -> isExpired(expirationTime, operationInstance))
         .map(OperationState::getId)
         .collect(Collectors.toSet());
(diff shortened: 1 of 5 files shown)


mhansonp (approved):

geode-core/src/main/java/org/apache/geode/management/internal/operation/OperationHistoryManager.java
```diff
@@ -96,10 +96,15 @@ private static boolean isExpired(long expirationTime, OperationState<?, ?> opera
...
 
   private OperationState<?, ?> validateLocator(OperationState<?, ?> operationState) {
+    if (operationState.getOperationEnd() != null) {
+      return operationState;
+    }
     if (isLocatorOffline(operationState)) {
       operationState.setOperationEnd(new Date(), null,
           new RuntimeException("Locator that initiated the Rest API operation is offline: "
               + operationState.getLocator()));
+      operationStateStore.recordEnd(operationState.getId(), operationState.getResult(),
```
dschneider-pivotal, line 106: I think you should call "recordEnd" here instead of "operationStateStore.recordEnd". Also it looks like recordEnd will in turn call operationState.setOperationEnd. So I think this whole if block could just be this one call: recordEnd(new Date(), null,
          new RuntimeException("Locator that initiated the Rest API operation is offline: "
              + operationState.getLocator()));

dschneider-pivotal (changes requested):

jinmeiliao (approved):
LGTM after putting in Darrel's suggestion.

geode-core/src/main/java/org/apache/geode/management/internal/operation/OperationHistoryManager.java
agingade: Changes are made as suggested. And the method name is changed to reflect the value is recorded.

geode-core/src/main/java/org/apache/geode/management/internal/operation/OperationHistoryManager.java
```diff
@@ -95,13 +95,13 @@ private static boolean isExpired(long expirationTime, OperationState<?, ?> opera
     return operationEnd.getTime() <= expirationTime;
   }
 
-  private OperationState<?, ?> validateLocator(OperationState<?, ?> operationState) {
-    if (isLocatorOffline(operationState)) {
-      operationState.setOperationEnd(new Date(), null,
+  private OperationState<?, ?> validateLocatorAndUpdateOperationState(
+      OperationState<?, ?> operationState) {
+    if (operationState.getOperationEnd() == null && isLocatorOffline(operationState)) {
```
dschneider-pivotal, line 100: get rid of the "getOperationEnd() == null" check here. It is already in isLocatorOffline.

dschneider-pivotal (changes requested):

geode-core/src/main/java/org/apache/geode/management/internal/operation/OperationHistoryManager.java
agingade: Good catch @dschneider-pivotal 
Removed the "operationState.getOperationEnd() == null" in isLocatorOffline()
Thanks.

dschneider-pivotal (approved):

PR #6919, opened 2021-09-29
<DO NOT REVIEW> Serialization filters

jchen21:
### For all changes:


geode-serialization/src/main/java/org/apache/geode/internal/serialization/filter/SanctionedSerializablesFilterPattern.java
```diff
@@ -108,6 +108,22 @@ private static StringJoiner dependenciesPattern() {
...
+        .add("org.apache.geode.internal.cache.CompactOfflineDiskStoreDUnitTest*")
+        .add("org.apache.geode.internal.cache.ParallelDiskStoreRecoveryDUnitTest*")
+        .add("org.apache.geode.internal.cache.PartitionedRegionSingleHopDUnitTest*")
+        .add("org.apache.geode.internal.cache.BucketServerLocation66")
+        .add("org.apache.geode.internal.cache.ValidateOfflineDiskStoreDUnitTest*")
+        .add("org.apache.geode.logging.internal.LoggingWithReconnectDistributedTest*")
+        .add("org.apache.geode.distributed.HostedLocatorsDUnitTest*")
+        .add("org.apache.geode.management.internal.cli.commands.ShutdownCommandDUnitTest*")
+        .add("org.apache.geode.management.internal.cli.commands.ShutdownCommandOverHttpDUnitTest*")
+        .add("org.apache.geode.test.dunit.**");
```
kirklund, line 127: We need to figure out a different solution for this. We shouldn't add dunit tests to the actual product pattern for sanctioned serializables.

I suspect that these tests may be starting a LocatorLauncher which then rejects deserializing classes in geode-core/src/distributedTest.

Tests such as `CompactOfflineDiskStoreDUnitTest` should instead do one of the following:
1. set the global serial filter to `"*"` in the VM that invokes `LocatorLauncher.start`
2. change to setting a global serial filter by passing -Djdk.serialFilter to a new Locator JVM from StartLocatorCommand instead of configuring it from LocatorLauncher.start
3. add an undocumented System Property to skip the configuring of serial filter in LocatorLauncher.start and set that in ChildVM.main
4. use `serializable-object-filter` to add the test and all dunit rules as user classes to the Locator's global serial filter (which I think we currently have commented out in LocatorLauncher)

I think [1] is the better option. We have plenty of tests so we don't really need these random dunit tests doing anything with the global serial filter. What do you think?

geode-core/src/integrationTest/java/org/apache/geode/internal/io/ServerJdkGlobalSerialFilterIntegrationTest.java
```diff
@@ -64,6 +65,8 @@ public void setUp() throws IOException {
   }
 
   @Test
+  @Ignore
+  // jdk.serialFilter does not apply to ServerLauncher
```
kirklund, line 69: This test may not be finished.

ServerLauncher should NOT configure global serial filter on Java 8 or Java 9.

Server JVM should however configure the jdk serial filter if it invokes `ManagementAgent.startAgent()`. The jdk serial filter work is fully implemented and merged to develop, but some of that code has been refactored so that the jdk serial filter and global serial filter share some classes and use similar internal APIs.

geode-core/src/integrationTest/java/org/apache/geode/internal/io/LocatorSerializableObjectFilterIntegrationTest.java
```diff
@@ -195,7 +195,7 @@ public void configuresValidateSerializableObjects_onJava8() {
         .isTrue();
     assertThat(isValidateSerializableObjectsConfigured())
         .as(VALIDATE_SERIALIZABLE_OBJECTS)
-        .isTrue();
+        .isTrue(); // ??? why not false
```
kirklund, line 198: `LocatorLauncher` should configure BOTH global serial filter AND `validate-serializable-objects` when `start` is invoked and the JVM is Java 8.

I bet my code doesn't yet configure `validate-serializable-objects`.

kirklund: It looks like `assumeThat(isJavaVersionAtLeast(JAVA_9)).isTrue();` on the first line of this test (`configuresValidateSerializableObjects_onJava8`) is wrong... it should be `isJavaVersionAtMost(JAVA_8)`.

Any tests that result in setting the global serial filter in a VM will need to either bounce that VM after the test or live in its own test class so that the JVM does away.

PR #6922, opened 2021-09-30
GEODE-9658: refactor test AuthInit to not use static field for creden…

jinmeiliao:
…tials.

Sometimes the test would have authentication failure that looks like the credentials are not updated as it should. Refactor this test utility to not use static field. Static variables are indeed shared between threads, but the changes made in one thread may not be visible to another thread immediately, making it seem like there are two copies of the variable.

### DIFF
--- a/geode-junit/src/main/java/org/apache/geode/security/UpdatableUserAuthInitialize.java
+++ b/geode-junit/src/main/java/org/apache/geode/security/UpdatableUserAuthInitialize.java
@@ -16,8 +16,8 @@
 package org.apache.geode.security;
 
 import java.util.Properties;
-import java.util.concurrent.atomic.AtomicReference;
 
+import org.apache.geode.LogWriter;
 import org.apache.geode.distributed.DistributedMember;
 
 /**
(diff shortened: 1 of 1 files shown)


geode-junit/src/main/java/org/apache/geode/security/UpdatableUserAuthInitialize.java
```diff
@@ -28,45 +28,49 @@
...
+  }
 
   @Override
   public Properties getCredentials(Properties securityProps, DistributedMember server,
       boolean isPeer) throws AuthenticationFailedException {
     Properties credentials = new Properties();
-    credentials.put("security-username", user.get());
-    credentials.put("security-password", user.get());
+    credentials.put("security-username", user);
+    credentials.put("security-password", user);
```
kirklund, line 50: While you're modifying this, you should change from `put` (Map) to `setProperty` (Properties). Both work, but `setProperty` is more correct for Properties.

kirklund (approved):

PR #6924, opened 2021-09-30
GEODE-9645: Don't notify servers of new DataSerializers

mhansonp:
on the client if in multi user authentication mode.


geode-core/src/main/java/org/apache/geode/cache/client/internal/EndpointManagerImpl.java
```diff
@@ -181,6 +182,11 @@ public void removeListener(EndpointManager.EndpointListener listener) {
     this.listener.removeListener(listener);
   }
 
+  @TestOnly
```
kirklund, line 185: Do we have guidelines on when to use `@TestOnly` versus `@VisibleForTesting`?

geode-core/src/test/java/org/apache/geode/internal/cache/PoolManagerTest.java
```diff
@@ -37,6 +65,112 @@ public void setUp() {
     assertThat(poolManager.getMap()).isEmpty();
   }
 
+  // in Mulituser Auth mode
+  @Test
+  public void Test_thatDataSerializerSynchronizationMessagesAreNotSent() {
```
kirklund, line 70: You should use camel case. The method names should start with lowercase unless there's some unusual reason why it helps with organization and understanding of a class.

geode-core/src/main/java/org/apache/geode/internal/cache/PoolManagerImpl.java
```diff
@@ -63,6 +64,11 @@ public static PoolManagerImpl getPMI() {
     return result;
   }
 
+  @TestOnly
+  public static void setImpl(PoolManagerImpl poolManager) {
+    impl = poolManager;
+  }
+
```
kirklund, line 71: I think we're going in the wrong direction with some of these TestOnly setters, especially when the constant or variable is annotated with `@MakeNotStatic` and it needs to have `final` removed to make it work.

geode-core/src/main/java/org/apache/geode/cache/client/internal/EndpointManagerImpl.java
mhansonp: No, but I don't want these API to be used at all outside of teset, not just to be private.

geode-core/src/main/java/org/apache/geode/internal/cache/PoolManagerImpl.java
mhansonp: I agree, I would love to work with you to make these tests better.

geode-core/src/main/java/org/apache/geode/cache/client/internal/EndpointManagerImpl.java
mhansonp: Removed unit tests.

DonalEvans (changes requested):
This seems like a potential departure from what the user might expect the behaviour to be. What are the implications of this change in terms of what the user sees happen when they register a DataSerializer? Also, there is no test coverage added as part of this PR, so it should probably be moved to draft mode until it's in a finished state.

mhansonp:
> This seems like a potential departure from what the user might expect the behaviour to be. What are the implications of this change in terms of what the user sees happen when they register a DataSerializer? Also, there is no test coverage added as part of this PR, so it should probably be moved to draft mode until it's in a finished state.

The behavior is correct. It was approved as there is no good solution when using multi-user auth. With regards to the testing, the JUnit testing I had proved to impact other seemingly unrelated tests. So I pulled them. The space where this code needs to work is full of global singletons.

PR #6925, opened 2021-10-01
GEODE-9587: Support MSETNX command

ezoerner:
### For all changes:

### DIFF
--- a/geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/string/MSetNXDUnitTest.java
+++ b/geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/string/MSetNXDUnitTest.java
@@ -23,7 +23,6 @@ import java.util.List;
 import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicBoolean;
 
-import org.apache.logging.log4j.Logger;
 import org.junit.After;
 import org.junit.AfterClass;
 import org.junit.BeforeClass;
@@ -31,21 +30,17 @@ import org.junit.ClassRule;
 import org.junit.Ignore;
 import org.junit.Test;
 import redis.clients.jedis.HostAndPort;
-import redis.clients.jedis.Jedis;
 import redis.clients.jedis.JedisCluster;
 
 import org.apache.geode.cache.control.RebalanceFactory;
 import org.apache.geode.cache.control.ResourceManager;
-import org.apache.geode.logging.internal.log4j.api.LogService;
 import org.apache.geode.redis.ConcurrentLoopingThreads;
 import org.apache.geode.test.dunit.rules.ClusterStartupRule;
 import org.apache.geode.test.dunit.rules.MemberVM;
 import org.apache.geode.test.dunit.rules.RedisClusterStartupRule;
 import org.apache.geode.test.junit.rules.ExecutorServiceRule;
 
-public class MSetnxDUnitTest {
-
-  private static final Logger logger = LogService.getLogger();
+public class MSetNXDUnitTest {
 
   @ClassRule
   public static RedisClusterStartupRule clusterStartUp = new RedisClusterStartupRule(4);
@@ -69,11 +64,7 @@ public class MSetnxDUnitTest {
     clusterStartUp.startRedisVM(3, locatorPort);
 
     int redisServerPort1 = clusterStartUp.getRedisPort(1);
-    // Ensure that buckets are created using a connection with a fairly high timeout since
-    // clusterNodes does not get retried.
-    new Jedis(BIND_ADDRESS, redisServerPort1, REDIS_CLIENT_TIMEOUT).clusterNodes();
-
-    jedis = new JedisCluster(new HostAndPort(BIND_ADDRESS, redisServerPort1), 5000, 20);
+    jedis = new JedisCluster(new HostAndPort(BIND_ADDRESS, redisServerPort1), REDIS_CLIENT_TIMEOUT);
   }
 
   @After
--- a/geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/string/MSetnxDUnitTest.java
+++ b/geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/string/MSetnxDUnitTest.java
@@ -1,0 +1,183 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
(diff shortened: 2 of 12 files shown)


geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/string/MSetnxDUnitTest.java
```diff
@@ -0,0 +1,183 @@
...
+import org.apache.geode.logging.internal.log4j.api.LogService;
+import org.apache.geode.redis.ConcurrentLoopingThreads;
+import org.apache.geode.test.dunit.rules.ClusterStartupRule;
+import org.apache.geode.test.dunit.rules.MemberVM;
+import org.apache.geode.test.dunit.rules.RedisClusterStartupRule;
+import org.apache.geode.test.junit.rules.ExecutorServiceRule;
+
+public class MSetnxDUnitTest {
+
+  private static final Logger logger = LogService.getLogger();
```
DonalEvans, line 48: This is never used in this class.

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/string/MSetnxDUnitTest.java
```diff
@@ -0,0 +1,183 @@
...
+    locator = clusterStartUp.startLocatorVM(0);
+    locatorPort = locator.getPort();
+    server1 = clusterStartUp.startRedisVM(1, locatorPort);
+    clusterStartUp.startRedisVM(2, locatorPort);
+    clusterStartUp.startRedisVM(3, locatorPort);
+
+    int redisServerPort1 = clusterStartUp.getRedisPort(1);
+    // Ensure that buckets are created using a connection with a fairly high timeout since
+    // clusterNodes does not get retried.
+    new Jedis(BIND_ADDRESS, redisServerPort1, REDIS_CLIENT_TIMEOUT).clusterNodes();
```
DonalEvans, line 74: I'm not sure I understand why this is needed. The test runs fine locally for me without this line.

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/string/MSetnxDUnitTest.java
```diff
@@ -0,0 +1,183 @@
...
+
+    jedis = new JedisCluster(new HostAndPort(BIND_ADDRESS, redisServerPort1), 5000, 20);
```
DonalEvans, line 76: This should be using `REDIS_CLIENT_TIMEOUT` rather than 5000, and unless there is a specific reason to use max attempts of 20, that should probably be removed.

DonalEvans (changes requested):

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/string/AbstractMSetExecutor.java
```diff
@@ -0,0 +1,70 @@
...
+
+    int numElements = commandElems.size() / 2;
+    List<RedisKey> keys = new ArrayList<>(numElements);
+    List<byte[]> values = new ArrayList<>(numElements);
+
+    RedisKey previousKey = null;
+    for (int i = 0; i < commandElems.size(); i += 2) {
+      RedisKey key = new RedisKey(commandElems.get(i));
+
+      if (previousKey != null && key.getBucketId() != previousKey.getBucketId()) {
```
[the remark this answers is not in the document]
ezoerner: I was discussing this with Jens as well. There is no technical reason for Geode to require that the keys be all hashed to the same slot, which is why there is just the bucketId check. I'm not sure its entirely obvious that we should throw an error here if they are not in the same slot just because Redis does. Is it a product mandate that we always return an error when Redis does?

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/string/MSetnxDUnitTest.java
ezoerner: This was just copied from `MSetDUnitTest`, and I'm not clear why this is needed either

geode-apis-compatible-with-redis/src/main/java/org/apache/geode/redis/internal/executor/string/AbstractMSetExecutor.java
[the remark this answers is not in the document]
ezoerner: BTW this same question also applies to MSET, RENAME, and RENAMENX.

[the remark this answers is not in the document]
DonalEvans: The approach we've been taking so far is to always match native Redis' behaviour (as seen by the user, at least) unless we're unable to do so for some reason. An example is how Redis handles integer value inputs from users, where they reject any values starting with + or the value -0 and return an error. We could in theory parse such values with no problem and not return an error, but the decision was made to match Redis' behaviour in this case, despite the fact that we could quite easily accept a wider range of inputs. (See `Coder.parseLong()`)

geode-apis-compatible-with-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/string/MSetNXDUnitTest.java
```diff
@@ -0,0 +1,174 @@
...
+    }
+    String[] keysAndValues1 = makeKeysAndValues(keys, "valueOne");
+    String[] keysAndValues2 = makeKeysAndValues(keys, "valueTwo");
+    AtomicBoolean running = new AtomicBoolean(true);
+
+    Future<?> future = executor.submit(() -> {
+      for (int i = 0; i < 20 && running.get(); i++) {
+        clusterStartUp.moveBucketForKey(keys[0], "server-3");
+        // Sleep for a bit so that MSETs can execute
+        Thread.sleep(2000);
```
kirklund, line 122: Another option that can be a bit cleaner and more efficient is to introduce a `CountDownLatch` above this block. All of these threads would then await the latch:
```
CountDownLatch latch = new CountDownLatch(1);
```
```
Future<?> future = executor.submit(() -> {
      for (int i = 0; i < 20 && running.get(); i++) {
        clusterStartUp.moveBucketForKey(keys[0], "server-3");
        // Sleep for a bit so that MSETs can execute
        latch.await(GeodeAwaitility.getTimeout().getSeconds(), TimeUnit.SECONDS);
```
And then later on where you're executing MSETs or whatever you want to sleep a bit for, you open that latch:
```
latch.countDown();
```
If there are _n_ threads concurrently doing something you want to pause for, then use `new CountDownLatch(n)` and have each of those threads `countDown` when you reach the mid-point or whatever you're pausing for.

ezoerner: Yeah, I was looking sideways at that sleep also, but in the interest in getting this ticket out of the critical path, I'm inclined to leave this for now. If you think this is important enough that it should be changed, can you open a ticket for it? Thanks.

DonalEvans (approved):

jdeppe-pivotal (approved):

PR #6927, opened 2021-10-01
GEODE-9663: throw and handle AuthenticationExpiredException at login time

jinmeiliao:
There is a time gap between credentials are gathered at the client and credentials are examined on the server, if between this time period, credentials expired (rare case, but it does happen a lot during stress tests), client operations will fail during the operations (even after user is authenticated, doing a put/get might need to open new connections and require authentication again). So we should allow "authenticate" method also throw AuthExpirationException if they want to give client a chance to retry (only retry once).

The PR handles retry from the client side if client receives an `AuthenticationExpiredException`. Older client won't receive that exception.


geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationMultiServerDUnitTest.java
```diff
@@ -253,6 +254,53 @@ public void registerInterestsWithMultiServers() throws Exception {
...
 
+  @Test
+  public void consecutivePut() throws Exception {
+    int locatorPort = locator.getPort();
+    // do consecutive puts using a client
+    ClientVM client = cluster.startClientVM(3,
+        c -> c.withProperty(SECURITY_CLIENT_AUTH_INIT, UpdatableUserAuthInitialize.class.getName())
+            .withCacheSetup(ccf -> ccf.setPoolMaxConnections(2))
+            .withLocatorConnection(locatorPort));
+    AsyncInvocation invokePut = client.invokeAsync(() -> {
```
kirklund, line 265: `AsyncInvocation<Void>`

geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationMultiServerDUnitTest.java
```diff
@@ -253,6 +254,53 @@ public void registerInterestsWithMultiServers() throws Exception {
...
+          ClusterStartupRule.getClientCache()
+              .createClientRegionFactory(ClientRegionShortcut.CACHING_PROXY)
+              .create(PARTITION_REGION);
+      IntStream.range(0, 1000).forEach(i -> proxyRegion.put("key" + i, "value" + i));
+    });
+
+    client.invoke(() -> {
+      // wait till at least 1/3 of the data is in the region to expire the user
+      await().until(() -> getProxyRegion() != null);
+      await().until(() -> getProxyRegion().size() > 10);
```
kirklund, line 277: Anytime you're using await for a value that can be more than two possible values, I would use an assertion so that the value can be seen if it ever fails.
```
await().untilAsserted(() -> assertThat(getProxyRegion().size()).isGreaterThan(10));
```

geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationMultiServerDUnitTest.java
```diff
@@ -253,6 +254,53 @@ public void registerInterestsWithMultiServers() throws Exception {
...
+
+    ExpirableSecurityManager securityManager = collectSecurityManagers(server1, server2);
+    Map<String, List<String>> authorizedOps = securityManager.getAuthorizedOps();
+    if (authorizedOps.size() == 1) {
+      // in case user1 has finished putting all 1000 values in the region before we can expire it.
+      return;
+    }
+    assertThat(authorizedOps).hasSize(2);
+    assertThat(authorizedOps.get("user1").size() + authorizedOps.get("user2").size())
+        .isEqualTo(1000);
```
kirklund, line 292: I recommend adding an `as` into this one so that if it fails, the failure message shows the size for both users:
```
assertThat(authorizedOps.get("user1").size() + authorizedOps.get("user2").size())
    .as(String.format("Combined sizes of user1 %s and user2 %s", 
        authorizedOps.get("user1").size(),
        authorizedOps.get("user2").size()))
    .isEqualTo(1000);
```

kirklund (approved):
Inlined a couple recommended changes to improve failure messages.

geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationMultiServerDUnitTest.java
```diff
@@ -253,6 +254,53 @@ public void registerInterestsWithMultiServers() throws Exception {
...
+      UpdatableUserAuthInitialize.setUser("user1");
+      Region<Object, Object> proxyRegion =
+          ClusterStartupRule.getClientCache()
+              .createClientRegionFactory(ClientRegionShortcut.CACHING_PROXY)
+              .create(PARTITION_REGION);
+      IntStream.range(0, 1000).forEach(i -> proxyRegion.put("key" + i, "value" + i));
+    });
+
+    client.invoke(() -> {
+      // wait till at least 1/3 of the data is in the region to expire the user
```
DonalEvans, line 275: This comment seems inaccurate, since we're putting 1000 entries into the region but only waiting for 10 of them before expiring the user.

geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationMultiServerDUnitTest.java
```diff
@@ -253,6 +254,53 @@ public void registerInterestsWithMultiServers() throws Exception {
...
+
+    expireUserOnAllVms("user1");
+    invokePut.await();
+
+    ExpirableSecurityManager securityManager = collectSecurityManagers(server1, server2);
+    Map<String, List<String>> authorizedOps = securityManager.getAuthorizedOps();
+    if (authorizedOps.size() == 1) {
+      // in case user1 has finished putting all 1000 values in the region before we can expire it.
+      return;
+    }
```
DonalEvans, line 289: This case could potentially be avoided by having the client do put ops in a while loop, with some latch that can be toggled to exit the loop once we've finished with the `expireUserOnAllVms()` method. Without guaranteeing that the user was expired before finishing the puts, we can't guarantee that this test is actually testing anything.

geode-core/src/test/java/org/apache/geode/cache/client/internal/AuthenticateUserOpTest.java
```diff
@@ -82,4 +92,54 @@ public void constructorWithProperties() throws Exception {
...
+
+  @Test
+  public void noAttempt_if_NotRequireCredentials() throws Exception {
+    when(server.getRequiresCredentials()).thenReturn(false);
+    impl.attempt(connection);
+    verify(impl, never()).parentAttempt(connection);
+  }
+
+  @Test
+  public void callPrentAttempt_IfRequireCredentials() throws Exception {
```
mhansonp, line 104: spelling parent

mhansonp (approved):
minor comments and +1s

geode-core/src/main/java/org/apache/geode/cache/client/internal/AuthenticateUserOp.java
```diff
@@ -176,13 +177,29 @@ protected InternalDistributedSystem getConnectedSystem() {
...
+        getMessage().clear();
+        try {
+          return parentAttempt(connection);
+        } catch (AuthenticationExpiredException second) {
+          throw new AuthenticationFailedException(second.getMessage(), second);
+        }
+      }
+    }
+
+    Object parentAttempt(Connection connection) throws Exception {
```
jchen21, line 198: `@VisibleForTesting`

geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationMultiServerDUnitTest.java
```diff
@@ -253,6 +254,53 @@ public void registerInterestsWithMultiServers() throws Exception {
...
+    int locatorPort = locator.getPort();
+    // do consecutive puts using a client
+    ClientVM client = cluster.startClientVM(3,
+        c -> c.withProperty(SECURITY_CLIENT_AUTH_INIT, UpdatableUserAuthInitialize.class.getName())
+            .withCacheSetup(ccf -> ccf.setPoolMaxConnections(2))
+            .withLocatorConnection(locatorPort));
+    AsyncInvocation invokePut = client.invokeAsync(() -> {
+      UpdatableUserAuthInitialize.setUser("user1");
+      Region<Object, Object> proxyRegion =
+          ClusterStartupRule.getClientCache()
```
jmelchio, line 268: `Objects.requireNonNull(ClusterStartupRule.getClientCache())` reduces warnings

geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationMultiServerDUnitTest.java
```diff
@@ -253,6 +254,53 @@ public void registerInterestsWithMultiServers() throws Exception {
...
+        .isEqualTo(1000);
+    Map<String, List<String>> unAuthorizedOps = securityManager.getUnAuthorizedOps();
+    assertThat(unAuthorizedOps).hasSize(1);
+    // user1 may not be unauthorized for just 1 operations, puts maybe done by different
+    // connections
+    assertThat(unAuthorizedOps.get("user1")).isNotEmpty();
+  }
+
+  private static Region<Object, Object> getProxyRegion() {
+    return ClusterStartupRule.getClientCache().getRegion(PARTITION_REGION);
```
jmelchio, line 301: `return Objects.requireNonNull(ClusterStartupRule.getClientCache()).getRegion(PARTITION_REGION);` reduces warnings

jmelchio (approved):
LGTM, just some minor suggestions.

dschneider-pivotal (approved):

geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationMultiServerDUnitTest.java
jinmeiliao: Just my personal preference: I feel like I rather live with the warning than clutter up the code.

DonalEvans: Just my 2 cents, but this warning shows up all over the place, so I wonder if it might be better to just make `ClusterStartupRule.getCache()` and `ClusterStartupRule.getClientCache()` throw an NPE if they're called with a null cache, since that's what would happen anyway.

jdeppe-pivotal (approved):

echobravopapa (approved):

PR #6933, opened 2021-10-04
GEODE-9623: Unsupported commands should not be returned by COMMAND when not enabled

jdeppe-pivotal:
- If the `enable-unsupported-commands` system property is not enabled
  then the COMMAND command should not return information about
  unsupported commands in its response.

### For all changes:

### DIFF
--- a/geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/server/CommandIntegrationTest.java
+++ b/geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/server/CommandIntegrationTest.java
@@ -84,19 +84,15 @@ public class CommandIntegrationTest {
   @Test
   public void commandDoesNotReturnUnsupported_whenUnsupportedCommandsAreDisabled() {
     radishServer.setEnableUnsupportedCommands(false);
-    try {
-      Map<String, CommandStructure> results = processRawCommands(radishClient.command());
+    Map<String, CommandStructure> results = processRawCommands(radishClient.command());
 
-      // Find an unsupported command
-      RedisCommandType someUnsupported = Arrays.stream(RedisCommandType.values())
-          .filter(RedisCommandType::isUnsupported).findFirst()
-          .orElseThrow(() -> new AssertionError("Could not find any UNSUPPORTED commands"));
+    // Find an unsupported command
+    RedisCommandType someUnsupported = Arrays.stream(RedisCommandType.values())
+        .filter(RedisCommandType::isUnsupported).findFirst()
+        .orElseThrow(() -> new AssertionError("Could not find any UNSUPPORTED commands"));
 
-      for (CommandStructure meta : results.values()) {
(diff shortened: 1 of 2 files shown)


dschneider-pivotal (approved):

DonalEvans (approved):

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/server/CommandIntegrationTest.java
```diff
@@ -19,20 +19,23 @@
...
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
 import io.lettuce.core.RedisClient;
 import io.lettuce.core.api.sync.RedisCommands;
+import junit.framework.AssertionFailedError;
```
kirklund, line 30: You should use `java.lang.AssertionError` instead. `junit.framework` is JUnit 1-1.3, while `org.junit` is JUnit 1.4. Anything in `junit.framework` should be considered deprecated.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/server/CommandIntegrationTest.java
```diff
@@ -73,6 +76,25 @@ public void commandReturnsResultsMatchingNativeRedis() {
...
+      RedisCommandType someUnsupported = Arrays.stream(RedisCommandType.values())
+          .filter(RedisCommandType::isUnsupported).findFirst()
+          .orElseThrow(() -> new AssertionFailedError("Could not find any UNSUPPORTED commands"));
+
+      for (CommandStructure meta : results.values()) {
+        assertThat(meta.name).isNotEqualToIgnoringCase(someUnsupported.name());
+      }
+    } finally {
+      radishServer.setEnableUnsupportedCommands(true);
+    }
```
kirklund, line 95: It's generally best and cleaner if you can avoid using finally-blocks for tear down. You should consider hoisting the finally-block to a dedicated tear down method instead:
```
@After
public void tearDown() {
  radishServer.setEnableUnsupportedCommands(true);
}
```
You can always add a conditional to the tear down if needed.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/server/CommandIntegrationTest.java
jdeppe-pivotal: Done.

geode-apis-compatible-with-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/server/CommandIntegrationTest.java
jdeppe-pivotal: Thanks - I've updated per your suggestion.

PR #6939, opened 2021-10-06
GEODE-9630: remove public setters from GatewaySender interface

onichols-pivotal:


kohlmu-pivotal (approved):
LGTM

mivanac (approved):

albertogpz (approved):

geode-core/src/main/java/org/apache/geode/cache/wan/GatewaySender.java
```diff
@@ -474,46 +474,4 @@
...
   boolean getEnforceThreadsConnectSameReceiver();
-
-
-  /**
-   * Set AlertThreshold for this GatewaySender.
-   *
-   * @since Geode 1.14
-   *
-   */
-  void setAlertThreshold(int alertThreshold);
```
jinmeiliao, line 485: this looks like a public class, can we remove these APIs?

onichols-pivotal: the @since comment was incorrect, these were added in 1.15 so we still have a chance to remove them before we do ship 1.15

jinmeiliao (approved):

PR #6942, opened 2021-10-06
GEODE-9621: Change spotlessGradle behavior for delcaring dependency v…

rhoughton-pivot:
…ersions

Instead of dropping the version string automatically, we throw and force
the user to add the version to the dependency constraint plugin.

### For all changes:


buildSrc/src/main/groovy/org/apache/geode/gradle/plugins/DependencyConstraints.groovy
```diff
@@ -67,6 +67,7 @@ class DependencyConstraints implements Plugin<Project> {
     deps.put("junit.version", "4.13.2")
     deps.put("junit-jupiter.version", "5.7.2")
     deps.put("cglib.version", "3.3.0")
+    deps.put("gradle-tooling-api.version", "5.1.1")
```
onichols-pivotal, line 70: I don't like that this is now separated from the comment that explains why it's such an old version

PR #6949, opened 2021-10-06
GEODE-9536: Disable jar deploy tests on Windows

demery-pivotal:
PROBLEM

Classloaders keep deployed jar files open. On Windows, this prevents
Geode and tests from deleting the files, causing tests in these classes
to fail:
- DeploySemanticVersionJarDUnitTest (geode-gfsh)
- DeploymentSemanticVersionJarDUnitTest (geode-assembly)

SOLUTION

Ignore these tests on Windows.

NOTE

This prepares for an upcoming PR to run all distributed tests (except
explicitly ignored ones like these) on Windows.


jdeppe-pivotal:
Should you give this PR a `windows` label so that it runs on Windows?

demery-pivotal:
> Should you give this PR a `windows` label so that it runs on Windows?

I don't think so. Our Windows CI jobs run only those distributed tests that have a `GfshTest` category. These classes don't have that category. So the tests are never run, and so the `IgnoreOnWindows` rule would never be invoked.

demery-pivotal:
I have reported the `MultiUserAPIDUnitTest` failure in the distributed test job: https://issues.apache.org/jira/browse/GEODE-9689. I am confident that the failure is unrelated to this PR.

PR #6971, opened 2021-10-11
GEODE-9536: Disable client version test on Windows

demery-pivotal:
PROBLEM

When a server receives a connection request with an unrecognized client
version, it sends an informative reply and immediately closes the
socket.

In this situation, the client wants to throw a
`ServerRefusedConnectionException` with a message that includes the
server's informative reply.

On Windows, when the client attempts to read the server's reply, the
socket throws a `SocketException` with the message "Connection reset".
As a result, the caller never receives the informative
`ServerRefusedConnectionException`.

SOLUTION

Disable this test on Windows until the underlying issue is fixed.  See
https://issues.apache.org/jira/browse/GEODE-9698

NOTE

This prepares for an upcoming PR to run all distributed tests (except
explicitly ignored ones like these) on Windows.


demery-pivotal:
> @demery-pivotal the new ticket shows 1.14 as affected version, I'm presuming 1.15 should be there as well or instead...

I've updated it to add 1.15. I think the original problem goes back a long way, but I'm not sure how far.

PR #6976, opened 2021-10-11
GEODE-9617: PartitionedRegionSingleHopDUnitTest fails with ConditionTimeoutException waiting for server to bucket map size

mhansonp:
Did some code cleanup up and made a change to Latch calls to check return values.

### DIFF
--- a/geode-core/src/distributedTest/java/org/apache/geode/cache30/PutAllMultiVmDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/cache30/PutAllMultiVmDUnitTest.java
@@ -22,8 +22,7 @@ package org.apache.geode.cache30;
 
 import static org.apache.geode.test.awaitility.GeodeAwaitility.await;
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.fail;
+import static org.assertj.core.api.Assertions.fail;
 
 import java.util.HashMap;
 import java.util.HashSet;
@@ -35,57 +34,48 @@ import java.util.TreeMap;
 
 import org.junit.Test;
 
-import org.apache.geode.cache.AttributesFactory;
 import org.apache.geode.cache.Cache;
 import org.apache.geode.cache.CacheException;
 import org.apache.geode.cache.CacheFactory;
 import org.apache.geode.cache.CacheTransactionManager;
 import org.apache.geode.cache.DataPolicy;
 import org.apache.geode.cache.Region;
-import org.apache.geode.cache.RegionAttributes;
 import org.apache.geode.cache.RegionDestroyedException;
+import org.apache.geode.cache.RegionFactory;
 import org.apache.geode.cache.Scope;
 import org.apache.geode.distributed.DistributedSystem;
 import org.apache.geode.test.dunit.AsyncInvocation;
-import org.apache.geode.test.dunit.Host;
 import org.apache.geode.test.dunit.Invoke;
-import org.apache.geode.test.dunit.SerializableRunnable;
 import org.apache.geode.test.dunit.VM;
 import org.apache.geode.test.dunit.internal.JUnit4DistributedTestCase;
 
 
-public class PutAllMultiVmDUnitTest extends JUnit4DistributedTestCase { // TODO: reformat
+public class PutAllMultiVmDUnitTest extends JUnit4DistributedTestCase {
 
   static Cache cache;
   static Properties props = new Properties();
-  static Properties propsWork = new Properties();
   static DistributedSystem ds = null;
-  static Region region;
-  static Region mirroredRegion;
+  static Region<Object, Object> region;
+  static Region<Object, Object> mirroredRegion;
   static CacheTransactionManager cacheTxnMgr;
 
   @Override
   public final void postSetUp() throws Exception {
-    Host host = Host.getHost(0);
-    VM vm0 = host.getVM(0);
-    VM vm1 = host.getVM(1);
+    VM vm0 = VM.getVM(0);
+    VM vm1 = VM.getVM(1);
     vm0.invoke(() -> PutAllMultiVmDUnitTest.createCache(DataPolicy.REPLICATE));
     vm1.invoke(() -> PutAllMultiVmDUnitTest.createCache(DataPolicy.REPLICATE));
   }
 
   @Override
   public final void preTearDown() throws Exception {
-    Host host = Host.getHost(0);
-    VM vm0 = host.getVM(0);
-    VM vm1 = host.getVM(1);
+    VM vm0 = VM.getVM(0);
+    VM vm1 = VM.getVM(1);
     vm0.invoke(PutAllMultiVmDUnitTest::closeCache);
     vm1.invoke(PutAllMultiVmDUnitTest::closeCache);
     cache = null;
-    Invoke.invokeInEveryVM(new SerializableRunnable() {
-      @Override
-      public void run() {
-        cache = null;
-      }
+    Invoke.invokeInEveryVM(() -> {
+      cache = null;
     });
   }
 
@@ -93,12 +83,13 @@ public class PutAllMultiVmDUnitTest extends JUnit4DistributedTestCase { // TODO:
     try {
       props.setProperty("log-level", "info");
       ds = (new PutAllMultiVmDUnitTest()).getSystem(props);
-      cache = CacheFactory.create(ds);
-      AttributesFactory factory = new AttributesFactory();
-      factory.setScope(Scope.DISTRIBUTED_ACK);
-      factory.setDataPolicy(dataPolicy);
-      RegionAttributes attr = factory.create();
-      region = cache.createRegion("map", attr);
+      CacheFactory cacheFactory = new CacheFactory(ds.getProperties());
+      cache = cacheFactory.create();
+
+      RegionFactory<Object, Object> regionFactory = cache.createRegionFactory();
+      regionFactory.setScope(Scope.DISTRIBUTED_ACK);
+      regionFactory.setDataPolicy(dataPolicy);
+      region = regionFactory.create("map");
     } catch (Exception ex) {
       ex.printStackTrace();
     }
@@ -106,11 +97,10 @@ public class PutAllMultiVmDUnitTest extends JUnit4DistributedTestCase { // TODO:
 
   public static void createMirroredRegion() {
     try {
-      AttributesFactory factory = new AttributesFactory();
-      factory.setDataPolicy(DataPolicy.REPLICATE);
-      factory.setScope(Scope.DISTRIBUTED_ACK);
-      RegionAttributes attr = factory.create();
-      mirroredRegion = cache.createRegion("mirrored", attr);
+      RegionFactory<Object, Object> regionFactory = cache.createRegionFactory();
+      regionFactory.setScope(Scope.DISTRIBUTED_ACK);
+      regionFactory.setDataPolicy(DataPolicy.REPLICATE);
+      region = regionFactory.create("mirrored");
     } catch (Exception ex) {
       ex.printStackTrace();
     }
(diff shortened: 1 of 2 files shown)


mhansonp:
Seems like the latch changes made the sticking point a little more obvious...

onichols-pivotal:

onichols-pivotal (changes requested):
First commit (and also PR title) must begin with `GEODE-nnnnn: `.  Please amend your initial commit and force-push.

onichols-pivotal (approved):

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionSingleHopDUnitTest.java
```diff
@@ -956,16 +935,16 @@ public void testClientMetadataForPersistentPrs() throws Exception {
       });
     }
 
-    LATCH.get().await(getTimeout().toMillis(), MILLISECONDS);
-
+    assertThat(LATCH.get().await(getTimeout().toMillis(), MILLISECONDS))
```
mhansonp, line 938: Another latch change

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionSingleHopDUnitTest.java
```diff
@@ -952,16 +915,16 @@ public void testClientMetadataForPersistentPrs() throws Exception {
       });
     }
 
-    LATCH.get().await(getTimeout().toMillis(), MILLISECONDS);
+    assertThat(LATCH.get().await(getTimeout().toMillis(), MILLISECONDS)).isTrue();
```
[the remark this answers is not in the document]
mhansonp: These changes to nothing to affect that. That will be a separate PR.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionSingleHopDUnitTest.java
```diff
@@ -85,21 +86,22 @@
...
 import org.apache.geode.test.dunit.rules.DistributedRestoreSystemProperties;
 import org.apache.geode.test.dunit.rules.DistributedRule;
 import org.apache.geode.test.junit.categories.ClientServerTest;
 import org.apache.geode.test.junit.rules.serializable.SerializableTemporaryFolder;
 
 @Category(ClientServerTest.class)
-@SuppressWarnings("serial")
 public class PartitionedRegionSingleHopDUnitTest implements Serializable {
-
+  protected static Logger logger = LogService.getLogger();
```
DonalEvans, line 104: This logger is not used in the test so should be removed.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionSingleHopDUnitTest.java
```diff
@@ -919,21 +897,22 @@ public void testClientMetadataForPersistentPrs() throws Exception {
...
     int locatorPort = DUnitEnv.get().getLocatorPort();
 
-    vm0.invoke(() -> createServer("disk", -1, 3, 4));
-    vm1.invoke(() -> createServer("disk", -1, 3, 4));
-    vm2.invoke(() -> createServer("disk", -1, 3, 4));
-    vm3.invoke(() -> createServer("disk", -1, 3, 4));
+    vm0.invoke((SerializableCallableIF<Integer>) this::createServer);
+    vm1.invoke((SerializableCallableIF<Integer>) this::createServer);
+    vm2.invoke((SerializableCallableIF<Integer>) this::createServer);
+    vm3.invoke((SerializableCallableIF<Integer>) this::createServer);
```
DonalEvans, line 903: These casts can be avoided if you use lambdas here:
```
    vm0.invoke(() -> createServer());
    vm1.invoke(() -> createServer());
    vm2.invoke(() -> createServer());
    vm3.invoke(() -> createServer());
```

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionSingleHopDUnitTest.java
```diff
@@ -956,16 +935,16 @@ public void testClientMetadataForPersistentPrs() throws Exception {
...
+        vm3.invokeAsync((SerializableCallableIF<Integer>) this::createServer);
     AsyncInvocation<Integer> createServerOnVM2 =
-        vm2.invokeAsync(() -> createServer("disk", -1, 3, 4));
+        vm2.invokeAsync((SerializableCallableIF<Integer>) this::createServer);
     AsyncInvocation<Integer> createServerOnVM1 =
-        vm1.invokeAsync(() -> createServer("disk", -1, 3, 4));
+        vm1.invokeAsync((SerializableCallableIF<Integer>) this::createServer);
     AsyncInvocation<Integer> createServerOnVM0 =
-        vm0.invokeAsync(() -> createServer("disk", -1, 3, 4));
+        vm0.invokeAsync((SerializableCallableIF<Integer>) this::createServer);
```
DonalEvans, line 947: More casts that can be avoided by using lambdas rather than method references.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionSingleHopDUnitTest.java
```diff
@@ -1405,7 +1385,8 @@ private void waitForLocalBucketsCreation() {
     await().untilAsserted(() -> assertThat(pr.getDataStore().getAllLocalBuckets()).hasSize(4));
```
DonalEvans, line 1385: While you're cleaning up the test code, warnings on lines 1328, 1333, 1338 and 1342 can be fixed by using `PutFunction<>()`.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionSingleHopDUnitTest.java
```diff
@@ -1501,11 +1482,11 @@ public String getId() {
...
+    public void execute(FunctionContext<T> context) {
       RegionFunctionContext rc = (RegionFunctionContext) context;
       Region<Object, Object> r = rc.getDataSet();
-      Set filter = rc.getFilter();
-      if (rc.getFilter() == null) {
+      Set<Object> filter = (Set<Object>) rc.getFilter();
+      if (filter == null) {
         for (int i = 0; i < 200; i++) {
           r.put(i, i);
         }
```
DonalEvans, line 1492: The `equals()` method for the Order class below is broken. Line 1604 should be:
```
      if (!(obj instanceof Order)) {
```

geode-core/src/distributedTest/java/org/apache/geode/cache30/PutAllMultiVmDUnitTest.java
```diff
@@ -347,22 +319,12 @@ public void testPutAllFromPreload() throws InterruptedException {
...
-          region.clear();
-        } catch (Exception ex) {
-          ex.printStackTrace();
-        }
-      }
-    };// end of clear
+    // end of clear
 
-    vm0.invoke(new CacheSerializableRunnable("testSimplePutAll1") {
+    vm0.invoke("testSimplePutAll1", new CacheSerializableRunnable() {
```
DonalEvans, line 327: This (and other uses of `CacheSerializableRunnable` below) can be simplified by using a lambda instead:
```
    vm0.invoke("testSimplePutAll1", () -> {
      int cntr = 0, cntr1 = 0;
      for (int i = 1; i < 6; i++) {
        region.put(i, "testSimplePutAll" + i);
        cntr++;
      }
...
    });
```

geode-core/src/distributedTest/java/org/apache/geode/cache30/PutAllMultiVmDUnitTest.java
```diff
@@ -486,24 +443,24 @@ public void run2() throws CacheException {
           // do nothing
```
DonalEvans, line 443: On line 441 above this, rather than using `fail()` when we expect a call to raise a throwable but it doesn't, it would be better to use `assertThatThrownBy()` to assert that we see the expected exception. This provides a more descriptive failure in the event that we don't see the expected exception. This comment also applies to the other 4 uses of `fail()` in this test.

DonalEvans (commented):
Per [the commit message format guidelines](https://cwiki.apache.org/confluence/display/GEODE/Commit+Message+Format) found on the wiki, could this PR name (or the final commit message when merging) be changed to describe what the commit is actually changing rather than just being a restatement of the ticket title? Maybe something like "GEODE-9617: Assert on successful latch await", since as far as I can see, none of the changes are intended to address the flaky test, just make it more obvious when it does fail?

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionSingleHopDUnitTest.java
mhansonp: Great! Thanks!

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionSingleHopDUnitTest.java
mhansonp: Isn't the above the intended approach of Java? It is the optimized form according to intellij

DonalEvans: If there was just one `createServer()` method in the class (or the zero-arg version of the method had a unique name) then we wouldn't need the cast here, but since there are multiple methods with different signatures but the same name, omitting the cast and just using a method reference results in ambiguity, since the compiler doesn't know which one you mean and throws an error.

It's true that method references are preferred in Java, but the performance optimizations are negligible in a test, and more readable code is a greater benefit here, I think. IntelliJ doesn't even highlight the lambda as replaceable with a method reference if you make this change, because I guess it knows that the compiler will complain if you do.

geode-core/src/distributedTest/java/org/apache/geode/cache30/PutAllMultiVmDUnitTest.java
mhansonp: You cannot do that because CacheSerializableRunnable throws a CacheException

`  @Override
  public void run() {
    try {
      run2();
    } catch (CacheException exception) {
      String message = "While invoking \"" + this + "\"";
      throw new CacheSerializableRunnableException(message, exception);
    }
  }
`

geode-core/src/distributedTest/java/org/apache/geode/cache30/PutAllMultiVmDUnitTest.java
mhansonp: Nice. Very clean.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionSingleHopDUnitTest.java
mhansonp: Ok, I switched Object to ?

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionSingleHopDUnitTest.java
mhansonp: Same as above.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionSingleHopDUnitTest.java
mhansonp: I agree the casts are ugly.

mhansonp: I will swap them to a lambda..

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionSingleHopDUnitTest.java
mhansonp: I will swap them to lambdas...

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/PartitionedRegionSingleHopDUnitTest.java
mhansonp: done

PR #6986, opened 2021-10-12
GEODE-9725: Remove membership port range feature from AvailablePort

demery-pivotal:
`AvailablePort` and `AvailablePortHelper` have methods to retrieve an
"available" port from the membership port range. This feature is both
unnecessary and inherently unreliable.

INHERENTLY UNRELIABLE

The default membership port range overlaps each OS's ephemeral port
range. An ephemeral port deemed to be "available" by one of these
methods can be put into use by another process even before the method
returns. It is not safe to rely on such a port to remain available after
it is checked.

UNNECESSARY

Currently the only callers that request ports from the membership port
range are two tests in `DistributedSystemDUnitTest`. In each case, the
test uses the returned ports to *set* the membership port range for a
member. There is no need for the ports to come from any particular
range.

There are no other uses of this feature, either in the product or in
tests.

SOLUTION

Remove the "membership port range" feature from `AvailablePort` and
`AvailablePortHelper`.

Change `DistributedSystemDUnitTest` to get available ports from the
"available port range." This range is safe to use (assuming all other
concurrently-running tests are well-behaved).

### DIFF
--- a/geode-core/src/distributedTest/java/org/apache/geode/distributed/DistributedSystemDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/distributed/DistributedSystemDUnitTest.java
@@ -101,7 +101,7 @@ public class DistributedSystemDUnitTest extends JUnit4DistributedTestCase {
     this.locatorPort = getRandomAvailableTCPPort();
     this.tcpPort = getRandomAvailableTCPPort();
 
-    int[] portRange = getRandomAvailableTCPPortRange(3, true);
+    int[] portRange = getRandomAvailableTCPPortRange(3);
     this.lowerBoundOfPortRange = portRange[0];
     this.upperBoundOfPortRange = portRange[portRange.length - 1];
   }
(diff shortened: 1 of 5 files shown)


demery-pivotal:
> I understand your "INHERENTLY UNRELIABLE" analysis. What I don't understand is: why does that argument apply only to ports in the membership port range? Seems to me that any method that purports to return an "available port range" is misguided. What am I missing here?

The key is that the membership port range overlaps with each operating system's ephemeral port range. The operating system is free to give those ports to anyone who asks for an ephemeral port. Many tests launch members in a way that uses ephemeral ports, making the risk significantly higher in CI than in normal use.

A port in any other range (such as the "available port" range) is likely to be put into use only if a caller specifically asks for that port. We have a fairly strong convention to use those ports only after checking their availability. And the way tests check availability significantly minimizes the risk of conflict. The risk is minimized by the build system allocating a distinct range of ports to each concurrently-executing test JVM, and by `AvailablePortHelper` allocating those ports round-robin instead of randomly.

PR #7007, opened 2021-10-17
GEODE-9742:fix resource leak due to Files.list and Files.walk

lujiefsi:
see https://issues.apache.org/jira/browse/GEODE-9742


kirklund:
You'll need to remove unused imports to get the build jobs to pass the spotless check.

You can use `spotlessApply` to automatically reformat code based on our spotless config:
`./gradlew spotlessApply`

You can use `spotlessCheck` to run the same format validation that failed in the build jobs for you PR:
`./gradlew spotlessCheck`

Note that `spotlessApply` is unable to remove unused imports, so you'll have to do that step manually.

PR #7009, opened 2021-10-18
GEODE-9744:fix bug like GEODE-9744

lujiefsi:
see https://issues.apache.org/jira/browse/GEODE-9744


jdeppe-pivotal:
Closing this as it appears to be stale.

PR #7010, opened 2021-10-18
GEODE-9632: Allow INDEX_THRESHOLD_SIZE System property to override CompiledValue.RESULT_LIMIT

mkevo:
### For all changes:

### DIFF
--- a/geode-core/src/integrationTest/java/org/apache/geode/cache/query/functional/LimitClauseJUnitTest.java
+++ b/geode-core/src/integrationTest/java/org/apache/geode/cache/query/functional/LimitClauseJUnitTest.java
@@ -673,7 +673,7 @@ public class LimitClauseJUnitTest {
       assertEquals(5, result.size());
       SelectResults wrapper = (SelectResults) result;
       assertEquals(5, wrapper.asSet().size());
-      assertTrue(observer.limitAppliedAtIndex);
+      assertFalse(observer.limitAppliedAtIndex);
     } catch (Exception e) {
       CacheUtils.getLogger().error(e);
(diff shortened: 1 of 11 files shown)


mkevo:
The problem in failing tests is that with a fix, the limit added in the query will be applied only on results, not on index results. Size of indexResults is defined by INDEX_THRESHOLD_SIZE. A new test case(_testLimitIsAppliedOnlyOnQueryResults_ in QueryWithRangeIndexDUnitTest.java) is added to verify it. 
Not sure what to do with failing tests, as they have assertions to check if limitAppliedAtIndex, and it will not apply as it checks INDEX_THRESHOLD_SIZE for limit at index.
Also, this can be fixed by the first proposal with fetching INDEX_THRESHOLD_SIZE every time the index is queried, but think that it is not the best solution as it will not take INDEX_THRESHOLD_SIZE into a count if it is not specified in the server startup command. This can lead to not collecting results in indexResults and the result of the query can be wrong.

Also, the PR name should be something like: _GEODE-9632: Add option to apply INDEX_THRESHOLD_SIZE on index results_, as it will not overwrite RESULT_LIMIT for printing results, just for collection of index results.

DonalEvans:
The reason that the tests are currently failing is that there are only 50 entries in the region but the default value for the index threshold limit is 100, so that limit is not being surpassed and so the `limitAppliedAtIndex` flag is not being set to true. Increasing the number of objects in the region to a value above 100 results in the tests passing. I think that this fix is fine, since the previous assertion was based on the incorrect behaviour which this PR is fixing.

DonalEvans (approved):

PR #7063, opened 2021-10-29
GEODE-9746: do not logout old subjects immediately when we get a new subject for the same uniqueId

jinmeiliao:
In some cases, client will send in Authentication request with existing uniqueId (in both re-auth or non-re-auth cases), we used to just replace the old subject with the new subject and log out the old subject immediately, but this old subject might already be bound with a running thread, hence causing "FailedToFindSubject" exception on that thread. We can't immediately log out the old subject, so we have to keep reference of them and log out them all at connection closing time.

This PR also includes a minor fix in CacheClientProxy for completeness of the logic

### DIFF
--- a/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/tier/sockets/ClientServerConnectDisconnectDistributedTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/tier/sockets/ClientServerConnectDisconnectDistributedTest.java
@@ -151,7 +151,7 @@ public class ClientServerConnectDisconnectDistributedTest implements Serializabl
     authorizations = new ArrayList<>();
     for (ServerConnection sc : acceptor.getAllServerConnections()) {
       ClientUserAuths auth = sc.getClientUserAuths();
-      assertThat(auth.getAllSubjects().size()).isNotEqualTo(0);
+      assertThat(auth.getAllSubjects()).isNotEmpty();
       authorizations.add(auth);
       for (Subject subject : auth.getAllSubjects()) {
         assertThat(subject.getPrincipal()).isNotNull();
@@ -204,7 +204,7 @@ public class ClientServerConnectDisconnectDistributedTest implements Serializabl
     }
 
     for (ClientUserAuths auth : authorizations) {
-      assertThat(auth.getAllSubjects().size()).isEqualTo(0);
+      assertThat(auth.getAllSubjects()).isEmpty();
     }
   }
 
--- a/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/tier/sockets/ClientServerConnectDisconnectDistributedTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/tier/sockets/ClientServerConnectDisconnectDistributedTest.java
@@ -151,9 +151,9 @@ public class ClientServerConnectDisconnectDistributedTest implements Serializabl
     authorizations = new ArrayList<>();
     for (ServerConnection sc : acceptor.getAllServerConnections()) {
       ClientUserAuths auth = sc.getClientUserAuths();
-      assertThat(auth.getSubjects().size()).isNotEqualTo(0);
+      assertThat(auth.getAllSubjects().size()).isNotEqualTo(0);
       authorizations.add(auth);
-      for (Subject subject : auth.getSubjects()) {
+      for (Subject subject : auth.getAllSubjects()) {
         assertThat(subject.getPrincipal()).isNotNull();
         assertThat(subject.getPrincipals()).isNotNull();
         assertThat(subject.isAuthenticated()).isTrue();
@@ -204,7 +204,7 @@ public class ClientServerConnectDisconnectDistributedTest implements Serializabl
     }
 
     for (ClientUserAuths auth : authorizations) {
-      assertThat(auth.getSubjects().size()).isEqualTo(0);
+      assertThat(auth.getAllSubjects().size()).isEqualTo(0);
     }
   }
 
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/CacheClientProxy.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/CacheClientProxy.java
@@ -223,7 +223,7 @@ public class CacheClientProxy implements ClientSession {
   private Subject subject;
 
   /**
-   * used for cq name to subject/auth mapping, always initialized in single/multi user casees
+   * used for cq name to subject/auth mapping, always initialized in single/multi user cases
    */
   private ClientUserAuths clientUserAuths;
 
@@ -783,9 +783,10 @@ public class CacheClientProxy implements ClientSession {
         else if (subject != null) {
           logger.debug("CacheClientProxy.close, logging out: " + subject.getPrincipal());
           subject.logout();
+          subject = null;
         }
         // for multiUser case, in non-durable case, we are closing the connection
-        else {
+        else if (clientUserAuths != null) {
           clientUserAuths.cleanup(true);
           clientUserAuths = null;
         }
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/CacheClientProxy.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/CacheClientProxy.java
@@ -14,6 +14,8 @@
  */
 package org.apache.geode.internal.cache.tier.sockets;
 
+import static org.apache.geode.logging.internal.spi.LoggingProvider.SECURITY_LOGGER_NAME;
+
 import java.io.IOException;
 import java.net.Socket;
 import java.net.SocketException;
@@ -94,6 +96,7 @@ import org.apache.geode.util.internal.GeodeGlossary;
 @SuppressWarnings("synthetic-access")
 public class CacheClientProxy implements ClientSession {
   private static final Logger logger = LogService.getLogger();
+  private static final Logger secureLogger = LogService.getLogger(SECURITY_LOGGER_NAME);
 
   @Immutable
   @VisibleForTesting
@@ -781,12 +784,13 @@ public class CacheClientProxy implements ClientSession {
         // single user case -- integrated security
         // connection is closed, so we can log out this subject
         else if (subject != null) {
-          logger.debug("CacheClientProxy.close, logging out: " + subject.getPrincipal());
+          secureLogger.debug("CacheClientProxy.close, logging out {}. ", subject.getPrincipal());
           subject.logout();
           subject = null;
         }
         // for multiUser case, in non-durable case, we are closing the connection
         else if (clientUserAuths != null) {
+          secureLogger.debug("CacheClientProxy.close, cleanup all client subjects. ");
           clientUserAuths.cleanup(true);
           clientUserAuths = null;
         }
@@ -979,8 +983,8 @@ public class CacheClientProxy implements ClientSession {
 
     // Logout the subject
     if (subject != null) {
-      logger.info(
-          "CacheClientProxy.closeOtherTransientFields, logging out: " + subject.getPrincipal());
+      secureLogger.debug(
+          "CacheClientProxy.closeOtherTransientFields, logging out {}. ", subject.getPrincipal());
       subject.logout();
     }
   }
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/CacheClientProxy.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/CacheClientProxy.java
@@ -310,7 +310,7 @@ public class CacheClientProxy implements ClientSession {
         acceptorId, notifyBySubscription, securityService, subject, statisticsClock,
         ccn.getCache().getInternalDistributedSystem().getStatisticsManager(),
         DEFAULT_CACHECLIENTPROXYSTATSFACTORY,
-        DEFAULT_MESSAGEDISPATCHERFACTORY);
+        DEFAULT_MESSAGEDISPATCHERFACTORY, ServerConnection.getClientUserAuths(proxyID));
   }
 
   @VisibleForTesting
@@ -320,7 +320,7 @@ public class CacheClientProxy implements ClientSession {
       SecurityService securityService, Subject subject, StatisticsClock statisticsClock,
       StatisticsFactory statisticsFactory,
       CacheClientProxyStatsFactory cacheClientProxyStatsFactory,
-      MessageDispatcherFactory messageDispatcherFactory)
+      MessageDispatcherFactory messageDispatcherFactory, ClientUserAuths clientUserAuths)
       throws CacheException {
     initializeTransientFields(socket, proxyID, isPrimary, clientConflation, clientVersion);
     _cacheClientNotifier = ccn;
@@ -345,11 +345,7 @@ public class CacheClientProxy implements ClientSession {
     _cacheClientNotifier.getAcceptorStats().incCurrentQueueConnections();
     creationDate = new Date();
     this.messageDispatcherFactory = messageDispatcherFactory;
-    initializeClientAuths();
-  }
-
-  void initializeClientAuths() {
-    clientUserAuths = ServerConnection.getClientUserAuths(proxyID);
+    this.clientUserAuths = clientUserAuths;
   }
 
   private void reinitializeClientAuths() {
@@ -774,14 +770,22 @@ public class CacheClientProxy implements ClientSession {
 
     connected = false;
 
-    // Close the Authorization callback (if any)
+    // Close the Authorization callback or subject if we are not keeping the proxy
     try {
       if (!pauseDurable) {
-        if (postAuthzCallback != null) {// for single user
+        // single user case -- old security
+        if (postAuthzCallback != null) {
           postAuthzCallback.close();
           postAuthzCallback = null;
         }
-        if (clientUserAuths != null) {// for multiple users
+        // single user case -- integrated security
+        // connection is closed, so we can log out this subject
+        else if (subject != null) {
+          logger.debug("CacheClientProxy.close, logging out: " + subject.getPrincipal());
+          subject.logout();
+        }
+        // for multiUser case, in non-durable case, we are closing the connection
+        else {
           clientUserAuths.cleanup(true);
           clientUserAuths = null;
         }
@@ -941,7 +945,8 @@ public class CacheClientProxy implements ClientSession {
     return false;
   }
 
-  private void closeTransientFields() {
+  @VisibleForTesting
+  protected void closeTransientFields() {
     if (!closeSocket()) {
       // The thread who closed the socket will be responsible to
       // releaseResourcesForAddress and clearClientInterestList
@@ -973,6 +978,8 @@ public class CacheClientProxy implements ClientSession {
 
     // Logout the subject
     if (subject != null) {
+      logger.info(
+          "CacheClientProxy.closeOtherTransientFields, logging out: " + subject.getPrincipal());
       subject.logout();
     }
   }
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientUserAuths.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientUserAuths.java
@@ -52,7 +52,7 @@ public class ClientUserAuths {
   // In the expirable credential case, there will be multiple
   // subjects created associated with one uniqueId. We always save the current subject to the top of
   // the list. The rest are "to-be-retired".
-  private final ConcurrentMap<Long, List<Subject>> uniqueIdVsSubject =
+  private final ConcurrentMap<Long, List<Subject>> uniqueIdVsSubjects =
       new ConcurrentHashMap<>();
   private final ConcurrentMap<String, Long> cqNameVsUniqueId = new ConcurrentHashMap<>();
 
@@ -82,13 +82,13 @@ public class ClientUserAuths {
     // we log out that subject immediately, that thread "authorize" would get null principal.
     synchronized (this) {
       List<Subject> subjects;
-      if (!uniqueIdVsSubject.containsKey(newId)) {
+      if (!uniqueIdVsSubjects.containsKey(newId)) {
         logger.debug("Subject of {} added.", newId);
         subjects = new ArrayList<>();
-        uniqueIdVsSubject.put(newId, subjects);
+        uniqueIdVsSubjects.put(newId, subjects);
       } else {
         logger.debug("Subject of {} replaced.", newId);
-        subjects = uniqueIdVsSubject.get(newId);
+        subjects = uniqueIdVsSubjects.get(newId);
       }
       // always add the latest subject to the top of the list;
       subjects.add(0, subject);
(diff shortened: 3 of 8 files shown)


jmelchio (approved):

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientUserAuths.java
```diff
@@ -52,17 +62,32 @@ public Long putUserAuth(UserAuthAttributes userAuthAttr) {
...
+      if (!uniqueIdVsSubject.containsKey(newId)) {
+        logger.debug("Subject of {} added.", newId);
+        subjects = new CopyOnWriteArrayList<>();
+        uniqueIdVsSubject.put(newId, subjects);
+      } else {
+        logger.debug("Subject of {} replaced.", newId);
+        subjects = uniqueIdVsSubject.get(newId);
+      }
+      // always add the latest subject to the top of the list;
+      subjects.add(0, subject);
```
jchen21, line 89: This will shift every element in the list, which could be costly. When adding a new subject, it makes a copy of the array. At the same time, the other thread could call `subjects.get(0)` and get the old subject before the new one is added.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientUserAuths.java
```diff
@@ -90,28 +115,29 @@ public UserAuthAttributes getUserAuthAttributes(final Long userId) {
...
+  public synchronized Subject getSubject(final Long userId) {
+    CopyOnWriteArrayList<Subject> subjects = uniqueIdVsSubject.get(userId);
+    if (subjects == null || subjects.isEmpty()) {
+      return null;
+    }
+    return subjects.get(0);
   }
 
-  public void removeSubject(final Long userId) {
+  public synchronized void removeSubject(final Long userId) {
```
jchen21, line 134: The extra synchronization is not necessary on `ConcurrentMap` and `CopyOnWriteArrayList`.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientUserAuths.java
```diff
@@ -121,7 +147,7 @@ public UserAuthAttributes getUserAuthAttributes(final String cqName) {
   public Subject getSubject(final String cqName) {
     Long uniqueId = cqNameVsUniqueId.get(cqName);
     if (uniqueId != null) {
-      return uniqueIdVsSubject.get(uniqueId);
+      return uniqueIdVsSubject.get(uniqueId).get(0);
```
jchen21, line 150: Need a null check before calling `get(0)`. Otherwise it could throw `NullPointerException`.

jchen21 (changes requested):

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientUserAuths.java
jmelchio: > This will shift every element in the list, which could be costly. When adding a new subject, it makes a copy of the array. At the same time, the other thread could call `subjects.get(0)` and get the old subject before the new one is added.

I think that `subjects.get(0)` is only invoked in a synchronized method which has the same lock scope as the `synchronized(this){}` block so that should prevent getting and putting to clash. 
Maybe the `synchronized(this){}` block could be limited to encapsulate just the `subjects.add(0, subject);` statement?

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/CacheClientProxy.java
```diff
@@ -774,14 +770,22 @@ protected boolean close(boolean checkQueue, boolean stoppedNormally) {
...
         }
-        if (clientUserAuths != null) {// for multiple users
+        // single user case -- integrated security
+        // connection is closed, so we can log out this subject
+        else if (subject != null) {
+          logger.debug("CacheClientProxy.close, logging out: " + subject.getPrincipal());
+          subject.logout();
+        }
+        // for multiUser case, in non-durable case, we are closing the connection
+        else {
```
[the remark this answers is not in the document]
jinmeiliao: `clientUserAuths` is always initialized, could not be null.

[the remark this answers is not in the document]
jinmeiliao: `clientUserAuths` is always initialized, should not be null, but we can add the null check.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientUserAuths.java
```diff
@@ -52,17 +62,32 @@ public Long putUserAuth(UserAuthAttributes userAuthAttr) {
...
       newId = getNextID();
     } else {
       newId = existingUniqueId;
     }
 
-    Subject oldSubject = uniqueIdVsSubject.put(newId, subject);
-    removeSubject(oldSubject);
-    logger.debug("Subject of {} added.", newId);
+    // we are saving all the subjects that's related to this uniqueId
+    // we cannot immediately log out the old subject of this userId because
```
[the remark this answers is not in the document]
jinmeiliao: yes.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientUserAuths.java
jinmeiliao: for performance concerns, after we fix multiple threads in the client could send in authentication requests, `putSubject` should only be called  multiple times with the same uniqueId when credentials expires, which is not that frequent. I think we ignore the cost issue.

put/get are synchronized on the object itself, so there should not be `get(0)` happening while other thread is doing the put.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientUserAuths.java
jinmeiliao: this prevents other thread doing `put/get` at the same time.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientUserAuths.java
jinmeiliao: good catch!

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientUserAuths.java
```diff
@@ -52,17 +62,32 @@ public Long putUserAuth(UserAuthAttributes userAuthAttr) {
...
+    // it might already be bound to another thread and doing operations. If
+    // we log out that subject immediately, that thread "authorize" would get null principal.
+    synchronized (this) {
```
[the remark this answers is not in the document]
jinmeiliao: we are preventing user doing `putSubject` `getSubject`, `removeSubject` at the same time. this `synchronize` block is necessary.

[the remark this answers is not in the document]
jchen21: `CopyOnWriteArrayList` offers built-in synchronization for all write access and it makes a copy of the underlying array for all write access. If you are trying to synchronize read and write access of the list with your own synchronization, it is not necessary to use `CopyOnWriteArrayList`. Because this introduces extra overhead of synchronization and copying the array.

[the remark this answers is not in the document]
jinmeiliao: that's true, it might be redundant using that list. I can change it to use regular list

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientUserAuths.java
jchen21: `CopyOnWriteArrayList` offers built-in synchronization for all write access and it makes a copy of the underlying array for all write access. If you are trying to synchronize read and write access of the list with your own synchronization, it is not necessary to use `CopyOnWriteArrayList`. Because this introduces extra overhead of synchronization and copying the array on write.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientUserAuths.java
```diff
@@ -90,28 +120,29 @@ public UserAuthAttributes getUserAuthAttributes(final Long userId) {
   }
 
   @VisibleForTesting
+  @TestOnly
   protected Collection<Subject> getSubjects() {
```
jchen21, line 124: It is better to make `getSubjects()` `synchronized` as well, even though it is for test only, in case something slips through the crack.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientUserAuths.java
DonalEvans: Why does the latest subject have to be inserted as the first element in the list? Could we instead add that subject to the end of the list, and then in `getSubject(final Long userId)` retrieve the last element in the list? This would be a simple solution to the performance issue that Jianxia described.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientUserAuths.java
DonalEvans: Could this be "getAllSubjects" to make it clear that it's getting all subjects for all unique IDs?

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientUserAuths.java
```diff
@@ -90,28 +120,29 @@ public UserAuthAttributes getUserAuthAttributes(final Long userId) {
...
+    List<Subject> subjects = uniqueIdVsSubject.get(userId);
+    if (subjects == null || subjects.isEmpty()) {
+      return null;
+    }
+    return subjects.get(0);
   }
 
-  public void removeSubject(final Long userId) {
+  public synchronized void removeSubject(final Long userId) {
     logger.debug("Subject of {} removed.", userId);
```
DonalEvans, line 140: Given that this call may result in multiple subjects being removed, would it be worth modifying the debug logging to give a little more information, such as how many subjects were removed (if any)?

geode-core/src/test/java/org/apache/geode/internal/cache/tier/sockets/CacheClientProxyTest.java
```diff
@@ -75,32 +79,41 @@ public void before() throws Exception {
...
-    proxy = spy(new CacheClientProxy(cache, notifier, socket, id, true, (byte) 1, version, 1L, true,
-        securityService, null, clock, statsFactory, proxyStatsFactory, dispatcherFactory));
-    proxy.getSubject("cq");
+    proxyWithMultiUser.getSubject("cq");
   }
 
   @Test
   public void deliverMessageWhenSubjectIsNotNull() {
     when(proxyStatsFactory.create(any(), any(), any()))
         .thenReturn(mock(CacheClientProxyStats.class));
```
DonalEvans, line 108: If this mocking is moved to the `@Before` method then you can avoid having to create a new `CacheClientProxy` on the following lines. This also applies to the `deliverMessageWhenSubjectIsNull()` test.

geode-core/src/test/java/org/apache/geode/internal/cache/tier/sockets/CacheClientProxyTest.java
```diff
@@ -109,23 +122,59 @@ public void deliverMessageWhenSubjectIsNotNull() {
...
+  }
+
+  @Test
+  public void close_singleUser() {
+    when(id.isDurable()).thenReturn(false);
+    CacheClientProxy spy = spy(proxyWithSingleUser);
+    doNothing().when(spy).closeTransientFields();
+    boolean keepProxy = spy.close(true, false);
+    assertThat(keepProxy).isFalse();
+    verify(subject, times(1)).logout();
```
DonalEvans, line 166: When `times()` is 1, it can be omitted, as the default behaviour of `verify()` with no second argument is to verify that the method was invoked once.

geode-core/src/test/java/org/apache/geode/internal/cache/tier/sockets/CacheClientProxyTest.java
```diff
@@ -109,23 +122,59 @@ public void deliverMessageWhenSubjectIsNotNull() {
...
+    verify(clientUserAuths, never()).cleanup(anyBoolean());
+
+    keepProxy = proxyWithMultiUser.close(true, false);
+    assertThat(keepProxy).isTrue();
+    verify(subject, never()).logout();
+    verify(clientUserAuths, never()).cleanup(anyBoolean());
+  }
+
+  @Test
+  public void close_singleUser() {
```
DonalEvans, line 160: Could this test name be a little more descriptive? Adding what the expected behaviour is would make it much clearer what the test is actually testing. This also applies to the `close_multiUser()` test.

geode-core/src/test/java/org/apache/geode/internal/cache/tier/sockets/ClientUserAuthsTest.java
```diff
@@ -45,37 +47,78 @@ public void before() throws Exception {
...
+  @Test
+  public void getSubjectReturnsTheLatestOne() {
+    Long id1 = auth.putSubject(subject1, -1);
+    assertThat(auth.getSubject(id1)).isSameAs(subject1);
+    auth.putSubject(subject2, id1);
+    assertThat(auth.getSubject(id1)).isSameAs(subject2);
+  }
+
+  @Test
+  public void getSubjectOfNonExistentId() {
```
DonalEvans, line 91: Could this test name include the expected behaviour? Something like "getSubjectWithNonExistentIdReturnsNull" maybe? This also applies to `removeSubjectOfNonExistentId()`, `getSubjectWithCq()` and `removeSubject()`.

DonalEvans (approved):
Some suggestions and a typo, but all optional.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientUserAuths.java
```diff
@@ -16,29 +16,44 @@
...
   private final ConcurrentMap<Long, UserAuthAttributes> uniqueIdVsUserAuth =
       new ConcurrentHashMap<>();
   private final ConcurrentMap<String, UserAuthAttributes> cqNameVsUserAuth =
       new ConcurrentHashMap<>();
-  private final ConcurrentMap<Long, Subject> uniqueIdVsSubject = new ConcurrentHashMap<>();
+  // use a list to store all the subjects that's created for this uniqueId
+  // In the expirable credential case, there will be multiple
+  // subjects created associated with one uniqueId. We always save the current subject to the top of
+  // the list. The rest are "to-be-retired".
+  private final ConcurrentMap<Long, List<Subject>> uniqueIdVsSubject =
```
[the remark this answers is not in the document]
jinmeiliao: done

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientUserAuths.java
jinmeiliao: done

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientUserAuths.java
jinmeiliao: good idea.

geode-core/src/test/java/org/apache/geode/internal/cache/tier/sockets/CacheClientProxyTest.java
jinmeiliao: done

jchen21 (approved):

mhansonp (approved):

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/tier/sockets/ClientServerConnectDisconnectDistributedTest.java
```diff
@@ -151,9 +151,9 @@ private void verifyServerConnectionSubjectsAreLoggedIn(AcceptorImpl acceptor) {
     authorizations = new ArrayList<>();
     for (ServerConnection sc : acceptor.getAllServerConnections()) {
       ClientUserAuths auth = sc.getClientUserAuths();
-      assertThat(auth.getSubjects().size()).isNotEqualTo(0);
+      assertThat(auth.getAllSubjects().size()).isNotEqualTo(0);
```
kirklund, line 154: The failure message will show the contents of the collection if you change to:
```
assertThat(auth.getAllSubjects()).isEmpty();
```

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/tier/sockets/ClientServerConnectDisconnectDistributedTest.java
```diff
@@ -204,7 +204,7 @@ private void verifyServerConnectionSubjectsAreLoggedOut() {
     }
 
     for (ClientUserAuths auth : authorizations) {
-      assertThat(auth.getSubjects().size()).isEqualTo(0);
+      assertThat(auth.getAllSubjects().size()).isEqualTo(0);
```
kirklund, line 207:
```
assertThat(auth.getAllSubjects()).isEmpty();
```

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/CacheClientProxy.java
```diff
@@ -774,14 +770,23 @@ protected boolean close(boolean checkQueue, boolean stoppedNormally) {
...
+        // single user case -- old security
+        if (postAuthzCallback != null) {
           postAuthzCallback.close();
           postAuthzCallback = null;
         }
-        if (clientUserAuths != null) {// for multiple users
+        // single user case -- integrated security
+        // connection is closed, so we can log out this subject
+        else if (subject != null) {
+          logger.debug("CacheClientProxy.close, logging out: " + subject.getPrincipal());
```
kirklund, line 784: This is probably exposing sensitive data to the log even though it's not the password. The User can set the log-level to debug which turns on these log statements.

Also, all security related logging in the product should use the SECURITY_LOGGER_NAME:
```
private static final Logger logger = LogService.getLogger(SECURITY_LOGGER_NAME);
```
The SECURITY_LOGGER_NAME gets redirected to the `ConfigurationProperties.SECURITY_LOG_FILE` if the user has set it.

Don't forget, the log4j2 uses `{}` in the syntax:
```
logger.debug("CacheClientProxy.close, logging out: {}", subject.getPrincipal());
```

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/CacheClientProxy.java
```diff
@@ -973,6 +979,8 @@ private void closeOtherTransientFields() {
 
     // Logout the subject
     if (subject != null) {
+      logger.info(
+          "CacheClientProxy.closeOtherTransientFields, logging out: " + subject.getPrincipal());
```
kirklund, line 983: This should use SECURITY_LOGGER_NAME.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientUserAuths.java
```diff
@@ -90,28 +120,29 @@ public UserAuthAttributes getUserAuthAttributes(final Long userId) {
...
-  }
-
-  public Subject getSubject(final Long userId) {
-    return uniqueIdVsSubject.get(userId);
-  }
-
-  public void removeSubject(final Long userId) {
-    logger.debug("Subject of {} removed.", userId);
-    removeSubject(uniqueIdVsSubject.remove(userId));
+  @TestOnly
```
kirklund, line 123: We should probably have some sort of discussion or tech talk about when `@VisibleForTesting` or `@TestOnly` should be used. Someone needs to own introducing `@TestOnly` and should add a wiki page or other info about when to use these annotations. They should also submit a PR to remove `@VisibleForTesting` and convert all uses to `@TestOnly`. Right now, we're just introducing a mess by using both annotations without any documented guidelines.

kirklund (changes requested):

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/tier/sockets/ClientServerConnectDisconnectDistributedTest.java
jinmeiliao: done

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/tier/sockets/ClientServerConnectDisconnectDistributedTest.java
jinmeiliao: done

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/CacheClientProxy.java
jinmeiliao: done

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientUserAuths.java
jinmeiliao: `@VisibleForTesting` tag was pre-existing, I added the `@TestOnly` because I found out it's only used for test, so `TestOnly` should be more appropriate. I removed the `VisibleForTesting` tag.

kirklund (approved):

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientUserAuths.java
```diff
@@ -199,9 +231,12 @@ public void cleanup(boolean fromCacheClientProxy) {
       }
     }
 
-    // Logout the subjects
-    for (final Long subjectId : uniqueIdVsSubject.keySet()) {
-      removeSubject(subjectId);
+    // for integrated security, doesn't matter if this is called from proxy
+    // or from the connection, we are closing the client connection
+    synchronized (this) {
```
agingade, line 236: As Jianxia pointed out earlier, use of synchronization can be avoided by using Concurrent Data structures (CopyOnWriteArrayList); any reason why we can not use them. 
Also, if using synchronized, can the synchronization be done with/on the "uniqueIdVsSubjects" instead on "this".

agingade (changes requested):

nabarunnag (approved):

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/ClientUserAuths.java
jinmeiliao: We want to prevent multiple threads calling this block of code and "removeSubject" at the same time, using concurrent data structure wouldn't guarantee that.

if we synchronize on "uniqueIdVsSubjects" here, we can't simply add "synchronize" on the other methods, since the other methods are synchronizing on the "this" object if we put "synchronize" on the method name.

agingade (approved):

PR #7067, opened 2021-11-01
GEODE-9792: avoid sending in multiple authentication request on the same connection by different threads

jinmeiliao:
It's observed that AuthenticationRequest will come in multiple times in the same connection by different threads. This PR would fix that.

### DIFF
--- a/geode-core/src/test/java/org/apache/geode/cache/client/internal/ConnectionFactoryImplTest.java
+++ b/geode-core/src/test/java/org/apache/geode/cache/client/internal/ConnectionFactoryImplTest.java
@@ -18,16 +18,23 @@
 package org.apache.geode.cache.client.internal;
 
 import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyLong;
+import static org.mockito.Mockito.doCallRealMethod;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.times;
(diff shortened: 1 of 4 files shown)


kirklund (approved):

jchen21 (approved):

boglesby (approved):

jmelchio (approved):

geode-core/src/test/java/org/apache/geode/cache/client/internal/ConnectionFactoryImplTest.java
```diff
@@ -100,4 +107,32 @@ public void authenticateIfRequired() {
...
+    CompletableFuture<Void> future1 =
+        executor.runAsync(() -> factory.authenticateIfRequired(connection));
+    CompletableFuture<Void> future2 =
+        executor.runAsync(() -> factory.authenticateIfRequired(connection));
+    future1.get();
+    future2.get();
+
+    // setUserId should only be called once by this two threads. The other time
+    // is called by the test setup.
+    verify(server, times(2)).setUserId(anyLong());
```
DonalEvans, line 136: I don't know if it's actually better, since clearing invocations on mocks is not a best practice, but to make this verification more intuitive, you can call `Mockito.clearInvocations(server);` on line 126 and then have this verification just be: `verify(server).setUserId(anyLong());` and remove the comment explaining why it's 2.

DonalEvans (approved):
Looks good, just a couple of small optional suggestions.

PR #7072, opened 2021-11-02
GEODE-9794: Radish PassiveExpirationManager can throw NPE

jdeppe-pivotal:
### For all changes:


DonalEvans (approved):
I'm a little confused as to how we get null values in the Redis region in the first place, since in theory, every key should have a RedisData object associated with it, otherwise the key gets removed. Is this a race condition where the region is being modified while we're attempting to expire a key?

dschneider-pivotal (approved):

PR #7076, opened 2021-11-03
GEODE-9795: allow opt-in to run benchmarks on a PR via github label

onichols-pivotal:
Just like you can already add the `windows` label to a PR to also run the Windows tests, and the `jdk8` label to run additional JDK8 checks; this lets you add the `benchmarks` label to also run the Benchmarks jobs on your PR.

### DIFF
--- a/ci/pipelines/pull-request/jinja.template.yml
+++ b/ci/pipelines/pull-request/jinja.template.yml
@@ -27,6 +27,10 @@ groups:
   - {{test.name}}-test-{{java_test_version.name}}
   {%- endfor %}
 {%- endfor %}
+{%- for run_var in (benchmarks.flavors) %}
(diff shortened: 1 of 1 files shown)


rhoughton-pivot (approved):
Looks good, and might help keep the develop pipeline clean if we can get the right PRs to have this label. Curious how much similarity there is between the `develop` and `pr` pipelines scripts now, for a little DRY refactor

dickcav (approved):

onichols-pivotal:
this doesn't work because benchmarks jobs were designed to re-check-out the geode source, not use the source provided by concourse.  so without a major overhaul of the way benchmarks works to use rsync instead, this is unfortunately not viable.

PR #7080, opened 2021-11-04
GEODE-9781: clean up radish sys props

dschneider-pivotal:
### For all changes:


geode-for-redis/src/main/java/org/apache/geode/redis/internal/RedisProperties.java
```diff
@@ -16,38 +16,90 @@
...
+    }
+  }
 
   public static String getStringSystemProperty(String propName, String defaultValue) {
-    String geodeValue = System.getProperty("geode." + propName, defaultValue);
-    String gemfireValue = System.getProperty("gemfire." + propName, defaultValue);
+    validatePropertyName(propName);
 
+    String geodeValue = System.getProperty(propName, defaultValue);
     if (StringUtils.isNotEmpty(geodeValue) && !geodeValue.equals(defaultValue)) {
```
DonalEvans, line 51: Do we actually need to check if the value here is equal the the default? If the property exists, we should just return it; checking if it's equal to the default just results in extra work and the same outcome in the end. This comment also applies to the check on line 56.

geode-for-redis/src/main/java/org/apache/geode/redis/internal/netty/ByteToCommandDecoder.java
```diff
@@ -51,16 +53,16 @@
...
+      RedisProperties.UNAUTHENTICATED_MAX_BULK_STRING_LENGTH;
 
   private static final int MAX_BULK_STRING_LENGTH = 512 * 1024 * 1024; // 512 MB
   // These 2 defaults are taken from native Redis
   public static final int UNAUTHENTICATED_MAX_ARRAY_SIZE =
-      Integer.getInteger(UNAUTHENTICATED_MAX_ARRAY_SIZE_PARAM, 10);
+      getIntegerSystemProperty(UNAUTHENTICATED_MAX_ARRAY_SIZE_PARAM, 10, 1);
   public static final int UNAUTHENTICATED_MAX_BULK_STRING_LENGTH =
-      Integer.getInteger(UNAUTHENTICATED_MAX_BULK_STRING_LENGTH_PARAM, 16384);
+      getIntegerSystemProperty(UNAUTHENTICATED_MAX_BULK_STRING_LENGTH_PARAM, 16384, 1);
```
DonalEvans, line 65: Could `UNAUTHENTICATED_MAX_ARRAY_SIZE_PARAM` and `UNAUTHENTICATED_MAX_BULK_STRING_LENGTH_PARAM` be replaced directly with `UNAUTHENTICATED_MAX_ARRAY_SIZE` and `UNAUTHENTICATED_MAX_BULK_STRING_LENGTH`? Having the same constant duplicated in two classes seems unnecessary.

DonalEvans (approved):
Looks good, just a couple of small optional comments.

geode-for-redis/src/main/java/org/apache/geode/redis/internal/RedisProperties.java
dschneider-pivotal: Are you suggesting that it should use System.getProperty(propName) instead of System.getProperty(propName, defaultValue)? I think that it is a good idea since it is simpler and since we never end up using geodeValue if it is defaultValue.

geode-for-redis/src/main/java/org/apache/geode/redis/internal/netty/ByteToCommandDecoder.java
dschneider-pivotal: They could. At the time I didn't because they are public and I worried that they were used outside of the geode code base. But they probably are not so I'll try getting rid of them.

DonalEvans (approved):

PR #7083, opened 2021-11-05
Wip/wan tx grouping module 1

albertogpz:
### For all changes:


geode-wan-txgrouping/src/distributedTest/java/org/apache/geode/internal/cache/wan/txgrouping/parallel/TxGroupingParallelDUnitTest.java
```diff
@@ -0,0 +1,506 @@
...
+    try {
+      GatewaySender sender = getGatewaySender(senderId);
+      final AbstractGatewaySender abstractSender = (AbstractGatewaySender) sender;
+      await().untilAsserted(() -> assertThat(abstractSender.getEventQueueSize()).isEqualTo(0));
+      await().untilAsserted(
+          () -> assertThat(abstractSender.getSecondaryEventQueueSize()).isEqualTo(0));
+    } finally {
+      exp.remove();
+      exp1.remove();
+    }
```
kirklund, line 470: `IgnoredException` implements `AutoCloseable` so you can use it in try-with-resources syntax. Using import-static and `addIgnoredException(Class)` tidies it up nicely:
```

try (IgnoredException ie1 = addIgnoredException(RegionDestroyedException.class);
     IgnoredException ie2 = addIgnoredException(ForceReattemptException.class)) {
  // ...
}

File: geode-wan-txgrouping/src/distributedTest/java/org/apache/geode/internal/cache/wan/txgrouping/serial/TxGroupingSerialDUnitTest.java [hunk 2]
```diff
@@ -0,0 +1,366 @@
...
+  }
+
+  private void checkStats_Failover(String senderId, final int eventsReceived) {
+    GatewaySenderStats statistics = getGatewaySenderStats(senderId);
+    assertThat(statistics.getEventsReceived()).isEqualTo(eventsReceived);
+    assertThat((statistics.getEventsQueued() + statistics.getUnprocessedTokensAddedByPrimary()
+        + statistics.getUnprocessedEventsRemovedByPrimary())).isEqualTo(eventsReceived);
+  }
+
+  private Boolean killPrimarySender(String senderId) {
```
kirklund, line 347: The dunit testing framework originally required primitive wrappers, so I think you may have copied some of this from elsewhere. It's better to just use simple primitives for return types and parameters now, such as `boolean` instead of `Boolean`.

geode-wan-txgrouping/src/main/java/org/apache/geode/cache/wan/internal/txgrouping/CommonTxGroupingGatewaySenderFactory.java
```diff
@@ -0,0 +1,36 @@
...
+
+import static java.lang.String.format;
+
+import org.jetbrains.annotations.NotNull;
+
+import org.apache.geode.cache.wan.internal.GatewaySenderTypeFactory;
+import org.apache.geode.internal.cache.wan.GatewaySenderException;
+import org.apache.geode.internal.cache.wan.MutableGatewaySenderAttributes;
+
+public abstract class CommonTxGroupingGatewaySenderFactory {
```
kirklund, line 26: I think you should make this an interface instead of an abstract class. The `static` validator method would then become a `default` implementation which would probably only ever be overridden by a test.

geode-wan-txgrouping/src/main/java/org/apache/geode/cache/wan/internal/txgrouping/parallel/TxGroupingParallelGatewaySenderQueue.java
```diff
@@ -0,0 +1,171 @@
...
+    int retries = 0;
+    while (true) {
+      for (Iterator<Map.Entry<TransactionId, Integer>> iter =
+          incompleteTransactionIdsInBatch.entrySet().iterator(); iter.hasNext();) {
+        Map.Entry<TransactionId, Integer> pendingTransaction = iter.next();
+        TransactionId transactionId = pendingTransaction.getKey();
+        int bucketId = pendingTransaction.getValue();
+        List<Object> events =
+            peekEventsWithTransactionId(partitionedRegion, bucketId, transactionId);
+        for (Object object : events) {
```
kirklund, line 72: Try to extract any nested loops to their own smaller methods. It helps keep things better organized and more readable.

kirklund (commented):
Inlined a few comments. Overall looks like a good direction! Here are some general comments and tips:

1. Prefer interfaces over abstract methods.
2. Standardize on one library instead of mixing them (such as AssertJ for assertions).
3. Is there a better word than "txgrouping" for module name and packages?
4. Declare variables and APIs with interfaces rather than implementations.
5. Always look for an interface to use instead of an implementation (such as InternalCache instead of GemFireCacheImpl) in both tests and product code.
6. Always minimize the scope of anything (class, method, variable, constant).

geode-wan-txgrouping/src/main/java/org/apache/geode/cache/wan/internal/txgrouping/serial/TxGroupingSerialGatewaySenderQueue.java
```diff
@@ -126,4 +148,110 @@ protected void postProcessBatch(final List<AsyncEvent<?, ?>> batch, final long l
...
+          break;
+        }
+      }
+    }
+
+    return elementsMatching;
+  }
+
+  @Override
+  public synchronized void remove() throws CacheException {
```
pivotal-jbarrett, line 186: At first glance it looks like a lot of the remove method content is reimplemented here. Is there a way break up the original with some extension points to deduce the duplication?

geode-wan/src/main/java/org/apache/geode/cache/wan/internal/GatewaySenderFactoryImpl.java
```diff
@@ -256,7 +257,13 @@ public GatewaySenderFactory setEnforceThreadsConnectSameReceiver(
 
   static void validate(final @NotNull InternalCache cache,
       final @NotNull GatewaySenderAttributesImpl attributes) {
-    final int myDSId = cache.getDistributionManager().getDistributedSystemId();
+    int myDSId;
+    if (cache instanceof CacheCreation) {
```
pivotal-jbarrett, line 261: Maybe `CacheCreation` should be fixed to behave like any other `InternalCache` rather than making consumer know subtile differences in behavior.

geode-wan-txgrouping/src/distributedTest/java/org/apache/geode/internal/cache/wan/txgrouping/cli/commands/CreateTxGroupingGatewaySenderDUnitTest.java
```diff
@@ -0,0 +1,109 @@
...
+    cluster.startServerVM(1, new Properties(), locator.getPort());
+    gfsh.connectAndVerify(locator);
+  }
+
+  @Test
+  public void createTxGroupingParallelGatewaySender() {
+    addIgnoredException("could not get remote locator");
+
+    String createCommandString =
+        "create gateway-sender --id=sender1 --remote-distributed-system-id=1 --parallel --group-transaction-events=true";
```
pivotal-jbarrett, line 53: This doesn't seem modular yet. Is this testing deprecated `--group-transaction-events` and work is pending to add at `--type=TxGroupingParallelGatewaySender`?

pivotal-jbarrett (changes requested):

geode-wan-txgrouping/src/main/java/org/apache/geode/cache/wan/internal/txgrouping/serial/TxGroupingSerialGatewaySenderQueue.java
pivotal-jbarrett: See https://github.com/Nordix/geode/pull/17

pivotal-jbarrett:
@albertogpz Can you please rebase this from develop. There as a fixed added there to address issues with Gradle builds.

pivotal-jbarrett:
@albertogpz we need to extract out the TX batching specific elements form the `GatewaySenderEventImpl` object. I replaced it with a implementation defined [metadata](https://github.com/pivotal-jbarrett/geode/blob/wip/wan-spi/geode-core/src/main/java/org/apache/geode/internal/cache/wan/GatewaySenderEventImpl.java#L183) object.  See [`TxBatchMetadata`](https://github.com/pivotal-jbarrett/geode/blob/wip/wan-spi/geode-wan-txbatch/src/main/java/org/apache/geode/cache/wan/internal/txbatch/TxBatchMetadata.java) from my previous example on making this a module. This gets set on the event object when constructed via an override on [AbstractGatewaySenderEventProcessor.java](https://github.com/pivotal-jbarrett/geode/blob/wip/wan-spi/geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySenderEventProcessor.java#L1310). See example [here](https://github.com/pivotal-jbarrett/geode/blob/wip/wan-spi/geode-wan-txbatch/src/main/java/org/apache/geode/cache/wan/internal/txbatch/parallel/TxBatchRemoteParallelGatewaySenderEventProcessor.java#L49).

geode-wan-txgrouping/src/distributedTest/java/org/apache/geode/internal/cache/wan/txgrouping/cli/commands/CreateTxGroupingGatewaySenderDUnitTest.java
pivotal-jbarrett: You can find an example of modularizing this in the first POC. Follow the new `type` property on the `create gateway-sender` command. 
https://github.com/pivotal-jbarrett/geode/blob/wip/wan-spi/geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/CreateGatewaySenderCommand.java#L83

geode-wan-txgrouping/src/main/java/org/apache/geode/cache/wan/internal/txgrouping/CommonTxGroupingGatewaySenderFactory.java
albertogpz: This is a utility class and its `validator` method can be called statically.
If I make it an interface I would have to create an instance of it in order to invoke it.
What would be the benefit?

albertogpz:
> Inlined a few comments. Overall looks like a good direction! Here are some general comments and tips:
> 
>     1. Prefer interfaces over abstract methods.
> 
>     2. Standardize on one library instead of mixing them (such as AssertJ for assertions).
> 
>     3. Is there a better word than "txgrouping" for module name and packages?
> 
>     4. Declare variables and APIs with interfaces rather than implementations.
> 
>     5. Always look for an interface to use instead of an implementation (such as InternalCache instead of GemFireCacheImpl) in both tests and product code.
> 
>     6. Always minimize the scope of anything (class, method, variable, constant).

Thanks for your comments.
Regarding the name, I do not have one that I really like. Other alternatives could be: txeventsgrouping, grouptxevents, txaware,...
If you like any of these or have any other proposal, please, let me know.

geode-wan-txgrouping/src/main/java/org/apache/geode/cache/wan/internal/txgrouping/serial/TxGroupingSerialGatewaySenderQueue.java
albertogpz: Nice!

geode-wan/src/main/java/org/apache/geode/cache/wan/internal/GatewaySenderFactoryImpl.java
albertogpz: You mean to implement in `CacheCreation` the `getDistributionManager()` method as follows?
```
  @Override
  public DistributionManager getDistributionManager() {
    return InternalDistributedSystem.getAnyInstance().getDistributionManager()
  }

```

albertogpz:
> @albertogpz we need to extract out the TX batching specific elements form the `GatewaySenderEventImpl` object. I replaced it with a implementation defined [metadata](https://github.com/pivotal-jbarrett/geode/blob/wip/wan-spi/geode-core/src/main/java/org/apache/geode/internal/cache/wan/GatewaySenderEventImpl.java#L183) object. See [`TxBatchMetadata`](https://github.com/pivotal-jbarrett/geode/blob/wip/wan-spi/geode-wan-txbatch/src/main/java/org/apache/geode/cache/wan/internal/txbatch/TxBatchMetadata.java) from my previous example on making this a module. This gets set on the event object when constructed via an override on [`AbstractGatewaySenderEventProcessor`](https://github.com/pivotal-jbarrett/geode/blob/wip/wan-spi/geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySenderEventProcessor.java#L1310). See example [here](https://github.com/pivotal-jbarrett/geode/blob/wip/wan-spi/geode-wan-txbatch/src/main/java/org/apache/geode/cache/wan/internal/txbatch/parallel/TxBatchRemoteParallelGatewaySenderEventProcessor.java#L49).

 I thought about modularizing the GatewaySenderEventImpl class so that the specifics of the feature were taken to a TxGroupingGatewaySenderEventImpl subclass containing the transactionId and the isLastEventInTransaction fields. Nevertheless given that version 1.14 of GatewaySenderEventImpl already contains the changes, if we went on that direction we would break backward compatibility with version 1.14.
This would have been the right move if I had implemented the feature in a modular way from the beginning.



> @albertogpz we need to extract out the TX batching specific elements form the `GatewaySenderEventImpl` object. I replaced it with a implementation defined [metadata](https://github.com/pivotal-jbarrett/geode/blob/wip/wan-spi/geode-core/src/main/java/org/apache/geode/internal/cache/wan/GatewaySenderEventImpl.java#L183) object. See [`TxBatchMetadata`](https://github.com/pivotal-jbarrett/geode/blob/wip/wan-spi/geode-wan-txbatch/src/main/java/org/apache/geode/cache/wan/internal/txbatch/TxBatchMetadata.java) from my previous example on making this a module. This gets set on the event object when constructed via an override on [`AbstractGatewaySenderEventProcessor`](https://github.com/pivotal-jbarrett/geode/blob/wip/wan-spi/geode-core/src/main/java/org/apache/geode/internal/cache/wan/AbstractGatewaySenderEventProcessor.java#L1310). See example [here](https://github.com/pivotal-jbarrett/geode/blob/wip/wan-spi/geode-wan-txbatch/src/main/java/org/apache/geode/cache/wan/internal/txbatch/parallel/TxBatchRemoteParallelGatewaySenderEventProcessor.java#L49).

@pivotal-jbarrett As I said in a previous e-mail, I do not think we can make these changes without breaking backward compatibility. The feature was production ready at 1.14 so we can expect that it is used by customers. So, we cannot just ignore those changes when interworking with 1.14 servers.
Am I missing something?

geode-wan-txgrouping/src/distributedTest/java/org/apache/geode/internal/cache/wan/txgrouping/cli/commands/CreateTxGroupingGatewaySenderDUnitTest.java
albertogpz: Done. We'll have to see how we reflect this in the documentation as we will need documentation specific for the module.

onichols-pivotal:

albertogpz:
> this PR appears to be abandoned, can it be closed?

It is not abandoned. Still waiting for comments after review changes. I have rebased the branch to remove the conflicts with develop.

semioticrobotic:
@albertogpz and @kirklund: I know this has been dormant for a bit, but I believe it still has merit and would certainly be happy to see it included in a future release. @JinwooHwang and I have been reviewing outstanding PRs and are wondering if you have a moment to return to this one and give us your assessment. Is this still something you'd like to see merged?

PR #7094, opened 2021-11-09
GEODE-9805: Do not log arguments of Radish AUTH command

DonalEvans:
- Rather than returning the arguments of the AUTH command, only return
 how many arguments there are when toString() is called

### For all changes:

### DIFF
--- a/geode-for-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/connection/AuthIntegrationTest.java
+++ b/geode-for-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/connection/AuthIntegrationTest.java
@@ -111,7 +111,6 @@ public class AuthIntegrationTest extends AbstractAuthIntegrationTest {
      * setting this value.
      */
     System.setProperty("io.netty.eventLoopThreads", "1");
-    port = AvailablePortHelper.getRandomAvailableTCPPort();
     CacheFactory cf = new CacheFactory();
     cf.set(LOG_LEVEL, "error");
     cf.set(MCAST_PORT, "0");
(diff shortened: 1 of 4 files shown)


dschneider-pivotal (approved):

geode-for-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/connection/AuthIntegrationTest.java
```diff
@@ -227,4 +244,85 @@ public void givenSecurityWithWritePermission_getCommandFails() throws Exception
...
+    jedis.auth(LOGGED_PASSWORD);
+    assertThatThrownBy(() -> jedis.auth(LOGGED_PASSWORD))
+        .hasMessageContaining(SERVER_ERROR_MESSAGE);
+
+    checkLogFileForPassword(logFile, LOGGED_PASSWORD);
+  }
+
+  private void createRadishServerWithLogFileAndSecurityManager(File logFile, String logLevel,
+      Class<?> securityManager) {
+    System.setProperty("io.netty.eventLoopThreads", "1");
```
jdeppe-pivotal, line 277: This isn't necessary.

geode-for-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/connection/AuthIntegrationTest.java
```diff
@@ -227,4 +244,85 @@ public void givenSecurityWithWritePermission_getCommandFails() throws Exception
...
+    port = AvailablePortHelper.getRandomAvailableTCPPort();
```
jdeppe-pivotal, line 278: I think it would be safer to move this down one line. I think there's a possibility that the cache creation would end up using the port just deemed free here.

jdeppe-pivotal (changes requested):

geode-for-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/connection/AuthIntegrationTest.java
DonalEvans: Removed

geode-for-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/connection/AuthIntegrationTest.java
DonalEvans: Moved

jdeppe-pivotal (approved):

PR #7103, opened 2021-11-10
GEODE-9851:  Make InterestType and DataPolicy enums.

pivotal-jbarrett:
### For all changes:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/InitialImageOperation.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/InitialImageOperation.java
@@ -51,6 +51,7 @@ import org.apache.geode.SystemFailure;
 import org.apache.geode.annotations.Immutable;
 import org.apache.geode.annotations.VisibleForTesting;
 import org.apache.geode.annotations.internal.MutableForTesting;
+import org.apache.geode.cache.DataPolicy;
 import org.apache.geode.cache.DiskAccessException;
 import org.apache.geode.cache.RegionDestroyedException;
 import org.apache.geode.cache.query.internal.CqStateImpl;
@@ -3748,7 +3749,7 @@ public class InitialImageOperation {
               // found.
               cqService.executeCq(e.getKey(), cq.getQueryString(),
                   ((CqStateImpl) cq.getState()).getState(), proxy.getProxyID(), ccn, cq.isDurable(),
-                  true, -1, emptyRegionMap);
+                  true, null, emptyRegionMap);
             } catch (Exception ex) {
               logger.info("Failed to register CQ during HARegion GII. CQ: {} {}", e.getKey(),
                   ex.getMessage(), ex);
@@ -3876,16 +3877,17 @@ public class InitialImageOperation {
             regionsWithInterest.add(regionName);
             if (allKey) {
               ccn.registerClientInterest(regionName, e.getValue(), proxy.getProxyID(), interestType,
-                  isDurable, updatesAsInvalidates, manageEmptyRegions, 0, false);
+                  isDurable, updatesAsInvalidates, manageEmptyRegions, DataPolicy.EMPTY, false);
             } else if (InterestType.REGULAR_EXPRESSION == interestType) {
               for (final Object o : asSet(e.getValue())) {
                 ccn.registerClientInterest(regionName, o, proxy.getProxyID(),
-                    interestType, isDurable, updatesAsInvalidates, manageEmptyRegions, 0, false);
+                    interestType, isDurable, updatesAsInvalidates, manageEmptyRegions,
+                    DataPolicy.EMPTY, false);
               }
             } else {
               ccn.registerClientInterest(regionName, new ArrayList<>(asSet(e.getValue())),
                   proxy.getProxyID(), isDurable, updatesAsInvalidates, manageEmptyRegions,
-                  /* TODO BUG!! interestType */ 0, false);
+                  DataPolicy.EMPTY, false);
             }
           }
         }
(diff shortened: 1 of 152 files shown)


jinmeiliao (approved):
LGTM

kirklund (approved):

mhansonp (approved):

dschneider-pivotal (approved):
The title of this PR is confusing. You didn't "make InterestType and DataPolicy enums". I think you should correct this in the description of GEODE-9851. My understanding is you are basically changing how these existing classes are used internally in geode by using an instance of the class instead of its ordinal value. But I didn't see any introduction of Java Enum instances.

pivotal-jbarrett:
> The title of this PR is confusing. You didn't "make InterestType and DataPolicy enums". I think you should correct this in the description of GEODE-9851. My understanding is you are basically changing how these existing classes are used internally in geode by using an instance of the class instead of its ordinal value. But I didn't see any introduction of Java Enum instances.

Good point.
`InterestType` was converted to an `enum` [here](https://github.com/apache/geode/pull/7103/files#diff-2dc76d4b46e63d7bdcd6debc3978915ee77d063a109d68c7e23a0ed3f9c58d47R23). The public version of this effectively same "enum" is unchanged to avoid API and ABI issues.
`DataPolicy` is public and couldn't be safely changed to an `enum` without breaking ABI, though I am still testing this issue. The API can largely be preserved upon conversion to an `enum`. In the mean time I have deprecated the methods that would be replaced by `enum` generated methods in the future, say Geode 2.0 where we can break the ABI. 

I will work on the description for a better commit message.

geode-core/src/main/java/org/apache/geode/internal/cache/InitialImageOperation.java
```diff
@@ -3874,16 +3877,17 @@ private void registerInterestKeys(Map<String, ?> regionKeys, boolean allKey, Loc
...
                 ccn.registerClientInterest(regionName, o, proxy.getProxyID(),
-                    interestType, isDurable, updatesAsInvalidates, manageEmptyRegions, 0, false);
+                    interestType, isDurable, updatesAsInvalidates, manageEmptyRegions,
+                    DataPolicy.EMPTY, false);
               }
             } else {
               ccn.registerClientInterest(regionName, new ArrayList<>(asSet(e.getValue())),
                   proxy.getProxyID(), isDurable, updatesAsInvalidates, manageEmptyRegions,
-                  interestType, false);
+                  DataPolicy.EMPTY, false);
```
DonalEvans, line 3890: This seems like a change in behaviour. Previously the value passed as the `regionDataPolicy` argument to `registerClientInterest()` wasn't fixed, as `interestType` could take values 0-4 (although if it was 1 then the `InterestType.REGULAR_EXPRESSION == interestType` check above would prevent us getting to here). Was this just a bug in the old code?

onichols-pivotal (approved):

geode-core/src/main/java/org/apache/geode/internal/cache/InitialImageOperation.java
pivotal-jbarrett: @DonalEvans winner winner chicken dinner! You found the bug that validates the need for the PR and the advocation for strong types, like enums, over weaker types, like int. If you look at this method, `registerClientInterest` has overrides that take parameters in slightly different orders but also takes an `int`. In one this `int` was the `InterestType` ordinal and the other it was expected to be, and is later converted to, `DataPolicy` ordinal. You will notice that the compiler was just fine with the bug that that `interestType` was passed for the `dataPolicy` since both were `int`. By converting the parameters to strong types we find this bug at compile time.

DonalEvans (approved):
Approving the files for which I'm a code owner.

mkevo (approved):

geode-core/src/main/java/org/apache/geode/internal/cache/tier/InterestType.java
```diff
@@ -14,36 +14,36 @@
...
 /**
- * Pre-defined interest list types supported by the system
- *
- * @version $Revision: 13166 $
- * @since GemFire 2.0.2
+ * Pre-defined interest list types supported by the system.
+ * Partial duplicate of {@link org.apache.geode.cache.operations.InterestType}.
  */
-public class InterestType {
+public enum InterestType {
```
gesterzhou, line 24: Maybe use Key(0), REGULAR_EXPRESSION(1) ...; int value; InterestType(int value) { this.value=value; } is a little bit better?

Just comments.

gesterzhou (approved):

nabarunnag (approved):

geode-core/src/main/java/org/apache/geode/internal/cache/InterestRegistrationEventImpl.java
```diff
@@ -73,84 +75,81 @@ public InterestRegistrationEventImpl() {}
...
-  public Set getKeysOfInterest() {
-    return this.keysOfInterest;
+  public Set<?> getKeysOfInterest() {
+    return keysOfInterest;
   }
 
   @Override
   public int getInterestType() {
-    return this.interestType;
+    return interestType;
```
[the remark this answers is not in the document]
pivotal-jbarrett: And here is a problem with a PR that has been waiting 2 weeks for reviews. I recall there was a good reason not to change it here but I can't recall. Let me dig into this again. This time I will leave a comment as to why it needs to stay an `int` here.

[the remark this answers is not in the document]
pivotal-jbarrett: Ok, this is the implementation of a public interface that has a method that returns the `InterestType` ordinal as `int`. We can't create a new method that returns the enum `InterestType` because it is internal. We could create a new method that returns the public enum-like `org.apache.geode.cache.operations.InterestType` but that felt like a ballooning scope.

PR #7107, opened 2021-11-11
GEODE-9806: added skip recovery of recovered bucket

mivanac:
### For all changes:

### DIFF
--- a/geode-core/src/test/java/org/apache/geode/internal/cache/ProxyBucketRegionTest.java
+++ b/geode-core/src/test/java/org/apache/geode/internal/cache/ProxyBucketRegionTest.java
@@ -15,11 +15,25 @@
 package org.apache.geode.internal.cache;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyBoolean;
+import static org.mockito.ArgumentMatchers.anyInt;
+import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
 import org.junit.Test;
 
+import org.apache.geode.CancelCriterion;
+import org.apache.geode.cache.DataPolicy;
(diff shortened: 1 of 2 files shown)


DonalEvans:
This PR needs tests to show that the new behaviour is working as expected. The JIRA ticket could also use some additional details, such as how to reproduce the issue.

geode-core/src/test/java/org/apache/geode/internal/cache/ProxyBucketRegionTest.java
```diff
@@ -32,4 +46,54 @@ public void shouldBeMockable() throws Exception {
 
     assertThat(mockProxyBucketRegion.getBucketAdvisor()).isSameAs(mockBucketAdvisor);
   }
+
+  @Test
+  public void testRecoverFromDisk() throws Exception {
```
DonalEvans, line 51: This test name needs to be more descriptive. It should say what behaviour is under test, what the conditions are, and what the expected behaviour is. Also, an exception is never thrown from this method, so this can be removed.

geode-core/src/test/java/org/apache/geode/internal/cache/ProxyBucketRegionTest.java
```diff
@@ -32,4 +46,54 @@ public void shouldBeMockable() throws Exception {
...
+    when(cache.getCancelCriterion()).thenReturn(cancel);
+    when(regionAdvisor.isInitialized()).thenReturn(true);
+
+    ProxyBucketRegion proxyBucketRegion =
+        new ProxyBucketRegion(0, partitionedRegion, internalRegionArguments);
+
+    proxyBucketRegion.recoverFromDisk();
+    proxyBucketRegion.recoverFromDisk();
+    verify(regionAdvisor, times(1))
+        .isInitialized();
```
DonalEvans, line 96: It's not at all clear to me what this is verifying, and this test passes even with the changes in `ProxyBucketRegion` reverted, so it doesn't seem to be testing the new behaviour. I feel like for this change it might be good to add an integration test that shows the changed behaviour at a higher level, by actually creating a colocated region and showing that already recovered buckets are not recovered again.

DonalEvans (changes requested):

DonalEvans (approved):

PR #7108, opened 2021-11-12
GEODE-9807: gfsh gateway sender stop in parallel (like start)

albertogpz:
### For all changes:


mhansonp (changes requested):
I don't see any unit tests of this functionality. The existing serial tests would not be sufficient.

Tests that show the threads get executed and cleaned up properly would be good regardless of the actual outcomes of the calls (failure, success).

jinmeiliao:
same comments as @mhansonp . Would need some unit tests for the refactor.

albertogpz:
> same comments as @mhansonp . Would need some unit tests for the refactor.

@mhansonp and @jinmeiliao I have added unit test cases. Anyhow, I was not very fond of adding them because in order to add them, I have open up the implementation by adding a new public method (getExecutorService()) to be able to override it in the spy I am using in the test.
I would have preferred to add a new constructor for testing to the class in which I could inject the dependencies for testing but it was not possible with gfsh command classes (at least, I did not find a way to do it).

geode-gfsh/src/test/java/org/apache/geode/management/internal/cli/commands/StopGatewaySenderCommandTest.java
```diff
@@ -0,0 +1,108 @@
...
+import org.junit.Test;
+import org.mockito.ArgumentCaptor;
+
+import org.apache.geode.distributed.DistributedMember;
+import org.apache.geode.management.ManagementService;
+import org.apache.geode.management.internal.SystemManagementService;
+import org.apache.geode.management.internal.functions.CliFunctionResult;
+import org.apache.geode.test.junit.rules.GfshParserRule;
+
+public class StopGatewaySenderCommandTest {
```
jinmeiliao, line 46: Nice test. Would be nice to add more tests to cover all branches. like `executorService.invokeAll` throws an exception, and mock up the list of future results to make sure it got displayed correctly ..

jinmeiliao (changes requested):

geode-gfsh/src/test/java/org/apache/geode/management/internal/cli/commands/StopGatewaySenderCommandTest.java
albertogpz: I refactored the code quite a bit (not sure if I over-engineered) in order to be able to test everything without breaking encapsulation.

onichols-pivotal (approved):

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/StopGatewaySenderOnMemberFactory.java
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package org.apache.geode.management.internal.cli.commands;
+
+interface StopGatewaySenderOnMemberFactory {
```
jinmeiliao, line 18: Is it really necessary to add those interfaces? Personally, I don't like adding more interfaces/implementations only for testing purposes. I thought change set was good, just missing a few more tests to cover all branches.  You can easily test your previous command changes by mocking up some dependencies.

geode-gfsh/src/test/java/org/apache/geode/management/internal/cli/commands/StopGatewaySenderCommandDelegateParallelImplTest.java
```diff
@@ -0,0 +1,126 @@
...
+import org.apache.geode.management.internal.cli.result.model.TabularResultModel;
+import org.apache.geode.management.internal.i18n.CliStrings;
+
+public class StopGatewaySenderCommandDelegateParallelImplTest {
+  private final String senderId = "sender1";
+  private Cache cache;
+  Set<DistributedMember> members;
+  ExecutorService executorService;
+  SystemManagementService managementService;
+  StopGatewaySenderOnMemberFactory stopperOnMemberFactory;
```
kirklund, line 54: These should be `private` unless you subclass the test itself in which case `protected` might be appropriate.

geode-gfsh/src/test/java/org/apache/geode/management/internal/cli/commands/StopGatewaySenderCommandTest.java
```diff
@@ -0,0 +1,91 @@
...
+  @ClassRule
+  public static GfshParserRule gfsh = new GfshParserRule();
+
+  private StopGatewaySenderCommand command;
+
+  @Before
+  public void before() {
+    command = spy(StopGatewaySenderCommand.class);
+    List<CliFunctionResult> functionResults = new ArrayList<>();
+    doReturn(functionResults).when(command).executeAndGetFunctionResult(any(), any(), any());
```
kirklund, line 51: Anytime you override an instance method on the component being tested, you're either not testing some part of the component or it suggests that the method represents a responsibility that should not be on this component.

You can avoid using `spy` and overriding the instance method by using a field of type `java.util.function.Function`, or a custom interface, that is injected into the constructor(s).

1. Add cleanResults field of type `java.util.function.Function` or custom inner-interface
```
private final java.util.function.Function<List<?>, List<CliFunctionResult>> cleanResults;
```
2. Add constructors to `GfshCommand`
```
  public GfshCommand() {
    this(CliFunctionResult::cleanResults);
  }

  protected GfshCommand(java.util.function.Function<List<?>, List<CliFunctionResult>> cleanResults) {
    this.cleanResults = cleanResults;
  }
```
3. Add constructors to `StopGatewaySenderCommand`
```
  @SuppressWarnings("unused") // invoked by spring shell
  public StopGatewaySenderCommand() {
    this(CliFunctionResult::cleanResults);
  }

  StopGatewaySenderCommand(Function<List<?>, List<CliFunctionResult>> cleanResults) {
    super(cleanResults);
  }
```
4. Change `GfshCommand.executeAndGetFunctionResult` to use the field `cleanResults`
```
  public List<CliFunctionResult> executeAndGetFunctionResult(Function<?> function, Object args,
      Set<DistributedMember> targetMembers) {
    ResultCollector<?, ?> rc = executeFunction(function, args, targetMembers);
    return cleanResults.apply((List<?>) rc.getResult());
  }
```
5. Pass in simple mock when constructing `StopGatewaySenderCommand` in test
```
Function<List<?>, List<CliFunctionResult>> cleanResults = mock(Function.class);
command = new StopGatewaySenderCommand(cleanResults);
```
If you need to observe `cleanResults` to validate interaction with that dependency, you simply move it from a local variable to a field:
```
private Function<List<?>, List<CliFunctionResult>> cleanResults;
// ...
cleanResults = mock(Function.class);
command = new StopGatewaySenderCommand(cleanResults);
```
Then you can perform any Mockito validation calls against `cleanResults` that you need to.

geode-gfsh/src/test/java/org/apache/geode/management/internal/cli/commands/StopGatewaySenderOnMemberWithBeanImplTest.java
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  @Test
+  @Parameters({"true", "false"})
+  public void executeStopGatewaySenderOnMemberNotRunningReturnsNotRunningError(
+      boolean isLocalMember) {
+    StopGatewaySenderOnMember stopperWithBean = new StopGatewaySenderOnMemberWithBeanImpl();
+    setUpMocks(isLocalMember, false);
+    doReturn(false).when(bean).isRunning();
+    ArrayList<String> result = stopperWithBean.executeStopGatewaySenderOnMember(senderId, cache,
```
kirklund, line 70: `List<String>`

geode-gfsh/src/test/java/org/apache/geode/management/internal/cli/commands/StopGatewaySenderOnMemberWithBeanImplTest.java
```diff
@@ -0,0 +1,113 @@
...
+        "GatewaySender sender1 is not running on member " + memberId + ".");
+  }
+
+  @Test
+  @Parameters({"true", "false"})
+  public void executeStopGatewaySenderOnMemberNotAvailableReturnsNotAvailableError(
+      boolean isLocalMember) {
+    StopGatewaySenderOnMember stopperWithBean = new StopGatewaySenderOnMemberWithBeanImpl();
+    setUpMocks(isLocalMember, true);
+    ArrayList<String> result = stopperWithBean.executeStopGatewaySenderOnMember(senderId, cache,
```
kirklund, line 82: `List<String>`

geode-gfsh/src/test/java/org/apache/geode/management/internal/cli/commands/StopGatewaySenderOnMemberWithBeanImplTest.java
```diff
@@ -0,0 +1,113 @@
...
+        "GatewaySender sender1 is not available on member " + memberId);
+  }
+
+  @Test
+  @Parameters({"true", "false"})
+  public void executeStopGatewaySenderOnMemberRunningReturnsOk(boolean isLocalMember) {
+    StopGatewaySenderOnMember stopperWithBean = new StopGatewaySenderOnMemberWithBeanImpl();
+    setUpMocks(isLocalMember, false);
+    doReturn(true).when(bean).isRunning();
+    ArrayList<String> result = stopperWithBean.executeStopGatewaySenderOnMember(senderId, cache,
```
kirklund, line 94: `List<String>`

geode-gfsh/src/test/java/org/apache/geode/management/internal/cli/commands/StopGatewaySenderOnMemberWithBeanImplTest.java
```diff
@@ -0,0 +1,113 @@
...
+        managementService, distributedMember);
+    assertThat(result).containsExactly(memberId, "OK",
+        "GatewaySender sender1 is stopped on member " + memberId);
+  }
+
+  private void setUpMocks(boolean isLocalMember, boolean beanMustBeNull) {
```
kirklund, line 100: How about more descriptive names like:
```
private GatewaySenderMXBean gatewaySenderMXBean(boolean isLocalMember, boolean mustBeNull) {
  GatewaySenderMXBean gatewaySenderMXBean = mustBeNull 
      ? null 
      : mock(GatewaySenderMXBean.class);
  // ...
  return gatewaySenderMXBean;
}
```
You can then delete the field `private GatewaySenderMXBean bean;` and change the tests do something like:
```
@Test
  @Parameters({"true", "false"})
  public void executeStopGatewaySenderOnMemberRunningReturnsOk(boolean isLocalMember) {
    GatewaySenderMXBean gatewaySenderMXBean = gatewaySenderMXBean(isLocalMember, false);
    when(gatewaySenderMXBean.isRunning()).thenReturn(true);

    StopGatewaySenderOnMember stopperWithBean = new StopGatewaySenderOnMemberWithBeanImpl();

    List<String> result = stopperWithBean.executeStopGatewaySenderOnMember(senderId, cache, managementService, distributedMember);

    assertThat(result)
        .containsExactly(memberId, "OK", "GatewaySender sender1 is stopped on member " + memberId);
  }
```

kirklund (changes requested):

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/StopGatewaySenderOnMemberFactory.java
albertogpz: @kirklund I went for the `Supplier` option.
@jinmeiliao I have removed some interfaces (by using the `Supplier` one as Kirk suggested) and I have moved the ones remaining to the class using them.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/StopGatewaySenderCommandDelegateParallelImpl.java
```diff
@@ -0,0 +1,111 @@
...
+    return buildResultModelFromMembersResponses(id, dsMembers, futures);
+  }
+
+  private ResultModel buildResultModelFromMembersResponses(String id,
+      Set<DistributedMember> dsMembers, List<Future<List<String>>> futures) {
+    ResultModel resultModel = new ResultModel();
+    TabularResultModel resultData = resultModel.addTable(CliStrings.STOP_GATEWAYSENDER);
+    Iterator<DistributedMember> memberIterator = dsMembers.iterator();
+    for (Future<List<String>> future : futures) {
+      DistributedMember member = memberIterator.next();
```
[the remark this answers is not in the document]
albertogpz: I have created a `List` from the `dsMembers` `Set` and I have passed it to the method so that we can be sure that the order between the `dsMembers` and the `futures` is the same.

geode-gfsh/src/test/java/org/apache/geode/management/internal/cli/commands/StopGatewaySenderCommandTest.java
albertogpz: @kirklund Thanks a lot for the suggestion. I finally have not needed to modify `GfshCommand`.
Nevertheless, I had to inject the command delegate and the `findMembers` test behavior using the constructor.
The outcome has not been very clean. If you have any suggestions to make it cleaner, please let me know.

Given that `findMembers` is not a static method, I could not pass it when calling the last constructor. Instead, I had to set it after I called it.
On the other hand, the `StopGatewaySenderCommandDelegateParallelImpl` constructor needed the a `ManagementSystem` which is obtained by calling the instance `getManagementSystem()` method. As a result, I had to set it to null in the Spring shell constructor and later set it to the right value if it was found null.

geode-gfsh/src/test/java/org/apache/geode/management/internal/cli/commands/StopGatewaySenderOnMemberWithBeanImplTest.java
albertogpz: Great suggestion!

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/StopGatewaySenderCommand.java
```diff
@@ -16,27 +16,40 @@
...
+  private BiFunction<String[], String[], Set<DistributedMember>> findMembers;
+
+  @SuppressWarnings("unused") // invoked by spring shell
+  public StopGatewaySenderCommand() {
+    this(null, null);
+    findMembers = this::findMembers;
+    commandDelegateSupplier = this::getCommandDelegate;
+  }
+
+  StopGatewaySenderCommand(Supplier<StopGatewaySenderCommandDelegate> commandDelegateSupplier,
```
jinmeiliao, line 46: is this test only?

jinmeiliao:
Normally the command classes are stateless, we don't usually have member variables for command classes. I guess I don't quite understand why you would want to have a delegate class to do all the work instead, why can't you just do all the work inside the command class, like every other command is doing? I hate to introduce another pattern for commands for no obvious reasons.

albertogpz:
> Normally the command classes are stateless, we don't usually have member variables for command classes. I guess I don't quite understand why you would want to have a delegate class to do all the work instead, why can't you just do all the work inside the command class, like every other command is doing? I hate to introduce another pattern for commands for no obvious reasons.

@jinmeiliao I have removed the delegate class. I agree with you, it is not needed. Regarding member variables, the ones added correspond to the dependencies that can be passed in the constructor to be able to unit test the class.

albertogpz:
> I don't see any unit tests of this functionality. The existing serial tests would not be sufficient.
> 
> Tests that show the threads get executed and cleaned up properly would be good regardless of the actual outcomes of the calls (failure, success).

@mhansonp Unit tests added.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/StopGatewaySenderCommand.java
albertogpz: I removed the delegate class. Now all unit tests are in `StopGatewaySenderCommandTest.java`.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/StopGatewaySenderCommand.java
```diff
@@ -36,7 +44,30 @@
...
 public class StopGatewaySenderCommand extends GfshCommand {
+  private final ExecutorService executorService;
+  private SystemManagementService managementService;
+  private BiFunction<String[], String[], Set<DistributedMember>> findMembers;
+  private final Supplier<StopGatewaySenderOnMember> stopperOnMemberFactory;
+
+  @SuppressWarnings("unused") // invoked by spring shell
+  public StopGatewaySenderCommand() {
+    this(newCachedThreadPool("Stop Sender Command Thread ", true),
+        StopGatewaySenderOnMemberWithBeanImpl::new, null, null);
```
jinmeiliao, line 55: `GfshCommand` has `findMember()` and `getManagementService()` you can stubbed out in your tests, I don't think you need the last two params in your "test constructor". Also, for the first two parameters, one you are passing the object itself, one is a supplier, I suggest you make it consistent, I would pass in both in object form (change the supplier to be an instance of `StopGatewaySenderOnMember`.

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/StopGatewaySenderCommand.java
```diff
@@ -36,7 +44,30 @@
...
+    findMembers = this::findMembers;
+  }
+
+  StopGatewaySenderCommand(
```
jinmeiliao, line 59: annotate this as `VisisbleForTesting`

jinmeiliao (changes requested):

geode-gfsh/src/main/java/org/apache/geode/management/internal/cli/commands/StopGatewaySenderCommand.java
albertogpz: I am passing the `SystemManagmentService` instance and the `findMembers()` function to avoid using a spy, which is not considered a good practice unless you are forced to because you are adding tests to legacy code.
I do not know in this case, given that the methods are provided by the base class, using a spy would be justified.

I have changed the code to pass to the constructor an instance of `StopGatewaySenderOnMember` instead of the supplier as you have suggested.

jinmeiliao: "Spying an object you are testing has been communicated to me as bad practice, but these warnings come with little to no explanation backing it up. I'm sure, like anything else, it can definitely be abused.". 
To me, in this case, spying on these methods to pass in a mock object is definitely better than passing null objects in the product calling code.

albertogpz: This is probably one of those cases in which using a spy is justified. I cannot give you strong arguments in this case that make the previous implementation without spy better, so I will follow your proposal.

jinmeiliao (approved):

mhansonp (approved):

rhoughton-pivot (approved):

gesterzhou (approved):

albertogpz:
@kirklund Do you agree with the changes done?

kirklund (approved):

PR #7111, opened 2021-11-15
GEODE-1537: Re-order ephemeral port acquisition to fix flaky DurableRegistrationDUnitTest

jdeppe-pivotal:
### For all changes:

### DIFF
--- a/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/tier/sockets/DurableRegistrationDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/tier/sockets/DurableRegistrationDUnitTest.java
@@ -117,12 +117,10 @@ public class DurableRegistrationDUnitTest extends JUnit4DistributedTestCase {
   public void testSimpleDurableClient() {
 
     // Step 1: Starting the servers
-    PORT1 = ((Integer) this.server1VM
-        .invoke(() -> CacheServerTestUtil.createCacheServer(regionName, new Boolean(true))))
-            .intValue();
-    PORT2 = ((Integer) this.server2VM
-        .invoke(() -> CacheServerTestUtil.createCacheServer(regionName, new Boolean(true))))
-            .intValue();
+    PORT1 = this.server1VM
+        .invoke(() -> CacheServerTestUtil.createCacheServer(regionName, Boolean.TRUE));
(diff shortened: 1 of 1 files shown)


geode-core/src/distributedTest/java/org/apache/geode/internal/cache/tier/sockets/DurableRegistrationDUnitTest.java
```diff
@@ -413,13 +411,10 @@ public void run2() throws CacheException {
...
     // Step 1: Start server1
+    PORT1 = this.server1VM
+        .invoke(() -> CacheServerTestUtil.createCacheServer(regionName, Boolean.TRUE));
     PORT2 = getRandomAvailableTCPPort();
 
-    PORT1 = ((Integer) this.server1VM
-        .invoke(() -> CacheServerTestUtil.createCacheServer(regionName, new Boolean(true))))
-            .intValue();
-
-
```
demery-pivotal, line 422: How does this fix the flakiness? Is `getRandomAvailableTCPPort()` returning ports in the ephemeral port range? In the latest CI failure reported in GEODE-1537, the port number seems to be in the available port range, and not in the ephemeral port range.

jdeppe-pivotal: I believe what's happening is that `PORT2` is assigned an available port but then the subsequent line, which starts a server, ends up using that port. Later on, when something else tries to use `PORT2` (for another server) a bind exception occurs because the port is already in use.

jdeppe-pivotal: This is just re-ordering the sequence so that the server can start up first and then an available port is retrieved.

agingade: Now I get it...Your comments are in relation to changes in line# 419 and line#519. Not related to line#120 :)

agingade (approved):

Bill (approved):
approved ✓

demery-pivotal (commented):
It looks as if I left my comments "pending," so you never saw them. 😧

PR #7113, opened 2021-11-16
GEODE-9809: stop monitoring of destroyed regions

mivanac:
### For all changes:

### DIFF
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/partitioned/PersistentBucketRecoverer.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/partitioned/PersistentBucketRecoverer.java
@@ -165,6 +165,9 @@ public class PersistentBucketRecoverer extends RecoveryRunnable implements Persi
     }
   }
 
+  List<RegionStatus> getRegions() {
+    return regions;
+  }
 
 
   /**
--- a/geode-core/src/main/java/org/apache/geode/internal/cache/partitioned/PersistentBucketRecoverer.java
+++ b/geode-core/src/main/java/org/apache/geode/internal/cache/partitioned/PersistentBucketRecoverer.java
@@ -17,7 +17,6 @@ package org.apache.geode.internal.cache.partitioned;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -29,6 +28,7 @@ import java.util.concurrent.TimeUnit;
 
 import org.apache.logging.log4j.Logger;
 
+import org.apache.geode.cache.RegionDestroyedException;
 import org.apache.geode.distributed.internal.membership.InternalDistributedMember;
 import org.apache.geode.internal.cache.BucketPersistenceAdvisor;
 import org.apache.geode.internal.cache.ColocationHelper;
@@ -100,7 +100,7 @@ public class PersistentBucketRecoverer extends RecoveryRunnable implements Persi
       }
     }
 
-    regions = Collections.unmodifiableList(allRegions);
+    regions = allRegions;
     allBucketsRecoveredFromDisk = new CountDownLatch(proxyBuckets);
     membershipChanged = true;
     addListeners();
@@ -174,6 +174,7 @@ public class PersistentBucketRecoverer extends RecoveryRunnable implements Persi
   @Override
   public void run2() {
     try {
+      List<RegionStatus> destroyedRegions = new ArrayList<>();
       boolean warningLogged = false;
       while (getLatchCount() > 0) {
         int sleepMillis = SLEEP_PERIOD;
@@ -182,12 +183,23 @@ public class PersistentBucketRecoverer extends RecoveryRunnable implements Persi
         if (!warningLogged) {
           sleepMillis = SLEEP_PERIOD / 2;
         }
+        if (regions.isEmpty()) {
+          break;
+        }
         Thread.sleep(sleepMillis);
 
         if (membershipChanged) {
           membershipChanged = false;
           for (RegionStatus region : regions) {
-            region.logWaitingForMembers();
+            try {
+              region.logWaitingForMembers();
+            } catch (RegionDestroyedException e) {
+              destroyedRegions.add(region);
+            }
+          }
+          if (!destroyedRegions.isEmpty()) {
+            regions.removeAll(destroyedRegions);
+            destroyedRegions.clear();
           }
           warningLogged = true;
         }
@@ -197,17 +209,23 @@ public class PersistentBucketRecoverer extends RecoveryRunnable implements Persi
       // Log and bail
       logger.error(e.getMessage(), e);
     } finally {
-      /*
-       * Our job is done. Stop listening to the bucket advisors.
-       */
-      removeListeners();
-      /*
-       * Make sure the recovery completion message was printed to the log.
-       */
-      for (RegionStatus region : regions) {
-        if (!region.loggedDoneMessage) {
-          region.logDoneMessage();
+      if (!regions.isEmpty()) {
+
+        /*
+         * Our job is done. Stop listening to the bucket advisors.
+         */
+        removeListeners();
+        /*
+         * Make sure the recovery completion message was printed to the log.
+         */
+
+        for (RegionStatus region : regions) {
+          if (!region.loggedDoneMessage) {
+            region.logDoneMessage();
+          }
         }
+
+        regions.clear();
       }
     }
   }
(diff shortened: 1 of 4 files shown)


geode-core/src/main/java/org/apache/geode/internal/cache/partitioned/PersistentBucketRecoverer.java
```diff
@@ -246,8 +264,12 @@ public RegionStatus(PartitionedRegion region) {
     }
 
     public void removeListeners() {
+      BucketPersistenceAdvisor tempBPA;
       for (ProxyBucketRegion proxyBucket : bucketRegions) {
-        proxyBucket.getPersistenceAdvisor().removeListener(PersistentBucketRecoverer.this);
+        tempBPA = proxyBucket.getPersistenceAdvisor();
+        if (!tempBPA.isClosed()) {
+          proxyBucket.getPersistenceAdvisor().removeListener(PersistentBucketRecoverer.this);
```
dschneider-pivotal, line 271: why do you refetch "proxyBucket.getPersistenceAdvisor()" instead of using tempBPA?
Also, would it be better to do this isClosed check in BucketPersistenceAdvisor?
Also why did you make this particular change? You have changed close to set the listeners to an empty collection. You could just have removeListener check if the set isEmpty and return. It seems like this method would not need to change at all

geode-core/src/main/java/org/apache/geode/internal/cache/persistence/PersistenceAdvisorImpl.java
```diff
@@ -1155,6 +1155,7 @@ public void close() {
     isClosed = true;
     persistentMemberManager.removeRevocationListener(profileChangeListener);
     cacheDistributionAdvisor.removeProfileChangeListener(profileChangeListener);
+    persistentStateListeners = Collections.emptySet();
```
dschneider-pivotal, line 1158: add unit test coverage for this change to this class

geode-core/src/main/java/org/apache/geode/internal/cache/partitioned/PersistentBucketRecoverer.java
```diff
@@ -182,12 +183,23 @@ public void run2() {
         if (!warningLogged) {
           sleepMillis = SLEEP_PERIOD / 2;
         }
+        if (regions.isEmpty()) {
```
dschneider-pivotal, line 186: I'm not sure about all these changes but this method needs unit test coverage for them.

dschneider-pivotal (changes requested):

geode-core/src/main/java/org/apache/geode/internal/cache/partitioned/PersistentBucketRecoverer.java
mivanac: Thanks for comments. Updated

mhansonp (changes requested):
As @dschneider-pivotal suggested, unit test changes would seem to be necessary.

geode-core/src/test/java/org/apache/geode/internal/cache/partitioned/PersistentBucketRecovererTest.java
```diff
@@ -72,4 +80,78 @@ public void latchCanBeCountedDown() {
...
+    bucs[4] = bucketRegion5;
+    when(regionAdvisor.getProxyBucketArray()).thenReturn(bucs);
+
+    BucketPersistenceAdvisor persistenceAdvisor = mock(BucketPersistenceAdvisor.class);
+    when(bucketRegion1.getPersistenceAdvisor()).thenReturn(persistenceAdvisor);
+    when(bucketRegion2.getPersistenceAdvisor()).thenReturn(persistenceAdvisor);
+    when(bucketRegion3.getPersistenceAdvisor()).thenReturn(persistenceAdvisor);
+    when(bucketRegion4.getPersistenceAdvisor()).thenReturn(persistenceAdvisor);
+    when(bucketRegion5.getPersistenceAdvisor()).thenReturn(persistenceAdvisor);
+    when(persistenceAdvisor.isClosed()).thenReturn(true);
```
mhansonp, line 144: Could you use a for loop to reduce duplication please?

mhansonp (approved):

mivanac:
Hi @dschneider-pivotal, are changes OK with you.

mivanac:
Hi @dschneider-pivotal, are changes OK with you, or there are more reasons to keep **Changes requested** on this PR.

geode-core/src/main/java/org/apache/geode/internal/cache/partitioned/PersistentBucketRecoverer.java
```diff
@@ -182,12 +186,23 @@ public void run2() {
...
+          break;
+        }
         Thread.sleep(sleepMillis);
 
         if (membershipChanged) {
           membershipChanged = false;
           for (RegionStatus region : regions) {
-            region.logWaitingForMembers();
+            try {
```
dschneider-pivotal, line 197: Instead of having a destroyedRegions collection, I think it would be better to use an Iterator on regions. Then in the catch you can just call iterator.remove.

dschneider-pivotal (changes requested):

geode-core/src/main/java/org/apache/geode/internal/cache/persistence/PersistenceAdvisorImpl.java
```diff
@@ -721,12 +721,19 @@ public void addListener(PersistentStateListener listener) {
   @Override
   public void removeListener(PersistentStateListener listener) {
     synchronized (this) {
+      if (persistentStateListeners.isEmpty()) {
```
dschneider-pivotal, line 724: even better would be:
`if (!persistentStateListeners.contains(listener)) { return; }`

geode-core/src/main/java/org/apache/geode/internal/cache/persistence/PersistenceAdvisorImpl.java
```diff
@@ -721,12 +721,19 @@ public void addListener(PersistentStateListener listener) {
...
+        return;
+      }
       Set<PersistentStateListener> tmpListeners = new HashSet<>(persistentStateListeners);
       tmpListeners.remove(listener);
       persistentStateListeners = Collections.unmodifiableSet(tmpListeners);
     }
   }
 
+  Set<PersistentStateListener> getPersistentStateListenerSet() {
```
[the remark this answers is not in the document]
dschneider-pivotal: actually the TestOnly annotation in org.jetbrains.annotations would be even better.

geode-core/src/main/java/org/apache/geode/internal/cache/partitioned/PersistentBucketRecoverer.java
mivanac: Thanks for comments. Updated

geode-core/src/main/java/org/apache/geode/internal/cache/persistence/PersistenceAdvisorImpl.java
mivanac: Thanks for comment. Updated.

geode-core/src/main/java/org/apache/geode/internal/cache/persistence/PersistenceAdvisorImpl.java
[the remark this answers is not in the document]
mivanac: Updated.

mivanac:
Hi @dschneider-pivotal, are changes OK with you.

mkevo (approved):

geode-core/src/main/java/org/apache/geode/internal/cache/partitioned/PersistentBucketRecoverer.java
```diff
@@ -197,17 +209,23 @@ public void run2() {
...
-       * Our job is done. Stop listening to the bucket advisors.
-       */
-      removeListeners();
-      /*
-       * Make sure the recovery completion message was printed to the log.
-       */
-      for (RegionStatus region : regions) {
-        if (!region.loggedDoneMessage) {
-          region.logDoneMessage();
+      if (!regions.isEmpty()) {
```
upthewaterspout, line 212: I'm not sure this if check is necessary. Everything inside here will work if regions is empty, and will have no effect if regions is not.

geode-core/src/main/java/org/apache/geode/internal/cache/partitioned/PersistentBucketRecoverer.java
```diff
@@ -197,17 +209,23 @@ public void run2() {
...
+         * Make sure the recovery completion message was printed to the log.
+         */
+
+        for (RegionStatus region : regions) {
+          if (!region.loggedDoneMessage) {
+            region.logDoneMessage();
+          }
         }
+
+        regions.clear();
```
upthewaterspout, line 228: I don't think this clear should be necessary either - once this thread exits this whole object should be garbage collected.

upthewaterspout (approved):
Approving this since in general the change looks good. I do have a few cleanup suggestions inline.

mivanac:
Hi @dschneider-pivotal, are changes OK with you.

dschneider-pivotal (approved):

PR #7114, opened 2021-11-16
GEODE-9642: skip RedundancyRecovery for colocated regions, if colocat…

mivanac:
…ion is not completed

Co-authored-by: albertogpz alberto.gomez@est.tech

In this solution, for each server, if colocation is not completed (due to any possible reason), we are skipping RedundancyRecovery. So, when last server is registering partitioned region, he will set colocation completed and notify all other servers, and he will trigger creation of buckets on all other servers.

### For all changes:

### DIFF
--- a/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/execute/PRColocationDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/execute/PRColocationDUnitTest.java
@@ -2593,6 +2593,14 @@ public class PRColocationDUnitTest extends JUnit4CacheTestCase {
 
   private static class MyResourceObserver extends ResourceObserverAdapter {
     Set<Region> recoveredRegions = new HashSet<Region>();
+    Set<Region> recoveryStartedOnRegions = new HashSet<Region>();
+
+    @Override
+    public void rebalancingOrRecoveryStarted(Region region) {
+      synchronized (this) {
+        recoveryStartedOnRegions.add(region);
+      }
+    }
 
     @Override
     public void rebalancingOrRecoveryFinished(Region region) {
@@ -2604,6 +2612,11 @@ public class PRColocationDUnitTest extends JUnit4CacheTestCase {
     public void waitForRegion(Region region, long timeout) throws InterruptedException {
       long start = System.currentTimeMillis();
       synchronized (this) {
+        long waitStart = (timeout / 10) - (System.currentTimeMillis() - start);
+        this.wait(waitStart);
+        if (!recoveryStartedOnRegions.contains(region)) {
+          return;
+        }
         while (!recoveredRegions.contains(region)) {
           long remaining = timeout - (System.currentTimeMillis() - start);
           assertTrue("Timeout waiting for region recovery", remaining > 0);
--- a/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/execute/PRColocationDUnitTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/execute/PRColocationDUnitTest.java
@@ -2612,10 +2612,12 @@ public class PRColocationDUnitTest extends JUnit4CacheTestCase {
     public void waitForRegion(Region region, long timeout) throws InterruptedException {
       long start = System.currentTimeMillis();
       synchronized (this) {
-        long waitStart = (timeout / 10) - (System.currentTimeMillis() - start);
-        this.wait(waitStart);
-        if (!recoveryStartedOnRegions.contains(region)) {
-          return;
+        while (!recoveryStartedOnRegions.contains(region)) {
+          long waitStart = (timeout / 10) - (System.currentTimeMillis() - start);
+          if (waitStart <= 0) {
+            return;
+          }
+          this.wait(waitStart);
         }
         while (!recoveredRegions.contains(region)) {
           long remaining = timeout - (System.currentTimeMillis() - start);
--- a/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/partitioned/PersistentColocatedPartitionedRegionDistributedTest.java
+++ b/geode-core/src/distributedTest/java/org/apache/geode/internal/cache/partitioned/PersistentColocatedPartitionedRegionDistributedTest.java
@@ -59,6 +59,7 @@ import java.util.function.Function;
 
 import junitparams.Parameters;
 import junitparams.naming.TestCaseName;
+import org.apache.commons.lang3.tuple.Pair;
 import org.apache.logging.log4j.Logger;
 import org.junit.After;
 import org.junit.Before;
@@ -122,6 +123,7 @@ public class PersistentColocatedPartitionedRegionDistributedTest implements Seri
 
   private static volatile InternalCache cache;
   private static volatile CountDownLatch latch;
+  private static volatile boolean recoveryExecuted;
 
   private final transient List<AsyncInvocation<Void>> asyncInvocations = new ArrayList<>();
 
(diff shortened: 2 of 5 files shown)


geode-wan/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/AlterRegionCommandWithRemoteLocator.java
```diff
@@ -207,8 +207,34 @@ public void whenAlteringMultipleRegionWithAlterCommandToAddGatewaySendersThenItS
...
+
+    gfsh.executeAndAssertThat(
+        "query --query=\"select key,value from " + SEPARATOR + "Positions.entries\"")
+        .statusIsSuccess();
+
+    gfsh.executeAndAssertThat(
+        "create gateway-sender --id=parallelPositions --remote-distributed-system-id=1 --enable-persistence=true --disk-store-name=data --parallel=true")
+        .statusIsSuccess();
 
+    GeodeAwaitility.await().until(() -> {
```
albertogpz, line 235: Shouldn't you add an atMost() condition here?

geode-wan/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/AlterRegionCommandWithRemoteLocator.java
```diff
@@ -207,8 +207,34 @@ public void whenAlteringMultipleRegionWithAlterCommandToAddGatewaySendersThenItS
...
+  }
 
+  /**
+   * Test execution of alter region command, when adding gw sender to already initialized region is
+   * not taking too much time.
+   * Verify that execution of command will not be halted, due to PRShadowRegion not available
+   * immediately in all servers.
+   */
+  @Test
+  public void alterInitializedRegionWithGwSenderOnManyServersDoesNotTakeTooLong() {
```
albertogpz, line 219: This test case seems to pass in my laptop without the fix. Should more servers be added?

geode-wan/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/AlterRegionCommandWithRemoteLocator.java
mivanac: In previos PR this was commented, and updated accordingly.

geode-wan/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/AlterRegionCommandWithRemoteLocator.java
mivanac: Also, this is test that reproduce fault in old PR. You can repeat test case several times, it will fail.

geode-wan/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/AlterRegionCommandWithRemoteLocator.java
mivanac: I agree with you.

geode-wan/src/distributedTest/java/org/apache/geode/management/internal/cli/commands/AlterRegionCommandWithRemoteLocator.java
mivanac: For sure, this would fail with changed timeout, but that cannot be changed. For details you can see comments in previous PR.

albertogpz (changes requested):
I have a general comment about how the tests have been modified.

I think that the problem with the modifications of the test cases is that it is not checked anymore if recovery is started at least in one server. It is only checked that if it is started, it is also finished, but it may happen (because it is not checked) that recovery is not started in any server.

I suggest an alternative change in which it is checked that recovery is finished in the last server started.

mivanac:
But recovery could be started, on more servers then just lust. In real situations when region is created, recovery will be started almost on all servers. So I am not sure that checking recovery only on one server is valid test.

albertogpz:
> It only depends if collocation is completed.

If servers are started sequentially, colocation will not be completed until the last one is started and that's why recovery will only be run in it.
If servers are started in parallel, it must be checked that recovery is executed at least in one of them.

As almost all test cases start the servers sequentially, I suggested to check recovery on the last one.
For test cases starting the servers in parallel, it must be checked that recovery is started in at least one of them or alternatively, the start could be changed to start all the servers in parallel except for one and then check that recovery was started on the last one.

mivanac:
We are talking about creation of child region on servers, not starting servers. Normal behavior, when you create region, is that it is created on all servers in parallel.

albertogpz:
> We are talking about creation of child region on servers, not starting servers. Normal behavior, when you create region, is that it is created on all servers in parallel.

You are right. I meant region creation on each server and not server start.

mivanac:
It would be good if any code owner could comment on this issue, and give his opinion on new solution, and what is expected from test.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/execute/PRColocationDUnitTest.java
```diff
@@ -2604,6 +2612,11 @@ public void rebalancingOrRecoveryFinished(Region region) {
     public void waitForRegion(Region region, long timeout) throws InterruptedException {
       long start = System.currentTimeMillis();
       synchronized (this) {
+        long waitStart = (timeout / 10) - (System.currentTimeMillis() - start);
+        this.wait(waitStart);
+        if (!recoveryStartedOnRegions.contains(region)) {
```
mhansonp, line 2617: I don't know the answer to this, but is there something we could wait for rather than just a timeout?

mhansonp (approved):

mhansonp (commented):
There  seems to be some debate here about the behavior changes. Waiting to see the result

mhansonp:
@albertogpz Did your concerns get addressed?

albertogpz:
> @albertogpz Did your concerns get addressed?

I think the modified tests still need to check that recovery was started at least in one of the servers.

DonalEvans:
> > @albertogpz Did your concerns get addressed?
> 
> I think the modified tests still need to check that recovery was started at least in one of the servers.

As far as I understand the tests, any time `checkIfRecoveryExecuted()` is called, it's implicitly checking that recovery was started, because if recovery is never started on any servers, then the calls to `recoveryStarted.await()` will time out and `recoveryExecuted` will not be set to true.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/partitioned/PersistentColocatedPartitionedRegionDistributedTest.java
```diff
@@ -2117,6 +2175,10 @@ private void tearDownPartitionedRegionObserver() {
     }
   }
 
+  private boolean getRecoveryStatus() {
+    return recoveryExecuted;
+  }
```
DonalEvans, line 2180: This can probably be inlined.

DonalEvans (approved):

albertogpz:
> > > @albertogpz Did your concerns get addressed?
> > 
> > 
> > I think the modified tests still need to check that recovery was started at least in one of the servers.
> 
> As far as I understand the tests, any time `checkIfRecoveryExecuted()` is called, it's implicitly checking that recovery was started, because if recovery is never started on any servers, then the calls to `recoveryStarted.await()` will time out and `recoveryExecuted` will not be set to true.

The problem is that the tests are not checking that recovery is started at least in one server. They are only checking that if recovery is started, then it will be done in a reasonable time.
But it could be that recovery is not started in any server and the tests will pass.

albertogpz (approved):

albertogpz:
> > @albertogpz Did your concerns get addressed?
> 
> I think the modified tests still need to check that recovery was started at least in one of the servers.

My concerns have been addressed. Now, recovery started in at least one server is addressed.

geode-core/src/distributedTest/java/org/apache/geode/internal/cache/execute/PRColocationDUnitTest.java
```diff
@@ -2604,6 +2612,13 @@ public void rebalancingOrRecoveryFinished(Region region) {
     public void waitForRegion(Region region, long timeout) throws InterruptedException {
       long start = System.currentTimeMillis();
       synchronized (this) {
+        while (!recoveryStartedOnRegions.contains(region)) {
```
mhansonp, line 2615: This should be simplified to an await.until...

mhansonp (changes requested):
See comment. Looks like you should use an await in one case..

PR #7115, opened 2021-11-16
GEODE-9812: Add Radish tests that kills multiple servers

jdeppe-pivotal:
- Despite data loss, the test ensures that operations continue without
  errors related to MOVED responses or missing slots.

### For all changes:

### DIFF
--- a/geode-for-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/string/StringsKillMultipleServersDUnitTest.java
+++ b/geode-for-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/string/StringsKillMultipleServersDUnitTest.java
@@ -15,7 +15,6 @@
 
 package org.apache.geode.redis.internal.executor.string;
 
-import static org.apache.geode.distributed.ConfigurationProperties.REDUNDANCY_ZONE;
 import static org.apache.geode.test.awaitility.GeodeAwaitility.await;
 import static org.apache.geode.test.dunit.rules.RedisClusterStartupRule.BIND_ADDRESS;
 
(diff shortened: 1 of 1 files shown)


dschneider-pivotal (approved):

geode-for-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/string/StringsKillMultipleServersDUnitTest.java
```diff
@@ -0,0 +1,111 @@
...
+  public void operationsShouldContinueAfterMultipleServerFailures() throws Exception {
+    AtomicBoolean running = new AtomicBoolean(true);
+    AtomicInteger counter = new AtomicInteger(0);
+
+    Future<Void> future1 = executor.submit(() -> doSetOps(running, counter));
+    Future<Void> future2 = executor.submit(() -> doGetOps(running, counter));
+    await().until(() -> counter.get() > 100);
+
+    cluster.crashVM(3);
+    cluster.crashVM(4);
```
DonalEvans, line 82: There is a possibility that buckets are distributed in such a way that server1 and server2 have all the secondary copies for the primary buckets hosted on server3 and server4, in which case, we wouldn't actually end up in a situation where we lose all copies of a bucket. Would it be possible to add in a check to ensure that we are definitely losing both primary and secondary buckets, which is what this test is ultimately trying to test?

jdeppe-pivotal: I've upped the min number of keys to 1000 which ensures that all buckets will contain at least 1 key. Also adding redundancy zones to further restrict where buckets can be place. Now this boils down to the likelihood of 64 buckets on server 3 having 0 secondaries on server 4 - I think that's around 1 / 2^64 :)

geode-for-redis/src/distributedTest/java/org/apache/geode/redis/internal/executor/string/StringsKillMultipleServersDUnitTest.java
```diff
@@ -49,10 +50,19 @@
...
+        .withProperty(REDUNDANCY_ZONE, "A"));
+    cluster.startRedisVM(2, x -> x
+        .withConnectionToLocator(locatorPort)
+        .withProperty(REDUNDANCY_ZONE, "B"));
+    cluster.startRedisVM(3, x -> x
+        .withConnectionToLocator(locatorPort)
+        .withProperty(REDUNDANCY_ZONE, "A"));
+    cluster.startRedisVM(4, x -> x
+        .withConnectionToLocator(locatorPort)
+        .withProperty(REDUNDANCY_ZONE, "B"));
```
DonalEvans, line 65: My understanding of redundancy zones may be wrong, but I think this change doesn't necessarily guarantee that we lose all copies of a bucket when crashing server3 and server4. If every secondary copy for server3 is on server2 and every secondary copy for server4 is on server1 (which is what will happen if we define these redundancy zones) then we can crash server3 and server4 but still have at least one copy of every bucket survive. I don't know how easy it would be to do, but we could be sure of crashing the "right" servers by finding which servers host the primary and the secondary for an arbitrary bucket and crashing both.

jdeppe-pivotal: So in that scenario the test would still pass but wouldn't be testing the right thing. However given that the likelihood of that happening is so incredibly small (pretty sure its around 1:2^64) I don't think it's worth over-complicating the test.

DonalEvans (approved):

PR #7116, opened 2021-11-16
GEODE-9811: UnavailableSecurityManagerException not handled properly

jmelchio:
- UnavailableSecurityManagerException is not handled on the server and
  will therefore stop client from failing over to alternate member for
  operations
- Fix adds a method to recreate the problem in DUnit testing as it is
  hard to recreate the problem under normal circumstances

### For all changes:


geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/command/Put70.java
```diff
@@ -411,6 +413,9 @@ public void cmdExecute(final Message clientMessage, final ServerConnection serve
       serverConnection.setAsTrue(RESPONDED);
       region.getCachePerfStats().incDeltaFullValuesRequested();
       return;
+    } catch (UnavailableSecurityManagerException e) {
```
jinmeiliao, line 416: So theoretically every command needs to do this?

geode-core/src/distributedTest/java/org/apache/geode/security/TestIntegratedSecurityService.java
```diff
@@ -0,0 +1,101 @@
...
+import org.apache.geode.internal.classloader.ClassPathLoader;
+import org.apache.geode.internal.security.IntegratedSecurityService;
+import org.apache.geode.internal.security.shiro.SecurityManagerProvider;
+import org.apache.geode.logging.internal.log4j.api.LogService;
+
+/**
+ * IntegratedSecurityService extension that allows for manipulation of behavior for testing
+ * handling of rare occurrences of responses of the security service in a more predictable way.
+ */
+public class TestIntegratedSecurityService extends IntegratedSecurityService {
```
jinmeiliao, line 38: I wonder if we need to do all these changes just for testing. Can we just use mocks and make sure the correct exceptions are thrown?

jinmeiliao (changes requested):
This seems to have only fixed the put command, theoretically any operations can encounter this. We need a fix that would work for all commands.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/command/Put70.java
jmelchio: @jinmeiliao yes, but the issue has only occurred on put70 command so far. 
There are (if I'm not mistaken) 17 other commands that can hit the same issue so I can make the change there too.

geode-core/src/distributedTest/java/org/apache/geode/security/TestIntegratedSecurityService.java
jmelchio: I'm not sure a mock can do all that this implementation does. It's used to collect test data which is used for verification.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/command/Put70.java
jinmeiliao: Is there a generic way to fix this without having to touch all the command files?

geode-core/src/distributedTest/java/org/apache/geode/security/TestIntegratedSecurityService.java
```diff
@@ -0,0 +1,101 @@
...
+    Subject subject = super.getSubject();
+    if (!getSubjectFailThrowable.get().isEmpty()
+        && getSubjectCallCount.get() > getSubjectFailInterval.get()) {
+      getSubjectTimesFailed.incrementAndGet();
+      logger.info("about to throw induced exception on call count: {}", getSubjectCallCount.get());
+      getSubjectCallCount.set(0);
+      try {
+        Class throwableClass = ClassPathLoader.getLatest().forName(getSubjectFailThrowable.get());
+        Constructor constructor = throwableClass.getConstructor(String.class);
+        throw (RuntimeException) constructor.newInstance("forced exception for test");
```
DonalEvans, line 66: Warnings here can be fixed by using:
```
        Class<? extends RuntimeException> throwableClass =
            uncheckedCast(ClassPathLoader.getLatest().forName(getSubjectFailThrowable.get()));
        Constructor<? extends RuntimeException> constructor =
            throwableClass.getConstructor(String.class);
        throw constructor.newInstance("forced exception for test");
```

geode-core/src/main/java/org/apache/geode/internal/cache/GemFireCacheImpl.java
```diff
@@ -1099,6 +1099,30 @@ public void lockDiskStore(String diskStoreName) {
     doLockDiskStore(diskStoreName);
   }
 
+  /**
+   * Having this method to return a SecurityServiceFactory allows users to replace the factory
+   * for testing purposes.
+   *
+   * In a test set the system property "org.apache.geode.internal.security.SecurityServiceFactory"
+   * to the class of choice ie. [Classname].class.getName() and in this class either implement
+   * SecurityServiceFactory or override the default implementation to create the desired behavior.
```
DonalEvans, line 1108: I'm really not a fan of the idea of introducing a system property solely to support testing. Is there any other approach that could be used here?

geode-core/src/distributedTest/java/org/apache/geode/security/SecurityManagerAvailabilityDUnitTest.java
```diff
@@ -0,0 +1,164 @@
...
+
+    assertThat((Integer) resultMap0.get(FAIL_TIMES)).isGreaterThan(0);
+    assertThat((String) resultMap0.get(FAIL_THROWABLE)).isNotEmpty();
+    assertThat((Integer) resultMap1.get(FAIL_TIMES)).isGreaterThan(0);
+    assertThat((String) resultMap1.get(FAIL_THROWABLE)).isNotEmpty();
+
+    Throwable error = (Throwable) clientResult.get(0);
+
+    assertThat(error instanceof Exception).isTrue();
+    assertThat(error.getMessage()).isEqualTo("no exception");
```
DonalEvans, line 161: These might be better as:
```
    assertThat(error).isInstanceOf(Exception.class);
    assertThat(error).hasMessage("no exception");
```
Also, I think that the first assertion is always true, since the exception can either be the `Exception` set on line 119 or some other exception with `UnavailableSecurityManagerException` as a cause, set on line 130, so this assertion isn't really telling us much. When I reverted the fix and reran the test, this assertion did not fail (although the one on the message contents did).

DonalEvans (changes requested):

geode-core/src/main/java/org/apache/geode/internal/cache/GemFireCacheImpl.java
jmelchio: @DonalEvans I agree with that but unfortunately the code is such that I have not been able to come up with an alternative approach.

geode-core/src/main/java/org/apache/geode/internal/security/IntegratedSecurityService.java
```diff
@@ -119,8 +121,12 @@ public Subject getSubject() {
       }
     }
 
-    // in other cases like rest call, client operations, we get it from the current thread
-    currentUser = SecurityUtils.getSubject();
+    try {
+      // in other cases like rest call, client operations, we get it from the current thread
+      currentUser = SecurityUtils.getSubject();
+    } catch (UnavailableSecurityManagerException e) {
+      throw new CacheClosedException("likely cache shutdown in progress", e);
```
DonalEvans, line 128: Is it possible to check the value of `GemfireCahceImpl.isClosing` here to confirm if we are actually closing the cache, rather than just assuming that we are?

DonalEvans (approved):

jinmeiliao (approved):

kamilla1201 (approved):
The files I'm code owner for look good

DonalEvans (approved):

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/CacheClientProxy.java
```diff
@@ -980,15 +979,6 @@ private void closeOtherTransientFields() {
     // Commented to fix bug 40259
     // this.clientVersion = null;
     closeNonDurableCqs();
-
```
jinmeiliao, line 983: These changes are reverting Barry's fix to log out the subject here, we are logging out the subject in the `close` method later on. So these changes are not needed here anymore.

geode-core/src/main/java/org/apache/geode/internal/security/SecurityServiceFactory.java
```diff
@@ -28,6 +28,8 @@
 import org.apache.geode.security.SecurityManager;
 
 public class SecurityServiceFactory {
+  public static final String SECURITY_SERVICE_SYSTEM_PROPERTY =
+      "org.apache.geode.internal.security.SecurityService";
```
agingade, line 32: Is this new property? Does it impact current behavior?
When we introduce system property we use SystemPropertyHelper class to define the system property...Is this not a requirement here...

jinmeiliao: This is only for testing

geode-core/src/distributedTest/java/org/apache/geode/security/SecurityManagerAvailabilityDUnitTest.java
```diff
@@ -0,0 +1,132 @@
...
+
+    server1 = clusterStartupRule.startServerVM(1,
+        s -> s.withSecurityManager(ExpirableSecurityManager.class)
+            .withProperty(SECURITY_LOG_LEVEL, "debug")
+            .withCredential("test", "test")
+            .withConnectionToLocator(locatorPort)
+            .withSystemProperty(SECURITY_SERVICE_SYSTEM_PROPERTY,
+                TestIntegratedSecurityService.class.getName()));
+    server2 = clusterStartupRule.startServerVM(2,
+        s -> s.withSecurityManager(ExpirableSecurityManager.class)
```
pivotal-jbarrett, line 72: On the surface this looks like duplicate code from lines 65-70. Perhaps assign the lambda to a local variable and use in both places rather than duplicate.

geode-core/src/main/java/org/apache/geode/internal/security/SecurityServiceFactory.java
```diff
@@ -51,6 +53,11 @@ public static SecurityService create(Properties securityProps, CacheConfig cache
 
   public static SecurityService create(Properties securityProps,
       SecurityManager preferredSecurityManager, PostProcessor preferredPostProcessor) {
+    String securityServiceClassName = System.getProperty(SECURITY_SERVICE_SYSTEM_PROPERTY);
```
pivotal-jbarrett, line 56: Is this system property get in a hot path? System property access is synchronized. This should probably be part of the static initializer for this class rather than on a per-call basis. System properties should not be mutated at runtime.

geode-core/src/main/java/org/apache/geode/internal/security/SecurityServiceFactory.java
```diff
@@ -51,6 +53,11 @@ public static SecurityService create(Properties securityProps, CacheConfig cache
...
+    if (securityServiceClassName != null) {
```
pivotal-jbarrett, line 57: If this is truly for testing purposes only please isolate this code in its own method with some indicator in the name that it is an intended testing only smell.
```java
final SecurityManager testOnlySecurityManager = whateverThisMethodIsDoingForTesting();
if (null != testOnlySecurityManager) {
  return testOnlySecurityManager;
}
```
Though I would seriously question why we need something like this in production code? Perhaps there is a different way to write the tests that doesn't require special test only code in production. Could a custom security manager delegate be used that delegates to this testing only instance when necessary?

pivotal-jbarrett (changes requested):

geode-core/src/main/java/org/apache/geode/internal/security/IntegratedSecurityService.java
jmelchio: At this point a `cache` object is not in scope and given that the issue we're addressing here is related to race conditions with cache requests on the one hand, and cache being either gone or in the process of being shut down on the other, I don't know if it's worth adding the functionality of bringing it in scope and checking its state.

geode-core/src/main/java/org/apache/geode/internal/security/SecurityServiceFactory.java
jinmeiliao: It's not a hot path, security service is only created once when cache is created.

geode-core/src/main/java/org/apache/geode/internal/security/SecurityServiceFactory.java
jinmeiliao: We will need to simulate a cache closed scenario while calling the "authorized" method, this can't be done in the `SecurityManager`. The only way to reproduce this in a predicable way is to plug in a test security service object. I have tried testing it the other way, none of them can guarantee the failure to happen.

I will put the logic in a method that clearly indicate this is for testing.

geode-core/src/main/java/org/apache/geode/internal/security/SecurityServiceFactory.java
```diff
@@ -49,8 +53,21 @@ public static SecurityService create(Properties securityProps, CacheConfig cache
...
+      return null;
+    }
+    return CallbackInstantiator.getObjectOfType(securityServiceClassName, SecurityService.class);
+  }
+
   public static SecurityService create(Properties securityProps,
       SecurityManager preferredSecurityManager, PostProcessor preferredPostProcessor) {
+    SecurityService testSecurityService = getTestSecurityService();
+    if (testSecurityService != null) {
+      return testSecurityService;
```
pivotal-jbarrett, line 68: If the purpose for this is to just create a specific `SecurityManager` for testing why isn't the normal method, using security properties, an acceptable way to configure a test `SecurityManager`? As written, this test only method completely bypasses all the other security initialization below resulting in a modification of behavior from the normal production behavior.

pivotal-jbarrett (changes requested):

geode-core/src/main/java/org/apache/geode/internal/security/SecurityServiceFactory.java
jinmeiliao: it's create a `SecurityManager`, it's to create a `SecurityService`, the test securityService is a child of the usual security service, only to override one method to throw an exception at intervals.

jinmeiliao: The alternative is that now that we have confirmed that by turning the `UnavailableSecurityManagerException` in the security service into `CacheClosedException` will cause the client to retry other servers, we can get rid of the dunit test and simply use a junit test to make sure the conversion of the exceptions.

jinmeiliao: Probably this other PR is more to the point: https://github.com/apache/geode/pull/7148

geode-core/src/distributedTest/java/org/apache/geode/security/SecurityManagerAvailabilityDUnitTest.java
```diff
@@ -0,0 +1,117 @@
...
+      UpdatableUserAuthInitialize.setUser("data1");
+      Region<Object, Object> region =
+          clientCache.createClientRegionFactory(ClientRegionShortcut.PROXY).create("region");
+
+      // even though we throw exceptions in the security service, the client will retry another
+      // server, so client wouldn't see any exceptions
+      IntStream.range(0, 1000).forEach(i -> region.put(i, "value" + i));
+    });
+
+    Integer timeFailedOnServer1 = server1.invoke(
```
kirklund, line 96: You should just use primitives (such as `int`) everywhere by default instead of wrapper classes like `Integer`.

geode-core/src/main/java/org/apache/geode/internal/cache/tier/sockets/MessageDispatcher.java
```diff
@@ -113,6 +113,8 @@
    */
   private volatile boolean _isStopped = true;
 
+  private volatile long wait_for_re_auth_start_time = -1;
```
kirklund, line 116: Let's do two things here. One `re` is too abbreviated. Spell it out. Also please switch to camel case even if some of the surrounding code uses the wrong case. (Camel case just means `waitForSomethingAuthStartTime`)

jinmeiliao: Thanks! This PR is gonna be split into two PRs. I will incorporate your comment in the coming PR.

PR #7119, opened 2021-11-17
Geode 9396

kamilla1201:
### For all changes:

### DIFF
--- a/geode-assembly/src/upgradeTest/java/org/apache/geode/management/DeploymentManagementUpgradeTest.java
+++ b/geode-assembly/src/upgradeTest/java/org/apache/geode/management/DeploymentManagementUpgradeTest.java
@@ -89,13 +89,13 @@ public class DeploymentManagementUpgradeTest {
     int locatorPort = ports[1];
     int jmxPort = ports[2];
     GfshExecution execute =
-        GfshScript.of(startLocatorCommand("test", locatorPort, jmxPort, httpPort, 0))
+        GfshScript.of(startLocatorCommand("test", "localhost", locatorPort, jmxPort, httpPort, 0))
             .and("deploy --jar=" + clusterJar.getAbsolutePath())
             .and("shutdown --include-locators")
             .execute(oldGfsh);
 
     // use the latest gfsh to start the locator in the same working dir
-    GfshScript.of(startLocatorCommand("test", locatorPort, jmxPort, httpPort, 0))
+    GfshScript.of(startLocatorCommand("test", "localhost", locatorPort, jmxPort, httpPort, 0))
         .execute(gfsh, execute.getWorkingDir());
 
     ClusterManagementService cms = new ClusterManagementServiceBuilder()
--- a/geode-assembly/src/upgradeTest/java/org/apache/geode/management/RollingUpgradeWithSslDUnitTest.java
+++ b/geode-assembly/src/upgradeTest/java/org/apache/geode/management/RollingUpgradeWithSslDUnitTest.java
@@ -1,0 +1,276 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.apache.geode.management;
+
+import static org.apache.geode.distributed.ConfigurationProperties.BIND_ADDRESS;
+import static org.apache.geode.distributed.ConfigurationProperties.SSL_ENABLED_COMPONENTS;
+import static org.apache.geode.distributed.ConfigurationProperties.SSL_ENDPOINT_IDENTIFICATION_ENABLED;
+import static org.apache.geode.distributed.ConfigurationProperties.SSL_KEYSTORE;
+import static org.apache.geode.distributed.ConfigurationProperties.SSL_KEYSTORE_PASSWORD;
+import static org.apache.geode.distributed.ConfigurationProperties.SSL_KEYSTORE_TYPE;
+import static org.apache.geode.distributed.ConfigurationProperties.SSL_REQUIRE_AUTHENTICATION;
+import static org.apache.geode.distributed.ConfigurationProperties.SSL_TRUSTSTORE;
+import static org.apache.geode.distributed.ConfigurationProperties.SSL_TRUSTSTORE_PASSWORD;
+import static org.apache.geode.distributed.ConfigurationProperties.SSL_TRUSTSTORE_TYPE;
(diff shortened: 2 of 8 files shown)


jinmeiliao (approved):

dschneider-pivotal (approved):

geode-core/src/main/java/org/apache/geode/internal/tcp/Connection.java
```diff
@@ -192,7 +196,7 @@
    */
   private final String conduitIdStr;
 
-  private InternalDistributedMember remoteAddr;
+  private InternalDistributedMember remoteMember;
```
Bill, line 199: Big improvement in naming here!

geode-core/src/main/java/org/apache/geode/internal/net/SocketCreator.java
```diff
@@ -373,14 +373,12 @@ void configureSSLEngine(SSLEngine engine, String hostName, int port, boolean cli
...
-    if (!clientSocket) {
-      engine.setNeedClientAuth(sslConfig.isRequireAuth());
-    }
-
     if (clientSocket) {
       if (checkAndEnableHostnameValidation(parameters)) {
         updateEngineWithParameters = true;
       }
+    } else {
+      engine.setNeedClientAuth(sslConfig.isRequireAuth());
```
Bill, line 381: This refactor improves readability.

geode-core/src/main/java/org/apache/geode/internal/tcp/Connection.java
```diff
@@ -1798,6 +1794,46 @@ private void createIoFilter(SocketChannel channel, boolean clientSocket) throws
...
+        && InetAddressValidator.getInstance().isValid(hostNameOrIP)) {
+      try {
+        return InetAddress.getByName(hostNameOrIP).getCanonicalHostName();
+      } catch (UnknownHostException e) {
+        // ignore
+      }
+    }
+    return hostNameOrIP;
+  }
+
```
Bill, line 1836: The crux of this PR. This looks good to me. Thanks for the clear comment. I feel like whoever (eventually) deprecates the 1.15 line will notice this when they grep for "1.15". Not quite as elegant as a version-specific deserialization method on MemberIdentifier, but this approach avoids problems with our client code (using MemberIdentifiers in region names).

Bill (approved):
Approved!

geode-assembly/src/upgradeTest/java/org/apache/geode/management/RollingUpgradeWithSslDUnitTest.java
```diff
@@ -0,0 +1,276 @@
...
+    oldGfsh.stopLocator(startupExecution, name);
+    GfshScript
+        .of(startLocatorCommandWithConfig(name, locatorPort, locatorJmxPort, connectedLocatorPort))
+        .execute(currentGfsh, tempFolder.getRoot());
+  }
+
+  private void upgradeServer(String name, int serverPort, int locatorPort,
+      GfshExecution startupExecution) {
+    oldGfsh.stopServer(startupExecution, name);
+    GfshScript.of(startServerCommandWithConfig(name, serverPort, locatorPort))
```
agingade, line 166: It will be nice to add verification to see old locator, server was successfully stopped and new locator and server started. I am thinking of case where the old locator and servers were not stopped. And in the verifyMembers we are only looking at any servers or locators (not version specific).

geode-assembly/src/upgradeTest/java/org/apache/geode/management/RollingUpgradeWithSslDUnitTest.java
```diff
@@ -0,0 +1,276 @@
...
+    verifyListMembers(locatorPort);
+    causeP2PTraffic(locatorPort);
+
+    upgradeLocator("locator2", locator2Port, locator2JmxPort, locatorPort, startupExecution);
+    verifyListMembers(locatorPort);
+    causeP2PTraffic(locatorPort);
+
+    // make sure servers can do rolling upgrade too
+    upgradeServer("server1", server1Port, locatorPort, startupExecution);
+    causeP2PTraffic(locatorPort);
```
agingade, line 148: This is not checking if the entry was made into server1 and read from server2; i am thinking thats the expectation here...Please correct me if I am missing anything.

geode-core/src/main/java/org/apache/geode/internal/tcp/Connection.java
```diff
@@ -1798,6 +1794,46 @@ private void createIoFilter(SocketChannel channel, boolean clientSocket) throws
     }
   }
 
+  private SSLEngine createSslEngine(final InetSocketAddress remoteAddress) {
+    final String hostName;
+    final boolean isSender = remoteMember != null;
```
agingade, line 1799: I am not expert in this area of the code...Earlier the caller used to pass true or pass based on sender or receiver connection; is the assumption is right; the earlier argument "clientSocket" passed reflects (or based on) remoteMember setting?

upthewaterspout (approved):

geode-assembly/src/upgradeTest/java/org/apache/geode/management/DeploymentManagementUpgradeTest.java
```diff
@@ -89,13 +89,13 @@ public void newLocatorCanReadOldConfigurationData() throws IOException {
     int locatorPort = ports[1];
     int jmxPort = ports[2];
     GfshExecution execute =
-        GfshScript.of(startLocatorCommand("test", locatorPort, jmxPort, httpPort, 0))
+        GfshScript.of(startLocatorCommand("test", "localhost", locatorPort, jmxPort, httpPort, 0))
```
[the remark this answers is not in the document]
kamilla1201: I needed to use a different value for RollingUpgradeWithSslDUnitTest.

geode-core/src/main/java/org/apache/geode/internal/tcp/Connection.java
kamilla1201: `createIoFilter` is used by both sender and receiver and that creates some confusion: it isn’t obvious what `clientSocket` really means, who and why sets it to true, and in what cases `remoteMember` can be null.
Yes, before this change, the sender set `clientSocket` to true and the receiver set it to false. The sender also sets the `remoteMember` while the receiver doesn’t (it only knows the remote address from the socket channel). So there is a correlation between `remoteMember` and `clientSocket`, and we wanted to “document” it in the code.

kirklund (commented):
Is `geode-assembly/src/upgradeTest/resources/org/apache/geode/management/gemfire.properties` used by all upgrade tests in geode-assembly?

I recommend renaming it to match the test such as `MyTest_gemfire.properties` and then copy it from resources to a TemporaryFolder for the test to use. The class `ResourceUtils` covers this use-case. Look for uses of `ResourceUtils` to see various examples.

kirklund (approved):

mhansonp (approved):

geode-core/src/main/java/org/apache/geode/internal/tcp/Connection.java
agingade: Can this be a possibility: by the time createSslEngine is called the remote member has left the cluster (or forced-disconnected)? or SocketChannel is closed...
Is it a good idea to have a test, where the remote member is made to exit; right before createSslEngine is called.
Sorry if I am suggesting cases thats not practical...Please ignore if thats the case...

geode-assembly/src/upgradeTest/java/org/apache/geode/management/RollingUpgradeWithSslDUnitTest.java
Bill: We'll consider these improvement suggestions in a future PR.

PR #7122, opened 2021-11-17
GEODE-9820: stopCQ should handle general exception same way as ExecuteCQ61

jinmeiliao:
Problem:
StopCQ with expired credentials does not trigger re-authentication.

Cause:
the server side doesn't send the AuthenticationExpiredException to the client with the `MessageType.EXCEPTION`, rather it only sends down only the error message with `MessageType.CQ_EXCEPTION_TYPE`, that's now how it's generally handled.

### DIFF
--- a/geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationBackwardCompatibleDUnitTest.java
+++ b/geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationBackwardCompatibleDUnitTest.java
@@ -45,6 +45,9 @@ import org.apache.geode.cache.client.ClientCache;
 import org.apache.geode.cache.client.ClientRegionFactory;
 import org.apache.geode.cache.client.ClientRegionShortcut;
 import org.apache.geode.cache.client.ServerOperationException;
+import org.apache.geode.cache.query.CqAttributesFactory;
+import org.apache.geode.cache.query.CqQuery;
+import org.apache.geode.cache.query.QueryService;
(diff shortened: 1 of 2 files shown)


agingade (approved):

geode-core/src/upgradeTest/java/org/apache/geode/security/AuthExpirationBackwardCompatibleDUnitTest.java
```diff
@@ -387,6 +390,72 @@ public void cqOlderClientWithClientInteractionWillDeliverEventEventually() throw
...
+    Map<String, List<String>> unAuthorizedOps = getSecurityManager().getUnAuthorizedOps();
+    Map<String, List<String>> authorizedOps = getSecurityManager().getAuthorizedOps();
+    assertThat(unAuthorizedOps.keySet()).containsExactly("user1");
+    assertThat(unAuthorizedOps.get("user1")).containsExactly("DATA:READ:region");
+    assertThat(authorizedOps.keySet()).containsExactly("user1", "user2");
+    assertThat(authorizedOps.get("user2")).containsExactly("DATA:READ:region");
+  }
+
+  @Test
+  public void stopCQ() throws Exception {
```
DonalEvans, line 427: This test name could be more descriptive. Something like "stopCQTriggersClientReAuthentication" maybe?

DonalEvans (approved):

mkevo (approved):

jdeppe-pivotal (approved):

PR #7126, opened 2021-11-18
GEODE-9822: Quorum Calculation Requires Majority

Bill:
[GEODE-9822](https://issues.apache.org/jira/browse/GEODE-9822)

The membership quorum calculation doesn't require a majority. As a result, a two-locator cluster will have a split-brain during a network partition. This PR fixes the quorum calculation to ensure that a majority is required on the "winning" side.


Bill:
the most recent acceptance test failure is GEODE-9877

davebarnes97 (approved):
Doc change LGTM

dschneider-pivotal (approved):

kamilla1201 (approved):
Very important fix! This problem exists not only in the cluster of two equal-weight members but in any cluster with an even number of equal-weight members. Or more generally in a cluster with even weight.

Bill:
In fact it's even slightly worse than that, @kamilla1201. Any set of members that can be partitioned (into more than one set), where any two-or-more, each still have 49% or more of the total weight, will result in a split-brain. How's that for a sentence?

PR #7127, opened 2021-11-18
GEODE-9669: Improve reporting of Radish mem_fragmentation_ratio

jdeppe-pivotal:
- For this implementation, this value is the ratio of memory available
  to the partitioned region (Java heap) vs. the memory used by data in
  the region.

### For all changes:

### DIFF
--- a/geode-for-redis/src/integrationTest/java/org/apache/geode/redis/internal/commands/executor/string/AbstractAppendMemoryIntegrationTest.java
+++ b/geode-for-redis/src/integrationTest/java/org/apache/geode/redis/internal/commands/executor/string/AbstractAppendMemoryIntegrationTest.java
@@ -49,6 +49,7 @@ public abstract class AbstractAppendMemoryIntegrationTest implements RedisIntegr
     int listSize = 100_000;
     String key = "key";
 
+    System.gc();
     Long startingMemValue = getUsedMemory(jedis);
 
     jedis.set(key, "initial");
--- a/geode-for-redis/src/integrationTest/java/org/apache/geode/redis/internal/commands/executor/string/AbstractAppendMemoryIntegrationTest.java
+++ b/geode-for-redis/src/integrationTest/java/org/apache/geode/redis/internal/commands/executor/string/AbstractAppendMemoryIntegrationTest.java
@@ -16,6 +16,7 @@ package org.apache.geode.redis.internal.commands.executor.string;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
+import java.time.Duration;
 import java.util.Map;
 
 import org.junit.After;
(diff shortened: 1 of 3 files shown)


nonbinaryprogrammer (approved):
nice documentation

DonalEvans (approved):

geode-for-redis/src/main/java/org/apache/geode/redis/internal/executor/server/InfoExecutor.java
```diff
@@ -127,13 +127,34 @@ private String getClientsSection(ExecutionHandlerContext context) {
...
 
+  /**
+   * Redis' fragmentation ratio is calculated from process_rss / used_memory. Essentially this is
+   * the ratio of physical memory being used vs. the amount of virtual memory the process
+   * requires. So, effectively an indication of memory pressure. If this number goes below 1.0 it
+   * would indicate that the process has started to swap memory which is bad.
+   * <p/>
+   * In a similar sense, the calculation for fragmentation here is a ratio of the amount of memory
+   * available to the partitioned region (Java heap) vs. the amount of memory consumed by data in
+   * the region. This ratio can only approach 1.0 and cannot go lower. However, the closer to 1.0
```
dschneider-pivotal, line 138: Actually, with the way you have implemented it, it can go below 1.0. This is because pr.getLocalMaxMemory is by default something like 90% of the jvm's max heap. But geode does not enforce this limit in any way. So we can add more data than that limit to the pr. 
Also, because we use heap memory for things that are not stored in the pr (think pubsub), I think it would be better for our "usedmemory" not to be limited to the getDataStore().currentAllocatedMemory(). 
For this fragmentation ratio I think we should ask the jvm for the max heap (Runtime.maxMemory), and then divide it by how much memory the jvm is using (Runtime.maxMemory()-Runtime.freeMemory()). I would not use Runtime.totalMemory because I have seen jvms (Azul) that set max and total the same but do have a meaningful freeMemory value. 
I wrote up another ticket discussing all this a couple of weeks ago (check the misc section). You don't need to do all that for this ticket. You could just focus on the fragmentation ratio.

dschneider-pivotal (approved):
I think you are also fixing GEODE-9800 with this PR so you could append the description of GEODE-9800 to GEODE-9669 and then mark it as a duplicate.

dschneider-pivotal:
@jdeppe-pivotal I wonder if the append test is failing because the jvm's free memory is not updated immediately? You could try having the test use geode awaitility when it is checking the usedmemory increases when an append is done.

geode-for-redis/src/integrationTest/java/org/apache/geode/redis/internal/commands/executor/string/AbstractAppendMemoryIntegrationTest.java
```diff
@@ -59,7 +60,8 @@ public void testAppend_actuallyIncreasesBucketSize() {
     info = RedisTestHelper.getInfo(jedis);
     Long finalMemValue = Long.valueOf(info.get("used_memory"));
 
-    assertThat(finalMemValue).isGreaterThan(previousMemValue);
+    GeodeAwaitility.await().atMost(Duration.ofSeconds(20))
+        .untilAsserted(() -> assertThat(finalMemValue).isGreaterThan(previousMemValue));
```
upthewaterspout, line 64: This test is going to be flaky some of the time. If you happened to sample `used_memory` at a time just before we do a GC, it may take a long time or we may never sample a value of used memory that is less than that value.

upthewaterspout (commented):
I'm ok with reporting a non-hard coded value here, but I'm not sure if this number will be meaningful in all cases - it really depends on the garbage collector someone is using.

PR #7133, opened 2021-11-19
GEODE-9638: Make test compiler not use files

demery-pivotal:
Fix `DeployedJarTest`, which intermittently throws
`DirectoryNotEmptyException` on Windows.

PROBLEM

`DeployedJarTest` uses Geode's custom Java compiler to create jar files.
When the compiler finishes compiling, it calls Apache commons
`FileUtils.deleteDirectory(dir)` to delete the temporary files it
created during compilation.

`FileUtils.deleteDirectory(dir)` first "cleans" `dir` by recursively
removing all files from it, then deletes `dir` itself. On Windows,
deleting `dir` sometimes throws `DirectoryNotEmptyException` even though
`FileUtils` has cleaned the directory. I do not know the cause of this.

SOLUTION

Change our custom Java compiler to create temporary files in memory
instead of on the file system.  That way there are no file system files
to delete.


agingade (approved):
Please ignore, if you have seen the below comment or already tried it...
From the net:

On Windows, having to delete twice is somewhat to be expected. Gradle watches project directories on Windows and Mac. On Windows, you can't delete a directory which is being watched. Gradle stops watching the directory when nothing is left to watch in the directory. So if you delete all the contents of a directory, then Gradle will stop watching it. It takes some time though for that to happen. Trying to delete the directory twice is an option which should work.

demery-pivotal:
> Please ignore, if you have seen the below comment or already tried it... From the net:
> 
> On Windows, having to delete twice is somewhat to be expected. Gradle watches project directories on Windows and Mac. On Windows, you can't delete a directory which is being watched. Gradle stops watching the directory when nothing is left to watch in the directory. So if you delete all the contents of a directory, then Gradle will stop watching it. It takes some time though for that to happen. Trying to delete the directory twice is an option which should work.

I haven't tried this, but I don't think Gradle's behavior is an issue for these particular tests. The compiler created a temporary, and Java creates those outside of the project directory. So Gradle wouldn't be watching those directories.

kirklund:
The only precheckin failure is NativeRedisRenameRedirectionsDUnitTest which matches [GEODE-9860](https://issues.apache.org/jira/browse/GEODE-9860)

kirklund (approved):

PR #7134, opened 2021-11-19
GEODE-9727: Updated COMMAND error message

Kris-10-0:
### For all changes:

### DIFF
--- a/geode-for-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/server/CommandIntegrationTest.java
+++ b/geode-for-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/server/CommandIntegrationTest.java
@@ -15,6 +15,7 @@
 
 package org.apache.geode.redis.internal.executor.server;
 
+import static org.apache.geode.redis.internal.RedisConstants.ERROR_UNKNOWN_COMMAND_COMMAND_SUBCOMMAND;
 import static org.apache.geode.test.dunit.rules.RedisClusterStartupRule.BIND_ADDRESS;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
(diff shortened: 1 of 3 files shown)


DonalEvans (approved):
Nice!

geode-for-redis/src/integrationTest/java/org/apache/geode/redis/internal/executor/server/CommandIntegrationTest.java
```diff
@@ -81,6 +87,32 @@ public void commandReturnsResultsMatchingNativeRedis() {
...
+  public void commandWithInvalidSubcommand_returnError() {
+    String invalidSubcommand = "fakeSubcommand";
+    RedisCodec<String, String> codec = StringCodec.UTF8;
+
+    CommandArgs<String, String> args =
+        new CommandArgs<>(codec).add(CommandType.COMMAND).add(invalidSubcommand);
+    assertThatThrownBy(
+        () -> radishClient.dispatch(CommandType.COMMAND, new NestedMultiOutput<>(codec), args))
+            .hasMessageContaining(
+                "ERR Unknown subcommand or wrong number of arguments for 'COMMAND'");
```
nonbinaryprogrammer, line 100: it would be good to use `ERROR_UNKNOWN_COMMAND_COMMAND_SUBCOMMAND` from `RedisConstants` so that if we change the error message it'll still work. I don't think we're really testing that the constant is right, just that the constant we expect gets returned.

geode-for-redis/src/main/java/org/apache/geode/redis/internal/executor/server/COMMANDCommandExecutor.java
```diff
@@ -34,9 +39,24 @@
  * CommandExecutor refers to the interface being implemented.
  */
 public class COMMANDCommandExecutor implements CommandExecutor {
+  private static final List<String> supportedSubcommands =
```
dschneider-pivotal, line 42: It would be good to mark this static with the @Immutable tag

PR #7143, opened 2021-11-24
GEODE-9808: Throw appropriate exception in AutoConnectionSourceImpl

DonalEvans:
### For all changes:

### DIFF
--- a/geode-core/src/integrationTest/java/org/apache/geode/cache/client/internal/AutoConnectionSourceImplJUnitTest.java
+++ b/geode-core/src/integrationTest/java/org/apache/geode/cache/client/internal/AutoConnectionSourceImplJUnitTest.java
@@ -18,10 +18,8 @@ package org.apache.geode.cache.client.internal;
 import static org.apache.geode.distributed.ConfigurationProperties.LOCATORS;
 import static org.apache.geode.distributed.ConfigurationProperties.MCAST_PORT;
 import static org.apache.geode.test.awaitility.GeodeAwaitility.await;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertNull;
-import static org.junit.Assert.fail;
+import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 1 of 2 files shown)


geode-core/src/integrationTest/java/org/apache/geode/cache/client/internal/AutoConnectionSourceImplJUnitTest.java
```diff
@@ -313,7 +296,7 @@ public void test_DiscoverLocators_whenOneLocatorWasShutdown() throws Exception {
       ArrayList<ServerLocation> locators = new ArrayList<>();
       locators.add(new ServerLocation(InetAddress.getLocalHost().getHostName(), secondPort));
       handler.nextLocatorListResponse = new LocatorListResponse(locators, false);
-      Thread.sleep(500);
+      Thread.sleep(1500);
```
kirklund, line 299: I think we should try to replace this sleep with a use of await. What background activity is happening during the call to sleep?

DonalEvans: Good call. I figured out what to await for here to allow the test to pass.

nabarunnag (approved):
Approved. 
Just for my edification, will there be a similar issue in run2 in UpdateLocatorListTask, where it calls updateLocatorList with the response but we never check the hasResults field and it simply starts accessing the members and calling methods on it.

DonalEvans:
> Approved. Just for my edification, will there be a similar issue in run2 in UpdateLocatorListTask, where it calls updateLocatorList with the response but we never check the hasResults field and it simply starts accessing the members and calling methods on it.

I don't believe so. If we get a response with no results, then the consequence will be that we'll just mark all our old known locators as bad and not add any new ones (because the response doesn't contain any). This behaviour is unaffected by the changes in this PR, and hasn't caused problems up till now, so I have to assume that it's okay.

Bill (approved):
Approved

